PR #4, opened 2015-06-16
Add figure and figcaption HTML elements to completion

kleinfreund:
This aims at adding the HTML `figure` and `figcaption` elements to the completion file. This is merely a draft, because I have some questions.

According to the Mozilla Developer Network, both elements ([figure](https://developer.mozilla.org/en/docs/Web/HTML/Element/figure#Attributes), [figcaption](https://developer.mozilla.org/en/docs/Web/HTML/Element/figcaption#Attributes)) only include the [global attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes)

I included the ones not marked as experimental but I left out all the event handler attributes, because that would add quite a mess and I'm not sure what's the policy here.

That said, there evolves a bigger question.

> Global attributes are attributes common to all HTML elements; they can be used on all elements, though the attributes may have no effect on some elements.

Should there not be a default list of attributes (the global ones) that are allowed on all elements? Then we would not need to include them for each element like it's done in the file.

Apart from this, I did not test this locally, yet. Is there something else I need to take care of?


HTML/html_completions.py
```diff
@@ -365,4 +367,3 @@ def test_expand_tag_attributes(self):
         # should give just one completion, and suppress word based completion
         self.assertEqual(completion_list, [('td.class', '<td class="class">$1</td>$0')])
         self.assertEqual(flags, sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS)
-
```
wesbos: Looks good - might want to keep this in - yo can set `"ensure_newline_at_eof_on_save": true,` in your prefs to prevent this from happening

kleinfreund: There were actually two newlines, that's just the superfluos one gone there. I have that enabled, because I don't want to mess with you vim users. :)

PR #13, opened 2015-06-18
Update Default.sublime-keymap

anyotherway:
// auto pair curly brackets
// current build does not space correctly, this fixes
// other consideration. after entering selector colon corect spacing added and ; but ; retype results ;; rather than st2 ignore and move past ;|


FichteFoll:
Do not like. 

This breaks the `{<enter>` work flow and leaves you with a leading and a trailing space. Furthermore, it's inconsistent with how sublime text does it in all other syntaxes.

sublimehq:
Closing for the reasons that @FichteFoll mentioned, however I've just pushed a change that should address the underlying concern

PR #15, opened 2015-06-18
Move brackets to the next line on __construct()

henriquemoody:


henriquemoody:
BTW, this is not a personal preference, according to [PSR-2](http://www.php-fig.org/psr/psr-2/):
> Opening braces for methods MUST go on the next line, and closing braces MUST go on the next line after the body.

PR #16, opened 2015-06-19
Bring HTML completions to the year 2015 :rocket:

kleinfreund:
**Changes made to html_completions.py**

Changes made to the file are mainly based upon the [HTML element reference](https://developer.mozilla.org/en/docs/Web/HTML/Element) from the Mozilla Developer Network. I did not make changes to tags and/or attributes that are marked as *obsolete* (I did neither remove any obsolete attributes nor add missing ones).

I tested whether the completion works for all tags inside `normal_tags` and it does.

- Adding new/missing tags: `main`, `content`, `data`, `datalist`, `details`, `element`, `embed`, `output`, `picture` (:rocket:), `progress`, `rp`, `rt`, `rtc`, `ruby`, `shadow`, `summary` and `video`
- Adding tags to `get_tag_to_attributes` map (so they get the global attributes)
- Adding HTML5 attributes (`download`, `media`, `ping`, etc.) if applicable
- Added special-cased completions for `audio`, `progress`, `source`, `track` and `video`
- consistent use of quotes (e.g. `''` only, escaped quotes `\"` were not touched as they’re used as actual output)
- Replacing `if` with `elif` in `get_attribute_completions`, because the conditions are exclusive

**Stats:**

|       | Original File | New File |
| ----- | -------------:| --------:|
| LoC   | 370           | 485      |
| Chars | 24864         | 20809    |

**Internal tests fail**

Note that I tried to run the internal test cases within that file by executing the following:

```python
import HTML.html_completions
HTML.html_completions.Unittest.run()
```

The tests fail with this:

```bash
  File "<string>", line 1, in <module>
  File "C:\Users\Philipp\dev\Sublime Text Build 3092 x64\Data\Packages\HTML\html_completions.py", line 420, in run
    test.test_simple_completion()
  File "C:\Users\Philipp\dev\Sublime Text Build 3092 x64\Data\Packages\HTML\html_completions.py", line 441, in test_simple_completion
    self.assertEqual(flags, sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS)
  File "./unittest/case.py", line 641, in assertEqual
  File "./unittest/case.py", line 634, in _baseAssertEqual
AssertionError: 0 != 3
reloading plugin HTML.html_completions
```

This does not look related to my changes. Running the same from within my installed version that does not have the changes results in the same.

---

Have a good day and let me know if there's something I need to adjust.

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -12,76 +12,172 @@ def match(rex, str):
 def make_completion(tag):
     # make it look like
     # ("table\tTag", "table>$1</table>"),
-    return (tag + "\tTag", tag + ">$0</" + tag + '>')
+    return (tag + '\tTag', tag + '>$0</' + tag + '>')
 
 def get_tag_to_attributes():
-    # maps tags to all available attributes
-    return (
-        {'a' : ['accesskey', 'charset', 'class', 'coords', 'dir', 'href', 'hreflang', 'id', 'lang', 'name', 'onblur', 'onclick', 'ondblclick', 'onfocus', 'onkeydown', 'onkeypress', 'onkeyup', 'onmousedown', 'onmousemove', 'onmouseout', 'onmouseover', 'onmouseup', 'rel', 'rev', 'shape', 'style', 'tabindex', 'target', 'title', 'type'],
-        'applet' : ['align', 'alt', 'archive', 'class', 'code', 'codebase', 'height', 'hspace', 'id', 'name', 'object', 'style', 'title', 'vspace', 'width'],
-        'area' : ['accesskey', 'alt', 'class', 'coords', 'dir', 'href', 'id', 'lang', 'nohref', 'onblur', 'onclick', 'ondblclick', 'onfocus', 'onkeydown', 'onkeypress', 'onkeyup', 'onmousedown', 'onmousemove', 'onmouseout', 'onmouseover', 'onmouseup', 'shape', 'style', 'tabindex', 'target', 'title'],
+    """
+    Returns a dictionary with attributes accociated to tags
+    This assumes that all tags can have global attributes as per MDN:
+    https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes
+    """
+
+    # Map tags to specific attributes applicable for that tag
+    tag_dict = {
+        'a' : ['charset', 'coords', 'download', 'href', 'hreflang', 'media', 'name', 'ping', 'rel', 'rev', 'shape', 'target', 'type'],
+        'abbr' : ['title'],
+        'address' : [],
+        'applet' : ['align', 'alt', 'archive', 'code', 'codebase', 'height', 'hspace', 'name', 'object', 'vspace', 'width'],
+        'area' : ['alt', 'coords', 'download', 'href', 'hreflang', 'media', 'nohref', 'rel', 'shape', 'target'],
+        'article' : [],
+        'aside' : [],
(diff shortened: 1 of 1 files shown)


HTML/html_completions.py
```diff
@@ -156,26 +251,46 @@ def get_completions(self, view, prefix, locations, is_inside_tag):
...
+            ('iframe\tTag', 'iframe src=\"$1\">$0</iframe>'),
+            ('img\tTag', 'img src=\"$1\">'),
+            ('link\tTag', 'link rel=\"stylesheet\" type=\"text/css\" href=\"$1\">'),
+            ('param\tTag', 'param name=\"$1\" value=\"$2\">'),
+            ('progress\tTag', 'progress value=\"$1\" max=\"$2\">'),
+            ('script\tTag', 'script type=\"${1:text/javascript}\">$0</script>'),
+            ('source\tTag', 'source src=\"$1\" type=\"$2\">'),
+            ('style\tTag', 'style type=\"${1:text/css}\">$0</style>'),
+            ('track\tTag', 'track kind=\"$1\" src=\"$2\">'),
+            ('video\tTag', 'video src=\"$1\">$0</video>')
```
FichteFoll: I just want to mention that these completions are inconsistent, which was not in the scope of your changes, but anyway.

kleinfreund: The completions for the `link` and `img` tag omit the closing tag and were there before. They don’t end in `/>` so I tried to stick with that.

Do you want me to add closing slashes to void tags like ``?

Also as far as I can tell, the `$0` is only used when tags are actually closed and the last tabbing position should be inside the tag pair. However `track`, `source` and `progress` omit the end tag as well.

wesbos: Ending self-closing tags with `/>` sort of died with HTML5 as it is valid to to just do ``

Some devs are still in the self close habbit, but its not necessary.

wesbos:
I can't speak of the Python code, but the logic behind the HTML5 standard tags is sound, so :+1:

HTML/html_completions.py
FichteFoll: I admit that I don't know enough about the specifications of all the new HTML5 tags and what you can actually place in them (it seems weird for me to have content for a video/audio tag for example), but I do know that not closing XML tags fucks with my mind (justifiably).

The $0 anchor would just be for "continue typing here, once all parameters have been filled" - and this is missing for img, link, param, progress, source and track.

wesbos: Yeah - it was a standard back when we validated our sites for be HTML 4.1 Strict or whatever - totally not necessary now. Feels good :)

kleinfreund: > The $0 anchor would just be for "continue typing here, once all parameters have been filled"

@FichteFoll As far as I observed it, having the last caret position after the complete tag is baked in somewhere. What you want is type `img`, hit <kbd>tab</kbd>, have the caret after `src="`, hit <kbd>tab</kbd> again and have the caret after ``, right?

If so, it's already working like that without the need to explicitly add `$0`.

FichteFoll: @kleinfreund Oh yeah, it is. Sorry, my bad. I should have known this. :cold_sweat:

kleinfreund: In my opinion, the time has long come that tags like `img` should be used as @wesbos already said. There is no longer a necessity for it and no one stops you if you have that habit. Boils down to adding one/two less characters. Not a big deal.

PR #21, opened 2015-06-20
[Makefile] Replace by a converted version of Makefile Improved

TheClams:
Remove old syntax style and replace it by a converted version of the plugin "Makefile Improved" by Kay-Uwe (Kiwi) Lorenz

This includes a few personal improvement:
- Fix ifeq/ifneq missing highlight
- Support more shell function (echo, setenv, ...)
- Highlight special target like .PHONY
- Allow variable declaration in the dependency list
- Support special variable with double $$
- And more minor details

This is far from perfect, but in my opinion, this is still a huge improvement compare to the original which highlights almost nothing

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -48,6 +48,9 @@ contexts:
     - match: \b(global)\b
       captures:
         1: storage.modifier.global.python
+    - match: \b(nonlocal)\b
+      captures:
+        1: storage.modifier.nonlocal.python        
     - match: \b(?:(import)|(from))\b
       captures:
         1: keyword.control.import.python


TheClams:
Sorry for the unrelated commit, I'll create a different branch next time ...

FichteFoll:
Two suggestions:

1. Split the conversion into a separate commit and then commit your changes onto that. This was it's way easier to see what you actually changed.
2. Add syntax test files where applicable. The C syntax is really huge, but the makefile is fairly small and a syntax test for at least your additions would be useful for verifying its correctness. Reviewing syntax definition is hard because of how complex they can be at times.

TheClams:
Sure. I have pushed the change for C in a different branch/pull request with two commit. 
For the syntax test of Makefile, I'll need to read more how it works.

Issue #8, opened 2015-06-17
Bundle snippets in a "Snippets" subdirectory?


FichteFoll:
Some packages contain quite a few snippets, which creates a lot of noise in the package folder. I propose to move all snippets, or if they exceed a certain threshold (e.g. 4), to a new subdirectory called "Snippets".

gerardroche:
Move them into their own package. Completions too. This allows the user to install alternative ones.

FichteFoll:
How about being able to ignore only a subset of a package in a new setting? With that you could,  for example, all .Sublime-snippet files with a pattern like `"ignore_package_files": {"Java": ["*.sublime-snippet"]}`.

That way it is more flexible and powerful and could be applied to any package.

jbrooksuk:
I think all snippets should belong in a `Snippets` directory, so I made a PR for it.

aziz:
Why should it only be limited to snippets? I always disliked the lack of file organizations in Sublime packages. 
Textmate 2 took a very structured approach
![](http://cl.ly/image/1P0R2W0s063D/Screen%20Shot%202015-06-18%20at%208.44.39%20PM.png)

Something in between would be great.

jbrooksuk:
I noticed that after moving the snippets the root directory was pretty sparse, so I don't think that there is much need to go further.

FichteFoll:
Besides `.tmPreferences`, all other resource types are fairly limited for each package (menus, key binds, command palette, ...). And even these rarely go beyond 5, so I would base this on individual judgement.

Additionally, things that can be added indefinitely (syntax definitions, themes, color schemes) are usually the core part of a package, and packages usually aim to not be bloated with 100s of color schemes. 

Anyway, for my own packages I adapted a structure to move all package resource files except plugins (which must be in the package root) in a "Package" subdirectory. This directory is then structured deeper into "Snippets" or "Syntax Definitions", if there are more files than a certain threshold. This has the advantage of having all repository-relevant files such as readme, license, files for Package Control are easy to catch in the root. Since we don't have any of these files here, I doubt this structure is necessary.

jbrooksuk:
IMHO all default packages should follow a convention. Not base it on some arbitrary amount of files before they get moved into a sub-directory. Basically, all or nothing.

gerardroche:
[...]

PR #22, opened 2015-06-20
Minor fixes to PHP syntax

jobedom:
I am the creator of the PHP Modern package. I've decided to deprecate it and add any enhancement it contains to the official PHP package, so I'll be sending PRs for any past and future improvements.

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -442,7 +442,7 @@ contexts:
       captures:
         1: storage.type.trait.php
         2: entity.name.type.trait.php
-    - match: \s*\b(break|c(ase|ontinue)|d(e(clare|fault)|ie|o)|e(lse(if)?|nd(declare|for(each)?|if|switch|while)|xit)|for(each)?|if|return|switch|use|while)\b
+    - match: \s*\b(break|c(ase|ontinue)|d(e(clare|fault)|ie|o)|e(lse(if)?|nd(declare|for(each)?|if|switch|while)|xit)|for(each)?|if|return|yield|switch|use|while)\b
       scope: keyword.control.php
     - match: (?i)\b((?:require|include)(?:_once)?)\b\s*
       captures:
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -460,7 +460,7 @@ contexts:
             3: punctuation.definition.variable.php
           pop: true
         - include: namespace
-    - match: \b(catch|try|throw|exception)\b
+    - match: \b(catch|try|throw|exception|finally)\b
       scope: keyword.control.exception.php
     - match: (?i)\b(function)\s*(&\s*)?(?=\()
       captures:
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -386,6 +386,8 @@ contexts:
                   pop: true
                 - match: \\
                   scope: punctuation.separator.inheritance.php
+                - match: '(?i)([a-z_][a-z_0-9]*)?(?=[^a-z0-9_\\])'
+                  scope: entity.other.inherited-class.php
         - match: \s*,\s*
     - match: '(?i)^\s*(abstract|final)?\s*(class)\s+([a-z0-9_]+)\s*'
       captures:
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -798,75 +798,97 @@ contexts:
               scope: constant.character.escape.php
         - match: "[$^+*]"
           scope: keyword.operator.regexp.php
-  sql-string-double-quoted:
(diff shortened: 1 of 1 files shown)


gerardroche:
Some of these are not minor fixes. Without extensive tests I don't see how they can be accepted.

For example trait should match `meta.trait.php storage.type.trait.php`. See https://github.com/gerardroche/sublime-php-grammar

jobedom:
I'll be happy to provide any needed tests. What is the procedure or actual mechanics of testing a grammar?

gerardroche:
http://www.sublimetext.com/docs/3/syntax.html

Also check out the tests in php-grammar https://github.com/gerardroche/sublime-php-grammar/tree/master/test/syntax

gerardroche:
break it down into smaller commits. The finally and trait keywords don't need  tests.

jobedom:
Great! Thanks a lot, gerardroche. I'll create needed tests.

Is your package sublime-php-grammar an unofficial one?

gerardroche:
Yes it's unofficial.

jobedom:
So I'm guessing it's not an easy path asking for integration into the official PHP package. I'm checking yours and maybe it has all I need.

gerardroche:
It's a tough one. I think most if not all the changes in php-grammar will be accepted. Even if a few small things are not accepted I would prefer to contribute to the default and run with that.

[Splitting the packages into their own repositories](https://github.com/sublimehq/Packages/issues/8) and the ability to [disable/override default completions and snippets](https://github.com/sublimehq/Packages/issues/8#issuecomment-113080703). These are prerequisite to porting php-grammar to the default package.

jobedom:
I get it. Thanks again for all the info.

rtheunissen:
Thanks for taking the time to do this @jobedom. I'm on the edge of switching to another editor simply because the PHP highlighting is broken, so I very much appreciate the effort to fix it. :+1:

Gerst20051:
:+1:

vinkla:
:+1:

jbrooksuk:
Yeah, I like whatever option improves PHP highlighting.

svivian:
Hey, just want to note a couple of issues with the current PHP syntax highlighter. If you've already fixed these then just ignore me :)

First, type hints and default arguments don't play nice. In the following code the `[]` gets marked in dark red (indicating an error):

    <?php
    function test(array $arr=[]) {
        // code
    }

Each works individually (i.e. `array $arr` or `$arr=[]`) but not together. Same goes for any type hint.

On a similar topic, since PHP7 beta was just released I guess you might want to add string/int/float/bool as keywords and ensure they work with type hints.

Finally, when using DocBlock comments I regularly get the "dark red background of death". As soon as I type `/**` everything below goes red because the following lines don't start with `*` (yet). Is it within the scope of this package to automatically format DocBlocks, or is that elsewhere? So when you type `/**` and press Enter, you get something like the following, where `|` is the cursor:

    /**
     * |
     */

gerardroche:
@svivian array type hints issue duplicate here https://github.com/SublimeTextIssues/DefaultPackages/issues/17 should be fixed in php-grammar https://github.com/gerardroche/sublime-php-grammar/commit/388eae9ffa8773dd59dc995a5f5baa7dbfdfafed 

@svivian php7 features would be good.

@svivian Try docblockr https://github.com/spadgos/sublime-jsdocs Perhaps the docblock feature you mention should be part of a php package or at least the basic feature. But the docblockr package covers all languages not just php.

svivian:
@gerardroche great, thanks for the info.

DocBlockr has the functionality I suggested which is great, but there is the still the issue with the dark red background. When I type `/**` a huge chunk of code goes red momentarily which is rather ugly.

Is that really necessary? It's perfectly valid PHP to have a comment starting `/**` with anything inside it.

gerardroche:
@svivian open an issue. I kinda agree with you. While it might be an invalid phpdoc it's not an invalid comment. Perhaps the scope could be changed to something else to indicate it's an invalid phpdoc but not an invalid comment.

rtheunissen:
@jobedom What's your status regarding this PR? I'm still using PHP Modern because the official package is too far behind. PHP 7 is ready and ST doesn't even support 5.4 properly. Thanks for your work on PHP Modern.

@gerardroche Do you know when we might expect some action here? I find it ridiculous that something like syntax highlighting is broken on arguably the most popular text editor for a widely-used language, 2 months after a partial patch was put forward. If the procedure to get these merged is that difficult, I don't see why these were made open in the first place.

gerardroche:
@rtheunissen The ST developer has disappeared again. PHP Grammar is now available on package control https://packagecontrol.io/packages/php-grammar.

wbond:
All of these features have been implemented in the master branch, along with tests.

In addition, the PHP syntax has improved performance, from removing regex constructs incompatible with the new Sublime Text regex engine.

Please take the time to check out your PHP with the new version and report any bugs you find, or features from PHP7 that are missing. https://github.com/sublimehq/Packages#installation has instructions on how to use the latest version of the syntax on your machine.

Issues should be filed at https://github.com/sublimehq/Packages/issues, and please be sure to include the a snippet of code that exposes the bug.

PR #27, opened 2015-06-26
[YAML] Fix end of block detection

TheClams:
Use first non-empty line to extract indentation level
Empty line inside a block do not pop out of the block

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -18,8 +18,12 @@ contexts:
         5: punctuation.separator.key-value.yaml
       push:
         - meta_scope: string.unquoted.block.yaml
-        - match: ^(?!^\1)|^(?=\1(-|\w+\s*:)|#)
-          pop: true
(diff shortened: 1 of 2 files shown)


sublimehq:
Looks good, this is a vast improvement on the current implementation. However, the hyphen handling appears incorrect. For example, the hyphen here should be treated as part of the string:

```
unquoted_block2: |-
    Start of the block
    Still in block
    - also: part of the string
```

Test link: http://yaml-online-parser.appspot.com/?yaml=unquoted_block2%3A+%7C-%0A++++Start+of+the+block%0A++++Still+in+block%0A++++-+also%3A+part+of+the+string%0A&type=json

TheClams:
Yes, I was not sure about this part (the YAML spec is really not easy to read ...), but since there was some kind of detection for this pattern in the original syntax, I did not want to removed it. Should be good now.

sublimehq:
Excellent, thanks for that. Fully agree on the topic of the YAML spec!

PR #28, opened 2015-06-27
[C] Collection of fixes for preprocessor, operator, namespace, ...

TheClams:
Highlight operator to be more coherent with other packages like python.
TODO: maybe add distinction for pointer/address operator


sublimehq:
I'll need to think about the "Add highlight for operators" commit, as it also highlights some tokens in type declarations that aren't operators.

It's fine in most cases, but IMHO it makes some type declarations, such as "const std::vector<const char *> &" unsightly in conjunction with Monokai.

TheClams:
Yes, I was not sure how to handle the pointer/address declaration/access.
Maybe I can try to at least identify only the & and * when used as boolean/mathematical operator ?

TheClams:
Trying to make the difference between an address and a boolean operator seems quite difficult, so in my next commit I will separate & and * from the other operators, this way it's easy for you to comment them if you don't like it. I personally thinks it's fine like that and the issues is more with Monokai using really limited scope and not making a difference between a keyword and an operator (but clearly all of this is really subjective ...)

TheClams:
I merged the latest contribution into this branch and added a fix for initilializer list in constructor.
I tried to split my commits to show that the biggest changes are actually just moving some pattern into a specific context to ease the fix. 
Please tell me if you need more support to merge this.

TheClams:
@wbond : When you have some time, I'd like to have some feedback on this PR: it fixes a significant number of issue listed on the [SublimeTextIssues](https://github.com/SublimeTextIssues/DefaultPackages/issues?q=is%3Aopen+is%3Aissue+label%3A%22C%3A+C%2FC%2B%2B%22) github repo, and one or two on this repo ( #201 and partially #177 ). 
The added syntax_test_c.c file shows clearly how broken the current syntax is around macro.

wbond:
Yes, I am planning on reviewing this in the near future.

C++/C.sublime-syntax
```diff
@@ -255,10 +157,27 @@ contexts:
...
+          captures:
+            1: meta.preprocessor.c
+            2: keyword.control.import.if.c
           pop: true
         - include: disabled
         - include: pragma-mark
+  operators:
+    - match: <\=|>\=|\=\=|<|>|\!\=
+      scope: keyword.operator.comparison.c
+    - match: \+\=|-\=|\*\=|/\=|//\=|%\=|&\=|\|\=|\^\=|>>\=|<<\=|\*\*\=
```
jbw3: @TheClams I don't think the operator highlighting is quite correct. I have a few suggestions:

1. You are highlighting `**`, `**=`, `//`, and `//=` which are neither C nor C++ operators (`//` starts a comment)
2. Since the comparison operators are first, I think they will prevent some of the arithmetic operators from being tagged properly. For example, I think the `<<` operator will be tagged as two `<` operators. I haven't tested this, though, so I don't know for sure. It would be a good idea to add some of these operators to the test file to look into this
3. Could you highlight the `!`, `&&`, and `||` operators as well? Thanks!

wbond:
There are definitely a bunch of nice fixes in here! My comments here are going to disproportionately cover issues that I think we still need to resolve.

Recently I worked a bunch on the Rust syntax, which shares a number of contexts. I think that continuing to harmonize these two syntaxes to utilize similar behavior will be good.

 1. We are going to need to explicitly handle generics syntax. Right now `<` and `>` are highlighted as comparison operators. In Rust I highlighted `<` as `punctuation.definition.generic.begin` and `>` as similar, but with `end`.
 2. `::` is being highlighted as two ternary operators. Additionally, using the default color scheme of Monokai, highlighting `::` as operators really kind of breaks up identifiers so they feel more like two variables being operated on. This also feels true for `->`. I'm not sure if we use a scope such as `punctuation.resolution` for `->`, `::` and `.`, thus allowing color schemes to give them a different flavor than things like `==`, `>=`, `/`, etc. This may be a change we could propagate to other languages also.
 3. There is the troubling issue of `*` and `&` in type definitions where they are highlighted as arithmetic operators. Without having looked much at the current implementation, I'm thinking we may have to more closely track the current state to try and properly parse type information to disambiguate it from operations. I don't know if this is all that possible.

FichteFoll:
Highlighting `->`, `::` and `.` as `punctuation.resolution` sounds like a good idea to me. `punctuation.accessor` (as used in JavaScript) would also be an option however. I agree that there should only be one of the two.

If you do not, I'll make sure to document these new scopes, especially the punctuation ones, eventually. Hopefully some tim in April.

TheClams:
1. Should be easy to add (I can do that later this week)
2. I noticed in other syntaxes that `punctuation.accessor` was used (And i used it here but only for `.`) I can extend that to `->` and `::` I also think that Monokai theme would gain a lot by having a color defined for keyword.operator : this is used in many syntaxes and when using Monokai having all operator highlighted as simple keyword does not look great at all (but I guess this is personnal taste)
3. I'm open to suggestion but I feel this would be really hard to do.

TheClams:
I implemented point 1 and 2. By the way in javascript the scope is not punctuation.accessor but keyword.operator.accessor : I think the puncuation scope is better so I used it, but it might be a good idea to update the Javascript one.

wbond:
@TheClams I wanted to give you a heads up that I've been working on this and the C/C++ syntaxes in general.

As I'm sure you well know, it is difficult to properly parse C, one line at a time. You have to guess with a bunch of heuristics to handle multi-line function definitions.

Many of these changes are huge improvements, but I'm also seeing a good number of false positives in terms of random identifiers in the middle of code being highlighted as function definitions. This then results in highlighting various curly braces and invalid/stray.

I am still trying to work on refactoring various things to improve the overall correctness of detecting functions, while keeping performance up. I've found the SQLite and Python source code to be pretty useful in finding various edge cases that break things.

TheClams:
Yes in some of my modification I removed some false positive on function definition but clearly there is still some cases where it could trigger (I think there is still too many parts where the top level language is included, while we should be more selective).
Anyway I won't touch this syntax anymore for the moment to avoid creating conflict with your work. Thanks for the heads up :)

wbond:
Ok, I did a big rewrite on top of this. Currently it is hosted at https://github.com/sublimehq/Packages/tree/c-rewrite. If you could try it out and report any issues, I'd love to get it in the next dev build.

At this moment, Obj C and Obj C++ are broken since they were written in such a way that `$top_level_main` was required, however that is no longer used. My next step is to get them fixes up, which should hopefully be relatively simple compared to all of the funkiness of the preprocessor and C++.

In general, C++ and C are almost completely separate now, due to the drastic differences. I did reuse some contexts that only highlight individual tokens, but anything that pushes into different scope isn't reusable across syntaxes.

TheClams:
I had a first quick look, checking some file that created problem with the initial syntax and it seems mainly ok.
I just found two inconsistent scoping around preprocessor:
- missing scope meta.preprocessor.* on #endif statement
- The # character has no scope for if/else/endif while it gets the scope keyword.control.* for #define, #error, ... 

I'll use this new syntax at work, and i'll report here if I found any new issues.

And one remark: I notice you are using variable.function for user defined function. I think the idea to give a different scope between standard function and user one is really good but this is not a very common scope in the existing color scheme and I can already see the ten incoming thread on the forum "C function not highlighted". It would be a good idea to add this scope to the default color schemes: looking on how it went with the Javascript change, it seems a large part of ST users are not very familiar with customizing their color scheme :P

FichteFoll:
Just FYI, we already went through the `varable.function` stuff with the JavaScript rewrite (and previously the integration of JavaScript Next). Most complaints were about missing highlighting of the dollar symbol though.

wbond:
@TheClams The issues you identified, plus a couple from #268 are resolved in the `c-rewrite` branch.

TheClams:
Great, it's looking good.
I worked on some C file today and did not notice any more problem :)

PR #33, opened 2015-06-29
Bring CSS completions to the year 2015 :rocket:

kleinfreund:
- Added ~81 new CSS properties (including `animation`, `columns`, `flex`, `transform`, `transition`, etc.)
- Changes to `CSS.sublime-syntax`:
  - Added missing/new properties to `rule-list`
  - Added missing/new property values to`property-values`
  - Added missing/new HTML tags to `selector`
  - Added viewport-percentage units `vw`, `vh`, `vmin`, `vmax` (and some other missing units like `turn`, `dppx`, `Hz`, `kHz`) to `numeric-values`
- Rewritten `parse_css_data()` so it parses a dictionary, not a huge string (which was essentially converted to a dictionary in the process anyway)
  - Removing some weird duplicates (identical)
  - Added global property values (`all`, `inherit`, `initial`, `unset`) to all properties
- Additional `COMMON_VALUES` for property values which are used multiple times in the dictionary

I tested highlighting and completions for all CSS property names and a couple of properties two (but not __all__ of them – it's a lot of them).

**Stats:**

| css_completions.py | Original      | New      |
| ------------------ | -------------:| --------:|
| LoC                | 195           | 412      |
| Chars              | 9283          | 17522    |

Ask me, if there is something unclear. Have a nice day.


CSS/CSS.sublime-syntax
```diff
@@ -168,12 +168,12 @@ contexts:
     - match: |-
         (?x)
             (?:-|\+)?(?:(?:[0-9]+(?:\.[0-9]+)?)|(?:\.[0-9]+))
-            ((?:px|pt|ch|cm|mm|in|r?em|ex|pc|deg|g?rad|dpi|dpcm|s)\b|%)?
+            ((?:px|pt|ch|cm|mm|in|r?em|ex|pc|deg|g?rad|turn|dpi|dpcm|dppx|s|v(w|h|min|max)|k?Hz)\b|%)?
```
wesbos: :+1:  Glad to see these units added

CSS/css_completions.py
```diff
@@ -2,174 +2,391 @@
...
+    'display': [
+        'none', 'inline', 'block', 'contents', 'list-item', 'inline-block',
+        'inline-table', 'table', 'table-cell', 'table-column',
+        'table-column-group', 'table-footer-group', 'table-header-group',
+        'table-row', 'table-row-group', 'flex', 'inline-flex', 'grid',
+        'inline-grid', 'ruby', 'ruby-base', 'ruby-text', 'ruby-base-container',
+        'ruby-text-container', 'run-in'
+    ],
+    'empty-cells': ['show', 'hide'],
+    'flex': ['none'],
```
wesbos: Perhaps this should be `<number>`?

wesbos:
Gave the CSS standards a look over and this is looking really good! 

I made a comment inline about flex

The other one that I think might be missing is `filter` - Although it's not 100% done yet, support is pretty good. 

https://developer.mozilla.org/en/docs/Web/CSS/filter

CSS/css_completions.py
kleinfreund: As of [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) the value setting `flex: 0 0 auto;` is `flex: none;` so I went with that. What do you think?

kleinfreund:
@wesbos Thanks for taking a look at this.

I actually thought about adding that as well for the same reason, but it seems that I forgot it along the lines. Maybe there are some other _experimental_ features worth adding.

I prepare a commit.

CSS/css_completions.py
wesbos: That is a the default flex, which is very rarely set because it's the default. 

When using the flex shorthand, It's more common to do `flex:1` or `flex: 1 3 400px`

kleinfreund: I see. I add `<number>`. It should be there.

kleinfreund:
Ah, yikes. I need to solve the merge conflicts because of [Updating for CSS3](https://github.com/sublimehq/Packages/commit/a40ec7128b40d31837068cc2917c1984ec525c84). Gonna take a while.

kleinfreund:
So I basically didn't notice the changes made to `CSS.sublime-syntax` and had quite come conflicts. I'm not sure whether I managed to bring this pr back in shape completely. This is my first time that something like this happened.

Long story short: I should've checked out the files I was editing in the remote repository. I did not do that.

Now there are some lost changes while rebasing and solving the merge conflicts. After this pr is approved/rejected, I'll have a look at the file and submit a new pr.

kleinfreund:
This should be it now. I have tested every single new property plus possible values. I ended up putting all new properties in a CSS file, so it's easy to see whether the syntax highlighting works or not. I also fixed some small mistakes in both files (superfluous `s` here, missing `pipe` there).

sublimehq:
Looking great!

PR #34, opened 2015-06-29
Improved `if` and `function` Javascript snippets.

mrmartineau:
- removed unneeded semicolons after them
- improved whitespace

It always bugged me that there were needless semicolons after these snippets..

### DIFF
--- a/JavaScript/function.sublime-snippet
+++ b/JavaScript/function.sublime-snippet
@@ -1,7 +1,5 @@
 <snippet>
-    <content><![CDATA[function($1) {
-	${0:$TM_SELECTED_TEXT}
-}]]></content>
+    <content><![CDATA[function($1) { ${0:$TM_SELECTED_TEXT} }]]></content>
     <tabTrigger>f</tabTrigger>
     <scope>source.js</scope>
     <description>Anonymous Function</description>
--- a/JavaScript/function.sublime-snippet
+++ b/JavaScript/function.sublime-snippet
@@ -1,5 +1,7 @@
 <snippet>
-    <content><![CDATA[function($1) {${0:$TM_SELECTED_TEXT}};]]></content>
+    <content><![CDATA[function($1) {
+	${0:$TM_SELECTED_TEXT}
+}]]></content>
     <tabTrigger>f</tabTrigger>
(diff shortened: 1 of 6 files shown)


JavaScript/function.sublime-snippet
```diff
@@ -1,5 +1,7 @@
 <snippet>
-    <content><![CDATA[function($1) {${0:$TM_SELECTED_TEXT}};]]></content>
+    <content><![CDATA[function($1) {
+	${0:$TM_SELECTED_TEXT}
+}]]></content>
```
FichteFoll: I don't think it's a good idea to always make this multi-line. You can easily turn the single-line anonymous function into a multi-line one by pressing enter/return once, but not the other way around.

FichteFoll:
Closes https://github.com/SublimeTextIssues/DefaultPackages/issues/37.

sublimehq:
I agree with @FichteFoll here: the changes in general are great, but adding the linebreaks is likely a step in the wrong direction

mrmartineau:
Your points are valid, how about I resubmit without the line breaks?

FichteFoll:
I'd like to mention again that the for snippets are missing in this pull request and still have semicolons.

mrmartineau:
@FichteFoll I didn't notice your edit about the `for` snippets & I didn't notice that they had semicolons as well.. I will submit another PR later today

mrmartineau:
removed the semicolons for `for` loops now :)

poke:
You should squash those changes and repush them (`--force`) so that the intermediary states are not included.

Also, the change in `function-(fun).sublime-snippet` is introducing a broken syntax as there is a line comment. The snippet should stay multi-line as it currently is. While I would personally welcome multi-line syntax for all the snippets, I accept that the consensus seems to be that they should stay one-line ones. But the ones that are currently multi-line shouldn’t be changed to one-liners.

Finally, is there a reason for the changed placeholder order in `if-___-else.sublime-snippet`? It seems a bit weird to first insert the else part…

mrmartineau:
I have created a new, better PR #43 which should be used instead. Thanks for your input, it is nice to know you all care about the details.

PR #37, opened 2015-07-01
Fix "fi" highlight in ShellScript and add some highlights.

jfcherng:
Left / Right = Before / After


ShellScript/Shell-Unix-Generic.sublime-syntax
```diff
@@ -449,7 +453,7 @@ contexts:
           pop: true
         - include: main
   pipeline:
-    - match: \b(time)\b
+    - match: '(?:^|\s)(a(?:[r]|wk)|b(?:[g]|zip)|c(?:[dp]|a[lt]|h(?:mod|own)|m([p]|ake))|d(?:[dfu]|c(?:_shell(-t|))|i(?:[r]|ff|rcmp))|expand|f(?:[g]|i(?:le|nd)|mt|usermount)|h(?:alt|ead|istory|ost(?:id|name|))|i([d]|n(dent|fo))|join|kill|l(?:[lns]|ast|ess|ocate)|m(?:[v]|a(?:[n]|il|ke)|kdir|o(?:re|unt))|news|p([sy]|asswd|erl|hp|ython[\d.]*)|r([m]|login|sh|who)|s(?:c(?:[p]|reen)|e[dq]|ftp|hl|ize|leep|ort|p(?:ell|lit)|sh(?:fs|))|t(?:[r]|a(?:[r]|il|lk)|e(?:[e]|lnet|st)|ime|ouch|ty)|u(?:mask|name|ptime|sers(?:add|))|w(?:[c]|ait|get|h(?:[o]|atis|ere|ich)|)|xz|zip)(?:$|\s)'
```
sublimehq: Unless there's a need for it, ```\b``` should be preferred over ``` (^|\s)``` (Likewise for ```(?:$|\s)```)

Also, and this is more of an FYI rather than something that I'm concerned about, but there's no performance benefit to doing ```a(r|wk)``` vs ```ar|awk``` in Sublime Text, and the latter is simpler to read and modify. Indeed, small character classes, especially with just a single character in them (e.g., ```[g]```), are a little slower than just writing out the alternatives.

jfcherng: I have expanded the regex by using commands from http://ss64.com/bash.
And add some common tools. (perl,python,php,java,jar,lua,git,svn,hg,autoconf,cmake,gcc,g++,clang,clang++,...)

ShellScript/Shell-Unix-Generic.sublime-syntax
```diff
@@ -418,6 +418,8 @@ contexts:
       push:
         - meta_scope: meta.scope.if-block.shell
         - match: \b(fi)\b
+          captures:
+            1: keyword.control.shell
```
aziz: Is there a reason why you're using captures here? Am I missing something? 
Cause I would simply use scope on it, like `scope: keyword.control.shell`.

jfcherng: Just because there are lots of captures which is used like this in the file...

ShellScript/Shell-Unix-Generic.sublime-syntax
```diff
@@ -449,7 +453,7 @@ contexts:
...
...on|fuser|g\+\+|gawk|gcc|getopts|git|grep|groupadd|groupdel|groupmd|groups|gzip|hash|head|help|hg|history|hostname|iconv|id|ifconfig|ifdown|ifup|import|insmod|install|jar|java|jobs|join|kill|killall|kmod|less|let|link|llvm|ln|local|locate|logname|loout|look|lpc|lpr|lprint|lprintd|lprintq|lprm|ls|lsod|lsof|mail|make|man|md5sum|mkdir|mkfifo|mkisofs|mknod|mmv|modinfo|modprobe|more|most|mount|mtools|mtr|mutt|mv|mysql|netstat|nice|nl|nohup|notify-se|nslookup|nstat|op|open|openssl|paswd|paste|pathchk|perl|php|ping|pkill|popd|pr|printcap|printenv|printf|ps|pushd|pv|pwd|py|python[\d.]*|quota|quotachec|quotactl|ram|rar|rcp|read|readarray|readonly|reboot|remsync|renae|renice|return|rev|rm|rmdir|rmmod|rsync|ruby|scp|screen|sdiff|sed|select|seq|set|sftp|sh|shift|shopt|shutdown|sleep|slocate|sort|source|split|sqlite3?|ssh|stat|strace|su|sudo|sum|susend|svn|sync|tail|tar|tcsh|tee|test|time|timeout|times|top|touch|tput|tr|tracerout|trap|true|tsort|tty|type|ulimit|umask|umount|unalias|uname|unexpand|uniq|units|unrar|unset|unshar|unil|updatedb|uptime|useradd|userdel|usermod|users|uudecode|uuencode|vmstat|wait|watch|wc|wget|whatis|whereis|which|while|who|whoami|write|xargs|xdg-open|xz|yum|zip|zsh)\b'
```
aziz: This wasn't a really good idea, now that I'm using this syntax my bash files are full of keywords. Not everything is a keyword. I want to differentiate between `shell keyword`, `shell builtin` and `external commands`.

jfcherng: I agree your concept.
I will separate them out.

jfcherng: Just having some question doing that.

For example, what is the category that `pwd` belongs to?
It's now classified as an `builtin` but on my machine it's in `/usr/bin`.
It should be `keyword`, `builtin` or `external commands`?

jfcherng: Discussion redirect to https://github.com/sublimehq/Packages/issues/46.

PR #45, opened 2015-07-02
Comment character for Batch Files

dufferzafar:
So you can do `Ctrl + /` to quickly comment/uncomment lines.

### DIFF
--- a/Batch File/Comments.tmPreferences
+++ b/Batch File/Comments.tmPreferences
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+    <dict>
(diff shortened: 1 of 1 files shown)


jamescostian:
I just ran into this today and was about to post on the forums about it!

dufferzafar:
I've had this for a long time. But found out about this repo yesterday. :)

sublimehq:
Surely "rem" is much more commonly used for comments than "::"? While both work, I'd think that most users would expect toggle comment to insert a rem.

FichteFoll:
In any case, both should be referenced in the tmpreferences.

dufferzafar:
I can change :: to rem, but I am not sure how I'd add support for both. Any help?

FichteFoll:
Add an additional "TM_COMMENT_START_2" property. http://docs.sublimetext.info/en/latest/reference/comments.html

wbond:
Thanks @dufferzafar for the contribution!

Per your suggestion @FichteFoll, I added `::` as the second version and `rem` is now the default. bf1acacad2d58d1fd378695868726e8c6d5bd1bd

PR #47, opened 2015-07-02
[ShellScript] Lots of fixes with a given test case.

jfcherng:
Edit: Moved to be a [plugin](https://packagecontrol.io/packages/ShellScriptImproved) ( https://github.com/jfcherng/Sublime-ShellScriptImproved )

-

Test case is from [here](https://github.com/SublimeTextIssues/DefaultPackages/issues/125#issuecomment-112616917).
Not all wrong highlights are fixed but quite a lot.

Left / Right = Before / After
all

Note that an extra rule is added to my `.tmTheme` in above screenshots in order to set text to its default color, i.e., white on dark theme and black on light theme mostly. (I can not find other way to set text to the default text color...) This resets the color of `TEXT` in `$(echo TEXT)` and other things like that.
```xml
        <dict>
            <key>name</key>
            <string>Reset Color</string>
            <key>scope</key>
            <string>string meta.reset.color, meta.reset.color</string>
            <key>settings</key>
            <dict>
                <key>foreground</key>
                <string>#FFFFFF</string>
            </dict>
        </dict>
```


jfcherng:
Need more tweeks.

jfcherng:
The following commands are considered as `builtin`s as well.
They are tested by `$ type COMMAND`.
```
bind, builtin, command, declare, enable, export, getopts, 
hash, help, let, local, printf, pwd, read, readarray, readonly, 
return, shopt, test, times, trap, type, ulimit
```

FichteFoll:
I propose splitting the command list variables into a multiline string, in a `(?x:)` group to allow linebreaks, and determine linebreak positions by first letter or max-width (of 80/100). This makes the list better diffable and readable, and editable.

aziz:
> I propose splitting the command list variables into a multiline string

:+1: for diffability! if that's even a word! :smile:

jfcherng:
> I propose splitting the command list variables into a multiline string
Definitely a good suggestion! :+1:

jfcherng:
Just a note. Maybe [this file](https://gist.githubusercontent.com/okdana/45d3fd437e2f97f2ac63/raw/a822842edf8acdb7af4d182bbe18545f36d4c7be/bashsyntax.sh) can be used as a test case.

jfcherng:
```bash
echo `echo git --version`
```
The command above is legal, which prints `git --version`, and left-recursive.
There is no way to make a LL(*) parser parse this correctly.

zx8:
@jfcherng: In the latest screenshot in your OP, it looks like `if` is being erroneously considered a keyword: http://i.imgur.com/DOSDtbo.png

jfcherng:
@zx8 Yes. Some wrong highlights are just not fixed yet.

jfcherng:
@zx8 Fixed. Screenshots updated.

ShellScript/Shell-Unix-Generic.sublime-syntax
```diff
@@ -14,6 +14,70 @@ file_extensions:
   - .textmate_init
 first_line_match: '^#!.*\b(bash|zsh|sh|tcsh)|^#\s*-\*-[^*]*mode:\s*shell-script[^*]*-\*-'
 scope: source.shell
+variables:
+  cmdPrevilege: '(?x:
```
zx8: Small nitpick/typo here - correct spelling is "privilege"

jfcherng: Fixed. Thanks.

leesei:
Awesome x 1000 :clap: :clap:.
Thanks for your efforts.

stevekuznetsov:
Why is this not being pulled in?

jfcherng:
@stevekuznetsov 
I consider this pull request is too aggressive and hard to maintain in my opinion.
(Currently, I use this PR for my temporarily solution since it works better on my machine though.)
We need someone to re-write a syntax file for Shell language to make grammars more clear.

wbond:
@jfcherng Based on your comment, you believe that trying to support the various shell syntax is too difficult using the current version as a base, correct?

There is also some discussion at #87 in regards to handling the various shell variants.

jfcherng:
@wbond Yes, keywords/grammars could be very different among different shells. 

But it depends on how detail we want the final .sublime-syntax file(s) to be. Currently, the .sublime-syntax highlights keywords and hardly considers grammars. Take `echo echo` as an example, the secondary `echo` is a string literal but not a keyword.

If I am asked to write .sublime-syntax files to support sh, bash, ksh, zsh, csh, tcsh, etc... I would be lazy to find out their common parts. Maybe I would finish one and copy/modify it to fit other shells.

txtsd:
I'd definitely like zsh support.
Here are the [zsh specific commands](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html).

gechr:
@jfcherng: Have been using your fork for a while now and only just updated it today after ~4 months. Noticed an issue where variables which share the same name as commands are being highlighted:

archey

jfcherng:
I have move those changes to be a plugin hence closed.
https://packagecontrol.io/packages/ShellScriptImproved

wbond:
@jfcherng I was under the impression that you considered these changes too complex to maintain?

At some point I would like to improve the existing ShellScript. Hopefully I can benefit from a bunch of the hard work you've done.

jfcherng:
@wond 

> I was under the impression that you considered these changes too complex to maintain?

Yes. My commits are under the official one (i.e., not from the ground up), so thoughts are constrained under its "framework." I consider it as a working one and a messy one as well.

> At some point I would like to improve the existing ShellScript. Hopefully I can benefit from a bunch of the hard work you've done.

Feel free to take advantages from those commits :)

bmarcaur:
These all seem like good improvements. To avoid letting the good be the enemy of the perfect can we consider merging these changes @wbond ?

jfcherng:
@wbond I made [a syntax test file](https://github.com/jfcherng/Sublime-ShellScriptImproved/blob/master/syntax_test_shellscript.sh) for `ShellScriptImproved`. It may help you when you cope with `ShellScript` although the scope naming may differ.

wbond:
@bmarcaur Because of the addition of so many look behinds and no assertions in the test file, I'd say no for now.

My plan is to rewrite this relatively soon, using only patterns compatible with the new regex engine and the scopes we've been standardizing on.

@jfcherng Definitely, thanks for letting me know!

PR #60, opened 2015-07-05
Add missing captures to pop patterns

### LINKED ISSUE #56: bug in latex syntax highlighting, since 3093
Is it just me?

screenshot 2015-07-05 00 06 19

FichteFoll:
As explained in #56, the converter did not consider copying the "captures" dict to the pop pattern that is pushed onto the stack. I altered the conversion script (and made it suitable for command-line use because there were a few flaws with it). For convenience, it's now possible to pass entire directories.

Fixes #56, closes #57.

---

Changes to `Default/convert_syntax.py`:
```diff
--- C:\Users\Fichte\AppData\Roaming\Sublime Text 3\Packages\Default\convert_syntax.py   Tue Jun  2 11:04:30 2015
+++ C:\Users\Fichte\AppData\Roaming\Sublime Text 3\Packages\Default\convert_syntax.py   Mon Jul  6 00:16:43 2015
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 import plistlib
 import os
 import sys
@@ -9,6 +11,7 @@
 
 def needs_yaml_quoting(s):
     return (s == "" or s[0] in "\"'%-:?@`&*!,#|>0123456789="
+        or s.startswith("<<")
         or s in ["true", "false", "null"]
         or "# " in s or ": " in s
         or "[" in s or "]" in s or "{" in s or "}" in s
@@ -213,8 +216,9 @@
             end_entry = {}
             end_entry["match"] = format_regex(p["end"])
             end_entry["pop"] = True
-            if "endCaptures" in p:
-                captures = format_captures(p["endCaptures"])
+            if "endCaptures" in p or "captures" in p:
+                captures = format_captures(
+                    p.get("endCaptures", p.get("captures")))
                 if "0" in captures:
                     end_entry["scope"] = captures["0"]
                     del captures["0"]
@@ -294,8 +298,12 @@
                 # ctx.append({"include-syntax": key})
                 ctx.append({"include": format_external_syntax(key)})
 
-        else:
-            raise Exception("unknown pattern type: ")
+        elif list(p.keys()) == ["comment"]:
+            comment = format_comment(p["comment"])
+            if comment:
+                ctx.append({"comment": comment})
+        else:
+            raise Exception("unknown pattern type: %s" % p.keys())
 
     return ctx
 
@@ -394,7 +402,7 @@
             data = to_yaml(convert(syn))
 
             # to_yaml will leave some trailing whitespace, remove it
-            data = "\n".join([l.rstrip() for l in data.splitlines()]) + "\n"
+            data = "\n".join(l.rstrip() for l in data.splitlines()) + "\n"
 
             v = self.window.new_file()
             v.set_name(os.path.basename(base) + ".sublime-syntax")
@@ -432,9 +440,12 @@
     import fnmatch
 
     class sublime:
-        def find_resources(pattern):
+        base_path = "."
+
+        @classmethod
+        def find_resources(cls, pattern):
             paths = []
-            for root, dirs, files in os.walk('.'):
+            for root, dirs, files in os.walk(cls.base_path):
                 for fname in files:
                     if fnmatch.fnmatch(fname, pattern):
                         path = os.path.join(root, fname)
@@ -445,23 +456,41 @@
                         paths.append(path)
             return paths
 
+        @staticmethod
         def load_resource(fname):
             with open(fname, "r", encoding="utf-8") as f:
                 return f.read()
 
 if __name__ == "__main__":
-    if len(sys.argv) < 2:
+    args = sys.argv[1:]
+
+    if not args:
         print("usage: convert_syntax.py files")
-    else:
-        for fname in sys.argv[1:]:
-            o = convert(fname)
-
+        print("       convert_syntax.py folder")
+    else:
+        filenames = []
+        for path in args:
+            if os.path.isdir(path):
+                sublime.base_path = path
+                filenames.extend(sublime.find_resources("*.tmLanguage"))
+            else:
+                filenames.append(path)
+
+        for fname in filenames:
+            outfile = os.path.splitext(fname)[0] + ".sublime-syntax"
+            if os.path.exists(outfile):
+                print("file already exists: " + outfile)
+                continue
+
+            data = convert(fname)
+            text = to_yaml(data)
             # verify that to_yaml produces valid yaml for this object
             if yaml:
-                assert(o == yaml.load(to_yaml(o)))
-
-            with open(os.path.splitext(fname)[0] + ".sublime-syntax", "w",
-                encoding="utf-8") as f:
-                data = to_yaml(convert(sys.argv[1]))
-                data = "\n".join([l.rstrip() for l in data.splitlines()]) + "\n"
-                f.write(data)
+                assert(data == yaml.load(text))
+
+            with open(outfile, "w", encoding="utf-8") as f:
+                # to_yaml will leave some trailing whitespace, remove it
+                text = "\n".join(l.rstrip() for l in text.splitlines()) + "\n"
+                f.write(text)
+
+            print("converted " + outfile)
```

The `<<` check was needed for Groovy.tmLanguage, because a `<<` designates the "merge" type: http://yaml.org/type/merge.html

Clojure.tmLanguage was using a lot of "patterns" that only consisted of a comment key, so I made the script recognize these and copy them over.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -93,6 +93,8 @@ contexts:
       push:
         - meta_scope: comment.block.documentation.source.cs
         - match: $\n?
+          captures:
+            0: punctuation.definition.comment.source.cs
           pop: true
         - include: scope:text.xml
     - match: /\*
@@ -101,6 +103,8 @@ contexts:
       push:
         - meta_scope: comment.block.source.cs
         - match: \*/\n?
+          captures:
+            0: punctuation.definition.comment.source.cs
           pop: true
     - match: //
       captures:
@@ -108,6 +112,8 @@ contexts:
       push:
         - meta_scope: comment.line.double-slash.source.cs
         - match: $\n?
+          captures:
+            1: punctuation.definition.comment.source.cs
           pop: true
   constants:
     - match: \b(true|false|null|this|base)\b
@@ -216,6 +222,8 @@ contexts:
           push:
             - meta_scope: meta.method.identifier.source.cs
             - match: "(?={)"
+              captures:
+                1: entity.name.function.source.cs
               pop: true
         - match: '(?=\w.*\s+[\w.]+\s*\{)'
           push:
--- a/ActionScript/ActionScript.sublime-syntax
+++ b/ActionScript/ActionScript.sublime-syntax
@@ -51,6 +51,8 @@ contexts:
       push:
         - meta_scope: comment.block.actionscript.2
         - match: \*/
+          captures:
+            0: punctuation.definition.comment.actionscript.2
           pop: true
     - match: (//).*$\n?
       scope: comment.line.double-slash.actionscript.2
--- a/AppleScript/AppleScript.sublime-syntax
+++ b/AppleScript/AppleScript.sublime-syntax
@@ -34,11 +34,11 @@ contexts:
         - include: main
     - match: |-
         ^(?x)
-          \s*(to|on)\s+           # "on" or "to"
-          (\w+)             # function name
-          (\()              # opening paren
-            ((?:[\s,:\{\}]*(?:\w+)?)*)  # parameters
-          (\))              # closing paren
+        	\s*(to|on)\s+ 					# "on" or "to"
+        	(\w+)							# function name
+        	(\()							# opening paren
+        		((?:[\s,:\{\}]*(?:\w+)?)*)	# parameters
+        	(\))							# closing paren
       comment: |
         This is not a very well-designed rule.  For now,
                                 we can leave it like this though, as it sorta works.
@@ -57,13 +57,13 @@ contexts:
         - include: main
     - match: |-
         ^(?x)
-          \s*(to|on)\s+           # "on" or "to"
-          (\w+)             # function name
-          (?:\s+
-            (of|in)\s+          # "of" or "in"
-            (\w+)           # direct parameter
-          )?
-          (?=\s+(above|against|apart\s+from|around|aside\s+from|at|below|beneath|beside|between|by|for|from|instead\s+of|into|on|onto|out\s+of|over|thru|under)\b)
+        	\s*(to|on)\s+ 					# "on" or "to"
+        	(\w+)							# function name
+        	(?:\s+
+        		(of|in)\s+					# "of" or "in"
+        		(\w+)						# direct parameter
+        	)?
+        	(?=\s+(above|against|apart\s+from|around|aside\s+from|at|below|beneath|beside|between|by|for|from|instead\s+of|into|on|onto|out\s+of|over|thru|under)\b)
       comment: "TODO: match `given` parameters"
       captures:
         1: keyword.control.function.applescript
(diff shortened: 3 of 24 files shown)


FichteFoll:
@sublimehq, when you are incorporating the changes, I suggest also taking a look at when to offer conversin of the "current syntax". Currently, it offers to convert a syntax only when you are *using* an old-style .tmLanguage syntax but not when you have one opened. I think that wanting to convert the opened syntax would be the more likely use case, so prioritizing the extension of the currently opened file sounds like a better idea (while still falling back to checking the extension of the syntax file used afterwards).

C#/C#.sublime-syntax
```diff
@@ -101,13 +103,17 @@ contexts:
...
+            0: punctuation.definition.comment.source.cs
           pop: true
     - match: //
       captures:
         1: punctuation.definition.comment.source.cs
       push:
         - meta_scope: comment.line.double-slash.source.cs
         - match: $\n?
+          captures:
+            1: punctuation.definition.comment.source.cs
```
eric-wieser: This looks like nonsense to me - there is no capture `1` in the above string

keith-hall: if you look at the PR text, you can see it was just a conversion from `tmLanguage`, and it was so long ago that presumably its not even relevant now? Did you check on `master`?

eric-wieser: On master this type of mistake is still present in some places. Perhaps it should be an error to specify a `captures` index which doesn't exist?

FichteFoll: You are correct in that this PR added some superfluous captures, but they are an artifact of the conversion and technically required for feature parity. It's just that in some cases (like this one here) the end pattern (now a pop-match) didn't actually use the `captures` of the pattern. If you'd like to go ahead and remove those that still exist, that'd be appreciated, but as it is they don't do anything bad.

keith-hall: That is unsurprising as that syntax is practically untouched since the conversion, no? Sometimes capture groups exist in the pattern but don't participate in a match, so if this were to be implemented it would have to work at regex compile time and not during lexing (probably I'm stating the obvious here, but just in case ;))

eric-wieser: > Sometimes capture groups exist in the pattern but don't participate in a match

That's not the case I'm suggesting become an error. Referring to group 1 in `(?:matches|(doesn't) match)` is fine, but referring to group 2 in that regex is knowably stupid at regex compilation/lex-time before attempting to match the regex against any text.

FichteFoll: >That's not the case I'm suggesting become an error / warning.

What are you suggesting then?

I was just explaining that captures referring to groups that don't exist (in pop matches) is an artifact of the old tmLanguage format (explained [on the unofficial docs](http://docs.sublimetext.info/en/latest/extensibility/syntaxdefs.html)) where `captures` applied to both begin and end patterns if no `beginCaptures` or `endCaptures` mapping has been provided. (I don't think this is documented.)

I'd be perfectly fine with ST reporting these as warnings in the console on compilation, as they hint at a user error more than anything. Regardless, that should be a suggestion made on https://github.com/SublimeTextIssues/Core.

PR #61, opened 2015-07-05
Convert remaining .tmLanguage files

FichteFoll:
Follow up to #23.

I don't exactly know why these syntaxes were omitted, but since I converted them all anyway (using the modified script in #60), I might as well propose this as a pull request.


jrappen:
Thanks for fixing this, I wasn't aware of the issue.

sublimehq:
Looks great, but could you please update the pull request  to leave "Plain text.tmLanguage" as-is. There are almost certainly plugins that test the current syntax against the string "Plain text.tmLanguage", and I'd rather not break them.

FichteFoll:
sure, I also thought about that but considering that the same applies to all the other syntaxes too, which is why I prioritized consistency.

FichteFoll:
Done.

PR #65, opened 2015-07-08
Recognize config.ru (Rackup file) as Ruby

berfarah:
The `config.ru` file is a standard deployment file for Rack and uses Ruby syntax.


sublimehq:
Thanks - the pull request won't merge automatically, so I manually made the same change in c97649b for simplicity

PR #66, opened 2015-07-08
[HTML] Consistent highlighting of punctuation in attributes

anlutro:
Currently, the `=` in id and style attributes gets special treatment by being highlighted. This change will make sure that all attributes are highlighted the same.


sublimehq:
Thanks for the pull request, I like the intent of the change, but the pull request itself wasn't working for me. I've pushed an equivalent change though. Will close this pull request for now.

anlutro:
Ah OK, I misunderstood the regex I suppose. As long as it got fixed I'm happy, thanks!

PR #67, opened 2015-07-08
Synchronized isn't a keyword in scala

djspiewak:
I almost never use it, but `synchronized` really isn't a keyword in Scala.  It's a method on `AnyRef`.


wbond:
This change was applied to the new `.sublime-syntax` definition in a3475970efa1422c21b3db43e28e4445059abd08

wbond:
BTW, thanks for the contribution!

PR #71, opened 2015-07-12
[HTML] Doctype scopes

aziz:
- Adding scope for the pop rule
- Fixing capture number

I guess the convertor has a bug when calculating the capture numbers in case of 
Regex options like `(?i)`
I've not investigated this though. It's just a guess! 

Example: 
```
- match: (?i:DOCTYPE)
```
Shouldn't have a capture index of 1.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -54,10 +54,11 @@ contexts:
       push:
         - meta_scope: meta.tag.sgml.html
         - match: ">"
+          scope: punctuation.definition.tag.html
           pop: true
         - match: (?i:DOCTYPE)
           captures:
-            1: entity.name.tag.doctype.html
+            0: entity.name.tag.doctype.html
           push:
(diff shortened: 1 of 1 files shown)


FichteFoll:
I suspect this was faulty in the initial version, otherwise it would have converted this single 0 capture to use the scope key.

aziz:
> otherwise it would have converted this single 0 capture to use the scope key.

I totally agree. All these `captures ` blocks with only a `0` key should be converted to use `scope` instead. 
Poking @jskinner

aziz:
OMG! There're lot's of these things around. Replacing them can dramatically reduce the syntax file in terms of LoC. 
in Ruby's case we've got 104 matches. Saves over 10% of LoC. 

Here's the regex I used to find them
``` ruby
/captures:\n\s+0:.+\n(?!=\s+\d)/
```

FichteFoll:
That's weird, because the script actually checks for these and should automatically turn them into `scope` lines. I'll take a look at it later.

PR #74, opened 2015-07-16
Replaced captures: 0: scope_name with scope: scope_name

aziz:
Saves over 600 lines of code and simplifies syntax files. 
Here are the three find and replace steps I took:

1. Replace captures with only 0 key that also has scope definition.
we shouldn't replace these ones. so I temporarily replace 0: with X:
Find: `(scope: )(.+)(\n\s+captures:\n\s+)(0:)(.+)(\n)(?!\s+\d)`
Replace: `\1\2\3X:\5\6`

2. Replace captures: 0: scope_name with scope: scope_name
Find: `captures:\n\s+0:(.+)\n(?!\s+\d)`
Replace: `scope:\1\n`

3. Bring back the X: scopes
Find: `(scope: )(.+)(\n\s+captures:\n\s+)(X:)(.+)(\n)(?!\s+\d)`
Replace: `\1\2\30:\5\6`

### DIFF
--- a/ASP/ASP.sublime-syntax
+++ b/ASP/ASP.sublime-syntax
@@ -35,13 +35,11 @@ contexts:
     - match: (?i:\b(Empty|False|Nothing|Null|True)\b)
       scope: constant.language.asp
     - match: '"'
-      captures:
-        0: punctuation.definition.string.begin.asp
+      scope: punctuation.definition.string.begin.asp
       push:
         - meta_scope: string.quoted.double.asp
         - match: '"(?!")'
-          captures:
-            0: punctuation.definition.string.end.asp
+          scope: punctuation.definition.string.end.asp
           pop: true
         - match: '""'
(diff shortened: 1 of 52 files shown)


wbond:
This large change conflicts with existing PRs and merged changes. I can do something like this, but I'd like to do it when there are fewer pending changes.

aziz:
It's all right. take your time. I appreciate that you're keeping these packages up to date.

jrappen:
@wbond not sure why you added sublimehq/packages@40100f2 instead of merging this PR

wbond:
Because a massive change set like this would break every open PR. And searching for and replacing these scope 0 during merges is easier than resolving merge conflicts.

wbond:
These are being fixed one at a time as I work on packages. As mentioned above, this is to prevent merge conflicts with existing pull requests, but also so I can review them and make other fixes along the way.

I do appreciate the work that went into this, though!

PR #76, opened 2015-07-17
[Java] Yet another captures index fix

aziz:


wbond:
That is not the correct scope for `(`. I addressed this issue in a recent commit.

PR #77, opened 2015-07-19
[PHP Source] cleaned up one match regexp

aziz:
- No need to be multiline
- Removed the unused named capture


FichteFoll:
Doesn't need to be quoted either.

aziz:
I think it does because of `{` and `}`.
I was looking at `needs_yaml_quoting` method in `convert_syntax.py`

FichteFoll:
It's fine, I'm working on a new yaml Def and this is perfectly valid because we're in a block context. needs_yaml_quoting is more on the safe side.

wbond:
I've been working on a bunch of fixes to PHP, and this is addressed in those changes.

PR #79, opened 2015-07-22
[TCL] Fix highlighting of escape character \\

TheClams:
Fix issue  SublimeTextIssues/DefaultPackages#134

Also add test file


wbond:
Thanks for this fix and the tests to go along with it!

PR #81, opened 2015-07-28
Add .gradle files to the list of Groovy extensions

sschuberth:
Gradle [1] is a build system that uses Groovy as its DSL.

[1] http://gradle.org/


sschuberth:
@sublimehq Would you mind having a look?

wbond:
Thanks!

PR #82, opened 2015-08-03
Add shell filetypes

gerardroche:


wbond:
Thanks for these additions!

PR #83, opened 2015-08-03
Move inc filetype to php

gerardroche:


wbond:
It seems `.inc` is used for both ASP and PHP.

Since there isn't a clear correct answer for this extension among server-side technologies, I am thinking we should leave it up to the end user to use the *Open all with current extension as...* functionality.

gerardroche:
+1

PR #85, opened 2015-08-05
[JavaScript] Add missing blank

chinesedfan:
Fix `Object-Method`, `Object-Value-JS`. Make their styles the same with `function` and `Object-key-key-value` snippets.


wbond:
Thanks for these tweaks!

PR #87, opened 2015-08-10
[ShellScript] Add support for ksh and csh to Shell-Unix-Generic

dannywarren:
* rename from "Shell Script (Bash)" to just "Shell Script" (not sure if you guys want to do this, but it seems to fit better since this is the "generic" shell syntax parser, and even the previously included sh and tcsh aren't at parity with bash either)
* add ksh and csh file extensions
* add ksh and csh (including common variants) to shebang line parser
* add "print" to builtins keyword list for ksh
* move "break" to control structures keyword list
* add ksh, csh and common variants to shell keyword list


wbond:
I think it gets tough to do a general shell syntax, as opposed to bash-specific. http://hyperpolyglot.org/unix-shells has some of the differences. However, there already was some (t)csh support in the existing syntax.

This isn't a decision, just recording some thoughts.

FichteFoll:
I don't know enough about bash variants, but would adding a common subset of all shell syntax and then adding specific variants on top of it feasable?

`.sublime-syntax` is better for this "include-and-monkeypatch" behavior, but only injecting patterns to certain contexts is not possible atm.

dannywarren:
@wbond: Agreed, I don't think there is an easy answer to this one.  The historical messiness of all the common unix shells makes it impossible to find a clean spot to split them in to different syntax definitions.  Even trying to nail down the differences between each flavor of ksh would be messy.

@FichteFoll: I was typing up a long rambling comment wondering if something like this might be possible.

Without something like this, trying to maintain a syntax definition for each shell variant and flavor and ignoring similarities would be a nightmare.  

Ignoring the differences between the shells and pushing for "good enough" definition that covers the similarities is definitely not right either.  Admittedly, that's what my patch moves towards, but I'm not sure I can argue in favor of it.

I'm also wondering if there are any shells that are just different enough from the other bash-like shells to make "include-and-monkeypatch" painful.  Fish comes close, but is still similar enough to make it worthwhile to share definitions.

dannywarren:
@wbond: Also, the link you provided is really helpful.  As someone who has to frequently hop between shells for scripting, that will cut down some of the context-switching penalty.

I used to use this doc, but it's rather shallow and out of date:

http://tldp.org/LDP/Bash-Beginners-Guide/html/x7369.html

wbond:
I think the path forward is going to be to write variations of the shell script syntax for bourne/bash, zsh and tcsh. Once we get a reasonable basis in place using the new regex engine, then hopefully we can get frequently shell script users to identify issues and submit PRs.

dannywarren:
Thanks Will!  I read up on the discussion over in pull #47 too, and I think you guys are going the right way with it maintainability wise.

I'll keep an eye on task #434, happy to help with any ksh-specific stuff once it gets to that point.

PR #89, opened 2015-08-13
Fix issue #140: Don't match keyword new when used as method reference.

StrangeNoises:
ie: In test for anonymous-classes-and-new, ignore 'new' when prefixed with two colons. Fixes problems when using eg: `stream.collect(Collectors.toCollection(ArrayList::new));`


wbond:
Thanks for this tweak. I added the original example code from https://github.com/SublimeTextIssues/DefaultPackages/issues/140 in e682b30de6dffedcb41d12237fd54a288cf6fff0.

PR #93, opened 2015-08-26
[C#] Rewriting from scratch

gwenzek:
Hi, I started from scratch a new syntax for C# on [a personal repository](https://github.com/gwenzek/csharpSyntax).
You'll find there several tests. I tried to use old scopes when they existed.

The principal features are: 
- Detection of variables declarations (that's a lot !)
- Detection of tag: `[TestMethod]`
- Support for new C# 6 syntax: `$"{1 + 2}"`, "x?.y", ...
- Detection of operators
- Detection of generic types
- Detection of labels
- Detection of function calls

I'll be happy to have some feedback on my work.


wbond:
Thanks for the work on this!

In order to reasonably consider these changes, we'll need some fairly extensive tests. These will consist of various C# files that are named a specific way, and include comments that indicate what scopes should be applied to what code. I recently added one in the C# folder for a bug someone reported. You can read more about syntax testing by scrolling to the Testing header on http://www.sublimetext.com/docs/3/syntax.html.

Additionally, with dev build 3098 of Sublime Text, a "Performance" variant of the Syntax Tests build system was introduced. This will allow timing the highlighting of the currently open file using the syntax currently selected. Could you find some large-ish C# files and compare the performance of before and after your tweaks, and post them here? That will help us ensure that the regular expressions don't have any performance issues.

gwenzek:
Thanks for the feedback!
As mentionned in my PR, I have made several tests that you can see in a separate repo.
Can you have a look at them in order to tell me if you find them sufficient?
I'll add them to my PR soon.

wbond:
Sorry, I did miss that! Those tests look like they cover a good amount of ground. If you could just include the existing test I created to ensure your version supports the same features, then I think we'll be in a good place.

Do note that I changed "annotations" to "attributes" in the scope names. From what I can tell, "attributes" is the correct term for C#. I'm obviously happy to be corrected since C# is not something I have worked very much with.

GuilhermeHideki:
Attributes seems the "right" word. MSDN link about: https://msdn.microsoft.com/en-us/library/z0w1kczw.aspx

I think you should include one test with the number suffixes because the sublime's version had the c/c++ LL "long long" but didn't have the "decimal" D

gwenzek:
I'm not sure it's a good idea to change the scope names to reflect the
vocabulary of a specific language.

My point is that people creating themes shouldn't have to know all the
languages and their vocabulary.

On the contrary, I used the same scope for the 'using' C# keyword than the
scope used in the python syntax for 'import'.

It would be really nice to have guidelines scopes naming. The Texmate
documentation have some but their aren't very detailed and let a lot of
things open to interpretation.

gwenzek:
Regarding performances, I tested on a 2600 lines C# files made of a concatenation of several files:

    Syntax "Packages/User/csharpSyntax/csharp.sublime-syntax" took an average of 51.3ms over 10 runs
    Syntax "Packages/C#/C#.sublime-syntax" took an average of 58.0ms over 10 runs

So my version seems a little bit faster even if the syntax itself is bigger (weird, I would have expected the contrary).

gwenzek:
Here is a gif showing the differences between the 'old' syntax and my proposition: 
Note how the old syntax was broken by generics. I also tried to be more coherent on naming:
`double` is now marked as a type not as a keyword, like in most other syntaxes.

Differences between 'old' syntax and mine

buildersbrewery:
Any updates regarding the status of this PR?

wbond:
@buildersbrewery If I haven't closed an issue, it means that I am still intending to do something with it, or am waiting on changes from the author.

In the case of this PR, I haven't yet gotten to it, but also there are a few things that would need to be resolved:

 1. The tests are written to test a different syntax file (see the `SYNTAX TEST` lines)
 2. There are merge conflicts with the master branch - existing tests should be preserved to prevent regressions
 3. Snippets are added in addition to tweaking the syntax (they should be a separate PR, and even then I'm not sure I want to open the floodgates of adding snippets to default packages)

I have not reviewed the actual syntax definition yet to see if there are incompatible regex constructs, or if scope names are consistent with the other languages that have been rewritten recently (C++ is only of the most recent, non-trivial examples).

gwenzek:
Thanks @wbond for the feedback,
I'll have a look at the C++ syntax.
I tried to be coherent with myself, but if you want people to follow conventions you should write them down 😃

What do you mean by incompatible regex? Is it just look-behind?  Or are they other patterns I should avoid?

wbond:
@gwenzek My plan is to write them down, I have just been using my experience of rewriting a bunch of the syntaxes to figure out what needs to be written down. I've covered a number of the languages with C-like syntax, however there are a number that I have not yet. I will definitely provide feedback about the scope names once I have time to review this PR in-depth. Part of this whole process of getting all of the syntaxes up-to-date is to make the scopes more consistent so we can document them for syntax and color scheme authors.

The Syntax Tests build system has a number of variants (ctrl+shift+b), one of which allows testing the compatibility with the new regex engine in Sublime Text. Moving forward we are removing all regexes that use incompatible features because the new regex engine is so much faster. This allows the highlighting of files and the indexer to run faster. There may be some edge cases of constructs from Oniguruma we aren't supporting in the new engine yet, so feel free to report those. We won't, however, be supporting lookbehinds, backreferences or atomic subgroups (those are generally unnecessary with the new engine). This has to do with the fact that the new engine is not a backtracking engine.

gwenzek:
@wbond I removed the backreferences, the syntax is now compatible with the new regex engine.

I added back the tests that you added in the meanwhile but I don't have exactly the same convention for scopes.
In particular I feel that the scope for "attribute" should be something else than `meta.method.attribute`. 
Attribute or decorators exists in several language, their should be a real scope for that.

Also I'd like to have a scope for the `.` I currently use `keyword.operator.accessor` as others but I think it should rather be a `punctuation.separator.something`, what do you think ?

My syntax does a great job at handling type parameters, variable declarations and method declarations.
It's also compatible with the last C# 6 syntax  so it really improves the readability of C# code.

buildersbrewery:
`keyword.operator.accessor` makes perfectly sense for languages where you access the x part of a vector via `float x = v.x;`

buildersbrewery:
Really depends on the context.

FichteFoll:
The current convention is to scope the dot as `punctuation.accessor.dot`.

gwenzek:
Thanks for the suggestion @FichteFoll ,
I renamed `keyword.operator.accessor` to `punctuation.accessor.dot` in  0e33931,
with a distinction between `dot` and `double-colon`

gwenzek:
Thanks for testing @michaelblyons.
I fixed the first bug, my look-ahead detecting function-call versus variable declaration was fooled by the `=`. 
I also corrected the misspelling of "reflection".

In the current version of this PR I mark `int?` and `string[]` as `support.type`.
For me the `?` and the `[]` are just a short-name for `Option<T>` and `Array<T>` so I don't feel they should receive a special treatment.
But it's quite easy to add a more precise scope, I'm just not sure what they should be.

The `typeof` receives the same scope as `as` and `is` which are more operators, maybe I could just use a `support.function`.

keith-hall:
When creating an anonymous type, like:

    new {hello = "world", foo = "bar"}

The `{` and `}` characters get the scope `punctuation.definition.array.begin.cs` and `punctuation.definition.array.end.cs` respectively. This may be a nitpick, but it's not an array... :)

Maybe it would be possible to check whether `[]` precedes the `{` character, so that:

    new[] { "hello", "world" }

will still be correctly recognized as an array.

C#/C#.sublime-syntax
```diff
@@ -1,276 +1,1593 @@
...
+  class_declaration:
+    - match: '\b(static|unsafe|abstract|partial|sealed)\b'
+      scope: storage.modifier.cs
+    - match: '{{visibility}}'
+      scope: storage.modifier.access.cs
+    - match: '\b(class)\s+{{name}}'
+      captures:
+        0: meta.class.identifier.cs
+        1: storage.type.class.class.cs
+        2: entity.name.type.class.cs
```
keith-hall: these days, this should just be `entity.name.class.cs`, without the `.type` after `entity.name`

gwenzek: Thanks, I wasn't aware the convention changed.

C#/C#.sublime-syntax
```diff
@@ -1,276 +1,1593 @@
...
+      push: tag_in
+    - match: \[
+      scope: punctuation.definition.tag.begin.cs
+      push: tag_in
+
+  tag_in:
+    - meta_scope: meta.attribute.cs
+    - match: '{{name}}(\()'
+      captures:
+        1: entity.name.tag
```
keith-hall: I'm not sure I agree with the scope `entity.name.tag` here. Perhaps it should just be `variable.other.type.attribute.cs` or something instead?

C#/C#.sublime-syntax
```diff
@@ -1,276 +1,1593 @@
...
+      scope: punctuation.separator.accessor.cs
+    - match: \]
+      scope: punctuation.definition.accessor.end.cs
+      pop: true
+    - include: line_of_code_in
+
+  type:
+    - include: type_common
+    - match: '{{name}}'
+      scope: variable.other.type
```
keith-hall: this is missing the `.cs` suffix. There are a few other instances of `scope: variable.other.type` where this applies too.

LoneBoco:
I submitted a patch in #590 for fixing the numerical suffixes and I noticed that your C# re-write doesn't support all valid suffixes.  I linked to various MSDN articles in #590 that supports my claims.  You are missing the decimal suffixes as well as some of the odd ulong suffixes.

gwenzek:
Thanks, I'll look at it.
If you have other impressions regarding my PR please let me know.

Here or at:
https://forum.sublimetext.com/t/looking-for-beta-tester-for-c-syntax/21373

michaelblyons:
I am curious about the choice of scope for the braces on class definitions.

```cs
private class ExampleClass { }
//                         ^ ^ These guys.
```

You're using `punctuation.section.class.begin.cs` and `end`, where the docs [appear to want `punctuation.section.block.begin`,][docs]  and where Rust (which is recommended as a guide) [uses `punctuation.definition.block.begin`.][rust]

I noticed this when the maintainer of the theme I use [set a scope][theme] for `punctuation.section.class.end` and now all my classes' end brace is a different color. If that's on him, I'll go take it up with him, but if it's because there isn't agreement on brace block scoping, I think we should talk about it here.

[docs]: https://www.sublimetext.com/docs/3/scope_naming.html#meta
[rust]: https://github.com/sublimehq/Packages/blob/master/Rust/syntax_test_rust.rs#L253
[theme]: https://github.com/oivva/boxy/commit/77db43304723e92cf6306440d8b0a99a1b3b7768#diff-369006d00eeadafe404fa9c83b0c8904R318

wbond:
@michaelblyons `punctuation.section.block.begin` is the preferred scope name. Many in the wild also use `punctuation.definition.block.begin`. I wrote Rust before we decided that `section` would be for larger sections and `definition` would be for more "atomic" type things, such as strings, variable names, etc. I spent some time updating a few of the languages, but have not made it through all of them yet.

We certainly aren't going to get all syntax authors and color scheme authors to 100% follow the recommendations, especially since many packages are developed and then not meaningfully maintained. My hope is that at least we can make it so intentional color scheme authors can at least provide a good experience for the default syntaxes.

LoneBoco:
I took a look at this language syntax in one of our projects and noticed that it completely broke due to a nested class.  I've attached a test case that shows how it breaks.

[syntax_break.txt](https://github.com/sublimehq/Packages/files/427929/syntax_break.txt)

gwenzek:
Thanks @LoneBoco I tested on my computer and I effectively had a problem due to an error introduced in the `[C#] add regions to symbol list` but it was due to the `#endregion` not the nested class.
I (force) pushed a fix so it's okay now.

Please could you try with the new version ?

gwenzek:
@michaelblyons Here are the scopes I use. I tried to be coherent with other syntaxes, but I choose those names a year ago and there were no written conventions at that point. I'm open to any proposition but I like the different scopes depending on the nature of the content of the braces.
I put different colors on those and therefore don't need plugins like "BracketHighlighter"

    namespace HelloWorld
    {
    ///<- punctuation.section.namespace
        class Hello
        {
    /// ^ punctuation.section.class

            static void Main()
            {
    ///     ^ punctuation.section.function
                someCode;
                {
    ///         ^ punctuation.section.block
                }
            }
        }
    }

LoneBoco:
@gwenzek It is much better now.  Here is another one I discovered.

1) The scope of DateTime as a class member variable and as a function variable are different.
2) The nullable operator on DateTime as a class member variable is not identified, but it is when it is a function variable.
3) The nullable operator is identified as a ternary operator when in an explicit cast.
4) The DateTime inside the explicit cast is identified as support.type.cs, but nowhere else.

[syntax_nullable.txt](https://github.com/sublimehq/Packages/files/428131/syntax_nullable.txt)

LoneBoco:
Enums are set to constant.numeric.enum.cs, but only when the enum contains non-numeric characters.  If it contains a number, it gets identified as variable.other.cs.  In my example, this causes one of the enums to not by colored by the default theme.

[syntax_enum.txt](https://github.com/sublimehq/Packages/files/430861/syntax_enum.txt)

LoneBoco:
Constants defined inside a function are not handled properly.
`void test()
{
    const int i = 1;
}`

LoneBoco:
Variables set inside the initializer of a for loop are not scoped in a way that colors their type (expecting something like support.type.cs).
`for (int i = 0; i < items.Count; ++i)`
`foreach (int test in collection)`

gwenzek:
Thanks @LoneBoco for the bug reports.

I added support for decimal literals.

I improved the detection of Enum to also support name containing digits.
The thing is I their is no definite way to detect enums so I just match
all-caps name.

I fixed the `const` not detected inside code blocks.

But I wasn't able to reproduce your problem with the `for` loop. I added
some tests in `GeneralStructure`.
For now I use the scope `variable.other.type` for marking types.
I think the convention would be to use `storage.type` but I'd like to
change it (because it's also the scope used for `const` or `var` which
aren't types). On Scala they use `support.type` for every type appearing in
the code and I find it much clearer.

LoneBoco:
Take this snippet:
```
public class Test
{
	public Test()
	{
		int y = 1;

		for (int x = items.Count - 1; x >= 0; x--)
		{}

		foreach (int x in items)
		{}
	}
}
```
On the "int" keyword for the y variable, we have this scope:
support.type.cs

On the int keyword in the for and foreach loop, we have these:
variable.other.type.cs

The variable.other.type.cs scope does not have any coloring on the default Monokai theme.
The C++ syntax uses storage.type.c for both instances.

LoneBoco:
C# supports the usage of "new" as a function modifier:
https://msdn.microsoft.com/en-us/library/435f1dw2.aspx

The syntax currently breaks horrible if you try to use it for that purpose.
```
public class BaseC
{
    public int x;
    public void Invoke() { }
}
public class DerivedC : BaseC
{
    new public void Invoke() { }
}
```

gwenzek:
@LoneBoco thanks for the feedback but you must be using an old version, I fixed this a month ago :-)

LoneBoco:
@gwenzek Yep, something weird was going on with git on my machine.  Verified its working now.  Seems to also have been the cause of the for loop problem I had.

LoneBoco:
Some issues I noticed with regions.
```
public class A
{
	#region < storage.type.section.cs
	#endregion

	#region < variable.other.section.cs
	public void Func()
	{
		#region Outer
		if (test)
		{
			#region Inner
			var a = 1;
			#endregion
		}
		else
		{

		}
		#endregion
	}
	#endregion
}
```

1. The second region has a different scope and isn't colored with the default syntax.

LoneBoco:
Another issue:
```c#
public class B
{
	#region//Test
	public void Test()
	{
	}
	#endregion
}
```
Visual Studio allows the above as a valid region.  Syntax highlighting goes crazy if it encounters an issue like the above.

wbond:
Thanks for the work the two of you are doing on this PR.

At some point in the next few weeks I am hoping to go through this and review it.

The things that will make it the review go most quickly will to ensure that:

 1. it passes the currently C# tests (I realize some scope names will likely need to be tweaked)
 2. it tries to follow out scope guidelines: http://www.sublimetext.com/docs/3/scope_naming.html
 3. it passes the `Regex Compatibility` variant of the `Syntax Tests` build system

gwenzek:
Thanks @wbond  for coming back on this.

1. the current tests are passed, the test file is part of the PR,
I have to check with the latest change but it should be ok thanks to LoneBocco.

3. the regex compatibility test is passing, I removed look behind a few months ago.

For the scopes I haven't started to refactor every thing so I have a bit of work todo.
I also got several questions:

1. I'd prefer to use Scala-like scope for methods and variables:
```cs
    Foo foo = new Foo();
//  ^^^ support.class
    var bar = new Bar();
//  ^^^ storage.type
```
instead of Java-like scopes:

```java
    Foo foo = new Foo();
//  ^^^ storage.type
    var bar = new Bar();
//  ^^^ storage.type
```
because we need  a clear distinction between `val` which is a keyword and `Foo` which is a type.
What's your opinion on this ? The documentation isn't really helpful on this part.

2. How do I scope annotations ? like in:
```cs
[SomeAnnotation("Hello")]
public void Foo() {}
```
I propose the same thing than in #605 ie: `storage.modifier.annotation`.

LoneBoco:
@gwenzek My #590 was merged just recently so you will need to update your syntax_test_c#.cs file.  It also doesn't pass since my PR updated the numeric suffixes.  This was tested against your changeset 8f44b4840d9b2890c4b47f0db6e8fb6290e9ac1d

gwenzek:
I rebased the commit and fixed the commit messages.
You're right @LoneBoco I lost the commit fixing the decimal numerals, I'll fix that

LoneBoco:
Rebasing everything has fixed all the problems I have mentioned above, @gwenzek.  Things are looking a lot better.

michaelblyons:
The scopes differ between constants defined at the class level and those inside methods:

```cs
public class Foo
{
    private const string _bar = "bar";
//          ^^^^^        storage.modifier.cs
//                ^^^^^^ support.type.cs
    protected void Baz()
    {
        const string stuff = "stuff";
//      ^^^^^              variable.type.other.cs
//            ^^^^^^       entity.name.variable.cs
//                   ^^^^^ entity.name.variable.cs
    }
}
```

FichteFoll:
> 
> I'd prefer to use Scala-like scope for methods and variables:
> 
> ```c#
    Foo foo = new Foo();
//  ^^^ support.class
    var bar = new Bar();
//  ^^^ storage.type
```
> 
> instead of Java-like scopes:
> 
> ```c#
    Foo foo = new Foo();
//  ^^^ storage.type
    var bar = new Bar();
//  ^^^ storage.type
```
> 
> because we need a clear distinction between var which is a keyword and Foo which is a type.
> What's your opinion on this ? The documentation isn't really helpful on this part.

See also #582.

This situation is tricky because both, staticly typed and dynamically typed (or inferred) declarations, are possible. For statically typed languages, you just use `storage.type` for everything pretty much, and then some specifications below that. For dynamically typed, you probably use `storage.type` for the keyword that indicates a declaration, like `var` or `let` and `const` in JavaScript.
This leaves us with a situation where they would both get the same two-part scope in a language where both is allowed.
Ideally, a different two-part storage scope like `storage.keyword` should be used for dynamic languages since, strictly speaking, they do not tell the type of a variable. They just indicate that we're declaring a variable but not its type. Unfortunately, this scope is unknown to anyone because I just made it up, so color schemes don't have support for it. Because of that, we could probably get away with just using a keyword scope like `keyword.other.storage` for dynamic declaration keywords.

Thus we'd do it as follows:

```c#
    Foo foo = new Foo();
//  ^^^ storage.type
    var bar = new Bar();
//  ^^^ storage.keyword
```

or

```c#
    Foo foo = new Foo();
//  ^^^ storage.type
    var bar = new Bar();
//  ^^^ keyword.other.storage
```

LoneBoco:
I found another lovely problem.  Attributes can be declared within a function's parameters, like so:
```c#
public class MyController
{
    [AcceptVerbs(HttpVerbs.Post)]
    [ValidateAntiForgeryToken]
    [JsonHandleError]
    public virtual ActionResult Process([ModelBinder(typeof(MyModelBinder))]
                                        [Bind(Prefix = "MyContent")] MyProcessModel model)
    {
    }
}
```
It looks like it identifies the first attribute correctly, but then fails on the second one.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1623 @@
...
+        - match: \}
+          scope: punctuation.section.trycatch.end.cs
+          pop: true
+        - include: code_block_in
+    - match: \S
+      scope: invalid.illegal
+      pop: true
+
+  using_block:
+    - meta_scope: meta.block.trycatch.using.cs
```
LoneBoco: Your using block is using trycatch identifiers.

gwenzek:
That's voluntary because a 'using' is like a 'try/finally'

gwenzek:
@LoneBoco I fixed the bug with attributes. The problem was the space between attributes, I had to insert a context for matching the attributes.

LoneBoco:
@gwenzek Another issue I've noticed with regions.
```c#
public class B
{
    #region//Test
    #endregion

    #region Test / Test
    #endregion
}
```
First one doesn't really have a problem, but comments should be disabled for the region name.
The second one has invalid.illegal.cs scope.

After #region, everything should be valid.

gwenzek:
@LoneBoco this is getting hairy !
Are you sure that `//Test` is a valid region name ? because it really looks like a comment !
For the second one it's because I use `[^/]*` to match region names. I'm not sure how I could match something with `/` in it. I'll look into it but I don't if it's possible.
But I can remove the `invalid` scope.

LoneBoco:
@gwenzek Yes, I am sure.  Visual Studio has no problems at all with any of those region names.
```
public class B
{
    #region//Test
    #endregion
}
```
Will get collapsed in Visual Studio as:
```
public class B
{
+   //Test
}
```
The + sign expands the selection, //Test has a box drawn around it.

My workplace has an absolutely massive C# code base that is worked on by probably a hundred people, so I see all sorts of weird things all day long.

FichteFoll:
> I'm not sure how I could match something with / in it.

`([^/]*|/[^/])*`

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+
+  comments_in:
+    - match: \b(?i:todo|hack|xxx)\b
+      scope: comment.line.todo
+    - match: '/!\\'
+      scope: comment.line.todo
+    - match: \b(https?://\S*)\b
+      scope: markup.underline.link.cs
+
+  git_merge:
```
wbond: I don't want to try to get into highlighting things like this in the default syntaxes.

Ideally this can be handled by a plugin that will scan the file and look for such patterns and alert the user.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
-  main:
-    - match: '^\s*(using)\s+([^ ;]*);'
-      scope: meta.keyword.using.source.cs
-      captures:
-        1: keyword.other.using.source.cs
-    - match: '^\s*((namespace)\s+([\w.]+))'
+  prototype:
+    - include: comments
+    - include: git_merge
+    - match: '#\s*'
```
wbond: A `#` anywhere in the syntax is a valid location for a preprocessor option? It isn't restricted to being the first thing on a line?

It probably also makes sense to scope the `#` with a variation of `punctuation.definition`.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+  visibility: \b(public|private|protected|internal|protected\s+internal)\b
+  base_type: (?:bool|byte|sbyte|char|decimal|double|float|int|uint|long|ulong|short|ushort|object|string|void)
+  type_suffix: ((?:\[,*\]|\*|\?)*)
+  generic_declaration: \s*(<[^(={};<]*>)?\s*
+  reserved: '(?:abstract|as|base|break|case|catch|checked|class|const|continue|default|delegate|do|else|enum|event|explicit|extern|finally|fixed|for|foreach|goto|if|implicit|in|interface|internal|is|lock|namespace|new|null|operator|out|override|params|private|protected|public|readonly|ref|return|sealed|sizeof|stackalloc|static|string|struct|switch|this|throw|try|typeof|unchecked|unsafe|using|virtual|volatile|while)'
+  name: '(@{{reserved}}|@{{base_type}}|@var|{{name_normal}})'
+  # name: '([A-Za-z_][A-Za-z_0-9]*)'
+  # name_normal: '[A-Za-z_][A-Za-z_0-9]*'
+  # name_normal: '(?:[A-Za-z_\p{L}]|unicode_char)[A-Za-z_0-9\p{L}]*'
+  start_char: '(?:{{unicode_char}}|[A-Za-z_\p{L}])'
```
wbond: `\p{L}` includes `A-Za-z`.

Can you really use `\u####` in identifiers?

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+  other_char: '(?:{{unicode_char}}|[A-Za-z_0-9\p{L}])'
```
wbond: Do identifiers only allow `0-9`, or other numbers also?

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+  name_normal: '{{start_char}}{{other_char}}*'
+  cap_name: '([A-Z]{{other_char}})'
```
wbond: Instead of `[A-Z]` here you probably want `\p{Lu}`, unless unicode characters are not allowed. If that is the case, `\p{L}` probably shouldn't be used in the variables above.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+        1: punctuation.definition.comment.documentation.cs
+      push: documentation
+    - match: '//'
+      scope: punctuation.definition.comment.cs
       push:
-        - meta_scope: meta.namespace.source.cs
-        - match: "}"
-          scope: punctuation.section.namespace.end.source.cs
+        - meta_scope: comment.line.double-slash.cs
+        - match: $
```
wbond: Typically we want to consume the `\n` here so that certain completions are not triggered while writing characters at the end of a single-line comment

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      push:
+        - meta_scope: comment.block.cs
+        - match: '\*/'
+          scope: punctuation.definition.comment.end.cs
+          pop: true
+        - include: comments_in
+
+  comments_in:
+    - match: \b(?i:todo|hack|xxx)\b
+      scope: comment.line.todo
```
wbond: Missing a trailing `.cs`

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+    - match: '/!\\'
+      scope: comment.line.todo
```
wbond: Missing a trailing `.cs`.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+    - match: \b(https?://\S*)\b
+      scope: markup.underline.link.cs
```
wbond: This may give the false sense that the URL can be clicked, no?

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+    - match: '^(>>>>>>>).*$'
+      captures:
+        1: meta.git.merge punctuation.section.merge.end.git
+
+  preprocessor_option:
+    - meta_scope: meta.preprocessor.cs
+    - match: '(define|undef)\s+{{name}}'
+      captures:
+        1: keyword.control.preprocessor.cs
+        2: constant.numeric.cs
```
wbond: There isn't a second capture in the pattern here. Generally, variables should not have capturing groups since it is hard for authors to remember.

Either way, `constant.numeric` seems like it is probably the wrong scope.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+    - match: '((el)?if)'
```
wbond: The outer `()` can be removed and a `\b` should be put at the beginning and the end.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
-          scope: punctuation.section.block.end.source.cs
-          pop: true
-        - include: code
-  builtinTypes:
-    - match: \b(bool|byte|sbyte|char|decimal|double|float|int|uint|long|ulong|object|short|ushort|string|void|class|struct|enum|interface)\b
-      scope: storage.type.source.cs
-  class:
-    - match: '(?=\w?[\w\s]*(?:class|struct|interface|enum)\s+\w+)'
+        - match: '\(|\)'
+          scope: punctuation.definition.expression.cs
```
wbond: According to our new scope naming guidelines, this should be `punctuation.section.group` or `punctuation.section.parens`.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+        - match: '&&|\|\||!'
+          scope: keyword.operator.logical.cs
+        - match: '\b(true|false)\b'
+          scope: constant.language.cs
+        - match: '{{name}}'
+          scope: constant.numeric.cs
```
wbond: `numeric` doesn't seem right here.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
-  builtinTypes:
-    - match: \b(bool|byte|sbyte|char|decimal|double|float|int|uint|long|ulong|object|short|ushort|string|void|class|struct|enum|interface)\b
-      scope: storage.type.source.cs
-  class:
-    - match: '(?=\w?[\w\s]*(?:class|struct|interface|enum)\s+\w+)'
+        - match: '\(|\)'
+          scope: punctuation.definition.expression.cs
+        - match: '&&|\|\||!'
+          scope: keyword.operator.logical.cs
+        - match: '\b(true|false)\b'
```
wbond: Are these keywords case sensitive?

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+          scope: constant.language.cs
+        - match: '{{name}}'
+          scope: constant.numeric.cs
+        - include: option_done
+    - match: '(else|endif)'
```
wbond: Add `\b` at the beginning and end

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      scope: keyword.control.preprocessor.cs
+      push: option_done
+    - match: '(error|warning)\s*'
```
wbond: Add `\b`, and change the `scope:` to
```
captures:
    1:
```
so that the whitespace doesn't get a scope.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      scope: keyword.other.preprocessor.cs
+      push: option_message
+    - match: '(region)\s*'
```
wbond: Same comments as above, add `\b` and don't add a scope to the optional trailing whitespace.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
-              captures:
-                1: storage.type.source.cs
-        - match: "{"
-          scope: punctuation.section.class.begin.source.cs
-          push:
-            - meta_scope: meta.class.body.source.cs
-            - match: "(?=})"
+            1: entity.name.section.cs
+        - include: option_done
+    - match: '(endregion)\s*'
```
wbond: `\b` at beginning, don't scope whitespace

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      scope: storage.type.section.cs
+      push:
+        - match: ([^/\s][^/]*?)\s*(?=//|/\*|$)
+          captures:
+            1: variable.other.section.cs
+        - include: option_done
+
+    - match: '(line)\s+(default|hidden)'
```
wbond: `\b` at beginning and end

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      captures:
+        1: keyword.other.preprocessor.cs
+        2: keyword.other.preprocessor.cs
+      push: option_done
+    - match: '(line)\s*([0-9]*)\s*(".*")?'
```
wbond: As written this will match `line0"foo"` – is that correct?

At the beginning it should have a `\b`. Perhaps `\s*([0-9]*)` should be `(?:\s+([0-9]+))?` and `\s*(".*")?` should be be a lookahead for `(?=")` that pushes into a string scope.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      captures:
+        1: keyword.other.preprocessor.cs
+        2: constant.numeric.cs
+        3: string.quoted.double.cs
+      push: option_done
+    - match: '(pragma\s+checksum)\s+(".*")\s+("\{)(.*)(\}")\s+("\{)(.*)(\}")'
```
wbond: Add a `\b` at the beginning.

The regex is getting really complicated. If possible, push into a context that handles the strings instead of trying to match in a single regex. You can exit the context as soon as something isn't valid syntax.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+        2: string.quoted.double.cs
+        3: punctuation.definition.hash.begin.cs
+        4: constant.numeric.cs
+        5: punctuation.definition.hash.begin.cs
+        6: punctuation.definition.hash.begin.cs
+        7: constant.numeric.cs
+        8: punctuation.definition.hash.begin.cs
+      push: option_done
+
+    - match: '(pragma)'
```
wbond: Remove `()` and add `\b` at beginning and end

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      scope: keyword.other.preprocessor.cs
+      push: option_done
+
+    - match: '(.*)'
```
wbond: Remove the `()`

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      scope: invalid.illegal.cs
+    - match: $
+      pop: true
+
+  option_message:
+    - meta_content_scope: string.unquoted.cs
+    - include: comments
```
wbond: It would seem odd that comments should be part of `string.unquoted`. Perhaps instead you should exit if you see `//` or `/*`?

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+    # skips opening
+    - match: ^\s*
+
+    - include: tag
+    - include: using
+    - include: class_declaration
+    - include: interface_declaration
+    - include: namespace_declaration
+    - include: delegate_declaration
+    # - include: method_declaration
```
wbond: This should be deleted

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+
+    # allows coloration of code outside a class
+    - include: code_block_in
+
+  using:
+    - match: '\b(using)\s+{{name}}\s*(=)\s*'
+      captures:
+        1: keyword.control.import.cs
+        2: entity.name.type.cs
+        3: storage.modifier.namespace.as.cs
```
wbond: This should probably be `keyword.operator.assignment`, right?

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      push: using_namespace
+    - match: '\b(using)\s+(static\b)?\s*{{name}}'
```
wbond: I think this would be more clear as:

```
\b(using)(?:\s+(static))?\s+{{name}}
```

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      captures:
+        1: keyword.control.import.cs
+        2: storage.modifier.namespace.cs
```
wbond: I don't know if `.namespace` adds much of value here

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+  using_namespace:
+    - match: '{{name}}'
+      scope: variable.namespace.cs
+    - match: \.
+      scope: punctuation.separator.namespace.cs
+    - match: '<'
+      scope: punctuation.definition.arguments.type.begin.cs
+      push: type_argument
+    - match: ';'
+      scope: punctuation.definition.namespace.end.cs
```
wbond: This should be `punctuation.terminator`.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+        3: variable.namespace.cs
+      push: using_namespace
+
+  using_namespace:
+    - match: '{{name}}'
+      scope: variable.namespace.cs
+    - match: \.
+      scope: punctuation.separator.namespace.cs
+    - match: '<'
+      scope: punctuation.definition.arguments.type.begin.cs
```
wbond: Is this a generic? If so, consider using `punctuation.definition.generic`. If not, `punctuation.section` probably makes more sense.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      pop: true
+    - match: '.*$'
+      scope: invalid.illegal.expected-namespace.cs
+      pop: true
+
+  namespace_declaration:
+    # package declaration
+    - match: '\b(namespace)\b'
+      captures:
+        0: meta.package.identifier.cs
```
wbond: Stylistically, let's use `scope:` instead of `captures: 0:`.

That said, this should be `meta.namespace` and should probably be a `meta_scope` on the pushed context.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+        1: storage.type.namespace
+        2: entity.name.namespace.cs
+        3: punctuation.accessor.dot.cs
+        4: entity.name.namespace.cs
```
wbond: Where are captures 2-4?

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      push:
+        - match: '{{name}}'
+          scope: entity.name.namespace.cs
+        - match: \.
+          scope: punctuation.separator.namespace.cs
+        - match: \{
+          scope: punctuation.section.namespace.begin.cs
+          set:
+            - meta_scope: meta.namespace.cs
```
wbond: This should be removed and moved up to the parent context.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+            - match: \}
+              scope: punctuation.section.namespace.end.cs
               pop: true
-            - include: method
-            - match: '='
-              scope: keyword.operator.assignment.cs
-              push:
+            - include: main
+        - match: (?=[^\s{])
```
wbond: This can just be `(?=\S)` since `{` is already matched above

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+          pop: true
+
+  class_declaration:
+    - match: '\b(static|unsafe|abstract|partial|sealed)\b'
+      scope: storage.modifier.cs
+    - match: '{{visibility}}'
+      scope: storage.modifier.access.cs
+    - match: '\b(class)\s+{{name}}'
```
wbond: There should be a fallback somewhere (there may already be) for base `class` to pick it up before someone has typed the space or the name after it.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+      captures:
+        0: meta.class.identifier.cs
```
wbond: `scope:` instead of `captures: 0:`

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+        1: storage.type.class.class.cs
```
wbond: There is an extra `.class` here

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
 %YAML 1.2
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
-name: "C#"
+# https://github.com/gwenzek/csharpSyntax
```
wbond: This should be removed since it isn't really relevant any longer, however you are free to add something like the MIT license here with your name/email.

With a significant rewrite like this, you will likely also get a mention in the changelog, like with your work on Scala.

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
   - cs
   - csx
 scope: source.cs
+
+
+variables:
+
+  bin_op: '(?:\+|->|-|\*|/|%|\|\||&&|\||&|\^|<<|>>|=>|<=|<|>=|>|==|!=|\?\?)'
+  unary_op: '(?:\+\+|--|-|~|!|&|\*)'
+  # name: '[A-Za-z_\p{L}][A-Za-z\p{L}_0-9]*'
```
wbond: This can be removed

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+  exponent: '[eE][+\-]?[0-9]+'
+  floating_point: '([0-9]+(?:\.[0-9]+)?(?:{{exponent}})?)([fFdDmM]?)'
+
+  # characters
+  unicode_char: '\\u[0-9a-fA-F]{,4}'
+  escaped_char: '(?:\\[btnfr"''\\]|{{unicode_char}}|\\x[0-9a-fA-F]{,4}||\\[0-9]{,3})'
+  # escaped_char: '\\{{unicode_char}}'
+  # char: '(?:{{escaped_char}}|.)'
+  # string: '(f|s|raw)?\"[^\"]*\"'
+  # literal: '({{number}}|{{char}}|{{string}})'
```
wbond: These commented lines can be removed

C#/C#.sublime-syntax
```diff
@@ -1,278 +1,1634 @@
...
+
+  visibility: \b(public|private|protected|internal|protected\s+internal)\b
+  base_type: (?:bool|byte|sbyte|char|decimal|double|float|int|uint|long|ulong|short|ushort|object|string|void)
+  type_suffix: ((?:\[,*\]|\*|\?)*)
+  generic_declaration: \s*(<[^(={};<]*>)?\s*
+  reserved: '(?:abstract|as|base|break|case|catch|checked|class|const|continue|default|delegate|do|else|enum|event|explicit|extern|finally|fixed|for|foreach|goto|if|implicit|in|interface|internal|is|lock|namespace|new|null|operator|out|override|params|private|protected|public|readonly|ref|return|sealed|sizeof|stackalloc|static|string|struct|switch|this|throw|try|typeof|unchecked|unsafe|using|virtual|volatile|while)'
+  name: '(@{{reserved}}|@{{base_type}}|@var|{{name_normal}})'
+  # name: '([A-Za-z_][A-Za-z_0-9]*)'
+  # name_normal: '[A-Za-z_][A-Za-z_0-9]*'
+  # name_normal: '(?:[A-Za-z_\p{L}]|unicode_char)[A-Za-z_0-9\p{L}]*'
```
wbond: These comments can be removed

LoneBoco:
@gwenzek Consider this for regions:
```YAML
    - match: '(region)\s*'
      scope: storage.type.section.cs
      push:
        - match: (.*)\s*
          captures:
            1: entity.name.section.cs
        - match: $
          pop: true
```
Turn off the prototype so we don't parse comments, then capture everything as the section name.

C#/C#.sublime-syntax
gwenzek: Thanks, do you think a link to the MIT license is enough or shall I put it entirely ?
Also shall we remove the "C#/Readme.txt" since the information it contains is not accurate anymore ?

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: Yes, it's crazy but the C# compiler replace all '\u####" by the corresponding unicode character in the identifier.
From https://msdn.microsoft.com/en-us/library/aa664670(v=vs.71).aspx : 

> the token "cl\u0061ss" is an identifier

C#/C#.sublime-syntax
gwenzek: I read the doc and the first character should be from: Lu, Ll, Lt, Lm, Lo, or Nl,
and following from: Lu, Ll, Lt, Lm, Lo, or Nl, Mn or Mc,  Nd,  Pc ,  Cf

So it's a mess !

C#/C#.sublime-syntax
gwenzek: You're right, `\p{Lu}` it is

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: Well with the right plugin it actually becomes clickable.
It's actually common to find URLs in comments that's why I try to provide a minimal support for it.
But I can remove it if you want.

C#/C#.sublime-syntax
gwenzek: It's actually pretty useful because it allows to merge from ST with a decent highlighting.
It's kind of a hack because sometimes the merge conflict still mess up the structure of the code,
but most of the time it provides a good experience.
Ideally the syntax engine could do it itself.

C#/C#.sublime-syntax
gwenzek: I changed it to `constant.other.flag.cs`.

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: I changed it to constant.other.flag.cs.

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: changed to "keyword.control.import.cs" for the two of them.

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done for the whole file

C#/C#.sublime-syntax
gwenzek: oups

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: There is one, it receives "keyword.other"

C#/C#.sublime-syntax
gwenzek: done

C#/C#.sublime-syntax
gwenzek: done

gwenzek:
@LoneBoco thanks but I found a better way: I push a context that pops out at the end of the line or at the next comment. The context is tagged with a "meta_content_scope".

LoneBoco:
@gwenzek Well, it doesn't complain anymore, but take this example:
```C#
public class Region
{
	#region//Test/Region//
	#endregion

	#region This is my region
	#endregion
}
```
The first region doesn't get the `entity.name.section.cs` scope so it doesn't appear in the symbols list.

gwenzek:
Yes but it's a comment.
I don't have access to Visual Studio but with Monodevelop it's marked as a
comment.

C#/C#.sublime-syntax
FichteFoll: I didn't get a mention in the changelog for YAML though :cry:

wbond: @FichteFoll That was a major oversight on our part, I'm sorry about that. We initially missed randy3k on the LaTeX one also. I'll see if I can put a list of major contributors and the syntaxes they've contributed in the readme.

C#/C#.sublime-syntax
FichteFoll: I believe that this sort of functionality should be provided by plugins, not syntax definitions.

LoneBoco:
@gwenzek Monodevelop is doing it wrong, then.
```C#
public class Region
{
	#region/*/Test/Region//
	#endregion

	#region This is my region
	#endregion
}
```
This completely breaks highlighting and it is valid.  Visual Studio compiles this code just fine.
works

As you can see, the C# compiler doesn't see this as commented code:
fails

Furthermore, according to the [C# Language Specification](https://www.microsoft.com/en-us/download/details.aspx?id=7029), this is what a region is:
<pre>
<code>
pp-message:
    new-line
    whitespace   input-characters<sub>opt</sub>   new-line<br>

pp-region:
    pp-start-region   conditional-section<sub>opt</sub>   pp-end-region

pp-start-region:
    whitespace<sub>opt</sub>   #   whitespace<sub>opt</sub>   region   pp-message

pp-end-region:
    whitespace<sub>opt</sub>   #   whitespace<sub>opt</sub>   endregion   pp-message
</code>
</pre>
As you can see, `pp-start-region` and `pp-end-region` start with optional whitespace, then a # character, then more optional whitespace, then the region/endregion keyword, and finally ends with `pp-message`.

`pp-message` can either be a new-line, or it can be whitespace, followed by optional input-characters, then ends in a new-line.

`input-characters`, as specified earlier in the specs, is `Any Unicode character except a new-line-character`.

<em>Technically</em> Visual Studio is wrong as it treats the whitespace in `pp-message` as optional and allows `#regionTEST` as valid, but the point is that inside a region/endregion, comments do not exist.

C#/C#.sublime-syntax
gwenzek: Yes: https://msdn.microsoft.com/en-us/library/4y6tbswk.aspx

LoneBoco:
@gwenzek Generic method constraints are not handled and break the syntax.
https://msdn.microsoft.com/en-us/library/d5x73970.aspx
```C#
public class A
{
    public virtual void Instantiate<T>(string componentId, out T component)
        where T : BaseComponent
    {
        var test = "Test";
    }
}
```

gwenzek:
@LoneBoco thanks for the Visual Studio screenshots.
I finally found the relevant passage in the [ECMA C# Specification](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-334.pdf) (page 100)
The MSDN C# reference isn't really precise ...

> A source line containing a #define, #undef, #if, #elif, #else, #endif, or #line
> directive can end  with a single-line comment.
> Delimited comments (the /* */ style of comments)
> are not permitted on source lines containing pre-processing directives. 

Also on page 105:

> Note that a `pp-message` can contain arbitrary text; 

I'll fix that soon.

Concerning the generic method it's a regression introduced by my very last "cleanup commented code" commit ! I'm looking into it.

michaelblyons:
I am pretty excited about this getting all the way through. I use it as my primary C# syntax now, but that has illuminated a couple more frustrations.

Fragments of code containing >1 function are not highlighted correctly:

```cs
string Permute(string input)
{
    char[] charArray = input.ToCharArray();
    Array.Reverse(charArray);
    return new string(charArray);
}

// this function has no highlighting
string DontPermute(string input)
{
    return input;
}
```

This guy doesn't work as a fragment either:

```cs
List<string> Frag
{
    get
    {
        var list = new List<string>();
        list.Add("foo");
        list.Add("bar");
        list.Add("baz");
        return list;
    } // <-- This and the next "}" are marked "invalid"
}
```

Both work inside a `class{...}` context.

keith-hall:
Can we get it to scope the placeholders in `string.Format("{0} {1}", "hello", "world")` like Python does? i.e. so the `{0}` and `{1}` should be `constant.other.placeholder`.

gwenzek:
@LoneBoco the regression regarding generics has been fixed.
@LoneBoco the #region now works as described in the ECMA C# reference.
@keith-hall #646 `"Hello {0}"` is now highlighted like in Python.

@michaelblyons Your first example is working with the current version. It has been added recently.
For the second one it's a bit more complicated, I'll look into it.

michaelblyons:
@gwenzek Thanks! I notice that the first case is still weird with access modifiers.

```cs
private string Permute(string input)
{
    char[] charArray = input.ToCharArray();
    Array.Reverse(charArray);
    return new string(charArray);
}

// this function has no highlighting
private string DontPermute(string input)
{
    return input;
}
```

gwenzek:
@michaelblyons thanks, the "visibility" modifiers weren't allowed inside code, so I added them.
I principally tested my syntax with fully compiling code (for browsing), so there might be other hiccup ahead.

LoneBoco:
@gwenzek Thanks, regions and generics are looking good.
```C#
public class B
{
    public void Test()
    {
        int test = 1;
        test = (test) ? 1 : 0;
    }
}
```
Found an odd issue with the ternary operator.  It looks like it thinks the `(test)` is a cast operator.  I'm not sure how difficult that would be to fix.  The test for casting probably needs to see if we have a variable name after it, or an opening parenthesis.

gwenzek:
The cast detection is a bit nasty and only works with simple types between
parentheses.
I should probably improve it.

keith-hall:
There is a slight inconsistency for generic types:

    private Dictionary<string, object> Data;

currently, the `<` is `punctuation.definition.arguments.type.begin.cs` but the `>` is `punctuation.definition.generic.end.cs`

keith-hall:
the Nullable shortcut `?` isn't scoped in property definitions, and is scoped as `keyword.operator.ternary.cs` for casts (but I know you plan to work on casts anyway):

    public DateTime? FirstRun
    {
      get
      {
        return (DateTime?) this.GetValue("start");
      }
      set
      {
        this.StoreValue("start", (object) value);
      }
    }

michaelblyons:
Is this going to get merged sometime soon? I've been using it for months.

From time to time I find bugs (generally minor), but I hesitate adding them to this ticket because I don't want to slow it down any further.

wbond:
@michaelblyons I will try to spend some time this week going through this again. I believe I only got part-way through last time.

wbond:
I started going through this, but rather than make @gwenzek go through and apply a bunch of things I know need to change, I am going to merge it and make the changes that I know need to be made for consistency.

Once that is done I can get feedback from anyone interested about tweaks where I may have changed something I shouldn't have.

Hopefully this will be more efficient than a bunch more back-and-forth.

BTW, thanks for your consistent effort on this @gwenzek, it is much appreciated! And thank you @LoneBoco and @michaelblyons for using it and providing feedback.

wbond:
Quick question: I presume the type `int` is built-in to the language and not added by a framework?

michaelblyons:
**RE `int`:** Yes, it's a built-in.
**RE merging:** Thanks so much!

gwenzek:
Thanks @wbond for taking up that work !

I know that a lot of things still have to be done, but I haven't got much time for open source lately,
and as I don't use C# at work anymore, my motivation kinda diminished.
Thanks for all the helpful feedback and sorry for not having took the time to finish it properly.

wbond:
Ok, so I just finished updating the rewrite to be consistent with our current scope guidelines. I added a ton of test assertions, and I think I checked the various bug reports in this PR, but I may have missed some.

If you use C#, please put the new syntax through its paces, and provide snippets of any broken highlighting so we can get it fixed up for the next dev build.

PR #94, opened 2015-08-26
Enhanced Scala syntax

gwenzek:
I have a created a new syntax for Scala on a [separate repository](https://github.com/gwenzek/scalaSublimeSyntax).
The syntax is on [Package Control](https://packagecontrol.io/packages/Scala%20Syntax) and has been downloaded by several users, and so far nobody complained.

My repository include one test file ( a bit messy but with important coverage ) that you can try.


djspiewak:
Do you have a high-level overview of what is different from the existing Scala mode?

gwenzek:
Yes, it's described in the [Readme](https://github.com/gwenzek/scalaSublimeSyntax/blob/master/README.md) of my Repo. You can try it on Package Control.

buildersbrewery:
Maybe you could add some tests? Compare http://www.sublimetext.com/docs/3/syntax.html#Testing

gwenzek:
Vb. Vvr
On Jul 15, 2016 10:42, "buildersbrewery" <notifications@github.com> wrote:

Maybe you could add some tests? Compare
http://www.sublimetext.com/docs/3/syntax.html#Testing

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/sublimehq/Packages/pull/94#issuecomment-232895760>, or mute
the thread

gwenzek:
@buildersbrewery done

gwenzek:
@djspiewak Please could you describe what you don't like about my syntax.
I see you're also trying to improve the Scala syntax, I think there is a lot you could use from this PR, even if some part certainly could be simplified.

djspiewak:
1. I don't like the bracket highlighting
2. I'm strongly averse to highlighting "intrinsics" or stdlib elements in the syntax mode.  Even the `Int`/`Boolean`/etc highlighting in the default mode bugs me
3. As I mentioned on #491, I want `val`, `case` and `for` variable declarations to be highlighted uniformly with function parameters, mirroring the scoping semantics
4. You highlight `else if` differently from `else`?
5. `val`s (and equivalently, `for` and `case` bindings) are not indexed

Also, more philosophically, I don't see a need to do a clean slate rewrite rather than starting with the existing mode.  Starting with the current mode has a lower impact on other syntax-twiddling tools like Ensime.

gwenzek:
@djspiewak, Thnaks for the detailed feedback.
1. I find it pretty useful, and it's an opt-in behavior since you have to adapt your theme to see it.
2. I agree with you, I just followed the behavior used by other syntaxes.
3. I chooses to use different scopes to not put too many variable in the symbol list which become unusable
4. Strange, I'm not sure if this is voluntary.
5. Same reason than 3, I hesitated to put all of them in the symbol list, for clarity reason. I think this kind of choice should be a user setting rather than a hardcoded choice for every user, but that's not possible for the moment.

(6). Totally agree with you, I made this syntax before the Scala.tmLanguage was converted to Scala.sublime-syntax. I don't have the courage to port all the changes, and negociate feature by feature with Wbond.  I wish you good luck with your PRs and hope that one day I could use the default Scala syntax in Sublime. For me the next big things for improving the syntax are better scopes for interpolated strings, allowing Utf8 variable names, and add a different scope for operators.

gwenzek:
Since the official Scala syntax as improved so much recently, I close this PR.
The ideas still needing implementation are tracked under #588.

PR #95, opened 2015-08-26
sublime-syntax.sublime-syntax

gwenzek:
Meta syntax :-)
I tried to cover as much as possible from the Onigruma syntax.

### DIFF
--- a/YAML/sublime-syntax.sublime-syntax
+++ b/YAML/sublime-syntax.sublime-syntax
@@ -8,7 +8,7 @@ first_line_match: ^%YAML( 1.\d)?$
 scope: source.yaml
 
 variables:
-  variable: '(\{\{)([a-zA-Z_0-9])(\}\})'
+  variable: '(\{\{)([a-zA-Z_0-9]*)(\}\})'
 
 contexts:
   comment:
@@ -27,11 +27,7 @@ contexts:
       captures:
         1: keyword.other.syntax
         2: punctuation.separator.key-value
-      push:
-        - match: $
-          pop: true
-        - include: comment
-        - include: regex
+      push: regex
 
     - match: ^(contexts)(:)
       captures:
@@ -39,8 +35,10 @@ contexts:
         2: punctuation.separator.key-value
       push: contexts_
 
-    - match: ^variables
-      scope: keyword.scope.variables.syntax
+    - match: ^(variables)(:)
+      captures:
+        1: keyword.scope.variables.syntax
+        2: punctuation.separator.key-value
       push: variables_
 
     # Numbers
@@ -77,7 +75,7 @@ contexts:
         2: keyword.operator.merge-key.yaml
         3: punctuation.definition.keyword.yaml
 
-    - match: "-"
+    - match: '-'
       scope: keyword.operator.symbol
 
     - match: (#).*?(?=%>)
@@ -92,36 +90,6 @@ contexts:
     - match: '^(?=\w)'
       pop: true
 
-    # double quoted regex
-    - match: '\s*([\w_]*)\s*(:)\s*(")'
-      captures:
-        1: entity.name.function.declaration
-        2: punctuation.separator.key-value
-        3: punctuation.definition.string.begin.yaml
-      push:
-        - meta_scope: string.quoted.double.yaml
-        - match: '"'
-          captures:
-            0: punctuation.definition.string.end.yaml
-          pop: true
-        - include: regex
-
-    # simple quoted regex
-    - match: '\s*([\w_]*)\s*(:)\s*('')'
-      captures:
-        1: entity.name.function.declaration
-        2: punctuation.separator.key-value
-        3: punctuation.definition.string.begin.yaml
-      push:
-        # - meta_scope: string.quoted.single.yaml
-        - match: "''"
-          scope: constant.character.escape.yaml
-        - match: "'"
-          captures:
-            0: punctuation.definition.string.end.yaml
-          pop: true
-        - include: regex
-
     # block regex
     - match: '(\s*)([\w_]*)\s*(:)\s*(>-)\s*$'
       captures:
@@ -132,20 +100,16 @@ contexts:
         # - meta_content_scope: string.unquoted.block.yaml
         - meta_content_scope: meta.regex.unquoted.syntax
         - include: comment
-        - include: regex
+        - include: regex_unquoted
         - match: '^(?!^\1)|^(?=\1(-|\w+\s*:)|#)'
           pop: true
 
-    # unquoted regex
-    - match: '\s*([\w_]*)\s*(:)\s*'
+    # inline regex
+    - match: '\s*([\w_]+)\s*(:)\s*'
       captures:
         1: entity.name.function.declaration
         2: punctuation.separator.key-value
-      push:
-        - match: $
-          pop: true
-        - include: comment
-        - include: regex
+      push: regex
 
   contexts_:
     - meta_scope: meta.contexts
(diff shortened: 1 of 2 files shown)


FichteFoll:
I have a few issues with that syntax.

I worked on a new yaml syntax definition in #90 and the lack of block scalars here is severe. Furthermore, quoted strings and their escaape mechanisms are not accounted for and plain scalar termination rules (which are not that trivial, see more in `YAML.sublime-syntax`) neither. Oniguruma regex support could also be improved and should be split into its own definition, as is [here](https://github.com/SublimeText/AAAPackageDev/blob/3386ee4277c391a89d255dfac041d035a2c1a64b/Syntax%20Definitions/Oniguruma%20Regular%20Expression.YAML-tmLanguage).

Also, I acutally intended to work on a syntax def for this on my own, since I also maintain AAAPackageDev and this seemed like a good fit, but I wanted to wait for my YAML def to be merged first because I intended to re-use some contexts I defined there (and had different things to do afterwards).

Furthermore, I'm interested in a stance from @sublimehq regarding special syntax highlighting of its own configuration formats in the shipped packages, including all the JSON formats. In the past they were not obviously (which is the reason why PackageDev exists). It also seems that package development here has stalled because pull requests haven't been merged or reviewed in a while.

gwenzek:
Thanks for your feedback @FichteFoll .

I'll work on scalar blocks.
I didn't give it too much thoughts because most of the time I quote my regexes.
I see you use a lot of 

     ns_tag_prefix: |- 
        (?x:
            !              {{ns_uri_char}}*
          | (?![,!\[\]{}]) {{ns_uri_char}}+
        )

in your YAML syntax, and they aren't correctly highlighted. I think simple forms like that with grouping parenthesis could be easily added.
I didn't split the regex syntax because YAML double quoted strings changed the behavior of escaped character and therefore change the way the regex should be highlighted. Maybe it's do-able and it would be nice to have the regex syntax usable somewhere else.

But I think it would be interesting to have a syntax for `.sublime-syntax`, more precise than a YAML syntax in order to match the few keywords allowed and the regexes.
I checked it on several `.sublime-syntax` and the regexes look good, could you be more precise ? 

I'd be happy to contribute to `AAAPackageDev` if @sublimehq think it's a better place to include such syntax.

FichteFoll:
>I checked it on several .sublime-syntax and the regexes look good, could you be more precise ?

What exactly do you mean here?

gwenzek:
You said:

>    Oniguruma regex support could also be improved 

So I was hoping that you could show me some examples of poorly highlighted code.

FichteFoll:
I prepared [this test file](https://github.com/SublimeText/AAAPackageDev/blob/3386ee4277c391a89d255dfac041d035a2c1a64b/tests/regexp.ore) for myself a while ago. It will serve as the base for syntax tests if I eventually start working on this again, but I want to wait a bit first due to the reasons outlined above.

gwenzek:
Thanks, I'll have a look at it and come back.

gwenzek:
Hi @FitcheFoll, I used your test set and improved the regexes.

I also made a few improvements on character classes and scope lists.

gschenck:
Highlighting is broken for 
- match: '(?<![\d.])\b0[xX][a-fA-F\d\.]+([pP][\-\+]?\d+)?|\b\d+(\.\d+)?([eE]-?\d+)?|\.\d+([eE]-?\d+)?'.

Problem is "-?"

gwenzek:
Hi @gschenck thanks for your report.
It looks fine to me, what's is the result you expected ?

I think the `?` in this context is an operator but the `-` is a character.
Is  `-?` having a special meaning in this context for Oniguruma regex ?

gschenck:
Please have a look at https://github.com/sublimehq/Packages/blob/master/Lua/Lua.sublime-syntax,
highlighting stops at line 20.

gwenzek:
@gschenck sorry I was testing with a more recent version of the syntax. I commited the new version here.

gschenck:
Works. Thank you.

YAML/sublime-syntax.sublime-syntax
```diff
@@ -453,9 +453,9 @@ contexts:
   variable_call:
     - match: '(\{\{)([a-zA-Z_0-9]*)(\}\})'
       captures:
-        1: punctuation.definition.parameters.begin.syntax
+        1: punctuation.definition.arguments.begin.syntax
```
keith-hall: for my use, I added the line: `0: variable.parameter` so that variables would be highlighted by the color scheme and thus be easier to spot :)

YAML/sublime-syntax.sublime-syntax
```diff
@@ -0,0 +1,532 @@
...
+    - match: '^(\s*)(captures)\s*(:)\s*'
+      captures:
+        2: keyword.control.flow.syntax
+        3: punctuation.separator.key-value
+      push:
+        - meta_content_scope: meta.capturing.yaml
+        - include: comments
+        - match: '^(\s*)(\d+)\s*(:)'
+          captures:
+            2: support.variable
```
keith-hall: for my use, I changed it from `support.variable` to `constant.numeric`, so that it would be highlighted by the color scheme :)

gwenzek:
Thanks @keith-hall for your suggestions, I added them in.
I also added a test file.
It was working fine a few build ago but with current build  (3111) the test file is not picked up by Sublime.

If any one has an idea on how to do this I'll be glad :smiley:

YAML/syntax_test_sublime-syntax.sublime-syntax
```diff
@@ -0,0 +1,44 @@
+## SYNTAX TEST "Packages/User/scalaSublimeSyntax/sublime-syntax".sublime-syntax"
```
wbond: This line should be `## SYNTAX TEST "YAML/sublime-syntax.sublime-syntax"`.

That is why your tests aren't being picked up.

YAML/sublime-syntax.sublime-syntax
```diff
@@ -0,0 +1,532 @@
...
+  - sublime-syntax
+first_line_match: ^%YAML( 1.\d)?$
+scope: source.yaml
+
+variables:
+  variable: '(\{\{)([a-zA-Z_0-9]*)(\}\})'
+  escaped_char: \\(?>[0atnvfre "/\\N_Lp]|x[0-9a-fA-F]{1,2}|x\{[0-9a-fA-F]{1,8}\}|u\d{4}|U\d{8}|c\d+|C-\d+|M-\d+|M-\\C-\d+)
+
+contexts:
+
```
keith-hall: one more suggestion :) to correctly scope the yaml directive (the very first line of the file) using `- include: scope:source.yaml#directive`

FichteFoll:
As a note, I added my own sublime-syntax highlighting to PackageDev, which has a lot of other goodies for syntax (test) development too. It's still in "alpha" because I want to finish all the other syntaxes as well, but it's very usable and an absolute must-have for anyone who works on syntax definition (not an under-statement).

See https://forum.sublimetext.com/t/looking-for-packagedev-alpha-testers-sublime-syntax-highlighting/24270 for the announcement thread

deathaxe:
As PackageDev is indeed a very helpful package not only for syntax development but handling all the different sublime-text-file-formats by providing functionalities which go far beyond syntax highlighting, I wonder whether this PR is still required to be open.

bumfo:
> As PackageDev is indeed a very helpful package not only for syntax development but handling all the different sublime-text-file-formats by providing functionalities which go far beyond syntax highlighting, I wonder whether this PR is still required to be open.

I think PackageDev shouldn't be required for basic syntax editing, maybe including their sublime-syntax.sublime-syntax with default Packages is an alternative.

FichteFoll:
> I think PackageDev shouldn't be required for basic syntax editing

It isn't. There is a syntax definition for YAML that you can easily use to edit syntaxes (and apparently wbond only uses that), so it's perfectly possible to edit syntax definitions without PackageDev. However, I wouldn't recommend it and installing a third-part package is arguably trivial.

Furthermore, it allows us to provide more functionality that goes beyond highlighting, as @deathaxe mentioned, like context completions, scope name completions and the syntax test stuff with a much higher release frequency than would be possible with having the package in the default packages. Of course, it would still be nice if ST shipped syntax definitions for its config files, but as the maintainer and main developer of PackageDev, I enjoy the freedom in deployment and independence I currently have.

bumfo:
> I think PackageDev shouldn't be required for basic syntax editing
> 
> It isn't. There is a syntax definition for YAML that you can easily use to edit syntaxes (and apparently wbond only uses that), so it's perfectly possible to edit syntax definitions without PackageDev. However, I wouldn't recommend it and installing a third-part package is arguably trivial.
> Furthermore, it allows us to provide more functionality that goes beyond highlighting, as @deathaxe mentioned, like context completions, scope name completions and the syntax test stuff with a much higher release frequency than would be possible with having the package in the default packages. Of course, it would still be nice if ST shipped syntax definitions for its config files, but as the maintainer and main developer of PackageDev, I enjoy the freedom in deployment and independence I currently have.

I appreciate the convenience provided by PackageDev, and the sublime-syntax.sublime-syntax from PackageDev is also full-featured. However, not everyone is looking for full support, nor all the other awesome features from PackageDev, when only basic editing of sublime-syntax is needed. 

Currently YAML isn't an alternative, as it doesn't support indexing. I think a default syntax with only basic functions should be shipped by default, and only when more features are needed, install the package. 

Installing a Package is, yes, trivial, but the overhead of installing features not needed causes many troubles & conflictions, which can all be avoided by including only necessary files.

jfcherng:
I would consider editing a `.sublime-syntax` file as advanced thing for a user.

Another reason why I don't like things to be added as default: It just updates SUPER slow due to ST's release cycle. Sure, people can manually update it locally but installing a package is far more simple...

I am not against approving this PR because I can still and will use `PackageDev` whatever.

FichteFoll:
The PR is basically a meta question whether to include a syntax definition at all at this point. At least that's how I perceive it. However, I'm not the person to decide here. 

PackageDev's syntax definition is not self-contained, because it uses the separate syntax definitions for Oniguruma regular expressions and the generic JSON-focused context library I built for many ST configuration files. They could of course be embedded or included, but it's not as simple as just taking a single file and be done with it.

While I would of course be okay with ST including a syntax definition for syntax definition by default, it then raises the question about the other files like settings or build system. And, because of what @jfcherng said, I would still want to develop PD in parallel because of quicker release cycles and the fact that many extra features provided through plugins are strictly dependent on tokenization of the files with the syntax definitions. I would also need to check how well they override each, which is based on the lexicographical order of the package names. 

@bunfo, maybe you could clarify on your areas of concern regarding conflicts of the features that PD provides?

wbond:
I think for right now it makes sense that https://packagecontrol.io/packages/PackageDev be the place for augmented syntaxes.

The release cadence is separate, there is a group of people who are enthusiastic about what it provides, and the average user does not get into writing syntaxes. If they do, the default YAML syntax does _work_, I mean, I use it. 😄

We may revisit this at some future point, but for now I am going to close.

gwenzek:
Good for me :-) I totally forgot about this PR.

PR #97, opened 2015-09-02
Fixes regex parse fail of bracket array syntax.

MarianoGappa:
The language file for PHP doesn't seem to support the "new" bracket syntax for arrays introduced on PHP 5.4 (http://php.net/manual/en/migration54.new-features.php).

Fortunately, the colouring still mostly works anyway; except when a function argument is typehinted.

This regex change appears to be successful in fixing it on my tests.

##### Before
Before

##### After
After

Note that I've tried the changes on the tm version; not the sublime one, but the regexes are the same.


rtheunissen:
:+1:

jbrooksuk:
:+1:

james2doyle:
+1

wbond:
Thanks for this! I fixed a YAML quoting issue and added your example as tests in 9ba46df9b0943380d5fb19d936780b6cc71b71cc.

PR #100, opened 2015-10-05
Added `puts` to the list of `keyword.other.special-method.ruby`

aziz:


wbond:
Hmm, so `puts` seems to be a method of [`Kernel`](http://ruby-doc.org/core-2.2.3/Kernel.html), as are many other methods.

Should all `Kernel` methods be `keyword.other.special-method.ruby`?

aziz:
I think it makes sense. I'll update this pull request if there's no objection to this idea.

wbond:
I added all previously unsupport methods from the `Kernel` and `Module` class in de66f5553206da67f06d8ed8b8b1e23d73c91ded

PR #101, opened 2015-10-13
[HTML] Improve completions for empty elements

philipbordallo:
## [HTML] Improve completions for empty elements

Some [empty elements](https://developer.mozilla.org/en-US/docs/Glossary/empty_element) are completing with a closing element.

For example `area` is currently completing to:

```html
<area></area>
```

When it should be completing to:

```html
<area>
```

This will fix completions for: `<area>`, `<col>`, `<base>`, `<meta>`, `<hr>`, `<br>`, `<wbr>`, `<input>`.

`meta`'s completion will render as:

```html
<meta charset="utf-8">
      ^-------------^
       Selected Text
```

Which means a `tab` keypress will jump to after the `>`, completing the example or a `delete` keypress will remove the example and allow for another attribute input. This way it's easier to input something other than just the default.

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -256,14 +256,14 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
         """
         default_list = []
         normal_tags = ([
-            'abbr', 'acronym', 'address', 'applet', 'area', 'article', 'aside',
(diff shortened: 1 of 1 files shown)


kleinfreund:
:+1:

wbond:
Like it, thanks!

PR #105, opened 2015-10-20
Adding new C++11 keywords and raw strings

jbw3:
Highlighting the following:
- C++11 keywords
- Raw strings
- Macros \_\_func\_\_, \_\_FUNCTION\_\_, \_\_LINE\_\_, and \_\_FILE\_\_

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -27,7 +27,7 @@ contexts:
       scope: storage.modifier.c++
     - match: \b(private:|protected:|public:)
       scope: storage.modifier.c++
-    - match: \b(catch|operator|try|throw|using)\b
+    - match: \b(catch|noexcept|operator|static_assert|try|throw|using)\b
       scope: keyword.control.c++
     - match: '\bdelete\b(\s*\[\])?|\bnew\b(?!])'
       scope: keyword.control.c++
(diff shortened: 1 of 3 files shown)


jfcherng:
Maybe the plugin [C++11](https://packagecontrol.io/packages/C%2B%2B11) could be take into consideration.

txtsd:
What's the holdup on this PR?

wbond:
@txtsd There are no tests included, which makes reviewing it more complicated. If you wish to contribute, you could write some tests against the current syntax to ensure this doesn't cause regressions. The readme for this repo has more info.

txtsd:
I'll take a look at it!

jbw3:
I've added tests for my changes as well as some regression tests.

FichteFoll:
Shouldn't the string prefixes be scoped as something like storage.type rather than punctuation? I can't compare with how it is in python at this moment, but it is certainly not just punctuation.

jbw3:
I've updated the string prefixes per @FichteFoll's request.

wbond:
Awesome, thanks for taking the time to add the tests and tweak the prefixes to be similar to Python's!

wbond:
I ended up rebasing these on top of master for a cleaner history. Thanks again!

PR #110, opened 2015-11-05
[Haskell] Add UnicodeSyntax symbols for type definitions

kerscher:
The GHC Haskell compiler [allows the use of unicode symbols](https://downloads.haskell.org/~ghc/7.2.1/docs/html/users_guide/syntax-extns.html) for some keywords when the UnicodeSyntax extension is enabled.

This commit adds support for the following symbols:
* `→` for `->`
* `⇒` for `=>`
* `∷` for `::`

The other characters demand deeper changes on the syntax definition but could be added. The symbols above comprise everything needed for type declarations, though.

### DIFF
--- a/Haskell/Haskell.sublime-syntax
+++ b/Haskell/Haskell.sublime-syntax
@@ -161,7 +161,7 @@ contexts:
         4: constant.character.escape.hexadecimal.haskell
         5: constant.character.escape.control.haskell
         6: punctuation.definition.string.end.haskell
-    - match: '^\s*([a-z_][a-zA-Z0-9_'']*|\([|!%$+\-.,=</>]+\))\s*(::)'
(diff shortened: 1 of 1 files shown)


wbond:
Could you provide a simple example of usage of these syntax forms so that we can put together a test that includes these new features?

kerscher:
Sure.
```haskell
(>$<) ∷ (Functor t) ⇒ t a → (a → b) → t b
t >$< f = f <$> t

(>*<) ∷ (Applicative t) ⇒ t a → t (a → b) → t b
t >*< f = f <*> t
```
Both are shown correctly highlighted here. This code compiles fine as well.

wbond:
Thanks for this. I've merged it in and adapted the tests in 9d2c09112a89fe349cc392623c2739b1fc02a2cb to ensure this doesn't revert in the future.

PR #111, opened 2015-11-08
[RegEx] Anchor highlighting bug fix

jbw3:
Fixing a bug where the ^ and $ anchors were incorrectly highlighted only when preceded with a backslash


wbond:
Thanks for this contribution. I added a test for it in 61b70589b614b6d98dab70a16f68a473f41b24c9.

Issue #40, opened 2015-07-01
The D Syntax Highlighter Does Not Recognize A Number Of Keywords


JackStouffer:
Many keywords in D are not recognized by the default D syntax highlighter and are displayed as regular text. Here is a list of all of the keywords that I have encountered that are not recognized:

* string
* shared
* __gshared
* ref
* immutable
* @safe
* @property
* @trusted
* @system
* @nogc
* pure
* nothrow
* ~~assert~~

~~While the last one is technically a function, the Python highlighter recognizes the assert statement and it would be nice if it was in D as well.~~

Also, it does not recognize long literals like `1_000_000_000` as numbers.

FichteFoll:
The Python `assert` is actually a keyword and designates a statement.

JackStouffer:
@FichteFoll you're right; I guess it could be left how it currently is. 

I should also mention that while `@safe @property @trusted @system @nogc` are all checks enforced by the compiler, there can also be user defined function attributes using the `@` syntax, much like Python. In a perfect world, these five native function attributes would have a different coloring than user defined function attributes, but that would be a purely a "nice to have" feature.

JackStouffer:
Also `synchronized` is marked as a type rather than a keyword.

wbond:
@JackStouffer

As I discovered while looking into #112, the D syntax is rather poor. In addition to the keywords you've found, the detection of methods gets tripped up on `typeof()` and `const()`. Additionally, in some situations `else if ()` is detected as a method definition.

I spent a little time working around some of those issues, however I haven't worked in D very much, so I'm not a very good candidate for making extensive changes.

If you are interested in contributing to the syntax, I can post my current tweaks. Ideally changes should include a test file to ensure we don't regress over time. Here is some more info about the format: http://www.sublimetext.com/docs/3/syntax.html.

wbond:
I've fixed a number of things in b84e922b408a8b6191d3c1ce6be8bc57f75f4594. If you check out the latest version and can provide some samples of any remaining failures so that I can add a test for them, I should be able to get them fixed up.

JackStouffer:
@wbond Cool! How can I use this in my version of sublime so I can test it?

wbond:
@JackStouffer There are some instructions at https://github.com/sublimehq/Packages#installation which should help.

JackStouffer:
Thanks for doing this.

This code sample shows all of the remaining problems:

```
__gshared int a = 5_000;
[...]

PR #112, opened 2015-11-12
Show D functions in the symbol list

Yoplitein:
Why was this disabled to begin with?

### DIFF
--- a/D/Symbol List-Method-Constructor.tmPreferences
+++ b/D/Symbol List-Method-Constructor.tmPreferences
@@ -9,7 +9,7 @@
 	<key>settings</key>
 	<dict>
 		<key>showInSymbolList</key>
-		<integer>0</integer>
+		<integer>1</integer>
 		<key>symbolTransformation</key>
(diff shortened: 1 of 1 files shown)


wbond:
From spending a little time with the D grammar, it seems it is of fairly low quality. Enabling the symbol list results in all sorts of incorrect junk showing up.

I spent a little time with an enhanced version from Package Control, which fixed some issues, however there were still issues detecting many methods, specifically those with typed defined via `typeof()` and `const()`/`const()*`.

Yoplitein:
Ah, I didn't even think to check the symbol list with the built in grammar. That is pretty bad! Perhaps the enhanced syntax could be merged in? I believe it fixes a lot of the issues mentioned in #40 too.

wbond:
In b84e922b408a8b6191d3c1ce6be8bc57f75f4594 I modified the syntax definition a bit to fix a number of glaring issues. If you see any more errors, please open a new issue and provide a code sample that shows the error.

Issue #32, opened 2015-06-29
Standard Package Structure


pradyunsg:
> This is really a broader version of #8, extending it to the whole package. The discussion there was headed this way and this is really a topic that deserves a separate issue of it's own. So, here it is.

The idea is that Packages in Sublime Text should be arranged in a standard way which may be optionally/partially enforced, like in other text-editors like TextMate 2 or Atom.

Since having too many files in the root directory of the Package creates a lot of noise, the various types of files should be moved into sub-directories in the package. While reducing noise in the root directory, it also makes the Package arrangement feel more organized.

~~Things worth more discussion (IMO):~~
Some initial questions regarding the structure:
 - Is there a need for a "/tests" directory?  
   Can tests for the Python code be placed in "/lib/tests" and syntax tests be placed in "/syntaxes/tests" instead?
 - How many .py files should be there in the root package directory?  
   Should it be only one, with the same name as Package ({name}/{name}.py) and it should include everything that it wishes Sublime Text (and other plugins?) to know about...
 - Naming scheme for the sub-folders.  
   Should they be CamelCase or snake_case or hypenated-names?

/cc @jappen @FichteFoll @aziz @jbrooksuk @gerardroche @sublimehq

---

### Initial Draft
(Based on @jrappen's [original suggestion](https://github.com/sublimehq/Packages/issues/8#issuecomment-113288614))
```/Package
    /build
        /linux
            *
        /osx
            *
        /windows
            *.exe
        *.sublime-build
    /color-schemes
        *.tmTheme
        *.hidden-tmTheme
    /commands
        *.sublime-commands
    /keymaps
        *.sublime-keymap
    /lib
        *.py
    /macros
        *.sublime-macro
        *.tmMacro
    /menus
        *.sublime-menu
    /messages
        install.txt
        *.txt
    /metadata
        *.tmPreferences
    /mousemaps
        *.sublime-mousemap
    /settings
        *.sublime-settings
    /snippets
        *.sublime-completions
        *.sublime-snippet
    /syntaxes
        *.tmLanguage
        *.hidden-tmLanguage
        *.sublime-syntax
    /tests
        *.*
    /templates
        *.*
    /themes
        *.sublime-theme
    .gitignore
    .gitattributes
    .no-sublime-package
    *.py
    CHANGELOG.md
    CONTRIBUTING.md
    LICENSE.md
    README.md
```

gerardroche:
[...]

PR #116, opened 2016-01-16
fix sublimehq/Packages#32

jrappen:
moved files, no source code changed


jrappen:
Issues still remaining:

* Sublime Text tries opening `LaTeX.tmLanguage` instead of `LaTeX.sublime-syntax`
* Sublime Text tries opening `Diff.tmLanguage` instead of `Diff.sublime-syntax`

compare how Sublime Text tries opening `Text.tmLanguage` instead of `Text.sublime-syntax` (see #61). This makes using `GoTo > Go to symbols in Project` unusable. Maybe @sublimehq could consider enabling Sublime Text to check for `*.sublime-syntax` for `Text`, `Diff` and `LaTeX`?

jrappen:
Any updates regarding this PR?

FichteFoll:
It is a hard-to-review pull request (many files touched) and the changes made, which we discussed in #32, need to be agreed upon by the maintainers first. I expect this not to happpen too soon.

PR #117, opened 2016-01-18
issue 172 - ensure assignment RHS is handled as code

StrangeNoises:
Ref: https://github.com/SublimeTextIssues/DefaultPackages/issues/172 which has screenshots etc.

This commit ensures the RHS of an assignment statement is handled as per the code: context, preventing mis-identification of e.g.: method *calls* as if they were method *definitions* when outside of a method.

Also removes a special-case exemption for the keyword `new` as a method name in a method definition, as the new, more conceptually consistent, way should make that unnecessary.

### DIFF
--- a/Java/syntax_test_java.java
+++ b/Java/syntax_test_java.java
@@ -5,9 +5,9 @@ public class SyntaxTest {
     private String memberString2 = new String("Hello");
     private String memberString3 = String.valueOf("Hello");
 //                                ^ meta.assignment.rhs.java
+//                                                 ^ string.quoted.double.java
     private int memberLpos = memberString3.indexOf("l");
 //                                                     ^ punctuation.terminator.java
-    private String memberClassPath = System.getProperty("java.class.path");
 
     public static void main(String... args) {
         String[] strings = new String[5];
--- a/Java/syntax_test_java.java
+++ b/Java/syntax_test_java.java
@@ -1,6 +1,14 @@
 // SYNTAX TEST "Java.sublime-syntax"
 
 public class SyntaxTest {
+    private String memberString = "Hello";
+    private String memberString2 = new String("Hello");
+    private String memberString3 = String.valueOf("Hello");
+//                                ^ meta.assignment.rhs.java
+    private int memberLpos = memberString3.indexOf("l");
+//                                                     ^ punctuation.terminator.java
+    private String memberClassPath = System.getProperty("java.class.path");
+
     public static void main(String... args) {
         String[] strings = new String[5];
 //                         ^ keyword.control.new.java
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -223,6 +223,11 @@ contexts:
       scope: keyword.operator.comparison.java
     - match: (=)
       scope: keyword.operator.assignment.java
+      push:
+        - meta_scope: meta.assignment.rhs.java
+        - match: ;
+          pop: true
+        - include: code
     - match: (\-\-|\+\+)
       scope: keyword.operator.increment-decrement.java
     - match: (\-|\+|\*|\/|%)
@@ -234,7 +239,7 @@ contexts:
     - match: ;
       scope: punctuation.terminator.java
   methods:
-    - match: '(?!new)(?=\w.*\s+)(?=[^=]+\()'
+    - match: '(?=\w.*\s+)(?=[^=]+\()'
       push:
         - meta_scope: meta.method.java
         - match: "}|(?=;)"
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -226,6 +226,7 @@ contexts:
       push:
         - meta_scope: meta.assignment.rhs.java
         - match: ;
+          scope: punctuation.terminator.java
           pop: true
         - include: code
     - match: (\-\-|\+\+)


StrangeNoises:
Typically soon after posting this, i see a minor improvement I would make to it, but not sure how in terms of github procedures to do so.

Basically for that new `meta.assignment.rhs.java` meta-scope, to ensure its terminating semicolon is correctly marked as per the `punctuation.terminator.java` scope. i.e.:

```yaml
    - match: (=)
      scope: keyword.operator.assignment.java
      push:
        - meta_scope: meta.assignment.rhs.java
        - match: ;
          scope: punctuation.terminator.java
          pop: true
        - include: code
```

The effect is presumably only visible if that punctuation scope has different syntax colouring. For my part, I only noticed because the way github presented the previous commit happened by chance to juxtapose my change with the *general* definition of that scope, right at the end of the keywords section. Without this additional change, when a semicolon appears at the end of an assignment (as it should always do) it is not recognised/marked as being `punctuation.terminator.java`, and with the change, it is.

For now this change remains uncommitted in my local working copy.

wbond:
To add more commits to a pull request, you just push them to the branch that you started the pull request from.

Could you also add a syntax test (or augment the existing one) to help ensure we don't regress in the future? You can see documentation about how to at the bottom of https://www.sublimetext.com/docs/3/syntax.html. There is an existing test file for Java named `syntax_test_java.java`.

StrangeNoises:
hm. ok, i committed & pushed that change, and, separately, the syntax test change.

Trying to build the syntax test however just says "Success: 3 tests passed". It said that before I started, when there were only two tests, and continues to say it now when there are four, so I don't know what's going on there. :-}

wbond:
That success message is somewhat confusing at first (it confused me also). It indicates that 3 different syntax test files ran successfully, not the number of assertions in the current test file.

You can check and make sure your test is running by changing one of the scope names and seeing if it fails.

StrangeNoises:
confirmed.

Java/syntax_test_java.java
```diff
@@ -1,6 +1,14 @@
 // SYNTAX TEST "Java.sublime-syntax"
 
 public class SyntaxTest {
+    private String memberString = "Hello";
+    private String memberString2 = new String("Hello");
+    private String memberString3 = String.valueOf("Hello");
```
wbond: Perhaps add a test here to ensure the string scope is correct for `"Hello"` since you identified that in your original bug report?

StrangeNoises:
done. It's pointing at the start of the string literal value itself. When it was pointing at the actual starting double-quote mark, one character back, it still passed the test; presumably because the string.quoted.double.java context is in force even though it's not the last scope in the list, that being, correctly I'm sure, `punctuation.definition.string.begin.java`.

Java/syntax_test_java.java
StrangeNoises: done. It's pointing at the start of the string literal value itself. When it was pointing at the actual starting double-quote mark, one character back, it still passed the test; presumably because the string.quoted.double.java context is in force even though it's not the last scope in the list, that being, correctly I'm sure, `punctuation.definition.string.begin.java`.

(edit: add this as a line comment instead)

wbond:
Awesome, thank you for the fixes and the tests to go with it!

StrangeNoises:
damn, i just found a problem with this. Relates to try-with-resources declarations for the last or only such declaration, i.e.:

```java
try (Stream lines = Files.lines(path)) {
    lines.forEach(System.out::println);
}
```

it matches the assignment RHS but lacking a semicolon at the end of that expression, it goes wrong thereafter. most obviously, subsequent methods aren't recognised as such. I think the RHS terminator needs to be able to identify the complete *expression* that follows, rather than trusting to the semicolon to mark the end of the *statement*. Got to think about how/if that can be done in regex.

wbond:
Could you open a new issue about this? I haven't looked at it in-depth, but my hunch is that perhaps the `try` block has a more restricted grammar and we can piece together rather than including all of `code`?

StrangeNoises:
yes i was starting to think something along the same lines, failing a way to reliably "consume" an expression without recourse to some kind of terminating sequence.

FichteFoll:
The other issue doesn't exist yet, so: You can actually just match the closing `)` in a lookahead to terminate the `meta.assignment.rhs.java` context (or rather the context it pushes). Just make sure that this special match is included before the one that you tweaked in this PR and it should never match.

wbond:
@FichteFoll Further issues were identified in #118 and then solved in #119 by limiting the scope of the changes to a class body outside of a method.

FichteFoll:
:no_mouth:

StrangeNoises:
i did initially try to terminate `meta.assignment.rhs.java` as you suggest, but it didn't work. I don't know if it *can't* work though and if it's just my regex-and-yaml-fu that's lacking. But the closing `)` seemed in some fashion to be consumed by its role in terminating the `parens:` context and appeared not to be available to the match expressions i was trying to use to terminate the `meta.assignment.rhs.java` code.

FichteFoll:
@StrangeNoises take a look at https://forum.sublimetext.com/t/syntax-highlight-how-to-match-already-matched-text/16354. I suspect this is what you were running into.

StrangeNoises:
Yes, that does work.

ie: In context `keywords:` I now have:

```yaml
    - match: (=)
      scope: keyword.operator.assignment.java
      push:
        - meta_scope: meta.assignment.rhs.java
        - match: (?=(;|\)))
          pop: true
        - include: code
```

(Still want to match the semicolon too, but with this trick we don't need to identify it ourselves as `punctuation.terminator.java`, because it isn't consumed, it gets identified as such where it should.)

I removed the `class-body-keywords:` context entirely, and of course removed it from the `class-body:` context too.

This seems to me a simpler solution, more minimal, and more logically "correct", in that `meta.assignment.rhs.java` is applied consistently, which may be a desirable trait longer-term. OTOH that also means the work of applying it is done everywhere, not just outside methods. SyntaxTest still passes unchanged, but then I also added another check to indicate that it is being applied correctly inside the try-with-resources declaration.

Do you want a PR with this?

FichteFoll:
I think it makes more sense syntax-wise to only match allowed statements in a class body, which most keywords do not fall under afaics. Thus, only matching the allowed set of statements in a class body and not "the entire thing", i.e. all keywords, is definitely clearer.

Assignments in method blocks, i.e. actual code, are of course also a thing, but they are much broader there. I don't exactly care how you would perform this, but this pattern should be applied to both of these contexts separately.

StrangeNoises:
well, the `class-body-keywords:` context wasn't there before all this, it was invented by me simply because I needed at the time to *avoid* matching `meta.assignment.rhs.java` inside methods, which in turn was only because my regex-fu wasn't good enough to have figured out the lookahead solution for RHS expressions that weren't terminated by a semicolon.

With the lookahead fix meaning I could match `meta.assignment.rhs.java` consistently everywhere, the `class-body-keywords:` solution became unnecessary, and that being the only reason I'd made it, I reversed it out.

In fact the more I think about it the more `class-body-keywords:` turns out to be wrong: Having it there included in `class-body:` didn't prevent the keywords in `keywords:` applying *too* (as `keywords:` is included in `code:`). `class-body-keywords:` was just providing an override assignment operator behaviour, overriding the one in `keywords:` so it would consume the expression before the `methods:` regex could mistakenly grab it. All other keywords would have remained valid as defined in `keywords:`.

In fact the naming is already a bit tortured: There are operators inside the `keywords:` context, including the assignment operator, and there are Java language keywords *not* listed there, but in e.g.: `storage-modifiers:` elsewhere which perhaps *should* just be a matched scope in `class-body-keywords:`. (And for instance I also added `default` to them, because it can appear as a modifier for non-abstract interface methods since Java 8, as well as being the none-of-the-above label in a switch statement).

Which is all to say that the more I look at it the more I can see *could* stand to be fixed, made more rational or correct according to the language, but sorting all that out would seem to be a bigger cleanup/rewrite task than I ever intended, and the latest fix described two comments up does IMHO represent a minimal but consistent fix to the original problem I was trying to fix. That all keywords were allowed in `class-body:` is a situation that existed before I started; this would leave it unchanged and leave this fix focused on solving the original problem. Maybe it's more appropriate to raise the desire for separate sets of keywords being available in different contexts as its own issue?

FichteFoll:
I can not say any more about this without actually looking into the syntax definition and I do not have the time to do that.

StrangeNoises:
think the best i can do is to put up a PR with the latest change which I feel is the right, and minimalist solution to the exact problem at hand. I'll include a diff of all the changes, including those of the previous two PRs so it's clear.

Issue #118, opened 2016-01-18
Java syntax: RHS termination not detected without semicolon


StrangeNoises:
PR #117 (from me, today) introduces a problem with try-with-resource syntax:

![screen shot 2016-01-18 at 17 43 24](https://cloud.githubusercontent.com/assets/1295001/12398347/014b86f4-be0b-11e5-898f-e7cfe9daeb6d.png)

That PR attempts to solve a problem with out-of-method assignment code being incorrectly parsed as a method definition. In so doing it creates a new meta-scope of `meta.assignment.rhs.java` which should encompass the expression to the right of the equals sign in an assignment.

There's a situation where this can go wrong. First discovered with respect to try-with-resource declarations, where the last, or only, declaration is not terminated by a semicolon. Thus:

```java
try (Stream<String> lines = Files.lines(path)) {
    lines.forEach(System.out::println);
}
```

It can also affect for-loops, but only if the final clause has an assignment. This trips it:

```java
for (int i = 0; i < 10; i+=2) {
    System.out.println(i);
}
```

Ideally we'd want a reliable way of "consuming" an expression without needing a terminator, but that may not be doable in regex. If the problem can be confined to `try (...)` and `for (...)` that might offer some more restricted-scope solutions.

wbond:
I am thinking the simplest way around this is to tweak the new push you introduced in #117 to only happen for assignments outside of a method body. I presume that `try` and `for` don't work outside of a method, correct?

StrangeNoises:
pretty much; they can conceivably happen in a static initialiser (valid but probably more honoured in the breach...) but I think the surrounding curly braces already put that in `code:` (Earlier, I'd tried surrounding the private and static declaration/assignments with curly braces and that fixed the syntax colouring although of course it wasn't correct java.)

I'm ok for doing it that way but I don't know how - i.e.: to do what I did for an = that's in `meta.class.body.java` but not in `meta.method.body.java`

Of course what I was originally trying was in `class-body:` and in `methods:` to just prevent it catching e.g.: `String.valueOf(` as a method definition as it's only a method *call*. But I couldn't make that work without killing its ability to find real method definitions.

wbond:
Completed by #119

PR #119, opened 2016-01-18
Fix for issue #118. class-body-keywords context for subset of keyword matches outside methods.

StrangeNoises:
Fix for issue #118: Restrict the use of the `meta.assignment.rhs.java` scope to the `class-body:` context, by having a separate `class-body-keywords:` context containing a restricted set of keyword matches for use in class-body only (as opposed to generally in code). Then the assignment operator in *there* can have the `meta.assignment.rhs.java` scope and the assignment operator in `keywords:` can revert to what it was before.

Currently `class-body-keywords:` only contains matches for the assignment operator and the semicolon terminator as none of the other matched keywords should appear outside of methods *except* in an assignment RHS.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -143,6 +143,7 @@ contexts:
     - include: methods
     - include: annotations
     - include: storage-modifiers
+    - include: class-body-keywords
     - include: code
   code:
     - include: comments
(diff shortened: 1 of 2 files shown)


wbond:
Excellent, thanks for this fix!

FichteFoll:
That is actually the proper approach :+1:

StrangeNoises:
it's not just a name difference; it's there to match the subset of symbols in `keywords:` that should appear in `class-body:` which is more than just assignment-equals, even if right now all I could see and think of that ought to go in besides assignment-equals was the semicolon terminator. I'm basically covering myself for when I realise what else I've forgotten in the next day or so. 😀

PR #120, opened 2016-01-20
use lookahead to properly identify RHS

StrangeNoises:
This represents what I think should be the final solution to the original problem.

For context, I'm including a diff showing the changes in the previous two PRs #117 and #119 as well so we can see the summarised difference of all the changes I've made as one. I wouldn't be surprised if there's a way to do this in github itself but for now I'll just paste the diff, for illustration purposes:

```diff
diff --git a/Java/Java.sublime-syntax b/Java/Java.sublime-syntax
index bf086fa..bff61b2 100644
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -223,6 +223,11 @@ contexts:
       scope: keyword.operator.comparison.java
     - match: (=)
       scope: keyword.operator.assignment.java
+      push:
+        - meta_scope: meta.assignment.rhs.java
+        - match: (?=(;|\)))
+          pop: true
+        - include: code
     - match: (\-\-|\+\+)
       scope: keyword.operator.increment-decrement.java
     - match: (\-|\+|\*|\/|%)
@@ -234,7 +239,7 @@ contexts:
     - match: ;
       scope: punctuation.terminator.java
   methods:
-    - match: '(?!new)(?=\w.*\s+)(?=[^=]+\()'
+    - match: '(?=\w.*\s+)(?=[^=]+\()'
       push:
         - meta_scope: meta.method.java
         - match: "}|(?=;)"
@@ -326,7 +331,7 @@ contexts:
     - match: \b(?:void|boolean|byte|char|short|int|float|long|double)\b
       scope: storage.type.primitive.java
   storage-modifiers:
-    - match: \b(public|private|protected|static|final|native|synchronized|strictfp|abstract|threadsafe|transient)\b
+    - match: \b(public|private|protected|static|final|native|synchronized|strictfp|abstract|threadsafe|transient|default)\b
       captures:
         1: storage.modifier.java
   strings:
diff --git a/Java/syntax_test_java.java b/Java/syntax_test_java.java
index db4649d..2cc793f 100644
--- a/Java/syntax_test_java.java
+++ b/Java/syntax_test_java.java
@@ -1,13 +1,30 @@
 // SYNTAX TEST "Java.sublime-syntax"
 
 public class SyntaxTest {
+    private String memberString = "Hello";
+    private String memberString2 = new String("Hello");
+    private String memberString3 = String.valueOf("Hello");
+//                                ^ meta.assignment.rhs.java
+//                                                 ^ string.quoted.double.java
+    private int memberLpos = memberString3.indexOf("l");
+//                                                     ^ punctuation.terminator.java
+
     public static void main(String... args) {
         String[] strings = new String[5];
+//                        ^ meta.assignment.rhs.java
 //                         ^ keyword.control.new.java
         printList(Arrays.stream(args)
             .collect(Collectors.toCollection(ArrayList::new)));
 //                                                      ^ meta.method.body.java - keyword.control.new.java
         anotherMethod();
+        try (Stream<String> lines = Files.lines(path)) {
+//                                 ^ meta.assignment.rhs.java
+//                                                   ^ meta.method.body.java - meta.assignment.rhs.java
+            lines.forEach(System.out::println);
+        }
+        for (int i = 0; i < 10; i+= 2) {
+            System.out.println(i);
+        }
     }
 
     private static void printList(List<String> args) {
```

edited to add links to the two previous PRs, for the sake of future comprehension.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -140,7 +140,6 @@ contexts:
     - include: comments
     - include: class
     - include: enums
-    - include: class-body-keywords
     - include: methods
     - include: annotations
     - include: storage-modifiers
@@ -224,6 +223,11 @@ contexts:
       scope: keyword.operator.comparison.java
     - match: (=)
       scope: keyword.operator.assignment.java
+      push:
+        - meta_scope: meta.assignment.rhs.java
+        - match: (?=(;|\)))
+          pop: true
+        - include: code
     - match: (\-\-|\+\+)
       scope: keyword.operator.increment-decrement.java
     - match: (\-|\+|\*|\/|%)
@@ -232,17 +236,6 @@ contexts:
       scope: keyword.operator.logical.java
     - match: (?<=\S)\.(?=\S)
       scope: keyword.operator.dereference.java
-    - match: ;
-      scope: punctuation.terminator.java
-  class-body-keywords:
-    - match: (=)
-      scope: keyword.operator.assignment.java
-      push:
-        - meta_scope: meta.assignment.rhs.java
-        - match: ;
-          scope: punctuation.terminator.java
-          pop: true
-        - include: code
     - match: ;
       scope: punctuation.terminator.java
   methods:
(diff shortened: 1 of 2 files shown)


wbond:
When I first viewed the pop pattern, I was concerned that nested function calls would break this. However, I can see there is a `parens` context which nicely ensures nested structures don't run into such issues.

Nice fix!

PR #121, opened 2016-01-22
[XML] support Unicode, more extensions, minor fixes

MattDMo:
Unicode:
* `a-zA-Z` replaced with `[:alpha:]`
* `a-zA-Z0-9` replaced with `[:alnum:]`

New extensions: plist, tmLanguage, tmTheme, tmPreferences, sublime-snippet

Minor fixes:
* closing tag of `<?xml version="1.0" encoding="UTF-8"?>` is now highlighted
* closing tag of `<!DOCTYPE ... >` is now highlighted

Old version:

Old version

New version:

New version

### DIFF
--- a/XML/syntax_test_XML.xml
+++ b/XML/syntax_test_XML.xml
@@ -1,0 +1,14 @@
+<!-- SYNTAX TEST "Packages/XML/XML.sublime-syntax" -->
+<?xml version="1.0" encoding="UTF-8"?>
+<!--                                ^ punctuation.definition.tag.end.xml -->
+
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!--                                                                                                 ^ punctuation.definition.tag.end.xml -->
+
+<tagname attr="stuff">Contents</tagname>
+<!-- ^ text.xml meta.tag.xml entity.name.tag.localname.xml  -->
+<!--      ^ text.xml meta.tag.xml entity.other.attribute-name.localname.xml -->
+
+<таĝñäᴹə ατţř="șƬűʃ⨍">Contents</таĝñäᴹə>
+<!-- ^ text.xml meta.tag.xml entity.name.tag.localname.xml  -->
+<!--      ^ text.xml meta.tag.xml entity.other.attribute-name.localname.xml -->
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -8,23 +8,30 @@ file_extensions:
   - dtml
   - rss
   - opml
+  - plist
+  - tmLanguage
+  - tmTheme
+  - tmPreferences
+  - sublime-snippet
 first_line_match: '^<\?xml '
 scope: text.xml
 contexts:
   main:
-    - match: '(<\?)\s*([-_a-zA-Z0-9]+)'
+    - match: '(<\?)\s*([-_[:alnum:]]+)'
       captures:
         1: punctuation.definition.tag.begin.xml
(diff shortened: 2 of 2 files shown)


wbond:
Could you provide a test for these tweaks?

MattDMo:
I added a fairly basic test file that passes without problems in Build 3095 on OS X. Let me know if you need anything more complex.

wbond:
Just an FYI, I'll need to do some performance testing to make sure that these changes don't have a big performance impact on large files. Many users tend to open large XML, HTML and JS files, so we just need to be a little more mindful of the performance of these syntaxes.

wbond:
In 3098 we introduced a new build variant of the Syntax Tests build system called "Performance". This will measure the time it takes to highlight the current buffer, using the current syntax, averaged over 10 runs.

If you feel like doing some tests of the current XML syntax and your proposed changes with some large files, that would certainly be helpful. If not, I will perform those tests at some point and post the results here.

MattDMo:
I didn't even see new builds were released, thanks for the heads up! I'll find some larger XML files and parse them to add some UTF-8 tags and whatnot, and let you know how the tests look.

XML/syntax_test_XML.xml
```diff
@@ -0,0 +1,14 @@
...
+<!--                                ^ punctuation.definition.tag.end.xml -->
+
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!--                                                                                                 ^ punctuation.definition.tag.end.xml -->
+
+<tagname attr="stuff">Contents</tagname>
+<!-- ^ text.xml meta.tag.xml entity.name.tag.localname.xml  -->
+<!--      ^ text.xml meta.tag.xml entity.other.attribute-name.localname.xml -->
+
+<таĝñäᴹə ατţř="șƬűʃ⨍">Contents</таĝñäᴹə>
```
keith-hall: the `ᴹ` in this `таĝñäᴹə` example is actually invalid according to Firefox and Python's `lxml` module.

wbond: @keith-hall That character appears to be http://www.fileformat.info/info/unicode/char/1d39/index.htm, which according to https://www.w3.org/TR/REC-xml/#d0e804 is valid both as the start of a name, and in the less restrictive case of part of a name that is not the first character.

Either way, I think handling of unicode XML tag names is going to depend on the library/software used. Some characters may be supported by some, and not by others, but we should highlight anything that has a semblance of being valid.

wbond:
For future reference, I found some large XML files at ftp://ftp.nlm.nih.gov/nlmdata/sample/toxspec/

XML/syntax_test_XML.xml
keith-hall: Makes sense - I agree :) thanks for looking into it - its interesting to me that not all implementations follow the spec ;)

keith-hall:
there is also a public 3.5 MiB file here http://www.dbis.informatik.uni-goettingen.de/Mondial/mondial.xml

wbond:
I ended up adding unicode support in fe943080e895bfbeedd5f6277dbacef33474b7ab, during which I was fixing false-positive matches on tag names. That commit used variables to reduce duplication

I tested with a 3MB XML file and there was no difference in performance. Due to some other changes in a recent dev build, in 1229bf0b9e64b7002b1c03c7dc2e3e97f226a8e9 I was able to remove the backreference which ended up improve performance by around 20% with the test file I was using.

The extensions you had added here are currently handled by the `hidden_extensions` setting in the XML.sublime-settings file.

I also improved `DOCTYPE` handling in some other recent commits, and added detection of some other invalid constructs.

PR #122, opened 2016-01-24
add info about overriding packages on Windows

guillermooo:
Junctions can be used to selectively override packages.


wbond:
Awesome, thanks!

PR #124, opened 2016-01-26
Change scope of CSS3 color names

kleinfreund:
Currently, all CSS3 color names are treated as deprecated and thus being highlighted as such in Sublime Text. Since these keywords are widely supported in all browser for a while now ([CSS3 Colors on caniuse.com](http://caniuse.com/#feat=css3-colors)) it’s time to recognize them as regular CSS color names.

- Current scope: `invalid.deprecated.color.w3c-non-standard-color-name.css`
- Proposed change: `support.constant.color.w3c-extended-color-keywords.css`

I just changed the scope and the comment (the referenced link 404’s anyway), because they’re not part of the 2.1 standard.

What do you think? Have I missed something?


wbond:
Could you add a syntax test file to confirm these changes and (hopefully) prevent regressions in the future? See the Testing section of http://www.sublimetext.com/docs/3/syntax.html for more info.

kleinfreund:
@wbond Sure, added the file. After lecture I’ll test this on my machine. Have to setup a separate Sublime Test with the Packages repository first.

kleinfreund:
I accidentally created the file in the wrong location. I'm on it. Sorry for the mess.

kleinfreund:
I accidentally created the syntax test file in the wrong location. Move it to the CSS directory now. Sorry for the mess.

@wbond Is there a way I can run the syntax tests locally?

wbond:
Yes, while you have the `syntax_test_*` file open, use *Tools > Build*, or cmd/ctrl+b.

kleinfreund:
Got it.

```
Success: 2 assertions in 1 files passed
[Finished]
```

Are those tests sufficient? Let me know if I missed something.

wbond:
Looks good, thanks!

PR #127, opened 2016-01-29
C# highlight the keyword 'partial'

hickford:


wbond:
Can you update `syntax_test_c#.cs` with an example of the usage and a check for the appropriate scope?

PR #128, opened 2016-02-03
Merged snippet files into completion files and cleaned up repository

jrappen:
This pull request:

* replaces `...` with `…` where possible (i.e. not in syntax files).
* merges snippet files for each language into a single completions file.
* cleanes up the main structure (compare #32)

Still needs merging of snippets for `Ruby`, `Rails` and `Groovy`. Let me know if you're interested.


jrappen:
When also done for the remaining languages this could bring the total file count across the whole repo down a couple hundred files.

FichteFoll:
Most of these should probably remain as snippets. As a general rule, if it contains a newline then it should probably be a snippet and not a completion, in a sense that completions only *complete* the word you are typing and potentially add some punctiation (e.g. parenthesis), whereas snippets are used to construct entire statements like for loops or conditionals.

There is no clear guideline on this, but this is how I feel about it and how I would describe it in a "how to write a package" guide.

wbond:
I just wanted to comment here and say that I've seen your PRs @jrappen, however I don't currently have an opinion about the structure changes, and I haven't discussed it with Jon yet.

My instinct for big changes like these is that we are going to break pretty much 100% of everyone's package overrides. I don't know if the tidiness is worth it. I feel like to accomplish file moves, we'd have to write a script to migrate all of the overrides to the new locations. But then in term of converting snippets to completions, wouldn't be lose the ability for overrides?

The other downside of big changes like this is that the PRs are super-brittle since any fix or tweak that is made to an individual package causes a conflict.

jrappen:
@wbond 

Take your time. I'll still need to merge the Rails snippets anyhow. The reason for this change is the changes proposed in #131. Due to single `*.sublime-completions` files being easier to maintain and functionally being the same as `*.tmSnippet`s, I started here with the streamlining first.
Tidiness eases maintenance as well. I am with you when you say that it might break the overrides for some, but if there actually was a **clear as well as strict** base structure that people could follow as a guidance (compare #32), then applying changes manually or programatically for changed paths wouldn't necessarily be such a great hassle, at least IMHO. On another note, the number of applied overrides should *hopefully* decrease over time while quality improves.
That's my thoughts for now.

FichteFoll:
>Due to single *.sublime-completions files being easier to maintain and functionally being the same as *.tmSnippets

If JSON supported multi-line strings or YAML was used (which also doesn't need escape sequences), I would agree.

jrappen:
@wbond done

jrappen:
Also compare the old structure...

old

with the new one...

new

Generated via https://github.com/veniversum/git-visualizer/

jrappen:
somehow related to https://forum.sublimetext.com/t/sugestion-ui-sidebar-improvements/18028 this PR brings down the number of files loaded by ST a lot, this will help performance

jrappen:
Let me know in case you would like to merge but I should redo this PR from a more recent commit as base.

jrappen:
@wbond any input on how to proceed from here?

wbond:
I'm not convinced that merging all snippets into completions files is a good idea since it will break existing overrides, and there isn't really a huge benefit, from what I can tell. Additionally, such sweeping changes will break pretty much all existing PRs.

Currently my focus is on making functional fixes and improvements, so changes like this are low priority by comparison. I'm not saying that we won't possibly make some such changes in the future, but for now my focus is on other things.

jrappen:
Fair enough. Let me know if you need help.

FichteFoll:
I just want to note again that snippets and completions are conceptually and functionally not the same. 
Conceptually, completions are supposed to "complete" what you want to type (mostly words) while snippets are used for recurring syntax constructs that may easily span multiple lines, which is something that completions do not. 

Functionally, snippets get prioritized if a prefix matches the entire tab trigger but have less priority than anything else otherwise. Snippets are also added to the command palette (with a "Snippets:" prefix) and much  easier to maintain when they contain line breaks because completions defined in json (and not yaml).

jrappen:
Ok. Should I close this then?

---

Any plans for a `*.sublime-snippets` file type which holds multiple snippets (like the `*.sublime-completions` file can do for completions)?

Maintenance would be easier and Sublime Text would have to load hundreds of files less.

---

My main goal anyhow remains sublimehq/Packages#131, which is basically the same idea like the one @FichteFoll talks about in sublimehq/Packages#297 - it makes following different styles easier. Especially when switching back and forth between multiple styles.

wbond:
I would be opposed to multiple snippets in a single file based on usability of overrides. Sublime Text 3 has a really nice override system, but it is file-based. Merging snippets into a single file means users who want to override a single snippet would have to miss out on any additions, or follow this repo and merge changes into their override.

From speaking with Jon, we are both on board with moving snippets into a subfolder named "Snippets". Currently any other organizational changes are still at a discussion phase.

PR #130, opened 2016-02-05
Update CSS syntax test to include all color property values

kleinfreund:
I added some test cases to the syntax test file and fixed the related things in the Sublime syntax file after running the build with the open syntax file.

**Changes to `syntax_test_css.css`:**

- Check for transparent and currentColor scopes
- Check for rgb, rgba, hsl, hsla scopes

**Changes to `CSS.sublime-syntax`:**

- For `color-values`, add a new match and scope for `transparent` and `currentColor`; Remove the two property values from the generic property value matches
- Split rgb and hsl matches because HSL colors cannot use the same regex

As for why HSL colors are different, see this comparison:

```css
.test {
    color: rgba(66%, 66%, 66%, 0.22);
    /*          ^ percentage in [0%,100%] OR numeric value in [0,255] */
    color: hsla(120, 50%, 50%, .77);
    /*          ^ arbitrary numeric value (no percentages) */
}
```

The first value in an HSL color is the *hue*. This is never a percentage but a degree represented by numerical values. Also, the values are not only in the range from 0° to 359°, they can overflow.<sup>[[w3.org]](https://www.w3.org/TR/css3-color/#hsl-color)</sup>


kleinfreund:
I forgot to test the most commonly used color value (hexadecimal). They’re added now. Also, I made the regular expression for matching `currentColor` case insensitive, because of the capitalized `C` which can easily be missed.

FichteFoll:
I tested this in FireFox and it seems that FF does not care about casing of the `currentColor` special value. It also accepts upper-case colors like `BLACK` so maybe CSS is case-insensitve? I don't know if that's in the spec though.

kleinfreund:
I observed the same. My point is just that this is a somewhat special case. Now that I think of it, it's maybe better to match `current[cC]olor`.

kleinfreund:
The spec at [syntax and basic data types – characters and case](https://www.w3.org/TR/CSS2/syndata.html#characters) states:

> All CSS syntax is case-insensitive within the ASCII range (i.e., [a-z] and [A-Z] are equivalent), except for parts that are not under the control of CSS.

wbond:
Thanks for all of the work on making the tests cover a number of different variations!

PR #132, opened 2016-02-08
Added array_column auto-complete

ahsankhatri:
There was a missing function for PHP in auto-complete named 'array_column' which is added now.


wbond:
Thanks!

PR #135, opened 2016-02-09
Harbour Syntax

lailton:


wbond:
As mentioned at https://github.com/sublimehq/Packages#new-packages, we aren't accepting new packages right now.

However, this syntax would probably make a great addition to https://packagecontrol.io/.

Issue #134, opened 2016-02-09
PHP .php.inc extension


KvanSteijn:
For all my projects i used .php.inc extension. Current version of Sublime doesn't support this extension.
See for more information about this extension: http://stackoverflow.com/questions/7129842/what-is-inc-and-why-to-use-it

kopeboy:
Also PHP files with other extensions (like .module - Drupal related) are not recognised and I can't even select the PHP syntax from the View menu!
So I basically get NO highlighting.. :-1:

wbond:
For the time being, you should be able to use the menu *View > Syntax > Open all with current extension as…*.

@kopeboy What happens when you try to manually select the PHP syntax?

kopeboy:
The problem was that I couldn't select the PHP syntax (option disabled, like many other formats).
That was weird, actually I can't replicate now :/ 

Ok then, Sublime wasn't recognising PHP automatically on .inc and .module files, but doing View > Syntax > Open all with current extension as…. works. Thank you.

jfcherng:
I remember that there is a way in a synatx file to set syntax depending the first line of the file.

wbond:
Yeah, we chose not to switch `.inc` files over to PHP since ASP uses that extension also. Unfortunately, the HTML syntax will be picked up even if we add `.php.inc` as an extension to PHP, I imagine due to the fact that HTML comes first alphabetically.

@jfcherng just submitted 9aadfea7adfeaaa6f80cbdc262ebe6664109b509, which will cause files starting with `<?php` to use the PHP syntax. Hopefully this should solve the issue for most people, and then others can use *Open all with current extension as…*.

PR #136, opened 2016-02-09
[PHP] Also set syntax to PHP if the first line starts with <?php

jfcherng:
As the title.
Related issue: https://github.com/sublimehq/Packages/issues/134

### DIFF
--- a/PHP/PHP.sublime-syntax
+++ b/PHP/PHP.sublime-syntax
@@ -9,7 +9,7 @@ file_extensions:
   - phps
   - phpt
   - phtml
-first_line_match: '^#!.*(?<!-)php[0-9]{0,1}\b'
+first_line_match: '^(#!.*[^-]php[0-9]?|<\?php)\b'
 scope: embedding.php
 contexts:
   main:


PHP/PHP.sublime-syntax
```diff
@@ -9,7 +9,7 @@ file_extensions:
   - phps
   - phpt
   - phtml
-first_line_match: '^#!.*(?<!-)php[0-9]{0,1}\b'
+first_line_match: '^(#!.*(?<!-)|<\?)php[0-9]{0,1}\b'
```
wbond: Right now this will match `#!/usr/bin/php`, `#!/usr/local/bin/php5` and `<?php`, but it will also match things like `<?php5`.

Perhaps we make the branch more explicit, and remove the lookbehind:

```
^(#!.*[^-]php[0-9]?|<\?php)\b
```

jfcherng:
@wbond Yes, updated.

wbond:
I tested this locally and it seems to work as intended.

hanoii:
I was looking into this and wanted to try the new commit. So I cloned the repo, did as per the homepage, linked it to where it says but I don't see picking up a php format on a .install file for example that starts with <?php.

I tried removing Cache/Packages/PHP but nothing changes. How do I know the new package files are being picked up before knowing if this is not actually working for me?

Using latest 3103.

BTW, @wbond extremely happy you are working on Sublime, will soon become a paying customer.

wbond:
@hanoii When I first tried it, I also ran into that issue, however after restarting it seemed to go away. We may have to debug a little more to figure out why this is happening.

hanoii:
Restarting doesn't make a difference, I don't know if the new syntax is not picked up or it's simply failing for some reason. If you need help debugging let me know. Is there any easy way of knowing what file is being picked up?

FichteFoll:
`sublime.load_resource(view.settings().get('syntax'))` should do, probably.

hanoii:
@FichteFoll ok, that somehow seems to work.

Ok, I can confirm the PHP syntax file is the latest one, but for some reason, the first_match_line is not really working.

ON .inc extension, that extension is added to the HTML syntax, I wonder if that's correct and I also wonder if the extension  takes precedence over the first_line_match, I think it should be the opposite.

I also have .install and .module files that are not assigned to any syntax and the first_line_match is definitely not working. The regex seems to work when used within sublime.

PR #139, opened 2016-02-09
[Python] Support for SnakeMake

karel-brinda:
Add support for main SnakeMake file name. SnakeMake (https://bitbucket.org/snakemake/snakemake/wiki/Home) is Python-based Make-like workflow management system.


wbond:
Thanks!

PR #144, opened 2016-02-09
Add missing keywords goto and const

mquigley-ck:
You wouldn't believe it, but `goto` and `const` are reserved keywords in Java even though they are illegal. 

Reference:
https://en.wikipedia.org/wiki/List_of_Java_keywords 
http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.9

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -215,7 +215,7 @@ contexts:
       scope: keyword.control.catch-exception.java
     - match: '\?|:'
       scope: keyword.control.java
-    - match: \b(return|break|case|continue|default|do|while|for|switch|if|else)\b
+    - match: \b(return|break|case|continue|default|do|while|for|switch|if|else|goto)\b
       scope: keyword.control.java
     - match: \b(instanceof)\b
       scope: keyword.operator.java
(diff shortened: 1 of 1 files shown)


wbond:
These should probably receive a scope of illegal also then, correct?

Once you've made that change, could you also add a test to ensure those are are scoped properly?

FichteFoll:
I didn't test, but I don't think they are actually illegal and just do "nothing".

djspiewak:
They're both rejected by the compiler, though with differing error messages.  They are invalid in all scopes.  Either an `illegal` scope assignment or `keyword` would make sense, though `illegal` is probably saner since neither `goto` nor `const` will *ever* be implemented by the compiler (they're just permanently reserved).

mquigley-ck:
These keywords are simply flagged as errors by the Java compiler. Therefore, they do nothing.

There is a bytecode opcode for _jump_ or _goto_ in the JVM bytecode language; however, you can't use the Java keyword `goto` in order to generate those.

djspiewak:
> These keywords are simply flagged as errors by the Java compiler. Therefore, they do nothing.

I think it is useful though to highlight them, either as keywords (which is what most IDEs do), or as illegal, since it is invalid to use them as identifier names.

wbond:
In d2eec42faa13383d70aaf3adba806f5682d0cf71 I made these marked as invalid.

Issue #145, opened 2016-02-09
[C++] Syntax highlighting different for code in C vs C++, build 3103


digikata:
For this code, the syntax highlighter correctly highlights when applied as C syntax, but misses the highlight for the last "#endif" when treating it as C++.

  #ifndef IGUARD_
  #define IGUARD_
  struct foo;
  struct foo*  alloc_foo();
  #endif

I'm running build 3103,  Linux 64bit on Ubuntu 14.04.3.

wbond:
This was resolved with PR #151

PR #151, opened 2016-02-10
[C++] Fix broken highlight after a function returning a struct (Issue #145)

TheClams:
Fix issue #145 (broken highlight after a function returning a struct)

Also include a syntax test file, to show the issue is now fixed.


wbond:
Awesome, and thank you for the test!

PR #155, opened 2016-02-10
[XML] Add missing scopes for captures

Briles:
Previously, the XML syntax was capturing tag ends for XML Declaration, DOCTYPE, and Entity Declaration but not applying any scopes to them. This patch corrects these occurrences by applying a `punctuation.definition.tag.end` scope before the contexts are popped.


wbond:
Could you add a test for this?

Briles:
I added a test that should cover the current XML syntax including the changes I made.

wbond:
Awesome, and thanks for taking the extra time with the test!

PR #156, opened 2016-02-11
Add filetypes to php7

vagkaefer:
Add filetypes to php7 and add adicional content to the auto completions.


wbond:
The whitespace in the completions seems inconsistent.

Also, are those completion parameters you added valid in older versions of PHP also?

vagkaefer:
Really I left some space incorrectly. Already removed them.

Regarding the functions:

- Bzcompress: These parameters are supported by these versions: PHP 4> = 4.0.4, PHP 5, PHP 7

- Gzdecode: PHP 5> = 5.4.0, PHP 7

wbond:
All of the other completions with multiple parameters have spaces between, so your changes should too.

The PHP versions sound pretty reasonable.

PR #157, opened 2016-02-11
[ShellScript] Reserved Word Kebab Case Highlight Fix

bmarcaur:
Previously the ShellScript highlighting behaved oddly when you used "reserved" words in a non-string token, e.g. function name, while using a "-" as the delimiter aka kebab case. For some reserved words like "while", it simply highlighted the word differently than its sibling and in other cases, like "for", it completely broke highlighting for the rest of the file.

Previous:
screen shot 2016-02-11 at 12 12 58 am

Current:
screen shot 2016-02-11 at 1 18 26 am

### DIFF
--- a/ShellScript/Shell-Unix-Generic.sublime-syntax
+++ b/ShellScript/Shell-Unix-Generic.sublime-syntax
@@ -111,7 +111,7 @@ contexts:
           pop: true
         - include: main
   function-definition:
-    - match: '\b(function)\s+([^\s\\]+)(?:\s*(\(\)))?'
+    - match: '(?<![-/])\b(function)\s+([^\s\\]+)(?:\s*(\(\)))?'
       captures:
         1: storage.type.function.shell
         2: entity.name.function.shell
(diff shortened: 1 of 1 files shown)


jfcherng:
I use [a more aggressive way](https://github.com/jfcherng/Packages/commit/7abcf6fb1e60d758476b78e0fec5dff8a7ec13b5) to fix this as in this [PR](https://github.com/jfcherng/Packages/commit/7abcf6fb1e60d758476b78e0fec5dff8a7ec13b5).

bmarcaur:
@jfcherng Do you want me to just close this PR in lieu of yours?

bmarcaur:
@jfcherng I also noticed that your syntax-test didn't explicitly hit any of my cases. What are your thoughts on committing the test file that you are working against, in your PR, and then adding my extra cases?

jfcherng:
@bmarcaur 
Yes, it's in the test case.

And it also works for your cases.


I don't think my PR is acceptable at this moment. We have to write down requirements for what we need instead of just saying a syntax highlighter.

bmarcaur:
@jfcherng Awesome, I also saw that you added the test cases to your [test](https://github.com/jfcherng/Packages/blob/fix-shell/ShellScript/syntax_test_shellscript.sh#L215). Unless you have any objections I am gonna go ahead and close this PR.

PR #163, opened 2016-02-11
[SQL] Add WITH as a sql keyword

danielhfrank:
Hi SublimeHQ! I'm aware that this is a little deficient with respect to pull request guidelines, but I didn't see a sql syntax test to start from. I defer to greater experts than myself on SQL or Sublime as to whether I've done this right, but my goal was to highlight the word "with" as a keyword, and that appears to be working locally over here.


wbond:
Great, thanks!

PR #164, opened 2016-02-12
[Rails] Rails 4 new filter action

berniechiu:


wbond:
Can you add a test, as mentioned at https://github.com/sublimehq/Packages#pull-requests?

wbond:
I believe this should be resolved by the merge PR #168

PR #165, opened 2016-02-12
[JavaScript] Renames Quasi literal to Template literal

subhaze:
**Renames Quasi literal to Template literal**  
  - Renamed literal-quasi definition to literal-string-template
  - Renamed all occurences of quasi in scopes to template
  - Added test cases
  - Scope name changes

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -74,7 +74,7 @@ contexts:
     - include: literal-prototype
     - include: literal-regexp
     - include: literal-number
-    - include: literal-quasi
+    - include: literal-string-template
     - include: literal-string
     - include: literal-language-constant
     - include: literal-language-variable
@@ -778,26 +778,27 @@ contexts:
       scope: punctuation.terminator.statement.js
     - match: ","
       scope: meta.delimiter.comma.js
-  literal-quasi:
+  literal-string-template:
     - match: '([a-zA-Z$_][\w$_]*)?(`)'
       captures:
-        1: entity.quasi.tag.name.js
-        2: punctuation.definition.quasi.begin.js
+        1: variable.function.tagged-template.js
+        2: punctuation.definition.string.template.begin.js
       push:
-        - meta_scope: string.quasi.js
+        - meta_scope: string.template.js
         - match: "`"
           captures:
-            0: punctuation.definition.quasi.end.js
+            0: punctuation.definition.string.template.end.js
           pop: true
         - include: string-content
         - match: '\${'
           captures:
-            0: punctuation.quasi.element.begin.js
+            0: punctuation.definition.template-expression.begin.js
           push:
-            - meta_scope: entity.quasi.element.js
+            - meta_scope: meta.template.expression.js
+            - meta_content_scope: source.js.embedded.expression
             - match: "}"
               captures:
-                0: punctuation.quasi.element.end.js
+                0: punctuation.definition.template-expression.end.js
               pop: true
             - include: expression
   literal-regexp:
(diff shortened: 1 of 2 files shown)


subhaze:
Forgot to add [JavaScript] at the beginning of commit, I can fix commit message if needed.

wbond:
Just so I don't get this wrong, could you provide an example of the syntax this highlights? I presume it is ES6. I'd like to add it to the tests so that we are building up a "contract" over time.

subhaze:
... Sorry I meant to look at the tests and add if something was missing/broke from the change. I'll have to look later, but no prob, will do!

subhaze:
A reference link for Templates https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings

subhaze:
@wbond updated PR to include additional test case.

subhaze:
Looking at this it may be better to change `entity.template.tag.name` to `variable.function.template.tag.name.js` since it is a function name. This way it's not just the same color as the string, but still keeps its meaning. Thoughts?

subhaze:
Now that I think about it, it would probably be good to make the template literal multiline in the test. I'm about to be AFK for a bit but will try to add that into the test case soon.

FichteFoll:
Ah, I made sure to check the open issues for this before opening my own, but I didn't catch this one.

FWIW, I reported this issue with some more detail and scope name suggestions in https://github.com/sublimehq/Packages/issues/169:

>
Currently, string interpolation in the form ``let str = `I'm a ${"string"}.`;`` has the following issues:

> 1. the meta name is `string.quasi`. I think this should be `string.interpolated.backtick`
1. backtick punctuation gets the scope names `punctuation.definition.quasi.begin`. Imo it should be `punctuation.definition.string.begin.interpolated` (and `end` respectively)
1. the pattern denotation is marked as `punctuation.quasi.element.begin`. I suggest `punctuation.definition.begin.section.interpolated` or similar.
1. the entire pattern is scoped as `entity.quasi.element`. Entitiy is definitely wrong here. I'm unsure atm what should be used instead (maybe `keyword`), but something with `embedded` would be a good start.

@subhaze, please consider these. If "template" is the official terminology, I can take that as well. I definitely recommend using `punctuation.definition.string.*` however.

subhaze:
@FichteFoll I agree on the 'element' bits completely. Since we're embedding expressions I've adjusted to scope names for this are to reflect that. Thought I'm not sure about `template.expression.embedded.js`

I'm pretty new to scope naming standards so more feedback is very welcomed.

Also, I was going to update the tests to check for multiline (it currently works) but it seems that the syntax tests wont allow for this as far as I can tell. In these template literals you can't break up multilines with comments... so... not sure how to test for multilines other than just adding two separate tests.

subhaze:
@wbond in hindsight I felt I shouldn't have moved this down to literal-string, just so it's easier to see what was actually changed and to help reduce any kind of merge conflicts that may happen later if this PR is brought it.

This PR should now provide a better story in the visual diff of what has actually changed.

FichteFoll:
I would:

- entity.name.function.template.tag.js → variable.function.tagged-template.js,
because it is a function call, not a function definition
- punctuation.definition.template.begin.js → punctuation.definition.string.template.begin.js (and end),
because we are inside a `string` scope and the selector `string punctuation.definition.string` (which I use for all strings) would match then.
- punctuation.template.expression.begin.js → punctuation.definition.template-expression.begin.js (and end),
because currently only `punctuation.definition` and `punctuation.separator` are commonly known
- `meta_scope: template.expression.embedded.js` → `meta_scope: meta.template.expression.js`
and add `meta_content_scope: source.js.embedded.expression` *or* embedded.js.expression (inspired by PHP syntax),
because only the content is javascript and these two variants have been used in other syntaxes before (except for the "expression" thing, I just made that up)

subhaze:
Thanks, I'll review the above (hopefully later today) and adjust the PR.

subhaze:
Updated scopes from notes above, rebased this branch on top of the current Packages master branch.

wbond:
Thank you both for the time spent working out improvements for this!

PR #168, opened 2016-02-12
[Ruby][Rails] Syntax improvements

tmichel:
## Rails

* Add support for some Rails 4 special methods (e.g. `before_action`)
* Fix highlighting when special methods are used as symbols

## Ruby

* Add pattern for constant definitions, apply `entity.name.type` so Go To Definition finds it
* Fix logical operators highlighting: `and`, `or`, `not` could be methods and in that case should not be highlighted.
* Add `prepend` as a special method. It is very similar to `include`.
* Add support for `%I` and `%i`: symbol arrays
* Remove `%S` and `%R` as they are not valid percent expressions
* Block variables can have modifiers (`*` and `&`), recognize these with the same scope as in methods.
* Add support for string formatting literals. E.g. `%s` or `%f`. Code copied for the Go syntax definition.

### DIFF
--- a/Rails/Ruby on Rails.sublime-syntax
+++ b/Rails/Ruby on Rails.sublime-syntax
@@ -87,10 +87,12 @@ contexts:
           pop: true
         - include: 'scope:source.ruby'
         - include: main
-    - match: \b(before_filter|skip_before_filter|skip_after_filter|after_filter|around_filter|filter|filter_parameter_logging|layout|require_dependency|render|render_action|render_text|render_file|render_template|render_nothing|render_component|render_without_layout|rescue_from|url_for|redirect_to|redirect_to_path|redirect_to_url|respond_to|helper|helper_method|model|service|observer|serialize|scaffold|verify|hide_action|append_view_path|prepend_view_path|view_paths)\b
+    - match: '\b(before_(filter|action)|skip_before_(filter|action)|skip_after_(filter|action)|after_(filter|action)|around_(filter|action)|filter|filter_parameter_logging|layout|require_dependency|render|render_action|render_text|render_file|render_template|render_nothing|render_component|render_without_layout|rescue_from|url_for|redirect_to|redirect_to_path|redirect_to_url|respond_to|helper|helper_method|model|observer|serialize|scaffold|verify|hide_action|append_view_path|prepend_view_path|view_paths)(?![?!:])\b'
+      comment: Uses negative lookahead to filter out symbols
       scope: support.function.actionpack.rails
-    - match: \b(named_scope|after_create|after_destroy|after_save|after_update|after_validation|after_validation_on_create|after_validation_on_update|before_create|before_destroy|before_save|before_update|before_validation|before_validation_on_create|before_validation_on_update|composed_of|belongs_to|has_one|has_many|has_and_belongs_to_many|validate|validate_on_create|validates_numericality_of|validate_on_update|validates_acceptance_of|validates_associated|validates_confirmation_of|validates_each|validates_format_of|validates_inclusion_of|validates_exclusion_of|validates_length_of|validates_presence_of|validates_size_of|validates_uniqueness_of|attr_protected|attr_accessible|attr_readonly)\b
+    - match: '\b(named_scope|default_scope|scope|after_create|after_destroy|after_save|after_update|after_validation|after_validation_on_create|after_validation_on_update|before_create|before_destroy|before_save|before_update|before_validation|before_validation_on_create|before_validation_on_update|composed_of|belongs_to|has_one|has_many|has_and_belongs_to_many|validate|validate_on_create|validates_numericality_of|validate_on_update|validates_acceptance_of|validates_associated|validates_confirmation_of|validates_each|validates_format_of|validates_inclusion_of|validates_exclusion_of|validates_length_of|validates_presence_of|validates_size_of|validates_uniqueness_of|validates|attr_protected|attr_accessible|attr_readonly)(?![?!:])\b'
+      comment: Uses negative lookahead to filter out symbols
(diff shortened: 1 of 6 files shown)


wbond:
We'll definitely need to add some tests for these changes. See https://github.com/sublimehq/Packages#pull-requests and https://www.sublimetext.com/docs/3/syntax.html#Testing for more info.

tmichel:
I'll add the tests for the modifications and the obvious ones that comes along. I think completely covering the Ruby & Rails syntax is out of scope in the pull request.

wbond:
Yes, adding some basic tests plus targeted tests for these changes is all that is needed. No need to test all existing edge cases.

tmichel:
@wbond I added tests and rebased it on `master` and incorporated the existing ruby tests.

Rails/Ruby on Rails.sublime-syntax
```diff
@@ -87,10 +87,12 @@ contexts:
           pop: true
         - include: 'scope:source.ruby'
         - include: main
-    - match: \b(before_filter|skip_before_filter|skip_after_filter|after_filter|around_filter|filter|filter_parameter_logging|layout|require_dependency|render|render_action|render_text|render_file|render_template|render_nothing|render_component|render_without_layout|rescue_from|url_for|redirect_to|redirect_to_path|redirect_to_url|respond_to|helper|helper_method|model|service|observer|serialize|scaffold|verify|hide_action|append_view_path|prepend_view_path|view_paths)\b
+    - match: '\b(before_(filter|action)|skip_before_(filter|action)|skip_after_(filter|action)|after_(filter|action)|around_(filter|action)|filter|filter_parameter_logging|layout|require_dependency|render|render_action|render_text|render_file|render_template|render_nothing|render_component|render_without_layout|rescue_from|url_for|redirect_to|redirect_to_path|redirect_to_url|respond_to|helper|helper_method|model|observer|serialize|scaffold|verify|hide_action|append_view_path|prepend_view_path|view_paths)(?![?!:])\b'
```
wbond: `(?![?!:])\b` will probably not function quite the way we want.

`\b` is a word boundary, so it won't trigger unless `?`, `!` or `:` are followed by a word character, or there is no symbol at the end of the method name.

Additionally, I presume the `(?![?!:])` is to handle Ruby 1.9 symbol keys, correct? If so, it should probably be: `\b(?![?!]?:)`. That will prevent matching the following: `after_filter!:`, `after_filter?:` and `after_filter:`. It will still match `after_filter :foo_bar`.

Ruby/Ruby.sublime-syntax
```diff
@@ -440,7 +449,7 @@ contexts:
             0: punctuation.definition.string.end.ruby
           pop: true
         - include: regex_sub
-    - match: '%[QWSR]?\('
+    - match: '%[QWI]?\('
```
wbond: What triggered the change here?

Ruby/Ruby.sublime-syntax
```diff
@@ -87,6 +87,11 @@ contexts:
         6: punctuation.separator.inheritance.ruby
         7: entity.other.inherited-class.module.third.ruby
         8: punctuation.separator.inheritance.ruby
+    - match: '([A-Z]\w*)(?=(,\s*[A-Z]\w*)*\s*=(?!\>))'
```
wbond: Let's add a test for this

Ruby/syntax_test_ruby.rb
```diff
@@ -1,8 +1,50 @@
...
+    #                  ^^^^ variable.other.readwrite.instance.ruby
+    #                                ^ keyword.operator.arithmetic.ruby
+    #                                  ^ meta.variable.block.ruby
+    end
+  end
+
+  def my_method
+    str = "number %d" %  11
+                # ^^ string.quoted.double.ruby constant.other.placeholder.ruby
+    str = 'number %d' % 11
```
wbond: Can we add tests for a separator, field width and field number to ensure those are working properly and aren't broken via future changes?

wbond:
Do your modifications handle the additions in #164 also?

Ruby/Ruby.sublime-syntax
tmichel: `%S` and `%R` are not valid expressions.

```
irb(main):012:0> %S(test)
SyntaxError: (irb):12: unknown type of %string
%S(test)
   ^

irb(main):013:0> %R(test)
SyntaxError: (irb):13: unknown type of %string
%R(test)
   ^
```

Rails/Ruby on Rails.sublime-syntax
tmichel: You are correct. This is added to handle Ruby 1.9 symbol keys. 

I don't quite understand why but either version works. I added tests for recognizing `after_filter!:` and `after_filter?:` as a symbols.

Ruby/Ruby.sublime-syntax
tmichel: Tests are in d28b28c.

Ruby/syntax_test_ruby.rb
tmichel: I revised my previous version. Originally I copied it from the Go syntax definition but Ruby's format string is a bit different. So I adapted the regexp to better cover Ruby's format string.

The only thing not matched is named arguments.

```ruby
sprintf("%<foo>d : %<bar>f", { :foo => 1, :bar => 2 }) #=> 1 : 2.000000
sprintf("%{foo}f", { :foo => 1 }) # => "1f"
```

tmichel:
> Do your modifications handle the additions in #164 also?

Yes. `xxx_filter` and `xxx_action` are both matched properly.

wbond:
Excellent. Thank you for all of your work on this!

PR #173, opened 2016-02-13
[C++] C++14 Digit Separators, Binary Literals, and Units

jbw3:
Highlighting digit separators and binary literals per the C++14 standard. Also, highlighting user-defined literals per the C++11 standard. This includes units defined in the STL (h, min, s, etc.) and custom literals defined by users.


wbond:
Thanks!

PR #174, opened 2016-02-13
[Pascal] Added dpr extension

LFriede:
dpr files are created by Delphi to store the initial unit, same syntax as pas files, so it should be included.


wbond:
Thanks!

LFriede:
You're welcome.

PR #175, opened 2016-02-13
[Batch File] Fixing a Comment Highlight Bug

jbw3:
Fixing a bug where @rem was not highlighted as a comment

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -20,7 +20,7 @@ contexts:
       scope: keyword.control.repeat.dosbatch
     - match: \b(?:EQU|NEQ|LSS|LEQ|GTR|GEQ)\b
       scope: keyword.operator.dosbatch
-    - match: (?:^|\s)((?i)rem)(?:$|\s.*$)
+    - match: (?<=^|\s|@)((?i)rem)(?:$|\s.*$)
       scope: comment.line.rem.dosbatch
       captures:
         1: keyword.command.rem.dosbatch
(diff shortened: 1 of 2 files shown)


kleinfreund:
@jbw3 Please be careful using `@` outside of code formatting on GitHub because it pings users – in this case it pinged the user rem.

jfcherng:
However, `@` is not exclusive to `rem`. The syntax file looks inconsistent using `\b` and `(^|\s)` since previous change. How about let `@` be a symbol, i.e., something like

```yaml
- match: ^\s*@
  scope: whatever
```

wbond:
@jfcherng I believe you are correct that this should be a general symbol, not just for `rem`.

jbw3:
Sounds good. I'll make this change as soon as I get a chance

wbond:
@jbw3 Why the change from `\b` to the lookbehind and lookahead?

In general, moving forward, we are likely to be trying to remove lookbehinds from all of the default syntaxes since it prevents usage of the new regex engine, thus resulting is slower parsing performance.

jfcherng:
@wbond It looks like it's because of my suggestion. lol...

wbond:
@jfcherng Ah, yes! :smile: I guess I missed the end of your comment.

I'll take a look at this in a day or two and see how involved a rewrite of it would be, such as was done with JavaScript.

jbw3:
@wbond I did make the changes per @jfcherng's request :) Should I keep them or go back to `\b`?

wbond:
I believe this issue has been fixed by recent improvements to Batch File

Issue #154, opened 2016-02-10
[JavaScript] Syntax highlight - how to get different colors for object literal key and a function value?


ahwayakchih:
As mentioned in #141, keys of object literals will get different color if value is a function.
I managed to modify Monokai Extended theme to prevent that, but i do not know how to keep key white, but function name colored.

Here's theme rule i added:

```xml
		<dict>
			<key>name</key>
			<string>JS: Object Key for Function Name</string>
			<key>scope</key>
			<string>meta.function.json.js entity.name.function.js, meta.function.json.arrow.js entity.name.function.js</string>
			<key>settings</key>
			<dict>
				<key>fontStyle</key>
				<string></string>
				<key>foreground</key>
				<string>#ffffff</string>
			</dict>
		</dict>
```

Here's test code:

```js
var obj = {
	one: () => 1,
	two: function two () {
		return 2;
	}
}
```

Here's result:

![screenshot from 2016-02-10 19 10 21](https://cloud.githubusercontent.com/assets/39827/12957218/de1cc0e8-d02b-11e5-8114-31a21453402a.png)

Is there a way to make function name colored, but keep object literal key color unchanged? It would be great if i could get something like this mockup:

![screenshot from 2016-02-10 19 10 21-mockup](https://cloud.githubusercontent.com/assets/39827/12957558/7da26108-d02d-11e5-9f92-fbb584df340c.png)

jonschlinkert:
can you paste the code in too? I'll play around with it and see if I can get what you want. 

fwiw, honestly the first one (`before`) seems like the most desirable, given that the value is a function for both properties. but that's just my own preference

FichteFoll:
I took a quick look and it is currently impossible to specifically target functions defined as object keys without also matching "normal" function definitions.

ahwayakchih:
@jonschlinkert i've updated post, to include code.

It does not look bad in case of example above, but it starts to look messy in something like this:

```js
var obj = {
    one: 1,
    two: () => 2,
    three: "3",
    four: function four () {
        return 4;
    }
}
```

But that's just a visual preference, and one could argue that such object literals should not be defined anyway. Thing is that the key is not highlighted if a function is passed through a variable, like this:

```js
var one = () => 1;
var obj = {
    one: one
}
```

`obj.one` is still a function (or a reference to a function), but highlighter does not know about that. Which kinda breaks rule.
[...]

PR #181, opened 2016-02-16
[JavaScript] Added function label entities

ahwayakchih:
As discussed in #154, added `entity.name.function.label.js` to clearly mark which part of function expression or declaration is a proper function name. Now functions have both `entity.name.function.js` (or `entity.name.function.method.js`) and `entity.name.function.label.js`, so this change is backwards compatible.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -355,7 +355,7 @@ contexts:
                 1: storage.type.js
                 2: storage.type.accessor.js
                 3: keyword.generator.asterisk.js
-                4: entity.name.function.method.js
+                4: entity.name.function.method.js entity.name.function.label.js
               push:
                 - meta_scope: meta.method.js
                 - match: (?<=\))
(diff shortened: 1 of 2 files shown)


wbond:
In this PR, only some of the function definition formats are tagged with the new scope.

However, I'd rather see the object literal key scopes get an addition so you can re-color them they way you want, as opposed to trying to color the actual function names for all of the other varieties of function definitions. This would result in fewer changes and would be more understandable for other color scheme developers and is more consistent with the existing `meta.object-literal.key` scopes.

You can see the changes in f6c42c7d4ade0687b3a7457205540bfcbf1f32f0. I tested them with a customized color scheme to ensure that you can use a selector to override the `entity.name.function.js` color.

Here is an example of the color scheme customization that was required for Monokai:

```xml
<dict>
	<key>name</key>
	<string>String</string>
	<key>scope</key>
	<string>string, meta.object-literal.key.js string entity.name.function.js</string>
	<key>settings</key>
	<dict>
		<key>foreground</key>
		<string>#E6DB74</string>
	</dict>
</dict>
<dict>
	<key>name</key>
	<string>Function label</string>
	<key>scope</key>
	<string>meta.object-literal.key.js entity.name.function.js</string>
	<key>settings</key>
	<dict>
		<key>foreground</key>
		<string>#F8F8F2</string>
	</dict>
</dict>
```

ahwayakchih:
Thanks.

jonschlinkert:
:+1: great, thanks friends!

PR #182, opened 2016-02-17
[JavaScript] Fix scoping for arrow functions

ccampbell:
I noticed in my last pull request I missed a couple of cases related to arrow functions returning different scopes than regular functions. 

I updated the scope and corresponding regex and added tests.

Hope you don't mind sending PRs as little things like this crop up :smile:

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -213,7 +213,7 @@ contexts:
         \s*(async)?
         \s*(?=\([^()]*\)\s*(=>))
       captures:
-        1: entity.name.class.js
+        1: support.class.js
         2: keyword.operator.accessor.js
         3: variable.other.readwrite.js entity.name.function.js
         4: storage.type.js
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -207,7 +207,7 @@ contexts:
         8: storage.type.function.arrow.js
     - match: |-
         (?x)
-        (\b_?[A-Z][$\w]*)?
+        (\b_?[a-zA-Z_?.$]*)?
         (\.)([_$a-zA-Z][$\w]*)
         \s*=
         \s*(async)?
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -207,7 +207,7 @@ contexts:
         8: storage.type.function.arrow.js
     - match: |-
         (?x)
-        (\b_?[a-zA-Z_?.$]*)?
+        (\b_?[a-zA-Z_.$]*)?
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -207,13 +207,13 @@ contexts:
         8: storage.type.function.arrow.js
     - match: |-
         (?x)
-        (\b_?[A-Z][$\w]*)?
+        (\b_?[a-zA-Z_?.$]*)?
```
wbond: Are `.` and `?` really valid characters in a class name?

ccampbell: Oh I just copied this regex from the one for non arrow functions. I think it is actually designed this way so that you can do things like.

```javascript
SomeThing.SomeOtherThing.AnotherThing.whatever = () => {}
```

Not the prettiest, but I am not sure if sublime syntaxes support repeating patterns like that.

Technically it should probably use the same pattern as the one on line 211 to actually match the correct characters.

wbond: Ok, What about the `?`?

ccampbell: I was just going to comment I have no idea about the question mark. I was just trying to search to find if it is some upcoming JavaScript feature, but I couldn't find anything. I am guessing that was just copied over from the javascript next package when it was originally copied into Sublime?

wbond: Alright, for now let's use `(\b_?[a-zA-Z_.$]*)?`.

I need to spend some time today on JavaScript.sublime-syntax since we are seeing indexing performance issues with the new additions in 3104. So there may end up being some significant changes upcoming.

ccampbell: Ah weird, I was the one who added that weird question mark thing, but I am pretty sure I copied it from an older syntax. Anyway I will update both of them.

As for the performance problems, do you think they were from my changes? I looked over them again and didn't see anything crazy unless it is something to do with negative/positive lookahead/lookbehinds making things slower.  Like changing

```diff
-    - match: \.(shift|sort|splice|unshift|pop|push|reverse)\b
 +    - match: (?<=\.)(shift|sort|splice|unshift|pop|push|reverse)\b(?=\()
```

ccampbell:
Okay I removed the question mark from the regular expressions so I think this should be good to go now. Interested in hearing what you discover with regards to the performance issues.

wbond:
I'm not seeing performance regressions with a file like jQuery. I am going to have to seek out some examples from users.

wbond:
These tests were added and the functionality implemented in the big rewrite from 409ccb469bca8bcc2a5c70c562d1db3428d3015f

ccampbell:
Awesome! I will check it out later :smile:

PR #183, opened 2016-02-18
[PHP] Added missing keywords (trait, callable, insteadof, string, int)

ptcong:
- Added missing keywords (trait, callable, insteadof)

trait keyword
```
trait Taggable
{
}
```

improve use meta.

Now `as` keywords not being `operator` keyword

Added `insteadof` keyword

```
class Xyz {
    use A, B, C {
        C:method2 insteadof B;
        B::method1 as anotherOne1;
        B::method2 as anotherOne2;
    }
}

```

Added `callable`, `int`, `string` type hint
```
public function abc(callable $var, int $var2, string $var3)
{
   
}
```

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -106,7 +106,7 @@ contexts:
     - include: comments
     - match: |-
         (?x)
-        \s*(array|callable|int|string) # Typehint
+        \s*(array|callable|int|string|bool|float) # Typehint
         \s*(&)?                     # Reference
         \s*((\$+)[a-zA-Z_\x{7f}-\x{ff}][a-zA-Z0-9_\x{7f}-\x{ff}]*) # The variable name
         \s*(=)  # A default value
@@ -131,7 +131,7 @@ contexts:
         - include: numbers
     - match: |-
         (?xi)
-        \s*(array|callable|int|string) # Typehint
+        \s*(array|callable|int|string|bool|float) # Typehint
         \s*(&)?                     # Reference
         \s*((\$+)[a-z_\x{7f}-\x{ff}][a-z0-9_\x{7f}-\x{ff}]*) # The variable name
         (?:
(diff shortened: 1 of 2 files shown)


wbond:
We’ll need tests for these additions. Check out http://www.sublimetext.com/docs/3/syntax.html#Testing for more info.

ptcong:
@wbond ok, i just made the tests

ptcong:
@wbond recheck the PR, i just add the tests

PHP/PHP Source.sublime-syntax
```diff
@@ -373,29 +373,54 @@ contexts:
         1: keyword.other.use.php
       push:
         - meta_scope: meta.use.php
-        - match: (?=;|(?:^\s*$))
+        - match: (?=;|(?:\s*$)|\s+)
           pop: true
```
wbond: This doesn't seem right. Does PHP really end a use statement as soon as a one or more spaces is present?

ptcong: Just for pop from
```
use Trait1, Trait2 {
    Trait1:methodA as methodX;
} [this is space]
protected $prop1;
```

ptcong:
@wbond i just remke this commit, it seems works fine now.

michaelblyons:
Can you prefix your pull request title with [PHP]?

ptcong:
@michaelblyons ok i just added [PHP] prefix for this PR

PHP/PHP Source.sublime-syntax
```diff
@@ -106,7 +106,7 @@ contexts:
     - include: comments
     - match: |-
         (?x)
-        \s*(array) # Typehint
+        \s*(array|callable|int|string) # Typehint
```
duncan3dc: What about `bool` and `float`? (http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration)

wbond:
I rebased after merging these in, so the hashes changed. Here are the new commits:

4397c636
004c616a
55be38a1

PR #184, opened 2016-02-18
[JavaScript] Removes jsx extension from list

subhaze:
jsx isn't part of core JS and not supported in this syntax file

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -5,7 +5,6 @@ name: JavaScript
 file_extensions:
   - js
   - htc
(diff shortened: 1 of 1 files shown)


wbond:
This was taken care of with the big rewrite in 409ccb469bca8bcc2a5c70c562d1db3428d3015f

subhaze:
Nice, pulled latest on the JS and PHP is blowing up on me now. I don't see an issue filed and about to head out. I'll check again later tonight and try to get something written down that's better than "broke PHP" :)

wbond:
Thanks for the heads up, I'll see if I can reproduce any issues

PR #194, opened 2016-02-20
[Groovy] Support comment toggling

ajoberstar:
Sublime Text has had a long-standing issue where comments in Groovy
code cannot be toggled with Ctrl+/. The workaround has been to copy
the comments preferences from the Java package and change the
source.java to source.groovy.

One source for the fix:
http://stackoverflow.com/questions/24567263/how-to-enable-groovy-comments-in-sublimetext-3-on-a-mac


wbond:
Excellent, thanks!

PR #225, opened 2016-03-02
[HTML] Updates tests to use new class-name scope

subhaze:
Updating tests to reflect changes in https://github.com/sublimehq/Packages/commit/9d2ff48c8df42b42874360661e2c719e6278f3ec

Edited on phone so double check me :p


wbond:
Thanks!

PR #228, opened 2016-03-04
Whitespace Fix for CSS Comments initiated with `Ctrl+/` in ST3 Build 3103

lunacodes:
CSS Comments previously displayed as `/*Comment Here*/` and now display as `/* Comment Here */`

Modified Comments.tmPreferences in CSS Package.
**17**: Changed `<string>/*</string>` to `<string>/* </string>`
**23**: Change `<string>*/</string>` to `<string> */</string>`

### DIFF
--- a/CSS/Comments.tmPreferences
+++ b/CSS/Comments.tmPreferences
@@ -14,13 +14,13 @@
 				<key>name</key>
 				<string>TM_COMMENT_START</string>
 				<key>value</key>
-				<string>/*</string>
+				<string>/* </string>
 			</dict>
 			<dict>
 				<key>name</key>
 				<string>TM_COMMENT_END</string>
 				<key>value</key>
-				<string>*/</string>
+				<string> */</string>
 			</dict>
 			<dict>
 				<key>name</key>


wbond:
I looked through all of the other `Comments.tmPreferences` files in the various languages, and as of right now none of them add spaces like this.

I am thinking it probably makes sense to keep it as-is. If you would like to add this for yourself, place the modified version of `CSS/Comments.tmPreferences` in `Packages/CSS/Comments.tmPreferences` and restart Sublime Text.

lunacodes:
Hi Will,

I looked through some other comment files for reference, and what I noticed is that single-line comments do have the space, while multi-line comments do not.  HTML appears to be an exception: https://github.com/sublimehq/Packages/blob/master/HTML/Comments.tmPreferences

Since CSS uses the same syntax for single & multi-line comments (as opposed to only having single-line comments, like Python), it seems like it would be useful to implement a TM_COMMENT_START & TM_COMMENT_START2 (example here: https://github.com/sublimehq/Packages/blob/master/JavaScript/Comments.tmPreferences) and have TM_COMMENT_START contain the space, and TM_COMMENT_START_2 not contain the space.

If you feel that's a constructive addition, I'd be happy to test it locally & submit a pull request with the code (and same for any other languages with this difference).

wbond:
I noticed that about the single line comments also. My guess is that it stems from whoever implemented the files first.

What `TM_COMMENT_START_2` (and `_3`) do is allow the comment code to allow uncommenting different styles of comments, for instance if a language allows `#` and `//` like PHP does. However, when you re-comment, it will always use the first version.

Because of that, I don't think that implementing `/*` and `/* ` will functionally really do anything. The one without the space will always be matched and used.

I'm hesitant to make changes for person preference (such as spacing) because of the sheer number of users. It is bound to annoy someone, or cause some side effect, so when presented with the option to keep something the same or change it, I'll tend to err on the side of keeping it the same. However, if there is a bug to fix, or an improvement to be made, I think it is easier to justify.

I realize this may not be the ideal outcome in your case, but with the override functionality, at least you can accomplish it for yourself.

lunacodes:
Thanks for all the clarification. 

In that case, if it were to be implemented: TM_COMMENT_START could include the space & TM_COMMENT_START_2 could exclude it & the uncommenting should still match. 

Just to clarify, I didn't submit due to personal preference - I submitted because it matches the standard comment style more accurately.  

For what it's worth: I've had this implemented locally for a while, and have found that ST still uncomments the /*comment here*/ syntax just fine for both single & multiline comments

Anyway, I understand your hesitancy - so up to you either way & thanks for the feedback & considering it!

PR #233, opened 2016-03-09
[HTML] Add support for custom tags

poke:
According to the [Custom Elements specification](https://www.w3.org/TR/custom-elements/#concepts), custom element tag names are required to contain a HYPEN-MINUS character.

This pull request adds the ability to match such custom elements properly.

Technically, the specification also requires that the tag names are lower-case, but I guess that would just cause further problems, so it’s easier to just allow dashes in upper- or mixed-case tag names as well.

I hope the weird indentation in the test is not a problem.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -124,6 +124,26 @@ contexts:
           scope: punctuation.definition.tag.end.html
           pop: true
         - include: tag-stuff
+    - match: (</?)([a-z_][a-z0-9:_]*-[a-z0-9:_-]+)
+      captures:
+        1: punctuation.definition.tag.begin.html
+        2: entity.name.tag.custom.html
+      push:
+        - meta_scope: meta.tag.custom.html
+        - match: '>'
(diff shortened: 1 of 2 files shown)


wbond:
Perhaps for custom tags use a scope such as `entity.name.tag.other.custom.html`.

We should probably also add a test to ensure that non-custom tags (without `-`) get `entity.name.tag.other.html`.

wbond:
Or a possibly better scope: `entity.name.tag.custom.html`.

poke:
Yeah, that would work too, good idea. Now that I understood how the syntax files work, I can work on that. I'll update the pull request tomorrow. (Alternatively, you can also merge this in and I'll expand on it then in a new PR given that this is new functionality)

poke:
Okay, here we go. Expanded on this idea a bit further :) The custom elements scope is now according to spec (except for all those extra characters in the other unicode ranges which are also not specified in the syntax file for the other tag names).

wbond:
Awesome, thanks!

PR #234, opened 2016-03-09
[HTML] added table, form and link syntax highlighting

hrzi:
Added custom scopes for table tags (table,tr,td...), form tags (form,input,button...) and a tag to be able to custom color them in themes.


wbond:
This seems to change a whole bunch of tags from `entity.name.tag.inline` to `entity.name.tag.block`.

Also, adding tests would be appropriate.

hrzi:
I changed the scopes and also added syntax test. Is this more acceptable now?

wbond:
Excellent, thanks!

Issue #17, opened 2015-06-19
Matlab OOP not supported?


mmagnuski:
On my sublime text 2 Matlab keywords like `classdef` or `properties` do not get highlighted.
For example in Matlab you can define object properties this way:
```matlab
properties (SetAccess = private, GetAccess = public)
    someproperty
end
```
This is not highlighted in my ST (you can see that GitHub matlab highlighter is also unaware of that :) )
If this is not supported in ST - I would love to help, but unfortunatelly I know nothing about how syntax highlighting is defined in ST...

FichteFoll:
Check out http://docs.sublimetext.info/en/latest/extensibility/syntaxdefs.html for what matlab uses currently and http://www.sublimetext.com/docs/3/syntax.html for the new format.

egeerardyn:
I would also love to see support for MATLAB classes. I haven't looked at the syntax definition format in detail, but it looks daunting.

Anyhow, this is some relevant documentation provided by the MathWorks: 
- http://nl.mathworks.com/help/matlab/ref/classdef.html
- http://nl.mathworks.com/help/matlab/matlab_oop/class-components.html

wbond:
This should be supported now in the master branch as of 4f2cb943c51a06573eccfa31a028c8f68d3d3b6d, and should be part of the next dev build.

PR #242, opened 2016-03-16
[Matlab] Add support for OOP syntax

TheClams:
Fix issue #17 by adding support for OOP syntax (classdef, properties,
...) and corresponding test file.

Fix scope for all support function
Fix scope for ans, i, j and pi which are constant and not keyword
Fix scope for fclose (function, not type)
Add scope for = (keyword.operator)


wbond:
This was rebased and committed as 4f2cb943c51a06573eccfa31a028c8f68d3d3b6d

wbond:
Thanks!

PR #248, opened 2016-03-18
[HTML/CSS] Fix custom tag names

poke:
Turns out I didn’t test the custom tag names well enough in #233 and #236. Custom tags that start with a otherwise known tag are not correctly recognized (e.g. `<form-group>`). I’ve added more tests for these cases, and fixed the syntax definitions to give priority to custom tag names first (otherwise I would have to change every other tag name definition).


wbond:
Thanks!

PR #252, opened 2016-03-18
Fix Typos and Misspellings

Briles:
Fixed all the typos and misspellings that I could find.

Some of them affected scopes while others were located within comments.


wbond:
Awesome, thanks for these!

PR #256, opened 2016-03-22
[CSS] Enhancements

Briles:
#### What's changed:

- Comment keys are now YAML comments

- All instances of `captures: 0:` were replaced with `scope:`

- Functions more closely follow the [spec](https://www.w3.org/TR/css3-color/#rgba-color) by no longer allowing whitespace after function name.

- Autocompleted comments now have standard whitespace - Fixes SublimeTextIssues/DefaultPackages#179
- Added the generic font families `cursive` & `fantasy` - [W3C](https://www.w3.org/TR/CSS2/fonts.html#generic-font-families)

- Pseudo-classes are no longer improperly scoped as custom tags

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;_the current custom tag regex will match most use-cases but should be improved to more closely follow the [spec] (http://w3c.github.io/webcomponents/spec/custom/#custom-elements-core-concepts)_

- Added the [namespace](https://www.w3.org/TR/css3-namespace) & [page](https://www.w3.org/TR/CSS2/page.html) at-rules

- Comment trailing punctuation `*/` is now scoped

- Added the `main` & `xmp` html tags

- Added the `speech` media type - [W3C](https://www.w3.org/TR/CSS21/media.html)

- Added the `rebeccapurple` color

- Improved test coverage *(some tests are currently failing)*

- Added completions for the following properties / values:
  - align-self: auto
  - alignment-baseline - [W3C](https://www.w3.org/TR/SVG11/text.html#AlignmentBaselineProperty)
  - baseline-shift - [W3C](https://www.w3.org/TR/SVG11/text.html#BaselineShiftProperty)
  - clip-path (71% [Browser Support](http://caniuse.com/#feat=css-clip-path))
  - clip-rule - [W3C](https://www.w3.org/TR/2014/WD-css-masking-1-20140213/#the-clip-rule)
  - color-interpolation - [W3C](https://www.w3.org/TR/SVG11/painting.html#ColorInterpolationProperty)
  - color-rendering - [W3C](https://www.w3.org/TR/SVG11/painting.html#ColorRenderingProperty)
  - content: counter()
  - cursor
  - display: table-caption (CSS 2.1)
  - dominant-baseline - [W3C](https://www.w3.org/TR/SVG11/text.html#DominantBaselineProperty)
  - image-rendering (85% [Browser Support](http://caniuse.com/#feat=css-crisp-edges))
  - list-style-type
  - mix-blend-mode (66% [Browser Support](http://caniuse.com/#feat=css-mixblendmode))
  - overflow-x
  - overflow-y
  - perspective - [W3C](https://www.w3.org/TR/css-transforms-1/#propdef-perspective)
  - perspective-origin - [W3C](https://www.w3.org/TR/css-transforms-1/#propdef-perspective-origin)
  - size
  - stroke-linejoin - [W3C](https://www.w3.org/TR/SVG11/painting.html#StrokeLinejoinProperty)
  - text-align-last (65% [Browser Support](http://caniuse.com/#feat=css-text-align-last))
  - text-orientation - [W3C](https://drafts.csswg.org/css-writing-modes-3/#text-orientation)
  - writing-mode (91% [Browser Support](http://caniuse.com/#feat=css-writing-mode))

#### Performance Tests

```
// 35K Lines // 740534 chars // 741,376 bytes
```
---
```
Syntax "CSS.sublime-syntax" took an average of 422.8ms over 10 runs
Syntax "CSS.sublime-syntax" took an average of 423.3ms over 10 runs
Syntax "CSS.sublime-syntax" took an average of 413.7ms over 10 runs
Syntax "CSS.sublime-syntax" took an average of 424.9ms over 10 runs
Syntax "CSS.sublime-syntax" took an average of 426.0ms over 10 runs
```
__Average ~ 422.14ms__

---

```
Syntax "CSS (Modified).sublime-syntax" took an average of 426.1ms over 10 runs
Syntax "CSS (Modified).sublime-syntax" took an average of 425.9ms over 10 runs
Syntax "CSS (Modified).sublime-syntax" took an average of 422.6ms over 10 runs
Syntax "CSS (Modified).sublime-syntax" took an average of 410.6ms over 10 runs
Syntax "CSS (Modified).sublime-syntax" took an average of 415.1ms over 10 runs
```
__Average ~ 420.06ms__

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -6,6 +6,26 @@ file_extensions:
   - css
   - css.erb
 scope: source.css
+variables:
+  custom_element_chars: |-
+    (?x:
+        [-_a-z0-9\xB7]
+      | \\\.
+      | [\xC0-\xD6]
+      | [\xD8-\xF6]
+      | [\x{000000F8}-\x{000002FF}]
+      | [\x{00000300}-\x{0000037D}]
+      | [\x{0000037F}-\x{00001FFF}]
+      | [\x{0000200C}-\x{0000200D}]
+      | [\x{0000203F}-\x{00002040}]
+      | [\x{00002070}-\x{0000218F}]
+      | [\x{00002C00}-\x{00002FEF}]
+      | [\x{00003001}-\x{0000D7FF}]
+      | [\x{0000F900}-\x{0000FDCF}]
+      | [\x{0000FDF0}-\x{0000FFFD}]
+      | [\x{00010000}-\x{000EFFFF}]
+    )
+
 contexts:
   main:
     - include: comment-block
@@ -665,7 +685,8 @@ contexts:
         - meta_scope: meta.selector.css
         - match: "(?=[/@{)])"
           pop: true
-        - match: '\b([a-z][a-z0-9_]*-[a-z0-9_-]+)\b'
+          # Custom Elements: http://w3c.github.io/webcomponents/spec/custom/#custom-elements-core-concepts
+        - match: '\b([a-z](?:{{custom_element_chars}})*-(?:{{custom_element_chars}})*)\b'
           scope: entity.name.tag.custom.css
         - match: '\b(a|abbr|acronym|address|applet|area|article|aside|audio|b|base|basefont|bdi|bdo|big|blockquote|body|br|button|canvas|caption|cite|code|col|colgroup|content|data|datalist|dd|del|details|dfn|dir|dialog|div|dl|dt|element|em|embed|eventsource|fieldset|figure|figcaption|footer|form|frame|frameset|h[1-6]|head|header|hgroup|hr|html|i|iframe|img|input|ins|isindex|kbd|keygen|label|legend|li|link|main|map|mark|menu|meta|meter|nav|noframes|noscript|object|ol|optgroup|option|output|p|param|picture|pre|progress|q|rp|rt|rtc|s|samp|script|section|select|shadow|small|source|span|strike|strong|style|sub|summary|sup|svg|table|tbody|td|template|textarea|tfoot|th|thead|time|title|tr|track|tt|u|ul|var|video|wbr|xmp)\b'
           scope: entity.name.tag.css
--- a/CSS/syntax_test_css.css
+++ b/CSS/syntax_test_css.css
@@ -240,7 +240,7 @@
 /*              ^ constant.other.color.hsl-value.css */
 }
 
-.test-custom-tags > div > span + custom-tag > div-custom-tag ~ form-custom-tag.classname:last-child:hover {}
+.test-custom-tags > div > span + cust·m-tÀg > div-cøstom-tag ~ form-Çust😀m-tag.classname:last-child:hover {}
 /*                ^ -entity.name.tag.custom.css */
 /*                  ^^^ entity.name.tag.css */
 /*                        ^^^^ entity.name.tag.css */


CSS/CSS.sublime-syntax
```diff
@@ -140,11 +144,13 @@ contexts:
         2: punctuation.definition.keyword.css
       push:
         - meta_scope: meta.at-rule.import.css
-        - match: '\s*((?=;|\}))'
+        - match: '\s*(;|\})'
```
wbond: I don't think a `}` should be scoped as `punctuation.terminator.rule.css`. I think this and the previous lookahead removed on line 45 (old) should probably remain as lookaheads. That way we don't need to duplicate the scope names for `;` and `}` in multiple places.

CSS/CSS.sublime-syntax
```diff
@@ -271,6 +278,36 @@ contexts:
         - match: '\s*(?=[{;])'
           pop: true
         - include: media-query
+# Namespace At-Rule: https://www.w3.org/TR/css3-namespace/
+  namespace:
+    - match: \s*((@)namespace)\s*\b(\S*)
+      captures:
+        1: keyword.control.at-rule.namespace.css
+        2: punctuation.definition.keyword.css
+        3: entity.other.namespace-prefix.css
```
wbond: I'd say let's skip adding the `punctuation.definition.keyword.css` for the `@` symbol here and on line 301. IMO, the `@` is really part of the keyword and not so much separate punctuation.

CSS/CSS.sublime-syntax
```diff
@@ -271,6 +278,36 @@ contexts:
...
+      push:
+        - meta_scope: meta.at-rule.namespace.css
+        - match: '\s*(;|$)'
+          captures:
+            1: punctuation.terminator.rule.css
```
wbond: I'd use a lookahead here: `(?=;|$)` so that we don't have to duplicate the scope for `;`.

CSS/Comments.tmPreferences
```diff
@@ -14,13 +14,13 @@
 				<key>name</key>
 				<string>TM_COMMENT_START</string>
 				<key>value</key>
-				<string>/*</string>
+				<string>/* </string>
```
wbond: This seems inconsistent with the other languages in the default packages, and is more of a personal preference. Since there isn't really a functional difference, I'd prefer to defer to it staying the same. Mostly because users are particular about "unnecessary" changes.

We discussed this some in #228.

wbond:
Could you give me a quick overview of what you understand the deficiencies to be in terms of custom tag handling? I'd be fine with spending some time on it, but if you already know what is deficient, it may save some time.

CSS/CSS.sublime-syntax
```diff
@@ -704,10 +739,6 @@ contexts:
             3: punctuation.section.function.css
             4: constant.numeric.css
             5: punctuation.section.function.css
-        - match: (:)(active|any-link|hover|local-link|link|visited|focus)\b
-          scope: entity.other.attribute-name.pseudo-class.css
-          captures:
-            1: punctuation.definition.entity.css
         - match: '(?i)(\[)\s*(-?[_a-z\\[[:^ascii:]]][_a-z0-9\-\\[[:^ascii:]]]*)(?:\s*([~|^$*]?=)\s*(?:(-?[_a-z\\[[:^ascii:]]][_a-z0-9\-\\[[:^ascii:]]]*)|((?>([''"])(?:[^\\]|\\.)*?(\6)))))?\s*(\])'
```
wbond: I realize that this isn't part of your changes, but when testing the regex compatibility, it noted that it does not recognize `^ascii`. I'm just trying to understand what this is supposed to match in the first place. Perhaps it is meant to match unicode letters?

Briles: `[:^ascii:]` is a POSIX shorthand for matching non-ASCII characters

CSS/CSS.sublime-syntax
Briles: The other at-rules had defined `@` as `punctuation`. Should the scopes also be removed from those? See lines 44, 121, 132, 144, 186, and 215.

CSS/Comments.tmPreferences
FichteFoll: In that case, I'd like to argue how trailing spaces are allowed for line comment markers but not for blocks.

Examples:
https://github.com/sublimehq/Packages/blob/20bb6038d0fb54cc76dc4ef22e810f80fba909e9/Python/Miscellaneous.tmPreferences#L23
https://github.com/sublimehq/Packages/blob/20bb6038d0fb54cc76dc4ef22e810f80fba909e9/Java/Comments.tmPreferences#L17

I did not test how toggling commands with whitespaces in block sections behaves however, although it works fine for line comments.

Briles:
Regarding custom tags, the current regex does not allow for matching the wide-range of characters that the spec allows. A custom tag must take the form `[a-z](PCENChar)*-(PCENChar)*` where `PCENChar` can be any of the following:

```
-
.
[0-9]
_
[a-z]
\xB7
[\xC0-\xD6]
[\xD8-\xF6]
[\xF8-\x2FF]
[\x300-\x37D]
[\x37F-\x1FFF]
[\x200C-\x200D]
[\x203F-\x2040]
[\x2070-\x218F]
[\x2C00-\x2FEF]
[\x3001-\xD7FF]
[\xF900-\xFDCF]
[\xFDF0-\xFFFD]
[\x10000-\xEFFFF]
```

Those unicode ranges include the characters `~ + : >`, and others, which are special in CSS and therefore must be escaped by `\` when used in a custom tag.

Thankfully most custom tags in current use take the form `foo-bar`.

The W3C [working draft](https://www.w3.org/TR/custom-elements/#dfn-custom-element) and [editor's draft](http://w3c.github.io/webcomponents/spec/custom/#valid-custom-element-name) have more information.

CSS/CSS.sublime-syntax
wbond: This syntax appears to be a perl/oniguruma extension: http://linux.die.net/man/1/perlre. It does not appear to be part of the original posix character class functionality. Thanks for help in identifying that!

CSS/Comments.tmPreferences
wbond: I'm not interested in changing the whitespace on any comment markers at this point. There doesn't seem to be any functional benefit, just personal preference.

CSS/CSS.sublime-syntax
wbond: I think consistency is best, so I'd leave it the way you have it to match the rest of the syntax. Thanks for providing those examples.

wbond:
> Those unicode ranges include the characters ~ + : >, and others, which are special in CSS and therefore must be escaped by \ when used in a custom tag.

The ranges you specified don't appear to include any of those characters. (0x7E, 0x2B, 0x3A, 0x3E). The only one that seems to conflict would be `.` (0x2E).

Briles:
I was mistaken. That certainly simplifies things.

Briles:
I made the changes you asked for and I also implemented a custom tag regex that exactly matches the spec.

CSS/CSS.sublime-syntax
```diff
@@ -6,6 +6,24 @@ file_extensions:
   - css
   - css.erb
 scope: source.css
+variables:
+  custom_element_chars: '(?x)
```
FichteFoll: Don't do this. If you want to use multi-line extended regular expressions in variables, **always** wrap the entire content in `(?x: ... )` and ensure there is no trailing whitespace at the end of the YAML string. This will ensure that inclusion of the variable won't have any side effects on the actual pattern. Check out [YAML.sublime-syntax](https://github.com/sublimehq/Packages/blob/master/YAML/YAML.sublime-syntax#L58-L69) for examples.

In addition to everey whitespace following the variable insertion becoming insignificant, you also add pipe junctions in the level the variable was included it. This can be worked around by always wrapping the variable inside some group, capturing or non-capturing, but imo it's good practice to always store and use "normal" variables as atomic, unless specifically prefixed/suffixed.

Briles: Thanks for this!

wbond:
I added a commit afterwards that removed incompatible regex patterns and fixed a test.

PR #257, opened 2016-03-22
[JavaScript] Added missing console, process, and mutator methods

Briles:
Added JavaScript methods:

- `console.timestamp()` - [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Console/timeStamp)
- `Array.prototype.copyWithin()` - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/copyWithin)
- `Array.prototype.fill()` - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill)

Added Node.js [Process](https://nodejs.org/dist/latest/docs/api/process.html) methods:

*all methods are supported in Node.js LTS 4.4.0 and Node.js 5.9.0 Stable*

- `process.disconnect()` - [4.x](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_disconnect)
- `process.getegid()` - [4.x](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_getegid)
- `process.geteuid()` - [4.x](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_geteuid)
- `process.send()` - [4.x](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_send_message_sendhandle_callback)
- `process.setegid()` - [4.x](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_setegid_id)
- `process.seteuid()` - [4.x](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_seteuid_id)

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1021,7 +1021,7 @@ contexts:
         - include: function-call-params
     - match: '(?={{identifier}}\s*\.\s*{{identifier}}\s*\()'
       push:
-        - match: \b(console)(?:(\.)(warn|info|log|error|time|timeEnd|assert|count|dir|group|groupCollapsed|groupEnd|profile|profileEnd|table|trace))?\b
+        - match: \b(console)(?:(\.)(warn|info|log|error|time|timeEnd|assert|count|dir|group|groupCollapsed|groupEnd|profile|profileEnd|table|trace|timeStamp))?\b
           captures:
             1: support.type.object.console.js
             2: punctuation.accessor.js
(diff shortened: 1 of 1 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -1029,7 +1029,7 @@ contexts:
           set:
             - meta_scope: meta.function-call.method.js
             - include: function-call-params
-        - match: \b(process)(?:(\.)(abort|chdir|cwd|exit|getgid|setgid|getuid|setuid|setgroups|getgroups|initgroups|kill|memoryUsage|nextTick|umask|uptime|hrtime))?\b
+        - match: \b(process)(?:(\.)(abort|chdir|cwd|disconnect|exit|[sg]et[e]?gid|[sg]et[e]?uid|send|[sg]etgroups|initgroups|kill|memoryUsage|nextTick|umask|uptime|hrtime))?\b
```
michaelblyons: What is the `[e]?` for on the `gid` and `uid` captures? If that change isn't necessary, can you do `[sg]et[gu]id`?

Briles: the `[e]?` is necessary to match both  `egid` & `euid` variants of `get` and `set` while still being able to match the non `e` variants. I could have done `[sg]et[e]?[gu]id` but I was considering the readability of it at the time.

FichteFoll: `[sg]ete?[gu]id` works too and is perfectly readable to someone versed in regular expressions, imo.

wbond:
Thanks again for the contribution!

In general, I think that the slightly longer, more searchable/readable regexes are preferred. I know there are a number of complicated examples in the default packages, but my own opinion is that we should make them simpler to read and process, rather than shorter.

PR #259, opened 2016-03-27
[Diff] Fix Changed & Deleted Punctuation

Briles:
The previous Diff syntax was scoping `changed` & `deleted` punctuation as `inserted`. These changes fix these issues and adds tests.


wbond:
Thank you!

PR #260, opened 2016-03-28
[Rust] Kleene star operator is optional

dten:
https://doc.rust-lang.org/reference.html#macros states
> The Kleene star operator consists of $ and parentheses, optionally followed by a separator token, followed by * or +. 

so the operator should not be required and should be optionally anything except * or +

### DIFF
--- a/Rust/syntax_test_rust.rs
+++ b/Rust/syntax_test_rust.rs
@@ -782,15 +782,33 @@ macro_rules! alternate_group (
 
 macro_rules! kleene_star {
     ($($arg:tt)+) => (
+//   ^ source.rust meta.macro.rust meta.block.rust meta.group.rust keyword.operator.rust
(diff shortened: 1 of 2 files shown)


Rust/syntax_test_rust.rs
```diff
@@ -779,3 +779,18 @@ macro_rules! alternate_group (
         println!("Test {}!", $a)
     )
 )
+
+macro_rules! kleene_star {
+    ($($arg:tt)+) => (
+        println!($($arg));
```
wbond: You'll just need to add comments here in like in the lines above that annotate what scopes the character above the `^` has, to ensure the highlighting is working properly.

http://www.sublimetext.com/docs/3/syntax.html#Testing has more info.

wbond:
Thanks for catching this!

dten:
I added assertions. That all good?

PR #263, opened 2016-03-30
[YAML] Fix lexing to not interpret unquoted values with colons as keys

wbond:


FichteFoll:
Reviewed :+1: and following are some details:

The commit that broke the tests is 1def87e88c3c21910e3cf33e06d4687418aee5fe by removing the possessive quantifier, which then caused the regex engine to backtrack until before the last colon in the sequence after which the final colon in the pattern would match successfully, thus rendering the whole token as a mapping key. That way every color in a plain scalar would trigger it to be detected as a mapping key.

The solution is to either make it possessive again (not supported by sregex as of now) or explicitly only match the final colon if it's followed by a whitespace character.

I expanded the test suite to also contain a key that is followed by a line break: https://github.com/FichteForks/Packages/commit/c0d6fc8a2be4afaa926f36baa0bdfb8e3e2c6fca

wbond:
Just for future reader reference, the sregex engine is not a backtracking engine.

I believe an alternative could be written that would either re-order `match` statements, or possibly split the regex into multiple `match` rules. This should allow the lexer to accomplish the same sort of abandoning behavior as a possessive quantifier.

wbond:
Also, I cherry-picked https://github.com/FichteForks/Packages/commit/c0d6fc8a2be4afaa926f36baa0bdfb8e3e2c6fca.

PR #265, opened 2016-03-30
modified: HTML/html_completions.py#L296

Saket-Komawar:
Removed the default "text/javascript" type from `<script>` tag in HTML/html_completions.py file.
As JavaScript is the default scripting Language in HTML.

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -293,7 +293,7 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
             ('meta\tTag', 'meta ${1:charset=\"utf-8\"}>'),
             ('param\tTag', 'param name=\"$1\" value=\"$2\">'),
             ('progress\tTag', 'progress value=\"$1\" max=\"$2\">'),
-            ('script\tTag', 'script type=\"${1:text/javascript}\">$0</script>'),
+            ('script\tTag', 'script>$1</script>'),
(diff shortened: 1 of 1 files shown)


FichteFoll:
Closes https://github.com/SublimeTextIssues/DefaultPackages/issues/185.

wbond:
I understand the desire to omit things that aren't required, however, this makes the completion useless for users who are using a different language. That is why I haven't merged it yet.

For these kinds of changes, I am tempted to leave them as-is since there it seems to be more of a personal preference.

FichteFoll:
>this makes the completion useless for users who are using a different language.

What kind of language are you thinking of? 

I also agree that this is a matter of preference however.

Saket-Komawar:
JavaScript in HTML files
--
Sent from myMail app for Android Thursday, 14 April 2016, 00:01pm +05:30 from FichteFoll < notifications@github.com> :

wbond:
I've seen it used for JS payloads (http://handlebarsjs.com) and run on the server (VBScript I think?).

wbond:
I've made a compromise between the two desirable positions on this.

The completion now includes a tab trigger `$2` to allow users to delete the entire `type=""` attribute. Use the completion, hit tab once, delete, hit tab again.

8cd729f3a6736b697a2d61e70f474a30ba15589f

Issue #273, opened 2016-04-08
[C/C++] Wrong scope for null character in single 


jfcherng:
```c
char a = '\0';
```

The scope of `\0` is currently `invalid.illegal.unknown-escape.c`.

I just tried that `'\0'`, `'\1'`, ... and `'\7'` can be compiled but `'\8'` and `'\9'` can not be compiled.
https://github.com/sublimehq/Packages/blob/480f61413bcc5db640c5680b924e523a5cec4718/C%2B%2B/C.sublime-syntax#L91
Change `[4-7]\d?` to `[0-7]\d?`?

wbond:
This is actually a bug with the regex engine (the previous match of `[0-3]\d{,2}` should match), and the fix will be part of the next dev build.

PR #274, opened 2016-04-08
[C/C++] Fix scope for escaped null character (\0)

jfcherng:
ref: https://github.com/sublimehq/Packages/issues/273


wbond:
This is actually a bug with the regex engine which has been fixed, and will be part of the next dev build.

PR #275, opened 2016-04-09
[C/C++] String escape sequence improvements

jbw3:
I made the following changes:
* Fixed issue where some octal escape sequences were highlighted as invalid
* Allowing hexadecimal escape sequences to be more that 2 digits long
* No longer highlighting \p (this is not a valid escape sequence in the C or C++ standards)

I followed the specification found [here](http://en.cppreference.com/w/cpp/language/escape). The only difference is I still highlight \e. This is not part of the C or C++ standards, but is supported by some compilers (I tested with Clang).

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -88,7 +88,7 @@ contexts:
         - include: string_escaped_char
 
   string_escaped_char:
-    - match: '\\(\\|[abefnprtv''"?]|[0-3]\d{,2}|[4-7]\d?|x[a-fA-F0-9]{,2}|u[a-fA-F0-9]{,4}|U[a-fA-F0-9]{,8})'
+    - match: \\(?:\\|[abefnrtv\'"?]|[0-7]{1,3}|x[\da-fA-F]+|u[\da-fA-F]{4}|U[\da-fA-F]{8})
       scope: constant.character.escape.c
     - match: \\.
       scope: invalid.illegal.unknown-escape.c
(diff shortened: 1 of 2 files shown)


jfcherng:
A related one [here](https://github.com/sublimehq/Packages/pull/274) so you may want to revise your PR. (I didn't review your PR carefully but it looks like maybe related to this.)

jbw3:
@jfcherng Thanks for pointing that bug out. That was probably why the octal escape sequences were not highlighted correctly. My PR still fixes the hexadecimal issue, though. I don't think that is related to the regex engine bug.

C++/C.sublime-syntax
```diff
@@ -88,7 +88,7 @@ contexts:
         - include: string_escaped_char
 
   string_escaped_char:
-    - match: '\\(\\|[abefnprtv''"?]|[0-3]\d{,2}|[4-7]\d?|x[a-fA-F0-9]{,2}|u[a-fA-F0-9]{,4}|U[a-fA-F0-9]{,8})'
+    - match: \\(?:\\|[abefnrtv\'"?]|[0-7]{1,3}|x[\da-fA-F]+|u[\da-fA-F]{4}|U[\da-fA-F]{8})
```
wbond: Since the regex engine used by Sublime Text is unicode aware, `\d` will match the decimal digits on http://www.fileformat.info/info/unicode/category/Nd/list.htm. Hence, in this situation I believe `0-9` is more correct.

wbond:
BTW, I tweaked these a little to make them more strict, mostly in regards to `\d`. 00c6f92f501fb10e62a8e83fc093a4d0dbaad46e

PR #276, opened 2016-04-09
[CSS] Improve Color Function Scopes

Briles:
This pull request:

- Fixes an issue where the `blue` argument of the rgb color function was incorrectly scoped when its value was >= 200
- Excludes commas from the color function argument scopes


FichteFoll:
With how the regular expressions are built, the comma before the alpha value suddenly became optional while the alpha value is mandatory.

I suggest you add test the following:
```
rgba(66%, 66%, 66% 0.22);
rgb(66%, 66%, 66%);
```

wbond:
I realize this wasn't a change that was introduced with this PR, but I feel like the numeric values in `rgb()`, etc should be `constant.numeric.css`, not `constant.other.color`. My guess is that `constant.other.color` was based on the hex color colors, but it just seems odd to me that numbers aren't highlighted as numbers.

Briles:
The funny thing is the `numeric-values` are being included so numerics will still be matched if the color regexes do not match. Also, `numeric-values` also matches the hex color values so `rgb(#FFF)` would get scoped.

PR #280, opened 2016-04-13
[Batch / Graphviz / JSON] Added tests

Briles:
Simply adds some test files


wbond:
`syntax_test_batch_file.bat` is missing scope names for a bunch of the assertions

Briles:
Should be good now, I removed the assertions without scopes.

wbond:
Thanks for these

PR #281, opened 2016-04-13
[RegExp] Support non-range numeric quantifiers

michaelblyons:
Allow support for `a{3}` in addition to the existing `a{3,}`, `a{2,3}`, and `a{,3}`. If it is a problem that this makes `a{}` a "valid" regex, let me know, and I'll fix it up.

### DIFF
--- a/Regular Expressions/RegExp.sublime-syntax
+++ b/Regular Expressions/RegExp.sublime-syntax
@@ -25,7 +25,7 @@ contexts:
     - include: escaped_char
     - include: charset
     - include: group
-    - match: '([?*+][?+]?)|\{\d*,\d*\}'
+    - match: '([?*+][?+]?)|\{\d*,?\d*\}'
       scope: keyword.operator.quantifier.regexp
 
   group:
--- a/Regular Expressions/RegExp.sublime-syntax
+++ b/Regular Expressions/RegExp.sublime-syntax
@@ -25,7 +25,7 @@ contexts:
     - include: escaped_char
     - include: charset
     - include: group
-    - match: '([?*+][?+]?)|\{\d*,?\d*\}'
+    - match: '([?*+][?+]?)|\{\d+(,\d*)?\}'
       scope: keyword.operator.quantifier.regexp
 
   group:
--- a/Regular Expressions/syntax_test_regexp.re
+++ b/Regular Expressions/syntax_test_regexp.re
@@ -94,11 +94,17 @@
 a{9}
 #^^^ keyword.operator.quantifier.regexp
 
-a{,9}
-#^^^^ keyword.operator.quantifier.regexp
-
 a{1,9}
 #^^^^^ keyword.operator.quantifier.regexp
 
 a{9,}
 #^^^^ keyword.operator.quantifier.regexp
+
+a{,9}
+#^^^^ - keyword.operator.quantifier.regexp
+
+a{,}
+#^^^ - keyword.operator.quantifier.regexp
+
+a{}
+#^^ - keyword.operator.quantifier.regexp
--- a/Regular Expressions/syntax_test_regexp.re
+++ b/Regular Expressions/syntax_test_regexp.re
@@ -90,3 +90,15 @@
 (?#foobar)
 #^^^^^^^^^ string.regexp.group
 #^^^^^^^^ comment.line.number-sign
+
+a{9}
+#^^^ keyword.operator.quantifier.regexp
+
+a{,9}
+#^^^^ keyword.operator.quantifier.regexp
+
+a{1,9}
+#^^^^^ keyword.operator.quantifier.regexp
+
+a{9,}
+#^^^^ keyword.operator.quantifier.regexp


Regular Expressions/RegExp.sublime-syntax
```diff
@@ -25,7 +25,7 @@ contexts:
     - include: escaped_char
     - include: charset
     - include: group
-    - match: '([?*+][?+]?)|\{\d*,\d*\}'
+    - match: '([?*+][?+]?)|\{\d*,?\d*\}'
```
jfcherng: A regex `{}` just means literal `{}` and so does `{,}`.

May use this one.
`- match: '([?*+][?+]?)|\{,\d+\}|\{\d+,?\d*\}'`

image

Regular Expressions/RegExp.sublime-syntax
```diff
@@ -25,7 +25,7 @@ contexts:
     - include: escaped_char
     - include: charset
     - include: group
-    - match: '([?*+][?+]?)|\{\d*,\d*\}'
+    - match: '([?*+][?+]?)|\{\d+(,\d*)?\}'
```
FichteFoll: How does this regex match `{,9}`, as indicated by the tests?

Regular Expressions/RegExp.sublime-syntax
```diff
@@ -25,7 +25,7 @@ contexts:
     - include: escaped_char
     - include: charset
     - include: group
-    - match: '([?*+][?+]?)|\{\d*,?\d*\}'
+    - match: '([?*+][?+]?)|\{\d+(,\d*)?\}'
```
jfcherng: It looks like `{,9}` may be a quantifier in some languages. Now it goes to wbond's decision.

--

Just tried in ST's console.
```python
>>> re.search(r'a{,9}', 'b') is None
False
```
So `a{,9}` matches `b` in Python 3.3.

--

Meanwhile, for Javascript, the console in Google Chrome,
```javascript
> /a{,9}/.test('b')
false
```
So, `a{,9}` is treated as plain text in Javascript.

FichteFoll:
Maybe those invalid quantifier groups should get highlighted as `invalid`? I don't think any regex engine allows those and just errors instead.

I was actually surprised to see that `{,9}` is not valid, but it also makes sense because specifying a lower boundary is always possible, which can't be said for the higher boundary.

Regular Expressions/RegExp.sublime-syntax
FichteFoll: `a{,9}` doesn't actually match `b` though. It's an empty match.

```py
>>> import re
>>> m = re.search(r'a{,9}', 'b')
>>> m.start(), m.end()
(0, 0)
```

jfcherng: Sorry for the confusion. What I want to say is it would get a False since `a{0}` can match a empty string in the string `b`.

michaelblyons:
@FichteFoll See @jfcherng's line comment: It's a legitimate string in both Python 3 and Chrome's Javascript engine, but interpreted by one as a quantifier and by the other as a string of characters.

Personally, I think `{,4}` is undesirable as a quantifier: Do you want 0-4 of the preceding token? Or 1-4? Adding one more character makes it unambiguous. Highlighting syntax from someone else's code is important, too, though.

jfcherng:
@michaelblyons Since @FichteFoll has found `comment: Matches Oniguruma's Ruby regexp syntax` in the syntax file, then the syntax file is for Oniguruma unless wbond changes his mind. The target becomes clear so we should following Oniguruma's rules. Python or Javascript is not important now. :+1:

FichteFoll:
@michaelblyons: I edited my post two times, if you missed that.

The situation at hand is similar to the one for "Shell Script" where it has to be decided whether one syntax definition should "rule them all" or multiple flavors should be added for the different engines. Python already has its own definition anyway. ST uses Boost for its find an replace feature (where regex highlighting is primarily used afaik, in the widgets) and other syntaxes most likely include the global `source.regexp` scope in their languages without keeping 100% feature parity.

wbond:
When a `{}` pair is not a valid quantifier, it is used as literal characters to match on. This is true for the various engines I have tried.

Currently this syntax is used by the find panel, which ST3 currently uses boost for. The only other place it is referenced in the default packages is via Clojure. I have no idea what engine Clojure uses.

While Oniguruma (and Sublime's new regex engine as of build 3110) support `{,#}`, it appears to be a non-standard extension and is not supported by boost or other engines.

Since the primary function of this syntax is to highlight what Sublime Text will actually interpret as regex characters in the Find and Find in Files panels, we should *not* highlight `{}`, or `{,#}`. Only `{#}`, `{#,}` and `{#,#}` should be highlighted.

wbond:
The comment about Oniguruma should be removed.

jfcherng:
So it becomes something like `\{\d+,?\d*\}`.

image

michaelblyons:
The test cases line up with @wbond's comment, and the syntax definition passes those tests. The matching expression could be changed to @jfcherng's suggestion if people like that better, but it doesn't need to.

Thanks to @FichteFoll for pointing out the Oniguruma reference. I can remove it or change it to boost.

wbond:
This looks good, thanks!

PR #286, opened 2016-04-16
[PHP] Add support for anonymous classes

duncan3dc:
* Converted the class definition to a reusable context (`class-definition`)
* Added some tests for standard class definition/instantiation (to avoid regressions)
* Added support for anonymous classes (http://php.net/manual/en/language.oop5.anonymous.php)


wbond:
Thanks for all of these pull requests. I'm going to pull in all of your work here and then go through some of the official docs and try to clean up some other issues I find.

duncan3dc:
OK, feel free to ping me with any issues, I don't mind making sure things are up to scratch

PR #289, opened 2016-04-16
[PHP] Add support for unicode codepoint escape syntax

duncan3dc:
A new unicode codepoint escape syntax was added to PHP 7
http://php.net/manual/en/migration70.new-features.php#migration70.new-features.unicode-codepoint-escape-syntax

Also added tests for some of the other string interpolation features


wbond:
I fixed the definition of unicodepoint via e69de84. I referenced http://php.net/manual/en/language.types.string.php for the correct definition.

duncan3dc:
Whoops sorry, not sure where I got the 2 character group definition from, good catch

PR #292, opened 2016-04-16
[PHP] Add support for the CSPRNG functions

duncan3dc:
PHP 7 added some new random functions
http://php.net/manual/en/ref.csprng.php


PHP/PHP Source.sublime-syntax
```diff
@@ -942,6 +942,8 @@ contexts:
     - match: (?i)\b(com_(set|create_guid|i(senum|nvoke)|pr(int_typeinfo|op(set|put|get))|event_sink|load(_typelib)?|addref|release|get(_active_object)?|message_pump)|variant_(s(ub|et(_type)?)|n(ot|eg)|c(a(st|t)|mp)|i(nt|div|mp)|or|d(iv|ate_(to_timestamp|from_timestamp))|pow|eqv|fix|a(nd|dd|bs)|round|get_type|xor|m(od|ul)))\b
       scope: support.function.com.php
     - match: (?i)\bctype_(space|cntrl|digit|upper|p(unct|rint)|lower|al(num|pha)|graph|xdigit)\b
+      scope: support.function.csprng.php
```
wbond: This line should be moved down so it is applied to the `random_` functions, not `ctype_`.

duncan3dc: Of course, sorry. Fixed now :+1:

PR #294, opened 2016-04-18
[Ruby] Do not match camelCase as a constant

azdavis:
Previously, the uppercase portion of a camelCase variable was matched as
a constant. Adding a simple `\b` to the match rule should fix this.

Before

b

After

a


wbond:
Could you add a test for this? There is info about the syntax at http://www.sublimetext.com/docs/3/syntax.html#Testing.

azdavis:
Added it.

wbond:
Thank you!

PR #296, opened 2016-04-19
[Python] Fix string block scope starting with ws

FichteFoll:
All whitespace until the first non-whitespace character was not
correctly scoped and will be visible for color schemes with a background
color for strings.

Fixes what I mentioned [here](https://github.com/sublimehq/Packages/commit/76c4d5fcf31956c41d061b4daaef60c1ad4b67a3#commitcomment-16871122).

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -561,7 +561,7 @@ contexts:
     - match: '([uU]?r)(""")'
       captures:
         1: storage.type.string.python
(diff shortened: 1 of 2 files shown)


wbond:
This doubles the scopes on the initial string delimiter. I think it would need to be `meta_content_scope`.

FichteFoll:
Good catch. I will remove the extra scope on the quotation marks instead.

FichteFoll:
Done.

wbond:
Thanks!

PR #297, opened 2016-04-19
[Python] Simplify single quotes regexps

FichteFoll:
I couldn't stand just looking at it ...

There are also countless other strings that use quotation marks without needing them, but it seems pointless to change multiple hundred lines just for that when it works "just fine" right now.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -434,7 +434,7 @@ contexts:
         - match: (?<=""")
           pop: true
         - include: string_quoted_double
-    - match: '^\s*(?=[uU]?[rR]?'''''')'
+    - match: ^\s*(?=[uU]?[rR]?''')
       push:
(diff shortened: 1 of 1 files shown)


wbond:
I personally prefer all strings be quoted…

FichteFoll:
By "all strings" I assume you only mean regular expressions for `match` keys, right? So you want me to change all `(?=''')` to `"(?=''')"`?

wbond:
Really my point it that this is just sort of a preference. :smile:

I think tweaking things as part of functional changes makes sense, but absent a style document, sending style-only changes probably won't do much other than waste extra time reviewing potentially useful changes.

FichteFoll:
>Really my point it that this is just sort of a preference.

Yes, obviously, but it is my humble opinion that the maintainer's preferences are to be followed in pull requests. 

Imo this  change still makes the expressions much more readable.

PR #299, opened 2016-04-19
[Ruby] Fix: account for equality with constants

tmichel:
In `Symbol === x` Symbol is now treated as a constant and not a constant definition.


wbond:
Thanks!

PR #300, opened 2016-04-19
[TeX] Fix inline Math mode

Eyenseo:
This fixes syntax highlighting in the following case:
\lstinline${$ foo \lstinline$}$

In this case the string highlighting would go until the end of the second lstinline, with the change it highlights all as expected.

I'm not sure if anything breaks by this change!
I'm not fluent in the syntax scheme - my document looks fine but I don't use math stuff often.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -66,7 +66,6 @@ contexts:
         - match: \\\$
           scope: constant.character.escape.tex
         - include: scope:text.tex.math
-        - include: main
     - match: '(\\)[A-Za-z@]+'
       scope: support.function.general.tex
       captures:


wbond:
At the current time, I don't have enough experience with TeX to review this.

Anyone else use the TeX.sublime-syntax that could comment on this?

wbond:
@Eyenseo We'd definitely need some sort of tests to confirm this is a valid change and isn't regressed. See https://www.sublimetext.com/docs/3/syntax.html#Testing for more info.

buildersbrewery:
There're these files for Ace, although very simple ones:

* https://github.com/ajaxorg/ace/blob/master/demo/kitchen-sink/docs/tex.tex
* https://github.com/ajaxorg/ace/blob/master/lib/ace/mode/_test/tokens_tex.json

naturally these are tested against:

* https://github.com/ajaxorg/ace/blob/master/lib/ace/mode/tex_highlight_rules.js

buildersbrewery:
Looking at https://github.com/area/language-latex the syntax there includes `$base` within math tags, similar to how it was before the suggested change from this PR.

Eyenseo:
I had a look at https://ace.c9.io/tool/mode_creator.html and inserted `\lstinline${$ foo \lstinline$}$` from my example, which gets highlighted the same way as it would in sublime with my patch. But their rules differ quite a lot from sublimes - I would not want to assume that my patch does not break something.

Currently I'm busy with studies so I can't learn how to test and write a test for tex, if no one else does it in the mean time, I will probably have time for it in a few months.

buildersbrewery:
I wasn't implying that this PR breaks anything. Simply linking tests and grammar files from other sources for comparison.

Eyenseo:
@buildersbrewery Yes, never thought otherwise! :)

buildersbrewery:
@randy3k is re-writing LaTeX syntax at #370 after having opened #340. Maybe this change could included there.

buildersbrewery:
@Eyenseo

@randy3k finished with #370, maybe you could check if that PR fixes your issue with inline math mode and then close this? I do hope @wbond will merge it before the next ST Build.

PR #301, opened 2016-04-21
[PHP] Improve php5.4 array syntax

ptcong:


wbond:
This uses back references, which are not compatible with the new regex engine.

Instead, it will have to be rewritten by enhancing the PHP syntax to check for `[` after an identifier or function/method call and treat that as an index dereference, and all other situations treat it as an array literal. Something similar was done for `/` in JavaScript, where it can be a division operator or the beginning of a regex literal.

wbond:
I started some work on refactoring the syntax into statements and expressions. There will likely be some more similar work, and then we should be in a place where this should be able to be implemented by only looking forward.

ptcong:
ok, waiting to see it

wbond:
The changes from this were added in 81606041f923dcae5caa37fce6aae3f98d46b2cc.

PR #303, opened 2016-04-21
Move all snippets into their own folders

jbrooksuk:
For consistency all `.sublime-snippets` for each language are now within a `Snippets` directory.


wbond:
Thanks!

PR #305, opened 2016-04-22
[JavaScript] Exit concise arrow function scope on comma

jcberquist:


jcberquist:
Incidentally, to run the tests I used the Windows x64 portable build of 3111, and copied the repository into the `Data` folder. Running the tests in `syntax_test_js.js` with no changes made I get the following:
```
Packages/JavaScript/syntax_test_js.js:35:19: [variable.other.readwrite] does not match scope [source.js meta.function-call.js meta.group.js meta.object-literal.js meta.function.declaration.js entity.name.function.js]
```
In fact, I can edit the test file all the way down to:
```
// SYNTAX TEST "Packages/JavaScript/JavaScript.sublime-syntax"
someFunction({
    prop1
//  ^ variable.other.readwrite
});
```
And I still get that failure message. Note that if there is more than one item on that middle line (e.g. `prop1, prop2`), only the last item gets the failure, which makes me think there is something up with newline matching when running tests.

If I inspect the scopes at line 35, column 19 (using CTRL+ALT+SHIFT+P) the scopes are listed as expected, including `variable.other.readwrite.js`. It is also clear that the scope reported in the failure is coming from the `method-declaration` context, because I can edit the `entity.name.function.js` match there (Line 955) and the failure message changes to match.

Do you know why this would be happening?

keith-hall:
@jcberquist I wonder if it could be related to this https://github.com/SublimeTextIssues/Core/issues/1152

FichteFoll:
You might be running into https://github.com/SublimeTextIssues/Core/issues/963.

wbond:
I can't reproduce this issue at all on my OS X machine. I even tried switching line endings and that made no difference.

Is it possible you have another version of the JavaScript syntax somewhere on your machine?

I'll be checking my Windows VM later to see if I can reproduce it there.

keith-hall:
In case it helps, I have tested on Linux with build 3103 and 3111 with the JavaScript syntax shipped with those builds, and this test file passes. On 3111 on Windows with the JavaScript syntax shipped with the build, I get the error...

js_test_fail_on_windows

jcberquist:
@keith-hall: thanks! I can confirm the same on my Windows 10 x64 machine. If I use Unix line endings the test passes, but not with Windows line endings. Which explains why I am seeing the errors if I git clone the repo into ST but not if I download a ZIP of it...git must be converting all of the line endings to Windows.

wbond:
Ah, so the test fails with Windows line endings, but the syntax highlighting is still correct.

Most likely the code in the test running is stripping `\n` with the test assertion lines and not `\r\n`. I'll get that fixed for the next dev build.

wbond:
Mac line endings make the test all kinds of broken. :smile:

wbond:
Ok, all of the syntax test issues related to non-unix line endings will be fixed in the next dev build.

PR #306, opened 2016-04-22
[CSS] Improved attribute selectors

Briles:
Attribute selectors now support namespaces & flags.

`string` scope inclusions were re-organized in the process

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -71,8 +71,7 @@ contexts:
       push:
         - meta_scope: meta.at-rule.charset.css
         - include: at-rule-punctuation
-        - include: string-double
-        - include: string-single
+        - include: literal-string
   color-values:
       # http://www.w3.org/TR/CSS21/syndata.html#value-def-color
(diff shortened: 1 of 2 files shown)


CSS/CSS.sublime-syntax
```diff
@@ -820,46 +813,33 @@ contexts:
...
+                1: entity.other.namespace-prefix.css
+                2: entity.name.namespace.wildcard.css
+                3: punctuation.separator.namespace.css
+            - match: '[^\\ =<>`~*|^$\]\[]+'
+              scope: entity.other.attribute-name.css
+            - match: '\s*([~*|^$]?=)\s*'
+              captures:
+                1: keyword.operator.attribute-selector.css
+              push:
+                - meta_content_scope: entity.other.attribute-value.css
```
wbond: `entity.other` doesn't feel like the correct scope to be applied here. The HTML syntax uses `string.quoted.double`, `string.quoted.single` or `string.unquoted`. I think we'd want to use that here also.

Briles:
`entity.other.attribute-value` is now `string.quoted/unquoted`

wbond:
Awesome, thanks for these improvements!

PR #307, opened 2016-04-22
[Rails] Stop HTML (Rails) from initially matching everything

gkemmey:
While working on supporting HTML (Rails) highlighting in the Markdown Extended package, I noticed I couldn't end the markdown code block once I include the `text.html.ruby` syntax. This looked to be because of the initial `match: ''`.

Normally, this isn't a problem when opening a `.html.erb` file, but it was when trying to apply the syntax highlighting to a section of the file as the Markdown Extended package does.

This change should more closely follow the patterns in the HTML (Erlang), HTML (Tcl), and HTML (ASP) syntax definitions as well.

### DIFF
--- a/Rails/HTML (Rails).sublime-syntax
+++ b/Rails/HTML (Rails).sublime-syntax
@@ -8,26 +8,26 @@ file_extensions:
 scope: text.html.ruby
 contexts:
   main:
-    - match: ''
-      push: 'scope:text.html.basic'
-      with_prototype:
-        - match: "<%+#"
+    - match: "<%+#"
+      captures:
+        0: punctuation.definition.comment.erb
+      push:
+        - meta_scope: comment.block.erb
+        - match: "%>"
+          pop: true
+    - include: scope:text.html.basic
+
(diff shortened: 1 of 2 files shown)


wbond:
Did you try a `with_prototype` when pushing into the HTML (Rails) syntax?

wbond:
I added your tests, with some additions in 54e56b701c5e17d905f8d8826106441f380980e2. If you pull that down and try your proposed changes, you'll see the effect of what changing the `with_prototype` construct has.

`with_prototype` injects the listed rules into every nested context of the included context. Thus, when the HTML context goes into a new child context for handling HTML tag attributes, the matching rules for the ERB tags are added.

To accomplish what you want with markdown extended, you'll need to use `with_prototype` in that context to include HTML (Rails) and tell it that when you hit the end marker of the markdown fenced code block, that you need to pop out of whatever nested context you are in until you get back up to the Markdown syntax.

gkemmey:
Got it. I've made the correct changes on the Markdown Extended end. Thanks for your help!

PR #308, opened 2016-04-23
[XML] attributes: allow whitespace before equals sign

keith-hall:
fixes https://github.com/SublimeTextIssues/DefaultPackages/issues/93

### DIFF
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -210,7 +210,7 @@ contexts:
         - include: entity
         - include: should-be-entity
   tag-stuff:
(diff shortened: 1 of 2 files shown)


keith-hall:
also, @wbond, what would your opinion be on giving the equals sign a scope, maybe something like `keyword.operator.equals.xml`? :)

wbond:
Possibly, although I imagine most users aren't going to all of a sudden want their markup `=` highlighted like operators.

Perhaps something like `punctuation.separator.attribute.xml` would make more sense?

We should move this to an issue so it doesn't get lost as I merge this change.

PR #310, opened 2016-04-23
[XML] assign scope to attribute equals sign and support attributes beginning on a new line

### LINKED ISSUE #309: [XML] attribute equals sign has no specific scope
from discussion https://github.com/sublimehq/Packages/pull/308

I think it would be a good idea to give the equals sign between the attribute name and value a scope, maybe something like `punctuation.separator.attribute.xml`, which shouldn't affect highlighting if I understand right because most color schemes don't color punctuation separators :)

keith-hall:
for consistency with the HTML syntax

fixes #309

### DIFF
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -210,7 +210,7 @@ contexts:
         - include: entity
         - include: should-be-entity
   tag-stuff:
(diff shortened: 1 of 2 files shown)


keith-hall:
updated to also support attributes that start at the beginning of the line

PR #312, opened 2016-04-23
[CSS] Added support for keyframe selectors

Briles:
Added support for numeric values and the keywords `from` & `to` within keyframes at-rules.

As a side effect, class and id selectors now follow the spec more closely

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -7,6 +7,14 @@ file_extensions:
   - css.erb
 scope: source.css
 variables:
+  # Many variable names taken directly from https://www.w3.org/TR/css3-selectors/#lex
+  unicode: '\\\h{1,6}[ \t\n\f]?'
+  escape: '(?:{{unicode}}|\\[^\n\f\h])'
+  nonascii: '[\p{L}\p{M}\p{S}\p{N}&&[^[:ascii:]]]'
+  nmstart: '(?:[[_a-zA-Z]{{nonascii}}]|{{escape}})'
+  nmchar: '(?:[[-\w]{{nonascii}}]|{{escape}})'
+  ident: '-?{{nmstart}}{{nmchar}}*'
+
   custom_element_chars: |-
     (?x:
         [-_a-z0-9\xB7]
@@ -204,6 +212,7 @@ contexts:
           captures:
             1: punctuation.definition.arbitrary-repetition.css
           pop: true
+  # https://drafts.csswg.org/css-animations/#keyframes
   keyframes:
     - match: (?=\s*@(?:-webkit-|-moz-|-o-)?keyframes\s*.*?)
       push:
@@ -232,6 +241,11 @@ contexts:
           push:
             - match: '(?=\})'
               pop: true
+            - match: '\s*(?:(from|to)|((?:\.[0-9]+|[0-9]+(?:\.[0-9]*)?)(%)))\s*,?\s*'
+              captures:
+                1: keyword.keyframe-selector.css
+                2: constant.numeric.css
+                3: keyword.other.unit.css
             - include: main
   media:
     - match: (?=\s*@media\s*.*?)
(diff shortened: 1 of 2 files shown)


CSS/CSS.sublime-syntax
```diff
@@ -7,6 +7,14 @@ file_extensions:
...
 scope: source.css
 variables:
+  # Many variable names taken directly from https://www.w3.org/TR/CSS22/grammar.html#scanner
+  h: '[0-9a-fA-F]'
+  unicode: '\\{{h}}{1,6}(?:\r\n|[ \t\r\n\f])?'
+  escape: '(?:{{unicode}}|\\[^\r\n\f0-9a-fA-F])'
+  nmstart: '(?:[_a-zA-Z]|[^[:ascii:]]|{{escape}})'
+  nmchar: '(?:[_a-zA-Z0-9-]|[^[:ascii:]]|{{escape}})'
+  ident: '-?{{nmstart}}{{nmchar}}*'
+
```
wbond: Instead of defining `h`, use the `\h` character class escape. It matches `0-9`, `a-f` and `A-F`. Sublime Text normalizes the line endings in buffers to `\n`, so `\r\n` is not necessary. Thus, `unicode` can be `\\\h{1,6}[ \t\n\f]?` and `escape` can use `{{unicode}}|\\[^\n\f\h]`

The regex engines in Sublime Text are unicode aware, thus `nmstart` should probably be `[[:alpha:]_]|{{escape}}`. That way only unicode alphabetic characters will be matched as the beginning of an identifier, and not things like zero-width spaces. `nmchar` should be `[[:alnum:]_-]|{{escape}}`.

Briles: Regarding `nmstart` & `nmchar`: By only matching alphabetic/alphanumeric characters, a large portion of the character table is unusable even though the spec allows it. Was this your intention?

I tested about 20 characters from various non-alphabetic ranges in the latest version of Chrome and all were parsed correctly as the first character of a classname, except for the space characters.

wbond: In the case that it is desired to support such characters, unicode property escapes should probably be used: http://www.regular-expressions.info/unicode.html#category.

My guess is something like `[\p{L}\p{M}\p{S}\p{N}_&&\P{Nd}]`. That will be letters, marks, symbols, numbers and `_`, but minus decimal digit numbers (`\P{Nd}`).

wbond: The `&&\P{Nd}` would only be required in `nmstart`, not `nmchar`.

Briles: The `mark` & `symbol` categories include characters that need to be escaped — `~!@$%^&*()+=,./';:"?><[]\{}|#​​` etc. — and subcategories & scripts don't offer enough control to match what's needed.

Why not match all non-ASCII except the zero-width & spacing characters like so:

```
nonascii: [^[:ascii:]\x{200B}-\x{200D}\x{FEFF}\p{Z}]
nmstart: ( [_a-zA-Z] | {{nonascii}} | {{escape}} )
nmchar: ( [-\w] | {{nonascii}} | {{escape}} )
```

wbond: Ok, so we can use character set negations (http://www.rexegg.com/regex-class-operations.html#subtraction_javaruby) and unions (http://www.rexegg.com/regex-class-operations.html#union) here. This should result in faster scanning since it doesn't need to branch as with the `|` operator:

```
nmstart: [[_a-zA-Z][\p{L}\p{M}\p{S}\p{N}&&[^[:ascii:]]]]
nmchar: [[-_a-zA-Z0-9][\p{L}\p{M}\p{S}\p{N}&&[^[:ascii:]]]]
```

These are a union of (the listed ASCII characters) and (the unicode characters with those properties, minus all ASCII). Thus effectively limiting matches of the unicode properties to non-ASCII codepoints. This prevents us from accidentally matching any of those ASCII symbols you listed. I specifically 

I suppose we still have an escape branch, but at least we aren't matching junk unicode like control characters, and line and paragraphic breaks.

wbond: I just updated by comment to split the nonascii character class into a variable.

Briles:
I implemented the changes, everything should be be good to merge now.

wbond:
Thanks for all of the work on this!

PR #313, opened 2016-04-24
[CSS] Support unicode escapes within strings

Briles:
Simple change that adds support for Unicode code point escape sequences within strings.


CSS/CSS.sublime-syntax
```diff
@@ -886,5 +886,5 @@ contexts:
   string-content:
     - match: \\\s*\n
       scope: constant.character.escape.newline.css
-    - match: \\.
+    - match: '\\([a-fA-F0-9]{1,6}|.)'
```
wbond: You can use `\h` instead of `[a-fA-F0-9]`.

Issue #315, opened 2016-04-26
[Regular Expressions] quantifiers aren't valid everywhere and should be scoped as such


keith-hall:
currently, the syntax for Regular Expressions sets the `keyword.operator.quantifier.regexp` scope in places where the quantifiers are illegal. Scoping them as `invalid.illegal` will make it consistent with the results and therefore will become much nicer to work with.

for example:

- at the beginning of the expression
  - using `*` or `?` or `+` in the find panel gives:

    > the repeat operator * cannot start a regular expression
  - using `{1,2}` in the find panel gives:

    > Invalid preceding regular expression prior to repetition operator
- after a previous quantifier or after a `|`:
  - using `hello**` or `hello|+` or `1{2}{3}` in the find panel gives:

    > Invalid preceding regular expression prior to repetition operator

Issue #314, opened 2016-04-25
[Regular Expressions] Dot (.) should have a scope


michaelblyons:
I think that the `.` character should have a scope when not a part of a `[]` expression or otherwise escaped. It is an important part of regular expression syntax, and it would be a shame not to scope it and have someone imagine `.` to be a literal in their find/replace.

I do not, however, have strong opinions about what its scope should be... `keyword.other.any.regexp`? `constant.character.non-newline.regexp`? Something else?

wbond:
This should be resolved by 0bb48851a1fe4f88498373a79e9e52789ce2b11f

PR #316, opened 2016-04-26
[Regular Expressions] scope invalid quantifiers

keith-hall:
I thought I would share my solution for issue #315.

It is not perfect in my opinion, because it's not DRY, so I would be interested to see if anyone has a better solution :) EDIT: this has since been corrected :+1: 

Other improvements:
- it scopes literal characters as `meta.literal.regexp`
- it also scopes the `.` character as `keyword.other.any.regexp` as per #314 
- correctly scope mode modifiers i.e. `(?x)` or `(?-i)` as `meta.mode-modifier.regexp`

(if you don't like any of the scope names, we can change them)

Regular expressions don't tend to get that big, but I did a performance measurement anyway on a 400 byte file:
- Before my changes:

  > Syntax "Packages/Regular Expressions/RegExp.sublime-syntax" took an average of 0.1ms over 10 runs
- After my changes:

  > Syntax "Packages/Regular Expressions/RegExp.sublime-syntax" took an average of 0.2ms over 10 runs

### DIFF
--- a/Regular Expressions/RegExp.sublime-syntax
+++ b/Regular Expressions/RegExp.sublime-syntax
@@ -10,9 +10,17 @@ variables:
   known_char_escape: '\\(?:[tnrfae]|[0-7]{3}|x\{\h{1,7}\}|x\h\h|c\d+)'
   invalid_char_escape: '\\[xcCM]'
   char_escape: '\\.'
+  ranged_quantifier: '\{\d+(,\d*)?\}'
+  lazy_or_possessive: '[?+]?'
+  character_quantifier: '[?*+]'
+  char_class: '\\(?:[wWsSdDhHvVXR]|[pP](?:\{[a-zA-Z_]+\}|(L&|[A-Z][a-z]?)))'
 
 contexts:
   main:
+    - match: ''
+      push: [base-literal, unexpected-quantifier-pop]
+
+  base:
     - include: character-class
     - include: special-escaped-char
     - include: backslashes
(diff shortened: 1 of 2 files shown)


wbond:
I haven't looked at this in depth, but having looked at the implementation of the new regex engine in Sublime Text, it *may* (or may not) be easier to create contexts where quantifiers are known to be invalid, rather than trying to push into a context in every situation a quantifier may be valid.

Places where quantifiers may not exist:

 - After another quantifier
 - After `|`
 - Beginning of a group
 - Beginning of regex
 - Others?

FichteFoll:
Looks rather nice after some diff reviewing. The only comment I have is that due to the way `group-start` is structured (and used), the opening and closing round braces will not have the same scope assigned as the content of the group, which will mostly be visible in comment groups.

wbond:
`\Q` and `\E` are used to create segments of the regex that don't require escaping. It would make sense those can not have a quantifier after them. `\K` restarts the match, so same thing there.

I'll look a little more in-depth later.

Squashing this all into one commit probably makes sense now, and just make sure you trim trailing whitespace.

keith-hall:
My first time rebasing, so hopefully I did a good job :)

wbond:
I looks like you merged in changes from master.

Probably the simplest way (now) is to copy the two files you modified and then do a fresh branch from master, and paste the contents over. Then you'll need to force push to GitHub.

In the future, to squash you need to run `git rebase -i {hash}` where `{hash}` is the commit hash of the commit before the first commit you want to squash. That will open an editor pane where you can either "pick" (keep as-is) commits, edit then, squash, etc.

michaelblyons:
@wbond Assuming Keith hadn't rebased, don't you have the option to `merge --squash` his branch into master? Is that undesirable for commit message reasons?

keith-hall:
It seems I need to brush up on my git skills - sorry for the inconvenience everyone, and thanks for your patience.

keith-hall:
I think it's better now :)

keith-hall:
@FichteFoll, regarding:
> the opening and closing round braces will not have the same scope assigned as the content of the group, which will mostly be visible in comment groups

do I understand right, that you want the entire `(?#comment)` to be scoped as `comment.line.number-sign`, where currently (even before my changes), it doesn't scope the parens as such?

keith-hall:
I took the opportunity to add some more tests - I haven't squashed the commit, so that you can see what has changed, but if you want me to squash it before you merge, I can - I know what I'm doing now ;) 
let me know if you have any questions while reviewing this PR :)

wbond:
No need to squash this commit, just previously there were like 10 commits, some of which were undone by later commits.

wbond:
Awesome, thanks for all of the work on this @keith-hall!

FichteFoll:
@keith-hall 
> do I understand right, that you want the entire (?#comment) to be scoped as comment.line.number-sign, where currently (even before my changes), it doesn't scope the parens as such?

Yes, exactly. `(?#` and `)` serve as comment punctuation here, which all other languages do highlgiht as part of the comment and is imo correct. If this wasn't addressed in another commit later, I'll create a separate issue for it so that it won't be forgotten.

PR #318, opened 2016-04-27
Detect functions in Shell scripts correctly

ariofrio:
The following functions are not correctly highlighted, and do not appear in Goto Symbol, before this change (and in the current version of Sublime, Build 3111):

```
function x() {
}
function y {
}
```

The problem is that `function` gets caught as a keyword in the pipeline context, which makes it impossible for the function-definition context to catch the cases that begin with `function`.

### DIFF
--- a/ShellScript/Shell-Unix-Generic.sublime-syntax
+++ b/ShellScript/Shell-Unix-Generic.sublime-syntax
@@ -20,7 +20,6 @@ scope: source.shell
 contexts:
   main:
     - include: comment
-    - include: pipeline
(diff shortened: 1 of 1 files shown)


wbond:
To ensure that we don't regress this in the future, could you write a `syntax_test` file? http://www.sublimetext.com/docs/3/syntax.html#Testing

rwols:
Related: https://github.com/sublimehq/Packages/pull/1057

wbond:
This (and many other things) were resolved by #1057

PR #320, opened 2016-04-27
highlight st build and license using badges

buildersbrewery:
see commit for details


wbond:
I'm concerned it would be confusing to list the latest tag, implying that the code here is the code from that version. The tags are really so it is possible to cross-reference what versions included what changes.

The license file seems pretty easy to find for those interested.

I think between the above issues, and the fact that I've run into shields.io being extremely slow (and even offline) in the past, I think we'll pass on the badges for now.

I do appreciate your work, however!

PR #321, opened 2016-04-28
[ASP] Added syntax test file and fixed typos

keith-hall:


wbond:
Thanks!

michaelblyons:
This is maybe dumb, but I can't get the assertions to run. If I <kbd>Cmd</kbd>-<kbd>B</kbd> on either the `ASP.sublime-syntax` or `HTML-ASP.sublime-syntax`, I get

> Success: 0 assertions in 0 files passed
> [Finished]

If I try it on `syntax_test_asp.asp`, I get

> The current file is not a \*.sublime-syntax, \*.tmLanguage or syntax_test\* file

Is there something obvious I'm missing?

keith-hall:
For me it works on the syntax test file or the `HTML-ASP.sublime-syntax` file. Could it be that you are using a build variant or something?

michaelblyons:
Symlinking it into the live packages directory allowed the tests to run. I didn't realize that was a requirement. Thanks!

wbond:
@michaelblyons Oh, yes, it does require that since the syntax is loaded via the resource loader and used to actually highlight the file.

FichteFoll:
It should have shown a different error message however, iirc.

wbond:
@FichteFoll Yes, you are correct, it should have shown an explicit error message, although some tweaks to the logic we made recently to deal with trying not to generate false positives.

I should probably augment it to throw an error if a test file is found, but no assertions are made, and presume it is a symlink issue.

PR #323, opened 2016-04-28
[CSS] Scoped combinators

Briles:
The combinators `>>>`, `>>`, `>`, `~`, `+` are now scoped

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -32,6 +32,8 @@ variables:
 
   color_alpha_value: '(?:\s*,\s*((?:(?:0?\.[0-9]+)|[0-1])))?'
 
+  combinators: '(?:>{1,3}|[~+])'
+
 contexts:
   main:
     - include: comment-block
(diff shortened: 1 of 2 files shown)


CSS/CSS.sublime-syntax
```diff
@@ -725,6 +727,11 @@ contexts:
             1: punctuation.definition.entity.css
         - match: \*
           scope: entity.name.tag.wildcard.css
+          # Combinators
+          # https://drafts.csswg.org/selectors-4/#combinators
+          # https://drafts.csswg.org/css-scoping/#deep-combinator
+        - match: '({{combinators}})(?!{{combinators}})'
```
wbond: Do we need a negative lookahead here?

Briles: The negative lookahead is to ensure consecutive combinators don't match: `++` etc.

wbond: Won't that just result in the first `+` being unscoped? Perhaps instead we should match any further combinators and mark them as illegal?

Briles:
I updated it to match invalid combinator combinations

wbond:
Awesome, thanks!

wbond:
I was just about to merge this, however I am wondering about the scope name, and existing color schemes. What about using `punctuation.combinator` or `punctuation.separator.combinator`? There will probably already be fairly high-contrast/color symbols to the left and right of these due to tag, class and ID names. By making it `keyword.operator`, a lot of existing packages have that kind of high-contract/color also.

Briles:
I changed the scope to `punctuation.separator.combinator.css` as this offers the most flexibility for color scheme writers.

PR #325, opened 2016-04-28
[RegExp] Highlight comment punctuation

FichteFoll:
As I mentioned in this other PR (https://github.com/sublimehq/Packages/pull/316#issuecomment-215554171), the definiting puntuation of inline-comments `(?#` and `)` were not highlighted as part of the comment.

I also changed the scope name to `comment.block` instead of `comment.line`.


wbond:
Thanks!

PR #326, opened 2016-04-29
[CSS] Consolidate property values

Briles:
Consolidated the blend-mode and isolation-mode property values with the other property values


wbond:
In general I'd like to try and avoid regular expressions that split words in the middle to try and reduce the number of characters in the file. It really should not make a bit of different from a performance perspective, however it can make maintenance hard since it can be hard to figure out where certain keywords are scoped since you can't search for the word.

That said, plenty of the packages already use such constructs, but hopefully we can make them easier to read moving forward.

PR #328, opened 2016-04-29
[ASP] support line continuation

keith-hall:
[ASP] add missing / operator
[ASP] fix hexadecimal constants
[ASP] support comments after method definition
[ASP] support method definition with no parenthesis

### DIFF
--- a/ASP/ASP.sublime-syntax
+++ b/ASP/ASP.sublime-syntax
@@ -47,7 +47,7 @@ contexts:
     - include: comments
     - match: (?i:\b(If|Then|Else|ElseIf|End If|While|Wend|For|To|Each|In|Step|Case|Select|End Select|Return|Continue|Do|Until|Loop|Next|With|Exit Do|Exit For|Exit Function|Exit Property|Exit Sub)\b)
       scope: keyword.control.asp
-    - match: '&[hH][0-9a-fA-F]+&?|\b(((([0-9]+\.?[0-9]*)|(\.[0-9]+))((e|E)(\+|-)?[0-9]+)?)(L|l|UL|ul|u|U|F|f)?)\b'
+    - match: '&[hH]\h+&?|\b(((([0-9]+\.?[0-9]*)|(\.[0-9]+))((e|E)(\+|-)?[0-9]+)?)(L|l|UL|ul|u|U|F|f)?)\b'
       scope: constant.numeric.asp
     - include: line_continuation_char
     - match: '[=><>+*^&/-]|\b(?i:(Mod|And|Not|Or|Xor|Is))\b'
@@ -93,7 +93,7 @@ contexts:
         - meta_scope: comment.line.apostrophe.asp
         - match: $\n?
           pop: true
-    - match: 'REM '
+    - match: \bREM\b
       scope: punctuation.definition.comment.asp
       push:
         - meta_scope: comment.line.rem.asp
--- a/ASP/ASP.sublime-syntax
+++ b/ASP/ASP.sublime-syntax
@@ -8,25 +8,49 @@ file_extensions:
 scope: source.asp
 contexts:
   main:
-    - match: '^\s*((?i:function|sub))\s*([a-zA-Z_]\w*)\s*(\()([^)]*)(\)).*\n?'
-      scope: meta.function.asp
+    - match: '^\s*((?i:function|sub))\s*'
       captures:
         1: storage.type.function.asp
-        2: entity.name.function.asp
-        3: punctuation.definition.parameters.asp
-        4: variable.parameter.function.asp
-        5: punctuation.definition.parameters.asp
-    - match: (').*$\n?
-      scope: comment.line.apostrophe.asp
-      captures:
-        1: punctuation.definition.comment.asp
-    - match: (REM ).*$\n?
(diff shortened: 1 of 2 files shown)


keith-hall:
Performance tests:
-----

Amended syntax on the amended test file:

  > Syntax "Packages/ASP/HTML-ASP.sublime-syntax" took an average of 1.3ms over 10 runs

Syntax before this PR on the amended test file: (despite not all the tests passing with the old syntax)

  > Syntax "Packages/ASP/HTML-ASP.sublime-syntax" took an average of 1.3ms over 10 runs

Difference:
- None

ASP/ASP.sublime-syntax
```diff
@@ -59,7 +83,29 @@ contexts:
...
       scope: support.function.vb.asp
-    - match: '\b((0(x|X)[0-9a-fA-F]*)|(([0-9]+\.?[0-9]*)|(\.[0-9]+))((e|E)(\+|-)?[0-9]+)?)(L|l|UL|ul|u|U|F|f)?\b'
-      scope: constant.numeric.asp
     - match: (?i:\b(vbTrue|vbFalse|vbCr|vbCrLf|vbFormFeed|vbLf|vbNewline|vbNullChar|vbNullString|vbTab|vbVerticalTab|vbBinaryCompare|vbTextCompare|vbSunday|vbMonday|vbTuesday|vbWednesday|vbThursday|vbFriday|vbSaturday|vbUseSystemDayOfWeek|vbFirstJan1|vbFirstFourDays|vbFirstFullWeek|vbGeneralDate|vbLongDate|vbShortDate|vbLongTime|vbShortTime|vbObjectError|vbEmpty|vbNull|vbInteger|vbLong|vbSingle|vbDouble|vbCurrency|vbDate|vbString|vbObject|vbError|vbBoolean|vbVariant|vbDataObject|vbDecimal|vbByte|vbArray)\b)
       scope: support.type.vb.asp
+
+  comments:
```
wbond: Generally for single line comments, it is better to push into a new context that is exited when a `\n` is consumed.

This allow a `with_prototype` to be pushed that will exit the language in the middle of a comment line, such as when the language is embedded in HTML. This would apply to both the `comment.line.apostrophe` and `comment.line.rem`.

ASP/ASP.sublime-syntax
```diff
@@ -59,7 +83,29 @@ contexts:
...
+    - match: (').*$\n?
+      scope: comment.line.apostrophe.asp
+      captures:
+        1: punctuation.definition.comment.asp
+    - match: (REM ).*$\n?
```
wbond: This may be better as `(REM)\s`. This would allow a `\t` after `REM` and wouldn't highlight the whitespace if the user defines a custom color for the `punctuation`.

Actually, combined with the comment above about pushing into a context, this should probably just be `match: \bREM\b`, unless the language requires whitespace after the `REM`.

ASP/ASP.sublime-syntax
```diff
@@ -8,25 +8,49 @@ file_extensions:
...
+            - match: \s*\S+.*$
+              scope: invalid.illegal.unexpected-token-after-method-declaration.asp
+              pop: true
+            - match: ''
+              pop: true
+    - include: comments
     - match: (?i:\b(If|Then|Else|ElseIf|End If|While|Wend|For|To|Each|In|Step|Case|Select|End Select|Return|Continue|Do|Until|Loop|Next|With|Exit Do|Exit For|Exit Function|Exit Property|Exit Sub)\b)
       scope: keyword.control.asp
-    - match: '=|>=|<|>|<|<>|\+|-|\*|\^|&|\b(?i:(Mod|And|Not|Or|Xor|Is))\b'
+    - match: '&[hH][0-9a-fA-F]+&?|\b(((([0-9]+\.?[0-9]*)|(\.[0-9]+))((e|E)(\+|-)?[0-9]+)?)(L|l|UL|ul|u|U|F|f)?)\b'
```
wbond: `[0-9a-fA-F]` can be reduced to `\h`.

wbond:
I made a couple of small comments on tweaks that would be good. I realize you just moved them around, but we might as well improve them in the process.

ASP/ASP.sublime-syntax
keith-hall: it doesn't require the whitespace after the `REM`, so I will make that change, thanks :)

ASP/ASP.sublime-syntax
keith-hall: ooh, how did I not know that? :D

keith-hall:
@wbond, thanks - all good suggestions. I have now made those changes :)

ASP/ASP.sublime-syntax
keith-hall: for (my/future reader) info: `\h` is specific to the Ruby regex flavor. In other flavors (Perl, PCRE etc.), it matches horizontal whitespace. http://www.regular-expressions.info/shorthand.html

wbond: Yes, Sublime Text uses the oniguruma regex engine that is used by Ruby, and our custom regex engine is implemented to be mostly compatible with that, aside from the architectural limitations of no backreferences or lookbehinds.

PR #333, opened 2016-04-29
[Regular Expressions] fix #332 and [Clojure] fix and add tests

### LINKED ISSUE #332: [Regular Expressions] First character in Find bar is always meta.literal
I'm having an odd experience with the RegExp definition, wherein this syntax test will pass:

```re
\b
# <- keyword.control.anchors

\\b
# <- constant.character.escape - keyword.control.anchors
```

But if I put those same strings into the "Find" bar, I get different results:

```re
\b
# <- meta.literal

\\b
# <- meta.literal
# ^ keyword.control.anchors
```

This is also true of other strings whose first character should be something other than `meta.literal`: (`^`, `(`, `[`).

The same thing happens for Python if I change the pushed `source.regexp.python` references to `source.regexp`<sup>1</sup>:

```py
regex = re.compile(r'\b\w{1,20}')
#                    ^^ meta.literal
```

The Find bar actually *searches* correctly, but its highlighting is broken. This only happens if it is the first character. Subsequent instances are fine.

------

<sup>1</sup> Why doesn't it use the main Regular Expressions syntax by default? I was surprised to discover that there is a Python-only RegExp syntax file that isn't as robust.

keith-hall:
fix #332

### DIFF
--- a/Clojure/Clojure.sublime-syntax
+++ b/Clojure/Clojure.sublime-syntax
@@ -1077,11 +1077,14 @@ contexts:
         0: punctuation.definition.string.begin.clojure
       push:
         - meta_scope: string.regexp.clojure
-        - match: '"'
-          captures:
-            0: punctuation.definition.string.end.clojure
-          pop: true
-        - include: scope:source.regexp
+        - include: scope:source.regexp#unexpected-quantifier
+        - match: ''
+          set:
+            - meta_scope: string.regexp.clojure
+            - match: '"'
+              scope: punctuation.definition.string.end.clojure
+              pop: true
+            - include: scope:source.regexp#base-literal
   string_escape:
     - match: '\\(u[0-9a-fA-F]{4}|b|t|n|f|r|"|''|\\|[0-3]?[0-7]{1,2}|(.))'
       scope: constant.character.escape.clojure
--- a/Clojure/Clojure.sublime-syntax
+++ b/Clojure/Clojure.sublime-syntax
@@ -1077,14 +1077,17 @@ contexts:
         0: punctuation.definition.string.begin.clojure
       push:
         - meta_scope: string.regexp.clojure
+        - match: '"'
+          scope: punctuation.definition.string.end.clojure
+          pop: true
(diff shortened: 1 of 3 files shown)


keith-hall:
for some reason, pushing a context (using an empty match) directly from `main` seems to cause the first character to not get the correct scope. The fix is to set the context instead of pushing it.

wbond:
Hmm, I think we'll want to do this a different way since this could break if this syntax is embedded in another language (like it is with Clojure right now).

I may have to dig in to the ST source and see what is causing the issue.

keith-hall:
> I may have to dig in to the ST source and see what is causing the issue.

that would be perfect ;)

keith-hall:
I think it is already broken for Clojure because Regular Expressions uses a push in main, and the Clojure syntax just includes the Regular Expressions main context, and there is no `with_prototype` or anything to pop it off the stack when the closing `"` character is reached.

Unfortunately, the Clojure syntax doesn't have any tests, or I would have discovered it while developing my changes. (and it also shows how important it is to run *all* syntax tests before making a PR, not just the tests for the syntax being worked on.)

Confirmed by checking the scopes in this example I modified from the first "hello world clojure" result in google:

    (def hello (fn [] #"Hello world"))

so as it stands, I think we will need to make some changes to the Clojure syntax anyway?

keith-hall:
It looks like we can work around this ST3 bug by changing the Regular Expressions main context to:

    main:
      - include: unexpected-quantifier
      - match: ''
        push: base-literal

instead of:

    main:
      - match: ''
        push: [base-literal, unexpected-quantifier-pop]

but it doesn't fix the Clojure syntax

keith-hall:
@wbond, in case it helps you investigate what is causing the bug in the ST source, I have logged an issue with the minimal steps required to repro here: https://github.com/SublimeTextIssues/Core/issues/1190

keith-hall:
I've made a fix for Clojure and added tests :)

wbond:
Excellent, thanks for figuring out how to fix this!

PR #334, opened 2016-04-29
[Regular Expressions] More named captures

michaelblyons:
http://www.boost.org/doc/libs/1_44_0/doc/html/xpressive/user_s_guide.html#boost_xpressive.user_s_guide.named_captures

### DIFF
--- a/Regular Expressions/RegExp.sublime-syntax
+++ b/Regular Expressions/RegExp.sublime-syntax
@@ -69,12 +69,12 @@ contexts:
         1: keyword.other.backref-and-recursion.regexp
         2: keyword.control.group.regexp
       pop: true
-    - match: '(\?&\w+)(\))'
+    - match: '(\?(?:P=|&)\w+)(\))'
       captures:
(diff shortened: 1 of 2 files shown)


keith-hall:
I deliberately didn't include these because I discovered they don't work in the find panel.

michaelblyons:
You're right, it crashes the Find utility.

I'm going to keep the branch around for reference, but I'll close the PR and leave some notes here:

```re
(?<foo>a+)(b+)(?&foo)
# Successful ST3 "Find"
# Crashes Python2
# Crashes all parsers at https://regex101.com/

(?P<foo>a+)(b+)(?P=foo)
# Crashes ST3 "Find" with error: "Invalid or unterminated Perl (?...) sequence."
# Successful Python2 search
# Successfully parsed by Python and PCRE at https://regex101.com/

wbond:
@michaelblyons Sublime Text uses boost regex for Find/Find in Files, and it is generally influenced by PCRE syntax, although not as exhaustive.

PR #337, opened 2016-04-29
[Python] Set PYTHONIOENCODING for all platforms

FichteFoll:
According to the docs for `sys.std{in,err,out}`:

>The character encoding is platform-dependent. Under Windows, if the stream is interactive (that is, if its isatty() method returns True), the console codepage is used, otherwise the ANSI code page. Under other platforms, the locale encoding is used (see locale.getpreferredencoding()).

Thus, PYTHONIOENCODING should be set for all platforms.

Iterates on https://github.com/sublimehq/Packages/pull/295.

### DIFF
--- a/Python/Python.sublime-build
+++ b/Python/Python.sublime-build
@@ -3,10 +3,7 @@
 	"file_regex": "^[ ]*File \"(...*?)\", line ([0-9]*)",
 	"selector": "source.python",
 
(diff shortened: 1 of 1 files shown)


wbond:
Just a note, this is mostly a semantic change since usually Linux and OS X have the locale set to a UTF-8 variant.

PR #338, opened 2016-04-30
[PHP] fix new ClassName() highlight

ptcong:


PHP/PHP Source.sublime-syntax
```diff
@@ -335,7 +335,7 @@ contexts:
             - match: (?=^\1\b)
               pop: true
   instantiation:
-    - match: '(?i)(new)\s+(class)\s*'
+    - match: '(?i)(new)\s+(class)\s+'
```
wbond: This will break valid code such as `$var = new class{}`;

wbond:
I believe I've fixed the issue via ba14c83f1bc3722378db9ba6ee9800d5a70d6b40 to continue allowing `$var = new class{};` to be properly highlighted.

PR #341, opened 2016-05-02
[CSS] Refined function support

Briles:
Not all functions were previously supported and those that were implemented were supporting keywords or values that are not compatible as an argument for that function.

For example:

1. the `cubic-bezier()` timing function was matching the keywords `end` & `start` which are only supported by the `steps()` timing function

2. `url()` was matching the property values `min-content` & `max-content` which are only supported by the `repeat()` and `minmax()` functions

---------

This PR changes some existing implementations:

1. **color functions** - `rgb(), rgba(), hsl(), hsla()`

 color function arguments no longer have specific scopes. *ie*: `constant.other.color.hsl-value.css` is now just `constant.numeric.css`. This seems more fitting as the arguments themselves are just numeric values.

2. **hex colors** were moved out of `numeric-values` and into `color-values`

3. **numeric-values** are now implemented in a modular fashion.

 This was necessary to allow for functions to `include` only the types of arguments they need.

--------

Things that need improvement / feedback:

1. Is the approach of including only needed types an improvement over the previous implementation?

 Ideally, all not-allowed arguments for each individual function would be scoped as `invalid` but such an implementation would require extensive work and a lot of duplicate/similar constructs.

 With the changes in this PR, the not-allowed arguments are simply not specifically scoped.

2. Grouping functions by the types of arguments *-VS-* providing a unique scope/meta_scope for each function

 *ie*: `translateX()` & `translateY()` are grouped because they both take a single `<number>` or `<length>` type  argument. This prevents each individual function from having unique scopes/meta_scopes like some of the other functions have, such as `hue-rotate()` which has its name scoped as `support.function.filter.hue-rotate.css` and pushes a `meta_scope` of `support.type.function.filter.css` because it belongs to the `filter` category of functions.

3. Inconsistent function scopes & meta_scopes

4. There are some scopes that are commented with a `TODO` tag that require feedback

--------

This PR adds support for the following functions:

```css
{
any-property: toggle();
background: linear-gradient();
background: radial-gradient();
background: repeating-linear-gradient();
background: repeating-radial-gradient();
color: alpha();
color: blackness();
color: blend();
color: blenda();
color: blue();
color: color-mod();
color: device-cmyk();
color: gray();
color: green();
color: hue();
color: hwb();
color: lightness();
color: red();
color: saturation();
color: shade();
color: tint();
color: whiteness();
filter: blur();
filter: brightness();
filter: contrast();
filter: drop-shadow();
filter: filter();
filter: grayscale();
filter: hue-rotate();
filter: invert();
filter: opacity();
filter: saturate();
filter: sepia();
transform: matrix();
transform: matrix3d();
transform: perspective();
transform: rotate();
transform: rotate3d();
transform: rotateX();
transform: rotateY();
transform: rotateZ();
transform: scale();
transform: scale3d();
transform: scaleX();
transform: scaleY();
transform: scaleZ();
transform: skew();
transform: skewX();
transform: skewY();
transform: translate();
transform: translate3d();
transform: translateX();
transform: translateY();
transform: translateZ();
}
```

----------------------
#### Performance:

`// 35K lines ; 1,427,552 chars ; 1.36MB`

Before:
```
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 214.7ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 212.6ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 212.7ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 212.5ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 213.7ms over 10 runs
```

**Average: 213.24ms**

--------------------

After:
```
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 221.1ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 221.4ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 219.8ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 220.1ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 220.1ms over 10 runs
```
**Average: 220.5ms**

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -15,6 +15,21 @@ variables:
   nmchar: '(?:[[-\w]{{nonascii}}]|{{escape}})'
   ident: '-?{{nmstart}}{{nmchar}}*'
 
+  # Types
+  # https://www.w3.org/TR/css3-values/#numeric-types
+  integer: '(?:[-+]?\d+)'
+  number: '[-+]?(?:(?:\d*\.\d+(?:[eE]{{integer}})*)|{{integer}})'
+
+  # Units
+  # https://www.w3.org/TR/css3-values/#lengths
+  font_relative_lengths: '(?i:em|ex|ch|rem)'
+  viewport_percentage_lengths: '(?i:vw|vh|vmin|vmax)'
+  absolute_lengths: '(?i:cm|mm|q|in|pt|pc|px|fr)'
+  angle_units: '(?i:deg|grad|rad|turn)'
+  duration_units: '(?i:s|ms)'
+  frequency_units: '(?i:Hz|kHz)'
+  resolution_units: '(?i:dpi|dpcm|dppx)'
+
   custom_element_chars: |-
     (?x:
         [-_a-z0-9\xB7]
@@ -34,13 +49,26 @@ variables:
       | [\x{00010000}-\x{000EFFFF}]
     )
 
-  rgb_decimal_value: '(?:0*(?:(?:2(?:[0-4][0-9]|5[0-5]))|(?:1?[0-9]{1,2})))'
-
-  color_percentage_range: '(?:100|[0-9]{1,2})%'
-
-  color_alpha_value: '(?:\s*,\s*((?:(?:0?\.[0-9]+)|[0-1])))?'
-
   combinators: '(?:>{1,3}|[~+])'
+
+  # Predefined Counter Styles
+  # https://drafts.csswg.org/css-counter-styles-3/#predefined-counters
+  counter_styles: |-
+    (?xi:
+        arabic-indic | armenian | bengali | cambodian | circle
+      | cjk-decimal | cjk-earthly-branch | cjk-heavenly-stem | decimal-leading-zero
+      | decimal | devanagari | disclosure-closed | disclosure-open | disc
+      | ethiopic-numeric | georgian | gujarati | gurmukhi | hebrew
+      | hiragana-iroha | hiragana | japanese-formal | japanese-informal
+      | kannada | katakana-iroha | katakana | khmer
+      | korean-hangul-formal | korean-hanja-formal | korean-hanja-informal | lao
+      | lower-alpha | lower-armenian | lower-greek | lower-latin | lower-roman
+      | malayalam | mongolian | myanmar | oriya | persian
+      | simp-chinese-formal | simp-chinese-informal
+      | square | tamil | telugu | thai | tibetan
+      | trad-chinese-formal | trad-chinese-informal
+      | upper-alpha | upper-armenian | upper-latin | upper-roman
+    )
 
 contexts:
   main:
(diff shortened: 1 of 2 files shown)


wbond:
I think making the scopes for color function arguments more generic is a good step forward and more consistent with other syntaxes.

Feedback items:

 1. Generally I'm starting to shy away from `invalid` scopes since it can be hard to implement them in such a way that they are consistently correct. I'd rather not give feedback than give incorrect feedback. Hopefully I'll be going through the various syntaxes and removing things like the invalid scope on `%` signs that aren't followed by a valid `sprintf()` placeholder char. Often times these highlight perfectly valid strings with `%` in them as invalid when they really aren't.
 2. I'd actually prefer generally not to get too specific with function names in scopes. I'd rather have some high-level categorization if necessary, but individual function names gets beyond what the syntax is supposed to be accomplishing. Really these syntaxes are for highlighting different types of syntax and extracting some high-level information for the indexer.
 3. For consistency with other syntaxes, a function name, the parens and all contents should have the context `meta.function-call.css`. I think I prefer `meta.group` since we can use that for any pair of parens in any language. Some syntaxes currently use `meta.group.css` for `(` to `)`, but others use `meta.parameters.css`. Depending on what meta scope is used the parens get `punctuation.definition.group.begin.css` and `punctuation.definition.group.end.css` or `punctuation.definition.parameters.begin.css` and `punctuation.definition.parameters.end.css`.
 4. On line 849, I wouldn't both limiting the valid syntax there. I don't have an opinion about line 985. Line 1505 looks good.

I think the scope `punctuation.operator` should be changed to `keyword.operator` for consistency with other syntaxes. This is relevant to `+`, `*`, etc inside of `calc()`. I don't know if there are ever such operators outside of `calc()`, but I don't recall such usage off of the top of my head.

I'm going to attempt to make some comments on the changes, but it seems GitHub thinks there are too many for me to view them, so I don't know if I'll be able to comment on specific lines.

wbond:
Yeah, so GitHub PR fail here. I'm going to leave comments on context names that you can hopefully reference.

 - `counter_styles:` - `(?xi)(?:` can be written as `(?xi:`
 - Can `color-values` have the negative lookahead on `blue|gray|green|red` removed by placing the context that matches the function version before `color-values`? Generally consuming the most specific syntax first helps improve performance since negative lookaheads seem to be a slight bit of a performance hit.
 - `custom-property-name:` - the leading `--` should probably not be `keyword.other`, but something like `punctuation.definition.custom-property`.
 - `color-adjuster-operators:` - just drop the `.assignment` part, that is just for `=`.
 - `comma-delimiter:` the consistent scope for this should be `punctuation.separator.css`.

Briles:
I went ahead and implemented changes that address the points you made.

1. The newly implemented `invalid` scopes were removed as they weren't matching all invalid cases. No previously implemented invalid scopes were touched.

2. Function names are now `support.function.{{category}}.css` where `category` is the group they belong to in the spec; *ie*: `rgb()` is  `support.function.color.css`. If a function has no group or is in a group by itself, the function name is used as the category; *ie*: `calc()` is `support.function.calc.css`.

3. Function meta scopes are now implemented in a manner consistent with other languages (JavaScript specifically). The function scopes are now:

    ```css
.function-meta {
        top: rgb(param1, 20px);
    /*       ^^^^^^^^^^^^^^^^^ meta.function-call.css                     */
    /*       ^^^               support.function.color.css                 */
    /*          ^^^^^^^^^^^^^^ meta.group.css                             */
    /*          ^            ^ punctuation.definition.group.begin/end.css */
}
    ```

4. `counter_styles` - `(?xi)(?:` is now `(?xi`

5. The `color-values` lookahead was removed. Comments were added to convey the importance of the order of `include: color-values`

6. `custom-property-name` leading `--` is now scoped as `punctuation.definition.custom-property.css`

7. `color-adjuster-operators` are now just `keyword.operator.css`

8. `meta.delimiter.comma.css` is now `punctuation.separator.css`

wbond:
Awesome, thanks for all of the work on this!

PR #343, opened 2016-05-02
[ASP] correctly close ASP scope when in a sub context

keith-hall:
[ASP] support mixed case "REM" comments
[ASP] better handling of methods missing closing parens


ASP/HTML-ASP.sublime-syntax
```diff
@@ -14,9 +14,10 @@ contexts:
...
-          scope: comment.line.apostrophe.asp
-          captures:
-            1: punctuation.definition.comment.asp
-        - include: scope:source.asp
+        - match: ''
+          push:
+            - include: scope:source.asp # don't directly push, so that the scope doesn't also contain 'source.asp ' as well as our 'source.asp.embedded.html '
+          with_prototype:
+            - match: "(?=%>)"
+              pop: true
```
wbond: I was thinking about this the other day – should we allow the ASP syntax to embed HTML inside of it instead of popping out of all nested context back up to the parent HTML context?

If we allow embedding HTML inside of nesting ASP contexts, then the ASP syntax should better be able to track when it is still inside of a function, etc. We would probably only want to push into a nesting HTML context if we have entered a nested ASP context. If we are still at the top-most level, we should probably still use the match of `%>` in the initial `<%=` push context.

Does the question make sense, or should I try to explain my idea a different way?

michaelblyons: I am invested in how this goes, because a C# HTML syntax I'm working on has a similar problem. (I've taken some cues from the ASP HTML syntax and some from the PHP syntax.) This is a valid pattern in that language:

```html
<div>
<% if (condition) { %>
    <p>Hello World</p>
<% } %>
</div>
```

The problem is that C# doesn't give the scope back to HTML in the `if{...}` body, even though the C# scope should be closed with `%>`. If you guys take a route with ASP that makes such patterns possible, I'm going to model the C# HTML after it.

I can't quite tell what @wbond is saying. 😕

keith-hall: I understand what you are proposing, and I think that it is a great idea.  However, the way the syntax works at the moment is mainly just very basic "keyword" matching - it isn't very intelligent; there are no pushed contexts while inside a function body or anything, and it currently allows all sorts of invalid keyword/control-flow combinations: `set exit for nothing / 2 "hello"  end ubound vbTextCompare = if`  anyone? ;)

So I think that when the ASP syntax is improved to parse it properly, it would be the perfect time (and probably necessary, even) to implement your suggestion. Until then, I had the idea to build it so that it has at least basic passing tests for valid code or known behaviors, to act as a solid foundation for further improvements.

This is the main reason for this PR, to test that a `%>` in a string literal is treated the same way as the ASP interpreter treats it - i.e. not as part of the string but to immediately close the ASP block.

keith-hall: @michaelblyons, I imagine it should work a bit like this, in case it helps to clarify:

    <div> 
    <% // root level <%, so the scope would be something like `text.html.basic source.cs.embedded`
       if (condition) { %> 
          <p>Hello World</p>
    </div>

jcberquist: I am interested in this, too. While I think it can work well to embed the HTML syntax, one of the challenges (I think) of embedding is that the HTML syntax includes the JavaScript and CSS syntaxes. So it is possible (if for instance the ASP/PHP/etc code is inside of an HTML `<script>` block) that what needs to be embedded is not HTML but JavaScript or CSS.

wbond: @jcberquist That is a VERY good point. However, if you are defining an ASP function inside of `<%=` inside of a `<script>` tag, and that function switches back to HTML, then… you are going to have other problems. :smile:

I think the distinction here is that is should only re-enter a new sub-context of HTML if it is inside of a nested ASP context like a function. Otherwise I do believe it should pop up to the original HTML syntax.

There may be implementation details I am not thinking of.

keith-hall: I just had a quick check with ASP to ensure that it is not valid to split up statements in separate blocks like this:

    <% Dim %><% var_name %>

and fortunately it isn't, so resetting to a new root code context each time is fine when not inside a nested block.

keith-hall: As you have expressed your interest (@jcberquist and @michaelblyons), you may want to know that I have [created a PR](https://github.com/sublimehq/Packages/pull/367) to make the changes for remembering the scope when moving between embedded HTML and ASP. So far it doesn't support being inside script or style tags etc. but its a step in the right direction, and I'd welcome your input on how it can be improved :)

PR #349, opened 2016-05-03
[CSS] Loosely match property names

Briles:
This change is purely for performance and maintainability. The functionality remains the same; all 858 in-use property names I tested were scoped the same before and after the changes.

----------------------
#### Performance:

`// 32K lines ; 671,221 chars ; 655KB`

Before:
```
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 311.4ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 306.3ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 297.6ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 307.5ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 310.6ms over 10 runs
```

Average: **306.68ms**

--------------------

After:
```
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 268.0ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 271.5ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 265.9ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 267.4ms over 10 runs
Syntax "Packages/CSS/CSS.sublime-syntax" took an average of 273.2ms over 10 runs
```
Average: **269.2ms**

Difference: **-37.48ms** :tada:


wbond:
Generally I like performance improvements.

I'm curious if there is any difference just making a big `|` separated list of property names? I don't know if the grouping makes any difference or not.

I kind of like it that only known properties are highlighted. I imagine it helps prevent typos, etc.

wbond:
Do you have the sample CSS file online somewhere? I'd love to mess around a little with it to see what affect different regexes may have on performance.

Briles:
I have my testing files at [https://github.com/Briles/syntax-testing-files](https://github.com/Briles/syntax-testing-files)

Briles:
I ran some tests using ST v3109. All tests were run on my 35K line test file in the repo I linked previously, unless otherwise noted. Units are in `ms`

Tests are described as follows:

- *Not Compacted* - `(foobar|food|foot)`
- *Max Compacted* - `foo(bar|d|t)`
- *Usage Sorted* - for these tests I used data from [Chrome Platform Status](https://www.chromestatus.com/metrics/css/popularity) to sort the property names by most used in descending order. The idea here was to increase performance by placing the more popular property names earlier in the capture group, allowing for an early match and exit.

Set 1
---
For the first set of tests I reduced the syntax file to a single capture that matched property values, colors, property names, and font names.

|         Test         | Average |  Min  |  Max  |
|:--------------------:|:-------:|:-----:|:-----:|
|        Max Compacted |  68.54  | 68.2  | 68.8  |
|        Not Compacted |  68.58  | 68.3  | 68.7  |
| Max Compacted - 140K |  270.96 | 270.2 | 271.8 |
| Not Compacted - 140K |  271.66 | 270.0 | 272.6 |

-----------------------------

Set 2
---
The second set of tests were performed using the latest syntax file from this repo. The only thing changed between tests was the `support.constant.property-name` capture contents

|            Test            | Average |  Min  |  Max  |
|:--------------------------:|:-------:|:-----:|:-----:|
| {{ident}} (this PR)        | 348.58  | 346.1 | 351.2 |
| Usage Sorted Not Compacted | 381.18  | 380.4 | 389.3 |
| No Change                  | 395.74  | 391.4 | 404.3 |
| Usage Sorted Max Compacted | 397.94  | 392.1 | 410.7 |
| Not Compacted              | 413.12  | 411.3 | 415.7 |
| Max Compacted              | 477.68  | 467.4 | 485.7 |

wbond:
So, Set 1 would alter the functionality of the syntax by getting rid of specific scope names for different syntax, correct?

It seems to me that Usage Sorted Not Compacted seems to me like the best choice. This way users will still see if they typo a property name, but there is a bit of performance improvement.

Thanks for taking the time to run through the variations, it is very helpful in making decisions.

Briles:
The reason for Set 1 was just to test the performance of the alternations by themselves. I concatenated the lists of property values, colors, property names, and font names to simply get the most alternations possible.

I was somewhat surprised to see no change in performance as constructing a trie from the alternations is a common optimization (at least in Perl).

The property values, font names, and colors could be combined, but this would affect each inclusion of `color-values` and any performance benefit would probably be lost as we would be including a large group of alternations instead of just color values.

Briles:
Property names are now ordered by usage

wbond:
I think the reason you did not see the performance gains from optimizations common to PCRE patterns is that the Sublime regex engine is non-backtracking. I imagine the tree-type structure helps prevent the backtracking. Sublime doesn't use backtracking, hence you don't need to try and prevent it. This is also why atomic subgroups are not generally required for correctness since they are normally used to prevent backtracking.

Thanks for all of the work on this and proving data for all of your tweaks.

PR #350, opened 2016-05-03
[Batch] More consistent strings & comments

Briles:
Implements strings & comments with scopes consistent with other languages.

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -8,6 +8,8 @@ file_extensions:
 scope: source.dosbatch
 contexts:
   main:
+    - include: comments
+    - include: strings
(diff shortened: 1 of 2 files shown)


Batch File/Batch File.sublime-syntax
```diff
@@ -20,22 +22,37 @@ contexts:
...
+    - match: '&>|\d*>&\d*|\d*(>>|>|<)|\d*<&|\d*<>'
+      scope: keyword.operator.redirect.shell
+
+  comments:
+    - match: '(:[+=,;:])'
+      scope: punctuation.definition.comment.dosbatch
+      push:
+        - meta_scope: comment.line.colon.dosbatch
+        - match: (?=\n)
+          pop: true
```
wbond: Recently we've been switching to line comments consuming the trailing linebreak. This is useful because the linebreak gets the `comment` scope, which allows certain features like completions from showing while typing a line comment out.

Briles:
Line comments now consume the trailing linebreak.

wbond:
Thanks!

PR #353, opened 2016-05-03
Create For loop using 'auto' as the iterator type

bryanj4:
I was stuck in between C++03 and C++11 with MS Visual C++10, so I couldn't use the 'for range' for loop and the forv.sublime-snippets was pointless. I could use 'auto' because it is implemented in VC++10 and gcc4.4 which are the main compilers.

### DIFF
--- a/C++/Snippets/fora.sublime-snippet
+++ b/C++/Snippets/fora.sublime-snippet
@@ -1,0 +1,13 @@
+<snippet>
+    <content><![CDATA[
+for(auto ${2:i} = ${1}.begin(); ${2:i} != ${1}.end(); ${2:i}++)
+{
+    ${3}
+}
+]]></content>
+    <!-- Optional: Set a tabTrigger to define how to trigger the snippet -->
+    <tabTrigger>fora</tabTrigger>
+    <!-- Optional: Set a scope to limit where the snippet will trigger -->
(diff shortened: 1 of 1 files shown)


C++/Snippets/fora.sublime-snippet
```diff
@@ -0,0 +1,13 @@
+<snippet>
+    <content><![CDATA[
+for(auto ${2:i} = ${1}.begin(); ${2:i} != ${1}.end(); ${2:i}++)
```
jfcherng: IMO, as in https://github.com/sublimehq/Packages/blob/master/C%2B%2B/Snippets/forv.sublime-snippet,
- No new line after `<![CDATA[`.
- A space between `for` and `(`.

C++/Snippets/fora.sublime-snippet
```diff
@@ -0,0 +1,13 @@
...
+{
+    ${3}
```
jfcherng: Indent with `Tab` before `${3}`. ST will convert it depends on the situation.

wbond:
I'm not sure I want to open the flood gates of adding snippets, largely because there tend to be many aspects that are personal preference.

IMO, these are better managed as packages available via Package Control, or implemented on a per-user basis.

wbond:
I'm going to close this since I don't think we should get into trying to add snippets to the default packages.

PR #355, opened 2016-05-04
[Batch] Updated commands, added variables

Briles:


Batch File/Batch File.sublime-syntax
```diff
@@ -25,6 +26,31 @@ contexts:
...
       scope: keyword.command.dosbatch
 
+  variables:
+    # global parameters
+    - match: '(%)((~([fdpnxsatz]|\$PATH:)*)?\d|\*)'
+      scope: variable.parameter.dosbatch
```
Briles: I'm somewhat torn on this scope; `variable.parameter` is typically used for functions while these are the arguments passed to the script.

Batch File/Batch File.sublime-syntax
```diff
@@ -22,6 +22,35 @@ contexts:
...
+      scope: keyword.command.dosbatch
+
+  variables:
+    # global parameters
+    - match: '(%)((~([fdpnxsatz]|\$PATH:)*)?\d|\*)'
+      scope: variable.parameter.dosbatch
```
Briles: I'm somewhat torn on this scope; `variable.parameter` is typically used for functions while these are the arguments passed to the script.

michaelblyons:
This is great! I had just finished variable identification on a branch of my own, but I like your system variables broken out into their own section.

I see a shortcomings, though. Spaces are a legitimate part of variables. I have this line in the code I was working on, and I've seen it other places, too:

```bat
:: Zero-padded timestamp 
SET t=%time: =0% 
```

I used this YAML for variables, but the substring/substitution bits have wonky scope names.

```yaml
variables:
  - match: '%\w+(?:(:)(?:(.=.)|(~-\d+|~\d+,\d+)))?%' 
    captures: 
      0: variable.other.percent-delimited.dosbatch 
      1: punctuation.separator.dosbatch 
      2: keyword.operator.substitution.dosbatch 
      3: keyword.operator.substring.dosbatch 
```

I think it could be adapted to add `!`-delimited variables and tag onto the system variables you set up.

Briles:
I added a far more robust implementation of variable matching than my original solution. Spaces, nested variables, string substitution, and substrings are all supported.

As a side effect, escape characters and numerics were also added.

Batch File/Batch File.sublime-syntax
wbond: This seems reasonable to me

wbond:
Thanks for all of these improvements!

PR #357, opened 2016-05-04
[Batch File] Scope for Goto labels

michaelblyons:
Hopefully the scopes are okay. I picked them explicitly to get picked up by the "Goto Symbol..." feature.

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -9,6 +9,7 @@ scope: source.dosbatch
 contexts:
   main:
     - include: comments
+    - include: labels
     - include: strings
     - include: operators
     - match: \b(?i)(?:append|assoc|at|attrib|break|cacls|cd|chcp|chdir|chkdsk|chkntfs|cls|cmd|color|comp|compact|convert|copy|date|del|dir|diskcomp|diskcopy|doskey|echo|endlocal|erase|fc|find|findstr|format|ftype|graftabl|help|keyb|label|md|mkdir|mode|more|move|path|pause|popd|print|prompt|pushd|rd|recover|ren|rename|replace|restore|rmdir|set|setlocal|shift|sort|start|subst|time|title|tree|type|ver|verify|vol|xcopy)\b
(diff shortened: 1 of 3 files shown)


Batch File/Batch File.sublime-syntax
```diff
@@ -34,6 +35,12 @@ contexts:
     - match: '<&?|>[&>]?'
       scope: keyword.operator.redirection.dosbatch
 
+  labels:
+    - match: '(?i):(\w+)\r?\n'
+      captures:
+        0: meta.function.dosbatch
+        1: entity.name.function.dosbatch
```
wbond: For consistency with other syntaxes, this should be:

```yaml
  labels:
    - match: '^\s*(:)([\w ]+)$
      captures:
        1: punctuation.separator.dosbatch
        2: entity.name.label.dosbatch
```

A meta scope shouldn't really be necessary since it is such a simple construct.

According to https://technet.microsoft.com/en-us/library/bb490914.aspx, a label can contain spaces and must be on a line by itself.

A `.tmPreferences` file should probably also be added to show `entity.name.label` in the symbol list.

Briles: Spaces and `+=,;:` are not allowed to be the first character in a label after the `:` but they can be using thereafter. Labels can also contain special characters (even non-ascii).

Should we differentiate between label definitions & label calls?

```batchfile
goto :label
call :label

:label
```

michaelblyons:
Updated capture, scopes, test, and `.tmPreferences`.

Batch File/Symbol List.tmPreferences
```diff
@@ -0,0 +1,17 @@
...
+    <string>Symbol List</string>
+    <key>scope</key>
+    <string>source.dosbatch entity.name.label.dosbatch</string>
+    <key>settings</key>
+    <dict>
+        <key>showInSymbolList</key>
+        <integer>1</integer>
+        <key>symbolTransformation</key>
+        <string>
+        </string>
```
wbond: These three lines can be removed since no transformation is being performed.

wbond:
The only other thing is that we use tabs for indentation of `.tmPreferences` files.

michaelblyons:
Spaces changed to tabs.

wbond:
Thanks!

PR #359, opened 2016-05-05
[Batch] Fix comments

Briles:
Fixes
- runaway scoping when REM is immediately followed by a newline
- Incorrect illegal characters within REM (the [docs](https://technet.microsoft.com/en-us/library/bb490986.aspx) say `(|)` are not allowed but the tests I did say otherwise)
- Allow : and a space to start a line comment

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -35,17 +35,16 @@ contexts:
       scope: keyword.operator.redirection.dosbatch
 
   comments:
-    - match: '(:[+=,;:])'
+    - match: '(:[+=,;: ])'
       scope: punctuation.definition.comment.dosbatch
       push:
         - meta_scope: comment.line.colon.dosbatch
         - match: \n
           pop: true
-      # REM command
-      # https://technet.microsoft.com/en-us/library/bb490986.aspx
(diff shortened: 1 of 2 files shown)


wbond:
The issue with `|`, `(` and `)` as I understand it is they are redirection operators, and `rem` is a command, so you can inadvertently execute code if they are included.

Briles:
It seems that was only a concern in earlier versions of cmd/dos. [This page](http://www.lagmonster.org/docs/DOS7/y-rem.html) from 1999 is the only other place I could find which expresses a concern for redirection characters within `rem`. Any idea how many people are using ST to write batch scripts for use on Windows 95? :laughing:

keith-hall:
Does ST even work on Win 95? ;)

wbond:
Well, the [official docs](https://technet.microsoft.com/en-us/library/bb490986.aspx) that were linked in the PR said `|`should not be used inside of `REM`. I think that is a perfect good reason to not allow them.

keith-hall:
Seems safer not to allow them to prevent bad coding habits. Who knows if DOS emulators etc. behave the same way? (DOSBox comes to mind for example) ;)

Briles:
The changes to illegal characters have been reverted.

PR #370, opened 2016-05-09
[LaTeX] rewrite syntaxes

### LINKED ISSUE #340: Rewrite the LaTeX syntax
The default LaTeX syntax is quite problematic, for examples
- #73, #300 
- https://github.com/jonschlinkert/sublime-markdown-extended/issues/119, 
- doesn't highlight `\newcommand{\bg}{\begin{equation}}` correctly.
- highlight `\text{}`and `\mbox{}` as string within math mode (changing color scheme does not fix this)
screen shot 2016-04-30 at 3 22 02 pm

There are 3 latex related packages now,
- https://github.com/SublimeText/LaTeXTools
- https://github.com/LaTeXing/LaTeXing
- https://github.com/randy3k/LaTeX-Plus

LaTeXTools uses the default LaTeX syntax and LaTeXing defines a wrapper syntax `LaTeXing` which imports the default LaTeX syntax. They both suffer both the same problems that the default LaTeX has.
Given the problems, I have created `LaTeX+` syntax from scratch in the package [LaTeX-Plus](https://github.com/randy3k/LaTeX-Plus/tree/master/syntax).

`LaTeX+` basically contains all features of the default `LaTeX` with the following two important changes.

1. It avoids the usage of $top_level_main, as indicted in #73, $top_level_main breaks a lot of things.

2. LaTeX math environments are highlighted as `string` scope, I found that it is a bad idea since it causes the problems of highlighting `\text{}`and `\mbox{}` as stated above.
In `LaTeX+`, the scope of math environment is defined as `meta.function.environment.math.latex` instead. Because of this change,  `LaTeX+` syntax is not compatible with `LaTeXTools` and `LaTeXing` commands which use scope `string.other.math.tex`. But the fix is easy, they just need to change the commands' scope from `string.other.math.tex` to `string.other.math.tex, meta.function.environment.math.latex`. This will ensure that the commands are compatible with `LaTeX+` and the default `LaTeX`.

To rewrite the LaTeX syntax, I suggest that we could use `LaTeX+` as a skeleton. @csch0 @msiniscalchi @ig0774 (the developers of LaTeXtools and LaTeXing) Do you have any comments?

randy3k:
close #340 

The follow up PR for #340, things to be done
- [x] syntax_test files
- [x] ~~~handle accents~~~ (It seems it is not necessary)
- [x] math scope naming (no objection?)
- [x] itemize/enumerate etc scope naming (no objection?)
- [x] remove incompatible regex patterns
- [x] fix `R/Rd (R Documentation)` and `Haskell/Literate Haskell`
- [x] cite commands
- [x] POSIX regex
- [x] minted package

There are many changes, I just list the important breaking changes
- math mode: `string.other.math.tex` -> `meta.environment.math`
- itemize/enumerate/description/list scopes are now
   - `meta.environment.list.itemize.latex`
   - `meta.environment.list.enumerate.latex`
   - `meta.environment.list.description.latex`
   - `meta.environment.list.list.latex`
- Added symbol lists for labels and sections
- The syntaxes for Beamer and Memoir were removed (I think no one is actually using these syntax files).
External packages' support can be added to the `packages` pattern of `LaTeX` syntax.

@csch0 @msiniscalchi @ig0774 Do you have any comments? This PR will certainly break your packages. We might have to discuss the naming convention of the scopes.


wbond:
I added a list item to remove incompatible regex patterns

wbond:
This will also affect the `R/Rd (R Documentation).sublime-syntax` and `Haskell/Literate Haskell.sublime-syntax` files.

randy3k:
Will fix them too.

r-stein:
Just a few points:

- It would be nice to add `\ensuremath` to the math commands
- I think it would be useful to add a meta scope for environments in which the tex commands are treated verbatim, e.g. lstlisting or verbatim. A usecase would be to define keybindings like:
```js
  {
    "keys": ["\""], "command": "insert_snippet", "args": {"contents": "``$0''"},  "context": [
      { "key": "selector", "operator": "equal", "operand": "text.tex.latex - (meta.environment.verbatim.latex, meta.environment.math.latex)" }
    ]
  }
```
- are you planing to support more packages (e.g. minted, comment,...) and additional listing languages?

randy3k:
- `\ensuremath` sounds like a good idea.

- verbatim now receives the scope `meta.function.verbatim.latex`, but to be consistent, I guess it should be renamed as `meta.environment.verbatim.latex`.

- the plan is to add packages which are commonly used. And adding such packages should help in highlighting or defining keybinds.

ig0774:
One regression I've noticed is that this syntax doesn't properly handle commands in optional arguments where there is a specific rule for the main command. For example, in a command like:

```latex
\makebox[\linewidth]{...}
```

`\linewidth` isn't properly recognised as a command.

randy3k:
The issue related optional arguments is new to this syntax as the original syntax doesn't detect optional arguments at all (at least for `\makebox`). I have an idea on how to fix it, but it will take some time since I want to do it more generally.

buildersbrewery:
Randy, Will has been using <http://www.regular-expressions.info/posixbrackets.html> in the latest re-writes for sublime syntaxes. For consistency please replace the parts where they apply.

r-stein:
General ST question: is it possible (in the colorscheme or elsewhere) to present a scope in super- or subscript position? It would be nice to show something like `$x^2$` as `$x^²$`.

FichteFoll:
No.
Only regular, bold and italics are supported. Textmate also supports underline, but Sublime Text does not render this.

r-stein:
Okay thanks. I personally think this and named foldings like `\alpha` to `α` and `\label{...}` to `[l]` increases the readability of tex files.

randy3k:
@r-stein

I would let other packages to handle it. Some users may not like this idea.

r-stein:
@randy3k As far ich know this is not possible in ST, it was just a general comment :)

jrappen:
> [`@FichteFoll`](https://github.com/fichtefoll): Only regular, bold and italics are supported. Textmate also supports underline, but Sublime Text does not render this.

I believe it's because underline is reserved for the spell check functionality.

randy3k:
Sorry for the delay, I should have some time this weekend to work on it.

FichteFoll:
@jrappen 
> I believe it's because underline is reserved for the spell check functionality.

Actually, those are "squiggly underlines" (via [color schemes](http://docs.sublimetext.info/en/latest/reference/color_schemes.html#brackets)). Underlines are still a thing for bracket/tag matching and the `View.add_regions` API though.

buildersbrewery:
@wbond add the `significant` label, please.

randy3k:
Just want to get an idea what languages should be supported for the `minted` package.
Also, should we keep `1stlisting`? It seems that `minted` is more popular (100k vs 26k search results via google).

ig0774:
I think `listings` (the package that defines `lstlistings`) may actually be more popular. It has 450k Google results and 2k questions on tex.stackexchange.com versus 350 for `minted`, which makes some sense since `listings` is pure LaTeX while `minted` depends on Python and Pygments. That said, I rarely typeset code... Aside from Python and Java, it may make sense to at least support LaTeX itself (if possible).

r-stein:
I think both packages are popular and therefore it would be nice to support both. Especially because I think, that the ability to highlight embedded languages is one of the strengths of Sublime over most other LaTeX editors.
Also for the supported languages I think it would be better to support more, than less. Nobody will say "It highlights to many languages". Disadvantages would be a big syntax file and I don't know about possible performance issues.
For the languages it would be good to support the top used languages and languages, which are typically used by LaTeX users/for LaTeX documents (SQL, Haskell, LISP, R,...).

r-stein:
Remark: `[:alpha:]` is different from `A-Za-z`. In python you can add the flag `re.UNICODE` and ST seems to do something similar. Hence `[:alpha:]` does also capture non-ascii word characters like ä, ö, ü... Therefore we should better use `A-Za-z` in most cases.

Example:

| `'(\\)[[:alpha:]@]+'` | `'(\\)[A-Za-z@]+'` |
|--- |--- |
| alpha | A-Za-z |

randy3k:
The rewrite is basically done except the extra language support for `minted` and `listings`. Please review it.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -4,554 +4,907 @@
...
+    - include: boxes
+    - include: scope:text.tex#math-commands
+    - include: scope:text.tex#math-operators
+    - include: scope:text.tex#math-characters
+    - include: scope:text.tex#math-numerics
+    - include: general-constants
+
+  inline-math:
+    - match: \$
+      scope: string.other.math.latex punctuation.definition.string.begin.latex
```
r-stein: Might be better to not use `string` for the math delimiters.

randy3k: The reason was just to retain the original highlight color (from the color scheme) of the dollar signs.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -4,554 +4,907 @@
...
+      push:
+        - meta_scope: meta.environment.math.latex
```
r-stein: I think it would be nice to add an other meta_scope specially for the different types. This could be useful for special keybindings and the scope is used for [BracketHighlighter rules](https://github.com/facelessuser/BracketHighlighter/blob/st3-2.14.2/bh_core.sublime-settings#L232).

randy3k: check 87a0d0b
👍

r-stein: I think its better, but we should also differ between `\(...\)` and `$...$`. The scopename could be similar to the convention of comments, i.e. `meta.environment.math.inline.dollar.latex`, `meta.environment.math.inline.paren.latex`, `meta.environment.math.block.brace.latex`, or something with `tex.latex` and `latex.latex` (for the commands, which come from tex and latex).

LaTeX/LaTeX.sublime-syntax
```diff
@@ -446,7 +446,7 @@ contexts:
     - match: \$
       scope: string.other.math.latex punctuation.definition.string.begin.latex
       push:
-        - meta_scope: meta.environment.math.latex
+        - meta_scope: meta.environment.math.latex meta.environment.math.inline.latex
```
FichteFoll: You don't need to provide both, `meta.environment.math.latex` and `meta.environment.math.block.latex`. The more specific one should be enough for 99.9% of all cases.

wbond:
Hey @csch0 and @msiniscalchi. This PR has done a bunch of work on the LaTeX package and syntaxes. My understanding is that it is likely to affect you and your packages.

Mind spending some time reviewing and providing feedback?

I'm no expert in LaTeX, or how it may affect you. If I don't hear anything back, I'm just going to have to assume it should be fine to merge in. Hopefully this can be a solid base to make further improvements or fixes to.

randy3k:
The two most critical changes are the scopes for math and list environments. They are now
`text.tex.latex meta.environment.math` and `text.tex.latex meta.environment.list` respectively. 

More specific meta scopes of math env and list are also provided. Here are some of the examples.

math
- `$....$`: `meta.environment.math.inline.dollar.latex`
- `$$....$$`: `meta.environment.math.block.dollar.latex`
- `\begin{equation}....\end{equation}`: `meta.environment.math.block.be.latex`

list
- `meta.environment.list.itemize.latex`
- `meta.environment.list.enumerate.latex`
- `meta.environment.list.description.latex`

wbond:
One high level question from just scanning the changes:

I noticed that `meta.group.braces` and `punctuation.definition.braces.begin` is used. We've been doing some work to try and standardize things like this to make it easier for color schemes to highlight such things. Is there a common semantic meaning to braces in LaTeX? Would a scope such as `meta.block` and `punctuation.definition.block.begin` make any sense? (That is what is used in C-style languages, I am just trying to figure out if that makes sense here).

Similarly we've been using `meta.group` and `punctuation.definition.group.begin`/`end` for `()`. Would such a "generic" scope make sense anywhere in this syntax, or is having more-specific scopes useful for the various packages?

randy3k:
The meaning of braces in LaTeX is like the parenthesis in C or python. I think it was why `meta.group.braces` was used in the original .tmlangauge.
Perhaps, we might use `meta.group.braces` and  `punctuation.definition.group.braces.begin`/`end`?

We might also need to define  `meta.group.brackets` and  `punctuation.definition.group.brackets.begin`/`end` for the `[` `]` pair.

By the way, should it be `meta.group.braces` or `meta.group.brace`?

randy3k:
Now, `meta.group.bracket` and `punctuation.definition.group.bracket.begin`/`end` are used for `[``]` pair and 
`meta.group.brace` and `punctuation.definition.group.brace.begin`/`end` are used for `[``]` pair.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -4,554 +4,1554 @@
...
+    - match: '(?:\s*)((\\)(?:url|href))(\{)([^}]*)(\})'
+      scope: meta.function.link.url.latex
+      captures:
+        1: support.function.url.latex
+        2: punctuation.definition.backslash.latex
+        3: punctuation.definition.group.brace.begin.latex
+        4: markup.underline.link.latex
+        5: punctuation.definition.group.brace.end.latex
+
+  invalid-quotations:
```
r-stein: This also highlights valid quotations like ``"`string"'`` to invalid. This supported by the [german babel package](https://en.wikibooks.org/wiki/LaTeX/Internationalization#German)

randy3k: Perhaps, we should remove the whole section invalid quotations? It seems there are more bizzard exceptions more complicated that ``` "` ``` and `"'`.
https://en.wikibooks.org/wiki/LaTeX/Internationalization#Cyrillic_script

r-stein: Yes, I don't think we can capture all valid quotations.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -4,554 +4,1545 @@
...
-        3: punctuation.definition.arguments.latex
+        2: punctuation.definition.backslash.latex
+        3: punctuation.definition.group.brace.begin.latex
         4: variable.parameter.function.latex
-        5: punctuation.definition.arguments.latex
+        5: punctuation.definition.group.brace.end.latex
       push:
-        - meta_scope: meta.function.environment.general.latex
-        - match: '((\\)end)(\{)(\4)(\})(?:\s*\n)?'
+        - meta_scope: meta.environment.math.block.be.latex
```
r-stein: I would change this to `meta_content_scope` otherwise the syntax highlighting also changes the braces of \begin{...} and \end{...}. In addition all math-completions should also only applied in the content.

LaTeX/TeX.sublime-syntax
```diff
@@ -7,73 +7,181 @@ file_extensions:
...
       scope: comment.line.percentage.tex
+
+  controls:
+    - match: ((\\)(else|fi|ftrue|if(case|cat|dim|eof|false|hbox|hmode|inner|mmode|num|odd|undefined|vbox|vmode|void|x)?))\b
       captures:
-        1: punctuation.definition.comment.tex
-    - match: '\{'
+        1: keyword.control.tex
+        2: punctuation.definition.backslash.tex
+    - match: '((\\)(?:input))'
```
r-stein: This also matches `\inputminted`, we might need to check that it is followed by a brace.

randy3k:
@r-stein 
 I don't have much time to work on this recently. I have added you as a collaborator, please do whatever reasonable to the PR.

FichteFoll:
I just tested this for the first time (because I was getting annoyed by quotation marks being highlighted as invalid in verbatim environments; thanks for the embedded highlighting!) and noticed that highlighting for `\cite` as well as `\ref` (and `\autoref`) is lost. Was that a conscious decision? If so, what was the reason?

As a suggestion, highlighting options like the optional argument to `\usepackage` would be nice.

Another suggestion, highlighting `\def`, `\newcommand` and `\renewcommand` as something like `storage.modifier` would also be nice, with the to-be-defined command as `entity.name.command`.

FichteFoll:
**Warning! Embedded code using `include` inside pushes, not `with_prototype`!**

Instead of 
```yaml
        - match: '(\{)(python|py)(\})'
          captures:
            1: punctuation.definition.group.brace.begin.latex
            2: variable.parameter.function.latex
            3: punctuation.definition.group.brace.end.latex
          push:
            - meta_content_scope: meta.environment.embedded.python.latex source.python.embedded
            - match: '(?=\\end\{minted\})'
              pop: true
            - include: scope:source.python
```
you must be using
```yaml
        - match: '(\{)(python|py)(\})'
          captures:
            1: punctuation.definition.group.brace.begin.latex
            2: variable.parameter.function.latex
            3: punctuation.definition.group.brace.end.latex
          push:
            - meta_content_scope: meta.environment.embedded.python.latex source.python.embedded
            - include: scope:source.python
          with_prototype:
            - match: '(?=\\end\{minted\})'
              pop: true
```

otherwise syntaxes with context-mangling might never close!

2016-06-29_21-26-04_sublime_text

randy3k:
@FichteFoll 
The loss of highlighting of \cite etc was not intended. The \cite and \ref use `constant.other` and \label uses `variable.parameter` for highlighting. I guess we should add those back.

You are right. We should use `with_prototype` to avoid any potential syntax errors. (But I don't have any highlight issue for your particular example...)

randy3k:
I have some difficulties in applying the `with_prototype` match..

```bash
Error loading syntax file "Packages/LaTeX/LaTeX.sublime-syntax": Apparent recursion within a with_prototype action: 25000 context sanity limit hit
```

Errors occurs so randomly so that I couldn't find a way to reproduce it. I used to have issue with embedding `source.php`, but strangely after some reboots, now it works with `source.php` but not `source.shell`...

wbond:
@randy3k `with_prototype` is hard to use for complex structures. It is very easy to create an include loop that recursively creates infinite versions of a context. It is also pretty easy to create context recursion because various syntaxes allow embedding other syntaxes in heredoc/multiline string contexts.

LaTeX/LaTeX.sublime-syntax
randy3k: I don't see why the braces of \begin{...} and \end{...} are affected. I guess the syntax highlight should not capture `meta.environment.math.xxx`, this scope should be used only in defining keybinds.

r-stein: I think it would be nice to highlight all math content in a specific color. But nonetheless also math keybindings should only be applied in the body not inside `\begin{...}` and `\end{...}`.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -523,7 +544,12 @@ contexts:
           pop: true
         - include: math-content
 
-    - match: '((\\)begin)(\{)\s*((?:align|alignat|aligned|alignedat|displaymath|eqnarray|equation|flalign|gather|gathered|math|multline|x?xalignat|split)\*?)\s*(\})'
+    - match: |-
+        ((\\)begin)(\{)\s*((?:
```
FichteFoll: You don't have extended mode enabled in this regex.

FichteFoll:
Thanks for these adjustments.

For `\usepackage` (`\documenttype`), it would be nice if you didn't highlight the option values (i.e. split on `=`) and also did it multi-line like here:
```
\usepackage[pdftex,
            %plainpages={false},
            bookmarks=true,
            unicode=true,
            bookmarksnumbered={true},
            % pagebackref=true,
            breaklinks=true,
            pdfstartview={FitBH}]{hyperref}
```

LaTeX/LaTeX.sublime-syntax
randy3k: Thank you for catching that.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -545,6 +545,7 @@ contexts:
         - include: math-content
 
     - match: |-
+        (:x)
```
FichteFoll: yeah, no :smile:

randy3k: There's a mistake again...fixed now.

LaTeX/LaTeX.sublime-syntax
randy3k: I think it is not a very big deal for now. I hesitate  to change because I don't know what you should do with `meta.environment.list`. For consistency, it seems that it should be
```
- meta_content_scope: meta.environment.list.itemize.latex
```

randy3k:
@FichteFoll's suggestion of using `with_prototype` somehow creates infinite loops, I guess it won't be fixed for now. However, I would like to understand why your python example was not highlighted correctly. Do you still experience the same bug at the latest commit?

FichteFoll:
It's likely I was working with an uncommitted and modified version of the Python syntax def. I also couldn't explain why highlighting was bogus. I'll look into it again once I have more time and start working on the Python syntax again, most likely mid-late August.

I agree that, with the way `with_prototype` works, it's unfortunately not feasable to use for inclusion of many (or specific) languages because of its tendency to create infinite compile loops.

FichteFoll:
It would be nice if `~`, the non-breaking space, would be highlighted.

FichteFoll:
Thanks.

I got another update regarding scoping of commands. In the old syntax, commands related to referencing were scoped as `keyword.control.<cmd-name>` (e.g. `\label` and `\cite`).

I liked that separate coloring since those command make connections across the entire document, which is why I want to suggest using `storage.type.label` for `\label` and `keyword.control.reference` or `keyword.other.reference/cite` for `\cite` and `\ref` respectively.

In a similar thought, the `\begin` and `\end` commands are very important for the document flow, which is why I suggest scoping these as `keyword.control.flow.begin` and `keyword.control.flow.end`.

randy3k:
Done.

buildersbrewery:
@randy3k @r-stein @FichteFoll thank you very much for this

FichteFoll:
There are still a few environments without the `keyword.control` treatment, notably `\begin{itemize}` and `begin{equation}`, align and most likely more. I'm not sure why those are matched separately in the first place.

r-stein:
@FichteFoll I think it is useful for special keybindings and snippets. I personally use `ctrl+enter` for "insert `\item $0` a line below" inside itemize/enumerate/list and `\item[$1] $0` inside description.

randy3k:
@FichteFoll Thanks for the testing. I agree with @r-stein.

Eyenseo:
@randy3k There is a problem with `\listings${$ Foo`. The second `$` is not recognized as the curly bracket is not balanced. This leads to the following text to be highlighted falsely until another 'math' environment closes that bracket.

foo

As I said in my pr (#300) I'm unsure if user of listings have to live with that - I myself am using a custom command (with options for the different languages I want to typeset) so it is not possible to rely on the `listings` name.

@buildersbrewery Thank you for letting me know! I will close my pull request as this is way better even if this 'issue' will not be fixed by it.

wbond:
This seems like it has been baking for a while. I'd love to get it merged in, and then further tweaks can be made as other bugs or possible enhancements are found.

I, personally, haven't reviewed it, partially because I currently know nothing of LaTeX, but also since I've been focusing on some other aspects of Sublime Text recently.

For all of the people who have, would you mind adding a 👍 reaction to this comment?

randy3k:
@Eyenseo
Is `\listings` a valid command? `\lstinline${$` is supported anyway. (I don't use listings personally)

@wbond 
I would love to have it merged.

wbond:
The only other thought I just had is: has anyone tested the symbol list to ensure that still functions as one would expect?

Eyenseo:
@randy3k `listings` could be a custom command I set up - so yes, it is valid. The official command is indeed `lstinline`.

randy3k:
@wbond This PR also contains changes to  `Symbol List*.tmPreferences` files. The symbol list should work for the current file and projectwisely.

randy3k:
My question should really be "is \listings official?". As it is not official, I don't think we should include it.

r-stein:
Currently `minted` supports more languages than `listings`, I think we should add the languages to `listings` as well.

randy3k:
It was because I have checked the supports of `listings` and `minted` and excluded some of the languages which are not supported by listings. Page 13 of http://tug.ctan.org/macros/latex2e/contrib/listings/listings.pdf.

Eyenseo:
@randy3k I do disagree with your reasoning - only supporting the "official" packages (I would prefer a complete/perfect solution). But as I said before it might be impossible to support this case and should not hinder this pull request.

If this won't be fixed before merging, I would open an issue for future reference as other people might run into the same issue.

In my case it is `\newcommand{\mySIn}{\lstinline[language=MyString]}`, `\newcommand{\myRIn}{\lstinline[language=MyRegex]}` and `\newcommand{\myMIn}{\lstinline[style=MyMacroStyle,postbreak=]}`.

Additionally the parameter to `\label` should probably be the same color as `\autoref`?

randy3k:
@Eyenseo While I agree with you that it is very useful to highlight custom defined command, I also want to tell you that it is not possible to define highlight patterns in run time (actually, it may be possible with some adhoc engineering). That's why only official commands should be considered here.

The highlight difference is because `\label` is defining the label and `\ref` is calling the label. If you do `Go To Symbol`, all the labels will be listed there.

Eyenseo:
@randy3k If you say that it is impossible with the current options I believe you. 

My reasoning is that `\label` defines a _label_ and `\ref` references a _label_ and I thought the color (yellow in my case) is used for labels. For me it's counter intuitive that `\label` has a different highlighting.

wbond:
I haven't looked at the current implementation, but theoretically you can extend the syntax with a new one that adds features. That depends on details of how the scope are implemented.

Eyenseo:
@wbond Is that on a per project base or per sublime text settings base? The former would be close enough for me.

wbond:
@Eyenseo This would be a new syntax based on LaTeX with a different name.

Eyenseo:
@wbond so per sublime setting base, better than nothing I guess ;) 
Thank you!

wbond:
@Eyenseo Yes, syntaxes are global to Sublime Text, and are not per-project.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -552,7 +552,7 @@ contexts:
         4: variable.parameter.function.latex
         5: punctuation.definition.group.brace.end.latex
       push:
-        - meta_scope: meta.environment.math.block.be.latex
+        - meta_content_scope: meta.environment.math.block.be.latex
```
randy3k: For consistency, shall this also be applied to `\[\]` and `\(\)` and `$$` and `$` pairs?
Though, it may break BracketHighlighter or make BracketHighlighter hard to highlight brackets.

r-stein: I would not apply this to those pairs, because it would break bracket highlighter and I don't see disadvantages for it.
Since it should have the same or a special highlighting and it has only 1-2 chars and is therefore to short to use hotkeys inside. However I agree, that it might look less consistent.
Nonetheless I think this is the best solution to do it like this.

FichteFoll:
I've been using this syntax for a while now and reported everything I noticed so far. This is obviously a great improvement over the current LaTeX syntax but people will need to adapt to changes since things are highlighted differently now (e.g. `entity.name.label` for `\label` definitions), but when JavaScript was changed they had to as well and that affected a lot more people.

The only remaining thing is inline Python highlighting which is seemingly broken "just for me". Since I did work on the syntax that relates to the resulting highlighting, I *should* be able to know why it behaves that way, but I don't just from thinking about it and I lack the time to actually investigate into it. I have a couple more changes planned for the Python syntax and will take a look at this again once I find time for that.

So far, this is getting a big :+1: from me.

ig0774:
I agree this is a major improvement over the previous syntax.

One small regression I've noticed is that commands in optional brackets after `\cite` commands don't get highlighted properly. (The same thing happens with `\ref`, but I don't think any of the major variants of `\ref` allow optional arguments anyways).

screen shot 2016-07-16 at 15 17 15

Not really a big a deal though.

randy3k:
@ig0774 fixed.

r-stein:
It might be worth to discuss to omit the scope `punctuation.definition.backslash.latex` for each `\` in the commands (`\command`), because ST seems to change the soft wrap behavior.
See also https://forum.sublimetext.com/t/word-wrapping-in-latex-files-st-st2/1403 and https://github.com/SublimeText/LaTeXTools/issues/643

randy3k:
Firstly, just want to clarify, it does not seem that I have the splitting issue.

Secondly, I would rather keep the scope for potential different highlight via color scheme. Some users may be interested in highlighting backslashes with a different color.

ig0774:
It still occurs for me with the latest version of the syntax:

screen shot

Frankly, however, I'm indifferent to what is done. For me, syntax highlighting still makes it clear its a command when the wrapping occurs, and I can see some small benefit to being able to highlight backslashes.

randy3k:
I see, I could reproduce it now. My suggestion is to not do soft wrap, do [hard wrap](https://github.com/randy3k/AutoWrap) instead. Hard wrap also improves the results in reverse synctex search from the pdf.

r-stein:
Small remark: math is currently not highlighted inside `\newcommand`:

example_math

randy3k:
@r-stein it is intended. The content insider the bracket can create all sort of issues.

Consider

```latex
\newcommand{\foo}{\begin{equation}}
\newcommand{\bla}{\[}
```

If we start to match the content, there will be difficult to close the scope, so only commands and constants are highlighted inside the bracket. This is actually the whole reason I started this re-write.

r-stein:
Ah okay I agree.

wbond:
Thank you for all of your work on this @randy3k.

And thank you to everyone who has been reviewing it and giving feedback.

musm:
I tried testing this out. I placed the LaTeX directory obtained from github to %appdata%/Sublime Text 3/Packages

Firing a tex document gives:
image

PR #372, opened 2016-05-10
[IDL] Add support for the Interactive Data Language (IDL)

cschreib:
**Not ready for merge.** 
I've created the pull request to gather your comments, and also to check if this is a wanted feature.

IDL is an old language that is (unfortunately) still widely used in data science, including astronomy, astrophysics and earth science.
More info on Wikipedia: https://en.wikipedia.org/wiki/IDL_%28programming_language%29

There is a unofficial package that offers support for this language in Sublime, however it is based on the older .tmLanguage syntax, and it does not properly highlights all the scopes and variable types. I've therefore started a rewrite from scratch (using the Lua .sublime-syntax file as a skeleton).

I will report back when the parser reaches my expectations. In the mean time, I think I might need help understanding the "Indent" file. It should be easy to support 90% of the cases, since IDL exclusively uses keywords for flow control, i.e:
```IDL
; Multi instruction construct
if b eq 1 then begin
    a = 2
endif ; or just 'end'

; Single instruction construct
if b eq 1 then a = 2
```

However this can become messy very quickly, since IDL allows any instruction to be split into multiple lines by inserting the '$' character essentially anywhere:
```IDL
; Single instruction construct split into multiple lines
if b eq 1 then $
    a = 2
if b eq 1 $
    then a = 2
if $
    b eq 1 $
    then $
    a = 2
```

This issue (plus the fact that multiple instructions can be combined on the same line with the "&" character) also make writing the parser quite difficult. Is there a standard and easy way to handle this?

### DIFF
--- a/IDL/Comments.tmPreferences
+++ b/IDL/Comments.tmPreferences
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>name</key>
+	<string>Comments</string>
+	<key>scope</key>
+	<string>source.idl</string>
+	<key>settings</key>
+	<dict>
+		<key>shellVariables</key>
+		<array>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_START</string>
+				<key>value</key>
+				<string>; </string>
+			</dict>
+		</array>
+	</dict>
+</dict>
+</plist>
--- a/IDL/Completion Rules.tmPreferences
+++ b/IDL/Completion Rules.tmPreferences
@@ -6,7 +6,7 @@
 	<key>settings</key>
 	<dict>
 		<key>cancelCompletion</key>
-		<string>^.*\b(then|end|do|else)$</string>
+		<string>^.*\b(then|do|else|for|while|case|switch|begin|end|function)$</string>
 	</dict>
 </dict>
 </plist>
--- a/IDL/Completion Rules.tmPreferences
+++ b/IDL/Completion Rules.tmPreferences
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>scope</key>
+	<string>source.idl</string>
+	<key>settings</key>
+	<dict>
+		<key>cancelCompletion</key>
+		<string>^.*\b(then|end|do|else)$</string>
+	</dict>
+</dict>
+</plist>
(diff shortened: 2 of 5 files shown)


wbond:
It seems to me that this is more of a niche language, and would be better served being hosted on Package Control. Currently there are hundreds of such syntaxes available for the "less-mainstream" languages. Supporting all of these as default packages would probably require the full-time work of multiple people.

With Package Control, you can target the version of Sublime Text that the package works with, so it would only show up to users running recent builds of ST3. Another thought would be to work on getting it into a good place and then contact the author of the other IDL syntax and see if they may let you take the name over.

I have not yet looked at the Lua syntax since I started working on maintaining these syntaxes, so I'm not sure what state it is in. In may require some cleanup. Generally:

 1. we are moving away from any constructs that aren't compatible with the new regex engine. You can use the Syntax Tests build system with the Regex Compatibility variant to check.
 2. The construct `captures: 0:` should be replaced by `scope:`.
 3. Using variables for common snippets of regex helps simplify maintenance

Unfortunately I haven't yet gotten to the point of working on official documentation about what scope names we are standardizing on for the default packages. This has to do with the fact that I am still going through them and ensuring that I'm familiar with all of the different constructs we need to handle, before I try to come up with a declaration of how things "should" be done.

In regards to your question about indenting, `decreaseIndentPattern` and `increaseIndentPattern` can get very messy, very quickly. Currently there are outstanding bugs with some of the major default syntaxes related to this. I've been trying to focus my effort towards improvement and fixes where the impact will be biggest, so I haven't addressed some of the edge cases yet. One thing that you can try to use in your favor is that the `decreaseIndentPattern` and `increaseIndentPattern` can be specified per scope. Most languages just use `source.idl`, however it may be possible to use meta scopes to further targets indent rules based on being in the middle of a certain type of construct. Additionally, you can use lookbehind assertions to check and see what characters are just before the linebreak.

cschreib:
Thank you for the tips, I'll try to follow the latest practice as best as I can. I'm also fine with releasing it as a separate package if you feel it doesn't belong to the default language set. Otherwise I can commit to maintain this specific language in the future if you have too much on your shoulders already.

cschreib:
I've finished the parser, following the recommendations you gave me. I managed to get rid of all the lookbehinds so the regex are compatible with the new engine. I also realized that no one had yet proposed a package in Package Control for this language; I suppose the one I have used so far was simply extracted directly from TextMate (http://wiki.macromates.com/Bundles/IDL).

So I'll create new package for now (https://github.com/wbond/package_control_channel/pull/5506). If ever you are interested in including IDL in the default supported languages, feel free to contact me or just copy the files. Thanks for your help.

PR #376, opened 2016-05-10
[Diff] Make scopes consistent

Briles:
I wasn't sure if placing `.diff` in the middle of scopes was intentional; regardless, it is inconsistent with the other syntaxes.

### DIFF
--- a/Diff/Diff.sublime-syntax
+++ b/Diff/Diff.sublime-syntax
@@ -24,28 +24,28 @@ contexts:
       captures:
         1: punctuation.definition.separator.diff
     - match: ^\d+(,\d+)*(a|d|c)\d+(,\d+)*$\n?
-      scope: meta.diff.range.normal
+      scope: meta.range.normal.diff
     - match: ^(@@)\s*(.+?)\s*(@@)($\n?)?
-      scope: meta.diff.range.unified
+      scope: meta.range.unified.diff
       captures:
         1: punctuation.definition.range.diff
         2: meta.toc-list.line-number.diff
         3: punctuation.definition.range.diff
     - match: '^(((\-{3}) .+ (\-{4}))|((\*{3}) .+ (\*{4})))$\n?'
-      scope: meta.diff.range.context
+      scope: meta.range.context.diff
       captures:
         3: punctuation.definition.range.diff
         4: punctuation.definition.range.diff
         6: punctuation.definition.range.diff
         7: punctuation.definition.range.diff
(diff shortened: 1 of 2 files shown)


wbond:
I agree this will be more consistent, but is sure to break some color schemes.

I should probably work on some migrations docs for the next beta release so color scheme authors can start moving towards the scopes we are standardizing on.

jrappen:
https://www.sublimetext.com/docs/3/scope_naming.html

jrappen:
any updates here?

jrappen:
@wbond This PR seems to follow the naming of scopes as described in the docs, as far as I can tell.

wbond:
@jrappen Yes, but that is not the issue. The issue is all of the color schemes that have created styles for these "backwards" scope names. If we "fix" this, many users will no longer have the diff highlighting they are used to.

jrappen:
You won't get all of the theme / scheme creators to update their packages anyway. Might as well move forward with stuff like this?

Mentioning creators of top theme / scheme packages:

/cc FYI @equinusocio @kkga @thinkpixellab @idleberg @jamiewilson @ihodev @daylerees @kenwheeler @itsthatguy

ihodev:
:+1: :wink:

wbond:
@jrappen My plan was to make a tweak in Sublime Text that would automatically rewrite the scope names when loading a color scheme.

wbond:
After discussing my idea with Jon, he pointed out that if we want to keep backwards compatibility, it should happen in this syntax, not in core. I think that is a far better idea than I had.

So, either this PR should add the new scopes, but leave the existing ones for backwards compat, or we just don't make a change.

jrappen:
To get moving, I submitted pull requests to:

* [Tomorrow Color Schemes](https://packagecontrol.io/packages/Tomorrow%20Color%20Schemes) via theymaybecoders/sublime-tomorrow-theme#9
* [1337 Color Scheme](https://packagecontrol.io/packages/1337%20Color%20Scheme) via MarkMichos/1337-Scheme#9

with [`.sublime-color-scheme`](https://www.sublimetext.com/docs/3/color_schemes.html) files converted by [PackageDev](https://github.com/SublimeText/PackageDev). This should make things easier going forward.

jrappen:
@Briles I'll have a look at gruvbox later this week. That will one take a bit longer.

wbond:
So I think we should either close this, or agree to implement backwards compat scope names, e.g. `meta.diff.range.normal` becomes `meta.range.normal.diff meta.diff.range.normal`.

PR #377, opened 2016-05-11
[Batch] Add SET /A command expressions

Briles:
Adds support for operators when using the `SET` command with the `/A` flag


wbond:
Thanks for rebasing and working out the conflicts!

PR #378, opened 2016-05-11
Added Laravel Blade support

jbrooksuk:
- [ ] Fix comments not working
- [ ] Improve tests


jbrooksuk:
Looking for feedback etc.

FichteFoll:
https://github.com/sublimehq/Packages#new-packages

jbrooksuk:
@FichteFoll I've already spoken to @wbond about this, it's more feedback related currently, then we'll see where to go.

buildersbrewery:
@wbond add the `significant` label, please.

Laravel Blade/Laravel Blade.sublime-syntax
```diff
@@ -0,0 +1,95 @@
...
+        - match: '-?(\}?\}\})|(!!)\}'
+          captures:
+            0: entity.name.tag.block.any.html
+          pop: true
+        - include: scope:source.php
+
+  inline-php:
+    - match: <\?(?i:php|=)?
+      captures:
+        0: punctuation.section.embedded.begin.php
```
wbond: Instead of `captures: 0:`, you can just use `scope:`.

Laravel Blade/Laravel Blade.sublime-syntax
```diff
@@ -0,0 +1,95 @@
...
+      push:
+        - meta_scope: meta.embedded.line.php
+        - meta_content_scope: source.php
+        - match: (\?)>
+          captures:
+            0: punctuation.section.embedded.end.php
+            1: source.php
```
wbond: This probably shouldn't be here since the question mark isn't really part of the source code.

Laravel Blade/Laravel Blade.sublime-syntax
```diff
@@ -0,0 +1,95 @@
...
+          pop: true
+
+  function-parameters:
+    - match: '\)'
+      scope: punctuation.definition.group.end.php
+
+  statements:
+    - match: '\{(!!)|(\{\{\{?)-?'
+      captures:
+        0: entity.name.tag.block.any.html
```
wbond: This isn't the HTML syntax, so you shouldn't use a scope ending in `.html` here. Additionally, there isn't any name present, so using `entity.name` is unlikely what you want to do.

Laravel Blade/Laravel Blade.sublime-syntax
```diff
@@ -0,0 +1,95 @@
...
+    - match: '@\{\{\{?-?'
+      captures:
+        0: entity.name.tag.block.any.html
+      push:
+        - meta_scope: scope:text.html.basic
+        - meta_content_scope: scope:text.html.basic
+        - match: '-?\}?\}\}'
+          captures:
+            0: entity.name.tag.block.any.html
+            1: entity.name.tag.block.any.html
```
wbond: You've got duplicated scopes here.

wbond:
Your test files don't seem to be setup as real syntax tests. Right now they appear to be something you are just eyeballing during development. Check out any of the `syntax_test_*` files in this repo to see how you can write tests to help prevent regressions.

wbond:
A note to others looking at this PR, currently there is not a plan to merge this into the default set of packages. The only other language we would like to support by default in the near term is Swift.

wbond:
I'm going to close this now, just as a housecleaning measure to try and keep the PR list a todo list for myself.

jbrooksuk:
Thanks for going through this @wbond!

Laravel Blade/Laravel Blade.sublime-syntax
jbrooksuk: The `source.php` part shouldn't be?

wbond: Right, it you look at the PHP syntax, `<?=`, `<?` and `<?php` don't have `source.php` applied to them.

PR #382, opened 2016-05-12
[PHP] show class name in Goto Symbol

### LINKED ISSUE #380: Build 3112 breaks goto definition for PHP class names
Prior to build 3112 it was possible to place the cursor over the string `ClassA` in test.php and perform the `goto_definition` command to be taken to the file and class declaration in ClassA.php.

It has been suggested that recent changes to the PHP syntax could have affected this https://forum.sublimetext.com/t/dev-build-3112/19778/10.

test.php:
```php
<?php

$a = new ClassA;
```

ClassA.php:
```php
<?php

class ClassA
{
    //
}
```

keith-hall:
re: #380

### DIFF
--- a/PHP/Symbol List.tmPreferences
+++ b/PHP/Symbol List.tmPreferences
@@ -9,6 +9,8 @@
 	<dict>
 		<key>showInSymbolList</key>
 		<integer>1</integer>
+		<key>showInIndexedSymbolList</key>
+		<integer>1</integer>
 	</dict>
 </dict>
 </plist>
--- a/PHP/Symbol List.tmPreferences
+++ b/PHP/Symbol List.tmPreferences
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Symbols List: Indent Functions</string>
 	<key>scope</key>
-	<string>entity.name.function.php, support.function.magic.php, entity.name.class.php</string>
+	<string>entity.name.function.php, support.function.magic.php, entity.name.class.php, entity.name.interface.php, entity.name.trait.php</string>
 	<key>settings</key>
 	<dict>
 		<key>showInSymbolList</key>
--- a/PHP/Symbol List.tmPreferences
+++ b/PHP/Symbol List.tmPreferences
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Symbols List: Indent Functions</string>
 	<key>scope</key>
-	<string>entity.name.function.php, support.function.magic.php</string>
+	<string>entity.name.function.php, support.function.magic.php, entity.name.class.php</string>
 	<key>settings</key>
 	<dict>
 		<key>showInSymbolList</key>


keith-hall:
Although my fix gets the class to show up in Goto Symbol, it seems that there is something wrong with the ST3 window lookup_symbol APIs, such that Goto Definition still doesn't work.

Using the test case from https://github.com/sublimehq/Packages/issues/380#issuecomment-218650244:

    >>> view.symbols()
    [((13, 19), 'ClassA')]

    >>> window.lookup_symbol_in_open_files('ClassA')
    []

    >>> window.lookup_symbol_in_index('ClassA')
    []

keith-hall:
nevermind, it's not a bug - need to use `showInIndexedSymbolList`

mrliptontea:
This is great, `expand_fqcn` and `find_use` commands from [PHP Companion](https://packagecontrol.io/packages/PHP%20Companion) are working again.

I think it's worth adding `entity.name.interface.php, entity.name.trait.php` for full functionality for all symbols after `use` and `implements` keywords.

And a little off topic, but any idea how to make aliases function properly? E.g. if I have
`use Illuminate\Support\Collection as BaseCollection;`
goto symbol should take me to `Collection`, instead, it finds all other classes that use the same alias (or `BaseCollection` if there's a class of that name).

jbrooksuk:
I agree with @mrliptontea, having interfaces and traits working would be amazing.

keith-hall:
@mrliptontea @jbrooksuk I have added a commit to include interfaces and traits to the symbol list.

@mrliptontea regarding aliases, I wonder if it would be possible to use a symbol transformation regex in a `.tmPreferences` file to achieve this. The entire `Illuminate\Support\Collection as BaseCollection` would need to be given a particular scope, for the transformation selector to target.  But it might mean that you'd never be able to type `BaseCollection` in Goto Anything, I'm not sure. Certainly something to play around with :)

mrliptontea:
@keith-hall Thanks! I hope this will be merged and released soon.

I don't really understand how this works, but could `symbolIndexTransformation`/`symbolTransformation` be used for aliases?
So in this example entire `Illuminate\Support\Collection as BaseCollection` would have a separate scope, that would map `BaseCollection` symbol to `Illuminate\Support\Collection`? Just thinking out loud.

hkdobrev:
@wbond Do you think this would go into a dev build soon? This is quite a regression which affects me when working with a PHP project when I'm used to finding classes with "Go to symbol".

mrliptontea:
@hkdobrev If it's affecting your current work, you don't have to wait. Create a file
`[Your Sublime Text 3]/Data/Packages/PHP/Symbol List.tmPreferences`,
paste [the fixed version] (https://github.com/sublimehq/Packages/blob/604a0e2/PHP/Symbol%20List.tmPreferences), and save the file. Immediately you should see index rebuilding in the status bar.

BTW you can override any Sublime's package (default or installed via package control) this way.

keith-hall:
My advice with custom overrides is to delete them when the next build comes out, otherwise if further changes are made to a file as part of a build, those changes won't take effect for you, and you may end up wondering why something isn't behaving the way you expect.

wbond:
This was merged for the beta 3114, so it will definitely be part of dev build 3115. I don't know when that will come out, as I am taking a little time off right now. Certainly feel free to use the instructions to use as an override for now.

hkdobrev:
@wbond Taking time off is perfectly fine. I'm curious why the release process is built like that so newer version could be on the beta channel, but not on the dev one. This is not the case with other software using multiple update channels. I think this should be reconsidered.

wbond:
@hkdobrev It almost never happens that way. We just didn't want to do another full dev build (and push it into the next week) for a trivial one-line fix. Generally, the latest dev build is promoted to a beta build.

PR #386, opened 2016-05-12
[Python] Improvements on highlighting with focus on keywords

FichteFoll:
I finally ~~found the time~~ made some time to work on this. There is still things left to be done, but especially keyword matching is greatly improved now for keywords that only occur in conjunction with other keywords (in statements, e.g. `for ... in ...:`). Additionally, list comprehensions are more properly matched and only where applicale and highlighting of `print` and `exec` is smarter.

While I was working on it, I also 

- allowed dots in dotted_names to be surrounded by spaces,
- made set literals have an appearance
- unified context formatting (underscores and separated with empty lines)
- added scopes for annotations, default values and all commas
- made assignments in expressions illegal, as well as storage modifiers
- made semicolons get matched
- more...?

---

**Performance tests** (with `Lib/_pydecimal.py` (6300 lines) as bundled in Python 3.5)

Before: ~105ms
After: ~105ms

---

What remains to be done is to make statements only match at the beginning of a line so that `for` or `def` may not appear in the middle of a line. There are a couple ways of doing this (I assume you know enough about this now thanks to the JavaScript rewrite), but this will be left up for later since it's another medium-scale change to scope management.

Item-access and function calls also need to be revisited, not only because they still contain look-behinds but also because of how inaccurate they are. While function calls in general are alright, item access sometimes gets confused with list literals.

Also, decorators are clearly lacking at this moment. They should be parsed like an inline-keyword (in that they allow expressions to follow but do not span more than one line unless explicitly continued or wrapped in a group). They should also not be scoped as `entity.name.function.decorator` imo.
I have a [change for this ready in a branch](https://github.com/FichteForks/Packages/commit/cd5628519bf67abab43a292a898ae6a38b279770) but it really takes some getting used to considering how long it has been "wrong". I will not be including this change in this PR and open a new one once this is merged.

---
This change is [Reviewable](https://reviewable.io/reviews/sublimehq/packages/386)

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -31,13 +31,13 @@ contexts:
     - include: classes
     - include: functions
     - include: decorators
+    - include: modifiers
     - include: expressions
 
   expressions:
     - include: comments
     - include: constants
     - include: numbers
-    - include: modifiers
     - include: keywords
     - include: operators
     - include: lambda
@@ -97,10 +97,19 @@ contexts:
       scope: constant.numeric.integer.decimal.python
 
   modifiers:
-    - match: \b(global)\b
-      scope: storage.modifier.global.python
-    - match: \b(nonlocal)\b
-      scope: storage.modifier.nonlocal.python
+    - match: \b(?:(global)|(nonlocal))\b
+      captures:
+        1: storage.modifier.global.python
+        2: storage.modifier.nonlocal.python
+      push:
+        - include: line_continuation
+        - match: $\n?
+          pop: true
+        - match: ','
+          scope: punctuation.separator.storage-list.python
+        - include: name
+        - match: \S+
+          scope: invalid.illegal.name.storage.python
 
   keywords:
     - match: \b(?:(import)|(from))\b
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -292,9 +292,9 @@ contexts:
         2: keyword.control.flow.yield-from.python
 
   assignments:
-    - match: \+\=|-\=|\*\=|/\=|//\=|%\=|&\=|\|\=|\^\=|>>\=|<<\=|\*\*\=
+    - match: \+=|-=|\*=|/=|//=|%=|&=|\|=|\^=|>>=|<<=|\*\*=
       scope: keyword.operator.assignment.augmented.python
-    - match: \=
+    - match: '='
       scope: keyword.operator.assignment.python
 
   operators:
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -37,7 +37,7 @@ contexts:
     - include: modifiers
     - include: assignments
     - match: ;
-      scope: punctuation.separator.statements.python
+      scope: punctuation.terminator.statement.python
     - include: expressions
 
   inline_statements:
@@ -127,7 +127,7 @@ contexts:
     - include: import_alias
     - include: name
     - match: \*
-      scope: keyword.other.import-all.python
+      scope: constant.language.import-all.python
     - match: \S+
       scope: invalid.illegal.name.import.python
 
@@ -139,7 +139,7 @@ contexts:
     # async for ... in ...:
     - match: \b(async +)?(for)\b
       captures:
-        1: keyword.control.flow.async.python
+        1: storage.modifier.async.python
         2: keyword.control.flow.for.python
       push:
         - meta_scope: meta.statement.for.python
@@ -160,7 +160,7 @@ contexts:
     # async with ... as ...:
     - match: \b(async +)?(with)\b
       captures:
-        1: keyword.control.flow.async.python
+        1: storage.modifier.async.python
         2: keyword.control.flow.with.python
       push:
         - meta_scope: meta.statement.with.python
@@ -200,8 +200,7 @@ contexts:
     - match: \b(elif|else|finally|if|try|while)\b
       scope: keyword.control.flow.python
 
-  expressions:
-    # Always include this last!
+  expressions: # Always include this last!
     - include: comments
     - include: constants
     - include: numbers
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1266,13 +1266,13 @@ contexts:
 
   inline_for:
     - match: \b(for)\b
-      scope: keyword.control.flow.for.comprehension.python
+      scope: keyword.control.flow.for.generator.python
       push:
-        - meta_scope: meta.comprehension.python
+        - meta_scope: meta.expression.generator.python
         - match: \b(in)\b
           scope: keyword.control.flow.for.in.python
           set:
-            - meta_scope: meta.comprehension.python
+            - meta_scope: meta.expression.generator.python
             - include: line_continuation
             - match: '(?=[)\]}])'
               pop: true
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -59,6 +59,10 @@ contexts:
         - include: expressions
     - match: \)
       scope: invalid.illegal.stray.brace.round.python
+    - match: \]
+      scope: invalid.illegal.stray.brace.square.python
+    - match: \}
+      scope: invalid.illegal.stray.brace.curly.python
     - include: lists
     - include: dictionaries_and_sets
 
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -524,7 +524,7 @@ contexts:
             - include: expressions
 
   lambda:
-    - match: \b(lambda)(?=\s+)
+    - match: \b(lambda)(?=\s+|:)
       scope: storage.type.function.inline.python
       push:
         - meta_scope: meta.function.inline.python
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -338,16 +338,17 @@ contexts:
       scope: keyword.operator.logical.python
 
   classes:
-    - match: '^\s*(class)\s+(?={{identifier}})'
+    - match: '^\s*(class)\b'
       captures:
         1: storage.type.class.python
       push:
         - meta_scope: meta.class.python
+        - include: line_continuation
+        - match: $
+          pop: true
         - match: ':'
           scope: punctuation.section.class.begin.python
           set: maybe-docstrings
-        - match: '(?=^\s*$)'
-          pop: true
         - match: "(?={{identifier}})"
           push:
             - meta_content_scope: entity.name.type.class.python
@@ -357,16 +358,18 @@ contexts:
         - match: '(?=\()'
           set:
             - match: \(
-              scope: meta.class.inheritance.python punctuation.definition.inheritance.begin.python
+              scope: punctuation.definition.inheritance.begin.python
               set:
-                - meta_content_scope: meta.class.inheritance.python
+                - meta_scope: meta.class.inheritance.python
                 - match: \)
                   scope: punctuation.definition.inheritance.end.python
                   set:
+                    - include: line_continuation
+                    - match: $
+                      pop: true
                     - match: ':'
                       scope: meta.class.python punctuation.section.class.begin.python
                       set: maybe-docstrings
-                    - include: comments
                     - match: (?=\S)
                       pop: true
                 - match: ':'
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -19,7 +19,7 @@ variables:
   # We support unicode here because Python 3 is the future
   identifier_continue: '[[:alnum:]_]'
   identifier: '\b[[:alpha:]_]{{identifier_continue}}*\b'
-  path: '({{identifier}}\.)*{{identifier}}'
+  path: '({{identifier}} *\. *)*{{identifier}}'
   sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER)\b
 
 contexts:
@@ -178,7 +178,7 @@ contexts:
                     - meta_scope: entity.other.inherited-class.python
                     - include: identifier-parts
                     - match: '{{identifier}}'
-                    - match: ''
+                    - match: '(?=\S)'
                       pop: true
                 - include: expressions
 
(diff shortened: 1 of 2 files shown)


keith-hall:
> keywords that only occur in conjunction with other keywords (in statements, e.g. `for ... in ...:`). 

`in` doesn't have to come after `for` - i.e. `if 'abc' in 'abcdefg':`... but `for` always has to be followed by `in`. Could you add a test to prove the former still works please? :)

FichteFoll:
I checked the syntax with some sample Python code and couldn't spot any oddities in highlighting (including the `in` operator) but I didn't add tests for all the things I did not change, which is mostly basic expression stuff and operators.

I'll cherry-pick the `in` operator as an exception because it also occurs in for statements.

By the way, navigating the syntax test file (with all the string tests) is incredibly annoying. I always fold those immediately but since ST does not remember folding states on restart, that info gets lost all the time. Imo a 700 line test file should be split into separate files, but I remember reading that @wbond preferred single files.

FichteFoll:
Made a couple more changes. The commit messages are quite self-explanatory imo.

I also found out that keyword arguments in class inheritance is not matched yet (for Py3 metaclasses).

FichteFoll:
I noticed that I added a couple of inconsistencies in scope names now. Notably, `meta.class` vs. `meta.statement.for` and `punctuation.section.function` vs. `punctuation.definition.block.for` (and variants).

Since there are no guidelines for this yet, I'll leave it up to you to decide and not unify them just now. I personally prefer my choices.

Anyway, I think that should be all for this PR. I'll try not to make any further changes unless I notice regressions.

wbond:
Definitely sounds like some great improvements!

I will be reviewing this, but it may be pushed back a little due to only being back part time right now, fixing bug reports from 3114 and the number of commits.

FichteFoll:
Made some more minor updates and fixes that are still related to the overall scope to this PR, notably multi-line lambda support.

Python/Python.sublime-syntax
```diff
@@ -15,38 +16,205 @@ file_extensions:
...
   statements:
+    - include: inline_statements
+    - include: block_statements
     - include: classes
     - include: functions
     - include: decorators
+    - include: modifiers
+    - include: assignments
+    - match: ;
+      scope: punctuation.separator.statements.python
```
wbond: I know you proposed this, but I've been sticking with `punctuation.terminator` for now in other syntaxes since that seems to be in wide usage

Python/Python.sublime-syntax
```diff
@@ -15,38 +16,205 @@ file_extensions:
...
+                - include: import_name_list
+        - include: dotted_name
+
+  import_name_list:
+    - match: ','
+      scope: punctuation.separator.import-list.python
+    - include: import_alias
+    - include: name
+    - match: \*
+      scope: keyword.other.import-all.python
```
wbond: In JavaScript I used `constant.other` here to try and give it a little bit of a different flavor than the keyword `import` and `from`. You could probably argue for either one, since neither are perfect. It isn't really a keyword or much of a constant. Thus I chose constant since it color schemes tend to treat them differently.

Python/Python.sublime-syntax
```diff
@@ -15,38 +16,205 @@ file_extensions:
...
+
+  import_alias:
+    - match: \b(as)\b
+      scope: keyword.control.import.as.python
+
+  block_statements:
+    # async for ... in ...:
+    - match: \b(async +)?(for)\b
+      captures:
+        1: keyword.control.flow.async.python
```
wbond: `async` should probably be `storage.modifier` here. Having some instances of `async` (before `def`) be one scope and others be a different scope would feel very off to me as a user. I think that `storage.modifier` is more correct than `keyword.control` since `async` is modifying how the `for` loop is being executed.

Python/Python.sublime-syntax
```diff
@@ -15,38 +16,205 @@ file_extensions:
...
+              pop: true
+            - include: expressions
+        - match: ':'
+          scope: invalid.illegal.missing-in.python
+          pop: true
+        - include: target_list
+    # async with ... as ...:
+    - match: \b(async +)?(with)\b
+      captures:
+        1: keyword.control.flow.async.python
```
wbond: Same here as above about `storage.modifier`

Python/Python.sublime-syntax
```diff
@@ -15,38 +16,205 @@ file_extensions:
...
+            - match: ':'
+              scope: punctuation.definition.block.except.python
+              pop: true
+            - include: name
+        - include: target_list
+    - match: \b(elif|else|finally|if|try|while)\b
+      scope: keyword.control.flow.python
+
   expressions:
+    # Always include this last!
```
wbond: Does this comment apply to `comments` or `expressions`?

Python/Python.sublime-syntax
```diff
@@ -55,11 +223,21 @@ contexts:
...
+    - match: \]
+      scope: invalid.illegal.stray.brace.square.python
+    - match: \}
+      scope: invalid.illegal.stray.brace.curly.python
     - include: lists
-    - include: dictionaries
+    - include: dictionaries_and_sets
+    # For debugging
+    # - match: \S
+    #   scope: invalid.illegal.unmatched
```
wbond: Did you intend to leave these lines here?

Python/Python.sublime-syntax
```diff
@@ -232,27 +403,30 @@ contexts:
     - match: '='
       scope: keyword.operator.assignment.python
       push:
+        - meta_scope: meta.function.default-value.python
```
wbond: I've been trying not to stack similarly named scopes (for instance here there will be `meta.function.parameters.python meta.function.default-value.python`). I'm not sure if there is a great solution to this or not.

Python/Python.sublime-syntax
```diff
@@ -232,27 +403,30 @@ contexts:
...
         - match: '(?=[,)])'
           pop: true
         - include: expressions
     - match: ':'
-      scope: punctuation.separator.annotation.python
+      scope: punctuation.separator.annotation.parameter.python
       push:
+        - meta_scope: meta.function.annotation.parameter.python
```
wbond: This is a similar scope stacking as just above where `meta.function` is stacked in the list.

Python/Python.sublime-syntax
MattDMo: `terminator` seems fine here, as that's what `;` really is. I'd stick with `separator` for commas, etc., as they truly are separators.

Python/Python.sublime-syntax
MattDMo: What about `meta.keyword`? Isn't this kind of what the `meta` "class" is for?

Python/Python.sublime-syntax
```diff
@@ -15,38 +16,205 @@ file_extensions:
...
+      scope: punctuation.separator.import-list.python
+    - include: import_alias
+    - include: name
+    - match: \*
+      scope: keyword.other.import-all.python
+    - match: \S+
+      scope: invalid.illegal.name.import.python
+
+  import_alias:
+    - match: \b(as)\b
```
MattDMo: Remember to scope this the same as `import` and `from`.

Python/Python.sublime-syntax
wbond: `meta` is more to create a scope that helps disambiguate contained scopes. It is used to kind of mark up larger sections of the code, and typically should not be colored via the color scheme, but can be used to make a more specific scope selector.

There is already a `meta.statement.import` on this, but it would be unmaintainable to have color schemes target things to that level for all of the different languages that should be supported.

Python/Python.sublime-syntax
```diff
@@ -342,34 +519,27 @@ contexts:
...
               pop: true
             - include: keyword_arguments
+            - match: ','
+              scope: punctuation.separator.parameters.python
+            - include: inline_for
             - include: expressions
 
   lambda:
-    - match: \b(lambda)(?=\s+)
+    - match: \b(lambda)
```
wbond: Won't this match identifiers such as `lambda_container`?

Python/Python.sublime-syntax
```diff
@@ -3,8 +3,9 @@
 name: Python
 file_extensions:
   - py
-  - rpy
   - pyw
+  - pyi
+  - rpy
```
MattDMo: PythonImproved also has `py3`, would you want to add that? Was there ever an equivalent `py2`?

Python/Python.sublime-syntax
```diff
@@ -342,34 +519,27 @@ contexts:
...
-              captures:
-                1: variable.parameter.python
-                2: punctuation.separator.parameters.python
-
-  late-keywords:
-    - match: \b(def|lambda)\b
-      scope: storage.type.function.python
-    - match: \b(class)\b
-      scope: storage.type.class.python
-    - match: \b(async)\b
```
wbond: Removing this causes a bare `async` to not highlight until `for`, `with` or `def` is typed.

Python/Python.sublime-syntax
```diff
@@ -1050,3 +1255,42 @@ contexts:
   strings:
     - include: string_quoted_double
     - include: string_quoted_single
+
+  inline_for:
+    - match: \b(for)\b
+      scope: keyword.control.flow.for.generator.python
+      push:
+        - meta_scope: meta.expression.generator.python
+        - match: \b(in)\b
```
MattDMo: Silly question, but does this `in` also cover comprehensions (list, dict, etc.) as well as generators?
```py
[function(item) for item in iterable]
```

Python/Python.sublime-syntax
MattDMo: Is this actually desirable? `async` on its own doesn't mean anything, and the lack of coloration could prompt the programmer to continue with `for`, `with`, or `def` (and maybe `yield` in the future...)

wbond: Generally it is nice to not have colors jumping around as the user is typing, so there are situations where we highlight things that aren't valid constructs. Trying to have the syntax provide linter-like feedback usually just leads to a less than desirable editing experience.

MattDMo: OK, fair enough. Although we do provide some sort of linting experience in JSON, for example, if someone enters a key: value pair, then a comma, then tries to continue typing non-whitespace on the same line. I don't know if that's equivalent. Still, you're the boss, boss :wink:

wbond: Yes, we do that in some situations. Sometimes it works ok, sometimes not so great. In general I think I fall more on the side of erring on the side of caution. Such as a few languages currently highlight a `%` in a string following by a character that is not a valid `printf()` formatter as invalid. However, may strings are not used with `printf()`, so it isn't really invalid.

Python/Python.sublime-syntax
FichteFoll: I haven't seen `py3` anywhere and it seems fairly non-standard. Do you have any sources on it?

Python/Python.sublime-syntax
FichteFoll: Indeed, there should be a \b.

Python/Python.sublime-syntax
FichteFoll: I'm ok with standardizing to  `terminator` later (e.g. a PR). Better be consistent than correct in this case, imo.

Python/Python.sublime-syntax
FichteFoll: I was kinda at a loss here since there wasn't a scope for this previously and I didn't look into other languages either, although something with a "special meaning" should probably get a keyword scope and not constant. It really doesn't matter though, so I'll adjust to the current standard.

Python/Python.sublime-syntax
FichteFoll: To `expressions`. The comment doesn't make sense if it was for `comment`, which is the first include.

Python/Python.sublime-syntax
FichteFoll: Yes, kind of. It's a good place to look for unmatched "anything" since in the current syntax almost everything if not really everything gets matched by some pattern.

Python/Python.sublime-syntax
wbond: Consistent with what? Other languages all use `punctuation.terminator` for `;`, as far as I know.

Languages using `punctuation.terminator`:

 - C/C++
 - JavaScript
 - CSS
 - Java
 - Objective-C
 - PHP
 - TCL

`punctuation.separator` is used almost exclusively for `,` and `:`.

Python/Python.sublime-syntax
FichteFoll: Imo meta scopes shouldn't be restricted in precision for a reason like this. We are still in the parameters "context" after all and just entered an even more specific meta scope, so the older one should remain.

Same for the annotation comment below.

Python/Python.sublime-syntax
FichteFoll: Unlike `def` and `class`, `async` will only become a proper keyword in 3.7. Especially in Python 2 code, it is still a valid identifier.

The other two keywords from that scope were specifically considered by making their contexts end early (on line breaks or comments).

Python/Python.sublime-syntax
FichteFoll: Yes. There are tests for this as well.

FichteFoll:
Review status: 0 of 2 files reviewed at latest revision, 12 unresolved discussions.

---

*[Python/Python.sublime-syntax, line 135 \[r5\]](https://reviewable.io:443/reviews/sublimehq/packages/386#-KJD-BL51fDkcJLQJP4p-r5-135:-KJD0NzlWu56oCbXUdyG:-324675483) ([raw file](https://github.com/sublimehq/packages/blob/b5a617e71f640038102b589ffba75571574d5ab9/Python/Python.sublime-syntax#L135)):*
<details><summary><i>Previously, MattDMo (Matt Morrison) wrote…</i></summary>
> Remember to scope this the same as `import` and `from`.
</details>
I'm using `keyword.control.import.as`. where `import` gets `keyword.control.import` and `from` gets `keyword.control.import.from`, so I'm not sure what you're referring to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sublimehq/packages/386)*

MattDMo:
I hadn't checked all three scopes in the full file at the time I made the comment, so just ignore it 😬

wbond:
I'm going to remove `meta.name` for now. Rationale:

 - `meta` scopes are for marking up "larger sections" of the document, to provide contextual information such as may be useful for targetting the usage of a more specific scope
 - currently, `meta.name` is applied to individual identifiers
 - it is likely to lead to color schemes targeting `meta.name` for highlighting
 - if we were to target such tokens, it is something I'd like to standardize, and something like `entity.other.identifier` may make more sense
 - with the current implementation it gets duplicated with `meta.dotted.name` for tokens like `int`, resulting in scopes of `meta.dotted.name meta.name storage.type`

wbond:
I'm also going to remove `meta.dotted.name` right now. I think this has better potential than `meta.name` since it is marking up a "path" that includes multiple identifiers and one or more accessors. However, `meta.name.dotted` wouldn't make sense it other languages such as C++.

We can discuss this more and then work on coming up with a scope name that we can apply across languages and then work on getting it applied to the various syntaxes that have been rewritten recently.

PR #392, opened 2016-05-13
[Scala] Misc improvements and additions

djspiewak:
- `macro` is (sort of) a keyword
- `???` isn't a keyword, but everyone treats it as if it is, so I think the highlighting is valid (note: implementing this unveiled a bug in Sublime's `\b` support, detailed in a comment)
- Parameter list pre-`:` highlighting (e.g. `foo: Int`) no longer applies at arbitrary points in the file.  Instead, parentheses are detected as a sub-mode, allowing the special highlighting to take place.  This means that the phrase `a :: b` is now *correctly* highlighted (except inside of parentheses, unfortunately; work in progress!)
  + Also, I removed the special override on the right-hand side of the `:`, so the `Int` in `foo: Int` highlights uniformly with how `Int` highlights in normal text (i.e. as a special primitive type)
- `def` identifier syntax now more closely matches the Scala specification (I still haven't fixed `val` or extraction).  This allows `def +(x: Int)` to highlight correctly.

Also, the YAML format is infinitely easier to edit than the old plist version.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -157,7 +157,7 @@ contexts:
   keywords:
     - match: \b(return|throw)\b
       scope: keyword.control.flow.jump.scala
-    - match: \b(else|if|do|while|for|yield|match|case)\b
+    - match: \b(else|if|do|while|for|yield|match|case|macro)\b
       scope: keyword.control.flow.scala
     - match: \b(catch|finally|try)\b
       scope: keyword.control.exception.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -19,8 +19,8 @@ contexts:
     - include: constants
     - include: char-literal
     - include: scala-symbol
-    - include: empty-parentheses
     - include: parameter-list
+    - include: empty-parentheses
     - include: qualifiedClassName
     - include: xml-literal
   block-comments:
@@ -175,10 +175,16 @@ contexts:
         - include: nest-curly-and-self
     - include: main
   parameter-list:
-    - match: '([a-zA-Z$_][a-zA-Z0-9$_]*)\s*:\s*([A-Za-z0-9][\w|_|?|\.]*)?,?'
-      captures:
-        1: variable.parameter
-        2: entity.name.class
+    - match: '\('
+      push:
+        - match: '\)'
+          pop: true
+        - match: '([a-zA-Z$_][a-zA-Z0-9$_]*)\s*:\s*'
+          captures:
+            1: variable.parameter
+            2: entity.name.class
+        - include: main
+        - include: parameter-list
   qualifiedClassName:
     - match: '(\b([A-Z][\w]*))'
       captures:
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -161,6 +161,8 @@ contexts:
       scope: keyword.control.flow.scala
     - match: \b(catch|finally|try)\b
       scope: keyword.control.exception.scala
+    - match: (\?\?\?)              # can't be guarded by \b for some reason (maybe bug?)
+      scope: keyword.control.exception.scala
   nest-curly-and-self:
     - match: '\{'
       captures:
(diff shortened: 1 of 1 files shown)


gwenzek:
Hi, did you try the Scala syntax I propose in #94 ? It doesn't support the macro keyword yet but a lot of other things have been improved.

djspiewak:
@gwenzek I haven't installed and played around with it, but right off the bat, there are several features that I'm not fond of.  The brace coloring based on syntactic element, for example, is technically cool but not something I really want.  Also I quite like the parameter highlighting of the default mode (even with its problems), so I would rather keep that.

My changes still don't correctly handle annotations (whereas your proposed mode does), but that's mostly because I haven't implemented it yet.  Obviously it's not hard to do (though harder if you want to get it right w.r.t. pattern matching).  My mode also correctly handles some other things, like type lambdas, which I don't think your mode addresses.

gwenzek:
@djspiewak thanks for your feedback.

The braces coloring is a feature that I enjoy a lot, but anyway it's opt-in since you have to choose a theme that support them. My syntax is just offering a new option here.

This type lambda `({type λ[α] = A => α})#λ` works for me, but it's not among the features I use really often, so if you have more complicated example that doesn't work please post them.

What do you don't like in my parameter highlighting ? For example in the current syntax the `=>` in:

    def map[B](f: A => B): Box[B] = Box(f(x.a))

won't have the same color than A and B next to it, which I find wrong.
And in this one it will say that `a` is a type and `b` a parameter:

    def map[b](f: a => b): Box[b] = Box(f(x.a))

All this issues are addressed by my PR.

wbond:
Generally, I'd rather see more smaller sets of changes (like this) because they are easier for me to review and wrap my head around, being unfamiliar with the language. I have quite a bit of experience with rewriting these syntaxes now, but the most time consuming can be to get a grasp of the language and what features are available and what syntax edge cases need to be handled. I generally have a pretty good handle of what scopes we should be using and constructs we can use to handle complex syntax.

One of the biggest things that will unstick this process and start moving it forward would be to create a somewhat extensive `syntax_test_scala.scala` file using our syntax test system. You could look at the Python package to see an example. Ideally it would cover the main points of the language so that we aren't making any large regressions as we clean this up and make it more correct. Here are some types of things I would expect covered by the syntax tests:

 - function definitions with parameters and return type if applicable
 - class or other data structure definitions (classes, structs, traits, interfaces, implementations, etc)
 - namespace/package/module constructs
 - literals (string, number, regexp, etc)
 - constants (bools, null/nil)
 - macros?
 - comments
 - basic types
 - keywords (`return`, `if`, `else`, etc)

djspiewak:
@wbond I don't mind breaking things up into smaller chunks.  I actually have almost as many commits outstanding locally as I have in this PR, but I didn't want to toss it all in at once.

:+1: for tests.  This will also give us a firmer basis on which we can argue about exactly *how* things should be highlighted.

Random question: locally, I also have an SBT highlighting mode which makes some assumptions about the contents of `.sbt` files (namely, that they're using a superset of Scala) and highlights them accordingly.  Would this be something you would be interested in seeing contributed as part of the core Scala mode?

wbond:
So does SBT just add well-known function names, or something like that?

djspiewak:
@wbond Partially.  SBT has several macros which are critical to its syntax.  Notably:

- A `.value` suffix
- The `project`, `settingKey` and `taskKey` functions
- The `in` quantifier
- Scoping denoted by the `in` quantifier and the `:=` operator
- Several legacy operators that are still useful (e.g. `<<=`, `<++=` and such)
- Artifact designators (`%` and `%%`)
- Artifact quantifier strings (literally syntax within string literals derived from Ivy)

I've been using my prototype mode (which has some issues!) for a while now, and it has been a palpably better experience than the default Scala mode when editing `.sbt` files.

FichteFoll:
From your comment, @djspiewak, I'd argue that including a syntax definition for .sbt files in the default Scala package makes sense. Ideally it would be based on the Scala syntax and use includes/pushes to not duplicate too many patterns.

djspiewak:
Closing this for the time being, as I will be splitting it into multiple separate PRs with tests.

PR #394, opened 2016-05-13
Add standard zsh file extensions

indrajitr:
These bring support to zsh standard files at par with the bash ones.


wbond:
This syntax really needs a rewrite since there are numerous bugs. However, as it is implemented now, it is focused on Bash. My understanding is that bash and zsh share a number of syntactic elements, but if we start automatically selecting the bash syntax for zsh, we are likely going to be highlighting things incorrectly.

I think the path forward probably involves making variations for bourne/bash, zsh, tcsh.

jrappen:
fixed in 34284ab

PR #395, opened 2016-05-13
Add sbt to the list of supported file extensions

indrajitr:
sbt follows a subset of scala syntax now.
So it makes sense to honor scala syntax in sbt files.


wbond:
Thanks!

PR #397, opened 2016-05-14
[Python] support line continuation in strings and assign meta scope to docstrings

keith-hall:


FichteFoll:
The common scope name for docstrings is `comment.block.documentation`.

keith-hall:
Thanks @FichteFoll, I have corrected the scope

wbond:
I haven't looked to see if this will merge cleanly with #386, so I am going to wait on merging.

FichteFoll:
Syntax changes should merge since I didn't change these parts, but the test file will have conflicts.

I will split the test file into multiple when #386 (and this) are merged.

wbond:
The downside of splitting test files is that now you have these very limited contexts within which tests run. That form of tests **is** useful to ensure there isn't a regression in the more straight-forward parts of the syntax.

By placing a bunch of syntax mixed all together, especially various statement and expression-level constructs, you can usually find errors in context manipulation (push/pop/set).

I realize that navigating it may be more of a pain, but I think the tests should favor catching regressions over moving around inside of the files.

vovkkk:
> The common scope name for docstrings is comment.block.documentation [1].

Why? 
I noticed today it looks like comment, thought it is a bug.
They are doc<b>strings</b>, hence must look like strings.


# ST2 :+1: 






# ST3 :-1: 






TextMate convention is not applicable here (and old tmLanguage even violate it as you see on screenshot above), because docstring in Python is an object per se (string, unicode, raw, bytes), while comment is not an object; so this convention might be acceptable for html, js, whatever, but not for Python.

keith-hall:
> They are doc**strings**, hence must look like strings.

How something looks is defined by the color scheme. The point of keeping the scope names consistent across languages is so that the color schemes don't need to have tons of crazy unmaintainable rules, and at the same time, ensuring the scopes are semantically correct.

This PR is so that color schemes will be able to correctly identify the docstrings by their scope name, and thus give them a specific color. It is up to the color scheme to decide whether they be colored the same as strings or comments, or completely differently. If your color scheme doesn't color it they way you like after this PR has been merged, you could either raise an issue on that color scheme's repo, or just override the color scheme with your own customized version. You can even use a selector to color Python docstrings differently to other language docstrings, because, as you say, they have more meaning in Python.

vovkkk:
> so that the color schemes don't need to have tons of crazy unmaintainable rules, and at the same time, ensuring the scopes are semantically correct.

This is exactly my point! strings must be strings, semantically and in every other aspect.

> You can even use a selector to color Python docstrings differently to other language docstrings, because, as you say, they have more meaning in Python.

Docstrings have same meaning in the most languages which actually have them.
In C, JavaScript, etc., there is no such thing as docstrings, there is block comment, however, which by its nature is a different entity.

Though my comment was more about https://github.com/sublimehq/Packages/commit/7a322b2de7bbbc71dad04f73f370fccb3cd52280 then I’ve noticed discussion in this pr, so

Oh now I see @wbond reaction to @keith-hall message, well, thank you guys for giving me another reason to keep using ST2.

wbond:
@vovkkk We've provided a clear, and reasonable way for you to color docstrings whatever way you want. I'm unclear what the issue is. In fact, you can use old, outdated syntax definitions for Python in Sublime Text 3 if you would like. Or you could create a custom override with the current Python syntax but using a different scope for docstrings. You really have so many options. It would seem like such a shame to throw away all of the bug fixes and improvements because you don't like the hidden, scope name we applied to a section of code. However, that is your right. :smile:

_Semantically_, docstrings are not like other strings, that is why they are called something different. They use the _syntax_ of strings, but affect the program in no way whatsoever, exactly like comments do. In fact, every other language uses a multi-line comment block to do the same thing as docstrings provide.

I hope this information ends up being useful for you and others in terms of understanding why we changed things, and how you can customize them as you see fit.

vovkkk:
> They use the syntax of strings, but affect the program in no way whatsoever, exactly like comments do.

That is not true. They are available in runtime, comments are not.
Docstrings are objects (of different types I mentioned it prev.), comments are not.

Docstring on module level looks like string, on class level — comment. Are they different?

In ST2 if you use `r'''\n'''` as docstring it has proper highlighting (and scopes), in ST3 you just throw it away fixing sth that was not broken and consider it improvement. :disappointed: 

Sorry, if I’m maybe too dramatic, ofc customisation is an option, this change simply feels like regression to me, so

FichteFoll:
> Docstring on module level looks like string, on class level — comment. Are they different?

No, this functionality just hasn't been added yet. It's on my todo.

andreixk:
this is still a bug in build 3126

FichteFoll:
@andreixk can you show us an example with code and a screenshot of how it renders for you? Can you also verify that you are indeed using the Python.sublime-syntax shipped with Sublime Text and not a third-party syntax or a custom override?

andreixk:
It's only broken in when trying to use multiline SQL statements joined with `\` in a Python file
I've checked, it's the same Python.sublime-syntax shipped with the app

7e8c5eb6baa95c334e8fef5c29c117e720e67112

FichteFoll:
Can reproduce. It's actually broken in everything that's not a normal thing (regular expression, even binary).

PR #400, opened 2016-05-16
[Haskell] comment fixes

### LINKED ISSUE #399: [Haskell] Some tweaks needed on comments
Hi!

I sometimes develop Haskell programs and I always use ST. There are two errors that I wish to be fixed when you can:

1) When you use the "toogle comment" option of ST, it just inserts two dashes in the beginning of each marked line. There are situations where you get a GHC error for doing this, so I think is better to put two dashes followed by one space on each line, to prevent those errors (sorry I don't find any GHC error use case, I've tried to and all my code compiles :(). Anyways, taking a look on [the Traversable typeclass code on Hackage](https://hackage.haskell.org/package/base-4.8.2.0/docs/src/Data.Traversable.html#Traversable) you can see they put a space after the two dashes on each line. I consider that the best practice for single line comments and I always do it that way.

2) ST does not hightlight well code snippers like:

```
{-
class (Functor t, Foldable t) => Traversable t where
{-# MINIMAL traverse | sequenceA #-}

-- | Map each element of a structure to an action,
-- evaluate these actions from left to right, and
-- collect the results. For a version that ignores
-- the results see 'Data.Foldable.traverse_'.
traverse :: Applicative f =>
(a -> f b)
-> t a
-> f (t b)
traverse f = sequenceA . fmap f
-- | Evaluate each action in the structure from
-- left to right, and collect the results.
-- For a version that ignores the results see
-- 'Data.Foldable.sequenceA_'.
sequenceA :: Applicative f => t (f a) -> f (t a)
sequenceA = traverse id
-}
```
On that code, there is a pragma indicating that the Foldable typeclass must be instantiated implementing one of those two methods. A pragma in Haskell has the form *{-# PragmaName #-}*. But the comments parser that ST uses does not parse that like a pragma expression but like the comment end *-}* and so the highlighting is incorrect. I see on ST code like not commented when it really is, and the code compiles fine.

keith-hall:
fixes #399

### DIFF
--- a/Haskell/Haskell.sublime-syntax
+++ b/Haskell/Haskell.sublime-syntax
@@ -94,13 +94,11 @@ contexts:
         1: punctuation.definition.preprocessor.c
     - include: pragma
     - match: '"'
-      captures:
-        0: punctuation.definition.string.begin.haskell
+      scope: punctuation.definition.string.begin.haskell
       push:
         - meta_scope: string.quoted.double.haskell
         - match: $|"
-          captures:
-            0: punctuation.definition.string.end.haskell
+          scope: punctuation.definition.string.end.haskell
           pop: true
         - match: '\\(NUL|SOH|STX|ETX|EOT|ENQ|ACK|BEL|BS|HT|LF|VT|FF|CR|SO|SI|DLE|DC1|DC2|DC3|DC4|NAK|SYN|ETB|CAN|EM|SUB|ESC|FS|GS|RS|US|SP|DEL|[abfnrtv\\\"''\&])'
           scope: constant.character.escape.haskell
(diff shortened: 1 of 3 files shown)


Haskell/Haskell.sublime-syntax
```diff
@@ -187,20 +183,25 @@ contexts:
       scope: punctuation.separator.comma.haskell
   block_comment:
     - match: '\{-(?!#)'
-      captures:
-        0: punctuation.definition.comment.haskell
+      scope: punctuation.definition.comment.begin.haskell
       push:
         - meta_scope: comment.block.haskell
+        - match: '\{-#'
```
wbond: Is this to handle commenting out preprocessor directives?

keith-hall: yes, to avoid having to change the comment block match which has a negative lookahead for `#`

keith-hall: * its to handle preprocessor directives inside a comment block

keith-hall: ugh, not explaining myself very well, sorry. It is so that preprocesor directives inside a comment block are correctly treated as nested comments. for more info, see the linked issue :)

wbond:
Ok, great. Thanks for taking the time to figure out solutions for #399!

PR #405, opened 2016-05-17
[JavaScript] get and set must be followed by a method name to be an accessor

### LINKED ISSUE #404: [JavaScript] Syntax highlighting in function name, params and statement
The params, and everything inside the function statement has the same scope (`source.js meta.class.js meta.block.js meta.function.declaration.js entity.name.function.js`), except for the `cacheName` parameter and `.set()` inside the function statement.

screen shot 2016-05-17 at 2 15 29 am

Here is another screen with more context. All of the function names should be the same scope and color, but some are white, one blue, one green. 

screen shot 2016-05-17 at 2 23 07 am

The `has` and `get` function names have the scope `source.js meta.function-call.js variable.function.js`, which is the same scope as `require` in `var foo = require('bar');`

Let me know if you need more detail.

keith-hall:
...otherwise they are just normal method names. fixes #404

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -940,7 +940,7 @@ contexts:
         - include: comments
 
   method-declaration:
-    - match: \b(get|set)\b\s*
+    - match: \b(get|set)\b(?!\s*\()\s*
       scope: meta.function.declaration.js
       captures:
         1: storage.type.accessor.js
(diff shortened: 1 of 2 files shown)


FichteFoll:
How about attempting to match method definitions before the get and set keywords?

keith-hall:
As `get` and `set` are valid identifiers, surely they would then always be seen as the method name and not as keywords?

I realize that my solution is not perfect, because the `(` could be defined on the next line, and it would therefore suffer the same problem as before this change.

Feel free to play around with it - if you come up with a better solution I'll close this PR.

FichteFoll:
They wouldn't be matched as the method name since there needs to be whitespace between `get|set` and the actual method name.

wbond:
Thanks!

PR #418, opened 2016-05-21
[C++] Ignore indention of multi-line block comments

### LINKED ISSUE #417: [C++] Incorrect indention after preprocessor statements and block comments
The following has been tested with C source and header files and indention using tabs (tab size 4).

For instance, pressing enter on a line containing a preprocessor statement like `#include <stdlib.h>` or `#define XYZ` will cause an indention of a single space on the next line. Pressing tab a few lines further down will also first insert a space and only then tabs, even if the lines in between are empty - but strangely only if the line is empty, if it already contains content a tab is inserted. If a preprocessor statement is indented with tabs itself the next line will just be indented by a single space.

    #include <stdlib.h>
     ^--- this line starts here after a single space
    ^--- pressing tab here will indent by a single space first, then tabs are inserted
    |--- but if the cursor is at the beginning of the line below (following the line above) a tab is inserted
    #include <stdio.h>

        #include <unistd.h>
     ^ --- we end up here (one space indented)

This is very similar for block comments:

    /* asdf */
     ^ --- we end up here instead of
    ^ --- here
    | --- pressing tab here will indent by a single space first

    /**
     * comment
     */
     ^--- we end up here instead of
    ^--- here

tobiasbrunner:
Lines matching ^\s+\*.*$ in multi-line block comments previously triggered a
matching indention on lines following it (even if e.g. preprocessor statements
were in between).

For example:

```
 /*
  * Copyright ...
  */
  ^--- this line is indented by one space
 #include <stdlib.h>
  ^--- this line too
```

Fixes #417.

### DIFF
--- a/C++/Indentation Rules.tmPreferences
+++ b/C++/Indentation Rules.tmPreferences
@@ -27,7 +27,7 @@
 		</string>
 
 		<key>unIndentedLinePattern</key>
-		<string>^\s*((/\*|.*\*/|//|#|template\b.*?&gt;(?!\(.*\))|@protocol|@interface(?!.*\{)|@implementation|@end).*)?$</string>
+		<string>^\s*((/\*|\*|.*\*/|//|#|template\b.*?&gt;(?!\(.*\))|@protocol|@interface(?!.*\{)|@implementation|@end).*)?$</string>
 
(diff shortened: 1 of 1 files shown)


wbond:
Awesome, thanks for the fix!

PR #429, opened 2016-05-24
[HTML] support _ in tag names

yangsu:
image


Briles:
These changes would probably be better suited as a part of `entity.name.tag.custom.html` as they are are custom tags, though the current custom tag regex already matches underscores.

According to the [spec](https://www.w3.org/TR/custom-elements/#valid-custom-element-name), a custom tag requires a hyphen.

yangsu:
Looking more closely at the syntax definition, it looks like this behavior is already handled. Will close the PR. Thanks for the feedback!

PR #443, opened 2016-06-01
[PHP] expand test coverage

keith-hall:
I just thought I would extend the PHP test coverage slightly:

- to prove that PHP inside HTML attribute values is still recognized as being part of that attribute value etc. and that the HTML `meta.tag` scope still applies
- to prove that HTML tags are still recognized when nested deeply in PHP blocks - the tags were already there but the tests weren't.


wbond:
Thanks!

PR #448, opened 2016-06-02
[JavaScript] Update symbol list with correct class name scope

jcberquist:


jcberquist:
The indexed symbol list already had the new scope `entity.name.class.js` so I guess no class names were being indexed :grinning:

PR #449, opened 2016-06-03
[JavaScript] Improve test coverage

Briles:
Improved test coverage & fixed the order of conditional variables


FichteFoll:
Talking about test coverage, I wonder if it would be possible to extend the syntax tester with a "test coverage" mode that would report a syntax's contexts that never match anything.

PR #456, opened 2016-06-06
[Java] add tests, improve generic type detection

### LINKED ISSUE #455: Java Storage Syntax Is Broken By Nested Types
In the following image, the second example is the correct variant; but it displays incorrectly due to the `.` character of the `FileManager.FileData` enum.&nbsp; Particularly, the `,` & `>` characters seem to be out of scope.

demo
&nbsp;

-----

&nbsp;
scope1

scope2
&nbsp;

-----

&nbsp;
scope3

scope4
&nbsp;

-----

&nbsp;
__ST Version:__ Stable Channel, Build 3114
__OS:__ Windows 10 Pro (64-bit), Build 10586

keith-hall:
fixes #455

### DIFF
--- a/Java/Indexed Symbol List.tmPreferences
+++ b/Java/Indexed Symbol List.tmPreferences
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>name</key>
+	<string>Symbol List: Classes</string>
+	<key>scope</key>
+	<string>source.java entity.name.class.java, source.java entity.name.function.java</string>
+	<key>settings</key>
+	<dict>
+		<key>showInIndexedSymbolList</key>
+		<integer>1</integer>
+	</dict>
+</dict>
+</plist>
--- a/Java/Java Server Pages (JSP).sublime-syntax
+++ b/Java/Java Server Pages (JSP).sublime-syntax
@@ -59,7 +59,7 @@ contexts:
                   scope: constant.character.escape.jsp
     - match: "(<%[!=]?)|(<jsp:scriptlet>|<jsp:expression>|<jsp:declaration>)"
       captures:
-        1: punctuation.section.embedded.jsp
+        1: punctuation.section.embedded.begin.jsp
         2: meta.tag.block.jsp
       push:
         - match: (?<=</jsp:scriptlet>|</jsp:expression>|</jsp:declaration>|%>)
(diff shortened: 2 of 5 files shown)


wbond:
We should ensure that class names are in the symbol list and index. It may also make sense to ensure other data structures are there too (I don't recall too much of what Java entails off of the top of my head).

FichteFoll:
Java's symbol selectors are targetting solely meta scopes for some reason.

wbond:
Yes, a lot of languages target the full class/function definition. This may be useful when scrolling through a list, but isn't so useful for GotoSymbol since it makes fuzzy matching less accurate.

It seems to me the meta scopes could be useful in the future if we implemented some way to create a tree structure for browsing code, however for the symbol indexer it is useless, and for the symbol list I think it is of dubious value.

FichteFoll:
Ah yes, that makes sense.

I can confirm that the indented symbol list is indeed quite useful as an overview (as it is in Python), but not for goto definition. These should probably be separated eventually.

wbond:
Just to clarify, Goto Symbol and Goto Definition are separate and have separate settings in `.tmPreferences` - `showInSymbolList` and `showInIndexedSymbolList`. There are also settings to manipulate the raw content, `symbolTransformation` and `symbolIndexTransformation`.

FichteFoll:
So you're saying that goto definition only looks at the IndexedSymbolList settings, even for the current file? That's very good to know.

Don't forget to add `entity.name.class` (and others?) to `Default/Indexed Symbol List.tmPreferences` then ;)

keith-hall:
I've updated the Java Indexed Symbol List. Enums are scoped the same as classes, so don't need any special rules.

FichteFoll:
In theory, these scopes should be covered by Default/Indexed Symbol List, since that contains `entity.name.function, entity.name.type` currently (and needs to be updated to list the other entities).

keith-hall:
@FichteFoll, but I notice that if I delete the `Packages/Java/Indexed Symbol List.tmPreferences` file, Goto Definition no longer works - it seems to not use the `Packages/Default/Indexed Symbol List.tmPreferences` file?

FichteFoll:
@keith-hall that's because the file doesn't include `entity.name.class` and others like `entity.name.enum` yet and still uses the only recently deprecated `entity.name.type`. Since the Default package is not part of this repo, this has to be bundled in a future build.

wbond:
Just to clarify, `entity.name.type` isn't deprecated for all uses, but `entity.name.class` should be used instead of `entity.name.type.class`.

I have been preferring to enumerate entity name scopes in the tmPreferences files for now so that users with older builds will still get correct functionality.

PR #459, opened 2016-06-06
[Python] Highlight decorators as inline-statements

FichteFoll:
This PR more appropriately matches decorators as "inline statements" and the `@` character as the designating keyword. It also gets rid of the odd `entity.name.function.decorator` scope (which is semantically something different) and instead applies a meta scope to the *entire* decorator construct, not just the initial identifier like before.

Before:

After:

(for Monokai, imagine both class names and decorators being green)


wbond:
This looks great, thanks!

PR #461, opened 2016-06-07
[Batch File] Ensure SET /A contexts are correctly pushed

Briles:
Previously, `/A num*=5` would get highlighted although the `SET` command was missing.
`SET /A` contexts should now only be pushed when the `SET` command is encountered.

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -49,25 +49,28 @@ contexts:
       scope: keyword.command.dosbatch
 
   command_set:
-    - match: '(?i)\bset\b'
+    - match: '(?i)\bSET\b'
       scope: keyword.command.dosbatch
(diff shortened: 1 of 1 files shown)


Batch File/Batch File.sublime-syntax
```diff
@@ -49,25 +49,28 @@ contexts:
...
-          push:
-            - meta_scope: string.quoted.double.dosbatch
-            - match: '"'
-              scope: punctuation.definition.string.end.dosbatch
-              pop: true
-            - include: inside_command_set
-            - include: command_set_group
-            - include: variables
-        - include: command_set_group
+      - match: '\n'
```
wbond: Shouldn't this block be indented one move level?

Briles: Should be OK now.

PR #464, opened 2016-06-08
[C++] Fixing incorrect XOR character

jbw3:
The XOR character in the operator overload regex is incorrect. It looks like a caret, but it's not.


kleinfreund:
Wow. This is … interesting. Any idea how this would happen? 

For reference: http://www.fileformat.info/info/unicode/char/02c6/index.htm

wbond:
@kleinfreund Probably some issue related to the Unicode input mode I was using in OS X when I wrote this.

PR #466, opened 2016-06-08
[Ruby] Support code on same line as heredoc begin token

wbond:
This will be merged right before we ship 3115 since it uses new functionality

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -1345,97 +1345,135 @@ contexts:
     # heredoc with embedded HTML and indented terminator
     - match: '(<<[-~]"?((?:[_\w]+_|)HTML)\b"?)'
       scope: punctuation.definition.string.begin.ruby
-      push:
-        - meta_scope: string.unquoted.embedded.html.ruby
-        - meta_content_scope: text.html.embedded.ruby
-        - match: ^\s*\2$
-          scope: punctuation.definition.string.end.ruby
-          pop: true
-        - include: scope:text.html.basic
-        - include: interpolated-ruby
-        - include: escaped-char
+      push: [heredoc-html, trailing-heredoc-start]
     # heredoc with embedded SQL and indented terminator
     - match: '(<<[-~]"?((?:[_\w]+_|)SQL)\b"?)'
       scope: punctuation.definition.string.begin.ruby
-      push:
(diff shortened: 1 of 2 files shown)


FichteFoll:
`clear_scopes: 2`

What does this do exactly?

wbond:
It removes N scopes from the scope name stack. Or you can pass `true` to clear the stack altogether, as used in #467.

In this case, it clears the two scopes added by the heredoc open token so that the code on the remainder of the line doesn't get highlighted like it is part of a string.

This could be useful for string interpolation, since we can remove the general string highlighting from the interpolated part.

FichteFoll:
I assume this works like a meta scope is only valid for the current context?

wbond:
Yes, this functions like `meta_scope` in that it only applies to the current context.

Ruby/Ruby.sublime-syntax
```diff
@@ -1345,97 +1345,135 @@ contexts:
...
+  #  - meta_scope: string.unquoted.embedded.shell.ruby
+  #  - meta_content_scope: text.shell.embedded.ruby
+  #  - match: ^\s*\2$
+  #    scope: punctuation.definition.string.end.ruby
+  #    pop: true
+  #  - include: Shell-Unix-Generic.sublime-syntax
+  #  - include: interpolated-ruby
+  #  - include: escaped-char
+
+  # This prevents clear_scopes from applying to the push token
```
trishume: @wbond what does this mean? Why is this necessary? This context seems useless to me.

I'm trying to implement `clear_scopes` in https://github.com/trishume/syntect and the fact that this context looks useless to me suggests I might be misunderstanding the semantics of either `clear_scopes` or push/pop operations. 

If you could explain to me why this is necessary I would really appreciate it. Also, what is a "push token" and why would `clear_scopes` apply to it?

wbond: `clear_scopes` is implemented in the same way as `meta_scope` in that it applies to the match the pushed, and to the match that pops.

The match that pushed is what I mean by "push token".

Without this, `clear_scopes` would clear the scope from the opening heredoc token, but we don't want that. I decided not to implement `clear_scopes_content` since this method can be used to emulate it.

trishume: Ok that makes sense, and in retrospect is explained in the docs, thanks for clearing up my confusion. Thanks!

PR #467, opened 2016-06-08
[PHP] Use clear_scopes to implement proper HTML nesting inside of PHP blocks

wbond:
This effectively reverts cd7f8c24ef8c785fc9ff01c1c157beadb35278d8, and will be merged right before we ship 3115 since it uses new functionality.

It has the side effect of also allowing literal HTML output inside of anonymous class methods and closures.

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -13,12 +13,6 @@ contexts:
   main:
     - include: statements
 
-  embedded:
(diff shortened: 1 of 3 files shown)


jfcherng:
Is `clear_scopes` a new API in 3115?

wbond:
Yes, it is a new feature of `.sublime-syntax` that will be part of 3115.

swashata:
Will this fix #478 ?

PR #469, opened 2016-06-09
[Lisp] Implemented operators & block comments

Briles:
- Implemented operators & block comments
- Scoped parentheses
- Converted `captures: 0` to `scope`
- Added tests


wbond:
Are lisp function names typically written in all uppercase? If not, it seems funny to convert them in the regexes to uppercase.

Briles:
Typically they are written in lowercase so I have converted them back to lowercase. I had written them in uppercase as I find it easier to scan (especially when using word boundaries).

wbond:
Thanks for this, especially starting the syntax tests!

PR #477, opened 2016-06-16
Update Java import and package keywords

ismell:
* Add a scope with .static for static imports. This allows scripts to extract static imports easily.
* Use scopes to simplify the regex.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -8,19 +8,42 @@ file_extensions:
 scope: source.java
 contexts:
   main:
-    - match: '^\s*(package)\b(?:\s*([^ ;$]+)\s*(;)?)?'
-      scope: meta.package.java
-      captures:
-        1: keyword.other.package.java
-        2: storage.modifier.package.java
-        3: punctuation.terminator.java
-    - match: '^\s*(import)\b(?:\s*([^ ;$]+)\s*(;)?)?'
-      scope: meta.import.java
-      captures:
-        1: keyword.other.import.java
-        2: storage.modifier.import.java
-        3: punctuation.terminator.java
+    - include: package
+    - include: import
     - include: code
+  package:
+    - match: '\bpackage\b'
+      scope: keyword.other.package.java
+      push:
+        - meta_scope: meta.package.java
+        - match: '[\w\.]+'
+          scope: support.other.package.java
+        - match: ;
+          scope: punctuation.terminator.java
+          pop: true
+        - match: (?=\n)
+          pop: true
+  import:
+    - match: \bimport\b
+      scope: keyword.other.import.java
+      push:
+        - meta_scope: meta.import.java
+        - match: \bstatic\b
+          scope: storage.modifier.static.java
+          push:
+            - match: '[\w\.]+'
+              scope: support.function.import.java
+            - match: (?=;)
+              pop: true
+            - match: (?=\n)
+              pop: true
+        - match: '[\w\.]+'
+          scope: support.class.import.java
+        - match: ;
+          scope: punctuation.terminator.java
+          pop: true
+        - match: (?=\n)
+          pop: true
   all-types:
     - include: primitive-arrays
     - include: primitive-types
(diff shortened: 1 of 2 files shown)


Java/Java.sublime-syntax
```diff
@@ -8,19 +8,42 @@ file_extensions:
...
+  import:
+    - match: \bimport\b
+      scope: keyword.control.import.java
+      push:
+        - meta_scope: meta.import.java
+        - match: \bstatic\b
+          scope: storage.modifier.static.java
+          push:
+            - match: '[\w\.]+'
+              scope: keyword.other.static.import.java
```
wbond: I don't believe this is a good usage of the scope `keyword.other`. That should only be applied to real keywords.

Perhaps something like `variable.other.import`? @keith-hall or @FichteFoll do you have any opinion on this? We could use a similar scope in Python and other languages also.

Java/Java.sublime-syntax
```diff
@@ -8,19 +8,42 @@ file_extensions:
...
+    - include: package
+    - include: import
     - include: code
+  package:
+    - match: '\bpackage\b'
+      scope: keyword.control.package.java
+      push:
+        - meta_scope: meta.package.java
+        - match: '[\w\.]+'
+          scope: keyword.other.package.java
```
wbond: Same comment about `keyword.other` applies here also.

Java/Java.sublime-syntax
```diff
@@ -8,19 +8,42 @@ file_extensions:
...
+          scope: storage.modifier.static.java
+          push:
+            - match: '[\w\.]+'
+              scope: keyword.other.static.import.java
+            - match: (?=;)
+              pop: true
+            - match: (?=\n)
+              pop: true
+        - match: '[\w\.]+'
+          scope: keyword.other.import.java
```
wbond: Same comment about `keyword.other` applies here also.

Java/Java.sublime-syntax
ismell: I'm open for suggestions. I didn't like the original storage.modifier.package.java because it's not a modifier. It's also not a variable either. It really is a class path. Maybe it should be `storage.type.import.java` ?

wbond: So while it isn't a variable per se, the textmate scopes have traditionally used `variable.` for various identifiers. One of the more popular ones is `variable.function` which is applied to the identifier in a function/method call.

`storage.type` is generally used for built-in data types such as `int`. Often user-defined classes get `support.class` and namespaces `support.namespace` (see PHP for examples). I'm not up on my Java, are imports always classes?

Perhaps we need a new, top-level scope such as `identifier`? Perhaps `identifier.function`, `identifier.generic.path`, `identifier.generic`, `identifier.import`, etc? In that case we'd probably leave `variable.parameter` and `variable.other.readwrite` as-is (or perhaps `variable.other.readwrite` becomes `variable.generic`?).

In many languages we won't really be able to identify all variables reliably, so we may want to recommend color scheme authors not apply highlighting to `identifier.generic`?

ismell: In Java imports are always fully qualified class paths. e.g., `foo.bar.SomeClass`. So maybe `support.class.import` makes sense. You can either add the class as an import and reference it as `SomeClass` in the file, or skip adding the import and use the fully qualified path.

Given the above, maybe the package path be `support.namespace`?

Static imports are always fully qualified static methods. e.g., `foo.bar.SomeClass.myStaticMethod`. They are not actual invocations so not sure if `variable.function` makes sense. How about `support.function.import`?

FichteFoll: I guess scoping the import identifiers as what they are makes sense for Java. That is, classes get the user-defined class scope and static methods get the function scope, whatever those are.

ismell:
PTAL

ismell:
ping

wbond:
I'm working on other things right at the moment, but I will review this again the next time I'm working on the packages.

wbond:
Awesome, thanks for your work on this!

ismell:
I got way more to come :)

PR #482, opened 2016-06-17
[Lisp] Update built-in variables, constants, functions, & keywords

Briles:
Updates the variables, constants, functions, & control keywords


wbond:
I'm not familiar with the Lisp community, but I do know there are different flavors of lisp.

Do these apply to a specific flavor, or are they widely supported?

Briles:
All of my Lisp changes so far have been aimed at supporting Common Lisp. 

I don't have any experience with Lisp, so I can't say how much carryover there is between dialects. Some googling reveals dialects do have different function, symbol, macro names, and even built-in types.

michaelblyons:
@Briles Did you scrape the builtins from a single source page, or were there several pages and then you sorted them? I'm having a go at some changes at #2387, but the last time I looked at Lisp was when MIT released their CS curricula in 2004 or so.

Briles:
@michaelblyons It's been so long since I've seen this so I don't remember much but I'm pretty sure I used this copy of [Common Lisp the Language, 2nd Edition](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/index.html) from a CMU repository.

PR #483, opened 2016-06-17
[CSS] Add document, supports, & counter-style at-rules

Briles:
- Adds the at-rules `@document`, `@supports`, & `@counter-style`
- Re-organizes at-rule constructs
- Adds `@document` supporting functions: `domain()`, `regexp()`, & `url-prefix()`

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -74,25 +74,21 @@ contexts:
   main:
     - include: comment-block
     - include: selector
-    - include: charset
-    - include: import
-    - include: font-face
-    - include: media
-    - include: custom-media
-    - include: keyframes
-    - include: namespace
-    - include: page
+    - include: at-rules
     - include: property-list
 
-  charset:
-    - match: \s*((@)charset\b)\s*
-      captures:
-        1: keyword.control.at-rule.charset.css
-        2: punctuation.definition.keyword.css
-      push:
-        - meta_scope: meta.at-rule.charset.css
-        - include: at-rule-punctuation
-        - include: literal-string
+  at-rules:
+    - include: at-charset
+    - include: at-counter-style
+    - include: at-custom-media
+    - include: at-document
+    - include: at-font-face
+    - include: at-import
+    - include: at-keyframes
+    - include: at-media
+    - include: at-namespace
+    - include: at-page
+    - include: at-supports
 
   # When including `color-values` and `color-adjuster-functions`, make sure it is
   # included after the color adjustors to prevent `color-values` from consuming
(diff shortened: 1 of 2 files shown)


CSS/CSS.sublime-syntax
```diff
@@ -125,7 +121,31 @@ contexts:
...
+
+  # @counter-style
+  # https://drafts.csswg.org/css-counter-styles-3/#the-counter-style-rule
+  at-counter-style:
+    - match: \s*((@)counter-style\b)\s+(?:(?i:\b(decimal|none)\b)|({{ident}}))\s*(?={)
+      captures:
+        1: keyword.control.at-rule.counter-style.css
+        2: punctuation.definition.keyword.css
+        3: invalid.illegal.counter-style-name.css
+        4: entity.other.counter-style-name.css
```
Briles: Would this scope `entity.other.counter-style-name.css`, be better as `entity.name.counter-style.css`?

Any changes would also apply to [entity.other.animation-name.css](https://github.com/Briles/Packages/blob/fdf35f282921b3532db9b2359820655c70307c99/CSS/CSS.sublime-syntax#L218)

CSS/CSS.sublime-syntax
```diff
@@ -603,14 +683,16 @@ contexts:
...
             )\b
           scope: support.type.property-name.css
     - match: (:)\s*
       captures:
         1: punctuation.separator.key-value.css
       push:
-        - meta_scope: meta.property-value.css
-        - match: '\s*(;|(?=\}))'
+        - meta_content_scope: meta.property-value.css
+        - match: '\s*(;)|(?=[})])'
```
Briles: I added `)` to the lookahead here to keep things simple for `@supports`. This now means that rules can be improperly terminated by `)` like so:

```css
.c {
    background: none)background: none
}
```

Should I duplicate the necessary [`rule-list-body`](https://github.com/Briles/Packages/blob/fdf35f282921b3532db9b2359820655c70307c99/CSS/CSS.sublime-syntax#L598) constructs so this improper termination doesn't happen?

aziz:
This is great that you're working on bringing new CSS features into the syntax. 
I would love to see more of these, specially the [CSS Nesting](http://tabatkins.github.io/specs/css-nesting/) syntax highlighting. It would be great if you can add more features to have better support for all [CSS Next](http://tabatkins.github.io/specs/css-nesting/) features

wbond:
Sorry for the delay on reviewing these changes

CSS/CSS.sublime-syntax
wbond: I'm inclined to say use an `entity.other` scope and not `entity.name`. The `entity.name` scopes (other than `entity.name.tag` and `entity.name.section`) are for types and data structures.

There will be some docs coming out with the next build that covers scope names some more. However, the recommendation is that color schemes highlight `entity.name` all the same except for the tag and section variations. Generally, these should all be in the symbol list and index also.

`entity.other`, however, is currently used for attribute names and class names used when dealing with inheritance. In a perfect world, I'd love to see the tag and section scopes moved into `entity.other`, but that would break pretty much every color scheme ever.

CSS/CSS.sublime-syntax
wbond: I'm not too concerned about it breaking highlighting since it isn't valid anyway. If we tried to gracefully handle each language and the common way typos would break syntax we'd have very long and complicated syntaxes.

CSS/CSS.sublime-syntax
```diff
@@ -603,14 +683,16 @@ contexts:
...
+              | symbols|speak-as
             )\b
           scope: support.type.property-name.css
     - match: (:)\s*
       captures:
         1: punctuation.separator.key-value.css
       push:
-        - meta_scope: meta.property-value.css
-        - match: '\s*(;|(?=\}))'
+        - meta_content_scope: meta.property-value.css
```
philipbordallo: This change seems to have broken completions, see #690

PR #484, opened 2016-06-20
[Pascal] Fixed one line comments eating keywords from next line

urraka:
One line comments grab the final `\n` which is apparently needed for matching a `\b` if there's a keyword right on the next line (with no trailing space). This is visually annoying and breaks jumping to symbols for functions/procedures that have a comment right before them.

### DIFF
--- a/Pascal/Pascal.sublime-syntax
+++ b/Pascal/Pascal.sublime-syntax
@@ -23,32 +23,22 @@ contexts:
         2: entity.name.function.pascal
     - match: '\b((0(x|X)[0-9a-fA-F]*)|(([0-9]+\.?[0-9]*)|(\.[0-9]+))((e|E)(\+|-)?[0-9]+)?)(L|l|UL|ul|u|U|F|f|ll|LL|ull|ULL)?\b'
       scope: constant.numeric.pascal
-    - match: '(^[ \t]+)?(?=--)'
+    - match: '(^[ \t]+)?(--)'
       captures:
         1: punctuation.whitespace.comment.leading.pascal
+        2: punctuation.definition.comment.pascal
       push:
-        - match: (?!\G)
+        - meta_scope: comment.line.double-dash.pascal.one
+        - match: \n
           pop: true
-        - match: "--"
-          captures:
-            0: punctuation.definition.comment.pascal
-          push:
-            - meta_scope: comment.line.double-dash.pascal.one
-            - match: '(?=\n)'
-              pop: true
-    - match: '(^[ \t]+)?(?=//)'
+    - match: '(^[ \t]+)?(//)'
       captures:
         1: punctuation.whitespace.comment.leading.pascal
+        2: punctuation.definition.comment.pascal
       push:
-        - match: (?!\G)
+        - meta_scope: comment.line.double-slash.pascal.two
+        - match: \n
           pop: true
-        - match: //
-          captures:
-            0: punctuation.definition.comment.pascal
-          push:
-            - meta_scope: comment.line.double-slash.pascal.two
-            - match: '(?=\n)'
-              pop: true
     - match: \(\*
       captures:
         0: punctuation.definition.comment.pascal
--- a/Pascal/Pascal.sublime-syntax
+++ b/Pascal/Pascal.sublime-syntax
@@ -34,7 +34,7 @@ contexts:
             0: punctuation.definition.comment.pascal
           push:
             - meta_scope: comment.line.double-dash.pascal.one
-            - match: \n
+            - match: '(?=\n)'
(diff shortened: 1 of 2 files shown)


keith-hall:
I think it would be better to change the keywords `\b` to `(?:\b|^)` so that the `\n` would still be scoped as a comment, for consistency with the other syntaxes and to prevent completions etc when typing a comment

keith-hall:
although, realistically, it seems like a bug that the start of a line doesn't count as a word boundary...

urraka:
I tried your suggestion but it doesn't seem to work. I'm not very familiar with how this works but feels like after parsing the comment line (including the `\n`) it just doesn't know that it's the start of a line anymore (I'm pretty sure that would count as `\b` aswell)

keith-hall:
I've tested with a separate, very simple minimal example, and it seems to work alright...

syntax definition:

    %YAML 1.2
    ---
    # See http://www.sublimetext.com/docs/3/syntax.html
    scope: source.test_word_boundary_after_newline
    contexts:
      main:
        - match: '#'
          push:
            - meta_scope: comment
            - match: '\n'
              pop: true
        - match: '\btest\b'
          scope: keyword

test file:

    # testing
    test


can you give an example of some Pascal code that isn't highlighted correctly please? Maybe even create a syntax test file?

urraka:
You can test this with the default Pascal syntax file:

```
// comment
procedure proc;
begin
	// another comment
end;
```

The keywords `procedure` and `end` are not recognized.

Pascal/Pascal.sublime-syntax
```diff
@@ -47,7 +47,7 @@ contexts:
             0: punctuation.definition.comment.pascal
           push:
```
keith-hall: changing this `push` to a `set` will fix the problem, no need to stop capturing `\n` as part of a comment

keith-hall: same here

keith-hall: same here. Not sure how necessary the whole `(?!\G)` is, maybe it can be removed?

urraka: Yep, that works a lot better. Maybe close this and commit that separately? There are 2 `push` that need to be replaced in lines `35` & `48` (for comments starting with `--` and `//`)

urraka: I'm not sure what the `\G` is for. Maybe not needed at all, but as I said, not very familiar with how this all works

urraka: From what I tested `(?!\G)` is not needed as long as we keep using `set` instead of `push`

keith-hall: one solution here would be to keep your `\n` -> `(?=\n)` changes, and also to change:

    - match: (?!\G)
      pop: true

to

    - match: \n
      scope: comment.line.double-slash.pascal
      pop: true

I'm not sure why Pascal is different from the other syntaxes in that it scopes the whitespace at the beginning of a line, before a comment as `punctuation.whitespace.comment.leading.pascal`, where other syntaxes don't scope it at all. It could be simplified quite a bit if this was removed and thus made consistent with the other syntaxes. What do you think @wbond ?

wbond: @keith-hall A bunch of the Textmate languages would match from the beginning of the line with whitespace before a comment line start. I feel like this has to do with how it's regex engine matched tokens, although I am far from an expert in that.

As we've been rewriting the syntaxes, I've been removing things such as this whenever possible.

wbond:
There is no test for this change, and we've actually been moving syntaxes *towards* consuming the newline via the comment context because it prevents the auto-complete window from appearing while writing a comment, and snippets from appearing while writing a comment.

urraka:
I ended up simplifying the whole thing. Basically, I removed the lookaheads for the `--` & `//` symbols on the first match. I'm not sure why it used a lookahead to begin with but seems fine without it.

There was no syntax test for Pascal so I added a simple one for this particular issue (it fails with the old syntax file but passes with the new one).

wbond:
@urraka Sorry for the delay in getting back to you. This change is looking good.

Could you just add an assertion or two to ensure that the comments are getting the correct scopes? It seems the assertions currently are just testing the keywords.

urraka:
@wbond Initially I had these assertions there but it just breaks the test for this particular bug: an assertion line between the pascal comment and the next keyword will make the keyword scope work just fine (with the non-fixed syntax), whearas without the assertion it fails.

Anyway, I'll go ahead and add some comment assertions separately.

wbond:
Perfect!

PR #485, opened 2016-06-21
[Scala] Added tests for existing Scala syntax support

djspiewak:
Doesn't change anything other than codifying most of what is currently there.  Some things aren't tested (such as the invalid string newline thing) since the test framework isn't quite able to handle it (so far as I can tell).


wbond:
Thank you for the tests, they are very helpful!

PR #486, opened 2016-06-21
[Scala] Symbolic method name support

djspiewak:
Fixes syntax highlighting for symbolic methods (e.g. `def +`) and methods with backticks.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -85,7 +85,7 @@ contexts:
     - match: |-
         (?x)
         \b(def)\s+
-        (([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`|[^\w\[\(\:\_\s]+)
+        (_*([a-zA-Z$][a-zA-Z0-9$]*|[^a-zA-Z0-9$`\s\(\)\[\]\{\};\.]+)(_+([a-zA-Z$][a-zA-Z0-9$]*|[^a-zA-Z0-9$`\s\(\)\[\]\{\};\._]+)?)*|`[^`]+`|[^\w\[\(\:\_\s]+)
       captures:
         1: keyword.declaration.scala
         2: entity.name.function.declaration
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -20,6 +20,30 @@ def foo(a: Int, b: Bar): Baz = 42
 //                       ^^^ entity.name.class
 //                             ^^ constant.numeric.scala
 
+   def +(a: Int)
+// ^^^ keyword.declaration.scala
+//     ^ entity.name.function.declaration
+
+   def `this is a test`(a: Int)
+// ^^^ keyword.declaration.scala
+//     ^^^^^^^^^^^^^^^^ entity.name.function.declaration
+
+   def ::(a: Int)
+// ^^^ keyword.declaration.scala
+//     ^^ entity.name.function.declaration
+
+   def foo_+(a: Int)
+// ^^^ keyword.declaration.scala
+//     ^^^^^ entity.name.function.declaration
+
+   def foo42_+(a: Int)
+// ^^^ keyword.declaration.scala
+//     ^^^^^^^ entity.name.function.declaration
+
+   def foo42_+_abc(a: Int)
+// ^^^ keyword.declaration.scala
+//     ^^^^^^^^^^^ entity.name.function.declaration
+
 val foo: Unit
 //^ keyword.declaration.stable.scala
 //  ^^^ entity.name.val.declaration
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -1,0 +1,224 @@
+// SYNTAX TEST "Packages/Scala/Scala.sublime-syntax"
+// <- source.scala comment.line.double-slash.scala
+
+package fubar
+// ^^^^ keyword.other.scoping
+//      ^^^^^ entity.name.package.scala
+
+import fubar.{Unit, Foo}
+// ^^^ keyword.other.import
+// <- meta.import.scala
+//     ^^^^^ variable.package.scala
+//            ^^^^ variable.import.scala
+
+def foo(a: Int, b: Bar): Baz = 42
+//^ keyword.declaration.scala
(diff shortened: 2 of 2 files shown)


Scala/Scala.sublime-syntax
```diff
@@ -85,7 +85,7 @@ contexts:
     - match: |-
         (?x)
         \b(def)\s+
-        (([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`|[^\w\[\(\:\_\s]+)
+        (_*([a-zA-Z$][a-zA-Z0-9$]*|[^a-zA-Z0-9$`\s\(\)\[\]\{\};\.]+)(_+([a-zA-Z$][a-zA-Z0-9$]*|[^a-zA-Z0-9$`\s\(\)\[\]\{\};\._]+)?)*|`[^`]+`|[^\w\[\(\:\_\s]+)
```
mkhl: This regex is in extended mode anyway, does it make sense to expand this a bit? Fully expanded, it looks like this:

```
(
	_*
	(
		[a-zA-Z$][a-zA-Z0-9$]*
	|
		[^a-zA-Z0-9$`\s\(\)\[\]\{\};\.]+
	)
	(
		_+
		(
			[a-zA-Z$][a-zA-Z0-9$]*
		|
			[^a-zA-Z0-9$`\s\(\)\[\]\{\};\._]+
		)?
	)*
|
	`[^`]+`
|
	[^\w\[\(\:\_\s]+
)
```

(…which seems a bit too verbose, there's probably sensible middle ground.)

Referring to that, the matchers at lines 6 and 13 (starting with `[^a-z`…) differ in that the former permits underscores (although that difference probably won't matter in practice).

Excluding underscores lets you write those classes as 
```
[^\w$`\s\(\)\[\]\{\};\.]
```

With this support for symbolic method names, can we drop the last branch of the match? It allows illegal method names like `]]` or `.;` and I can't think of any symbolic names that aren't handled by the matchers you added.

djspiewak: I do think that would make things more readable.  With that said, I'm planning on making a pass over this whole thing and collapsing all of the "matching the same thing, but not actually the same regex" situations that we have in the current Scala mode using variables.  When I do that, I will format the regexes in a more readable way.  However, I don't want to actually make that change until all of my PRs have landed, since I have a *lot* of pending work right now and a common-regex collapse cuts across all of it.

mkhl:
> Regarding the elimination of underscore forbidding…

Just to be clear, I’m in favour of forbidding underscores inside the large character classes. I talked about "excluding" them because those classes are negated. I believe that matching underscores *only* in the `_*` and `_+` portions is unambiguous and faster.

> Also the regex as written forbids certain legal method names (e.g. `foo_`)

I don’t think it does; I was wondering why the question mark was there on line 14 (`)?`), but that is precisely what permits trailing underscores.

djspiewak:
> Just to be clear, I’m in favour of forbidding underscores inside the large character classes. I talked about "excluding" them because those classes are negated. I believe that matching underscores only in the `_*` and `_+` portions is unambiguous and faster.

Oh…  Hmm…  This bears experimentation.  I wonder if it's worth doing that in this PR, or waiting until my PR where I unify all the disparate regexes?  I'm tempted to say the latter, but I'm willing to be persuaded otherwise.

> I don’t think it does; I was wondering why the question mark was there on line 14 ()?), but that is precisely what permits trailing underscores.

You're right.  I was thinking of another case:

foo_2

That's with the current regex.  So there's a bit of work to do.

mkhl:
> Oh… Hmm… This bears experimentation. I wonder if it's worth doing that in this PR, or waiting until my PR where I unify all the disparate regexes? I'm tempted to say the latter, but I'm willing to be persuaded otherwise.

Yeah, probably the latter. I guess I’m trying to skip ahead—as noted elsewhere I want these changes in the TextMate bundle 😄

> `def foo_2()`

Ah, I see. I also found a few more illegal names it permits, like `+_+` and `foo_+_bar`.

I assume you’ve seen [Dale’s attempt to simplify the syntax spec](https://gist.github.com/dwijnand/3fcce60c8ac68abc0953fbb62acbda24)?
Following that I arrived at

```
(	`[^`]+`								# string literal
|	[^\w$`\s\(\)\[\]\{\};\.]+			# operator
|	[a-zA-Z$_][a-zA-Z0-9$_]*			# identifier
	((?<=_)[^\w$`\s\(\)\[\]\{\};\.])?	# trailing operator
)
```

Does that seem reasonable to you?

djspiewak:
It looks to me like that regex would reject the following:

```scala
def +_abc
```

I'm also dubious of using lookahead to catch the underscore.  It should be possible to encode this pattern without using backtracking.

mkhl:
> It looks to me like that regex would reject the following: `def +_abc`

So does the compiler:

```
error: '=' expected but identifier found.
def +_abc = {
     ^
```

djspiewak:
> So does the compiler

So it does.  :-)

Scala/Scala.sublime-syntax
wbond: Can we reformat this regex now for the sake of review? It appears to me there is a new capture group added, but the current formatting makes it hard to check.

wbond: In the interest of continuing to review these Scala PRs, I'll reformat and deal with any changes necessary.

djspiewak: @wbond I intend to take a pass over all of the regexes in the Scala mode and both unify and reformat them.  I haven't done it yet because that change touches *all* of my PRs, and I don't want to do repeated work.

wbond: Right, but I need to review them before merging. I'll just deal with the conflicts and resolve them myself, no problem.

wbond:
When pulling these down for testing I ended up having to rebase, which wiped the merge commit. The two unique commits are in as 80e43e65fdd763254feb7a0783f43056d65546ed and 675d28736226b805bd7ae9697a45ad0492cb3205.

I also fixed the identifier regex with info from http://www.scala-lang.org/files/archive/spec/2.11/01-lexical-syntax.html in 1aeadb4c0f7430f88947de39a892a629067296d4 and 06e9f9cf5c7fb33a4f66056db80903ea224d0ef1. It uses variables and character class intersection for the most part. `foo_2` is now accepted. Additionally, `foo_+_bar` is rejected.

PR #487, opened 2016-06-21
[Scala] Avoid highlighting colon prefixes as parameters

djspiewak:
…mostly.

This is still a bit of a work in progress, but it's still a significant improvement over what is there *now*.  The semantic here is to look for enclosing parentheses.  So still a few false positives, but far fewer than before.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -138,7 +138,7 @@ contexts:
           pop: true
         - match: '([a-zA-Z$_][a-zA-Z0-9$_]*)\s*:\s*'
           captures:
-            1: entity.name.variable variable.parameter
+            1: entity.name.parameter variable.parameter
         - include: main
     - match: '(?=[:\{=\n])'
       pop: true
@@ -169,7 +169,7 @@ contexts:
           pop: true
         - match: '([a-zA-Z$_][a-zA-Z0-9$_]*)\s*:\s*'
           captures:
-            1: entity.name.variable variable.parameter
+            1: entity.name.parameter variable.parameter
         - include: main
     - match: '(?=([\{\n]|extends))'
       pop: true
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -20,8 +20,8 @@ contexts:
     - include: constants
     - include: char-literal
     - include: scala-symbol
-    - include: empty-parentheses
     - include: parameter-list
+    - include: empty-parentheses
     - include: qualifiedClassName
     - include: xml-literal
   block-comments:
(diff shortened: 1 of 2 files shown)


Scala/Scala.sublime-syntax
```diff
@@ -138,7 +138,7 @@ contexts:
           pop: true
         - match: '([a-zA-Z$_][a-zA-Z0-9$_]*)\s*:\s*'
           captures:
-            1: entity.name.variable variable.parameter
+            1: entity.name.parameter variable.parameter
```
wbond: `entity.name` should be used here at all. That is for the definition of things like classes, function, etc. `variable.parameter` is the correct scope to use. I just won't cherry-pick this commit, and instead remove the `entity.name.variable` part.

wbond:
This unique commits from this PR were cherry-picked as:

 - 32a78f3a3aa0854d5eeba990b22a0dd59a2a61ac
 - a58581cc47c7decb0a7e5e7ad9df8528d0a58b90
 - f7ca210a74581a81a4f73814480f690e980997de

I then fixed the parameter scope names in e0057d48f85fb405de7166f55d4f9feec00b470e.

Scala/Scala.sublime-syntax
djspiewak: This is one of those weird interactions between my PRs.  Once all the PRs land, I'll go back and make sure that the interaction actually worked out.

Scala does not make a distinction between entities that are parameters and entities that are objects or variables, and the distinction between these two and entities that are functions is microscopic.  For that reason, I would rather actually use `entity.name.parameter`.  This is more strongly seen in the pattern matching PR.  `variable.parameter` is fine, but it still needs to be considered by symbol indexing.

wbond: Indexing every parameter in a codebase feels wrong, but then again, I don't know Scala at all.

wbond: However, for code highlighting, it seems it would be better to leave it as `variable.parameter` and instead add a `.tmPreferences` file that indicates `variable.parameter` should be indexed.

djspiewak: I'm fine with that.

djspiewak: > Indexing every parameter in a codebase feels wrong, but then again, I don't know Scala at all.

It is a little noisy, in practice, but not so noisy that I feel it's necessarily wrong.  Are function parameters indexed in the Java, Ruby, Python or Haskell modes?

wbond: No, function parameters are never indexed to my knowledge. I've recently worked on a whole bunch, including C, C++, Go, Rust, Python, Ruby, Java, PHP, TCL and JavaScript. Parameters were never indexed in any of them before I started, or now.

It may just be that you make the decision to add a `.tmPreferences` file to you user package that turns on indexing/symbol list for the scope `source variable.parameter`.

djspiewak: Ok, so let's not index function parameters then.  The slightly-weird part about that though is the indexer *will* consider variables in pattern matching or for-comprehensions (once #491 is merged), and those elements are highlighted identically to function parameters.  In practice, indexing these pattern variables has been enormously useful (indexing function parameters hasn't been quite as vital), so I'd definitely like to keep that functionality.

PR #488, opened 2016-06-21
[Scala] Keyword highlighting

djspiewak:
The Scala mode does this weird thing where it only highlights certain keywords in context (e.g. `class` is only highlighted in `class Foo`, not when by itself).  This is deceptive since these keywords aren't valid identifiers.  This PR adds fallback highlighting support for those keywords.

Also adds support for highlighting `macro` and `???`, the latter of which isn't actually a keyword, but most people treat it as if it is.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -158,7 +158,7 @@ contexts:
   keywords:
     - match: \b(return|throw)\b
       scope: keyword.control.flow.jump.scala
-    - match: \b(else|if|do|while|for|yield|match|case)\b
+    - match: \b(else|if|do|while|for|yield|match|case|macro)\b
       scope: keyword.control.flow.scala
     - match: \b(catch|finally|try)\b
       scope: keyword.control.exception.scala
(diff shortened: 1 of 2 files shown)


NeQuissimus:
LGTM 

Curious about the `\b` not working with `??? `

michaelblyons:
`?` isn't a "word character." You'd need it to adjoin a word character for the `\b` (word boundary) to match: 

```
What?
   /\ `\b` matches here.
? ? ??.
   (but nowhere on this line)
```

Scala/Scala.sublime-syntax
```diff
@@ -9,12 +9,12 @@ scope: source.scala
 contexts:
   main:
     - include: storage-modifiers
-    - include: keywords
     - include: declarations
     - include: inheritance
     - include: imports
     - include: comments
     - include: block-comments
+    - include: keywords
```
mkhl: Adding the new keywords this high can prevent other patterns from matching, e.g. I believe your matching of `new` would prevent `initialization` from matching its expected `new` keyword.

I’d also recommend keeping the same scope for keywords that you’re matching without context now, e.g. `type` is `keyword.declaration.scala` in `declarations`.

It’s a pity the scopes have diverged from the TextMate grammar so much…

wbond:
I added an alternative implementation of `late-keywords` in f83d366c91a338741635bd4bdfb3be2b5722c09b, along with `macro` and `???`.

I cherry-picked the tests from 3a89da2292cd47ed8bb9bfc847498da098e463ea.

I fixed various test failures in bce1507fc653cb00f9acc22eb80128d4c845de35.

PR #490, opened 2016-06-21
Fix scope mismatch between Ruby syntax and Haml

trishume:
I found this while investigating all the scopes in the repo where order mattered and this was the only one. Turns out it is a mistake and this should match the one in the Ruby syntax file.

So fun fact: in all the default syntaxes order of the scope doesn't matter when checking equality. Not sure if this is true for all prefixes yet though.

@wbond

### DIFF
--- a/Rails/Ruby Haml.sublime-syntax
+++ b/Rails/Ruby Haml.sublime-syntax
@@ -72,7 +72,7 @@ contexts:
         - match: '((do|\{)( \|[^|]+\|)?)$|$|^(?!.*\|\s*$)'
           captures:
             1: source.ruby.embedded.html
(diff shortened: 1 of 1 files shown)


wbond:
A number of the scopes from the initial Textmate `.tmLanguage` import were in a funky order. As we've been improving the syntaxes, I've been making a point of cleaning up instances where the language is not listed last.

However, there are some examples where a variation is tacked onto the end, as it with `text.html`. In this case it allows writing color schemes and plugins that target `text.html` without having to worry about what flavor it is (`text.html.basic`, `text.html.asp`, `text.html.ruby`, etc).

PR #491, opened 2016-06-21
[Scala] Pattern match highlighting for case, val and for

djspiewak:
Scala's variable definition syntax is quite rich, and Sublime doesn't (currently) handle it very well.  Additionally, several special syntactic forms are available in both `val` declarations and `case` pattern matches (the syntax is actually identical) which are not highlighted at all by Sublime.  Finally, complex patterns in `val` declarations and *any* patterns in `case` pattern matches are not currently considered for "go to definition", which is annoying.

This PR fixes all that.

- case example
- val examples

The orange italic parameters in the above are both considered for symbol matching, which enables go to definition jumping into pattern match extractors, as well as provides more accurate auto-complete.

There is a slight cost here.  Specifically, when you type `case`, you open up a new scope which highlights the *rest of your file* differently.  This scope is closed by either `=>` or `⇒`, but it does make the process of actually *typing* Scala in a left-to-right fashion a bit different.  If you use the `case`<kbd>TAB</kbd> snippet, this problem is resolved.  The additional wrinkle is if you write a syntactically *invalid* pattern match, you may see some weird syntax highlighting.  For example:

```scala
case f: Int => Boolean :: _ => f
```

This is not valid Scala, but it's a relatively easy mistake to make.  What will happen here is the first `=>` will terminate the pattern match scope and the `_` character will not be correctly highlighted.  I don't feel too bad about this since Scala will actually not compile the above, and the fix for the syntax highlighting is identical to the fix for the Scala parser: wrap it in parentheses:

```scala
case (f: Int => Boolean) :: _ => f
```

~~This PR does not currently address the problem of `for`-comprehensions, which are another instance of pattern matching syntax, but unfortunately one which is far more syntactically complex.  Specifically, I would need to do some sort of backtracking scope push on the `<-` and `←` characters, defining everything from the start of the line to that character to be in pattern match scope while within a `for` block, but any line which does *not* contain these characters gets parsed normally.  As far as I can tell, `.sublime-syntax` does not have a way of expressing this sort of thing.~~

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -433,6 +433,10 @@ contexts:
         - meta_scope: source.scala
         - match: '`'
           pop: true
+    - match: '\b[a-z][a-zA-Z0-9_]*\.'
+      scope: source.scala
+    - match: '\.[a-z][a-zA-Z0-9_]*\b'
+      scope: source.scala
     - match: '\b[a-z][a-zA-Z0-9_]*\b'
       scope: entity.name.parameter variable.parameter
     - match: '\b[A-Z][a-zA-Z0-9_]*'
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -272,7 +272,7 @@ contexts:
           pop: true
         - include: for-parens-body
   for-braces-body:
-    - match: '^(?=([^<←=\{]|<[^\-])+(<-|←|=[\s\b]))'
+    - match: '^(?=([^<←=\{\}]|<[^\-])+(<-|←|=[\s\b]))'
       push:
         - match: '^\s*(val)\s+'
           scope: keyword.declaration.stable.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -36,6 +36,8 @@ contexts:
     - include: char-literal
     - include: scala-symbol
     - include: empty-parentheses
+    - include: braces
+    - include: parameter-list
     - include: qualifiedClassName
     - include: xml-literal
     - include: late-keywords
@@ -140,6 +142,17 @@ contexts:
           captures:
             1: keyword.control.flow.scala
         - include: pattern-match
+  braces:
+    - match: \(
+      push:
+        - match: \)
+          pop: true
+        - include: main
+    - match: \{
+      push:
+        - match: \}
+          pop: true
+        - include: main
   function-type-parameter-list:
     - match: '(?=\()'
       set: function-parameter-list
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -134,7 +134,7 @@ contexts:
             - match: '\)'
               pop: true
             - include: pattern-match
-        - match: '=>'
+        - match: '=>|⇒'
           pop: true
         - include: pattern-match
   function-type-parameter-list:
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -119,6 +119,18 @@ contexts:
       captures:
         1: keyword.other.scoping.scala
         2: entity.name.package.scala
+    - match: '\b(case)\b'
+      captures:
+        0: keyword.declaration.scala
+      push:
+        - match: '\('
+          push:
+            - match: '\)'
+              pop: true
+            - include: pattern-match
+        - match: '=>'
+          pop: true
+        - include: pattern-match
   function-type-parameter-list:
     - match: '(?=\()'
       set: function-parameter-list
@@ -225,7 +237,7 @@ contexts:
   keywords:
     - match: \b(return|throw)\b
       scope: keyword.control.flow.jump.scala
-    - match: \b(else|if|do|while|for|yield|match|case)\b
+    - match: \b(else|if|do|while|for|yield|match)\b
       scope: keyword.control.flow.scala
     - match: \b(catch|finally|try)\b
       scope: keyword.control.exception.scala
@@ -332,3 +344,23 @@ contexts:
           pop: true
         - include: xml-literal
         - include: xml-attribute
+  pattern-match:
+    - include: comments
+    - include: block-comments
+    - include: constants
+    - include: char-literal
+    - include: scala-symbol
+    - include: strings
+    - include: xml-literal
+    - match: '`'
+      push:
+        - meta_scope: source.scala
+        - match: '`'
+          pop: true
+    - match: '\b[a-z][a-zA-Z0-9_]*\b'
+      scope: entity.name.parameter variable.parameter
+    - match: '\b[A-Z][a-zA-Z0-9_]*'
+      captures:
+        0: entity.name.class
+    - match: '@|_'
+      scope: keyword
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -25,6 +25,7 @@ contexts:
     - include: storage-modifiers
     - include: declarations
     - include: inheritance
+    - include: for-comprehension
     - include: keywords
     - include: imports
     - include: comments
@@ -242,6 +243,40 @@ contexts:
       captures:
         1: keyword.scala
         2: entity.name.class.scala
+  for-comprehension:
+    - match: '\b(for)\s*\{'
+      captures:
+        1: keyword.control.flow.scala
+      push:
+        - match: '\}'
+          pop: true
+        - include: for-body
+    - match: '\b(for)\s*\('
+      captures:
+        1: keyword.control.flow.scala
+      push:
+        - match: '\)'
+          pop: true
+        - include: for-body
+  for-body:
+    - match: '^(?=([^<←=]|<[^\-])+(<-|←|=))'
+      push:
+        - match: '^\s*(val)\s+'
+          scope: keyword.declaration.stable.scala
+        - match: <-|←|=
+          pop: true
+        - include: pattern-match
+    - match: '\{'
+      push:
+        - match: '\}'
+          pop: true
+        - include: main
+    - match: '\('
+      push:
+        - match: '\)'
+          pop: true
+        - include: main
+    - include: main
   keywords:
     - match: \b(return|throw)\b
       scope: keyword.control.flow.jump.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -272,7 +272,7 @@ contexts:
           pop: true
         - include: for-parens-body
   for-braces-body:
-    - match: '^(?=([^<←=]|<[^\-])+(<-|←|=))'
+    - match: '^(?=([^<←=\{]|<[^\-])+(<-|←|=[\s\b]))'
       push:
         - match: '^\s*(val)\s+'
           scope: keyword.declaration.stable.scala
@@ -438,5 +438,10 @@ contexts:
     - match: '\b[A-Z][a-zA-Z0-9_]*'
       captures:
         0: entity.name.class
+    - match: \[
+      push:
+        - match: \]
+          pop: true
+        - include: main
     - match: '@|_'
       scope: keyword
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -250,15 +250,15 @@ contexts:
       push:
         - match: '\}'
           pop: true
-        - include: for-body
+        - include: for-braces-body
     - match: '\b(for)\s*\('
       captures:
         1: keyword.control.flow.scala
       push:
         - match: '\)'
           pop: true
-        - include: for-body
-  for-body:
+        - include: for-parens-body
+  for-braces-body:
     - match: '^(?=([^<←=]|<[^\-])+(<-|←|=))'
       push:
         - match: '^\s*(val)\s+'
@@ -277,6 +277,25 @@ contexts:
           pop: true
         - include: main
     - include: main
+  for-parens-body:
+    - match: '<-|←|='
+      push:
+        - match: ;
+          pop: true
+        - match: '\{'
+          push:
+            - match: '\}'
+              pop: true
+            - include: main
+        - match: '\('
+          push:
+            - match: '\)'
+              pop: true
+            - include: main
+        - include: main
+    - match: '\b(val)\b'
+      scope: keyword.declaration.stable.scala
+    - include: pattern-match
   keywords:
     - match: \b(return|throw)\b
       scope: keyword.control.flow.jump.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -23,9 +23,9 @@ variables:
 contexts:
   main:
     - include: storage-modifiers
-    - include: keywords
     - include: declarations
     - include: inheritance
+    - include: keywords
     - include: imports
     - include: comments
     - include: block-comments
@@ -95,7 +95,7 @@ contexts:
         1: keyword.declaration.scala
         2: entity.name.function.scala
       push: function-type-parameter-list
-    - match: '(case)?\b(class|trait|object)\s+([^\s\{\(\[]+)'
+    - match: '\b(case\s+)?(class|trait|object)\s+([^\s\{\(\[]+)'
       captures:
         1: keyword.declaration.scala
         2: keyword.declaration.scala
@@ -355,6 +355,7 @@ contexts:
   pattern-match:
     - include: comments
     - include: block-comments
+    - include: keywords
     - include: constants
     - include: char-literal
     - include: scala-symbol
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -134,8 +134,10 @@ contexts:
             - match: '\)'
               pop: true
             - include: pattern-match
-        - match: '=>|⇒'
+        - match: '=>|⇒|\b(if)\b'
           pop: true
+          captures:
+            1: keyword.control.flow.scala
         - include: pattern-match
   function-type-parameter-list:
     - match: '(?=\()'
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -105,11 +105,17 @@ contexts:
       captures:
         1: keyword.declaration.scala
         2: entity.name.type.scala
-    - match: '\b(?:(val)|(var))\s+(([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`)?'
+    - match: '\b(?:(var))\s+(([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`)?'
+      captures:
+        1: keyword.declaration.volatile.scala
+        2: entity.name.var.declaration variable.parameter
+    - match: '\b(val)\b'
       captures:
         1: keyword.declaration.stable.scala
-        2: keyword.declaration.volatile.scala
-        3: entity.name.val.scala
+      push:
+        - match: '=|\n'
+          pop: true
+        - include: pattern-match
     - match: '\b(package object)\s+([^\s\{\(\[]+)'
       captures:
         1: keyword.declaration.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -98,7 +98,7 @@ contexts:
         1: keyword.declaration.scala
         2: entity.name.function.scala
       push: function-type-parameter-list
-    - match: '\b(case\s+)?(class|trait|object)\s+([^\s\{\(\[]+)'
+    - match: '\b(case\s+)?(class|trait|object)(?:\s+([^\s\{\(\[]+))?'
       captures:
         1: keyword.declaration.scala
         2: keyword.declaration.scala
--- a/Scala/Symbols.tmPreferences
+++ b/Scala/Symbols.tmPreferences
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Symbol List</string>
 	<key>scope</key>
-	<string>entity.name.function.declaration, entity.name.class.declaration, entity.name.val.declaration, entity.name.type.declaration</string>
+	<string>entity.name.function.declaration, entity.name.class.declaration, entity.name.val.declaration, entity.name.type.declaration, entity.name.parameter</string>
 	<key>settings</key>
 	<dict>
 		<key>showInSymbolList</key>
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -347,7 +347,7 @@ object Foo
    case abc @ `abc` =>
 //      ^^^ entity.name variable.parameter
 //          ^ keyword
-//            ^^^^^ source.scala
+//            ^^^^^ - variable.parameter
 
    case foo: (Int => Boolean) :: _ =>
 //                               ^ keyword
@@ -387,27 +387,27 @@ object Foo
 //      ^^^ constant.language.scala
 
    case _ ⇒ _
-//          ^ source.scala
+//          ^ - keyword
 
    case _ if stuff =>
 //        ^^ keyword.control.flow.scala
-//           ^^^^^ source.scala
+//           ^^^^^ - variable.parameter
 
    val abc @ `abc`
 // ^^^ keyword.declaration.stable.scala
 //     ^^^ entity.name variable.parameter
 //         ^ keyword
-//           ^^^^^ source.scala
+//           ^^^^^ - variable.parameter
 
    _
-// ^ source.scala
+// ^ - keyword
 
    val ble @ `abc` = _
 // ^^^ keyword.declaration.stable.scala
 //     ^^^ entity.name variable.parameter
 //         ^ keyword
-//           ^^^^^ source.scala
-//                   ^ source.scala
+//           ^^^^^ - variable.parameter
+//                   ^ - keyword
 
    case object Thingy extends Other
 // ^^^^ keyword.declaration.scala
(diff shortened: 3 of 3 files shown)


djspiewak:
**Update** I figured out how to do lookahead scoping, so `for`-comprehensions now highlight correctly (both brace- and paren-style) and are subject to symbol indexing.

Scala/Scala.sublime-syntax
```diff
@@ -89,7 +90,7 @@ contexts:
       captures:
         1: keyword.declaration.scala
         2: entity.name.function.declaration
-    - match: '(case)?\b(class|trait|object)\s+([^\s\{\(\[]+)'
+    - match: '\b(case\s+)?(class|trait|object)\s+([^\s\{\(\[]+)'
```
mkhl: You could avoid assigning the keyword scope to whitespace by adding a non-capturing group: `(?:(case)\s+)?`

Scala/Scala.sublime-syntax
```diff
@@ -98,11 +99,17 @@ contexts:
       captures:
         1: keyword.declaration.scala
         2: entity.name.type.declaration
-    - match: '\b(?:(val)|(var))\s+(([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`)?'
+    - match: '\b(?:(var))\s+(([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`)?'
```
mkhl: The non-capturing group around `var` seems unnecessary: `\b(var)\s+` should be fine.

Scala/Scala.sublime-syntax
```diff
@@ -265,3 +349,24 @@ contexts:
...
+    - include: char-literal
+    - include: scala-symbol
+    - include: strings
+    - include: xml-literal
+    - match: '`'
+      push:
+        - meta_scope: source.scala
+        - match: '`'
+          pop: true
+    - match: '\b[a-z][a-zA-Z0-9_]*\b'
```
mkhl: Shouldn't dollar signs be allowed here as well?

Scala/Scala.sublime-syntax
djspiewak: Or, without introducing backtracking: `((case)\s+)?`.

Scala/Scala.sublime-syntax
djspiewak: I agree.

Scala/Scala.sublime-syntax
djspiewak: Yep. This falls into the same category as "all the other regexes that are slightly, but not completely, inconsistent and wrong".  Extracting everything into variables is the way to fix that, but I only want to do that once, rather than once per outstanding PR.

djspiewak:
As a quick FYI, there are currently problems with this PR and the following two snippets:

```scala
          back <- Traverse[Option]
            .traverse[FetchAlg.Free, CurrentConditions](res) { loc =>
```

Almost all of the second line is mis-colored.

```scala
  lazy val baseSettings: Seq[Def.Setting[_]] =
```

Just the underscore.

Both problems are pretty easy to solve.  Just noting here.

Scala/Scala.sublime-syntax
FichteFoll: Non-capturing groups do not introduce backtracking (that would be terrible!). You might be thinking of atomic groups.

djspiewak: @FichteFoll Oh of course you're right.  Well, given that `.sublime-syntax` allows arbitrary group indexing in the `captures` clause, it doesn't much matter, but I suppose the right way to do it is as @mkhl suggested, with a non-capturing group.

djspiewak:
Ping @wbond. Any chance of some review and/or merge on a few of these [scala] PRs?  I have several more things I'd like to do, but I'm basically stuck due to the fact that everything is split up into separate PRs, and a lot of what I want to do cuts across them.

wbond:
I've been working on minihtml recently, but I will address these the next time I come back to the Packages. Sorry for the delay.

djspiewak:
Outstanding issue with this PR:

```scala
for {
  x <- stuff
} yield x.copy(field = value)

_
```

The trailing underscore will be highlighted as a keyword, as if it were in a pattern match position.  In fact, the *rest of the file* will be highlighted as pattern match mode.  This is a consequence of a strange interaction in regex priorities between scoping and backtracking suffix match.  It can be fixed by the user by removing the spaces surrounding the `=`, but it would be better if the mode just behaved correctly.  I don't think this is too hard to fix.

djspiewak:
@wbond For conflict resolution with master, do you prefer rebase or merge?

wbond:
Considering all of the cherry-picking and tweaks I've been doing during the PRs, it probably makes sense to rebase this PR, then I can check it out. I'm going to be committing a whole bunch of scope fixes for Scala in a little bit, so you may want to wait an hour or so to ensure the latest changes are up.

djspiewak:
@wbond Will do.

wbond:
Ok, I've fixed the scope name errors I found. There is definitely a bit of work left to do on parameters and various identifier names. For instance, some parameter types are `entity.name` (which is incorrect), while others are `storage.type`.

gwenzek:
@wbond Which scope do you recommend for types like `Foo` in `val abc: Foo` ?
`storage.type` is already use for marking `val`and `def` I think it would be an error to use it with for marking `Foo`. 
I think a dedicated scope should be created, like `variable.type`.

gwenzek:
@djspiewak why do you use `variable.parameter` for val declaration like in: `val List((x1: Int, y1), (x2, y2)) = ((10, 5), (1, 2))`. I used the scope `entity.name.val.scala`

wbond:
@gwenzek Some of the syntaxes currently use `storage.type` just for built-in types, and nothing for user types. Others use `support.type` or `support.class` for user types. It partially depends on the history of the syntax, and is hard to change, since it tends to significantly alter the look of code, which users don't typically like.

I don't know if `variable.type` is the best choice. Currently we are using `variable.function` for user function names, however this seems less than ideal also. There had been some discussion of creating a top-level `identifier.` scope for class and type names that are from the user.

djspiewak:
@gwenzek I would be fine with `entity.name.val.scala`, though remember that val declarations and pattern match declarations (`case`) and `for`-comprehension generators are *all* the same scope, so the `val` part actually doesn't make as much sense as it first seems.

djspiewak:
@wbond Rebased.  I think some of the scope names may have regressed under certain circumstances as a consequence.  I had to adjust the class declaration matching a bit, since the regex wasn't matching the partial fragment `case class` properly (caught by a pre-existing test).

djspiewak:
@wbond Something I did discover from this: your entity name changes have regressed symbol indexing MASSIVELY for this mode.  It's no longer possible to jump to class or object declarations.  In fact, I think the only indexing which is working at this point is function names.  I'll be fixing locally for now just so I can make the editor work again, but I'd be happy to commit upstream.  Would you like that in a separate PR?

wbond:
@djspiewak This line is the issue (and is totally incorrect currently): https://github.com/sublimehq/Packages/blob/master/Scala/Symbols.tmPreferences#L7. It should limit its scope to source.scala, and then get rid of the `.declaration` parts.

djspiewak:
There's a slight semantic flaw in this as it stands:

```scala
x match {
  case Foo => 42
}

val Foo = 42
```

The `val` introduces a new variable named `Foo`.  The `case` checks to see if `x` is equal to `Foo`.  To make matters more confusing:

```scala
val (x, Foo) = y
```

Now the `val` introduces a single variable named `x` while checking to see if `y` is a `Tuple2` where the second value is equal to `Foo`.

So it's a very slight difference in the way the scoping needs to be assigned, since in the first example, the `Foo` in `val Foo` should be indexed, but the `Foo` in `case Foo` from the first example or `val (x, Foo)` in the second example should *not* be indexed.

wbond:
So looking into details of Scala, `var` is for defining a variable, and `val` is for defining an immutable variable.

The correct scopes for the identifiers , respectively is `variable.other.readwrite` and `variable.other.constant`. In some situations it won't be possible to know which is which. Some syntaxes just assign `variable.other.readwrite` to every identifier that doesn't receive another scope name. Most color scheme highlight these the same.

Neither of these should receive an `entity.name` scope. If a user wishes variable declarations to show in the symbol list, they can override the `.tmPreferences` file and include the scope when a variable is defined. It may be that we need to create a new sub-scope of `variable.other.readwrite` such as `variable.other.readwrite.declaration` to allow for this to happen. Alternatively another approach would be to add a `meta.variable.declaration` scope to the code where a variable is defined. Then the `.tmPreferences` file could add `source.scala meta.variable.declaration variable.other.readwrite` to the syntax list and index.

wbond:
In terms of all of the different ways that identifiers can be used, and the ambiguity in that, sometimes we just have to punt on highlighting and leave it as `source.scala`. What is most important is keywords, string and numeric literals, language constants and function and data structure definitions.

Things beyond that are nice, but just sometimes can't happen. For instance with Ruby we can't reliably scope method calls, because a bare identifier can be a method call.

djspiewak:
@wbond From a Scala semantic standpoint, there is no difference between `foo` and `bar` in the following:

```scala
class Stuff {
  val foo = 42

  def bar = 42
}
```

Why should `bar` be indexed and `foo` should not?

wbond:
Neither should be. Only functions and classes/structs/enums/unions/interfaces/traits are generally indexed. Sometimes constants are, but that tends to be rare.

djspiewak:
@wbond That's exactly my point.  *Both* the `val` *and* the `def` in my example are functions.  Scala does not distinguish between these cases because there is no distinction, semantically.

wbond:
42 is a function?

djspiewak:
@wbond It can be viewed as a function, yes, but `foo` and `bar` are functions in the above.  That's the important bit.  Neither of them are 42.

wbond:
If every instance of `val` is a function, then sure, it should be indexed.

wbond:
So what does `42` represent in the code above?

djspiewak:
@wbond Every instance of a `val` is a function, though I would accept an argument that nested `val` and `def` shouldn't be indexed (but should nested classes?).

So where is the line then?

```scala
val (x, y) = stuff()
```

Should `x` and `y` be indexed in the above?  They're both functions.

```scala
stuff() match {
  case (x, y) => ...
}
```

How about here?

```scala
for {
  (x, y) <- stuff()
}
```

Or here?

In the code above, `42` is an integer literal, but it is also an object, and thus can be made to be a function.  More importantly though, `foo` and `bar` are both methods of class `Stuff`.

Here's an example of nested things:

```scala
class Stuff {
  def foo(x: Int) = {
    class Inner { ... }
    def bar(y: Int) = x + y
    val baz = 42
    
    ...
  }
}
```

Which things should be indexed?  Why or why not?

djspiewak:
An example to clarify the "made to be a function" line…

```scala
implicit class RichInt(val self: Int) extends AnyVal {
  def apply(x: Int): Int = self + x
}
```

As long as the above is in scope, any integer is actually also a function which adds itself to its argument:

```scala
40(2)  // => 42
```

This pattern works with anything that is an object, and since everything is an object…  So you see why this can get a bit semantically muddled.  :-)

wbond:
We don't index fields of structs, or similar things. So, it would seem to me that in your final example `baz` is not a function, it acts semantically as a member/attribute/field.

I presume there is an effective difference between `val` and `def`, otherwise why would the language have both? Sure, in JavaScript, any variable may hold a function, and in Ruby any identifier may be a method. However, we only highlight and index things that are definitely functions that way. Otherwise once everything is highlight, effectively nothing is highlighted.

We tend to err on the side of "this is almost definitely a function" instead of "this maybe a function". For instance, most types in C and C++ are not highlighted because it is not trivial to know which identifiers are types and which are variables. Similarly in Ruby we highlight things that definitely seem to be classes as such, but things which may be any of a number we do nothing with.

In the case of the final code snippet you presented, I would expect `Stuff` and `Inner` to be indexed as classes, and `foo` and `bar` to be indexed as functions. `baz` is effectively a field, even if the implementation of it allows operator overloading. Sort of how in Ruby accessing an attr is a message being sent, like "calling" a method is.

djspiewak:
> I presume there is an effective difference between val and def, otherwise why would the language have both? Sure, in JavaScript, any variable may hold a function, and in Ruby any identifier may be a method. However, we only highlight and index things that are definitely functions that way. Otherwise once everything is highlight, effectively nothing is highlighted.

The difference is primarily legacy due to the way the JVM works.  It affects `trait` linearization pre-Java 8, and it affects the precise bytecode that is generated.  There is also some really minor and weird type inference differences, which most people don't know about and can probably be put down as compiler bugs.  From a standpoint of pure language semantics though, they are the same, except that `def` can have parameters while `val` cannot.  Thus, a more uniform language design, free of the need to hint things to the JVM for performance/other reasons, would have done away with the `val` keyword altogether and just used `def`.

In fact, this kind of change (swapping a `val` to a `def` and vice versa) is extremely common in practical Scala codebases.  It is for this reason that I feel it is enormously important that `val` is indexed in all scopes, because it would be intensely annoying to make that change purely to hint optimization semantics to the JVM and be punished by having the symbol removed from my editor's index.

> In the case of the final code snippet you presented, I would expect Stuff and Inner to be indexed as classes, and foo and bar to be indexed as functions. baz is effectively a field, even if the implementation of it allows operator overloading. Sort of how in Ruby accessing an attr is a message being sent, like "calling" a method is.

So to clarify:

```scala
class Stuff {
  val foo = 42
  def bar = {
    def bin = 42
    val baz = 42
  }
}
```

You would index `foo`, `bar` and `bin`, but not `baz`?

djspiewak:
@wbond Another example to point out how dangerous this line of distinction is:

```scala
def foo = {
  lazy val bar = new Stuff {}
  object baz extends Stuff
}
```

Would you index both `bar` and `baz`?  Or only one of them?  The bytecode they produce is almost identical.

wbond:
I would not index `foo` or `baz`, since they use `val` and are assign a literal constant value. I would consider indexing a `val` if it was assigned to an object literal – we do things like that in JavaScript. An object field name is highlighted and indexed as a function name when the right-hand-side is `function` or an arrow function.

djspiewak:
> I would not index foo or baz, since they use val and are assign a literal constant value. I would consider indexing a val if it was assigned to an object literal – we do things like that in JavaScript. An object field name is highlighted and indexed as a function name when the right-hand-side is function or an arrow function.

```scala
class Stuff {
  val foo = thing()
  def bar = {
    def bin = otherThing()
    val baz = moreDifferentThings()
  }
}
```

How about now?

```scala
class Super {
  def foo: Int
}

class Child extends Super {
  val foo = 42
}
```

Do you only index the `foo` in `Super`?  What if I refactor `Child` to the following:

```scala
class Child extends Super {
  def foo = 42
}
```

*Literally* identical bytecode aside from the metadata field, but is `foo` in `Child` now indexed?

wbond:
What does `new` mean in Scala? What does `object` mean?

Assuming they use similar definitions to other popular languages, I would say neither since they are instances.

djspiewak:
> What does new mean in Scala? What does object mean?

Scala's `new` is identical to Java's.  Scala's `object` is a combination `val` and `class` declaration.  So in other words, there is almost no difference between the following three snippets:

```scala
// first
lazy val foo = new Stuff {}

// second
object foo extends Stuff

// third
class foo extends Stuff
lazy val foo = new foo
```

I mean, I wouldn't normally give a class a lower-case name, but there's nothing invalid about it.

wbond:
I'm sure you can come up with convoluted examples of all sorts of things, just like you can in Ruby, Python, JavaScript or a handful of other dynamic languages that have first-class classes and functions, or blur the distinction between them.

In short, we index:

 - functions
 - types
 - data structures

Once you get into assigning such constructs to other identifiers, the support drops off unless we can unambiguously determine an identifier definitely points to a construct that is normally on the index.

wbond:
In

```scala
// first
lazy val foo = new Stuff {}

// second
object foo extends Stuff

// third
class foo extends Stuff
lazy val foo = new foo
```

I would only index `foo`.

djspiewak:
> In short, we index…

Ok…  So here's what I would suggest we do then.  This would not be the most internally consistent index set, but it sort of works, and I think it bridges the gap between your index guidelines and the actual Scala language semantics:

```scala
val foo = stuff
def foo = stuff
val (foo, bar) = stuff
lazy val foo = stuff
```

All of the above are indexed, in all scopes, as they are all functions.

```scala
class Foo
trait Bar
object Baz
type Bin = Stuff
```

All of the above are indexed, in all scopes, as they are all types.

It's at this point that I have no idea what you mean by "data structures", but I assume JavaScript records, for which Scala has no analog.

```scala
stuff match {
  case (foo, bar) => ...
}

for {
  (foo, bar) <- stuff
}
```

Nothing in the above is indexed.  Technically speaking, `foo` and `bar` in both of the above blocks are semantically basically the same as a `val`, and they are using the same pattern matching deconstruction as the `val` declaration syntax does, but I couldn't unambiguously say that they're functions in the same way that `val` parameters or `def` declarations are.

```scala
def foo(bar: Int) = ...
```

For the same reason, `bar` in the above would not be indexed.

Assuming you agree to the above, there is only one remaining complication:

```scala
package foo

package bar {
  // your stuff here
}

package object baz {
   // oh dear...
}
```

I would propose that neither the `foo` nor the `bar` in the above should be indexed (they're namespaces, not types), but the `baz` should be (it's a type).  I wouldn't fight if you wanted to index `bar` though.

> I would only index foo.

lol…

wbond:
>> I would only index foo.
> lol…

Please check the comment above, as I clarified.

Anyway, your comments don't address the gist of mine, which clarified that we only index things we know are functions. The details of the JVM implementation and bytecodes and all of that really don't matter one bit.

What truly matters is how these are practically used in the real-world, which I am not experienced enough in Scala to make a judgement call on. However, based on your comments here, I don't think you have enough experience with how Sublime Text scopes, the symbol list and indexer are generally used.

I'm going to push pause on this PR since it is sprawling on and the we are just arguing past each other at this point.

Hopefully we can get some other Scala users in here, and use some info from the Atom and VS Code syntaxes for Scala (if VS Code even has one). I can see that GitHub's syntax highlighting for Scala seems to be a mix of our two current positions.

djspiewak:
Oh, I just realized my proposal doesn't address `var`…

Selfishly, I don't care whether or not `var` is indexed, since I simply don't use it.  Practically though, it would probably be most consistent with your guidelines to *not* index it in any scope, despite the fact that it's not that uncommon among Scala newcomers to use `val` and `var` interchangeably, and also despite the fact that `var` in a class (or trait or object) is technically a function like any other.

> What truly matters is how these are practically used in the real-world, which I am not experienced enough in Scala to make a judgement call on. However, based on your comments here, I don't think you have enough experience with how Sublime Text scopes, the symbol list and indexer are generally used.

What I have practical experience in is using the editor, using ctags, using the indexer, and using Scala.  I don't have experience using the indexer on other languages (I use other languages in sublime, just not anywhere near the extent to which I use Scala).  And I don't care at all about scopes, though I do share the desire to ensure that all of the language modes are consistent.

If I can't convince you to make the editor as good as it can be for Scala, then I will simply fork the mode and go away.  I don't *need* my changes to be integrated into the upstream Sublime core, it just makes the world a better place for other Scala users.

GitHub's Scala highlighting is broken.  It's pretty closely based on Sublime's original Scala mode, which was also broken and in pretty much the same ways.  Ctags attempts to index Scala `val`s (not sure about `var`s) in all scopes, but their regular expressions are broken and they don't correctly handle pattern match extraction (in fact, they will actually build invalid indexes).  It doesn't try to handle `for` or `case`.  Eclipse and IntelliJ both allow you to jump to symbols that are defined as variables, pattern matches, `var`s, whatever.  If it's a name, you can get to it.  I have no idea what Atom does, but I would guess it pretty closely matches Sublime's old behavior, which closely matched the ctags behavior (i.e. functional enough to be useful, but still broken in weird ways).

The most comprehensively correct syntax highlighting mode for Scala in current existence is the Emacs mode, and it doesn't address indexing (since everyone basically just uses ctags).  It does correctly highlight pattern matching, though, and I believe it also correctly highlights `for`-comprehensions.

gwenzek:
I'm just trying to give my two cents. TL;DR I agree with Wbond.

In Scala, `def`marks a function/method. `val` are value computed once and that can't be modified afterwards. You can have functions with 0 arguments that can be used in code like a `val`.

The indexing in Sublime have two goals, filling the symbol list and allowing the 'goto' command to work. For the symbol list to be useful, I prefer when only important elements are present in it, like classes and top level functions. `val` are generally one line of code, i'm not sure if they deserve an entry on the symbol list.

In IntelliJ and Eclipse the goto command allows to jump to the definition of any symbol. This use a complex analysis chain. In Sublime only the name of the symbol is used. If every `val` is indexed the goto command will also become unusable because there will be too many name collisions.

So I suggest that only class and `def` should be indexed, it's maybe not always coherent since a `val` can override a `def` but it seems to be the best we can do without implementing a Scala analysis chain in Sublime.

Another option would be to allow the user to choose the scopes he wish to index with a simple setting.

gwenzek:
Otherwise I think this PR is good, so I hope you'll come to an agreement 😃

djspiewak:
> In IntelliJ and Eclipse the goto command allows to jump to the definition of any symbol. This use a complex analysis chain. In Sublime only the name of the symbol is used.

Press Cmd-Shift-R in Eclipse (IntelliJ has an identical function, but I don't remember the keybind).  It's literally identical to Sublime's goto symbol, and as with Sublime, it only considers the name.

wbond:
> Another option would be to allow the user to choose the scopes he wish to index with a simple setting.

This is already the way it is. We ship a default list of scopes, but users can create `.tmPreferences` file and select what scopes they want in the symbol list.

gwenzek:
@wbond I'm aware of that, it's just not straight forward since it involves creating a tmPreferences rather than editing a Json settings file.

@djspiewak I'm not used to Eclipse, I use only IntelliJ. The question is, what happens if you ask 'goto symbol' on `a.id` and you have 23 `id` defined in your project.

schrepfler:
In Scala land, @djspiewak is very much of an authority. I'd listen to him.

djspiewak:
@wbond I think if the setting could be made slightly simpler or more accessible, that would go a long way.  A good example of this is the new user preferences behavior (what happens when you hit Cmd-,), which is fantastic.  Something like that for hinting users along to adjust which scopes are indexed would solve almost all problems here.

As long as current (in the stable release) behavior isn't regressed, I would be very happen with that kind of setup.  Current behavior is to index `val` (incorrectly, but still indexed).

@gwenzek No question that jumping to symbol is always going to be better in full IDEs (or editor extension modes like ENSIME).

wbond:
I don't have anything to announce at this point, but my hope is to deprecate `.tmPreferences` for a better system in the future.

FichteFoll:
>deprecate .tmPreferences for a better system in the future.

Yes, please! I've been asking for this for years.

wbond:
@schrepfler I have no doubt that @djspiewak seems to be quite familiar with Scala.

@djspiewak My task here is to make the best Scala syntax and symbol indexing that we can for Sublime Text (as with other default syntaxes). However, more does not always mean better. My concern is over having almost every identifier indexed, resulting in the index being semi-useless. The art of user experience is nuanced. I am not claiming to be the utmost expert, but my intuition leads me to believe we have not yet identified the correct balance.

djspiewak:
@wbond I don't dispute that indexing everything is probably bad.  However, practically speaking, current stable Sublime indexes all vals, defs, classes, traits and objects. Ctags does exactly the same. So does Eclipse, in so far as comparisons are relevant.  By correcting some scopes, this PR would fix the broken val indexing while broadly keeping the existing semantics (its relatively easy to not index case and for).  So in other words, identical behavior to stable… except not buggy.  I don't feel like that should be controversial, at least not as a starting point.

Removing val from the index set would be a significant step backwards for the tool as I use it, and also for anyone else who expects ctags-like semantics.

coltfred:
I agree with @djspiewak here, without `val` being indexed Ctrl + Shift + r becomes useless. val and def are about evaluation semantics, not about what they "point" to.

for instance:
```
def add(x:Int, y:Int) = x + y
val add1 = add(1, _)
```
`add1` is a function which adds 1. Putting def in the front of `add1` only changes the evaluation semantics, not the program. Currently sublime indexes vals and even in very large code bases with all the external sources exploded indexing works like a champ.

subramanyamchitti:
The ability to jump to val's like  @djspiewak  mentioned is important to me.

wbond:
I'm glad some Scala users are showing up! Hopefully you can all stick around and help provide info and feedback as improvements are made.

timperrett:
Hi there! Long time Scala+Sublime user here. Jumping to `val` is certainly important and would not want to loose that functionality, as one does not necessarily know apriori if the definition site is `def` or `val`. A good use case for this is the sbt source code - navigating it without `val` indexing would be a nightmare. 

Thanks for the consideration, and thanks for sublime - its a fantastic tool.

wbond:
I've merged this in, with changes necessary to match scopes with existing syntaxes. I've also made some further fixes and improvements upon the Scala syntax.

Thanks @djspiewak for all of your work on this!

With the next dev build there will be documentation about standard scope names, which should provide useful information in terms of selecting the correct scope names for various syntactic elements.

To all of the Scala users who have arrived here to comment on the issue, please stick around and report errors or constructs that need improved handling. Thanks!

wbond:
@djspiewak I'm sure you'll notice that a number of colors of various elements have changed since you authored them, and probably even in general.

I realize this can feel jarring, but what we've done here is (as far as I am aware), it assign correct scopes following the standards that the general community uses. Using a customized color scheme, it should be possible in most cases to change colors back to what you had before, if you prefer that aesthetic.

wbond:
I should also note that indexing should be working properly, as I understand it. I did not change the indexing or `val`, or anything else. Instead, non-declaration of class names were changed to the correct `support.class` scope, which means they will not appear in the symbol list.

Additionally, I added package names to the symbol list and index.

djspiewak:
@wbond I haven't looked yet, but I'm guessing you didn't change the indexing of the following?

```scala
x match {
  case (foo, bar) => ...
}
```

In my original patch, `foo` and `bar` were indexed, since they are in fact highlighted with the same context as `val (foo, bar) = x`, which *are* indexed.  Given what you've told me of Sublime's index conventions, I don't believe `case` or `for` should be indexed (highlighted correctly, yes, but not indexed), which requires some changes from the PR as I left it.

I can make the changes to remove them from the index while still highlighting them reasonably (probably as `variable.parameter` when used in `case` or `for`, while `val` declarations will not be affected).  I can also get started on implementing the type region identification changes I referenced, so that constants don't all become blue.

djspiewak:
@wbond I'll open a new issue, but master is now SUPER buggy.  To wit, `class`, `def` and `val` now highlight as `storage.type`, when they're not types but keywords.  The `support.class` stuff is sort of insane and over-applied, as we previously discussed, but that was expected and can be fixed.  But the fact that keywords are now highlighted as storage types can't be right.

wbond:
No, these are the correct scopes. You can check out other languages to see how those scopes are used. C, Go, Rust, Python, etc.

gwenzek:
According to [Textmate documentation](http://manual.macromates.com/en/language_grammars) the scope `storage.type` is used to describe "the type of something, class, function, int, var, etc."

Here "type" doesn't have the same meaning than in Scala. In Python "def" also has the scope "storage.type". The problem IMHO is routed in those Textmate guidelines that where written for C like languages where the return type of a function is used as a function marker. `int foo()` vs `def foo(): Int`. So I find "storage.type" correct here even if the name is misleading.

But I find weird than in "case class" the two words don't receive the same scope ("keword" and "storage" respectively). Same thing for `f: A => Int` where `A` is marked as a "support.class", `=>` has no scope while it's a type used as an operator and `Int`is scopes as "storage.type.primitive". For me the "storage.type.primitive" is something new that collide with the well established "support.type". If you create a new scope it should be for user defined types rather than the standard types.
Other way, using "support.type" for every type is also an option.

I also think that the scopes for interpolated string should be improved. For me the whole block of code in a string should be marked with a special scope to allow themes creators to 
NOT color the code part in the string color.

djspiewak:
@wbond Didn't you say that the Ruby mode has been recently redone and may be considered relatively authoritative?  `class` and `def` (correctly!) highlight as keywords in that language.

Also, `def` has no analogue in the following C fragment:

```c
void main(char **argv);
```

Obviously, `void` should be highlighted as `storage.type`, since that's exactly what it is.  But `def` is not a type, nor is it declaring the inference of a type.  It is a syntactic element, exactly as it is in Ruby.

PR #494, opened 2016-06-22
[RegExp] Comments in extended mode and scope adjustments

FichteFoll:
I implemented comment parsing for the regular expression syntax. This required tracking the extended mode state within groups and I had to duplicate a lot of base\* and group\* contexts, but it's working very nicely now.

Additionally, I changed a few minor things such as a more appropriate scope name for mode modifiers.


wbond:
Thanks!

PR #498, opened 2016-06-23
[Python] With statements with multiple resources

### LINKED ISSUE #496: python: double with as on same line erroneously claimed as illegal name
Works in default 3114, not with the latest downloaded from here.

FichteFoll:
Fixes #496.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -166,22 +166,7 @@ contexts:
       captures:
         1: storage.modifier.async.python
         2: keyword.control.flow.with.python
-      push:
-        - meta_scope: meta.statement.with.python
-        - include: line-continuation-or-pop
-        - match: \b(as)\b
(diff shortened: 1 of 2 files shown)


jcberquist:
I had just been staring at some Python code that had this issue, and your pull request with the fix showed up :smile: Very nice - it is really only as I have watched these syntaxes get rewritten that I have grasped the power of being able to use `set` to move between contexts.

FichteFoll:
My current favorite with context mangling is what @keith-hall did with the [unexpected quantifier context](https://github.com/sublimehq/Packages/blob/211a27c8c6b5fb42a85a5139028ca5efc5470e58/Regular%20Expressions/RegExp.sublime-syntax#L67) in the RegExp syntax (and my recent additions to that regarding extended mode in #494). It made me realize what pushing multiple contexts actually does, since I had the wrong idea of it previousl.

PR #507, opened 2016-07-02
[Makefile] Add first_line_match for makefile shebang

dlech:


Makefile/Makefile.sublime-syntax
```diff
@@ -12,6 +12,7 @@ file_extensions:
   - OCamlMakefile
   - mak
   - mk
+first_line_match: '^#!/usr/bin/(make|env\s+make)'
```
wbond: This should probably be something more like: `'^#!/usr/bin/(make|env\s+make)\b'` to ensure `make` isn't part of a longer word.

wbond:
Do people not use the `/usr/bin/env make` format?

dlech:
I don't think so. I found a StackOverflow question about this and the conclusion was no. I also tried it and it didn't work because you have to pass the `-f` command line option to the `make` command itself. So, I'm pretty sure no one uses it.

wbond:
The only other thought would be to match `/usr/local/bin/make`, although that seems unlikely in practice. I also don't imagine distros ship `make` as `/bin/make`.

dlech:
Well, we can always change it if we find out there is another case.

If you think this is good enough for now, I can squash if you don't use the squash merge feature of GitHub.

Issue #508, opened 2016-07-04
[Python] == is scoped as keyword.operator.assignment, not keyword.operator.comparison


MattDMo:
This tests all the comparison operators:
```python
1 > 2
1 >= 2
1 < 2
1 <= 2
1 == 2
1 != 2
```
Using PythonImproved, everything is scoped correctly:
![comparison pi](https://cloud.githubusercontent.com/assets/1776358/16548782/a40cb624-4164-11e6-821f-6e73d9a0b4d6.png)

However, using the latest `Python.sublime-syntax`, `==` is scoped as `keyword.operator.assignment`, which is orange, as you can see from an actual assignment below:
![comparison ss](https://cloud.githubusercontent.com/assets/1776358/16548803/f80e2d5c-4164-11e6-8c7e-d8f1e87f393f.png)

A PR that I am about to submit fixes this. The old code is
```yaml
320  - match: '='
321    scope: keyword.operator.assignment.python
```
while the fix is simply:
```yaml
320  - match: '=(?!=)'
321    scope: keyword.operator.assignment.python
```

wbond:
Fixed by #509

PR #509, opened 2016-07-04
[Python] fix for keyword.operator.assignment/keyword.operator.comparison issue #508

MattDMo:
Pretty self-explanatory. Single-line change in `sublime-syntax` file, two tests added, passes with flying colors.


FichteFoll:
If we match the comparison operator before the assignment one this fix won't require a look ahead.

PR #512, opened 2016-07-06
Some additional syntax file extensions

gerardroche:


Python/Python.sublime-syntax
```diff
@@ -3,8 +3,10 @@
 name: Python
 file_extensions:
   - py
+  - py3
   - pyw
   - pyi
+  - pt
```
FichteFoll: Can you provide any reference for where this extension is used and what it means?

I found `py3` in [some SO question](http://stackoverflow.com/a/18032741/1327727), but I've never actually seen it. Still, I guess that counts.

keith-hall: There was a proposal for it [on the Python mailing list](https://mail.python.org/pipermail/python-list/2001-July/082404.html). I guess it (or something very similar) was implemented, because I have heard of the py3 file extension before.

gerardroche:
I just added all the additional filetypes I found in the Boxy Theme extensions settings. So I'm ok with whatever. If in doubt, leave it out. :)

wbond:
Let's remove `pt` from Python. Otherwise the changes seem fine.

PR #513, opened 2016-07-06
Theme file icon support

gerardroche:
See https://forum.sublimetext.com/t/sublime-text-3-file-icons-in-sidebar/21134

This is just the first PR for the most basic support. I'll prepare the rest later.


oivva:
I think we should wait for @wbond answer in https://forum.sublimetext.com/t/sublime-text-3-file-icons-in-sidebar/21134. If that files will be added to default packages, I'll remove this files.

wbond:
I haven't looked into themes and icons yet, nor discussed it with Jon.

That said, it seems it would make more sense to deal with this in core, trying to find an icon file in the format `file_type_<lowercase_syntax>` and falling back to a generic icon if that files is not part of the theme.

gerardroche:
I'm going to close this PR because these changes would only break themes that don't have specific icons.

The forum discussion: https://forum.sublimetext.com/t/sublime-text-3-file-icons-in-sidebar/21134

PR #519, opened 2016-07-11
[PHP] Ensure namespace elements and separators are scoped with their class

duncan3dc:
Some of the recent changes meant that the leading slash was not scoped with the class name, this meant tooling was unable to tell if the code was using a local `Exception` class or the global `\Exception` class.

This pr fixes that, along with scoping the entire namespace as the class too

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -119,7 +119,9 @@ contexts:
       captures:
         1: keyword.control.exception.catch.php
       push:
-        - meta_scope: meta.catch.php
+        - meta_scope: meta.catch.php meta.path.php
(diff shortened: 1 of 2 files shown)


wbond:
I'm not sure this is the correct solution.

It strikes me that adding a `meta.path.php` scope to the complete `\MyNS\ClassName` is probably more of what you want.

duncan3dc:
Sounds good, I'll take a stab at that

duncan3dc:
I've re-written it to use `meta.path.php` now, how does that look?

wbond:
I added `meta.path` and tests for what appears to be the rest of the situations in PHP where namespaces are valid.

PR #520, opened 2016-07-11
[PHP] Tidy up of the example code to ensure it's valid PHP

duncan3dc:
Using invalid PHP in the tests seems like something that would trip people up when writing tests and changing syntax rules.

* The reserved keyword 'namespace' cannot be used as a sub-namespace
* Add missing semi-colons
* ~~Complex expressions within strings are not supported, eg "{$a . $b}" is not valid PHP~~
* Empty if() statements are not valid
* Remove duplicated function/class names

### DIFF
--- a/PHP/syntax_test_php.php
+++ b/PHP/syntax_test_php.php
@@ -47,102 +47,102 @@ use My\Full\Classname as /**/ Another # Foo baz
 //        ^^^^^^ support.class.php - constant.other - entity.name - support.function.php - support.other.namespace
 //              ^ punctuation.terminator.expression.php - meta.use
 
-use function /**/ some\namespace\fn_a;
+use function /**/ some\nspace\fn_a;
 // <- keyword.other.use
-//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.use
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.use
 //  ^^^^^^^^ storage.type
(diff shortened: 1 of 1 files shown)


duncan3dc:
There's quite a few issues remaining, but most are about duplicate function/class names, etc.
Would you object to me cleaning this up? As they _are_ valid PHP I don't have a strong reason to argue they should be, it just seems like the whole file should be lintable by `php`

wbond:
I don't have a problem with the file being lintable by PHP. The key thing is just that we don't lose any of the functionality of the tests, in terms of constructs that are tested.

Complex expressions are allowed if it is a variable lookup (the `{}` is prefixed with a `$`). Most likely I just missed the `$` when typing that. This ensures we are allowing such valid constructs.

duncan3dc:
Thanks, I'll take a look at making some other changes in a separate pr then.

Could you provide an example of such an expression that is supported by php?

wbond:
```php
$helloworld = "Hello World!";
$a = 'hello';
$b = 'world';
print("${$a . $b}");
```

duncan3dc:
Thanks :+1: I've pushed a change for that now (a791234bc9780c7b4a4c7fb7ad4e0c8c752eba01), but I've not got a dev environment handy to test it at the moment, I'll check it out tonight unless you beat me to it

duncan3dc:
Correcting the test code causes all kinds of other failures, so I've removed that point for this pr

wbond:
Thanks!

Issue #470, opened 2016-06-11
[PHP] Long array and short array scope are not same


ptcong:
I think long array and short array syntax should have same scope, but currently, it give different scopes

`$a = array(|);`
`embedding.php text.html.basic meta.embedded.block.php source.php meta.array.empty.php punctuation.definition.array.end.php`

`$a = array[|]` 
`embedding.php text.html.basic meta.embedded.block.php source.php punctuation.section.array.end.php`

```
$a = array(
    embedding.php text.html.basic meta.embedded.block.php source.php meta.array.php 
);
```

```
$a = [
    embedding.php text.html.basic meta.embedded.block.php source.php 
];
```

wbond:
I believe this is fixed by 81606041f923dcae5caa37fce6aae3f98d46b2cc

wbond:
Also 410027a7dc96e71bc81f94494b365927aaddaa9c

PR #521, opened 2016-07-11
[PHP] Ensure the two types of arrays have consistent scopes

duncan3dc:
Fix for #470

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -239,11 +239,17 @@ contexts:
         1: keyword.control.goto.php
         2: support.other.php
     - include: string-backtick
+    - match: '(\[)(\])'
+      scope: meta.array.empty.php
+      captures:
+        1: punctuation.definition.array.begin.php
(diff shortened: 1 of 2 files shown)


wbond:
Part of this is that we need to do some work to standardize on `section` versus `definition`. From discussing it some with @FichteFoll, I believe we are going to use `definition` for atomic constructs, such as strings or variables. Things like blocks, arrays, functions, etc I believe we are going to use `section`.

This change will need to be made in many different syntaxes here in the default packages.

Another aspect of this is that I want to get the current state of Sublime Text's "standard" scope names documented for the sake of syntax and color scheme authors. I had been holding off to ensure I had a handle on the landscape. Having gone through a fair number of the default syntaxes, I can probably take a stab at a first version of that documentation soon.

wbond:
This and some other `[]` scoping were fixed in 410027a7dc96e71bc81f94494b365927aaddaa9c and 81606041f923dcae5caa37fce6aae3f98d46b2cc.

PR #524, opened 2016-07-13
[Go] Stops the default keyword from becoming entity.name.label

jakelucas:
A small change to fix the highlighting of the default keyword in the Go syntax.

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -23,7 +23,7 @@ contexts:
     - include: expressions-early
     - include: initializers
     - include: block
-    - match: ^\s*({{identifier}})(:)(?!=)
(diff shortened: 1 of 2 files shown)


keith-hall:
please can you add some more tests in to check that the `:` after `default` is correctly scoped as `punctuation.separator.go`, and that labels are (still) scoped as `entity.name.label.go`? because it seems as though there are/were no previously existing tests for this :)

jakelucas:
I've just pushed some more tests along with some minor changes to make sure that the `,` and `:` characters were being parsed as `punctuation.separator`.

If I've missed anything just let me know and I'll have another look.

wbond:
Thanks!

PR #528, opened 2016-07-14
[Python] Enhancements

SirNickolas:
[PEP 3127](https://www.python.org/dev/peps/pep-3127/) proposed binary and octal prefixes for integers.

[Waf](https://waf.io/) is a build system which is written in Python and uses Python scripts to control the building process.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -279,9 +279,9 @@ contexts:
       scope: constant.numeric.integer.long.hexadecimal.python
     - match: \b(?i:(0x\h*))
       scope: constant.numeric.integer.hexadecimal.python
-    - match: '\b(?i:(0o?[0-7]+)L)'
+    - match: '\b(?i:0(o|[0-7])[0-7]*L)'
       scope: constant.numeric.integer.long.octal.python
-    - match: '\b(?i:0o?[0-7]+)'
+    - match: '\b(?i:0(o|[0-7])[0-7]*)'
       scope: constant.numeric.integer.octal.python
     - match: \b(?i:0b[01]*L)
       scope: constant.numeric.integer.long.binary.python
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -278,10 +278,14 @@ contexts:
       scope: constant.numeric.integer.long.hexadecimal.python
     - match: \b(?i:(0x\h*))
       scope: constant.numeric.integer.hexadecimal.python
-    - match: '\b(?i:(0[0-7]+)L)'
+    - match: '\b(?i:(0o?[0-7]+)L)'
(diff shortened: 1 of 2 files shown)


wbond:
Thanks for contributing!

Could you add some tests for the integer forms?

buildersbrewery:
@SirNickolas http://www.sublimetext.com/docs/3/syntax.html#Testing

Python/Python.sublime-syntax
```diff
@@ -14,6 +14,7 @@ file_extensions:
   - gyp
   - gypi
   - Snakefile
+  - wscript
```
keith-hall: my understanding from reading https://waf.io/book/#_using_several_scripts_and_folders is that `wscript` is not a file extension, but the full filename of a Waf project description file.

keith-hall: and these files should be picked up as Python by the `first_line_match` anyway: https://waf.io/book/#_declaring_waf_commands

SirNickolas:
> `wscript` is not a file extension, but the full filename of a Waf project description file.

Indeed it is.

> and these files should be picked up as Python by the `first_line_match` anyway

`wscript`s are not executables, and they are not invoked directly, but via `waf` command. It compiles and `exec`s their code. So the first line may actually be any valid Python code.

keith-hall:
>> wscript is not a file extension, but the full filename of a Waf project description file.

> Indeed it is.

so it should be removed from the syntax definition's file extensions then, as it won't work

> the first line may actually be any valid Python code.

maybe there is no reliable way to do it without using a plugin to check the full filename

wbond:
> so it should be removed from the syntax definition's file extensions then, as it won't work

No, the Ruby syntax supports an extensive number of extension-less filenames. I believe the filename checker checks the whole filename if there is no `.` in it.

keith-hall:
> I believe the filename checker checks the whole filename if there is no . in it.

oh I didn't know that. But it does mean that it will incorrectly treat files with a wscript extension as python...

SirNickolas:
> as it won't work

Well, but it does work. Look, just a line above there is `Snakefile` definition, which is exactly the same - a file without an extension, just named "Snakefile".

> it will incorrectly treat files with a wscript extension as python

Who and for what reason will assign a `.wscript` extension to a file which is not a `wscript`?

wbond:
I just looked online and found no reference to the `.wscript` extension. There is `WScript.exe`, but it uses `.wsf`, `.vbs` or `.js`.

keith-hall:
In which case, I retract my comments :) sorry for causing a fuss!

wbond:
@keith-hall No worries, I (and I'm sure many others in the community) appreciate all of the time you take double-checking things!

wbond:
Excellent, thanks for this @SirNickolas!

PR #529, opened 2016-07-15
[Python] Relax regex comments a bit

SirNickolas:
Added a few punctuation symbols, none of which is a regex meta-character.

Also, I suppose dash should be escaped, because `[!-:]` actually means `[!"#$%&'()*+,\-./0123456789:]`.

### DIFF
--- a/Python/Regular Expressions (Python).sublime-syntax
+++ b/Python/Regular Expressions (Python).sublime-syntax
@@ -6,67 +6,77 @@ scope: source.regexp.python
 hidden: true
 contexts:
   main:
-    - match: '\\[bBAZzG]|\^|\$'
-      scope: keyword.control.anchor.regexp
-    - match: '\\[1-9][0-9]?'
-      scope: keyword.other.back-reference.regexp
-    - match: '[?+*][?+]?|\{(\d+,\d+|\d+,|,\d+|\d+)\}\??'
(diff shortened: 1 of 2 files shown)


wbond:
As I understand it, with these changes the following characters are the ones now not allowed in a regex comment:

 - `"`
 - `#`
 - `$`
 - `'`
 - `(`
 - `)`
 - `*`
 - `+`
 - `[`
 - `]`
 - `\`
 - `^`
 - `{`
 - `}`
 - `|`

In short, it seems like various regex operators and quotes.

What prompted you to not allow quotes after first allowing them?

Also, perhaps this character class would make more sense to users as a negated class since the list would be more obviously a set of characters that have meaning in regex?

SirNickolas:
Oh, of course negated class would be much clearer. I'll fix it, thanks.

I removed quotes after realizing that single line regexes may contain comments (or something that looks like comments) too.

That's how it renders in `master`:


In `py-regex`:


I also managed to get this behaviour:


Which one is better, in your opinion?

wbond:
I think the py-regex branch is fine.

Ideally we could improve the regex handling to know if we are in extended mode, but that currently has minimal return on investment compared to spending the effort on some other syntaxes which are currently rather broken.

SirNickolas:
Updated and got rid of lookbehind.

PR #532, opened 2016-07-18
[HTML] and [CSS] completions improvements

keith-hall:
- don't suggest HTML attribute names when typing inside a HTML attribute value string
- suggest CSS when typing at the end of a HTML style attribute, just before the closing quote

EDIT: also:
- make `style="..."` and `style='...'` consistent in HTML - previously, the single quotes were being scoped as `source.css`, which is incorrect.

### DIFF
--- a/CSS/css_completions.py
+++ b/CSS/css_completions.py
@@ -426,7 +426,7 @@ class CSSCompletions(sublime_plugin.EventListener):
     regex = None
 
     def on_query_completions(self, view, prefix, locations):
-        selector_scope = "source.css - meta.selector.css, text.html meta.attribute-with-value.style.html string.quoted punctuation.definition.string.end.html"
+        selector_scope = "source.css - meta.selector.css, text.html meta.attribute-with-value.style.html string.quoted - punctuation.definition.string.begin.html" # match inside a CSS document and inside the style attribute of a HTML tag, including just before the quote that closes the attribute value
         prop_name_scope = "meta.property-name.css"
         prop_value_scope = "meta.property-value.css"
         loc = locations[0]
--- a/CSS/css_completions.py
+++ b/CSS/css_completions.py
@@ -426,7 +426,7 @@ class CSSCompletions(sublime_plugin.EventListener):
     regex = None
 
     def on_query_completions(self, view, prefix, locations):
-        selector_scope = "source.css - meta.selector.css"
+        selector_scope = "source.css - meta.selector.css, text.html meta.attribute-with-value.style.html string.quoted punctuation.definition.string.end.html"
         prop_name_scope = "meta.property-name.css"
         prop_value_scope = "meta.property-value.css"
         loc = locations[0]
--- a/CSS/css_completions.py
+++ b/CSS/css_completions.py
@@ -426,7 +426,12 @@ class CSSCompletions(sublime_plugin.EventListener):
     regex = None
 
     def on_query_completions(self, view, prefix, locations):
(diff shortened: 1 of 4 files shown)


keith-hall:
Related bug report on SO: http://stackoverflow.com/q/38426242/4473405

CSS/css_completions.py
```diff
@@ -426,7 +426,7 @@ class CSSCompletions(sublime_plugin.EventListener):
     regex = None
 
     def on_query_completions(self, view, prefix, locations):
-        selector_scope = "source.css - meta.selector.css"
+        selector_scope = "source.css - meta.selector.css, text.html meta.attribute-with-value.style.html string.quoted punctuation.definition.string.end.html"
```
keith-hall: Let's imagine we have the following HTML:

    <div style="float: "></div>

and the caret is immediately before the closing `"` character. Using the `view.scope_name(point)` API / `view.match_selector(point, selector)` API (where the `point` is the beginning of the selection region) will return the scope of the `"` character, i.e. `punctuation.definition.string.end.html`

If the caret position is immediately after the `"` character, it would return the scope of the next character to the right, in this example case a `>`, so `punctuation.definition.tag.end.html`.

Therefore, as we want completion of CSS to still occur while the next character is the closing quote, we use the scope selector `punctuation.definition.string.end.html`, which will trigger just before the closing quote of the attribute, and not just after it :)

Probably to really see the difference, I suggest trying it out, both with and without this change. :)

wbond: In that case, shouldn't it be:

```
text.html meta.attribute-with-value.style.html string.quoted -punctuation.definition.string.start.html
```

So that we get completions for the whole string, but not before the opening quote?

keith-hall: that would work too, and is probably clearer (y)

my solution worked because the rest of the attribute is already covered by the `source.css`[...] selector.

keith-hall:
Updated the CSS completions file to use a more understandable scope selector, as discussed.

CSS/css_completions.py
```diff
@@ -426,7 +426,7 @@ class CSSCompletions(sublime_plugin.EventListener):
     regex = None
 
     def on_query_completions(self, view, prefix, locations):
-        selector_scope = "source.css - meta.selector.css"
+        selector_scope = "source.css - meta.selector.css, text.html meta.attribute-with-value.style.html string.quoted - punctuation.definition.string.begin.html" # match inside a CSS document and inside the style attribute of a HTML tag, including just before the quote that closes the attribute value
```
wbond: Can you place the comment on the previous line and wrap the string to 120 chars?

I realize it isn't obvious with the current public/private repo split, but I recently updated all plugin code shipped in the default packages to pass flake8 with a max line-length of 120 chars.

HTML/HTML.sublime-syntax
```diff
@@ -345,7 +345,8 @@ contexts:
         - match: "'"
           scope: string.quoted.single punctuation.definition.string.begin.html
           set:
-            - meta_scope: meta.attribute-with-value.style.html source.css
+            - meta_scope: meta.attribute-with-value.style.html
+            - meta_content_scope: source.css
```
wbond: We should probably add a test for this.

keith-hall:
@wbond done :)

wbond:
Great, thanks!

lp177:
Hi,

Is it possible to extend this pattern for match the style tag ? ( not trigger on build linux 3114 )

Issue #506, opened 2016-06-28
[Ruby] Incorrect scope of nested constants used outside of class definition contexts


mitio:
This might be related to the changes made because of the discussion in #458. The examples below are from build 3118.

Let's take the following simple example:

``` ruby
Foo::Bar
```

Here `Foo` has the correct scope of `source.ruby support.class.ruby`, but `Bar` is scoped only as `source.ruby` which is incomplete. In this case, `Bar` is also a constant (it might contain a class object, module object or some other value/object).

`Foo`:

<img width="267" alt="screenshot 2016-06-28 19 48 59" src="https://cloud.githubusercontent.com/assets/129307/16424557/60d64676-3d69-11e6-80ae-dcd012adedc9.png">

`Bar`:

<img width="247" alt="screenshot 2016-06-28 19 50 35" src="https://cloud.githubusercontent.com/assets/129307/16424617/98dd0064-3d69-11e6-9d40-f38765cefb5b.png">

Just for reference, in the following case only `Baz` (the last constant in the chain) is not scoped as a constant:

``` ruby
Foo::Bar::Baz
```

PR #533, opened 2016-07-18
[Scala] Fixed symbol indexing

djspiewak:
Addresses regression in scope name changes.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -93,27 +93,27 @@ contexts:
     - match: '\b(def)\s+({{id}})'
       captures:
         1: keyword.declaration.scala
-        2: entity.name.function.scala
+        2: entity.name.function.declaration.scala
       push: function-type-parameter-list
     - match: '(case)?\b(class|trait|object)\s+([^\s\{\(\[]+)'
       captures:
         1: keyword.declaration.scala
         2: keyword.declaration.scala
-        3: entity.name.class.scala
+        3: entity.name.class.declaration.scala
       push: class-type-parameter-list
     - match: '\b(type)\s+(([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`)'
       captures:
         1: keyword.declaration.scala
-        2: entity.name.type.scala
+        2: entity.name.type.declaration.scala
     - match: '\b(?:(val)|(var))\s+(([a-zA-Z$_][a-zA-Z0-9$_]*(_[^a-zA-Z0-9\s]+)?)|`.*`)?'
       captures:
         1: keyword.declaration.stable.scala
         2: keyword.declaration.volatile.scala
-        3: entity.name.val.scala
+        3: entity.name.val.declaration.scala
     - match: '\b(package object)\s+([^\s\{\(\[]+)'
       captures:
         1: keyword.declaration.scala
-        2: entity.name.class.scala
+        2: entity.name.class..declarationscala
     - match: '\b(package)\s+([\w\.]+)'
       scope: meta.package.scala
       captures:
--- a/Scala/Symbols.tmPreferences
+++ b/Scala/Symbols.tmPreferences
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Symbol List</string>
 	<key>scope</key>
-	<string>entity.name.function.declaration, entity.name.class.declaration, entity.name.val.declaration, entity.name.type.declaration</string>
+	<string>source.scala entity.name.function, source.scala entity.name.class, source.scala entity.name.val, source.scala entity.name.type</string>
 	<key>settings</key>
 	<dict>
 		<key>showInSymbolList</key>
--- a/Scala/Symbols.tmPreferences
+++ b/Scala/Symbols.tmPreferences
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Symbol List</string>
 	<key>scope</key>
-	<string>source.scala entity.name.function, source.scala entity.name.class, source.scala entity.name.val, source.scala entity.name.type</string>
+	<string>source.scala entity.name.function.declaration, source.scala entity.name.class.declaration, source.scala entity.name.val.declaration, source.scala entity.name.type.declaration</string>
 	<key>settings</key>
 	<dict>
 		<key>showInSymbolList</key>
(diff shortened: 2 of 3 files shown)


djspiewak:
Don't merge this yet. It's still not right, since non-declarations get highlighted as `entity.name.class`.  Uggghhhhhhh…

wbond:
Adding `.declaration` back in is not the correct solution. Instead you should be removing non-declaration instance of `entity.name`. `entity.name` is *only* for declarations.

djspiewak:
And what is the correct scope for non declarations?

wbond:
@djspiewak Of which token? For user-written code there is either no scope, or `support.class`, `support.type`, `support.namespace`, etc.

I've written up documentation for scope naming that will be part of the next release. I'm not sure exactly when that will be landing, however.

djspiewak:
@wbond Are you sure?  That's a fairly substantial change to the mode, in terms of how it highlights things in a user-visible way.  Monokai highlights `entity.name.class` as green, and `support.class` as italic and light blue.  In practice, this means that a significant percentage of the average `.scala` file just swapped from green and straight to blue and slanted, which is very jarring (at least to me) and not necessarily aesthetically pleasing.

wbond:
Yes, I am sure. I just spent around six months reading, writing and rewriting syntaxes and working on the highlighting engine itself.

You can certainly create a customized version of the Monokai if you want all of your class names to be green everywhere. I know changes to the visual "feel" of a syntax can be weird, but it seems Scala was just very much incorrectly written in the past. That is, unless I am misunderstanding and every instance of these class names is also a definition of them.

djspiewak:
It's not a definition of them, but we're in weird territory here.

There's another problem, which is that *most* class name-ish usage in code (i.e. not definitions) is actually objects or constants.  For example:

```scala
object Foo {
  val Bar = 42
}

println(Foo.Bar)
```

Focus on the usage site.  Currently, this highlights both `Foo` and `Bar` as `entity.name.class`, which is wrong both from the `entity.name` standpoint and because neither of them are classes.  `support.class` is the scope which is assigned on my branch now, but that's not right *either* because, again, neither of them are classes.

Semantically, unless found in type position (which is another ball of wax), upper-case identifiers are actually closer to Ruby constants than they are to Java classes.  The current modes appear to be a bit inconsistent about this.  Ruby highlights definition sites but not use sites for constants (which is probably correct).  Java highlights all instances of constant-ish things as `constant.other.java`, which in practice makes them look just like int literals (probably incorrect).  Scala currently highlights everything as `entity.name.class` (probably incorrect), but on my branch would highlight everything as `support.class` (exactly as incorrect).

I'm fine with things not being green.  But blue is not right either.

wbond:
The Java syntax is in poor shape and hasn't been rewritten either, I wouldn't use it as a benchmark for anything.

The Ruby syntax has been worked on recently. Due to the ambiguous nature, in cases where a token is known to be constant-ish thing and must be a class or module, it is highlighted as `support.class`. This isn't ideal, but then again there is no way in Ruby to know if it is a module or a class. Currently there is a bug about the gray area of constant-like identifiers that may be a class, module or constant – #506. It is pretty easy to detect the assignment of a constant, but not when it is dereferenced.

Due to the fact that we are dealing with hundreds of existing color schemes, we need to try and not stray too far from existing scopes, lest we break highlighting for everyone. Ideally we are going to need to add a few more "standard" scopes to deal with languages that push the boundaries a little. I'm fine with discussing that some more, and trying to come up with a migration plan, but for now we need to work within our current constraints.

Since I'm not familiar with all of the different ways that upper-case identifiers are used in scala, I don't know if context information would help. For instance, is `.` only ever to dereference a constant, or can it be used with classes also? In short, the options to work with are basically:

 - `support.type`
 - `support.class`
 - `support.constant`

I left out `support.function` since it is clearly wrong.

djspiewak:
`.` can be used to reference both classes and constants.  `#` may only be used to reference classes (i.e. it can only occur in type position).  Generally speaking, I think the following rules more-or-less apply:

- If it is a single "expression" (loosely defined for now) following a `\b:\b`, then it is in type position and pretty much everything is either `support.type` or `support.class` (no way to reliably distinguish)
- If it is within `[...]`, then it is in type position and the same rules apply
- If it is a single "expression" following a `new`, then it is a `support.class`.  Note that this is complicated, since things like the following are valid: `new (Foo ~> Bar)`.  Semantically, the `support.class` would be the `~>`, while the `Foo` and `Bar` are either `support.class` or `support.type` (no way to tell, syntactically)

Beyond that, we don't really know anything at all.  In other words, the fragment `Foo.Bar` without any other context could be `support.class` (e.g. if it is preceded by a `new`), or `support.type` (e.g. if it is preceded by a single `:`), or it could be dereferencing a constant-ish thing (`Bar`) from a constant-ish thing.

The safest thing to do, more than likely, is to try to build out a definition for "expression" in a type position (specifically, the three instances I referenced) and highlight those instances as either `support.class` or `support.type`.  Outside of those scopes, highlight upper-case things as `support.constant` or simply `source.scala`.

djspiewak:
Separate comment to propose migration strategy (and note that I'm willing to continue working on this).

- [ ] Remove incorrect upper-case highlighting from Scala altogether
- [ ] Refactor regexes to have an `ucase-id` and `lcase-id` and use them everywhere
- [ ] Enrich `new` highlighting with a new context for type parsing
- [ ] Enrich `[]` with the same context
- [ ] Find a way of reliably detecting (without massive performance penalties) preceding `:` to open type parsing context
- [ ] Highlight all other upper-case identifier stuff as `support.constant`, provided that `support.constant` isn't a really dumb-looking scope (if it is, just leave them as `source.scala` or something)
- [ ] Decide on appropriate scoping for pattern match and function parameters
- [ ] *(pie in the sky)* Find a clever way of detecting trailing `=>` so we can finally highlight lambda parameters correctly

The second point is really important, because the amount of insanity we are currently going through with inconsistent identifier detection is really dumb.

I am willing to (and would like to do) the work involved in this, but I don't really want to do it until after #491 lands, due to merge conflicts (and the fact that I have become personally reliant on the functionality in #491, so I can't dogfood any changes for the above list until it lands).  So I think that the best thing to do for #533 would be the first point on this list: remove the auto-magical upper-case highlighting stuff.  Completing the list will bring back a fairly significant, correct subset of the current (incorrect) highlighting.

wbond:
I'll fix the symbol indexing in such a way that it is consistent with other languages.

#491 is going to need a partial rewrite, just from looking at the scopes used in the tests. I'll spend some time learning about Scala so that we can move it forward.

djspiewak:
@wbond It's easy to fix the scopes in #491.  As I mentioned in the PR though, it's very very useful to have symbol indexing pick up pattern-match variables (`case`, `for` and `val`).  Considering that `val` was already (incorrectly!) indexed, that isn't a change so much as a bug fix.

wbond:
We should continue the discussion about #491 over there.

PR #539, opened 2016-07-20
[Python] scope escaped { and } characters in strings

### LINKED ISSUE #536: python: {{ and }} not handled correctly
```Python
>>> print '{{!}}{0}'.format(1)
{!}1
```

keith-hall:
fixes #536


SirNickolas:
Braces need escaping, don't they?
`\{\{|\}\}`

keith-hall:
Only if they are ambiguous. But, probably a good idea to escape them - I'll update it, thanks.

PR #540, opened 2016-07-21
[Regular Expressions] support relative capture groups

keith-hall:
`\g` allows to use `-` for previous capture groups
`(?\d)` allows to use `-` for previous capture groups and `+` for next capture groups

example:

> hellohellowowwow

would be matched by:

    (hello)\g-1(?+1)(wow)


FichteFoll:
You need to add this in line 163 as well.

keith-hall:
@FichteFoll good spot - thanks. I've pushed another commit with this and a test for extended mode.

PR #543, opened 2016-07-22
[LaTeX] Add match for \path command to make it behave like \url and \href

merlinthered:
Text inside the \path command was not treated special. This broke highlighting if the path contained special characters (e.g. $).

I'm not sure if there are multiple \path commands in different packages that behave differently. I am using the one from the hyperref package to typeset file paths. In my case, they sometimes contain environment variables, starting with $. Unlike for the \url command, which is highlighted correctly, a $ inside a \path would lead to the following text being highlighted as if it was set in math mode. This pull request fixes this.

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -573,7 +573,7 @@ contexts:
         - include: general-1-argument-with-options
 
   url:
-    - match: '((\\)(?:url|href))(\{)([^}]*)(\})'
+    - match: '((\\)(?:url|href|path))(\{)([^}]*)(\})'
       scope: meta.function.link.url.latex
       captures:
         1: support.function.url.latex
--- a/LaTeX/syntax_test_latex.tex
+++ b/LaTeX/syntax_test_latex.tex
@@ -100,6 +100,11 @@
 % ^ support.function.url.latex
 %     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ markup.underline.link.latex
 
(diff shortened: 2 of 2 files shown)


wbond:
I just merged #370, so this would need to be rebased on the current master, if it still applies.

FichteFoll:
#370 did not address this issue.

wbond:
@merlinthered Are you planning on seeing if you can revise this to work with the new, revised LaTeX syntax?

merlinthered:
I've rebased my change on the current master. The current master does not have the same problems with highlighting inside a \path command as the old syntax definition. However, I assume this is because the new syntax definition is not yet finished, as there are a number of other problems with highlighting now that did not occur before (In one of my files that I checked, some commands are not highlighted at all).

This pull request is still valid, because I think the \path command should be explicitly handled like the \url and \href commands, regardless of what the new syntax definition does in general.

FichteFoll:
@merlinthered, please do report these issues.

merlinthered:
@FichteFoll I've done so in #568. It was a special case that just happened to occur often in my code. It's not related to this issue at all.

wbond:
@merlinthered Could you add a test for your addition? I realize it is a simple addition, but I'm not super familiar with LaTeX, so extra tests help me feel better about merging in changes

merlinthered:
Done. I more or less just copy/pasted from the \url command because it should behave exactly the same. Added a $ in the test case because that's what made me do this in the first place.

wbond:
Thanks!

PR #546, opened 2016-07-24
[Regular Expressions] support conditionals

### LINKED ISSUE #517: [Regular Expressions] if/else conditionals marked as invalid
1. `[^a-zA-Z]`

Not latin letters

Different metasymbols `[` and `^` have identical scope `source.regexp keyword.control.set.regexp`. 
<br />

2.`[а-яА-ЯЁё]` 

Ёё

`Ёё` has to have same scope as at `А-Яа-я`, if not to specify its, some realization of regular expressions passes a Cyrillic letters `Ё` and `ё`. Range of letters of the Russian alphabet is specified in regular expressions as ` [а-яА-ЯЁё]`. For example, see [article](https://habrahabr.ru/post/123845/).
<br />
3. `(<[Aa]\s+[^>]+>\s*)?<[Ii][Mm][Gg]\s+[^>]+>(?(1)\s*</[Aa]>)`

Question mark

`?` highlighted how Invalid, its scope `source.regexp meta.group.regexp invalid.illegal.unexpected-quantifier.regexp `. But pattern is work, see https://regex101.com/r/rW8zR4/1 . Question mark with red invalid background found in other working regular expressions, for example:

Question mark 2
<br />
4. `(?<=[href|src]=")https?:\/\/.*?(?=")`

OR

Metasymbol `|` and literals `href` and `src` have identical scope `source.regexp meta.group.regexp`. 

Thanks.

keith-hall:
fixes #517

### DIFF
--- a/Regular Expressions/syntax_test_regexp.re
+++ b/Regular Expressions/syntax_test_regexp.re
@@ -387,6 +387,9 @@ hello++
 #                ^ keyword.operator.alternation.regexp
 #                 ^ meta.literal.regexp
 #                  ^ keyword.control.group.regexp
+(?(DEFINE)(?<a>abcd))(?&a)
+#^^^^^^^^^ keyword.other.conditional.definition.regexp
+#                     ^^^ keyword.other.backref-and-recursion.regexp
 
 (?x)
 (?<test>a)?b(?('test')c|d)
@@ -445,4 +448,7 @@ hello++
 #                ^ keyword.operator.alternation.regexp
 #                 ^ meta.literal.regexp
 #                  ^ keyword.control.group.regexp
+(?(DEFINE)(?<a>abcd))(?&a)
+#^^^^^^^^^ keyword.other.conditional.definition.regexp
+#                     ^^^ keyword.other.backref-and-recursion.regexp
 (?-x)
--- a/Regular Expressions/syntax_test_regexp.re
+++ b/Regular Expressions/syntax_test_regexp.re
@@ -324,3 +324,125 @@ hello++
 #    ^^ keyword.control.character-class.regexp
 #      ^ keyword.operator.quantifier.regexp
 #       ^^ constant.character.escape.regexp - keyword.control.set.regexp
+
+(?<test>a)?b(?('test')c|d)
+#            ^^^^^^^^^ keyword.other.backref-and-recursion.conditional.regexp
+(?<test>a)?b(?(<test>)c|d)
+#            ^^^^^^^^^ keyword.other.backref-and-recursion.conditional.regexp
+(a)?b(?(1)c|d)
+#     ^^^^ keyword.other.backref-and-recursion.conditional.regexp
+(?(?!\d)a|b)
+#^ keyword.other.conditional.regexp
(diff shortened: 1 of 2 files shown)


keith-hall:
maybe we should apply a separate meta scope to the group that starts with `(?(`, what do people think?
groups currently have `meta.group` or `meta.group.extended`, I was thinking perhaps `meta.group.conditional[.extended]`?
on a semi-related note, I was thinking a `|` should be `keyword.operator.alternation` instead of just `keyword.operator`, what do you think?

Regular Expressions/syntax_test_regexp.re
```diff
@@ -324,3 +324,83 @@ hello++
...
+# ^ keyword.control.group.regexp
+#  ^^ constant.other.assertion.regexp
+#    ^^ keyword.control.character-class.regexp
+#      ^ keyword.control.group.regexp
+#       ^^ keyword.control.character-class.regexp
+#         ^ meta.literal.regexp
+#          ^ keyword.operator.regexp
+#           ^ meta.literal.regexp
+#            ^ keyword.control.group.regexp
+(?(?<=\dc)a|b)
```
FichteFoll: Is `(?<=...)` a legit condition? It's not listed on http://www.boost.org/doc/libs/1_44_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html.

keith-hall: it is, it's a lookbehind :) unless I'm blind and have made a typo I can't spot atm :) http://www.boost.org/doc/libs/1_44_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html#boost_regex.syntax.perl_syntax.lookbehind

keith-hall: either way, it works for me in ST3's find panel :)

FichteFoll: Yes, but in [this section about conditionals](http://www.boost.org/doc/libs/1_44_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html#boost_regex.syntax.perl_syntax.conditional_expressions) it says only foward look-aheads are supported.

> condition may be either: a forward lookahead assert, the index of a marked sub-expression (the condition becomes true if the sub-expression has been matched), or an index of a recursion (the condition become true if we are executing directly inside the specified recursion).

keith-hall: hmm, even [the latest version of the page](http://www.boost.org/doc/libs/1_61_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html) doesn't mention it, but it works in ST3...

FichteFoll: Oh, by the way, the latest version 1.61.0 supports more things:
http://www.boost.org/doc/libs/1_61_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html#boost_regex.syntax.perl_syntax.conditional_expressions

FichteFoll: Yeah, it works for me as well in ST. Weird stuff.

PR #550, opened 2016-07-25
[Scala] Fixed object/class/trait/def/val/var/type keywords

djspiewak:
They now scope consistently with other modes, and with the semantics of the language itself.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -98,26 +98,26 @@ contexts:
         - include: main
     - match: '\b(def)\s+({{id}})'
       captures:
-        1: storage.type.function.scala
+        1: keyword.control.def.scala
         2: entity.name.function.scala
       push: function-type-parameter-list
     - match: '\b(case\s+)?(class|trait|object)(?:\s+([^\s\{\(\[]+))?'
       captures:
         1: keyword.other.declaration.scala
-        2: storage.type.class.scala
+        2: keyword.control.class.scala
         3: entity.name.class.scala
       push: class-type-parameter-list
     - match: '\b(type)\s+({{id}})'
       captures:
-        1: storage.type.scala
+        1: keyword.control.type.scala
         2: entity.name.type.scala
     - match: '\b(var)\s+({{id}})'
       captures:
-        1: storage.type.volatile.scala
+        1: keyword.control.volatile.scala
         2: entity.name.var.scala
     - match: '\b(val)\b'
       captures:
-        1: storage.type.stable.scala
+        1: keyword.control.stable.scala
       push:
         - match: '(?=[=\n])'
           pop: true
@@ -125,7 +125,7 @@ contexts:
     - match: '\b(package)\s+(object)\s+({{id}})'
       captures:
         1: keyword.control.scala
-        2: storage.type.class.scala
+        2: keyword.control.class.scala
         3: entity.name.class.scala
     - match: '\b(package)\s+([\w\.]+)'
       scope: meta.namespace.scala
@@ -327,13 +327,13 @@ contexts:
     - match: \b(extends|with|forSome)\b
       scope: keyword.declaration.scala
     - match: \b(class|trait|object)\b
-      scope: storage.type.class.scala
+      scope: keyword.control.class.scala
     - match: \bdef\b
-      scope: storage.type.function.scala
+      scope: keyword.control.def.scala
     - match: \btype\b
-      scope: storage.type.scala
+      scope: keyword.control.type.scala
     - match: \bvar\b
-      scope: storage.type.volatile.scala
+      scope: keyword.control.volatile.scala
     - match: \b(package)\b
       scope: keyword.control.scala
     - match: \b(new)\b
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -12,7 +12,7 @@ import fubar.{Unit, Foo}
 //            ^^^^ variable.import.scala
 
 def foo(a: Int, b: Bar): Baz = 42
-//^ storage.type.function.scala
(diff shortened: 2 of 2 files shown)


wbond:
No, these changes are not consistent with other syntaxes, or the official TextMate documentation on scopes.

If this is something you'd like to get behind and make a case for we'd be looking at changing the definitions on the many, if not a majority, of the languages in the default syntaxes.

A change like this will affect the highlighting of probably > 50% of Sublime Text users, so it isn't something I am going to merge without heavy buy in from popular color scheme authors.

djspiewak:
@wbond **THEN FIX THE RUBY MODE**

Is the Ruby mode highlighting `def` and `class` incorrectly?  All I did was change the Scala mode to highlight these keywords identically to the Ruby mode, which you have previously said was authoritative in its use of scopes.

The semantics of the Ruby keywords `def`, `class` and `module` are *precisely* the same as the semantics for Scala's `def`/`val`/`class`/`object`/etc.  There is no difference.  They are *not* analogous to the `var` keyword in C#, or the `void` keyword in C.  Highlighting these keywords in Scala as types is basically indefensible.  If the official TextMate documentation on scopes dictates that `def` in Scala should be highlighted as `support.type`, then either the documentation is wrong or your interpretation of it is.  Either way, it needs to change, because what you're proposing is analogous to highlighting `return` as `constant.numeric`.

wbond:
@djspiewak Yes, I do believe the Ruby scopes are currently incorrect. We've been working on fixing the vast inconsistencies for and bug in the syntaxes for over six months. You can look at https://github.com/sublimehq/Packages/graphs/contributors to see the amount of work that has gone into this repo since January.

There are a couple of things going on here that I think may be causing confusion and/or frustration around the issue of scope names:

 1. Scope names are largely historical, and those in heavy usage can't be broken wily nily (see `variable.function` for a function name)
 2. When talking about "keyword" and "type", the terms are somewhat loose, and aren't necessarily a 1-to-1 correlation. For instance, thinking about the "storage type" of an identifier in JavaScript. It might be assign an integer, or a string, or a function. Thus, while technically "def" isn't a type, it represents what type of a construct the identifier references.

I wish the process had more clarity and was simple enough that we could just search and replace various elements of the syntax definitions to improve them, however it is more of a complicated process because backwards compatibility is important. We have to temper this with consistency. Many of the syntaxes used completely different scopes, so when users changed languages, or when author created color schemes, there was much inconsistency.

I'll be the first to admit this process isn't 100% scientific, and I'm sure we've made missteps along the way. I am certainly open to discussing ideas for how to improve things. However, I must note that this process is something that involves many syntaxes and many end users, so it isn't always simple, or fast, and there aren't always clear answers. Sometimes I have to defer to the opinion of contributors who have been, and are continuing to contribute to these syntaxes as a whole. There are many users who want their one bug fixed, but we need to make sure the decisions being made that aren't just going to have to be undone in the near future.

djspiewak:
More broadly, I think the key point is that the following should *not* be colored identically:

```scala
def
Int
Foo
class
```

If we remove the `Int`, they actually all have the same scope!  That's very, *very*, very wrong and doesn't make sense for any language.  As I said, it's very analogous to saying that `return` should have the same scope as `42`, which is nonsense in all languages.

I do think that `support.type` makes sense for *some* instances of this sort of thing.  For example, in C#:

```c-sharp
int x = 42;
var y = 12;
```

It seems clear that `var` should be `support.type`, since it is a type placeholder.  C and Java, of course, don't have a notion of placeholders at all, and their declaration syntax always starts with a type or a modifier, so this discussion never comes up.  But it seems very clear that in Scala, Ruby, Python, SML, and many other languages where these declarations are simply syntax, it should be highlighted as a keyword.

This is what I mean by either the guidelines or the interpretation being wrong.  It's not that I'm sitting here saying that I would *prefer* one thing or another, it's that the guidelines are objectively wrong.  If the guidelines stated that strings which are surrounded by syntactic elements that are not `"` should never be highlighted (e.g. Scala's triple-quotes, or Ruby's single-quotes), wouldn't that be wrong?  Of course it would.

At the end of the day, scopes are all about "*this* looks like *that*".  It doesn't make any sense for `class` to look like `Foo`, under any circumstances.

djspiewak:
As a continuation to this…  In fairness, it does seem that some of my frustration on this point is simply related to Monokai itself not actually distinguishing certain scopes.  The fact that `storage.type` and `support.type` are highlighted identically is sort of a quick of the color scheme, and other color schemes appear to distinguish these elements.  `storage.type` isn't used for anything other than these specific markers, and so in general, it is a distinct coloring from `support.type`.

With that said, `def` and `class` in Scala really are very very much unrelated to `void` and `struct` in C, and so it's very weird that they would be lumped under the same syntactic guideline.

FichteFoll:
I am chiming in here while disregarding any effects that a change of `storage.type` to a `keyword.` scope could cause to many users. For me in particular, both scopes already have a very similar color.

This post is solely intended to find out the appropriate solutions without considering any debt from previous scoping conventions.

---

In Python, `def` and `class` *are* keywords as well, but they not only indicate a different type but also have different effects on the following block (and what goes into the parens following the identifier). To me, this is a pretty sure indicator that scoping both `def` and `class` as `storage.type` is wrong since they actually have keyword-like effects on code. `storage.type` should most likely only be used for staticly typed languages like C and Java.

The exception to this would be the `async` "keyword", which certainly should get a `storage` scope, although `storage.modifier`. Furthermore, the `global` and `nonlocal` keywords are semantically tied to referring to storage, so it makes sense for those to get a storage scope too. `storage.scope` or similar?

In JavaScript we have `var`,  `let` and `const`. I would most likely mark those as `storage.type` and give `class` and `function` a `keyword.` scope for the same reason as for Python.

PR #551, opened 2016-07-25
[Scala] Removed variables in case/for declarations from indexing

djspiewak:
The exception to this right now is the following:

```scala
for (x <- 42; val y = 24)
```

The `x` will not be indexed (which is the intent), but the `y` will be (which is unintended).  Removing the `val` will remove the `y` from the index and change the highlighting slightly, which is annoying.

This is a step forward though, and it constrains the index back to what it is on stable, which is to say only `val`/`var`/`class`/`trait`/`object`/`def`.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -121,7 +121,7 @@ contexts:
       push:
         - match: '(?=[=\n])'
           pop: true
-        - include: pattern-match
+        - include: val-pattern-match
     - match: '\b(package)\s+(object)\s+({{id}})'
       captures:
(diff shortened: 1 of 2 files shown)


djspiewak:
Rebased to remove the dependency on #550, in deference to the fact that style guideline changes take a really long time.

djspiewak:
@wbond Ping.  Reminder that this is a bugfix on indexing and no longer messes up the scoping of declaration keywords.  Also it's the head element in my PR dependency chain.

djspiewak:
Superseded by #563.

PR #561, opened 2016-07-31
[Scala] Highlighting of type expressions

djspiewak:
Dependent on #551.

This broadly fixes the issue with all upper-case tokens being highlighted as `support.class`, since in practice very few of them are.  Now, upper-case tokens in expression position are highlighted as `support.constant`, which is both idiomatic and far-and-away the most common scenario, while upper-case tokens in *type* position are highlighted as `support.class` and symbolic tokens in type position are highlighted as `support.type`.  For example:

```scala
val x: Foo1 => Baz1 = Foo2.bar(Baz2)
```

In the above, `Foo1` and `Baz1` are highlighted as `support.class`, `=>` is highlighted as `support.type` (though as discussed in another issue, `keyword.operator` might make more sense for this specific instance), and `Foo2` and `Baz2` are highlighted as `support.constant`.

This PR also fixes an issue with the following sort of expressions:

```scala
new Foo
```

In the above on master, `Foo` is highlighted as `entity.name.class`, which is incorrect since `new` signals instantiation, which is a usage site and not a declaration site.  The proper highlighting is `support.class`, which is the highlighting which is implemented by this PR.

There are some edge cases here if people use extremely non-standard style (e.g. random newlines will screw things up), but by and large it seems to work extremely well.  I have been using this locally on a number of code bases with non-trivial type expressions, and modulo some bugs (all of which I fixed), it seems to work extremely well and makes the highlighting far more pleasant.

Indexing is not changed by this PR, aside from the removal of instantiation from the index set (since it isn't a name declaration of any sort).  Note that there are a couple regexes here which can benefit from some factoring into variables, which is a change I made in another branch since it got tangled up with some other things.  More PRs incoming.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -200,6 +200,7 @@ contexts:
       push: function-tparams-brackets
     - match: '\]'
       pop: true
+    - include: type-constraints
     - include: delimited-type-expression
 
   function-parameter-list:
@@ -246,6 +247,7 @@ contexts:
       pop: true
     - match: this
       scope: variable.language.scala
+    - include: type-constraints
     - include: delimited-type-expression
 
   class-parameter-list:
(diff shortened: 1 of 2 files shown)


djspiewak:
Superseded by #563.

PR #562, opened 2016-07-31
[Scala] Lambda Highlighting

djspiewak:
Dependent on #561 

More or less mirrors the JavaScript highlighting.  For example:

```scala
xs map { x => x + 1 }
```

The `x` to the left of the `=>` is highlighted as `variable.parameter` (it's the parameter to the lambda), while `=>` is highlighted as `storage.type.function.arrow`, which mirrors JavaScript.  I thought about using a different scope here, since the scoping guidelines are a bit unclear on this point, but in practice this one seems to work extremely well.  I also considered changing the scope of the `=>` in `case ... =>` expressions to *also* be `support.type.function.arrow`, which wouldn't necessarily be inaccurate (though `case` does not really define a lambda), but in practice the different scoping actually helps a *ton* in reading code.  I was actually really surprised at the difference this makes.

Due to the way the lookahead works, the `=>` must be on the same line as the lambda parameters, which is very definitely idiomatic and far and away the most common way to write Scala.

This PR also adjusts a few pre-existing regexes that dealt with operator lookaround, notably the type prefix identification in #561.  Due to the fact that lambda parameter blocks introduce a new delimited type expression scope, this PR is inextricably dependent on #561.

I also moved comments into the prototype section, sort of randomly.  I was tired of including them everywhere.  There might be one or two lingering issues from this, but I think I got them all.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -200,6 +200,7 @@ contexts:
       push: function-tparams-brackets
     - match: '\]'
       pop: true
+    - include: type-constraints
     - include: delimited-type-expression
 
   function-parameter-list:
@@ -246,6 +247,7 @@ contexts:
       pop: true
     - match: this
       scope: variable.language.scala
+    - include: type-constraints
     - include: delimited-type-expression
 
   class-parameter-list:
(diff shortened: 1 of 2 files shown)


djspiewak:
Superseded by #563.

Issue #560, opened 2016-07-28
[Scala] "new Foo" highlighted as entity.name


djspiewak:
I have a fix for this locally that is part of a broader overhaul of type highlighting.  Just filing this bug to make the issue known.

`new` in Scala is purely instantiation, not declaration, so the correct highlighting for `Foo` in `new Foo` would be `support.class` or `support.type`.  `new` *can* declare new classes, but these classes are anonymous (identical to Java/C# semantics) and do not have a name.

wbond:
I believe this was fixed in #563

Issue #558, opened 2016-07-27
[Scala] Highlighting lambda parameters


djspiewak:
I'm honestly not certain if there even is an efficient way of doing this.  I'd like to be able to highlight lambda parameters as `variable.parameter`.  Unfortunately, the syntax is fairly… diverse.

```scala
{ a => ... }
{ (a, b) => ... }
{ a: Int => ... }
{ (a: Int, b: Int) => ... }

(a) => ...
(a, b) => ...
(a: Int) => ...
(a: Int, b: Int) => ...

a => ...
a: Int => ...
```

`a` and `b` are lambda parameters in all of the above examples.  The ellipses may be replaced by any expression (including other lambdas).

Basically, the marker token here is the `=>`, but we need to backtrack, perhaps quite a distance, in order to find the parameters.  That… doesn't sound like a recipe for efficiency.

One thing that might help here is I have a mostly-finished implementation of type expression parsing on my local fork, which eliminates the need to worry about `=>` appearing in types, and the pattern match parsing already eliminates the need to worry about it in `case` expressions.

wbond:
As long as there isn't a line break between the beginning of the parameters and the `=>`, we can handle it no problem. We tokenize one line at a time for efficiency sake, and to allow for resumable operation. This means we don't have to re-tokenize the entire file when you add a single character on line 200 of a 10,000 line file.

In terms of performance, Jon has written a very fast, custom built regex engine for syntax highlighting. It does not use backtracking, and thus is very fast – It can highlight the 6.6MB `sqlite.c` in around 900ms on my Mac. The contexts and matches required for handling this will likely be simplerthan what we use to try and handle C++ with C preprocessor macros thrown in.

djspiewak:
There *can* be a newline which separates the parameters and the `=>`.  For example:

```scala
{
(a: Int)
=>
42
}
```

That's a valid lambda.  But it's *very* unusual and would be considered bad style by almost everyone.  I'm completely ok with saying that lambdas defined in this way just don't get syntax highlighted.

What would this kind of functionality look like?  I've taken a crack at it in the past, and it's always fallen short in a few cases.

wbond:
I would start with the version JavaScript uses:

 - https://github.com/sublimehq/Packages/blob/master/JavaScript/JavaScript.sublime-syntax#L16
 - https://github.com/sublimehq/Packages/blob/master/JavaScript/JavaScript.sublime-syntax#L781-L784
[...]

Issue #549, opened 2016-07-25
[Scala] Everything upper-case is now support.class


djspiewak:
Wrong far more often than it is right.  I have a way to fix this, but it'll take somewhat longer.

wbond:
Previously it was `entity.name.class` which was definitely incorrect. The new scope is the "usage" equivalent of the previous scope.

If you can provide some examples of where it is incorrect we can determine how they fit in with the TextMate scope conventions to see if there is a better fit.

djspiewak:
I agree that `support.class` is slightly less wrong than `entity.name.class`.  So it's a step in the right direction.  This is mostly a ticket to track efforts to go further in that direction.

I can come up with a whole bunch of examples, but generally-speaking I would propose that we need to identify type expressions more explicitly.  Types can only appear in certain syntactically-delimited positions in Scala (e.g. within square brackets, following `:`, etc).  Outside of those positions, upper-case identifiers should probably be highlighted as `support.constant`, since that's going to be true 99.9% of the time in practical Scala (someone using upper-case for a variable is really just wrong, and it almost never happens).  For example:

```scala
val Foo = 42

Foo + 12
```

Thus, the first `Foo` would highlight as `entity.name.parameter`, as with any other constant variable declared with `val`, and the second `Foo` would highlight as `support.constant`.  A counter-example:

```scala
abc: Foo
```

This is a type ascription, and the `Foo` here should be highlighted as `support.type` or `support.class` (whatever makes more sense for sublime), and not as a constant.

I have a branch locally where I've been experimenting with this, and it seems that in nearly all cases, I can syntactically scope the type expressions, which means that my proposal is probably achievable.

djspiewak:
Here's my current proposal on this one: https://gist.github.com/aec7d52de2efbf983a2378b3d0733da6  This test passes on my local branch.  Ignore the `def`, `val` etc stuff, since obviously we're tracking that in #550.

wbond:
The new [Scope Naming documentation](http://www.sublimetext.com/docs/3/scope_naming.html) is probably worth spending some time reviewing.

Scoping `=>` feels wrong to me. Depending on its significance, I would expect `keyword.operator` or `punctuation.separator`. The first seems far more likely than the second. http://stackoverflow.com/a/7890032/230074 seems to indicate that it is a keyword also.

djspiewak:
[...]

PR #563, opened 2016-07-31
[Scala] Significant Improvements

djspiewak:
Several adjustments, fixes, enhancements and general better-ness:

- Indexing
  + `for`/`case` variables are no longer indexed
  + Initialization (i.e. `new`) are no longer erroneously indexed (#560)
  + Upper-case `val` declarations are indexed in base position, and considered constants in more complex patterns
- Type expressions are identified and parsed (subsumes #549)
  + Upper-case identifiers are now highlighted as constant references
  + Upper-case types are now highlighted as class references
  + Symbols in type position are highlighted as types
  + Underscores in type position… don't have a highlighting
  + Bounds (`<:`, `>:`, `<%`, `+`, `-` and `:`) are highlighted as special operators in type position
  + Dereferencing is identified (`#` and `.`)
  + Type lambdas are more stable now
  + `.type` is given special highlighting, since it is not a declaration when it appears in type position
  + `_*` is given special highlighting (also in pattern matching).  Note that we don't specially highlight the `*` type suffix yet, though we probably should.
- Lambdas are identified and highlighted (#558)
  + There are limitations to this.  Specifically, if your parameter block contains more than three nested parentheses sets, it will not be identified
  + Another limitation is that the parameters must be on the same line as the `=>` token
- Identifier regexes have been significantly improved
  + Invalid identifiers will no longer parse as operators (e.g. `:`, `_`, `<-` and others)
  + Prefix `_` characters are allowed (e.g. `_foo` is a valid identifier)
- The `_` operator is now scoped as a language variable in expression and pattern positions
- Assignment in all its forms has been given the appropriate scoping
- The `()` constant is now highlighted as such (similar to `true`/`false`)
- Numeric regexes have been rewritten entirely, based on the spec
  + Color schemes which highlight floats/integers/hex distinctly will now do so with Scala literals
  + The expression `42.toString` will no longer incorrectly highlight the `.` as part of the numeric
  + One deviation from the Scala spec here: `-12` will *not* highlight the `-` as part of the numeric.  This is sort of semantically justifiable (it can be viewed as a unary operator on `Int`), and it doesn't look horrible.  The main reason I did it is because it would require extreme contortion of the `op` regex
  + Octal literals are not highlighted, as they are no longer part of the language (e.g. `042` will not scope as anything).  I considered scoping these as `invalid.deprecated`, since that's exactly what they now are, but I can't remember how long ago they were removed.
- **All lookbehind has been removed.**  Previously, lookbehind was used for strings and imports, for no particularly good reason.  I would expect this should lead to significantly better performance, both on indexing and on highlighting, though I haven't actually noticed a difference.  Also, I have occasionally seen the indexer crash on some code (not very repeatably, and I saw it in master as often as on this PR), and when the indexer has crashed it produced a stack trace indicating that Oniguruma is still being used.  I wonder if we have some sort of regex weirdness somewhere that is forcing the fallback?
- Import matching has been completely rewritten and is now more robust
- `Nil` and `None` are no longer highlighted as primitives (since they aren't)
- A lot of random word boundary and keyword stuff that *was* broken is now fixed
  + As an example, classes/traits/objects who's name ends with "`extends`" or "`with`" no longer completely break highlighting at their declaration site
- Handling of default parameters in method and class definitions is now substantially more robust
- Comments are now more consistently handled (there may still be some latent bugs with this, but I think I got them all)
- Punctuation is now given the appropriate scoping (e.g. `.` is scoped as `punctuation.accessor.scala`).  Most color schemes ignore this, but it's good practice.
- Interpolated expressions within `${...}` blocks inside of strings now highlight with base scopes, rather than as if they were just part of the string.  (#580)
- `eq` is now considered a keyword (in particular, a word operator similar to Ruby's `and`)

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -7,26 +7,51 @@ file_extensions:
   - sbt
 scope: source.scala
 variables:
-  # lookahead/behind for operators
-  oplookbehind: '(?<=[[[:alpha:]]0-9\s\(\)\[\]\{\}])'
-  oplookahead: '(?=[[[:alpha:]]0-9\s\(\)\[\]\{\}])'
+  # lookaround for operators
+  nonopchar: '[[[:alpha:]]0-9\s\(\)\[\]\{\}]'
   # From http://www.scala-lang.org/files/archive/spec/2.11/01-lexical-syntax.html
   disallowed_as_operator: '[^\w\[\]\(\)\{\}''";,.`_\s]'
   operator_character: '[\p{Sm}\p{So}[{{disallowed_as_operator}}&&[\x20-\x7F]]]'
   upper: '[$_\p{Lu}]'
   # This is "letter", but without _ so we can ensure it is not last
   idcont: '[$\p{Lu}\p{Ll}\p{Lt}\p{Lo}\p{Nl}0-9]'
-  op: '(?:{{operator_character}}+)'
+
+  # {{operator_character}}+ \ {:, =, =>, <-, @, ←, ⇒, #}
+  op: |-
+    (?x:
+      [[^:=<@←⇒#]&&{{operator_character}}]+|
+      =[[^>]&&{{operator_character}}]+|
+      =>{{operator_character}}+|
+      <[[^\-]&&{{operator_character}}]+|
+      <-{{operator_character}}+|
+      [:@←⇒#]{{operator_character}}+
+    )
+
+  # {{operator_character}}+ \ {:, =, =>, <-, @, ←, ⇒, #, <%, <:, :<, +, -}
+  typeop: |-
+    (?x:
+      [[^:=<@←⇒#+\-]&&{{operator_character}}]+|
+      =[[^>]&&{{operator_character}}]+|
+      =>{{operator_character}}+|
+      <[[^\-%:]&&{{operator_character}}]+|
+      <-{{operator_character}}+|
+      :[^<]{{operator_character}}+|
+      :<{{operator_character}}+|
+      [@←⇒#+\-]{{operator_character}}+
+    )
+
   idrest: '(?:(?:{{idcont}}|_(?=[^{{operator_character}}]))*(?:_{{op}})?)'
   varid: '(?:\p{Ll}{{idrest}})'
   boundvarid: '(?:`{{varid}}`|{{varid}})'
   plainid: '(?:{{upper}}{{idrest}}|{{varid}}|{{op}})'
+  typeplainid: '(?:{{upper}}{{idrest}}|{{varid}}|{{typeop}})'
   id: '(?:{{plainid}}|`[^`]+`)'
+  typeid: '(?:{{typeplainid}}|`[^`]+`)'
   alphaid: (?:{{upper}}{{idrest}}|{{varid}})
   # Custom productions
   rightarrow: '=>|⇒'
   upperid: '(?:\b\p{Lu}{{idrest}})'
-  typeprefix: '{{oplookbehind}}(:){{oplookahead}}\s*'
+  typeprefix: '(:)\s*'
   # hack to cover up to three levels of nested parentheses
   withinparens: '\((?:[^\(\)]|\((?:[^\(\)]|\([^\(\)]*\))*\))*\)'
 
@@ -52,6 +77,8 @@ contexts:
     - include: keywords
     - include: imports
     - include: strings
+    - include: xml-literal
+    - include: operators
     - include: initialization
     - include: ascription
   main-post-lambdas:
@@ -60,7 +87,6 @@ contexts:
     - include: scala-symbol
     - include: empty-parentheses
     - include: braces
-    - include: xml-literal
     - include: late-keywords
 
   block-comments:
@@ -105,9 +131,8 @@ contexts:
   lambdas:
     - match: '(?=({{id}}|{{id}}\s*:\s*{{id}}|{{id}}\s*:\s*{{withinparens}}|{{withinparens}})\s*(?:{{rightarrow}})[[[:alpha:]]0-9\s\)\]\}])'
       push:
-        - match: '{{oplookbehind}}({{rightarrow}}){{oplookahead}}'
-          captures:
-            1: storage.type.function.arrow.scala
+        - match: '{{rightarrow}}'
+          scope: storage.type.function.arrow.scala
           pop: true
         - include: lambda-declaration
   lambda-declaration-base:
@@ -131,7 +156,7 @@ contexts:
   lambda-declaration:
     - match: '{{typeprefix}}'
       push:
-        - match: '(?={{oplookbehind}}({{rightarrow}}){{oplookahead}})'
+        - match: '(?={{nonopchar}}({{rightarrow}}){{nonopchar}})'
           pop: true
         - include: delimited-type-expression
     - include: lambda-declaration-base
@@ -177,7 +202,7 @@ contexts:
             - match: '\]'
               pop: true
             - include: delimited-type-expression
-        - match: '{{oplookbehind}}(<:|>:){{oplookahead}}'
+        - match: '(<:|>:)'
           scope: keyword.operator
           set:
             - match: '(?=[\n\}\)\]])'
@@ -392,6 +417,12 @@ contexts:
           pop: true
         - include: delimited-type-expression
 
+  # we need to greedily capture operators to prevent (e.g.) :: being matched as an ascription
+  # it would be incorrect to scope operators specially, since they are just identifiers
+  # by pulling them out HERE though and refusing to scope them, we emulate lookbehind
+  operators:
+    - match: '{{op}}'   # no explicit scope, just pulling it out
+
   initialization:
     - match: '\b(new)(?:\s+|\b)'
       captures:
@@ -465,7 +496,7 @@ contexts:
       scope: keyword.other.scala
     - match: \?\?\?
       scope: keyword.other.scala
-    - match: '{{oplookbehind}}(=){{oplookahead}}'
+    - match: '='
       scope: keyword.operator.assignment.scala
 
   late-keywords:
@@ -685,8 +716,10 @@ contexts:
         - match: \}
           pop: true
         - include: declarations
+    - match: =>|⇒
+      scope: keyword.operator.arrow.scala
   type-constraints:
-    - match: '{{oplookbehind}}(<:|>:|<%|\+|-|:){{oplookahead}}'
+    - match: '<:|>:|<%|\+|-|:'
       scope: keyword.operator.scala
   delimited-type-expression:
     - match: "[αβ]"     # just here for type lambdas
@@ -699,7 +732,7 @@ contexts:
     - match: '{{upperid}}'
       scope: support.class.scala
     - match: _
-    - match: '{{id}}'
+    - match: '{{typeid}}'
       scope: support.type.scala
     - include: base-type-expression
   single-type-expression:
@@ -713,7 +746,7 @@ contexts:
     - match: '{{upperid}}'
       scope: support.class.scala
       set: single-type-expression-tail
-    - match: '{{id}}'
+    - match: '{{typeid}}'
       scope: support.type.scala
       set: single-type-expression-tail
     - include: base-type-expression
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -47,6 +47,7 @@ variables:
   plainid: '(?:{{upper}}{{idrest}}|{{varid}}|{{op}})'
   typeplainid: '(?:{{upper}}{{idrest}}|{{varid}}|{{typeop}})'
   id: '(?:{{plainid}}|`[^`]+`)'
+  idorunder: '(?:{{id}}|_)'
   typeid: '(?:{{typeplainid}}|`[^`]+`)'
   alphaid: (?:{{upper}}{{idrest}}|{{varid}})
   # Custom productions
@@ -130,7 +131,7 @@ contexts:
       push: single-type-expression
 
   lambdas:
-    - match: '(?=(_|{{id}}|{{id}}\s*:\s*{{id}}|{{id}}\s*:\s*{{withinparens}}|{{withinparens}})\s*(?:{{rightarrow}})[[[:alpha:]]0-9\s\)\]\}])'
+    - match: '(?=({{idorunder}}|{{idorunder}}\s*:\s*{{idorunder}}|{{idorunder}}\s*:\s*{{withinparens}}|{{withinparens}})\s*(?:{{rightarrow}})[[[:alpha:]]0-9\s\)\]\}])'
       push:
         - match: '{{rightarrow}}'
           scope: storage.type.function.arrow.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -557,6 +557,10 @@ contexts:
       scope: keyword.operator.assignment.scala
     - match: '_'
       scope: variable.language.scala
+    - match: '\.'
+      scope: punctuation.accessor.scala
+    - match: ','
+      scope: punctuation.separator.scala
 
   nest-curly-and-self:
     - match: '\{'
@@ -687,6 +691,8 @@ contexts:
       scope: keyword.operator.other.scala
     - match: '_'
       scope: variable.language.scala
+    - match: ','
+      scope: punctuation.separator.scala
   val-pattern-match:
     - match: '{{upperid}}(?=[\s=])'
       scope: entity.name.parameter.scala
@@ -747,6 +753,8 @@ contexts:
       scope: keyword.operator.other.scala
     - match: '_'
       scope: variable.language.scala
+    - match: ','
+      scope: punctuation.separator.scala
 
   base-type-expression:
     - match: '\(\{\s*type\s+λ\[α(\[_\])?(,\s*β(\[_\])?)?\]\s*='
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -200,6 +200,7 @@ contexts:
       push: function-tparams-brackets
     - match: '\]'
       pop: true
+    - include: type-constraints
     - include: delimited-type-expression
 
   function-parameter-list:
@@ -246,6 +247,7 @@ contexts:
       pop: true
     - match: this
       scope: variable.language.scala
+    - include: type-constraints
     - include: delimited-type-expression
 
   class-parameter-list:
@@ -566,7 +568,7 @@ contexts:
       push:
         - match: \]
           pop: true
-        - include: main
+        - include: delimited-type-expression
     - match: '@|_'
       scope: keyword.other.scala
 
@@ -593,6 +595,9 @@ contexts:
         - match: \}
           pop: true
         - include: main
+  type-constraints:
+    - match: '(?<=[a-zA-Z0-9\s\)\[\]\}])(<:|>:|<%|\+|-|:)(?=[a-zA-Z0-9\s\)\]\}])'
+      scope: keyword.operator
   delimited-type-expression:
     - match: "[αβ]"     # just here for type lambdas
       scope: comment.block.empty.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -227,7 +227,7 @@ contexts:
         1: storage.type.scala
         2: entity.name.type.scala
       push:
-        - match: '(?=[\n\}\)\]])'
+        - match: '(?=[\n;\}\)\]])'
           pop: true
         - match: '\['
           push:
@@ -243,7 +243,7 @@ contexts:
         - match: '='
           scope: keyword.operator.assignment.scala
           set:
-            - match: '(?=[\n\}\)\]])'
+            - match: '(?=[\n;\}\)\]])'
               pop: true
             - include: delimited-type-expression
     - match: '\b(var)\s+({{id}})'
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -108,7 +108,7 @@ contexts:
       scope: constant.character.literal.scala
 
   comments:
-    - match: (//).*$\n?
+    - match: (//).*$
       scope: comment.line.double-slash.scala
       captures:
         1: punctuation.definition.comment.scala
@@ -118,7 +118,7 @@ contexts:
       scope: punctuation.definition.comment.scala
       push:
         - meta_scope: comment.block.documentation.scala
-        - match: \*/(\s*\n)?
+        - match: \*/\s*$
           scope: punctuation.definition.comment.scala
           pop: true
         - match: (@\w+\s)
@@ -314,12 +314,10 @@ contexts:
       push: function-tparams-brackets
     - match: ':'
       push:
-        - match: '(?=[\B\s]=[\B\s])'
+        - match: '(?=[\{\};\n]|{{nonopchar}}={{nonopchar}})'
           pop: true
         - include: delimited-type-expression
-        - match: '(?=[\{\n])'
-          pop: true
-    - match: '(?=[\{=\n])'
+    - match: '(?=[\{\};\n]|{{nonopchar}}={{nonopchar}})'
       pop: true
 
   function-tparams-brackets:
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -106,23 +106,30 @@ contexts:
             1: storage.type.function.arrow.scala
           pop: true
         - include: lambda-declaration
-  lambda-declaration:
+  lambda-declaration-base:
     - match: \(
       push:
         - match: \)
           pop: true
-        - include: lambda-declaration
+        - include: lambda-declaration-parens
+    - match: '{{id}}'
+      scope: variable.parameter
+  lambda-declaration-parens:
     - match: '{{typeprefix}}'
       push:
-        - match: '(?=(?<=[a-zA-Z0-9\s\)\]\}])(=>)(?=[a-zA-Z0-9\s\)\]\}]))'
-          pop: true
         - match: ','
           pop: true
         - match: '(?=\))'
           pop: true
         - include: delimited-type-expression
-    - match: '{{id}}'
-      scope: variable.parameter
+    - include: lambda-declaration-base
+  lambda-declaration:
+    - match: '{{typeprefix}}'
+      push:
+        - match: '(?=(?<=[a-zA-Z0-9\s\)\]\}])(=>)(?=[a-zA-Z0-9\s\)\]\}]))'
+          pop: true
+        - include: delimited-type-expression
+    - include: lambda-declaration-base
 
   base-types:
     - match: \b(Unit|Boolean|Byte|Char|Short|Int|Float|Long|Double)\b
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -521,8 +521,6 @@ contexts:
       scope: keyword.other.scala
     - match: \?\?\?
       scope: keyword.other.scala
-    - match: '='
-      scope: keyword.operator.assignment.scala
 
   late-keywords:
     - match: \b(extends|with|forSome)\b
@@ -537,6 +535,8 @@ contexts:
       scope: storage.type.volatile.scala
     - match: \b(package)\b
       scope: keyword.control.scala
+    - match: '='
+      scope: keyword.operator.assignment.scala
 
   nest-curly-and-self:
     - match: '\{'
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -490,11 +490,13 @@ contexts:
   strings:
     - match: '"""'
       push:
+        - meta_include_prototype: false
         - meta_scope: string.quoted.triple.scala
         - match: '"""(?!")'
           pop: true
     - match: (?<!\\)"
       push:
+        - meta_include_prototype: false
         - meta_scope: string.quoted.double.scala
         - match: '"'
           pop: true
@@ -506,6 +508,7 @@ contexts:
       captures:
         1: support.function.scala
       push:
+        - meta_include_prototype: false
         - meta_scope: string.quoted.triple.interpolated.scala
         - match: '"""(?!")'
           pop: true
@@ -514,6 +517,7 @@ contexts:
       captures:
         1: support.function.scala
       push:
+        - meta_include_prototype: false
         - meta_scope: string.quoted.interpolated.scala
         - match: '"'
           pop: true
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -828,4 +828,4 @@ contexts:
       scope: punctuation.accessor.scala
       set: single-type-expression
     - match: '(?=\S)'
-      pop: true
+      set: try-dispatch       # this is needed for initialization (new ...) and doesn't HURT elsewhere
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -121,6 +121,7 @@ contexts:
     - match: '{{typeprefix}}'
       push:
         - match: ','
+          scope: punctuation.separator.scala
           pop: true
         - match: '(?=\))'
           pop: true
@@ -182,6 +183,7 @@ contexts:
               pop: true
             - include: delimited-type-expression
         - match: '='
+          scope: keyword.operator.assignment.scala
           set:
             - match: '(?=[\n\}\)\]])'
               pop: true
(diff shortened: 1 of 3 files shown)


gwenzek:
I'd feel totally lost if `Nil` and `None` where not highlighted as `constant.language`

Scala/Scala.sublime-syntax
```diff
@@ -608,18 +608,25 @@ contexts:
...
     - match: '{{id}}'
       scope: support.type.scala
-      pop: true
+      set: single-type-expression-tail
     - include: base-type-expression
     - match: '(?=[\s,\)\}\]])'
       pop: true
+  single-type-expression-tail:
+    - match: '[\.#]'
+      scope: punctuation.separator
```
wbond: This should probably be `punctuation.accessor`.

Scala/Scala.sublime-syntax
```diff
@@ -260,6 +260,8 @@ contexts:
       push:
         - match: '\)'
           pop: true
+        - match: '\b(val)\b'
```
wbond: No capture is required here

Scala/Scala.sublime-syntax
```diff
@@ -22,6 +22,7 @@ variables:
   alphaid: (?:{{upper}}{{idrest}}|{{varid}})
   # Custom productions
   upperid: '(?:\b\p{Lu}{{idrest}})'
+  typeprefix: '(?<=[a-zA-Z0-9\s\)\]\}])(:)(?=[a-zA-Z0-9\s\)\]\}])\s*'
```
wbond: We don't use lookbehinds since it forces highlighting to fall back to the slow oniguruma regex engine

Scala/Scala.sublime-syntax
```diff
@@ -249,16 +304,26 @@ contexts:
                 3: variable.import.renamed-to.scala
             - match: '([^\s.,}]+)'
               scope: variable.import.scala
+
   inheritance:
     - match: '(extends|with)\s+([^\s\{\(\[\]]+)'
       captures:
         1: keyword.declaration.scala
         2: entity.other.inherited-class.scala
+    - match: '(extends|with)\s+\('
```
wbond: There should probably be a leading `\b` before `(`

Scala/Scala.sublime-syntax
```diff
@@ -452,18 +524,26 @@ contexts:
     - match: '@|_'
       scope: keyword.other.scala
   val-pattern-match:
+    - match: '{{upperid}}'
+      scope: entity.name.parameter
```
wbond: `.scala` need to be added to the end of this scope name.

Scala/Scala.sublime-syntax
```diff
@@ -111,6 +122,14 @@ contexts:
       captures:
         1: storage.type.scala
         2: entity.name.type.scala
+      push:
+        - match: '\n'
+          pop: true
+        - match: '='
```
wbond: This should have a scope, such as `keyword.operator.assignment.scala`

Scala/Scala.sublime-syntax
```diff
@@ -156,35 +181,56 @@ contexts:
...
-        - match: '({{alphaid}})\s*:\s*'
+        - match: '({{alphaid}})(?=\s*:)'
           captures:
             1: variable.parameter.scala
+          push:
+            - match: '{{typeprefix}}'
+              set:
+                - match: '(?=[,=\)])'
+                  pop: true
+                - match: ','
```
wbond: This should have the scope `punctuation.separator.scala`

Scala/Scala.sublime-syntax
```diff
@@ -156,35 +181,56 @@ contexts:
...
+                  pop: true
+                - include: delimited-type-expression
         - include: main
-    - match: '(?=[:\{=\n])'
+    - match: ':'
```
wbond: Should have a scope, probably `punctuation.separator.scala`.

Scala/Scala.sublime-syntax
```diff
@@ -194,31 +240,40 @@ contexts:
...
     - match: '(?=([\{\n]|extends))'
       pop: true
+
   class-tparams-brackets:
     - match: '\['
       push: class-tparams-brackets
     - match: '\]'
       pop: true
-    - include: main
+    - match: this
```
wbond: This should probably have `\b` before and after

Scala/Scala.sublime-syntax
```diff
@@ -194,31 +240,40 @@ contexts:
...
-        - match: '({{alphaid}})\s*:\s*'
+        - match: '({{alphaid}})(?=\s*:)'
           captures:
             1: variable.parameter.scala
+          push:
+            - match: '{{typeprefix}}'
+              set:
+                - match: '(?=[,=\)])'
+                  pop: true
+                - match: ','
```
wbond: Needs a scope name

wbond:
I was going through commits and making comments, but it appears they have all been edited by later commits. You can disregard those and I'll make comments on the current state of the PR.

Scala/Scala.sublime-syntax
```diff
@@ -7,6 +7,9 @@ file_extensions:
   - sbt
 scope: source.scala
 variables:
+  # lookahead/behind for operators
+  oplookbehind: '(?<=[[[:alpha:]]0-9\s\(\)\[\]\{\}])'
```
wbond: We don't use lookbehinds because it forces us to fallback to the slow oniguruma regex engine. Instead, push into a context whenever a construct could be valid.

Scala/Scala.sublime-syntax
```diff
@@ -77,25 +94,62 @@ contexts:
...
+      push:
+        - match: \)
+          pop: true
+        - include: lambda-declaration-parens
+    - match: '{{id}}'
+      scope: variable.parameter
+  lambda-declaration-parens:
+    - match: '{{typeprefix}}'
+      push:
+        - match: ','
```
wbond: Needs a scope, probably `punctuation.separator.scala`

Scala/Scala.sublime-syntax
```diff
@@ -77,25 +94,62 @@ contexts:
...
+        - match: '(?={{oplookbehind}}(=>){{oplookahead}})'
+          pop: true
+        - include: delimited-type-expression
+    - include: lambda-declaration-base
+
+  base-types:
+    - match: \b(Unit|Boolean|Byte|Char|Short|Int|Float|Long|Double)\b
+      scope: storage.type.primitive.scala
   constants:
+    - match: '(?<![[[:alpha:]]0-9`\)])\(\)'
```
wbond: Needs to be reconstructed so it doesn't use a lookbehind

Scala/Scala.sublime-syntax
```diff
@@ -111,6 +165,25 @@ contexts:
...
+            - match: '\]'
+              pop: true
+            - include: delimited-type-expression
+        - match: '{{oplookbehind}}(<:|>:){{oplookahead}}'
+          scope: keyword.operator
+          set:
+            - match: '(?=[\n\}\)\]])'
+              pop: true
+            - include: delimited-type-expression
+        - match: '='
```
wbond: Needs a scope, probably `keyword.operator.assignment.scala`

Scala/Scala.sublime-syntax
```diff
@@ -140,12 +213,25 @@ contexts:
             - match: '\)'
               pop: true
             - include: pattern-match
-        - match: '=>|⇒|\b(if)\b'
+        - match: '\b(if)\b'
+          captures:
+            1: keyword.control.flow.scala
+          set:
+            - match: '=>|⇒'
```
wbond: Needs a scope, possibly a `keyword` variant (I realized this may not be 100% correct according to scala semantics, but I think it would make sense to most users)

Scala/Scala.sublime-syntax
```diff
@@ -140,12 +213,25 @@ contexts:
...
+              pop: true
+            - include: main-no-lambdas
+        - match: '=>|⇒'
```
wbond: Same here as above

Scala/Scala.sublime-syntax
```diff
@@ -156,35 +242,57 @@ contexts:
...
           pop: true
         - include: main
+
   function-type-parameter-list:
     - match: '(?=\()'
       set: function-parameter-list
     - match: '\['
       push: function-tparams-brackets
-    - match: '(?=[:\{=\n])'
+    - match: ':'
```
wbond: Needs a scope, probably `punctuation.separator.scala`

Scala/Scala.sublime-syntax
```diff
@@ -156,35 +242,57 @@ contexts:
...
-        - match: '({{alphaid}})\s*:\s*'
+        - match: '({{alphaid}})(?=\s*:)'
           captures:
             1: variable.parameter.scala
+          push:
+            - match: '{{typeprefix}}'
+              set:
+                - match: '(?=[=\)])'
+                  pop: true
+                - match: ','
```
wbond: Needs a scope, probably `punctuation.separator.scala`

Scala/Scala.sublime-syntax
```diff
@@ -194,31 +302,43 @@ contexts:
...
+          scope: storage.type.scala
+        - match: '({{alphaid}})(?=\s*:)'
           captures:
             1: variable.parameter.scala
+          push:
+            - match: '{{typeprefix}}'
+              set:
+                - match: '(?=[=\)])'
+                  pop: true
+                - match: ','
```
wbond: Needs a scope, probably `punctuation.separator.scala`

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
       push:
         - match: '`'
           pop: true
-    - match: '\b[a-z][a-zA-Z0-9_]*\.'
-    - match: '\.[a-z][a-zA-Z0-9_]*\b'
+    - match: '{{varid}}\.'
+    - match: '\.{{varid}}'
```
wbond: The `.` should probably get the scope `punctuation.accessor.scala` in both above rules

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
     - match: '\b{{varid}}'
       scope: entity.name.parameter.scala
+    - match: \[
+      push:
+        - match: \]
+          pop: true
+        - include: main
+    - match: '@|_'
```
wbond: Is `_` really a keyword, or is it like many other languages where it is a convention for an unused variable?

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+      scope: keyword.other.scala
+  val-pattern-match:
+    - match: '{{upperid}}(?=[\s=])'
+      scope: entity.name.parameter
```
wbond: Need `.scala` at the end of the scope

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
       scope: support.class.scala
+
+  pattern-match:
+    - include: keywords
+    - include: constants
+    - include: char-literal
+    - include: scala-symbol
+    - include: strings
+    - include: xml-literal
+    - match: '`'
```
wbond: This should have some scopes - is it a string of some sort?

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+      push:
+        - match: '`'
+          pop: true
+    - match: '{{varid}}\.'
+    - match: '\.{{varid}}'
```
wbond: Probably `punctuation.accessor.scala` for the `.` in the above two matches

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+    - match: \[
+      push:
+        - match: \]
+          pop: true
+        - include: delimited-type-expression
+    - match: \{
+      push:
+        - match: \}
+          pop: true
+        - include: main
```
wbond: Usually `main` isn't the right scope to include in such a situation (although sometimes it is). Can this really contain any scala construct?

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+  type-constraints:
+    - match: '{{oplookbehind}}(<:|>:|<%|\+|-|:){{oplookahead}}'
+      scope: keyword.operator
```
wbond: Need `.scala` at the end

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+  delimited-type-expression:
+    - match: "[αβ]"     # just here for type lambdas
+      scope: comment.block.empty.scala
+    - match: '[\.#]'
+      scope: punctuation.separator
```
wbond: Should probably be `punctuation.accessor.scala`

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+    - include: base-types
+    - match: '\b(forSome)\b'
+      scope: keyword.declaration.scala
+    - match: '{{upperid}}'
+      scope: support.class.scala
+    - match: _
```
wbond: In another place this was `keyword` (and I left a comment asking about that), just make sure the two situations are consistent or at least correct.

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+      scope: support.type.scala
+      set: single-type-expression-tail
+    - include: base-type-expression
+    - match: '(?=[\s,\)\}\]])'
+      pop: true
+  single-type-expression-tail:
+    - match: '[\.#]'
+      scope: punctuation.separator
+      set: single-type-expression
+    - match: '\s+'
```
wbond: This seems like an odd rule to include

Scala/Scala.sublime-syntax
```diff
@@ -440,16 +578,123 @@ contexts:
...
+      set: single-type-expression-tail
+    - match: '{{id}}'
+      scope: support.type.scala
+      set: single-type-expression-tail
+    - include: base-type-expression
+    - match: '(?=[\s,\)\}\]])'
+      pop: true
+  single-type-expression-tail:
+    - match: '[\.#]'
+      scope: punctuation.separator
```
wbond: This should probably be `punctuation.accessor.scala`, or at least needs `.scala` at the end

wbond:
Ok, I've left notes on items I noticed from a cursory glance

Scala/Scala.sublime-syntax
djspiewak: Is there a better solution?  I need to be able to highlight `:`

Scala/Scala.sublime-syntax
djspiewak: It should.  The normal use-case for it doesn't matter, but it could be a subset of another word.

Scala/Scala.sublime-syntax
djspiewak: Ehhhhhh… no.  I think we might be able to get away with `declarations`.  I'll have to look at what's there though.

Scala/Scala.sublime-syntax
djspiewak: That's going to be a problem.  `:` and `=>` could be valid following almost any token.  There really isn't a way for us to backtrack over a context (is there?).

Scala/Scala.sublime-syntax
djspiewak: I'm trying to decide whether or not it should be highlighted as a `keyword` in all scopes, since it has different meanings in different positions (e.g. it absolutely should *not* be highlighted as `keyword` in a type context).  I could see the argument for highlighting it as a keyword in all expression and pattern contexts though.

Scala/Scala.sublime-syntax
djspiewak: I'm not sure I follow…

Scala/Scala.sublime-syntax
djspiewak: I thought about it.  In fact, I was going to make a change, but in practice highlighting the lambda `=>` differently from the version of the same operator is surprisingly helpful in reading code.  They *are* different semantics (the `case` version doesn't declare a function).  `keyword.operator` might make sense though.

Scala/Scala.sublime-syntax
```diff
@@ -249,16 +368,26 @@ contexts:
...
+      captures:
+        1: keyword.declaration.scala
+      push:
+        - match: '\)'
+          pop: true
+        - include: delimited-type-expression
+
   initialization:
-    - match: '\b(new)\s+({{id}})'
+    - match: '\b(new)\s+'
```
djspiewak: I think `late-keywords` already takes care of that, but it's reasonable to do it here.

Scala/Scala.sublime-syntax
djspiewak: It depends on the context.  In pattern match position, it is a convention for an unused variable except when it is followed by a `*`, in which case it denotes a varargs extraction.  However, because `_` is not a valid *dereference* identifier, it's a slightly stronger convention than one would normally assume.  Also, there is no checking for repeated use of `_` in patterns, whereas a normal identifier would be forbidden.

So in other words, yeah, it really is a keyword, just a particularly odd one.

Scala/Scala.sublime-syntax
djspiewak: It denotes a strictly referenced identifier, as opposed to an identifier that may be rebound.

djspiewak:
I think the biggest issue you raised is the lookbehind.  This is a problem for both `:` and `=>` since these operators are important to identify in advance (in the case of `=>`, quite significantly in advance), and they can follow almost anything, *and* they are valid suffixes (and prefixes) of other operators (e.g. `+:`, `::`, `==>`, etc).

I'm not 100% certain how to resolve that with contexts, or if it's even possible without significantly reorganizing the mode.

Scala/Scala.sublime-syntax
wbond: Backtracking is generally bad for performance.

Instead, consume the `::` and the `==>` with a match rule. Then the second `:` and the trailing `=>` won't be possible to be matched against.

Scala/Scala.sublime-syntax
wbond: Now that we aren't matching the id here, if we do add the `\b` we can remove it from `late-keywords`.

Scala/Scala.sublime-syntax
wbond: That doesn't sound like a keyword to me, just a "special" identifier. If you wanted to highlight it specially, I would expect a scope of `variable.language.scala` might be more accurate.

Scala/Scala.sublime-syntax
wbond: Ok, so the `` ` `` should probably be scoped as `punctuation.definition.identifier.scala` then. Or possibly some other `punctuation.definition.` scope.

Scala/Scala.sublime-syntax
djspiewak: There should, but you'll note that this line is almost a copy/paste of the line above it, which has been in the mode since the dawn of time.  :-)

wbond:
The equivalent of a positive lookbehind is to push into a special context after the item you were looking backwards for. This context should have a match rule containing the match, minus the look behind. Then there should be a blank match (or a non-whitespace lookahead) to pop out of the context.

The equivalent of a negative lookbehind is to push into a special context after the item you were looking backwards for and match the alternative meaning of the symbol (assuming that is why you are using a negative lookbehind). Again, use a blank match (or a non-whitespace lookahead) to pop out of the context.

Moving forward we are committed to removing all lookbehinds and other incompatible patterns from our default packages. This is because they have a significant impact on performance, especially in regards to indexing projects.

wbond:
Here is an example of the approach I mentioned: https://github.com/sublimehq/Packages/blob/master/JavaScript/JavaScript.sublime-syntax#L393-L417

It handles the situations in JavaScript where a `{` can be the beginning of a block or object literal, and `/` can be the division operator or the start of a regex.

djspiewak:
I understand the lookbehind equivalences.  The problem we have here though appears in the following expression:

```scala
(x: Int, y: Int) => x
```

There are actually two *nested* instances of lookbehind here.  The first is with the `:` operator.  As the user types this from left to right, the `x: Int` and `y: Int` fragments will be parsed as type ascription, as the `:` operator is identified in expression position.  But as soon as the `=>` operator is entered, the meaning of the parentheses block completely changes and `:` is no longer ascription, but type denotation for the `x` and `y` parameters.

In other words, if I find a `=>`, I will need to "unconsume" a set of tokens prior to it.  There is a similar problem with the `for` comprehension matching, but I was able to resolve it with lookahead since the unconsumed token set always stretches to the beginning of the line.

Scala/Scala.sublime-syntax
wbond: Then let’s add it to both! :smile:

In general I tried not to deviate into fixes that need to be made that were not touched by this PR (there are many related to parentheses and brackets, plus meta scopes). In this case, it is pretty minor, so we might as well include them both.

Scala/Scala.sublime-syntax
djspiewak: `variable.language.scala` would be fine.

Scala/Scala.sublime-syntax
djspiewak: Well, the problem is that those operators were just random examples that I constructed on the fly.  Scala allows *arbitrary* symbolic operators (truly arbitrary).  The only restriction is that a couple operators are reserved.  Single-character `:` and `=>` are among them.  This is literally the symbolic analogue to keywords (e.g. `def` is a valid identifier, but the keyword takes precedence).  The problem is that we can't leverage `\b` to pull out symbolic keywords ahead of time, which is why I needed the backtracking in the first place.

In general, I hate backtracking.  I've done a lot of academic work in the field of parser theory, so I'm very familiar with how horrible it is.  :-)  I just couldn't find another solution in this particular case.

Scala/Scala.sublime-syntax
```diff
@@ -77,25 +94,62 @@ contexts:
...
+        - match: ','
+          pop: true
+        - match: '(?=\))'
+          pop: true
+        - include: delimited-type-expression
+    - include: lambda-declaration-base
+  lambda-declaration:
+    - match: '{{typeprefix}}'
+      push:
+        - match: '(?={{oplookbehind}}(=>){{oplookahead}})'
```
wbond: You seem to have a lookahead containing a lookbehind here, BTW.

djspiewak: Yep.  Lookception.  I did that just because of laziness.  I do need the lookahead, but obviously due to the lookahead, the lookbehind (and also the nested lookahead) are unnecessary.

Scala/Scala.sublime-syntax
wbond: I don't have time right ATM to look into the specifics of your implementation, but I do believe we can use a context push/pop to handle situations where you only want matches to happen if the preceding characters matched a certain pattern.

Here are some more examples (I linked to the JavaScript one in the main PR comments):

 - https://github.com/sublimehq/Packages/commit/dba0eaed9285d36590013ea5c5135fd2b90f9642
 - https://github.com/sublimehq/Packages/blob/master/Ruby/Ruby.sublime-syntax#L1042-L1061
 - https://github.com/sublimehq/Packages/blob/master/Ruby/Ruby.sublime-syntax#L316-L379

wbond:
> In other words, if I find a =>, I will need to "unconsume" a set of tokens prior to it.

Use a lookahead to detect that situation in advance.

djspiewak:
> Use a lookahead to detect that situation in advance.

Actually, even lookahead is problematic because I need to be able to test for arbitrarily nested parentheses.  I'm covering that now by pretending that "arbitrary" is "3", but that's obviously limited.

After refamiliarizing myself with this…  The lookbehind is really just to catch operator matching, not lambda block identification.  Lambdas are identified solely with lookahead.  The operator matching still seems extremely problematic though.  None of the examples you've linked look like they even approach the problem in Scala.  I'm not giving up, I just don't see a way of doing it yet.

djspiewak:
@wbond I think I fixed the lookbehind issue.  At the very least, the tests pass.  I also adjusted the semantics in a couple areas that were not quite correct.  I basically rejiggered the `op` (and now also `typeop`) regular expressions to exclude primitive operators via progressive negation (no backtracking).  I expect this will be extremely fast, though it is ugly.

One thing that I *anticipate* you might disagree with is the `operators` context, which simply greedily parses out operator identifiers *before* any of the "other" operator parsing can kick in, but without assigning a scope.  I strongly believe these are the correct semantics.  `&!!^%` is no different of an identifier than `fubar`, and should not be highlighted differently when in common use.  So I simply capture the match and let it fall through as `source.scala`, exactly the same way that `fubar` simply falls through altogether uncaptured and unscoped.

`=>` is now highlighted as `keyword.operator.arrow.scala`, which is uniform with the way it is highlighted in imports, except when it is the delimiter for a lambda (where it is highlighted as `storage.type.function.arrow`).  `=` is highlighted as `keyword.operator.assignment.scala`, as is `<-`.  I haven't finished going through all of the uses of `_`, but I think I've gotten them into a slightly more sane state.  Remember that this symbol does have very different meaning in different contexts (it is a wildcard type, a wildcard import, a "no name" pattern match, a function value delimiter, and more).  I'm not exactly sure what scope to give it when used as an expression; either `keyword.operator.other` or `variable.language.scala` could be correct.

Oh yeah, and I've given up on adjusting the separate PRs.  Since you left your comments here, I figured it was better to just update this one.  So everything is on the `feature/unit` branch.

There is still exactly one use of lookbehind: highlighting `()`.  I think this is pretty easily handled though by having a pre-scope, as this case is really quite similar to the regex matching case in Python.  It wouldn't be the end of the world if we couldn't highlight `()`.  It's really nice, but if we can't do it, I won't cry over its loss.

Another thing I haven't done yet is take an overview of *all* of the instances where punctuation could be matched.  For example, `.` in expression position is uncaptured, and is thus scoped inconsistently with `.` in type position.  That should be cleaned up to make the mode more internally consistent.

Oh yeah…  Could you take a look at the `withinparens` variable and tell me if I'm missing an obvious way to do this better?  I tried self-referential groups from oniguruma and it didn't work.  This is required for the lambda lookahead.  There is a test case which illuminates the need for it.

Finally, there's still a TODO in the test suite related to `val` inside of a `for` with parentheses rather than braces.  I mean to fix that, though it is a lower priority since this particular syntactic form is incredibly rare in practice.

Thoughts on removing `Nil` and `None` from the `constant.language` scope?  I know @gwenzek doesn't like the idea.  :-)  They aren't primitive in any sense of the word (they're just elements of the standard library).  The correct scope for them would be `support.constant` according to the scope guidelines, and this is precisely the scope they will be assigned if I simply remove the hard-coding.

djspiewak:
@wbond Oh, it also appears that we were *already* using negative lookbehind to handle interpolated strings.

djspiewak:
@wbond Sorry for the spam…  All lookbehind has now been removed.  Ready for second review.

gwenzek:
@djspiewak I won't mind a `support.constant`, I just want some kind of highlighting for `None` and `Nil`  :-) 

For me the difference between `support.constant` and `constant.language` isn't quite clear anyway.

And I like the idea of scoping lambda parameters, even if some edge cases aren't supported.

But I think that scoping operators like `+`, `===` and more complex ones like the Scalaz operators really improves code readability.
Maybe you could use a special scope so the users can decide in their color scheme.

djspiewak:
@gwenzek It would be relatively trivial to scope operators differently from regular identifiers (though it would get weird if you did things like `+(...)`, which is legal).  What scope would you give them though?  I don't see symbolic code as being any different from ascii, and from a semantic standpoint, it certainly isn't.  Though there is some precedent, since ascii types highlight as `support.class` and symbolic types as `support.type`.

I wouldn't mind tossing a meta scope on symbolic operators or something like that.  I'm curious as to what @wbond thinks.

`support.constant` is what all the other upper-case things are now scoped as.  So just nuking the `constant.language` bit would make `Nil` and `None` fall through to that catch-all, which is more accurate in my opinion.

djspiewak:
Numerics were driving me nuts.  Fixed them (also made the scopes align with the guidelines).  It's a self-contained change, so we can pull it out if we want.

gwenzek:
@djspiewak I was just thinking about operator in infix position.
For `+(...)` scoping them like functions might be more logical. Maybe `variable.function.operator` to give more control to the user.

djspiewak:
@gwenzek I thought about it, but doing that is really no more accurate than styling infix alphabetical names as `variable.function`.  In other words, it might be true, and it's likely to be true more often than not, but it's also not really guaranteed to be true and I think it is likely to distract more than it will add value.

djspiewak:
I added scoping for the `_*` "magic" type ascription and pattern match, which is a special type that indicates the presence of varargs (but doesn't actually denote varargs in and of itself).  Currently it's scoped as `keyword.operator.other.scala`, which I'm not convinced is the appropriate scoping.  It's not a `support.type`, but it's also not really a `keyword`.  `support.type` would probably look better in most color schemes, but it's hard to say.

djspiewak:
@wbond I renamed the PR and updated the description to match, since it has sort of become a bit of a grab bag (mostly stemming from the fact that all the review happened here, and thus so did all the fixes).  I tried to be as comprehensive as possible in the changelog, both for your benefit and the benefit of interested observers.

jrappen:
To be frank, I haven't checked the differences between this and #94, but if this superseeds #94 maybe @gwenzek could close that other PR as well?

djspiewak:
@jrappen This PR is an incremental change (albeit a large one) from the current Scala mode.  #94 is a rewrite from scratch.

Glancing at some of the features listed on #94…  What is still missing here includes the following:

- Unicode character escapes (e.g. `\u2101`)
- Scoping for annotations (outside of primitives like `@volatile`)
- Better handling for XML
- Meta scoping
- Sane scaladoc highlighting

#94 has numerous other features as well, but I personally would not be interested in them.  But nevertheless, it seems clear that this PR does not supersede #94.

Scala/Scala.sublime-syntax
```diff
@@ -111,6 +226,26 @@ contexts:
...
+      push:
+        - match: '(?=[\n;\}\)\]])'
+          pop: true
+        - match: '\['
+          push:
+            - match: '\]'
+              pop: true
+            - include: delimited-type-expression
+        - match: '(<:|>:)'
+          scope: keyword.operator
```
wbond: Missing `.scala` at the end

djspiewak: I think there's still a few of these (non-`.scala` scopes) floating around.  I'll try to take a pass through and clean them all up.

Scala/Scala.sublime-syntax
```diff
@@ -194,71 +370,97 @@ contexts:
...
+      push:
+        - match: '\)'
+          pop: true
+        - include: delimited-type-expression
+
+  # we need to greedily capture operators to prevent (e.g.) :: being matched as an ascription
+  # it would be incorrect to scope operators specially, since they are just identifiers
+  # by pulling them out HERE though and refusing to scope them, we emulate lookbehind
+  operators:
+    - match: '{{op}}'   # no explicit scope, just pulling it out
```
wbond: Why not scope these operators as `keyword.operator.scala`?

djspiewak: Well, they aren't keywords any more than `JFrame` is a keyword in Java.  They're just identifiers like any other.  The only reason that I'm matching them *at all* is to ensure that I can match the primitive operators (the ones that *are* keywords) later on, since the primitive operators are valid substrings of non-primitive operators (and should not be matched in that case).  It's basically a backwards way of emulating word boundaries for symbolic identifiers.

wbond:
I think one of the last questions I have before merging this in is if `val` identifiers are referred to as "parameters". Right now the scope for these is `entity.name.parameter` – should it be `entity.name.val`?

djspiewak:
@wbond `val` identifiers are *not* referred to as "parameters".  I think I've been keeping it as `entity.name.parameter` because you kept it as `entity.name.parameter` because I set it that way back several PRs ago without knowing what the better scope would be.

`entity.name.val` might make sense, or `entity.name.stable` or something like that.

wbond:
Thanks for all of the work that went into this!

It sounds like the next steps after merging this are probably to look through #94 and pull in some features that @gwenzek implemented in that rewrite.

I just wanted to note @gwenzek, I appreciate your time in reviewing these changes and everything you are contributing. I'm sorry that the effort you put into the full rewrite in #94 won't be used directly, but I am sure we will pull in at least some of your features and tests.

djspiewak:
@wbond @gwenzek I'll open an issue to track some of the things that I know of in #94 that we would really like to pull in.  Some variant of [this](https://github.com/sublimehq/Packages/pull/563#issuecomment-239681953) probably.

gwenzek:
Thanks @wbond  and congrats to both of you for having greatly improved Scala support in Sublime !

Now that the syntax seems more stable, I'll start on porting the improvements of #94 that are sill relevant.
I should have some time to work on this next week

keith-hall:
@djspiewak sorry if this comes across as nit-picky, but please could I ask you to prefix all your future commits with the language tag i.e. `[Scala]` just like we do for issues and PRs. It just makes it so much easier to browse the recent commits / commit history and tell which syntax(es) were affected :) thanks!

PR #570, opened 2016-08-04
[Batch File] support set with /p switch

keith-hall:


Batch File/Batch File.sublime-syntax
```diff
@@ -101,6 +101,20 @@ contexts:
               - include: command_set_group
               - include: variables
           - include: command_set_group
+      - match: '\s+/[pP]\s+' # SET from standard input
+        push:
+          - match: '(?=\n)'
```
wbond: This should probably be `'(?=$\n|{{command_terminators}})'`, or maybe a variation that does allow `<`?

keith-hall: you are absolutely right, thanks!

keith-hall:
Thanks for the feedback @wbond, I have pushed an updated commit with your suggestion :)

PR #576, opened 2016-08-08
[XML] add [sublime-snippet] syntax definition

keith-hall:
...which will highlight as plain XML initially, until the content is reached, whereby it will also highlight (nested) placeholders and their optional regular expression replacements

### DIFF
--- a/XML/Snippet Content.sublime-syntax
+++ b/XML/Snippet Content.sublime-syntax
@@ -3,6 +3,7 @@
 # See http://www.sublimetext.com/docs/3/syntax.html
 scope: text.sublime-snippet-content
 name: Sublime Text Snippet Content
+hidden: true
 contexts:
   main:
     - include: possible_caret
--- a/XML/Snippet Content.sublime-syntax
+++ b/XML/Snippet Content.sublime-syntax
@@ -1,0 +1,78 @@
+%YAML 1.2
+---
+# See http://www.sublimetext.com/docs/3/syntax.html
+scope: text.sublime-snippet-content
+name: Sublime Text Snippet Content
+contexts:
+  main:
+    - include: possible_caret
+
+  possible_caret:
+    - match: '\\\$'
+      scope: constant.character.escape.sublime-snippet-content
+    - match: '\$(?=\{|\w)'
+      scope: keyword.other.caret.sublime-snippet-content
+      push: caret
+    - match: '\$'
+      scope: invalid.illegal.unescaped-dollar.sublime-snippet-content
+
+  caret:
+    - match: '(?:(\d+)|(PARAM\d+|SELECTION|TM_CURRENT_LINE|TM_CURRENT_WORD|TM_FILENAME|TM_FILEPATH|TM_FULLNAME|TM_LINE_INDEX|TM_LINE_NUMBER|TM_SELECTED_TEXT|TM_SOFT_TABS|TM_TAB_SIZE)\b)' # http://docs.sublimetext.info/en/latest/extensibility/snippets.html?highlight=snippet#environment-variables
+      captures:
+        1: constant.numeric.sublime-snippet-content
+        2: constant.language.sublime-snippet-content
+      pop: true
+    - match: '\{'
+      scope: keyword.other.caret.begin.sublime-snippet-content
+      set: [inside_caret, caret]
+    - match: '\w+'
+      scope: constant.other.sublime-snippet-content #invalid.illegal.unknown-variable.sublime-snippet-content
+      pop: true
+
+  inside_caret:
+    - match: '\}'
+      scope: keyword.other.caret.end.sublime-snippet-content
+      pop: true
+    - include: possible_caret
+    - match: ':'
+      scope: keyword.operator.alternation.sublime-snippet-content
+    - match: '/'
+      scope: keyword.operator.regexp.begin.sublime-snippet-content
+      push:
+        - match: '/'
+          scope: keyword.operator.regexp.end.sublime-snippet-content
+          set:
+            - match: '(/)([gims-]*)(?=\})' # x mode doesn't seem to work properly in snippets
+              captures:
+                1: keyword.operator.replacement.end.sublime-snippet-content
+                2: storage.modifier.mode.regexp.sublime-snippet-content
+              pop: true
+            - include: regex_replacement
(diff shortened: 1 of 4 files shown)


keith-hall:
Note that the syntax test file deliberately doesn't have a `sublime-snippet` extension - so that ST doesn't try to read it as an actual snippet! (it would fail due to being invalid XML as it contains the syntax test header line...)

jrappen:
Maybe it would make sense to suggest a fixed syntax test file extension to Jon and Will?

XML/Snippet Content.sublime-syntax
```diff
@@ -0,0 +1,78 @@
...
+          set:
+            - match: '(/)([gims-]*)(?=\})' # x mode doesn't seem to work properly in snippets
+              captures:
+                1: keyword.operator.replacement.end.sublime-snippet-content
+                2: storage.modifier.mode.regexp.sublime-snippet-content
+              pop: true
+            - include: regex_replacement
+        - include: scope:source.regexp#base-literal
+
+  regex_replacement: # http://www.boost.org/doc/libs/1_56_0/libs/regex/doc/html/boost_regex/format/boost_format_syntax.html
```
keith-hall: Note that, if [regex replacements in the find and replace panel supported the same functionality as snippets](https://forum.sublimetext.com/t/conditional-insertions-using-editors-find-replace/13871) then this section would be better off in it's own syntax file in the Regular Expressions package. This would [enable the regex replacement syntax to be used to syntax highlight the replace field in the find and replace dialog](https://forum.sublimetext.com/t/regex-syntax-highlighting-in-the-replace-field/169). also https://github.com/SublimeTextIssues/Core/issues/1079

keith-hall:
> Maybe it would make sense to suggest a fixed syntax test file extension to Jon and Will?

I get by with the following tiny plugin, which, when a syntax_test file is opened, will set the syntax to the one specified in the syntax test header line - maybe it could be added to the Default package that ships with ST3:

    import sublime, sublime_plugin
    
    class HighlightSyntaxTestsListener(sublime_plugin.EventListener):
        def on_load_async(self, view):
            if view.file_name().startswith('syntax_test_'):
                first_line = view.substr(view.line(0))
                comment, file = first_line.split(' SYNTAX TEST ', 1)
                file = file.split('"')[1]
                view.set_syntax_file(file)

XML/Snippet.sublime-syntax
```diff
@@ -0,0 +1,26 @@
+%YAML 1.2
+---
+# See http://www.sublimetext.com/docs/3/syntax.html
+file_extensions:
+  - sublime-snippet
+name: Sublime Text Snippet
+scope: text.xml.sublime-snippet
+contexts:
+  main:
+    - match: '(<)(content)(>)(<!\[CDATA\[)'
```
FichteFoll: The `<![CDATA[` part may be provided on the next line (or even be skipped entirely) afaik.

FichteFoll:
Linking a few reference implementation of snippet syntaxes (which you probably checked already):

- https://github.com/SublimeText/PackageDev/blob/dab5f66346cd9e36e08210d8d67a3bc1811e8eea/Syntax%20Definitions/Sublime%20Snippet%20(Raw).YAML-tmLanguage
- https://github.com/bobthecow/sublime-sane-snippets/blob/a0891a9e46f15afa5fc30570f3d93e8d55740b05/SaneSnippet.YAML-tmLanguage

FichteFoll:
By the way, I'd like to use this for the next PackageDev update by the way, unless Jon decides he wants special highlighting for ST's internal configuration to ship by default – if I may.

Related: https://github.com/sublimehq/Packages/pull/95

XML/Snippet.sublime-syntax
keith-hall: @FichteFoll thanks for the heads up - I've done some testing and you are correct. Also, if more than one CDATA section is used, ST ignores any but the first, so I've marked any others as illegal.

wbond:
I think this probably makes sense to add. I will check it out further and try it on my machine. From a quick glance at the code, I am wondering if a suffix of `.snippet` would make more sense than `.sublime-snippet-content`. The later seems very long/wordy.

FichteFoll:
If it makes sense to add this, does that mean that syntax definitions for other resource files should be added as well (thinking about sublime-syntax, sublime-keymap, sublime-commands, sublime-mousemap, sublime-menu, sublime-project) ...?

XML/Snippet.sublime-syntax
```diff
@@ -0,0 +1,54 @@
+%YAML 1.2
+---
+# See http://www.sublimetext.com/docs/3/syntax.html
+file_extensions:
+  - sublime-snippet
+name: Sublime Text Snippet (XML)
+scope: text.xml.sublime-snippet
```
keith-hall: Note that the `new_snippet` command in `Packages/Default/new_templates.py` currently references the normal XML syntax, and this should be changed in the ST3 codebase after merging.

    v.assign_syntax('Packages/XML/XML.sublime-syntax')

keith-hall:
@wbond there are currently 2 syntax definitions included in this PR, one for the snippet XML and one for the `<content>` tag inside the XML. It is useful to split them because maybe one day, a syntax definition for keybindings could reference the snippet content syntax for the `insert_snippet` command. Currently, the scopes are `sublime-snippet` and `sublime-snippet-content`. We could shorten them, but I would like us to make sure that it is clear which scope refers to what. Maybe `snippet` and `snippet-content` would achieve both goals?

FichteFoll:
In packagedev we use raw-snippet iirc.

keith-hall:
Did you ever try this out @wbond ? :) The main reason why I'd like this definition added is because a lot of people put an unescaped `$` in their snippets and wonder why it doesn't work the way they expect.

wbond:
I haven't yet - been primarily working on non-package stuff recently.

wbond:
Based on the decision in #95, I think I am going to decide to not include this for now. My understanding is that https://packagecontrol.io/packages/PackageDev is the goto package to include augmented syntaxes and helpers. I think that makes sense for the time being.

We may revisit this in the future and decide that including syntaxes and helpers for the built in formats makes sense. For now, users can use the default XML syntax, or install something like Package Dev.

PR #577, opened 2016-08-10
Add entity.name.constant to Java symbol index

ismell:
I realize the current syntax definition doesn't use this yet. I'm working on
upstreaming my Java syntax changes.


wbond:
This isn't currently used, so there isn't value in adding it now. If you have some changes that will benefit from this, we can add it as part of that PR.

Issue #434, opened 2016-05-26
Rewrite Shell Script (Bash)


wbond:
Probably one of the buggiest syntaxes right now is `Shell Script (Bash)`. Some current issues/PRs:
 
 - #46 
 - #47
 - #87
 - #157
 - #190
 - #231
 - #318
 - #394

Work was done on improving the current state of things by @jfcherng with https://packagecontrol.io/packages/ShellScriptImproved. He indicated that he didn't think that was maintainable long term, hence this issue.

Issues that need to be addressed with the rewrite:

 - [ ] Update regular expressions to only use new engine
 - [ ] Consume identifiers including `-` so partial commands/functions aren't highlighted
 - [ ] Create variations for TCSH and ZSH (http://hyperpolyglot.org/unix-shells)
 - [ ] Fix first line detection
 - [ ] Fix detection of function definitions
 - [ ] Detect command/operator names and don't highlight when used as params (e.g. `dd if`)
 - [ ] Handle `$()`, `$(())` and ``` `` ``` expansions
 - [ ] Heredocs

There are surely many more.

michaelblyons:
Wishlist:

- [ ] `$(())` can include other expressions.

      ```sh
     foo=$(($(cat bar.txt | wc -l) + 5))
      ```

- [ ] Escape sequences within strings recognize the "color" patterns and don't just capture the `\e`.

      ```sh
     echo -e "\e[31;44mRed on Blue\e[0m"
      ```

jfcherng:
Here's one that I do not fix.
```bash
empty=$((echo hello) > output.txt)
```

MattDMo:
It would be great if the new syntax had a separate scope for ALLCAPS_CONSTANTS. (`[[:upper:]][[:upper:][:digit:]_]*` should do it IIRC.)
Also, don't dump everything into `keyword.other.shell`, it's rather pointless. It would also be kind of cool if you could match unquoted strings containing `/` as paths, but that may have too many exceptions to be useful.

jrappen:
looking at #46 @jfcherng linked [this test file](https://gist.githubusercontent.com/okdana/45d3fd437e2f97f2ac63/raw/a822842edf8acdb7af4d182bbe18545f36d4c7be/bashsyntax.sh) by @okdana

jfcherng:
[...]

Issue #261, opened 2016-03-29
Inconsistencies with .sublime-settings' "extensions" and the syntax def's "file_extensions" lists


FichteFoll:
The following packages have differences in this regard:

- C++
- Lisp
- XML

Imo they should all be in the syntax definitions.

---

Additionally, the XML package's settings file sports a setting named `hidden_extensions`. What does it do exactly?

Briles:
The `hidden_extensions` setting seems to hide the extensions when opening a file of type or saving a file as type.

The `hidden_extensions` will not be listed with the XML syntax for the following commands:
- `File > Save As`
- `File > Open File`

These are the only effects I could find after limited testing; there may be more.

wbond:
Removed `extensions` after speaking with Jon. At this point that setting should only be used by users to configure certain file types to open with a specific syntax.

FichteFoll:
All right.

`.sublime-syntax` files don't support an equivalent to the `hidden_extensions` setting, do they?

I'm not sure where I would want to use it though. Maybe someone else can thing of an example.

FichteFoll:
Created a separate issue for this at https://github.com/SublimeTextIssues/Core/issues/1326.

The most practical use case seems to be for a long list of extension support (for automatic highlighting) such as https://github.com/sublimehq/Packages/pull/581 without cluttering the file dialogs, as @Briles explained.

PR #581, opened 2016-08-13
[ShellScript] add missing extensions

jrappen:
see commit for details

### DIFF
--- a/ShellScript/Shell-Unix-Generic.sublime-syntax
+++ b/ShellScript/Shell-Unix-Generic.sublime-syntax
@@ -3,18 +3,56 @@
 # http://www.sublimetext.com/docs/3/syntax.html
 name: Shell Script (Bash)
 file_extensions:
-  - sh
+  - ackrc
+  - aliases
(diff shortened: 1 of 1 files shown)


jrappen:
addresses #434

FichteFoll:
I think the long extension (or filename list) for shellscript should be using the `hidden_extensions` setting instead of adding them all to the syntax. See https://github.com/sublimehq/Packages/issues/261.

wbond:
Some of these are just plain wrong, such as [editorconfig](http://editorconfig.org/).

wbond:
I apologize if my comment came off as harsh.

By scanning the list, it seemed like many of these are not shell scripts, but rather various config formats - some of them like INI, others single-line regex patterns.

The Shell Script syntax is in line for an overhaul, and it desperately needs it. However, after the overhaul I don't want to try to be shoehorning various config syntaxes in with bash support. We may even end up splitting bash, zsh and tcsh variants out into their own syntaxes.

Issue #442, opened 2016-05-31
[HTML/XML?] Add meta scopes for tag closing


FichteFoll:
#441 just made me aware of the fact that closing tags, at least in HTML, have no meta scope (or otherwise district selector). I would probably be advantageous to have a meta scope for closing tags for example to highlight them in a different color or render them in italics.

keith-hall:
Not sure if you also want a separate scope for self closing tags or not, but just wanted to note that this won't be possible for self closing tags that have the `/>` "punctuation" on a separate line, due to the regexs only being able to see one line at a time.

i.e.

    <foobar
        attr="hello" />

FichteFoll:
Yes, that is why I didn't specifically include self-closing tags as well. I'm not a huge fan of giant look-aheads either, but sometimes they just can't be avoided.

The closing punctuation of the self-closing tag should get a distinct scope however.

LeonardoGentile:
Related to #439?

wbond:
This issue has been solved by a previous commit, hence it is being closed.

Issue #438, opened 2016-05-30
[HTML] Question: auto_complete_trigger discrepency


keith-hall:
In the Default `Preferences.sublime-settings` shipped with Sublime Text, there is the following preference:

```
"auto_complete_triggers": [ {"selector": "text.html", "characters": "<"} ]
```

however, as things stand, triggering HTML auto completion when/after typing `<` has no effect, because:
- The [`html_completions.py` file returns nothing if there is no `prefix` - i.e. nothing typed after the `<` character.](https://github.com/sublimehq/Packages/blob/6a5e2d46d01df3c5da0d3cd181ea62fffdc7aa76/HTML/html_completions.py#L236)

So my question is, was it intended that it would show all available tags? Or is it just in the default preferences to serve as an example of how to configure this preference?

It's presence confused me somewhat, because it triggers even when in a `source` scope i.e. `embedding.php text.html.basic meta.embedded.block.php source.php`, except that the HTML tags are [correctly not listed by `on_query_completions` because the first thing it does is check that it is not in a source scope.](https://github.com/sublimehq/Packages/blob/6a5e2d46d01df3c5da0d3cd181ea62fffdc7aa76/HTML/html_completions.py#L202). However, other plugins may still decide to show completions at this stage, so maybe the preference should be either removed or the selector changed to `text.html - source`?

I thought I would ask about it here, to know whether it is a potential issue that it does not popup a completions window for all HTML tags when typing the <kbd>&lt;</kbd>.

sentience:
This definitely worked (popping up a list of all tags) in Sublime Text 2. I'm currently updating my lynda.com video course for ST3 and was surprised to find this not working in the current build.

michaelblyons:
Is this fixed by recent changes?

keith-hall:
Partially - typing `<` in HTML now shows available tags in the completion popup, but ST still brings up autocomplete when typing `<` in a `text.html source` scope, for example:

- create a new file
- set syntax to PHP
- type `html` <kbd>Tab</kbd>
- <kbd>Esc</kbd> out the snippet field mode
- type `php` <kbd>Tab</kbd>
- type `if (a <` and see autocomplete popup undesirably (correctly without HTML tag suggestions except those from the buffer)

However, this is now a core issue (`Packages/Default/Preferences.sublime-preferences`) so maybe makes sense for me to open a new issue on the core tracker about changing the default `auto_complete_triggers` and close this one.

michaelblyons:
[...]

PR #586, opened 2016-08-17
[HTML] [CSS] completions before </style> and [HTML] meta.tag scope

keith-hall:
[HTML] meta.tag scopes were missing on style and script tags.

suggest [CSS] completions immediately before `</style>`. This was suggested at https://github.com/sublimehq/Packages/pull/532#issuecomment-240342983

Note: this change will still require auto-completion to be triggered manually when the cursor is just before `</style>` (when using ST without any modifications). This is just like the other PR, whereby typing just before the closing `"` in a style attribute also requires a manual trigger.  There are a few reasons for this:
- when ST checks scopes [for autocompletion], it checks at the character after the caret, which would be `<` in this case, with the scope `text.html.basic meta.tag.style.html punctuation.definition.tag.begin.html`
- In the default `auto_complete_selector` preference, `punctuation.definition.tag.begin` is specifically excluded from the `meta.tag` scope.
- only the CSS itself has a `source` scope, which is picked up by the default `auto_complete_selector` preference. The character that closes the CSS scope in the HTML document (`<` here, `"` in the style attribute) doesn't.

Obviously, users can work around the above "limitation" by overriding their preference. But anyway, this PR also fixes the problem whereby HTML tags were being offered when manually triggering the autocompletion immediately before the `</style>`.

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -199,7 +199,8 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
 
     def on_query_completions(self, view, prefix, locations):
         # Only trigger within HTML
-        if not view.match_selector(locations[0], "text.html - source - string.quoted - meta.tag.style punctuation.definition.tag.begin"):
+        if not view.match_selector(locations[0],
+           "text.html - source - string.quoted - meta.tag.style.end punctuation.definition.tag.begin"):
             return []
 
         # check if we are inside a tag
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -199,7 +199,7 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
 
     def on_query_completions(self, view, prefix, locations):
         # Only trigger within HTML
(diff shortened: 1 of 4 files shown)


HTML/html_completions.py
```diff
@@ -199,7 +199,7 @@ def __init__(self):
 
     def on_query_completions(self, view, prefix, locations):
         # Only trigger within HTML
-        if not view.match_selector(locations[0], "text.html - source - string.quoted"):
+        if not view.match_selector(locations[0], "text.html - source - string.quoted - meta.tag.style punctuation.definition.tag.begin"):
```
keith-hall: Note: this has the side effect of showing buffer completions when typing a new tag immediately before the `<style type="text/css">` i.e. type `<`. I'd welcome any ideas for combating this. I am thinking to have a separate scope for `meta.tag.style.begin.html` and `meta.tag.style.end.html`, as per https://github.com/sublimehq/Packages/issues/442.

keith-hall: Related: https://github.com/sublimehq/Packages/issues/438

PR #591, opened 2016-08-18
[Scala] Only index scoped packages

### LINKED ISSUE #552: [Scala] Proposal: packages should only be indexed if followed by a scope
Scala has two different package syntaxes:

```scala
package fubar

// your stuff here!
```

And this one:

```scala
package fubar {
  // your stuff here
}
```

As of master, sublime currently indexes *both* of these syntaxes.  I think it's basically useless to index the first one, since it's going to appear more or less verbatim at the top of every file (e.g. if you attempt to Cmd-Shift-R "fubar", you're going to get every file in your project).  However, it does seem quite useful to index the *second* version, since this syntax when used generally denotes a package which is only defined in this one place.

djspiewak:
This fixes #552.  To be clear:

```scala
package stuff    // not indexed

package stuff2 {
  // indexed
}
```

I'm using the scopes `entity.name.namespace.header` and `entity.name.namespace.scoped`.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -264,11 +264,20 @@ contexts:
         1: keyword.control.scala
         2: storage.type.class.scala
         3: entity.name.class.scala
-    - match: '\b(package)\s+([\w\.]+)'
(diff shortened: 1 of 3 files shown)


wbond:
Thanks!

PR #593, opened 2016-08-20
[Scala] Bail out on backticks in patmat if hitting EOL

djspiewak:
Newlines are not allowed in backtick identifiers


wbond:
A `$` in `[]` is just a normal `$`, not end of line.

djspiewak:
> A $ in [] is just a normal $, not end of line

Huh…  No idea how I got that wrong.  Will fix.

PR #594, opened 2016-08-20
[Scala] Fixed lambda lookahead for bare ascriptions with type parameters

djspiewak:
Lambda lookahead didn't correctly identify the following scenario:

```scala
x: List[Int] => ???
```

This is now using the same hack that I had to use for parentheses matched lookahead, which means it will break with over 3 levels of bracket nesting.


FichteFoll:
It should be possible to recursively apply regular expressions in cases like this, using named group references.

wbond:
sregex doesn't support named groups, nor recursion currently. This should handle the majority of types.

PR #595, opened 2016-08-20
[D] Small fixes

vladdeSV:
#### Add class documentation field
* D relies heavily on commented code

#### Improve struct documentation field
* Functions the same way as `method` does

#### Fix warning
* There was a stray }

#### Improve while
* To match rest of snippets, a space was removed


wbond:
Thanks!

PR #599, opened 2016-08-24
[Scala] Fixed nested scaladoc

djspiewak:
Scala allows nested comment blocks.  We correctly handle this for regular comments, but scaladoc is (still) sort of a mess and we weren't correctly nesting the scopes.  Fixed.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -92,28 +92,6 @@ contexts:
     - include: late-operators
 
   block-comments:
-    - match: /\*
-      push:
-        - meta_scope: comment.block.scala
-        - match: \*/
-          pop: true
-        - include: block-comments
-        - match: |-
-            (?x)
-            			(?! /\*)
-            			(?! \*/)
-
-  char-literal:
-    - match: '''\\?.'''
-      scope: constant.character.literal.scala
-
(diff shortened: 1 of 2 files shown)


Scala/Scala.sublime-syntax
```diff
@@ -92,6 +92,18 @@ contexts:
     - include: late-operators
 
   block-comments:
+    - match: (^\s*)?/\*\*
+      scope: punctuation.definition.comment.scala
+      push:
+        - meta_scope: comment.block.documentation.scala
+        - match: \*/\s*$
```
wbond: Won't this potentially run into an error like what was fixed with #596? Perhaps if you want to enforce nothing on the same line, mark everything other than a space after `*/` as invalid?

djspiewak: These two PRs are essentially in conflict.  I can just merge them into one if you would like.  They each fix a separate issue with the same functionality.

wbond: I've already merged in #596. Mind making that fix to this PR also?

djspiewak: Willdo over the weekend.

djspiewak:
@wbond done

PR #600, opened 2016-08-24
[YAML] comment.line fix, scope changes

infininight:
Ported this grammar back to TextMate and had a few changes I made along the way. Note: I am not in any way familiar with SublimeText and did not test these changes. I believe I did the tests correctly but did not test them either.

233860f78a55fe0c5a8f873ed6754ec38e91ac39: The comment.line rule was matching way too much, it caught any line with trailing whitespace and any empty line. I totally rewrote the rule and made it a begin/end style rule to allow for injections.

The rest are just minor scope changes.

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -271,7 +271,7 @@ contexts:
       - match: (&)({{ns_anchor_name}})(\S+)?
         captures:
           1: keyword.control.property.anchor.yaml punctuation.anchor.yaml
-          2: entity.name.other.anchor.yaml
+          2: entity.name.type.anchor.yaml
           3: invalid.illegal.character.anchor.yaml
         pop: true
       # !Tag Handle
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -411,7 +411,7 @@ contexts:
         - include: flow-pair
         - include: flow-node
         - match: :(?=\s|$|{{c_flow_indicator}})
-          scope: punctuation.separator.mapping.key-value.yaml
+          scope: punctuation.separator.key-value.mapping.yaml
           set: flow-pair-value
     # Attempt to match plain-in scalars and highlight as "entity.name.tag",
     # if followed by a colon
@@ -442,7 +442,7 @@ contexts:
               pop: true
     - include: flow-node
     - match: :(?=\s|$|{{c_flow_indicator}}) # Empty mapping keys allowed
-      scope: meta.flow-pair.yaml punctuation.separator.mapping.key-value.yaml
+      scope: meta.flow-pair.yaml punctuation.separator.key-value.mapping.yaml
       push: flow-pair-value
 
   flow-pair-value:
@@ -496,7 +496,7 @@ contexts:
           pop: true
         - match: ^ *(:)
           captures:
-            1: punctuation.separator.mapping.key-value.yaml
+            1: punctuation.separator.key-value.mapping.yaml
           pop: true
         - match: ':'
           scope: invalid.illegal.expected-newline.yaml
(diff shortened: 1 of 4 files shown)


FichteFoll:
Regarding comment matching: I initially followed the definitions provided by the spec for [`l-comment`](http://www.yaml.org/spec/1.2/spec.html#l-comment) but I agree that this is not optimal for highlighting and should be removed. The fact that it was a consuming match and didn't allow for injections is also a good catch.

Re `punctuation.separator.mapping.key-value` vs `punctuation.separator.key-value.mapping`, I'm undecided on this one. To me, a "mapping" is more general than "key-value" pair specifically, but I haven't seen any of these used in practice by other syntax definitions or color schemes. Could you point me to some, if you have any?

The other changes don't need any comments. :+1:

YAML/YAML.sublime-syntax
```diff
@@ -527,16 +527,23 @@ contexts:
...
   comment:
     # http://www.yaml.org/spec/1.2/spec.html#comment//
     - match: | # l-comment
         (?x)
-        (?: ^ [ \t]* | [ \t]+ )
-        (?:(\#) \p{Print}* )?
-        (\n|\z)
-      scope: comment.line.number-sign.yaml
+        (?: ( ^ [ \t]* ) | [ \t]+ )
+        (?=#\p{Print}*$)
```
wbond: The `#` in extended mode denotes a comment, so nothing after that is matching. In fact the regex doesn't even compile.

wbond: Additionally, this regex pretty much hoses the YAML definition.

screen shot 2016-09-21 at 5 04 05 am

wbond:
`punctuation.separator.key-value` is used by JavaScript, and seems more abstract than `punctuation.separator.mapping` since `key-value` pairs can happen in dictionaries, mappings, objects, etc.

We should update the JSON syntax to use `punctuation.separator.key-value`, but that does not need to be part of this PR.

YAML/YAML.sublime-syntax
```diff
@@ -271,7 +271,7 @@ contexts:
       - match: (&)({{ns_anchor_name}})(\S+)?
         captures:
           1: keyword.control.property.anchor.yaml punctuation.anchor.yaml
-          2: entity.name.other.anchor.yaml
+          2: entity.name.type.anchor.yaml
```
infininight: OK, this is a difference between the TextMate and Sublime conventions, defer to your choice.

YAML/YAML.sublime-syntax
infininight: Ah correct, in the TextMate format grammar I had this as a non-extended and only changed it because the original rule here was extended.

YAML/YAML.sublime-syntax
```diff
@@ -527,16 +527,23 @@ contexts:
...
         (?x)
-        (?: ^ [ \t]* | [ \t]+ )
-        (?:(\#) \p{Print}* )?
-        (\n|\z)
-      scope: comment.line.number-sign.yaml
+        (?: ( ^ [ \t]* ) | [ \t]+ )
+        (?=#\p{Print}*$)
       captures:
-        1: punctuation.definition.comment.line.number-sign.yaml
+        1: punctuation.whitespace.comment.leading.yaml
```
infininight: In Textmate we use this scope to be able to have commands, with scope as the input, match an entire block of comment lines. This allows for things like reflowing the comment text, etc. If this isn't used in Sublime the rule can be simplified.

wbond: Yes, I just saw you are porting tweaks from Textmate – thanks for taking the time!

infininight:
I have commented on a few of the bits of feedback above, I am going to leave this for someone else to correct the minor issues that remain since I cannot test in Sublime itself which makes ensuring issues are corrected more difficult.

wbond:
This was superseded by #627

Issue #588, opened 2016-08-17
[Scala] Further Enhancements


djspiewak:
There are some further enhancements which are included in #94 which should be ported into the mainline Scala mode.  Off the top of my head:

- [ ] Unicode character escapes in strings (e.g. `\u2101`)
- [ ] Scoping for annotations (outside of primitives like `@volatile`)
- [ ] Better handling for XML
- [ ] Meta scoping
- [ ] Sane scaladoc highlighting

I understand there are several other features also implemented by #94, but I'm personally of the *opinion* that they're probably not worth pulling in (e.g. specifically target-scoped braces).

This is the place for opinions though!  Ping @gwenzek for thoughts, opinions and insights.  Anything I missed?  Anything that should be added to the above?

gwenzek:
Thanks for making this list.

I'd also like to add:

- [ ] Scoping of function call `f(a)`
- [ ] Scoping of operators in infix position `a + b`
- [ ] Similar scopes for `s"$foo"` and `s"${foo}"`

Regarding the symbol list I'd like to improve it with the following points:

- [ ] Full function signature in  symbol list
- [ ] Indentation in symbol list

I think the point you mentioned are the most consensual, so I'll start by that.

gwenzek:
Below are two screenshots showing what the symbol list could look like (I don't have `val` here but it's just an example).

**Current symbol list**:
![Current symbol list](https://cloud.githubusercontent.com/assets/5920036/17756132/5abcfafe-64de-11e6-96cc-918c17570f52.png)

**Proposed symbol list**
![Proposed symbol list](https://cloud.githubusercontent.com/assets/5920036/17755981/78d16288-64dd-11e6-8fbd-9b93bc8bca08.png)

IMHO the second one is more readable. The problem comes from too long signatures that are strangely abbreviated, instead of just truncated.

djspiewak:
> - Full function signature in symbol list
> - Indentation in symbol list

Yes yes yes!  I took a crack at doing this a few weeks ago.  This is one thing that the Python mode does tremendously well.

Unfortunately, Scala's function and value declaration syntax is context-free, whereas Python's is regular, which means that we're not going to make things quite as good as we might like.

> Below are two screenshots showing what the symbol list could look like

I would be in favor of removing the `trait`, `class`, `def` stuff, as well as likely the parameter names.  As I said, the Python mode does a great job here, but I'm not sure how far in that direction we can go.

djspiewak:
[...]

PR #605, opened 2016-08-28
[Scala] add support for annotations

gwenzek:
As mentioned in #588, I propose to add support for annotations.
I based myself on http://www.scala-lang.org/files/archive/spec/2.11/11-annotations.html .

I used the scope `storage.modifier.annotation.scala` on annotations.
This provide coloring in most themes due to `storage.modifier`and is easily customizable.

``` scala
@deprecated("Use D", "1.0") class C { ... }
// ^^ storage.modifier.annotation
//            ^^ string

@transient @volatile var m: Int
// ^^ storage.modifier.annotation

String @local
//     ^^ storage.modifier.annotation

(e: @unchecked) match { ... }
//  ^^ storage.modifier.annotation
//              ^^^^^ keyword.control.flow.scala

// more complex:
@scala.beans.BeanProperty
//  ^^^^^^^^^^^^^^^^^^^^^ storage.modifier.annotation

(e: Int @unchecked) match { ... }
//  ^^ storage.type.primitive
//      ^^ storage.modifier.annotation
//                  ^^^^^ keyword.control.flow.scala

trait Function0[@specialized(Unit, Int, Double) T] {
//               ^^ storage.modifier.annotation
//                           ^^ storage.type.primitive
//                                              ^ support.class
  def apply: T
}

```

**Note**
I match annotations with `{{plainid}}` regex, but we might want to support something more complex in the future.

I handle `@specialized` separately because it's the only annotation I know that uses type parameters between `(`.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -166,9 +166,7 @@ contexts:
       push:
         - match: ','
           scope: punctuation.separator.arguments.scala
-        - include: strings
-        - include: constants
-        - include: initialization
+        - include: main
         - match: \)
           scope: punctuation.definition.arguments.end.scala
           pop: true
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -179,7 +179,7 @@ contexts:
           scope: punctuation.definition.arguments.end.scala
           pop: true
         - include: delimited-type-expression
-    - match: (?!\()
+    - match: (?!\(|\[)
       pop: true
 
   lambdas:
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -149,20 +149,6 @@ contexts:
       push: single-type-expression
 
   annotation:
-    - match: '@specialized'
-      scope: storage.modifier.annotation.scala
-      push:
-      - match: \(
-        scope: punctuation.definition.arguments.begin.scala
-        set:
-          - include: base-types
-          - match: ','
-            scope: punctuation.separator.arguments.scala
-          - match: \)
-            scope: punctuation.definition.arguments.end.scala
-            pop: true
-      - match: (?!\()
-        pop: true
     - match: '@{{plainid}}\.'
       scope: storage.modifier.annotation.scala
       push:
@@ -178,7 +164,7 @@ contexts:
   annotation-parameters:
     - match: \(
       scope: punctuation.definition.arguments.begin.scala
-      set:
+      push:
         - match: ','
           scope: punctuation.separator.arguments.scala
         - include: strings
@@ -187,6 +173,13 @@ contexts:
         - match: \)
           scope: punctuation.definition.arguments.end.scala
           pop: true
+    - match: \[
+      scope: punctuation.definition.arguments.begin.scala
+      push:
+        - match: \]
+          scope: punctuation.definition.arguments.end.scala
+          pop: true
+        - include: delimited-type-expression
     - match: (?!\()
       pop: true
 
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -148,43 +148,45 @@ contexts:
       push: single-type-expression
 
   annotation:
-    - match: '(@){{plainid}}\.'
+    - match: '(@)({{plainid}})(\.)'
+      scope: storage.modifier.annotation.scala
       captures:
-        0: storage.modifier.annotation.scala
         1: punctuation.definition.annotation.scala
+        3: punctuation.accessor.scala
       push:
         - meta_scope: meta.annotation.scala
-        - match: '{{plainid}}\.'
+        - match: '({{plainid}})(\.)'
           scope: storage.modifier.annotation.scala
+          captures:
+            2: punctuation.accessor.scala
         - match: '{{plainid}}'
           scope: storage.modifier.annotation.scala
           set: annotation-parameters
-    - match: '(@){{plainid}}'
+    - match: '(@)({{plainid}})'
+      scope: storage.modifier.annotation.scala
       captures:
-        0: storage.modifier.annotation.scala
         1: punctuation.definition.annotation.scala
       push: annotation-parameters
 
   annotation-parameters:
     - meta_scope: meta.annotation.scala
     - match: \(
-      scope: punctuation.definition.arguments.annotation.begin.scala
+      scope: punctuation.section.arguments.annotation.begin.scala
       push:
         - match: ','
           scope: punctuation.separator.arguments.annotation.scala
         - match: \)
-          scope: punctuation.definition.arguments.annotation.end.scala
+          scope: punctuation.section.arguments.annotation.end.scala
           pop: true
         - include: main
     - match: \[
-      scope: punctuation.definition.arguments.annotation.begin.scala
+      scope: punctuation.section.arguments.annotation.begin.scala
       push:
         - match: \]
-          scope: punctuation.definition.arguments.annotation.end.scala
+          scope: punctuation.section.arguments.annotation.end.scala
           pop: true
         - include: delimited-type-expression
-    - match: \s*
-    - match: (?![(\[])
+    - match: (?!\s*[(\[])
       pop: true
 
   lambdas:
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -103,7 +103,6 @@ contexts:
   main-post-lambdas:
     - include: late-keywords
     - include: constants
-    - include: char-literal
     - include: scala-symbol
     - include: braces
     - include: late-operators
@@ -169,7 +168,7 @@ contexts:
           scope: punctuation.separator.arguments.scala
         - include: strings
         - include: constants
-        - include: char-literal
+        - include: initialization
         - match: \)
           scope: punctuation.definition.arguments.end.scala
           pop: true
@@ -244,6 +243,7 @@ contexts:
     - include: base-types
   constants:
     - include: base-constants
+    - include: char-literal
     # other upper-case stuff highlights as constant
     - match: '{{upperid}}'
       scope: support.constant.scala
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -162,19 +162,19 @@ contexts:
 
   annotation-parameters:
     - match: \(
-      scope: punctuation.definition.arguments.begin.scala
+      scope: punctuation.definition.arguments.annotation.begin.scala
       push:
         - match: ','
-          scope: punctuation.separator.arguments.scala
+          scope: punctuation.separator.arguments.annotation.scala
         - include: main
         - match: \)
-          scope: punctuation.definition.arguments.end.scala
+          scope: punctuation.definition.arguments.annotation.end.scala
           pop: true
     - match: \[
-      scope: punctuation.definition.arguments.begin.scala
+      scope: punctuation.definition.arguments.annotation.begin.scala
       push:
         - match: \]
-          scope: punctuation.definition.arguments.end.scala
+          scope: punctuation.definition.arguments.annotation.end.scala
           pop: true
         - include: delimited-type-expression
     - match: (?!\(|\[)
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -1120,13 +1120,17 @@ contexts:
     - match: '{{typeid}}'
       scope: support.type.scala
       set: single-type-expression-tail
+    - match: (?=@{{plainid}})
+      set: single-type-expression-tail
     - include: base-type-expression
-    # the '@' indicates the beginning of an annotation
-    - match: '(?=@)'
-      pop: true
-    - match: '(?=[\s$,\)\}\]])'
+    - match: '(?=[\s,\)\}\]])'
       pop: true
   single-type-expression-tail:
+    - match: (?=@{{plainid}})
+      set:
+        - include: annotation
+        - match: '(?=\S)'
+          set: single-type-expression-tail
     - match: '[\.#]'
       scope: punctuation.accessor.scala
       set: single-type-expression
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -148,28 +148,34 @@ contexts:
       push: single-type-expression
 
   annotation:
-    - match: '@{{plainid}}\.'
-      scope: storage.modifier.annotation.scala
+    - match: '(@){{plainid}}\.'
+      captures:
+        0: storage.modifier.annotation.scala
+        1: punctuation.definition.annotation.scala
       push:
+        - meta_scope: meta.annotation.scala
         - match: '{{plainid}}\.'
           scope: storage.modifier.annotation.scala
         - match: '{{plainid}}'
           scope: storage.modifier.annotation.scala
           set: annotation-parameters
-    - match: '@{{plainid}}'
-      scope: storage.modifier.annotation.scala
+    - match: '(@){{plainid}}'
+      captures:
+        0: storage.modifier.annotation.scala
+        1: punctuation.definition.annotation.scala
       push: annotation-parameters
 
   annotation-parameters:
+    - meta_scope: meta.annotation.scala
     - match: \(
       scope: punctuation.definition.arguments.annotation.begin.scala
       push:
         - match: ','
           scope: punctuation.separator.arguments.annotation.scala
-        - include: main
         - match: \)
           scope: punctuation.definition.arguments.annotation.end.scala
           pop: true
+        - include: main
     - match: \[
       scope: punctuation.definition.arguments.annotation.begin.scala
       push:
@@ -177,7 +183,8 @@ contexts:
           scope: punctuation.definition.arguments.annotation.end.scala
           pop: true
         - include: delimited-type-expression
-    - match: (?!\(|\[)
+    - match: \s*
+    - match: (?![(\[])
       pop: true
 
   lambdas:
@@ -1112,8 +1119,8 @@ contexts:
       scope: support.type.scala
       set: single-type-expression-tail
     - include: base-type-expression
+    # the '@' indicates the beginning of an annotation
     - match: '(?=@)'
-      comment: the '@' indicates the beginning of an annotation
       pop: true
     - match: '(?=[\s$,\)\}\]])'
       pop: true
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -149,21 +149,21 @@ contexts:
 
   annotation:
     - match: '(@)({{plainid}})(\.)'
-      scope: storage.modifier.annotation.scala
+      scope: variable.annotation.scala
       captures:
         1: punctuation.definition.annotation.scala
         3: punctuation.accessor.scala
       push:
         - meta_scope: meta.annotation.scala
         - match: '({{plainid}})(\.)'
-          scope: storage.modifier.annotation.scala
+          scope: variable.annotation.scala
           captures:
             2: punctuation.accessor.scala
         - match: '{{plainid}}'
-          scope: storage.modifier.annotation.scala
+          scope: variable.annotation.scala
           set: annotation-parameters
     - match: '(@)({{plainid}})'
-      scope: storage.modifier.annotation.scala
+      scope: variable.annotation.scala
       captures:
         1: punctuation.definition.annotation.scala
       push: annotation-parameters
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -1098,7 +1098,6 @@ contexts:
       scope: support.type.scala
     - include: base-type-expression
   single-type-expression:
-    - include: annotation
     - match: '\b(type)\b'
       scope: keyword.other.scala
     - match: '[\x{03B1}\x{03B2}]'     # just here for type lambdas
@@ -1115,6 +1114,9 @@ contexts:
       scope: support.type.scala
       set: single-type-expression-tail
     - include: base-type-expression
+    - match: '(?=@)'
+      comment: the '@' indicates the beginning of an annotation
+      pop: true
     - match: '(?=[\s$,\)\}\]])'
       pop: true
   single-type-expression-tail:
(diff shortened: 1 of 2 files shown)


Scala/Scala.sublime-syntax
```diff
@@ -813,6 +857,7 @@ contexts:
       scope: support.type.scala
     - include: base-type-expression
   single-type-expression:
+    - include: annotation
```
djspiewak: This isn't going to work quite correctly, and also I think it's probably wrong.  For example:

```scala
x: @specialized Foo
```

Does that parse?  I actually don't know, but I don't think it does.  

Anyway, the reason it won't work correctly is because the space following `@specialized` will break out of the `single-type-expression` context!  If the above *does* parse, then you need to have a bit more machinery that parses the annotation, consumes the whitespace and then `set`s back into `single-type-expression`.  It isn't valid to ascribe *just* an annotation as a type, so that probably works.

Scala/Scala.sublime-syntax
```diff
@@ -130,6 +131,48 @@ contexts:
     - match: '{{typeprefix}}'
       push: single-type-expression
 
+  annotation:
+    - match: '@specialized'
```
djspiewak: This really shouldn't be special-cased.  First off, there's nothing particularly magical about `@specialized`, and other annotations *can* take type parameters in this manner.  Second, it is extremely common in practice to import-rename `@specialized` to `@sp` or `@spec`, for concision.  Your special-casing doesn't catch this.

I would rather simply support parenthetical parameters on annotations.  You'll have to do a bit of lookahead for it, I think, but that's not too bad.

djspiewak:
I agree that this fits well as a subscope of `storage.modifier`.  That's semantically closest to what it is, and fits well with how most color schemes use that scope.

Scala/Scala.sublime-syntax
gwenzek: Thanks for the review.
I just tried with Scala 2.10 and 2.11 and `(e: Int @unchecked)` compiles but `(e: @unchecked Int)` doesn't ( "error: ')' expected but identifier found." )
The `annotation` context doesn't popup after the annotation and we still are in the `single-type-expression` so it works with  `(e: Int @unchecked)`.
With `(e: @unchecked Foo)` the `Foo` appears as a `support.constant` when the user would expect a type, which hints the compile error.

Scala/Scala.sublime-syntax
gwenzek: >  other annotations *can* take type parameters in this manner

I'm not really familiar with this kind of annotations. I mainly use annotations with string/int parameters. I haven't found documentation on annotations having types inside "()".
If you have pointers or code sample, I'll be happy to have a look at them.

My plan is to support  `@test[A, B]`, `@test(1, "foo")` and `@test(a, b)`
But I'm not sure how to handles `@test(Int, Foo)` because it's not really different from the previous.

Scala/Scala.sublime-syntax
djspiewak: I think it would be better to either highlight it as `invalid` or simply remove the highlighting for annotations inside of single-type-expression.

djspiewak:
Here's a link to the way in which you can get types (or really anything) in annotation parens: http://docs.scala-lang.org/overviews/macros/annotations.html

djspiewak:
I think what you can probably do to handle annotation parentheticals:

```yaml
  - match: '@{{id}}'
    push: try-ann-params

try-ann-params:
  - match: '\('
    push:
      - match: '\)'
        pop: true
      - include: primitives
      - include: delimited-type-expression
  - match: '(?=\S)'
    pop: true
```

You'll need to extract numeric, boolean and string literals into a new context, which shouldn't take much refactoring.  Also, you'll want to match `classOf`, since that will result in a class literal rather than a type.

gwenzek:
> You'll need to extract numeric, boolean and string literals into a new context, `literals`

Isn't the context `constants` already doing that ?

gwenzek:
So I removed the special handling of `@specialized`, and "annotation" is not included in the "single-type-expression" context anymore.
The supported annotation parameters are any of the "constants" context `new ...` expressions and strings.

This allow to highlight properly the following edge cases:

```scala
@foo(new AnyRef { })
@foo(List(1, "foo"))

(e: Int @unchecked)
(e: @unchecked)

@specialized(Unit, Int, Double)
```

djspiewak:
> Isn't the context constants already doing that ?

Sort of.  It doesn't handle strings (which you want), and it *does* handle upper-case constants (which you don't want).  So I think we need a bit of reorganization there.

djspiewak:
> `@foo(new AnyRef { })`
> `@foo(List(1, "foo"))`

I don't believe that either of these are actually legal outside of macro annotations.  Do you feel it's worth highlighting those cases?

Scala/Scala.sublime-syntax
```diff
@@ -867,6 +866,9 @@ contexts:
       scope: support.type.scala
       set: single-type-expression-tail
     - include: base-type-expression
+    - match: '(?=@)'
```
djspiewak: Why not just do this in `single-type-expression-tail`?  This is how we handle the following case:

```scala
e: Foo with Bar
```

djspiewak:
@gwenzek I gave it some thought (well, a lot of thought), and I think you should just allow arbitrary expressions within annotation parentheses.  The problem we have here is that macro annotations allow people to do *anything* and we can't really detect that.  This is a syntax highlighting mode, so we have to be really aggressive about what we highlight.

The only downside to allowing arbitrary expressions within annotation parentheticals is it means that the types in `@specialized(Unit, Int, Double)` will scope as constants rather than types.  I'm not sure that's particularly terrible (e.g. in Monokai, the only difference is the former is italic while the latter is not), but it is a thing.  Ultimately, we can't really tell what people *really* want there, and `@specialized` is specifically taking constants and remapping them as types through AST rewriting, so it's a bit unfair to expect Sublime to sort that all out.

So tl;dr: I would literally treat it as a `try-dispatch`.

gwenzek:
I agree with you, that's why I finally settled for matching 'strings', 'constants' and 'initialization'.
If you think it's better to push a 'main' context I can do it, but is it really allowed to put *anything* like a class definition in an annotation ?

djspiewak:
I don't think you can put a class definition in an annotation, but you can definitely put arbitrary expressions (e.g. ascriptions, lambdas, etc).

gwenzek:
Ok, I allowed anything as annotation parameter using `main` context, as you did in `try-dispatch`.

I think the PR is good to be merged, but maybe I could collapse the commits since I ended with 6 different commits for what is finally a minor code modification.

djspiewak:
@gwenzek Oh, btw, should we remove the special-casing of `@volatile` with this PR?

gwenzek:
@wbond The PR is in good shape now and has been validated by @djspiewak .
If you're okay, I'll collapse the commits into one and you'll be able to merge it !

Scala/Scala.sublime-syntax
```diff
@@ -1039,6 +1074,9 @@ contexts:
...
+      comment: the '@' indicates the beginning of an annotation
```
wbond: With YAML defs we are moving away from `comment` keys and just writing a YAML comment.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,39 @@ contexts:
     - match: '{{typeprefix}}'
       push: single-type-expression
 
+  annotation:
+    - match: '@{{plainid}}\.'
```
wbond: The `@` symbol should probably have the scope `punctuation.definition.annotation.scala`, like how variables do in PHP.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,39 @@ contexts:
...
+      scope: storage.modifier.annotation.scala
+      push:
+        - match: '{{plainid}}\.'
+          scope: storage.modifier.annotation.scala
+        - match: '{{plainid}}'
+          scope: storage.modifier.annotation.scala
+          set: annotation-parameters
+    - match: '@{{plainid}}'
```
wbond: Add the `punctuation` scope to `@` here also.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,39 @@ contexts:
...
+      scope: storage.modifier.annotation.scala
+      push: annotation-parameters
+
+  annotation-parameters:
+    - match: \(
+      scope: punctuation.definition.arguments.annotation.begin.scala
```
wbond: This and related scopes should start with `punctuation.section.`. We are using `punctuation.definition` for more single-token/atomic type things and `punctuation.section` for larger areas, like braces, parens, etc.

This is an ongoing change, so there will be plenty of examples that still use `punctuation.definition` for this type of construct.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,39 @@ contexts:
     - match: '{{typeprefix}}'
       push: single-type-expression
 
+  annotation:
+    - match: '@{{plainid}}\.'
+      scope: storage.modifier.annotation.scala
+      push:
```
wbond: Let's add a `meta.` scope for the annotation.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,39 @@ contexts:
...
+        - match: '{{plainid}}\.'
+          scope: storage.modifier.annotation.scala
+        - match: '{{plainid}}'
+          scope: storage.modifier.annotation.scala
+          set: annotation-parameters
+    - match: '@{{plainid}}'
+      scope: storage.modifier.annotation.scala
+      push: annotation-parameters
+
+  annotation-parameters:
```
wbond: The same `meta.` scope should be added here as in the match in the `annotation` context.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,39 @@ contexts:
...
+    - match: \(
+      scope: punctuation.definition.arguments.annotation.begin.scala
+      push:
+        - match: ','
+          scope: punctuation.separator.arguments.annotation.scala
+        - include: main
+        - match: \)
```
wbond: This match should probably be moved up above `include: main` in case a match for `\)` was added there in the future.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,39 @@ contexts:
...
+          scope: punctuation.definition.arguments.annotation.end.scala
+          pop: true
+    - match: \[
+      scope: punctuation.definition.arguments.annotation.begin.scala
+      push:
+        - match: \]
+          scope: punctuation.definition.arguments.annotation.end.scala
+          pop: true
+        - include: delimited-type-expression
+    - match: (?!\(|\[)
```
wbond: Is whitespace allowed between `@identifier` and `(`? If so, this should allow `\s+` also.

Scala/syntax_test_scala.scala
```diff
@@ -1189,3 +1189,38 @@ def <(a: Int) = 42
      &amp;
 //   ^^^^^ constant.character.entity.xml - meta.tag.xml
    </foo>
+
+// annotation examples from: http://www.scala-lang.org/files/archive/spec/2.11/11-annotations.html
+@deprecated("Use D", "1.0") class C { ... }
```
wbond: Add assertions to test the `@`, `,` and the `meta.` scope.

Scala/syntax_test_scala.scala
```diff
@@ -1189,3 +1189,38 @@ def <(a: Int) = 42
...
+// ^^ storage.modifier.annotation
+//            ^^ string
+
+@transient @volatile var m: Int
```
wbond: Add assertions for the second annotation here to ensure we a properly allowing two back-to-back.

Scala/Scala.sublime-syntax
djspiewak: I'd like it if we could give them this scope, but *overlapping* with the `storage.modifier.annotation` scope, so that the `@` doesn't end up "randomly not colored" in most color schemes.

wbond: Yes, that is correct, the `@` should be in addition to the `storage.modifier.annotation.scala` scope.

djspiewak:
@gwenzek Can you add a test for the following?

```scala
x: Foo @volatile with Bar
```

I think my comment from earlier got lost in the wash.  I'm not sure the above will correctly highlight with your current implementation.

gwenzek:
@wbond  I added your suggestions 

@djspiewak done and it works

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
     - match: '{{typeprefix}}'
       push: single-type-expression
 
+  annotation:
+    - match: '(@){{plainid}}\.'
```
wbond: Sorry, missed this the first time: the `.` should be `punctuation.accessor.scala`.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
...
+      captures:
+        0: storage.modifier.annotation.scala
+        1: punctuation.definition.annotation.scala
+      push:
+        - meta_scope: meta.annotation.scala
+        - match: '{{plainid}}\.'
```
wbond: Again, `punctuation.accessor.scala` on the `.`.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
     - match: '{{typeprefix}}'
       push: single-type-expression
 
+  annotation:
+    - match: '(@){{plainid}}\.'
+      captures:
+        0: storage.modifier.annotation.scala
```
wbond: Stylistically we've been moving away from `0` captures and instead using `scope: ` for when a scope applies to the whole match.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
...
+      push:
+        - meta_scope: meta.annotation.scala
+        - match: '{{plainid}}\.'
+          scope: storage.modifier.annotation.scala
+        - match: '{{plainid}}'
+          scope: storage.modifier.annotation.scala
+          set: annotation-parameters
+    - match: '(@){{plainid}}'
+      captures:
+        0: storage.modifier.annotation.scala
```
wbond: Move to `scope: `.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
...
+        1: punctuation.definition.annotation.scala
+      push: annotation-parameters
+
+  annotation-parameters:
+    - meta_scope: meta.annotation.scala
+    - match: \(
+      scope: punctuation.definition.arguments.annotation.begin.scala
```
wbond: Need this to be `punctuation.section.arguments...`

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
...
+      push:
+        - match: ','
+          scope: punctuation.separator.arguments.annotation.scala
+        - match: \)
+          scope: punctuation.definition.arguments.annotation.end.scala
```
wbond: Also `punctuation.section.arguments...`.

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
...
+          pop: true
+        - include: main
+    - match: \[
+      scope: punctuation.definition.arguments.annotation.begin.scala
```
wbond: Another `punctuation.section...`

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
...
+      push:
+        - match: \]
+          scope: punctuation.definition.arguments.annotation.end.scala
```
wbond: `punctuation.section...`

Scala/Scala.sublime-syntax
```diff
@@ -147,6 +147,46 @@ contexts:
...
+          pop: true
+        - include: delimited-type-expression
+    - match: \s*
```
wbond: I would probably move this as part of the next negative lookahead, e.g. `(?![(\[\s])`

Scala/syntax_test_scala.scala
wbond: This still needs to be addressed. The `meta` scope should probably be tested on a few of the different forms to make sure it terminates properly.

gwenzek:
@wbond sorry I shouldn't have committed at 1am ! I'll finish that soon.

Scala/Scala.sublime-syntax
FichteFoll: How about specializing this to `punctuation.accesor.dot.scala`, like it's done for comment punctuation?

I suggested this a few months ago already for C# iirc.

gwenzek: done.

@FichteFoll I kept "punctuation.accessor.scala" in order to stay coherent with the rest of the file. This scope renaming could go in a separate PR.

Scala/Scala.sublime-syntax
gwenzek: done

Scala/Scala.sublime-syntax
gwenzek: done

Scala/Scala.sublime-syntax
gwenzek: done

Scala/Scala.sublime-syntax
gwenzek: done

Scala/Scala.sublime-syntax
gwenzek: done

Scala/Scala.sublime-syntax
gwenzek: done

gwenzek:
Hi @wbond , I think this version should be ok.
I added tests for the "meta.scope" tell me if you think it's enough.

djspiewak:
The following is incorrectly highlighted with this PR:

```scala
x: Foo @abc with Bar
```

To see why it is incorrect, remove the annotation and note the scoping on Bar.

djspiewak:
@gwenzek I've PR'd (to you) a fix for the test case I referenced.

@wbond I've been dogfooding this branch for a few months now.  Aside from the aforementioned broken case (and its fix), it really looks ready to go.

gwenzek:
Thanks Daniel, I'll have a look a it soon.

I'm sorry I didn't took the time to finish that, I've had less time for
this kind of thing this last months.

wbond:
Sorry for the delay on this, and thanks for your work on it.

We did some discussion about annotation scopes at #709. I made a decision after I gathered some feedback: https://github.com/sublimehq/Packages/issues/709#issuecomment-266835130.

If you want to update this PR with those, great! If you won't have time for a while, I can pull this in and make the scope name tweaks myself.

djspiewak:
@wbond If you do end up pulling this in manually, be sure to grab my fix! (mentioned above) Without it, certain types involving `with` will be incorrectly scoped.

wbond:
@djspiewak Will do, thanks!

Scala/Scala.sublime-syntax
gwenzek: Done

gwenzek:
I included @djspiewak fix, renamed the annotation scope to `variable.annotation` and rebased on master.
I didn't introduce `meta.annotation.identifier` nor `meta.annotation.parameters` but  the first one would be equivalent to `meta.annotation variable.annotation` .

wbond:
@gwenzek Thanks for the tweaks!

The `meta.annotation.identifier` scope primarily exists for a multi-part identifier that includes accessors. I suppose it depends on the language for if it is needed. However, it probably wouldn't be a bad idea to use `meta.annotation.identifier variable.annotation` for a single identifier token. This way plugins can rely on always using `meta.annotation.identifier` to fully resolve the annotation name.

For instance, with Python a decorator could be `@module.ClassName.static_method(True)`. The entirety of `module.ClassName.static_method` would get `meta.annotation.identifier`, while only `static_method` would get `variable.annotation`, and the `.` would get `punctuation.accessor`. The part `(True)` would get `meta.annotation.parameters`, `(` would get `punctuation.section.parens.begin` and `True` would get `constant.language`.

Does that make sense?

FichteFoll:
I agree, except that `(` should get `punctuation.section.arguments.begin`.

wbond:
> I agree, except that ( should get punctuation.section.arguments.begin.

Currently the guidelines suggest `group` or `parens`. For most of the paired punctuation we've stayed away from semantic names since the semantics of the symbols can vary greatly between syntaxes. I guess the exception so far has been `generic`. I think in some situations we've used `angle` or `angles`.

FichteFoll:
Hm, I took that from the Python syntax definition after you worked on it. It uses `punctuation.section.parameters` and (in a PR of mine) `punctuation.section.arguments`, similar to how `meta.arguments` is used. I don't really care what we chose to do since all these punctiation tokens most likely also have a meta-scope associated with them, though.

wbond:
Yes, I am sure there are syntaxes that have been worked on in the past year that break the guidelines. I know Rust needs to be revisited for some things, and I imagine there are probably some tweaks that should be made in C/C++ and JavaScript.

> This reads more like an argument for my position, honestly.

As in, let's not use names for what they mean in the language, but rather what symbol they are. That way color schemes don't need to handle `arguments`, `parameters`, `tuple`, `group`, `cast`, etc. Hopefully they can just target `parens` and `group` and be done.

I suppose some color scheme authors will want to highlight `(`, `[` `{` all the same for macro parameters in Rust (they are all viable), which would be an argument for `parameters`. But I guess is someone really wants to do that they should be able to use a meta scope to target specific instances of `punctuation.section`.

----

Overall I think the best we can do is try to be consistent, and I think more abstract names are probably better, but this isn't a position I hold too strongly.

FichteFoll:
So, this is kind of related to what we discussed in #727, except it appears that I'm of the opposite opinion in this one. This actually makes me kind of wonder whether the decision in #727 of using `punctuation.accessor.dot` is the better one after all ... I'll need to think more about this.

I don't quite get `group`, though, which *is* semantic.

By the way, the guidelines don't have anything for `definition` vs `scope` either and you probably want to add that too when working on them.

wbond:
@FichteFoll Let's open a new issue if you want to continue discussing more ideas for clarifying the guidelines.

For now I am going to merge this and we can tweak it later as need be.

wbond:
I made some tweaks in eaf8321 that apply `meta.annotation.identifier` as I had originally intended, removed `variable.annotation` from the `@` symbol and only applies `variable.annotation` to the final identifier in the annotation name.

wbond:
@gwenzek Thanks for all of your work on this!

PR #606, opened 2016-08-28
[Scala] improve string highlighting

gwenzek:
As mentioned in #588 unicode escaped characters were improperly highlighted.
I also added support for octal characters, and check for improper escape sequences.

```scala
  "escaped chars: \u1221 \125 \n"
//                ^^^^^^ constant.character.escape.scala
//                        ^^^ constant.character.escape.scala
//                            ^^ constant.character.escape.scala

  "bad escaping: \p"
//               ^ invalid.illegal.lone-escape.scala

  """escaped in triple: \u1221 \125 \n"""
//                      ^^^^^^ constant.character.escape.scala
//                             ^^^ - constant.character.escape.scala
//                                  ^^ - constant.character.escape.scala
```

I also added support for "f-string" with a "printf" like escaping. I'm not sure which scope to use here.
```scala
   f"formatted: x: $x%+,.3f ca"
// ^ support.function
//                  ^ variable.other.scala
//                   ^^^^^^ constant.other.formatting.scala

   f"formatted: date: $x%T "
// ^ support.function
//                    ^ variable.other.scala
//                      ^^ constant.other.formatting.scala
```

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -622,6 +622,17 @@ contexts:
         - match: \n
           scope: invalid.string.newline.scala
         - include: escaped
+    - match: '(f)(""")'
+      captures:
+        1: support.function.scala
+        2: punctuation.definition.string.begin.scala
+      push:
+        - include: f_string
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.interpolated.scala
+        - match: '"""'
+          scope: punctuation.definition.string.end.scala
+          pop: true
     - match: '({{alphaid}})(""")'
       captures:
         1: support.function.scala
@@ -639,7 +650,15 @@ contexts:
       captures:
         1: support.function.scala
         2: punctuation.definition.string.begin.scala
-      push: f_string
+      push:
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.interpolated.scala
+        - include: f_string
+        - match: '"'
+          scope: punctuation.definition.string.end.scala
+          pop: true
+        - match: \n
+          scope: invalid.string.newline.scala
     - match: '({{alphaid}})"'
       captures:
         1: support.function.scala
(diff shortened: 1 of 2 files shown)


djspiewak:
@gwenzek :+1: to the escaping changes.  While you're at it, would you mind adding `punctuation` scope to the string delimiters?  (look at the scoping for `'` in YAML mode for an example)  This matches other languages, and it's on my TODO list, but since you're already in that code.

I'm not so sure about specifically styling the `f` string macro.  In theory it's not too different from `s`, but it's also incredibly infrequently used.  Doesn't affect me too much one way or another, though it is probably worth scoping both the double- and triple-quoted variants.

gwenzek:
@djspiewak I added the `punctuation` scope.

I'm not sure either about the `f`string. I just found it in the Scala documentation when I wrote my syntax.

FichteFoll:
As an aside, the string modifiers in Python use a `storage.modifier` scope.

djspiewak:
@FichteFoll I think `support.constant` makes more sense in Scala's case, given what it is semantically.  Also, `storage.type` is *really* overloaded.

FichteFoll:
Well, it's special syntax in Python and the most appropriate scope name in my opinion.

Since it's not special syntax in, say, JavaScript, it would probably need a `variable.function` scope there. Maybe that's already the case, too.

djspiewak:
@FichteFoll `support.function` actually probably makes more sense than `support.constant`.  Obviously in Scala it's not really built-in or a special syntax.

wbond:
So the `f` just needs to be scoped `variable.function` for this to be complete, correct?

gwenzek:
Well until now every `foo` in `foo"some string"` was scoped as `support.function` so I conserved this scope. As it stands this PR is consistent with itself, but if you want I can rename the scope to `variable.function`.

I rebased onto the latest changes, the only conflict was due to me adding an EOF character in the test file.

djspiewak:
@wbond @gwenzek I think it should remain consistent with the existing string interpolation highlighting.  `support.function` seems like a more appropriate scope in any case.

wbond:
Thanks!

PR #613, opened 2016-09-04
[Scala] < is a valid operator (when not followed by alphabeticals)

djspiewak:
In master, `<` (by itself) is not considered a valid identifier.  This PR fixes this case while respecting XML literals.  Note that there is one situation that will be improperly highlighted at present.  Consider the following:

```scala
x<y
x< y
x <y
x < y
```

The first of the four will be improperly highlighted (it will assume that `<` indicates an XML literal, rather than an operator).  All of the remaining three will be handled correctly by the syntax highlighting, including the third (which *correctly* identifies `<` as part of an XML literal).  I'm not sure how to handle the first one without some combinations of contortions or lookbehind, and it's super-rare anyway – as demonstrated by the fact that Martin Odersky was one of two people (total!) to correctly identify the parsing rules which govern this token – so I'm not going to stress out about it.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -22,6 +22,7 @@ variables:
       [[^:=<@\x{2190}\x{21D2}#]&&{{operator_character}}]{{operator_character}}*|
       =[[^>]&&{{operator_character}}]+|
       =>{{operator_character}}+|
+      <(?!{{operator_character}}|[[:alpha:]])|
       <[[^\-]&&{{operator_character}}]+|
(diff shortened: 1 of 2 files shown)


wbond:
It probably makes sense to start "organizing" the syntax test file into sections as opposed to just appending new test cases at the end for each PR. This will help make it so ever PR doesn't conflict once one has been merged.

djspiewak:
@wbond I agree.  Though actually just allowing the tail of the file to have a newline also fixes that problem.  Git doesn't seem to deal well with newline-less file termination.

PR #614, opened 2016-09-04
[Scala] Rewrote XML support

djspiewak:
As mentioned in #588.  This fixes some long-standing bugs in XML literals, including failure to correctly scope embedded Scala within attribute declarations, as well as generally not obeying the XML spec on names, entities, and really much of anything.  Depends on #613, since I didn't want to redo the lookahead changes.


wbond:
I ended up cherry picking the four new commits from here and then making some further tweaks on top of them. Thanks!

PR #623, opened 2016-09-17
[Scala] Reworked auto-indentation

djspiewak:
Scala's default auto-indentation has long been broken.  This makes it… uh… less broken.

Notably, this PR provides actual working support for auto-indenting `case` expressions.  For example:

``` scala
thing match {
  case foo =>
    stuff1
    stuff2

  case bar =>
    stuff3

    stuff4
}

println("hello world")
```

The above represents the idiomatic (and basically universally-accepted) indentation for `case` expressions.  It's very tricky to do this though, since the _first_ `case` represents an indent, while _all subsequent_ `case`s represent both an indent and a dedent.  So… tricky.

We do it with meta scoping.  There's a new `meta` scope which is applied by the mode, identifying first cases vs non-first cases.  That meta scoping is then used by the auto-indentation regular expressions to swap off the default case auto-indent (which only indents) and swap in the non-first auto-indent (which both indents and dedents).  Hurray?

Oh btw, for the record, scope-based filtering of the auto-indentation rules is determined by the scopes which are applied to the `\n` character on the line.  So you have to make sure that any such meta scoping is applied at the _end_ of the line, and not the beginning (this is the source of some contortion in the contexts).

There's one place where this trick doesn't work though.  If you take the above, select all and reindent it, you'll get the following:

``` scala
thing match {
  case foo =>
    stuff1
    stuff2

  case bar =>
    stuff3

    stuff4
  }

  println("hello world")
```

Yeah…  This is sadness.  It comes from the fact that Sublime does not provide a way to automatically indent (or in this case, dedent) multiple levels per line.  I made a suggestion on the forums as how this could be solved, but other than @FichteFoll, not many people chimed in.  (basically, I think indentation should be determined directly by meta scope (e.g. `meta.indent.1`, or perhaps repeated applications of `meta.indent`), without the clumsy secondary regexes in plist files)

This limitation also shows up in a couple other areas.  For example, the standard double-curly problem:

``` scala
{
  {
    foo
  }}

  bar
```

Hint: the above is _not_ correct, it's just what Sublime tries to do by default.

Anyway, I've been using these auto-indent rules for about a month now on varying code-bases, and I'm pretty confident they are about as correct as they can possibly be.  I haven't had any real "wtf" moments other than the closing curly on a pattern match, which I can't really do anything about.

I haven't addressed some specifically weird cases, like class or function declaration line wrapping.  This is intentional, mostly because those cases are extremely complicated and not at all standard in their handling.  If you want to auto-indent those things, use scalafmt.

### DIFF
--- a/Scala/Indent-case.tmPreferences
+++ b/Scala/Indent-case.tmPreferences
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+  <key>name</key>
+  <string>Indent (case)</string>
+  <key>scope</key>
+  <string>source.scala meta.block.case.scala - comment</string>
+  <key>settings</key>
+  <dict>
+    <key>decreaseIndentPattern</key>
+    <string>(?x)
+        ^ (.*\*/)? \s* \} .* $
+      | ^ (.*\*/)? \s* \) .* $
+      | ^ \s* case\b .* $
+    </string>
+  </dict>
+</dict>
+</plist>
--- a/Scala/Indent-case.tmPreferences
+++ b/Scala/Indent-case.tmPreferences
@@ -4,7 +4,7 @@
   <key>name</key>
   <string>Indent (case)</string>
   <key>scope</key>
-  <string>source.scala meta.block.case.scala - comment</string>
+  <string>source.scala meta.block.case.late - comment</string>
   <key>settings</key>
   <dict>
     <key>decreaseIndentPattern</key>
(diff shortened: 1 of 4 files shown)


keith-hall:
here is a link to the aforementioned forum post, in case people are interested :)

https://forum.sublimetext.com/t/configure-auto-indent-with-multiple-scopes-per-line/22408

djspiewak:
I found something that gets auto-indented quite poorly with these rules:

```scala
x match {
  case abc =>
  case sdf =>
    sdf match {
      case oiu =>
        here
    }
}
```

That kind of breaks pretty horribly, because the inner `case` is inside the non-first meta scope for the second outer case.  So I need to do some slightly more careful scope popping.

djspiewak:
This needs more work, so I'm closing it for now.  I'll come back to it later.

PR #624, opened 2016-09-18
[HTML] allow HTML autocompletion inside PHP HTML heredoc strings

keith-hall:
allow HTML autocompletion inside PHP HTML heredoc strings

i.e.

```
<?php echo <<<HTML
<!-- here, when user types for e.g. `sect`, it should suggest <section> -->
HTML;
?>
```


wbond:
I'm going to hold off on this until after the next beta since I feel it may need testing for possible side-effects.

PR #641, opened 2016-09-26
Allow spaces in shebang lines.

xavierholt:
Spaces are allowed between the `#!` and `/` bits - this'll make syntax detection friendlier for those of us who think the space helps readability.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -5,7 +5,7 @@ name: JavaScript
 file_extensions:
   - js
   - htc
-first_line_match: ^#!\s*/.*\b(node|js)$\n?
+first_line_match: ^#!\s*/.*\b(node|js)\b
 scope: source.js
 variables:
   identifier: '[_$[:alpha:]][_$[:alnum:]]*'
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -5,7 +5,7 @@ name: JavaScript
 file_extensions:
   - js
   - htc
-first_line_match: ^#!/.*\b(node|js)$\n?
+first_line_match: ^#!\s*/.*\b(node|js)$\n?
 scope: source.js
 variables:
   identifier: '[_$[:alpha:]][_$[:alnum:]]*'
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -68,7 +68,7 @@ file_extensions:
   - thor
   - Thorfile
   - Vagrantfile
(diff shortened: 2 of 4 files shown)


Ruby/Ruby.sublime-syntax
```diff
@@ -68,7 +68,7 @@ file_extensions:
   - thor
   - Thorfile
   - Vagrantfile
-first_line_match: ^#!/.*\bruby
+first_line_match: ^#!\s*/.*\bruby\b
```
xavierholt: I added the trailing `\b` here based on the Python regex, but then realized that all the first line regices are pretty different.  I'm happy to revert if this might break anything.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -5,7 +5,7 @@ name: JavaScript
 file_extensions:
   - js
   - htc
-first_line_match: ^#!/.*\b(node|js)$\n?
+first_line_match: ^#!\s*/.*\b(node|js)$\n?
```
xavierholt: The `node` and `js` interpreters aren't allowed to take `--flags` here?

Ruby/Ruby.sublime-syntax
wbond: I don't think I ever saw a binary name other than `ruby` in the few years I spent working in Ruby land. Instead, all of the versioning seems to be handled by shell script manipulating the `PATH` and similar tricks.

djspiewak: I've seen `erb` in shebang lines.  Also `jruby`.  But both of those are vanishingly rare.

xavierholt: I'll tweak this to allow `jruby` when I get home.  I don't think we should add `erb`, though - that'll most likely be some other language with a few embedded Ruby snippets and won't play nice with the Ruby syntax file.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -5,7 +5,7 @@ name: JavaScript
 file_extensions:
   - js
   - htc
-first_line_match: ^#!/.*\b(node|js)$\n?
+first_line_match: ^#!\s*/.*\b(node|js)\b
```
xavierholt: Went ahead and changed this to be a little more permissive.

xavierholt:
Updated according to our conversation about Ruby interpreters.

Ruby/Ruby.sublime-syntax
xavierholt: Changed this a day or so ago, but the Github UI seems confused...  Maybe the rebase + force push confused it.  The correct version shows in the "Files Changed" tab.

wbond:
Thanks!

PR #652, opened 2016-10-01
[Ruby] make `.rb` the default file extension

smanolloff:
When saving new files, the suggested extension is `.Appraisals`, apparently because its first in the list -- a regression introduced by 84396b2cfc7ad2ddb289c3d5bfecf4ae6a1a4f7d.

Moving `.rb` to the top solves this issue

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -41,6 +41,7 @@ name: Ruby
 # ternary operator rule, precedence stuff, symbol rule.
 # but also consider 'a.b?(:c)' ??
 file_extensions:
+  - rb
   - Appraisals
   - Berksfile
   - Brewfile
@@ -60,7 +61,6 @@ file_extensions:
   - rake
   - Rakefile
   - Rantfile
-  - rb
   - rbx
   - rjs
   - ruby.rail


jodyheavener:
I'm encountering this issue again unfortunately. I'm not sure how to check the version of my package manager, but my Sublime version is **Stable Channel, Build 3126**. Happy to create a new issue or provide more info if it is helpful.

keith-hall:
this change was merged after the latest ST build was released, so unless you have manually overridden it, you are probably still using the syntax definition that came with build 3126, which doesn't contain this fix.

beechnut:
For future readers who run into this issue:

- Go to [the directory where language files are stored](https://forum.sublimetext.com/t/where-are-tmlanguage-files-stored-for-st3/8712)
- Change the filename of Ruby.sublime-package to Ruby.zip—it's just a specially-named archive.
- Unzip Ruby.zip.
- Find Ruby.sublime-syntax and edit it. Reorder your file so it looks like [this](https://github.com/sublimehq/Packages/pull/652/commits/e7139d82277079558659defdd85b2019ba6cebe6).
- Save the file.
- I had trouble re-zipping with Finder on macOS, so I used Terminal to zip up the folder. Ensuring the archive is named `Ruby.sublime-package`. 
- Quit ST3 and reopen.

wbond:
@beechnut What version of ST3 are you running? This should be fixed in the latest beta and dev builds.

wbond:
Also, be aware that any edits to a `.sublime-package` file will be overwritten the next time you upgrade Sublime Text.

FichteFoll:
@wbond the latest beta build is from September 2016 but this was merged in October.

wbond:
@FichteFoll Hah, good point. I thought we released 3126 in late October. My bad!

PR #657, opened 2016-10-04
[Markdown] Fix invalid first_line_match in MultiMarkdown.sublime-syntax

jfcherng:
Recently I received an issue about a `first_line_match` compiling error [here](https://github.com/jfcherng/Sublime-AutoSetSyntax/issues/6) hence I make this PR.

The `first_line_match` in `MultiMarkdown.sublime-syntax`:
- Cannot be compiled with Python's `re` module.
```python
>>> re.compile(r'^Format:\s*(?i:complete)\s*$')
  File "<string>", line 1, in <module>
  File "./python3.3/re.py", line 214, in compile
  File "./python3.3/re.py", line 283, in _compile
  File "./python3.3/sre_compile.py", line 491, in compile
  File "./python3.3/sre_parse.py", line 747, in parse
  File "./python3.3/sre_parse.py", line 359, in _parse_sub
  File "./python3.3/sre_parse.py", line 708, in _parse
sre_constants.error: unknown extension
```
- According to [this](https://sublimetext.userecho.com/topics/4710-multimarkdown-mode-not-recognizing-format-complete-on-first-line-should-be-case-insensitive-for-keys/) report, the whole regex should be case insensitive.

### DIFF
--- a/Markdown/MultiMarkdown.sublime-syntax
+++ b/Markdown/MultiMarkdown.sublime-syntax
@@ -2,7 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: MultiMarkdown
-first_line_match: ^Format:\s*(?i:complete)\s*$
+first_line_match: (?i)^format:\s*complete\s*$
 scope: text.html.markdown.multimarkdown
 contexts:
   main:


keith-hall:
@jfcherng if I understand right, the regex wasn't/isn't invalid per se, just Python's `re` module doesn't support a few features. So ideally, we would try to keep the `first_line_match`es in the default packages always compatible with Python's `re` module so that they will be compatible with your plugin :)

jfcherng:
@keith-hall Mmm... I am not sure for other cases. But in this PR, I cannot see a obvious compatibility/performance issue if it follows rules for Python's `re` module.

keith-hall:
I didn't mean to imply that I thought there would be performance issues, I have no objection to this PR :)

jfcherng:
@keith-hall Yeah. I know that.

wbond:
This change shouldn't break anything.

Just know that `first_line_match` are used with the Oniguruma regex engine. This is a holdover from `.tmLangauge` and the fact that TextMate uses Oniguruma for all of its matching.

Kristinita:
@wbond, @keith-hall, are you sure, that changes merged? I install newest Sublime Text x64 portable version [**from official site**](https://www.sublimetext.com/3) → I install Package Control, Package Resource Viewer and AutoSetSyntax → I get [**same error**](https://github.com/jfcherng/Sublime-AutoSetSyntax/issues/6). <kbd>Ctrl+Shift+P</kbd> → `Package Resource Viewer: Open Resource` → `Markdown` → `MultiMarkdown.sublime-syntax` → I see line 5:

	first_line_match: ^Format:\s*(?i:complete)\s*$

Thanks.

keith-hall:
@Kristinita the latest ST3 release (Build 3126) was on the 23rd September 2016. This PR was merged on 10th October 2016. September < October, so this hasn't made it into an ST release yet.

michaelblyons:
@Kristinita You can follow the instructions in [README.md][readme] to use the new syntax definitions before they are released.

[readme]: https://github.com/sublimehq/Packages/blob/master/README.md

PR #658, opened 2016-10-05
[Python] Fix regex syntax

### LINKED ISSUE #656: Regex parenthesis bug since 3126
Hi,
Since 3126, it seems there's an issue with regex Scope.
Opened parenthesis equals **punctuation.definition.group.regexp** but closed parenthesis equals **meta.group.regexp** :
opened :
image
closed :
image

Just to clarify, opened **and closed** parenthesis were scoped **punctuation.definition.group.regexp** in build 3114.
Could you please have a look ?

Regards !

FichteFoll:
For some reason, the entire main context would push (and pop) an anonymous context for non-whitespace characters and exit it again for whitespace characters. That means that including the main context could lead to the including context not being able to pop out of its context properly, causing issues like #656.

Additionally, I made a quick pass over punctuation scope names and added "begin" and "end" scope leafs.

Fixes #656.

### DIFF
--- a/Python/Regular Expressions (Python).sublime-syntax
+++ b/Python/Regular Expressions (Python).sublime-syntax
@@ -26,9 +26,11 @@ contexts:
         - match: \|
           scope: keyword.operator.or.regexp
         - match: '\(\?\#'
+          scope: punctuation.definition.comment.begin.regexp
           push:
             - meta_scope: comment.block.regexp
             - match: \)
+              scope: punctuation.definition.comment.end.regexp
               pop: true
         - match: '\(\?[iLmsux]+\)'
           scope: keyword.other.option-toggle.regexp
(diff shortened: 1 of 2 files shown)


wbond:
I believe the push/pop was an attempt to properly handle highlighting in https://github.com/sublimehq/Packages/blob/master/Python/syntax_test_python.py#L713-L719. After the `#` shouldn't be considered a comment. I haven't pulled down these changes to see if it breaks that. Either way we should add some tests for that.

Technically comments should only happen in [`X`/`VERBOSE`](https://docs.python.org/2/library/re.html#re.X) mode, however that flag always comes after the pattern, and usually that flag is specified with multi-line patterns, so there isn't really a great way to handle it other than heuristics.

narusFR:
Hello there,
I'm concerned about this issue.
Could you please indicate me what's still needed for this to be pushed with an update ?
Thank you !

wbond:
There haven't been any releases recently as I've been working on some bigger, behind-the-scenes things.

My plan is to review open PRs before our next dev build.

narusFR:
Hello, this bug is quite frustrating to me, but you know... It's not that bad ... however I was still wondering if is there a target date for the next releases ? I understand it's not an easy thing, and I imagine there's a lot of stuff going on, I'm just asking.
Regards.

wbond:
This fix has been part of dev builds 3127-3132.

narusFR:
Thanks for the recent stable release fixing this, keep being awesome ;)

PR #659, opened 2016-10-05
[ASP] support method definitions containing explicit array parens

keith-hall:
In ASP, a method could be defined like:

    Sub Test1 (expect_array)
        Response.Write expect_array(1)
    End Sub

and it could also be defined like:

    Sub Test2 (expect_array())
        Response.Write(expect_array(1))
    End Sub

This PR is to ensure the latter case is highlighted correctly, as it is valid. (Interestingly, even in this case, ASP doesn't actually enforce that an array has to be passed, so it is mainly just a visual clue for the developer...)


wbond:
Thanks!

PR #664, opened 2016-10-10
[OCaml] Add 'let open' matching

### LINKED ISSUE #661: [OCaml] incorrect highlighting of `let open Something in`
Hello, I'm just learning OCaml and trying out examples from "Real World OCaml" book, here's one of them:

``` ocaml
open Core.Std

let do_hash file =
  In_channel.with_file file ~f:(fun ic ->
    let open Cryptokit in
    hash_channel (Hash.md5 ()) ic
    |> transform_string (Hexa.encode ())
    |> print_endline
  )

let spec =
  let open Command.Spec in
  empty
  +> anon ("filename" %: string)
```

Currently `let open Cryptokit in` is highlighted incorrectly. Even more, all syntax highlighting after it is broken.

P.S. I'm also learning to write syntax definitions for sublime text. So if I can help, where should I start?

faergeek:
It fixes #661. It will incorrectly highlight function definition like this `let open s = 5 + s`, but at least it does not break highlighting of everything after it.

### DIFF
--- a/OCaml/OCaml.sublime-syntax
+++ b/OCaml/OCaml.sublime-syntax
@@ -6,15 +6,22 @@ file_extensions:
   - ml
   - mli
 scope: source.ocaml
+variables:
+  identifier: '[a-z][a-zA-Z0-9''_]*'
+  capIdentifier: '[A-Z][a-zA-Z0-9''_]*'
 contexts:
   main:
-    - match: '\b(let)\s+(module)\s+([A-Z][a-zA-Z0-9''_]*)\s*(=)'
+    - match: '\b(let)\s+(module)\s+({{capIdentifier}})\s*(=)'
       scope: meta.module.binding
       captures:
         1: keyword.other.module-binding.ocaml
         2: keyword.other.module-definition.ocaml
         3: support.other.module.ocaml
         4: punctuation.separator.module-binding.ocaml
+    - match: \b(let|and)\s+(open)\s+{{capIdentifier}}(\.{{capIdentifier}})*\b
+      captures:
+        1: keyword.other.ocaml
+        2: keyword.control.import.ocaml
     - match: '\b(let|and)\s+(?!\(\*)((rec\s+)([a-z_][a-zA-Z0-9_'']*)\b|([a-z_][a-zA-Z0-9_'']*|\([^)]+\))(?=\s)((?=\s*=\s*(?=fun(?:ction)\b))|(?!\s*=)))'
       captures:
         1: keyword.other.function-definition.ocaml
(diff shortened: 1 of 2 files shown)


wbond:
Let's add a test or two for this. http://www.sublimetext.com/docs/3/syntax.html#testing has info about the test syntax.

If you felt so inclined, a basic test file that covered some basic like function names and variable definitions would be a great help!

I'd love to help getting this fix working even for `let open s = `. Are you aware of why this isn't working? I haven't worked with OCaml before (or this syntax definition), but I'd be happy to help improving it if you can help fill in some details for me.

faergeek:
Tests are great and easy to write, I'll start writing one a little later. I assume we could start with just this case for now? Do you have any additional guides for scope naming? I know only about this https://www.sublimetext.com/docs/3/scope_naming.html . Here (in OCaml) are many scopes meaning of which I do not fully understand

The problem with `let open s = 5` (function definition, although it does not use it's argument `s`) vs `let open SomeModule in` (somewhat like binding module during expression evaluation) is that we can't distinguish them, if they're written in strange way.

We could just assume that it starts with line like in those two examples, based on which we can assume what it is

```ocaml
let open s = # function
  let five = 5 in
    five

let open s in # module binding
s.some_fn ()
```

It's the same problem as detecting arrow function definition (and matching function name as `entity.name.function`) in JavaScript, if it's written on multiple lines. I don't know of any bulletproof solution.

wbond:
Yes, we can start with just a single test for now.

The scope naming documentation you linked to is our current set of guidelines. I don't think anyone has worked on OCaml recently, so there are surely many deviations.

It seems in this situation we should be able to use the presence of the `in` keyword and absence of `=` as a distinguishing factor. Will it be correct 100% of the time? No, but then again most of the syntaxes have edge cases that are very hard to handle. It will almost always be possible to come up with contrived examples that break. Ideally we just try to handle as much as we can.

faergeek:
I think something like this should work pretty good in most cases

faergeek:
Ping :-)

faergeek:
@wbond Do I need to do something more for this PR to be merged?

wbond:
Sorry this sat for a while! I've been working on some core stuff and hadn't circled back here to the packages for a while.

Definitely feel free to make other enhancements to the syntax. Currently we don't have a resident OCaml user around here making improvements, so your help would surely be appreciated.

faergeek:
@wbond Thanks. But now I use vim for that. There are some features for OCaml, which you can't find anywhere (except also Emacs, of course), e.g. merlin integration.

So I'm not sure that I will actively contribute to improve OCaml syntax here.

PR #666, opened 2016-10-12
[LaTeX] Support for breqn environments

neok-m4700:
breqn package is very popular together with amsmath for mathematical typesetting in LaTeX


wbond:
Thanks!

PR #673, opened 2016-10-25
[Rust] Cargo error format update

dten:
The error format from Cargo changed
https://github.com/rust-lang/sublime-rust/issues/84
so updating this package to match the new format. maybe should do both?


wbond:
Thanks!

PR #674, opened 2016-10-30
[Python] Various changes

### LINKED ISSUE #500: python: open brackets detector too impatient?
Write this.

``` Python
y = list()

def x():
    return
```

Now delete the closing bracket of `list()`. Both `def` and `return` are immediately marked as errors.
I think Sublime should perhaps not do this while the line is still being edited.

FichteFoll:
I found some time to finally fix a few shortcomings of the Python syntax that I noted down what feels like ages ago.

Notable changes (not exhaustive):
- Post-expression context was extended and is now applied to a lot more sub-expressions.
- Unclosed function definitions (or similar constructs) don't cause the entire file to look like a christmas tree (fixes #500)
- Added matrix multiplication operator
- Split test files into two (see commit msg for reasoning)

Remaining things to do on my list (in a different PR):
- f-strings
- improve current .format string matching (it's rather naive)
- match `abc = lambda:` as symbol
- keywords for class inheritance (notably `metaclass`)
- unpacking operators in function calls, definitions and collection literals
- match type hinting constructs (PEP 484)?


FichteFoll:
The string test file wasn't run (because I forgot to include the header), so I made a couple fixes and rebased.

wbond:
Awesome!

PR #676, opened 2016-11-02
[XML] Permit unescaped > in content

bobince:
Greater-than is actually valid in both text content and attribute values (see CharData and AttValue productions in XML 1.0 spec). Change should-be-entity pattern to reflect this.

The exception is that the sequence ]]> must not appear in text content (but it's OK in an attribute value, for some reason). Add missing-entity detection for that inside text content.

### DIFF
--- a/XML/syntax_test_xml.xml
+++ b/XML/syntax_test_xml.xml
@@ -131,7 +131,7 @@
      text
 <!-- ^^^^ text -->
      >
-<!-- ^ text -->
+<!-- ^ text - punctuation - illegal -->
 
      <![CDATA[<!DOCTYPE catalog plist "dtd">]]>
 <!-- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ string.unquoted.cdata - meta.tag - keyword -->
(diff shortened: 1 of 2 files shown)


XML/syntax_test_xml.xml
```diff
@@ -130,6 +130,8 @@
 <!--                           ^ punctuation.definition.tag.end -->
      text
 <!-- ^^^^ text -->
+     >
+<!-- ^ text -->
```
keith-hall: a better test here would be `text - illegal - punctuation`, to avoid regressions - `text` will always be true due to the base scope, `text.xml`

XML/syntax_test_xml.xml
```diff
@@ -182,6 +184,9 @@
 <!--                            ^ punctuation.definition.string.begin -->
 <!--                             ^^^^^ string.quoted -->
 
+     <element validinattr="a > b ]]> c"/>
+<!--                       ^^^^^^^^^^^ string.quoted -->
```
keith-hall: I think it would be a good idea to also add `- illegal - punctuation` here :)

wbond:
@bobince Could make the enhancements that @keith-hall suggested?

bobince:
Sorry, was afk for a bit! Yes, done—thanks, I didn't realise one could negative-assert like that.

wbond:
Thanks!

PR #677, opened 2016-11-02
[XML] Handle prolog and PIs separately

bobince:
Currently we are using the same match for the <?xml ... ?> prolog at the start of a file, and <?other ... ?> processing instructions in document content. They look similar, but they're separate in the XML syntax and PIs don't have attribute/quoted-value constructs inside them.

Match <?xml with the current handling, and add simpler handling for processing instructions which doesn't look for attr-like structure in the content. This prevents eg stray apostrophes in PI content from triggering invalidity.

Exclude any targets beginning with XML (case insensitive) from being PIs as per the PITarget production, so they don't match the prolog. Tweak spaces to better match spec: ban space after <? for both prolog and PI; require space after <?xml, and require either space after PITarget or empty PI, so that invalid PI name chars are detected.

### DIFF
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -25,7 +25,7 @@ variables:
 
 contexts:
   main:
-    - match: '(<\?)\s*({{qualified_name}})'
+    - match: '(<\?)(xml)(?=\s)'
       captures:
         1: punctuation.definition.tag.begin.xml
         2: entity.name.tag.xml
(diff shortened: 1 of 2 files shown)


XML/XML.sublime-syntax
```diff
@@ -25,7 +25,7 @@ variables:
 
 contexts:
   main:
-    - match: '(<\?)\s*({{qualified_name}})'
+    - match: '(<\?)(xml)(?=\s)'
```
wbond: Should this also match mixed case `xml`?

It seems with these changes, this will only match lowercase, and the new match won't match any capitalization of `xml`, leaving `XML` matched by nothing. Is that correct?

bobince:
> leaving `XML` matched by nothing. Is that correct?

Surprisingly yes, this is correct. `<?XML ... ?>` is not well-formed as it matches neither the XMLDecl nor the PI productions.

wbond:
Ok, thanks for the confirmation!

keith-hall:
For reference, here are the relevant parts of the XML RFC:
https://www.w3.org/TR/REC-xml/#NT-PITarget
https://www.w3.org/TR/REC-xml/#NT-XMLDecl

@bobince is right :)

wbond:
Thanks!

PR #679, opened 2016-11-02
Regexp scopes

FichteFoll:
Added `entity.name.capture-group` and `variable.other.backref-and-recursion` scopes.

Tests where kinda all around the place, so I grouped them a little like in the Python tests.


wbond:
Thanks!

PR #680, opened 2016-11-03
Update Scala.sublime-syntax

jpz:
This small change fixes a bug with the display of the comment `/**/` which was functioning as an unclosed block-comment.

This is fixed by moving the matching of `/**/` ahead of `/**`

### DIFF
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -306,6 +306,10 @@ type Foo = Bar[A] forSome { type A }
 // <- comment.block.documentation.scala
 */
 
+  /**/0xff
+//^^^^ comment.block.empty.scala
+//    ^^^^ - comment
+
    if
 // ^^ keyword.control.flow.scala
 
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -108,6 +108,8 @@ contexts:
     - include: late-operators
 
   block-comments:
+    - match: /\*\*/
+      scope: comment.block.empty.scala
     - match: /\*\*
       scope: punctuation.definition.comment.scala
       push:
@@ -130,8 +132,6 @@ contexts:
             (?x)
             			(?! /\*)
             			(?! \*/)
-    - match: /\*\*/
-      scope: comment.block.empty.scala punctuation.definition.comment.scala
 
   char-literal:
     - match: '''({{escaped_char}}|.)'''


jpz:
Before and after:

image

image

wbond:
We’ll definitely want to add a test for this.

djspiewak:
He beat me too it. :-D Adding a test that some identifier on the line after is *not* a comment should turn the trick.

FichteFoll:
Can also target the following newline character.

jpz:
Thanks for the reviews - I've added a test.  

I added a trailing value to the test (a simple hex value) to demonstrate that the remainder of the line is normal code.

image

Scala/syntax_test_scala.scala
```diff
@@ -306,6 +306,10 @@ type Foo = Bar[A] forSome { type A }
 // <- comment.block.documentation.scala
 */
 
+  /**/0xff
+//^^^^ comment.block.empty.scala
+//    ^^^^ - thereafter, not a comment
```
FichteFoll: Change this line to

```
//    ^^^^ - comment 
```

jpz: It demonstrates the comment has terminated and text thereafter is not regarded as a comment.  How does describing the hexadecimal number as a comment help in this case?

Kind regards

keith-hall: @jpz it sounds like there is some confusion about how the tests and scope selectors work. If the line starts with a comment, and contains `^` or `<-` then it is a test assertion line (described at http://www.sublimetext.com/docs/3/syntax.html#testing). The rest of the line then dictates the scope selector. Because the scope selector is `- comment` and not just `comment`, it means, assert that the scope does *not* match `comment`. I've never really found any good documentation on scope selectors, but an answer I wrote on SO may be useful: http://stackoverflow.com/questions/37406338/sublime-text-modifying-tmtheme-file/37407588#37407588

jpz: Thanks for explaining and pointing me in the direction of the documentation I should have read :)

Will amend the PR tomorrow.

jpz: done

jpz:
Ready to merge @wbond I think

wbond:
Thanks for the patch @jpz and the reviews @FichteFoll and @djspiewak!

PR #687, opened 2016-11-11
[Scala] Scoping for the varargs type operator

djspiewak:
A suffix `*` is the syntax by which you declare varargs function parameters in Scala.  Confusingly, infix `*` is not a primitive operator, and may be declared in user-land.  So that makes the lookahead a bit tricky.


Scala/Scala.sublime-syntax
```diff
@@ -390,6 +390,8 @@ contexts:
           push:
             - match: '{{typeprefix}}'
               set:
+                - match: '\*\s*(?=[\),=])'
+                  scope: keyword.operator.varargs.scala
```
wbond: Only the `*` should be scoped here, not the spaces.

Scala/Scala.sublime-syntax
```diff
@@ -452,6 +454,8 @@ contexts:
...
```
wbond: Same here, use a capture so the spaces aren't scoped.

Scala/Scala.sublime-syntax
djspiewak: Derp.  You're right of course.

wbond:
Thanks!

PR #688, opened 2016-11-11
[Scala] Corrected issue with anonymous AnyRef inner class scoping

djspiewak:
Scala supports a C#-like syntax for declaring anonymous inner classes with structural types that extend `AnyRef`.  Like so:

```scala
val x = new {
  def foo = 42
}

x.foo  // => 42
```

This syntax is very rarely used, because it's quite slow to evaluate, but it does exist.  We were previously scoping this incorrectly, since the braces were considered part of a valid single-line type declaration.  This meant that the class body was scoped as a type, which is obviously incorrect.  It is now fixed.

Amusingly, as you can see from the above snippet, we weren't the only ones to forget about this construct when building a syntax highlighting for Scala.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -524,7 +524,15 @@ contexts:
     - match: '\b(new)(?:\s+|\b)'
       captures:
         1: keyword.other.scala
-      push: single-type-expression
+      push:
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #696, opened 2016-11-20
[LaTeX] Enable percent sign comments in pure TeX

r-stein:
Currently comment toggling only works in LaTeX and not in pure TeX (https://github.com/SublimeText/LaTeXTools/issues/949).


wbond:
Thanks!

PR #697, opened 2016-11-20
[Scala] Fixed single-line class declarations with bodies

djspiewak:
For example;

```scala
class Foo { val bar = 42 }
```

The `{` was not properly detected, and so the entire line was considered part of the class meta definition, meaning that the `val`, the `42` and so on were not properly highlighted by the `main` context.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -419,7 +419,7 @@ contexts:
       set: class-parameter-list
     - match: '\['
       push: class-tparams-brackets
-    - match: '(?=extends|with)'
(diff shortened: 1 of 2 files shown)


djspiewak:
Worth noting that this functionality (but not the test) is superceded by #725

wbond:
@djspiewak Should I merge this, or is the test part of #725?

djspiewak:
@wbond I think it's worth merging.  It doesn't really conflict with #725 outside of a trivial two-character bit.  I didn't include the test in #725, but there are other tests in that PR which are very similar.

PR #702, opened 2016-11-23
[PHP] Fix functions completion with void parameter

gwarnants:
The initial PHP completions file has many errors, especially about functions having no parameters.
ie: this extract from the PHP documentation (http://php.net/manual/en/function.func-get-args.php)

`
array func_get_args ( void )
`

was translated like this in the completions file:

`
		{ "trigger": "func_get_args", "contents": "func_get_args(${1:oid})" },
`

but the _${1:oid}_ has no sense here. So it result of an extra "oid parameter" appearing on many PHP completions hints. The line should be like this:

`
		{ "trigger": "func_get_args", "contents": "func_get_args()" },
`

My fix will correct all functions that matches this pattern.


wbond:
Awesome, thanks for fixing this!

Issue #685, opened 2016-11-04
[LaTeX Syntax] newcolumntype in package array


RomanKarwacik:
If I use the command `\newcolumntype{L}{>{$}l<{$}}` the highlighting is a bit off:
![newcolumntype1](https://cloud.githubusercontent.com/assets/8593000/20015718/d2b636f8-a2bc-11e6-962a-9219c6956ed0.PNG)

Elliot2560:
I'm having the exact same issue. Went through a bunch of debugging and found out that the root of the issue was in the syntax definition. It seems that the second `$` in each of the commands is not matched because there is a hanging `{`. The same behaviour occurs with sequences like `$ { $` (granted, this is not a valid use case).

r-stein:
It may be an idea to automatically end the math mode if we match an unescaped and unmatched `}`. This is not the perfect solution, but would help in such cases and not break the whole remaining file.

Elliot2560:
I considered that, but I don't think you can really have unmatched delimiters unless in macros, so I don't think it's wise to exit math mode this way. LaTeX is a bit similar to C because without basically preforming preprocessing, it's impossible to perfect cover all changes on the default syntax highlighting due to macros. So I think it's more realistic to just cover the default `newcommand` and commands from popular packages, which in this case is `>{}`, `<{}`, etc., from `tabular`. Or maybe adding preprocessing utilities to LaTeXTools? :P

FichteFoll:
The blessing and the curse of a powerful macro system in programming languages.

I should check out how Rust's macro system works just so I can understand how it does things differently.

wbond:
This issue has been solved by a previous commit, hence it is being closed.

PR #708, opened 2016-11-26
[LaTeX] Added syntax for array preamble options

Elliot2560:
Solves [issue #685 ](https://github.com/sublimehq/Packages/issues/685). Syntax for two macro commands are added for use with the `array` package. (cf. [stack exchange](https://tex.stackexchange.com/questions/257128/how-does-the-newcolumntype-command-work))


wbond:
Pinging @r-stein, @randy3k and @FichteFoll for review/feedback

LaTeX/LaTeX.sublime-syntax
```diff
@@ -225,6 +225,8 @@ contexts:
         - include: general-commands
         - include: macro-braces
 
+    - include: array-preamble
```
r-stein: This should be at the [packages context](https://github.com/Elliot2560/Packages/blob/65b6d484b4de3bc79d8150f582eeb1c46332968d/LaTeX/LaTeX.sublime-syntax#L772)

LaTeX/LaTeX.sublime-syntax
```diff
@@ -1655,3 +1657,30 @@ contexts:
        3: punctuation.definition.group.brace.begin.latex
        4: entity.name.function.frame.latex
        5: punctuation.definition.group.brace.end.latex
+
+  array-preamble:
+    - match: '(>)\s*(\{)'
```
r-stein: As far I know this is only possible inside column definitions, i.e. `\newcolumntype{X}{>{...}l}` and `begin{tabular}{|l|>{...}l}`, so we should not match this everywhere.

Elliot2560:
@r-stein Thank you for your review! I thought including all column types, but that is not the issue that I was trying to address. The main problem is that the `>{}` and `<{}` syntax should be treated like macros, since according to the [package doc](http://mirror.jmu.edu/pub/CTAN/macros/latex/required/tools/array.pdf), these can contain singled out environment delimiters. Though `!{}` and `@{}` are similar, they should not contain singled out environment delimiters. Perhaps I will add support of the latter two, since on a second thought, even though they shouldn't provide a use case like `>{$}`, syntactically they should be the same as the former two. Regardless, `l`, `p{}`, `m{}`, etc. should not be treated the same.

LaTeX/LaTeX.sublime-syntax
Elliot2560: Thank you, I didn't fully understand the structure of the file before. I will update the PR.

r-stein:
I agree, that this is not necessary to fix that bug. However we should add popular packages (as the array package) and not just fix visible bugs. You will always be able to create a highlighting bug anyway as you can with `#ifdef` in C.
If we have something like `\begin{tabular}{|l|@{See: }>{$}l<{$}|}` and only highlight `>` with `support.function...` and not the `@` this does not look reasonable to me.
I suggest:

| sign                | scope               |
| ---                 | ---                 |
| l, r, c             | keyword.other...    |
| \|                  | keyword.operator... |
| p, @, m, n, >, <, ! | support.function... |

I don't want do discourace you. If you are not so experienced with writing syntax files, I can also have a look at it.

LaTeX/syntax_test_latex.tex
```diff
@@ -91,7 +98,7 @@ \subsubsection{name}
 % ^ support.function.reference.latex keyword.other.reference.latex
 %        ^ constant.other.reference
 
-\cite{my:bib:key}
+\cite{my:bib:key }
```
randy3k: please remove the space.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -1656,3 +1657,137 @@ contexts:
...
+            4: variable.parameter.function.tabular.latex
+            5: punctuation.definition.group.brace.end.latex
+          pop: true
+        - include: main
+
+  array-preamble:
+    - match: 'l|r|c'
+      scope: keyword.other.column-type.latex
+
+    - match: 'p|m|b'
```
r-stein: We may change this to `'(?:p|m|b)(?=\s*\{)'` to ensure it is followed by a brace, otherwise the syntax breaks for a missing `{...}`.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -1656,3 +1657,137 @@ contexts:
...
+        2: punctuation.definition.backslash.latex
+        3: punctuation.definition.group.brace.begin.latex
+        4: variable.parameter.function.latex
+        5: punctuation.definition.group.brace.end.latex
+        6: punctuation.definition.group.bracket.begin.latex
+        7: punctuation.definition.group.bracket.end.latex
+        8: punctuation.definition.group.brace.begin.latex
+      push:
+        - meta_scope: meta.environment.tabular.latex
+        - match: '(?<=\{)'
```
r-stein: I think we should avoid lookbehind pattern, because it slows down the regex parser. @wbond can you confirm this?
This could be done via an other context and a `set`-chain.

LaTeX/syntax_test_latex.tex
```diff
@@ -42,6 +42,13 @@
 %             ^support.function.latex entity.name.newcommand.latex
 %                               ^ support.function.general.latex
 
+\newcolumntype{x}{>{$}c<{$}}
```
r-stein: You could also test, that `meta.function.newcolumntype.latex` stretches over the whole command.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -1656,3 +1657,137 @@ contexts:
        3: punctuation.definition.group.brace.begin.latex
        4: entity.name.function.frame.latex
        5: punctuation.definition.group.brace.end.latex
+
+  # support for array package
+  array:
```
r-stein: I would rename this to pkgarray, because array is also an environment.

LaTeX/LaTeX.sublime-syntax
randy3k: Good catch

FichteFoll: It will indeed slow down the parser. This can be checked with a Syntax Test build variant.

FichteFoll:
I only reviewed this briefly but will trust @r-stein's judgement (and also agree on his comments). I also rate my LaTeX expertise as average.

Elliot2560:
I just now noticed the discussion about lookbehind from [a previous Scala PR](https://github.com/sublimehq/Packages/pull/563). I didn't quite get the solution using blank match to pop the context, which I think is also what's used in the `general-*-argument` contexts here. I used nested set instead, per @r-stein 's advice. Now at least the tests pass... The only issue I see with this is that the `meta.environment.tabular.latex` scope does not include the `\begin{tabular}` statement anymore. I could fix that by adding this scope to each layer of push/set, but I'm not quite sure if that would split up the scope and introduce error were it to be used in packages.

Elliot2560:
As it turns out sublime merges adjacent identical scopes, so just adding the tabular scope to each level of the set chain fixes the scoping issue.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -1706,10 +1706,11 @@ contexts:
         6: punctuation.definition.group.bracket.begin.latex
         7: punctuation.definition.group.bracket.end.latex
       push:
+        - meta_scope: meta.environment.tabular.latex
         - match: '\{'
           scope: punctuation.definition.group.brace.begin.latex
           set:
-            - meta_scope: meta.function.column-spec.latex
+            - meta_scope: meta.function.column-spec.latex meta.environment.tabular.latex
```
r-stein: This should be `meta.environment.tabular.latex meta.function.column-spec.latex`

LaTeX/syntax_test_latex.tex
```diff
@@ -382,6 +390,27 @@ \subsubsection{name}
 %                    ^ source.python.embedded
 
 
+% PACKAGE: array
+% The array package extends array and tabular environments
+% and allows for macros in format specifications
+
+\begin{tabular}[t]{|x|@{See: }>{$}l<{$}|}
```
r-stein: Would be good to add
``` tex
% ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.environment.tabular.latex - meta.environment.tabular.latex meta.environment.tabular.latex
```
to avoid duplicates scopes and ensure the scope is over the whole command.

r-stein:
At the braces the scope is duplicated. I don't know the best practise and how to solve this, would be nice if @FichteFoll could have a look how to preserve the meta scope over the set-chain.

FichteFoll:
Duplicating the scope is the normal procedure.

r-stein:
Sorry I have phrased myself wrong, the resulting scope at the braces is

text.tex.latex **meta.environment.tabular.latex meta.environment.tabular.latex** meta.function.column-spec.latex punctuation.definition.group.brace.begin.latex

instead of

text.tex.latex **meta.environment.tabular.latex** meta.function.column-spec.latex punctuation.definition.group.brace.begin.latex

FichteFoll:
I'll look into it when I'm at home (and don't forget).

wbond:
I'm working on merging PRs, but it appears this still need a little work, correct?

LaTeX/LaTeX.sublime-syntax
```diff
@@ -1656,3 +1657,137 @@ contexts:
...
+      captures:
+        1: support.function.begin.latex keyword.control.flow.begin.latex
+        2: punctuation.definition.backslash.latex
+        3: punctuation.definition.group.brace.begin.latex
+        4: variable.parameter.function.latex
+        5: punctuation.definition.group.brace.end.latex
+        6: punctuation.definition.group.bracket.begin.latex
+        7: punctuation.definition.group.bracket.end.latex
+      push:
+        - meta_scope: meta.environment.tabular.latex
```
r-stein: @wbond From my point of view the only remaining problem is, that this scope does not span correctly over the following set-chain. E.g. you have

text.tex.latex **meta.environment.tabular.latex meta.environment.tabular.latex** meta.function.column-spec.latex punctuation.definition.group.brace.begin.latex

instead of

text.tex.latex **meta.environment.tabular.latex** meta.function.column-spec.latex punctuation.definition.group.brace.begin.latex

at the braces. Hence it is not ready to pull, but only a little work left.

FichteFoll:
```
From 3ed205ea829fbdfec33d88e818195cc35bec5529 Mon Sep 17 00:00:00 2001
From: FichteFoll <fichtefoll2@googlemail.com>
Date: Wed, 14 Dec 2016 14:35:44 +0100
Subject: [PATCH] [LaTeX] Define meta.environment.tabular like align

The align environment does not include the \begin and \end commands in
its meta scope. Imo it makes sense not to do this for the tabular
environment too.

Note that "lstlisting" and "minted" currently do *not* do this and
include the \begin and \end matches in their meta scopes.
---
 LaTeX/LaTeX.sublime-syntax  | 5 ++---
 LaTeX/syntax_test_latex.tex | 4 ++++
 2 files changed, 6 insertions(+), 3 deletions(-)

diff --git a/LaTeX/LaTeX.sublime-syntax b/LaTeX/LaTeX.sublime-syntax
index 2de2090..ac2a038 100644
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -1706,16 +1706,15 @@ contexts:
         6: punctuation.definition.group.bracket.begin.latex
         7: punctuation.definition.group.bracket.end.latex
       push:
-        - meta_scope: meta.environment.tabular.latex
         - match: '\{'
           scope: punctuation.definition.group.brace.begin.latex
           set:
-            - meta_scope: meta.function.column-spec.latex meta.environment.tabular.latex
+            - meta_scope: meta.function.column-spec.latex
             - include: array-preamble
             - match: '\}'
               scope: punctuation.definition.group.brace.end.latex
               set:
-                - meta_scope: meta.environment.tabular.latex
+                - meta_content_scope: meta.environment.tabular.latex
                 - match: '((\\)end)(\{)(tabular)(\})'
                   captures:
                     1: support.function.end.latex keyword.control.flow.end.latex
diff --git a/LaTeX/syntax_test_latex.tex b/LaTeX/syntax_test_latex.tex
index 9bc56af..79de675 100644
--- a/LaTeX/syntax_test_latex.tex
+++ b/LaTeX/syntax_test_latex.tex
@@ -395,19 +395,23 @@ def my_function():
 % and allows for macros in format specifications
 
 \begin{tabular}[t]{|x|@{See: }>{$}l<{$}|}
+% ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - meta.environment.tabular.latex
 %                 ^^^^^^^^^^^^^^^^^^^^^^^ meta.function.column-spec.latex
 %                  ^ keyword.operator.inter-column-line.latex
 %                     ^ support.function.inter-column-nospace.latex
 
 \hline
+%^^^^^^ meta.environment.tabular.latex
 
 a & b
 % <- meta.environment.tabular.latex
+%^^^^^ meta.environment.tabular.latex
 
 \hline
 
 \end{tabular}
 %     ^ variable.parameter.function.latex
+%^^^^^^^^^^^^ - meta.environment.tabular.latex
 
 
 
-- 
2.9.0.windows.1
 ```

Elliot2560:
I disagree that meta scopes for environments should not include the begin and end commands. I can't find if there is an official LaTeX reference that settles this semantic issue. However I think since the begin and end commands specify the behaviour of the text enclosed between them, they should be part of the environment. It is true that the `amsmath` environments do not include the begin and end commands, but the other environments, including the lists, inline and block math, and environments defined in other packages, all include the delimiters in the `meta.environment` scope. That's why I decided to include the two commands in the tabular meta scope.

The other potential problem is, the syntax tests cannot test if the scopes are duplicated. Even when they are the tests still pass. I suspect the testing utility just makes use of the `view.match_selector` method, which does not care how many times a selector is matched.

keith-hall:
@Elliot2560 Scope selectors allow you to check for duplicate scopes like this:

    desired.scope - desired.scope desired.scope

example (found using regex search: `\^\s+[\w.-]+\s+-\s+([\w.-]+)\s+\1`): https://github.com/sublimehq/Packages/blob/358be560407d7c3ef2a273d4905aa4aad0f63180/PHP/syntax_test_php.php#L388

Elliot2560:
@keith-hall Thank you! I added that to the test now.

jrappen:
@Elliot2560 Thanks for this PR. Are you done here? If so, please ping Will.

r-stein:
I have now created 2 PRs (https://github.com/Elliot2560/Packages/pull/1 https://github.com/Elliot2560/Packages/pull/2). If at least https://github.com/Elliot2560/Packages/pull/1 is merged, I will approve it.

PR #713, opened 2016-11-29
[YAML] Don't scope leading whitespace as comment

FichteFoll:
Just realized I had this branch lying around without opening a PR.

Should be self-explanatory.


wbond:
Thanks!

PR #714, opened 2016-11-30
[Scala] Replaced the Scala snippets

djspiewak:
The old set of Scala snippets were really not very useful.  In most cases, they produced highly unidiomatic code.  In several cases, they produced code which was either actively discouraged by the language or simply so rarely used as to be unsuitable for templatization.   Many of the snippets literally expanded their tab trigger into "their tab trigger + one space".  The only one which I ever used on a regular basis was `case`.

The problem(?) is that Scala isn't a very boilerplate-ish language, so there isn't a ton that benefits from templating.  However, there are a *few* things, most of them falling into the same category as `case`, and I have added those templates.  I have a few other templates which I have locally which I personally find useful (e.g.`tlam` and `alph`), but I doubt those are widely applicable.


wbond:
@gwenzek I'm totally unqualified to evaluate these. If you have a couple of minutes, I'd be grateful for any feedback, pro or con.

wbond:
I haven't seen any feedback here, so I am going to merge and we can deal with feedback afterwards

gwenzek:
Sorry for missing this.
I agree that the old snippets weren't really useful. These ones looks better, even if some like the "var" snippet only save you two chars.

PR #720, opened 2016-12-05
[LaTeX] missing iota and Iota commands

randy3k:
the `\iota` and `\Iota` are missing.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -114,7 +114,7 @@ contexts:
         - include: macro-braces
 
   greeks:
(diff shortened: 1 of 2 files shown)


Elliot2560:
I don't think `\Iota` command is used. It's just `I`.

randy3k:
Oh, you are right.

randy3k:
removed `\Iota`.

wbond:
Can we add a test for this?

FichteFoll:
That would imply we need a test for every of these constants. (I didn't check whether there are any tests for Greek symbols at all.)

wbond:
I would prefer to add tests for new additions or bug fixes because it means that the original implementation missed it. That further implies it is more of an edge case, or didn't come to mind to the original implementer. This means it is more likely to be missed in future modifications, so a test would be useful in preventing regressions.

I am not going to expect someone submitting a bug fix to add tests for all combinations of existing code. But ideally our test "coverage" will increase over time, not decrease.

randy3k:
I have to admit that it was my fault to not include `\iota` in the first round (it is also because it is not one of the commonly used greek letters`. I will add a test for it.

wbond:
Thanks!

PR #721, opened 2016-12-05
[Java] Misc Fixes

djspiewak:
Still several things to be addressed, especially related to the over-use of `storage.type`, but it's getting better.  Notably fixed here:

- Scoping on `class`/`enum`/`interface` to be in line with the guidelines
  + Also improved user-experience while typing with better lookahead
- Added `volatile` to the set of keywords
- Scoping for function calls
- Tweaked `extends`/`implements` scoping to be uniform with Scala's `extends`/`with`

Further high priority things to address in later PRs:

- `storage.type` is applied to any identifier-ish thing in random code
- Generic brackets (`<`/`>`) are not properly scoped in types
- Lambdas are entirely unscoped
- Annotations are barely scoped

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -105,6 +105,9 @@ contexts:
             - match: "}"
               pop: true
             - include: class-body
+  functions:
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #722, opened 2016-12-08
[LaTeX] improve tests for parbox and makebox

randy3k:
For the general-*-arguments, the context is popped by the match '\}',  the non-space are not necessary
In other places, `set` was used in the match argument, those non-space matches are also not needed

PS: the test for `\parbox` is also improved.

### DIFF
--- a/LaTeX/syntax_test_latex.tex
+++ b/LaTeX/syntax_test_latex.tex
@@ -266,10 +266,14 @@ $\alpha_$
 \mbox{text}{text}
 %       ^ meta.function.box.latex
 %             ^ -meta.function.box.latex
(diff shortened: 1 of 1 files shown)


wbond:
Won't this cause the state machine to get stuck in the `general-1-argument-with-options` (and similar) context if no `{` or `[` is found?

I think it other syntaxes we either pop out upon no match, or mark as an error. That way the whole rest of the document isn't in some funky state.

randy3k:
You are definitely correct. I have almost missed it.  So now I have just improved the tests.

wbond:
Thanks!

PR #723, opened 2016-12-09
[ASP] scope member access

keith-hall:
scope member access just like PHP does (for `var->member`), in ASP the syntax is `var.member`

previously, ASP was scoping the `member` part as `variable.other.asp`, now it is more specific, as `variable.other.member.asp`, to make it consistent with the PHP syntax, and the scope naming guidelines at http://www.sublimetext.com/docs/3/scope_naming.html#variable


wbond:
Thanks!

Perhaps in the future split unrelated scope changes into another commit so it is easier to review?

PR #724, opened 2016-12-09
[ShellScript] Keyword matching fixes and syntax tests

carlosefr:
The changes in this pull request prevent most keywords from being matched in a bunch of wrong contexts. It isn't fool proof, but makes it more palatable.

I understand that the ShellScript syntax is due to a complete rewrite, however I think it's still useful to make it a little saner while that rewrite doesn't happen. Anyway, I'm using this and I guess it doesn't hurt to open a pull request for it... :)


FichteFoll:
Using look-behinds hits very hard on performance due to the internal sregex enging not being compatible with them and ST falling back to Oniguruma.

carlosefr:
The first commit has some examples of bad behavior:

  * In `rpm -q yum-utils` the word "yum" is highlighted.
  * In `rpm --import ...` the word "import" is highlighted.

There were already a few lookbehinds in there. What do you think the alternative should be here? Would captures be better?

I don't notice too much of a performance hit (or any, for that matter) but I'm not editing very large shell scripts.

FichteFoll:
Ideally, look-behinds should just go away since you can emulate them using context mangling with empty match patterns. That's partially why a rewrite is needed.

Regarding tests, I was mostly referring to implementing *actual* tests via http://www.sublimetext.com/docs/3/syntax.html#testing.

carlosefr:
Ok, but in the context of minimal changes to improve the current syntax definition, is it objectively better to replace (most) look-behinds with captures?

As for tests, I'll take a look at the documentation.

carlosefr:
I've added a bunch of test cases.

I also found a few more bugs like `while` or `printf` being matched as `keyword.other` instead of a block or a builtin function respectively, and fixed those too.

FichteFoll:
So, this isn't your fault since it was already like that before, but I just noticed that "common commands" all get a `keyword` scope. That is semantically wrong and should be some `support` scope, just like the built-in functions, imo.

carlosefr:
I noticed that too, but I'm wondering if it breaks the looks on themes people are using, and if it isn't a change that should happen only with a bigger rewrite. But I'll look into it.

Also, that list of commands is too big, I think. It mixes commands that take other commands as arguments as part of their normal operation (eg. `xargs`, `sudo`) and commands highlighted just for the sake of it (eg. `install`, `zip`). These should be broken into separate scopes, IMHO.

Any suggestions for the names of these scopes? Is there a hard nomenclatures for this? Examples: `support.command.shell` for the likes of `sudo` and `support.other.shell` for the likes of `zip`?

FichteFoll:
Scope name suggestions sound good to me.

Ultimately, it is on @wbond to decide whether or not the scope name should be changed (and imo it should, since `keyword` is really, really wrong). The default color schemes do need an overhaul anyway (`support` is not highlighted by Monokai for example, but this is on Will's radar).

jfcherng:
Will this somehow become another ShellScript Improved eventually?

carlosefr:
@jfcherng: Right now I'm doing this for myself because the inconsistencies in this syntax definition annoy me personally and because I think any improvement is better than no improvement while we wait for a proper rewrite that may never happen (like I said above).

I have no intention of maintaining this separately for public consumption. If it happens to not be acceptable for merging, I'll just consider this as my local configuration (meaning no promises on compatibility, consistency or anything other than my personal preferences).

carlosefr:
I've been using this personally for a while, feel that it's definitely an improvement over the version shipped with ST3 and haven't seen any regressions or relevant issues that I felt the urge to fix (or that were fixable without a complete rewrite).

Can this be considered for inclusion?

wbond:
Unfortunately, I don't want to increase the number of look behinds, since it is important to me that the default syntaxes get rid of those for performance reasons.

My hope is to do a ground-up rewrite of shell script (with variants for bash, zsh, csh, etc) in the next month or two. The tests from this PR will definitely be helpful in that.

carlosefr:
OK, fair enough.

FichteFoll:
Having used the shell script syntax in the past few weeks, I can definitely feel the pain. It's in a usable state but there are so many things wrong with it that a rewrite is definitely needed. Unfortunately, I have other priorities right now.

rwols:
Related: https://github.com/sublimehq/Packages/pull/1057

wbond:
#1057 should hopefully be a good place for us to continue further work from. I appreciate your willingness to contribute @carlosefr, and hopefully the revised Bash syntax from @rwols (along with @jfcherng and @okdana) can be something that you'll find useful to work from.

Issue #694, opened 2016-11-17
[Python] Python 3.6 hits 2016-12-16 (with new syntax)


FichteFoll:
According to [PEP 494](https://www.python.org/dev/peps/pep-0494/), Python 3.6 will hit in one month with a couple new syntax features to sport. Nothing breaking, however. The list of changes is available here: https://docs.python.org/3.6/whatsnew/3.6.html.

Syntax changes:

- [x] https://www.python.org/dev/peps/pep-0498 (https://github.com/sublimehq/Packages/issues/726)
- [x] https://www.python.org/dev/peps/pep-0515 (#728)
- [x] https://www.python.org/dev/peps/pep-0526 (#730)
- [x] https://www.python.org/dev/peps/pep-0530 (#729)

The remaining changes would include adding a couple identifiers to special highlighting, which is trivial.

I am volunteering to work on this as soon as either my other PRs are merged or I start using 3.6 (which will most likely happen on the release day). Or in case you intend to plan a new dev release soon.

wbond:
This appears to be done. Thanks @FichteFoll!

PR #726, opened 2016-12-10
[Python] f-strings

FichteFoll:
Adds support for the largest syntax addition to Python 3.6. Normal string placeholders have also been revised.

See #694.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -668,8 +668,22 @@ contexts:
       scope: support.type.python
 
   constant-placeholder:
-    - match: '%(?i:\([a-z_]+\))?#?0?\-? ?\+?([0-9]*|\*)(\.([0-9]*|\*))?[hlL]?[acdeEfFgGiorsuxX%]'
+    - match: |-
+        (?x)
+        %
+          ( \( ({{identifier}}) \) )? # mapping key
+          \#?            # alternate form
+          0?             # pad with zeros
+          \-?            # left-adjust
+          \ ?            # implicit sign
+          [+-]?          # sign
+          (\d*|\*)       # width
+          (\. (\d*|\*))? # precision
+          [hlL]?         # length modifier (but ignored)
+          [acdeEfFgGiorsuxX%]
       scope: constant.other.placeholder.python
+      captures:
+        2: variable.other.placeholder.python
     - match: '%(?:[aAwdbBmyYHIpMSfzZjUWcxX%]|-[dmHIMSj])'
       scope: constant.other.placeholder.python
     - match: '\{\{|\}\}'
@@ -682,9 +696,9 @@ contexts:
           (:            # format_spec
             (.? [<>=^])?       # fill align
             [ +-]?             # sign
-            \#?
+            \#?                # alternate form
             \d*                # width
-            ,?
+            ,?                 # thousands separator
             (\.\d+)?           # precision
             [bcdeEfFgGnosxX%]? # type
           )?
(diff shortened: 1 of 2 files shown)


Python/syntax_test_python_strings.py
```diff
@@ -445,3 +430,120 @@
...
+
+# FIXME The opening brace incorrectly gets its scope stack reset
+# due to https://github.com/SublimeTextIssues/Core/issues/1526.
+# Thus, the following tests on it are failing.
+
+f"{something}"
+#^^^^^^^^^^^^ meta.string.interpolated
+# <- storage.type.string
+#^ punctuation.definition.string.begin
+# ^ punctuation.definition.interpolation.begin
```
wbond: This should be `punctuation.section.interpolation`, right?

Python/Python.sublime-syntax
```diff
@@ -842,6 +837,107 @@ contexts:
...
+      scope: invalid.illegal.empty-expression.python
+    - match: (?=\{)
+      push: f-string-replacement
+    - match: \}
+      scope: invalid.illegal.stray-brace.python
+
+  f-string-replacement:
+    - clear_scopes: 1
+    - match: \}
+      scope: meta.interpolation.python punctuation.definition.interpolation.end.python
```
wbond: I think this should be `punctuation.section.interpolation`

Python/Python.sublime-syntax
```diff
@@ -842,6 +837,107 @@ contexts:
...
+      pop: true
+    - match: \{
+      scope: punctuation.definition.interpolation.begin.python
```
wbond: `punctuation.section` here also

Python/syntax_test_python_strings.py
```diff
@@ -445,3 +430,120 @@
...
+#              ^ source - meta, string, source source
+
+f"{True!a:02f}"
+#^^^^^^^^^^^^^^ meta.string.interpolated
+# ^ - source source.python.embedded
+#  ^^^^ source source.python.embedded constant.language
+#      ^^^^^^^ - source source.python.embedded
+#      ^^ storage.modifier.conversion - constant.other.format-spec
+#        ^^^^ constant.other.format-spec
+#            ^ punctuation.definition.interpolation.end
```
wbond: `punctuation.section` here

Python/syntax_test_python_strings.py
```diff
@@ -445,3 +430,120 @@
...
+# FIXME The opening brace incorrectly gets its scope stack reset
+# due to https://github.com/SublimeTextIssues/Core/issues/1526.
+# Thus, the following tests on it are failing.
+
+f"{something}"
+#^^^^^^^^^^^^ meta.string.interpolated
+# <- storage.type.string
+#^ punctuation.definition.string.begin
+# ^ punctuation.definition.interpolation.begin
+#           ^ punctuation.definition.interpolation.end
```
wbond: `punctuation.section` on these two lines

Python/syntax_test_python_strings.py
FichteFoll: Well, this one is difficult since you cannot split interpolated "sections" over lines in single-line strings but you can split multi-line strings. Unfortunately, the syntax definition doesn't differ between these cases yet. I'll take a look soon.

I wonder whether I should use `section` and `definition` for them or just `section`, though.

FichteFoll:
I decided I won't implement newline popping in single-line f-string interpolations for now and wait until I can use with_prototype instead.

wbond:
Thanks!

Issue #709, opened 2016-11-28
Annotations/Attributes Scope Naming


wbond:
Currently the following syntaxes that are part of the default (or will become part of the default) have constructs called annotations, attributes or decorators that modify a function/class/type/etc:

 - Java
 - Python
 - Rust
 - C#
 - Scala
 - Swift

They are not all 100% functionally equivalent, but all seem to generally include:

 1. Some sort of punctuation to signal the construct
 2. An identifier
 3. Optional other distinct tokens including things like parens, strings, integers, etc

Currently the scopes are very inconsistent across syntaxes, and I'd like to get it standardized so users and color scheme authors can benefit. Here are links to some existing discussions about them:

 - https://github.com/sublimehq/Packages/pull/605
 - https://github.com/sublimehq/Packages/issues/704
 - https://github.com/sublimehq/Packages/pull/93#issuecomment-176951572
 - https://github.com/rust-lang/sublime-rust/pull/124
 
My thoughts on the current state of annotations/attributes:

 1. While historically they have been highlighted as comments in some syntaxes, I believe this is a disservice to the user since they can contain multiple tokens
 2. The entirety of the construct should get `meta.annotation` or `meta.attribute`.
 3. I think that a consistent `punctuation.` scope should be applied to the leading symbol
 4. The identifier is probably the crux of the discussion. Java currently uses `storage.type`. The Scala PR has proposed `storage.modifier.annotation`. Rust is currently scoping as a `comment.`. Python scopes the `@`, but just applied a `meta.` scope to the rest. I forget where C# is right now. We don't yet have Swift in the default packages.
 5. All tokens after the identifier should probably get the standard scopes.

Based on this, I think the decisions that need to be made are:

 1. **Should we use `meta.annotation` or `meta.attribute` as the meta scope?** It seems C#, Rust, Swift use the term attribute. Java and Scala use annotation. (Python's decorator construct is slightly different, but I feel it probably makes sense to scope it similarly.)
 2. **What should the scope of the identifier be?** I think `storage.type` is used too much in general. `storage.modifier.annotation` feels a little too semantically strict to have this be a general-purpose scope name, but has the benefit of being implemented by existing color schemes. `variable.annotation` would be consistent with `variable.function`, but the later has always felt a little odd. The other idea is to start using a new top-level scope name of `identifier.` that allows for more fine-grained control of identifier in various contexts. The downside of this is that no color schemes currently support it.

[...]

PR #727, opened 2016-12-10
[Java] More fixes!

### LINKED ISSUE #481: Remove Incompatible Regex Patterns
The following languages need to be improved to remove regex patterns that cause us to hit the Oniguruma library:

 - [ ] AppleScript
 - [ ] Clojure
 - [ ] D
 - [ ] Erlang
 - [ ] Groovy
 - [ ] Java
 - [ ] LaTeX (in progress at #370)
 - [ ] Lua
 - [ ] Makefile
 - [ ] Markdown
 - [ ] Matlab
 - [ ] OCaml
 - [ ] Perl
 - [ ] Python (minor)
 - [ ] Scala
 - [ ] ShellScript (see #434)
 - [ ] Textile (minor)

As noted above, some are minor tweaks, while others probably need a major overhaul.

djspiewak:
Lots of stuff.  In no particular order:

- Removed all of the lookbehind (updates #481)
- Added support for lambdas (I'm pretty sure anyway…)
- Fixed scoping of `?`, `extends` and `super` inside of generics
- Removed a metric boatload of `storage.type`
- Fixed some random punctuation scoping that was incorrect (including `.`)
- Added kitchy `invalid` scoping on primitive-instantiated generics (it's pretty safe to say these will always be invalid, given how generics are implemented on the JVM)
- Implemented full-state field declaration parsing

There's a bit of a question here though, and I don't know the answer.  Specifically:

```java
private String memberString = "Hello";
```

What should the scope of `String` be?  Clearly, if we replace `String` with `int`, the answer is `storage.type.primitive.java` with some additional meta scoping for fun.  But it's a lot less clear what we do if it's a non-primitive type.

The C++ mode says "no scoping at all", which is certainly a perspective I can rationalize.  The Scala mode says either `support.class.java` or `support.type.java` (probably the former, since Java is *super* class-oriented).  The strict reading of the scoping guidelines says `storage.type.java`.

Initially, I tried `storage.type`, but then I ran into the following signature:

```java
private Map<Foo, Bar> map = null;
```

Should the `storage.type` just cover the `Map`?  Or should it cover the whole type?  Both look very weird.  And what if the type contains a wildcard, resulting in some `keyword.operator` scoping in the *middle* of our `storage.type`?  That doesn't make any sense.

By the same token, extending the Scala `support.class`/`support.type` regime to Java also has some problems.  `<` is just a more syntactically heavyweight token than `[`, which means that Java's generics just seem more visually intrusive when the operators are not colored uniformly with the surrounding types.  Or maybe I'm just used to reading Scala.  Either way, it seems pretty darn weird.

I left the `support.type` and `support.class` scopes commented out, just so I can remember where they need to be rematerialized.  For the time being, we're just going to go with the C++ mode's minimalist scoping, rather than Scala's rich (and complicated) scoping.  Java types are a lot closer to C++ types (in terms of average complexity) than Scala types, so I think that's somewhat justifiable, even if it does make code look a little plain.

Anyway, I think the only remaining item for the Java mode is to fix annotations, which you could argue is pending a resolution on #709.

---

**Disclaimer**  I do not use Java on a regular basis.  Or at all, really.  So unlike the Scala mode, I'm not dogfooding any of this.  I'm happy to fix bugs and corner cases, but I'm probably not going to find them myself.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -7,11 +7,15 @@ file_extensions:
   - bsh
 scope: source.java
 variables:
-  primitives: \b(void|boolean|byte|char|short|int|float|long|double)\b
+  primitives: \b(boolean|byte|char|short|int|float|long|double)\b
+  primitives_with_void: \bvoid\b|{{primitives}}
 
   # source: http://stackoverflow.com/a/5205467/9815
   id: '(?:[\p{L}_$][\p{L}\p{N}_$]*)'
   qualified_id: '(?:(?:{{id}}\.)*{{id}})'
+
+  # utility lookaround
+  within_generics: '(?:<(?:[^<>]|<(?:[^<>]|<[^<>]*>)*>)*>)'  # handles 3 levels nesting
 
 contexts:
   prototype:
@@ -79,47 +83,119 @@ contexts:
           scope: punctuation.separator.property.java
     - match: '@\w*'
       scope: storage.type.annotation.java
+
   anonymous-classes-and-new:
-    - match: \b(?<!::)new\b
+    - match: \bnew\b
       scope: keyword.control.new.java
-      push:
-        - match: '(?<=\)|\])(?!\s*{)|(?<=})|(?=;)'
+      push: instantiation
+
+  instantiation:
+    - match: '{{primitives}}'
+      scope: storage.type.primitive.java
+      set: instantiation-arrays
+    - match: '{{qualified_id}}'
+      scope: support.class.java
+      set: instantiation-generics
+    - match: '(?=\S)'
+      pop: true
+
+  instantiation-generics:
+    - match: '<'
+      scope: punctuation.definition.generic.begin.java
+      set:
+        - meta_scope: meta.generic.java
+        - match: '>'
+          scope: punctuation.definition.generic.end.java
+          set: instantiation-arrays
+        - include: generic-instantiation
+    - match: '(?=\S)'
+      set: instantiation-arrays
+
+  instantiation-arrays:
+    - match: \[
+      scope: punctuation.section.brackets.begin.java
+      set:
+        - meta_scope: meta.brackets.java
+        - match: \]
+          scope: punctuation.section.brackets.end.java
           pop: true
-        - match: '(\w+)\s*(?=\[)'
-          captures:
-            1: support.class.java
-          push:
-            - match: '}|(?=;|\))'
-              pop: true
-            - match: '\['
-              push:
-                - match: '\]'
-                  pop: true
-                - include: code
-            - match: "{"
-              push:
-                - match: "(?=})"
-                  pop: true
-                - include: code
-        - match: (?=\w.*(\(|$))
-          push:
-            - match: \)
-              pop: true
-            # - include: object-types
-            - match: \(
-              push:
-                - match: (?=\))
-                  pop: true
-                - include: code
-        - match: "{"
-          push:
-            - meta_scope: meta.inner-class.java
-            - match: "}"
-              pop: true
-            - include: class-body
+        - include: code
+    - match: \(
+      scope: punctuation.section.parens.begin.java
+      set:
+        - meta_scope: meta.parens.java
+        - match: \)
+          scope: punctuation.section.parens.end.java
+          set: instantiation-anon-class
+        - include: code
+    - match: '(?=\S)'
+      pop: true
+
+  instantiation-anon-class:
+    - match: \{
+      scope: punctuation.section.braces.begin.java
+      push:
+        - meta_scope: meta.class.anonymous-inner-class.java
+        - match: \}
+          scope: punctuation.section.braces.end.java
+          pop: true
+        - include: class-body
+    - match: (?=\S)
+      pop: true
+
   functions:
-    - match: '\b([a-zA-Z](?:[^\.<>\s]*))(?=(\s*<.+>)?\s*\()'
+    - match: '\b{{id}}(?=(\s*<.+>)?\s*\()'
       scope: variable.function.java
+
+  fields:
+    - match: '{{primitives}}(?=(\s*\[\s*\])?\s+{{id}})'
+      scope: storage.type.primitive.java
+      push: field-definition-arrays
+    - match: '\b{{qualified_id}}(?=(\s*{{within_generics}})?(\s*\[\s*\])?\s+{{id}})'
+      scope: storage.type.class.java
+      push: field-definition-generics
+
+  field-definition-generics:
+    - match: '<'
+      scope: punctuation.definition.generic.begin.java
+      set:
+        - meta_scope: meta.generic.java
+        - match: '>'
+          scope: punctuation.definition.generic.end.java
+          set: field-definition-arrays
+        - include: generic-instantiation
+    - match: '(?=\S)'
+      set: field-definition-arrays
+
+  field-definition-arrays:
+    - match: \[
+      scope: punctuation.section.brackets.begin.java
+      set:
+        - match: \]
+          scope: punctuation.section.brackets.end.java
+          set: field-definition-name
+    - match: '(?=\S)'
+      set: field-definition-name
+
+  field-definition-name:
+    - match: '{{id}}'
+      pop: true
+
+  generic-instantiation:
+    - match: \?
+      scope: keyword.operator.wildcard.java
+    - match: \b(extends|super)\b
+      scope: keyword.declaration.extends.java
+    - match: '{{qualified_id}}'
+      scope: support.type.java
+    - match: '<'
+      scope: punctuation.definition.generic.begin.java
+      push:
+        - match: '>'
+          scope: punctuation.definition.generic.end.java
+          pop: true
+        - include: generic-instantiation
+
   assertions:
     - match: \b(assert)\b
       scope: keyword.control.assert.java
@@ -184,6 +260,7 @@ contexts:
     - include: constants-and-special-vars
     - include: anonymous-classes-and-new
     - include: keywords
+    - include: fields
     - include: functions
     - include: storage-modifiers
     - include: strings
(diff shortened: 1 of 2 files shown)


keith-hall:
With `private Map<Foo, Bar> map = null;`, I would personally apply a meta scope to the whole `Map<Foo, Bar>`, and scope `Map`, `Foo` and `Bar` separately as types, and give punctuation scopes to the `<`, `,` and `>`.

djspiewak:
Would you scope them as `support.class`?

djspiewak:
Restored scoping and employed `support.class` and `support.type`.

Java/Java.sublime-syntax
```diff
@@ -67,47 +85,134 @@ contexts:
...
-    - match: \b(?<!::)new\b
+    - match: \bnew\b
       scope: keyword.control.new.java
+      push: instantiation
+
+  instantiation:
+    - match: '{{primitives}}'
+      scope: storage.type.primitive.java
+      set: instantiation-arrays
+    - match: '{{qualified_id}}'
```
wbond: For this I'd like to use a look-ahead and them push into a scope where we match `.` as `punctuation.accessor`.

Java/Java.sublime-syntax
```diff
@@ -227,7 +338,15 @@ contexts:
                 - match: "}"
                   pop: true
                 - include: class-body
+
   keywords:
+    - match: '::'
```
wbond: This should probably be either `punctuation.accesor` (that would be for more "color" consistency at the expense of semantic correctness), or `keyword.operator` for semantic correctness, correct?

Java/Java.sublime-syntax
```diff
@@ -227,7 +338,15 @@ contexts:
...
+      scope: keyword.control.method-reference.java
+      push:
+        - match: '{{id}}'
```
wbond: Is there a scope we could add here that would be useful?

Java/Java.sublime-syntax
```diff
@@ -270,8 +390,17 @@ contexts:
           push:
```
wbond: Add a scope for the `(`?

Java/Java.sublime-syntax
```diff
@@ -270,8 +390,17 @@ contexts:
...
             - meta_scope: meta.method.identifier.java
             - match: \)
+              scope: meta.method.identifier.java
```
wbond: Probably should add a `punctuation` scope for the `)`.

Java/Java.sublime-syntax
```diff
@@ -285,69 +414,144 @@ contexts:
...
-          push:
-            - meta_scope: storage.type.generic.java
-            - match: '>|[^\w\s,\[\]<]'
+        - match: '<'
+          scope: punctuation.definition.generic.begin.java
+          set:
+            - meta_scope: meta.generic.java
+            - match: '>'
+              scope: punctuation.definition.generic.end.java
               pop: true
```
wbond: Should we be including a context here to scope the types in the generic list?

Java/Java.sublime-syntax
```diff
@@ -285,69 +414,144 @@ contexts:
...
+      scope: storage.modifier.java
+    - match: '{{primitives}}'
+      scope: storage.type.primitive.java
+      set: method-parameter-after
+    - match: '{{qualified_id}}'
+      scope: support.class.java
+      set: method-parameter-after
+    - match: ','
+      scope: punctuation.separator.java
+    - match: \)
```
wbond: Let's add a `punctuation` scope here

Java/Java.sublime-syntax
```diff
@@ -365,21 +570,34 @@ contexts:
...
+        - match: (?=\{|;)
           pop: true
-        - include: object-types
+        - match: '{{qualified_id}}'
+          scope: support.class.java
+          push:
+            - match: '<'
+              scope: punctuation.definition.generic.begin.java
+              set:
+                - match: '>'
```
wbond: We should add `meta.generic.java` to this, right?

Java/Java.sublime-syntax
djspiewak: I think `keyword.operator.method-reference` might make sense.  `punctuation.accessor` would also be sort of accurate.  This is a deference operator which returns a pointer rather than invoking the method.  I don't have a strong feeling about it, but I will say that my color scheme (which colors `keyword.operator` and not `punctuation`) makes the test example look nicer with a `keyword`-based scope than a `punctuation` one.

Java/Java.sublime-syntax
djspiewak: Most definitely.  This is a bug.

Java/Java.sublime-syntax
djspiewak: Yep

Java/Java.sublime-syntax
wbond: From reading about this it seemed as though most authors considered it more of an operator, in that it transforms something, rather than a basic accessor. So let's go with `keyword`.

Java/Java.sublime-syntax
wbond: Yeah, now that you confirmed this is for static references to methods, I believe that would be the correct scope.

Java/Java.sublime-syntax
djspiewak: Missed this in my initial scan of your review…

This would be a trickier change, since we're nesting `{{qualified_id}}` matches within *other* lookahead contexts (for field matching).  It's not impossible, and I agree it's the right thing to do, it's just probably going to be very messy.  If it's alright with you, I'd rather tilt that windmill in a separate PR, since it's going to require some deeper changes.

Java/Java.sublime-syntax
djspiewak: Oh, correction: it's already included on line 429.  I was puzzled when I attempted to add a failing test and it passed.  :-)

djspiewak:
@wbond I believe everything except the `qualified_id` bit is now addressed.  As I said, due to the way this interacts with the other lookahead contexts, I would prefer to address that in another PR where I can fix some of the other structural issues with the mode.

Java/Java.sublime-syntax
FichteFoll: >but I will say that my color scheme (which colors keyword.operator and not punctuation) makes the test example look nicer with a keyword-based scope than a punctuation one.

That's totally subjective though. I'm fine with my Python code not showing dots as blue (for `keyword.operator`).

Either way, this is apparently a discussion about whether or not accessors are in fact "operators". I vote for a `keyword.operator.accessor` scope for all of them, if we were to decide on them being operators.
By the way, semantically this has the same scope as dot access for attributes like `this.field_name`, but I don't think those are scoped as `keyword.operator`, are they?

FichteFoll:
Note that #481 will be closed when this PR is merged becuase it includes `fixes #481` in the OP. I highly suggest editing that to read `updates #481` instead.

Java/Java.sublime-syntax
djspiewak: > By the way, semantically this has the same scope as dot access for attributes

It doesn't really though.  The best analogue is honestly the `->` vs `.` operators in C, except sort of inverted.  The `.` operator in C directly accesses a member, while the `->` operator dereferences a pointer and *then* accesses the member.  The `.` operator in Java accesses a member function and applies it, while the `::` operator accesses a member function and creates a pointer to it (without application).

In the C/C++ mode, both `.` and `->` are given scope `punctuation.accessor`.  In Java, `.` is (now) given `punctuation.accessor` while `::` is given `keyword.operator.method-reference`.  That's certainly a bit inconsistent.

By another perspective though, the `::` operator in Java is analogous to the `&` operator in C (creating a pointer), which is scoped as `keyword.operator.c`.

djspiewak:
@FichteFoll Done.

Java/Java.sublime-syntax
wbond: I agree with both of you, and as you can see from my original comment I didn't have a strong opinion either way.

For consistency with other punctuation operators, including the `::` in PHP, let's go with `punctuation.accessor`. That was my first instinct, however the docs I had been reading about it made it sound as if it was doing more. I think the visual consistency and style-ability wins here over pure semantic correctness. Kind of like how we scope docstrings in Python as comments because they are used as comments, but technically are just strings.

djspiewak: @wbond I'll make the change.  I preferred `keyword.operator` before we were scoping the identifier on the right (since it offered a strong visual distinction), but now that we're scoping the id, there's no real reason to avoid the punctuation scope.

Either way, I don't have much of a horse in the race since I don't use Java at all.  :-D

wbond: Well, I certainly appreciate all of these fixes you are doing considering you never really use it!

djspiewak:
@wbond Done.  If you approve the deferral (to a subsequent PR) on the `qualified_id` issue, this should be good to go.

Java/Java.sublime-syntax
```diff
@@ -227,7 +338,16 @@ contexts:
                 - match: "}"
                   pop: true
                 - include: class-body
+
   keywords:
+    - match: '::'
+      scope: punctuation.accessor.java
```
FichteFoll: Would you mind changing this to `punctuation.accessor.double-colon.java`?

Java/Java.sublime-syntax
```diff
@@ -253,25 +373,33 @@ contexts:
       scope: keyword.operator.arithmetic.java
     - match: (!|&&|\|\|)
       scope: keyword.operator.logical.java
-    - match: (?<=\S)\.(?=\S)
-      scope: keyword.operator.dereference.java
+    - match: \.
+      scope: punctuation.accessor.java
```
FichteFoll: And this to `punctuation.accessor.dot`.

FichteFoll: Great. Some suggestions from my side.

`.` → dot
`::` → double-colon
`->` → arrow
`=>` → fat-arrow ? (probably not used as an accessor anywhere but whatever)
`\` → backslash (yes, Moonscript uses this...)

wbond: PHP uses `\` for namespace "accessor", in the way that C++ uses `::`, although right now I believe in PHP it is scoped as a `punctuation.separator.namespace`.

wbond: Additionally, `:` is used by Lua. Anyway, we can probably continue any further conversation about this on a new issue rather than flooding this PR.

djspiewak:
@FichteFoll Revised to use `dot` and `double-colon`.  I agree on your proposed guidelines, btw.  Using the syntactic name for these operators rather than a language-specific semantic term seems the most sane.

wbond:
Thanks!

PR #730, opened 2016-12-11
[Python] Variable annotations

FichteFoll:
https://www.python.org/dev/peps/pep-0526

Chose the simple route.

Ref #694

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -70,6 +70,8 @@ contexts:
       scope: keyword.control.flow.return.python
     - match: \b(break|continue|pass)\b
       scope: keyword.control.flow.python
+    - match: ':'
+      scope: punctuation.separator.annotation.variable.python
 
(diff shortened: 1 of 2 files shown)


wbond:
Do you think at some point we should treat `[]` in annotations as `punctuation.definition.generic`? I suppose technically it isn't strictly that since you can put any valid AST there, right? It is just that item access is being used as if it were a generic structure, correct?

FichteFoll:
This isn't relevant to this issue at all, is it?

Anyway, item access is not valid in Python decorators. You may only directly use an identifier or a single function call. https://docs.python.org/3/reference/compound_stmts.html#function

FichteFoll:
You are right that any expression is valid for annotations, so scoping `[]` as `generic` may be wrong in occasions.

I have type hints revolving around `Dict`, `Tuple` etc. as a separate feature on my TODO.

PR #731, opened 2016-12-14
[ASP] add missing Comments.tmPreferences file

keith-hall:
add missing `Comments.tmPreferences` file to allow "toggle comment" functionality to work inside ASP files


wbond:
Thanks!

PR #732, opened 2016-12-14
[Python] Use punctuation.accessor.dot

FichteFoll:
Found this in my unpushed branches~

Imo it's a good idea to incldue the puncutation character in this scope name, similar to the comment punctuation. People might want different colors for `::` scopes in C++/Rust than for `->` in C or Php. Similarly, dots deserve their own scope name.


wbond:
Thanks!

PR #733, opened 2016-12-14
[YAML] Change block scalar indicator scopes

FichteFoll:
Also from my unpushed branches.

`keyword.control.flow` and `storage.modifier` seem more appropriate and the characters are important enough to warrant a colored scope.


wbond:
Sounds good!

PR #740, opened 2016-12-16
[Scala] Meta scopes for case expressions

djspiewak:
There aren't a lot of guidelines in the scope documentation as to what meta scopes are appropriate here, so I took a crack at it.  Could probably add one or two more overlapping scopes, since I think it's probably more useful to avoid overlapping the `case` and `=>` tokens (since avoiding the overlap allows block selection of the pattern or block body without getting a spare token).

Critically, these meta-scopes distinguish the first `case` block from later ones, making it possible to implement tooling (such as auto-indentation) which correctly handles this edge case.


wbond:
Thanks!

Issue #737, opened 2016-12-15
RFC: Annotation qualifier scope names


wbond:
With new guidelines for annotations in #709, we didn't address the issue of scoping identifier qualifiers, i.e. "paths". Example identifiers with qualifiers:

Python

```python
@ClassName.method(True)
def my_func():
    pass
```

Scalar

```scala
@scala.beans.BeanProperty
var status = ""
```

Java

```java
@qualifier.ClassName
class Foo { }
```

My intention is to scope the final identifier as `variable.annotation`, with `.`, `::`, etc as `punctuation.accessor` and the `@` (or other symbol) as `punctuation.definition.annotation`.

The question raised in #735 is if the whole qualified identifier should get `variable.annotation`, or if classes/namespaces/modules should get scopes they would in other parts of the syntax. For instance something like `support.class` in some cases.

What do you think @djspiewak, @gwenzek, @FichteFoll, @keith-hall, @michaelblyons?

keith-hall:
I personally think they should get the scopes they would in other parts of the syntax - if users/color scheme authors want, they can make the whole annotation the same color by utilizing the meta scope, while allowing others to have it presented in the same colors as they have come to expect from the main document.

djspiewak:
Quoting myself from the other issue, so that the discussion is in one place:

> That's fair, though I think syntactically it might look a little weird because the `@` is to the far *left* (whereas function calls have nothing to the left).  Additionally, in the original version of the proposal, the "annotation scope" (whatever it ended up being) was supposed to be applied to the `@`, overlapping the `punctuation` scope.  Is that changed now?
>
> If it's unchanged, it will end up with this rather odd situation where `variable.annotation` starts, then stops after one character and is replaced by "normal" scoping, and then starts again for a single identifier.
>
> FWIW, nearly all editors which have any scoping at all for Java/Scala annotations tend to scope the entire token string, including quantifiers.

FichteFoll:
It would look as much out of place as with just `scala.beans.BeanProperty` in a normal statement, so I'm with @keith-hall. By the way, I *will* add a dimmed background highlight for my color scheme that targets `meta.annotation` and will include potential parameters, as I already have currently.

I posted an example of this in #704:

![example](https://cloud.githubusercontent.com/assets/931051/20636727/a1e9af66-b372-11e6-9113-a9958d5ee148.png)

djspiewak:
[...]

PR #743, opened 2016-12-19
[Python] Decorator refactoring

FichteFoll:
**Requires https://github.com/SublimeTextIssues/Core/issues/1190 to be fixed before merging!**

---

When trying to implement proper scoping according to #709 and #737, I noticed that I would have to rebuild almost the entire `function-calls` scope for decorators only with different scope names. Since I religiously hate duplicated code (especially considering I would have to duplicate it *twice*), I worked on refactoring how qualified names are matched instead and `qualified-name-until-leaf` was born. 

Unfortunately, it is affected by the bug described in https://github.com/SublimeTextIssues/Core/issues/1190 and thus is not ready to be merged yet. The following 17 failing tests should be fixed when this bug is fixed:

```
Packages/Python/syntax_test_python.py:102:2: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:3: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:4: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:5: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:6: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:7: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:8: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:9: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:102:10: [meta.qualified-name variable.function] does not match scope [source.python meta.function-call.python]
Packages/Python/syntax_test_python.py:508:7: [meta.annotation] does not match scope [source.python invalid.illegal.character.python]
Packages/Python/syntax_test_python.py:508:8: [meta.annotation] does not match scope [source.python]
Packages/Python/syntax_test_python.py:508:9: [meta.annotation] does not match scope [source.python]
Packages/Python/syntax_test_python.py:508:6: [meta.qualified-name variable.annotation.function] does not match scope [source.python meta.annotation.python meta.annotation.python meta.qualified-name.python]
Packages/Python/syntax_test_python.py:508:7: [meta.qualified-name variable.annotation.function] does not match scope [source.python invalid.illegal.character.python]
Packages/Python/syntax_test_python.py:508:8: [meta.qualified-name variable.annotation.function] does not match scope [source.python]
Packages/Python/syntax_test_python.py:508:9: [meta.qualified-name variable.annotation.function] does not match scope [source.python]
Packages/Python/syntax_test_python.py:520:10: [invalid.illegal.character] does not match scope [source.python meta.structure.list.python punctuation.section.list.begin.python]
FAILED: 17 of 3525 assertions in 2 files failed
```

Note: I am unsure if I split the commits correctly (and tests are passing after each of them).

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -91,7 +91,7 @@ contexts:
         - meta_scope: meta.statement.import.python
         - include: line-continuation-or-pop
         - include: import-alias
-        - include: dotted-name
+        - include: qualified-name
         - match: ','
           scope: punctuation.separator.import-list.python
         - match: (?=\S)
@@ -127,7 +127,7 @@ contexts:
                   pop: true
             - match: (?=\S)
               pop: true
-        - include: dotted-name
+        - include: qualified-name
         - match: (\.)
           scope: punctuation.accessor.dot.python
         - match: (?=\S)
@@ -257,13 +257,13 @@ contexts:
   line-expressions: # Always include this last!
     - include: expressions-common
     - include: line-continuation
-    - include: dotted-name
+    - include: qualified-name
 
   expressions: # Always include this last!
     # Differs from the line scope in that invalid-name matches will pop the current context
     - include: expressions-common
     - include: illegal-names-pop
-    - include: dotted-name
+    - include: qualified-name
     - match: '(\.) *(?={{identifier}})'
       captures:
         1: punctuation.accessor.dot.python
@@ -444,9 +444,6 @@ contexts:
                   pop: true
             - include: expressions
 
-  path:
-    - match: (?={{path}})
-
   functions:
     - match: '^\s*(?:(async)\s+)?(def)\b'
       captures:
@@ -526,7 +523,7 @@ contexts:
           scope: keyword.other.decorator.python
         - include: line-continuation-or-pop
         - include: function-calls
-        - include: dotted-name
+        - include: qualified-name
 
   item-access:
     - match: '(?={{path}}\s*\[)'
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -519,14 +516,48 @@ contexts:
...
+          scope: punctuation.definition.annotation.python
+        - match: $
+          pop: true
         - include: line-continuation-or-pop
-        - include: function-calls
-        - include: dotted-name
+        - match: (?=\.?\s*{{path}}\s*\() # now we do
+          set: [decorator-function-call-wrapper, qualified-name-until-leaf]
+        - match: (?=\.?\s*{{path}})
+          push: [decorator-wrapper, qualified-name-until-leaf]
```
wbond: So this (and two lines above) is where you are pushing two contexts, the later of which does not consume anything?

FichteFoll: Exactly. The latter may or may not consume something, but due to the bug it always consumes the first character.

FichteFoll:
Note that after this is merged, https://github.com/FichteForks/Packages/tree/python_unpacking-operators can *probably* be merged too. It's been a while since I looked into this and I obviously can't test it myself.

PR #744, opened 2016-12-19
[Scala] Better local symbol indexing

djspiewak:
Makes the local symbol index somewhat prettier and nicer to navigate.  This is similar to what the Python mode does, but adjusted for the fact that Scala has a bazillion more indexed constructs.

I didn't change the global index (again, similar to Python), since that seems to be the convention.  Oh, this also makes a token effort to add some meta scoping to class declarations.  I have more plans in this direction, but one thing at a time.


Scala/syntax_test_scala.scala
```diff
@@ -550,6 +551,7 @@ type Foo = Bar[A] forSome { type A }
 =>     // this is here to act as a random terminator to the above partial syntax
 
    case class Thingy(abc: Int) extends Other
+// ^^^^^^^^^^^^^^^^^ meta.class.identifier.scala
```
wbond: It seems odd that the `case` keyword would be considered part of the `meta.class` scope.

djspiewak: I agree that seems quite odd.  The reason is that `case class` is a very different construct from `class`.  Different in a way that `final class` or `private class` would not be.  With that said, I'm not 100% convinced that this difference is significant enough to merit inclusion in the local symbol list.  So if you feel strongly the other way (i.e. that `case` should be excluded), I can change it.

PR #755, opened 2016-12-23
Change default Markdown file extension from "mdown" to "md"

twhb:
Requesting feedback from Sublime Text and other users.

The vast majority of Markdown files I see use .md; GitHub and BitBucket default to .md; Wikipedia lists .md first. Unless I'm in a bubble, .md seems to be much more common. Changing the default to the more common value will save time typing names for new files, and occasionally prevent the mistake of leaving it at the wrong value.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -3,10 +3,10 @@
 # http://www.sublimetext.com/docs/3/syntax.html
 name: Markdown
 file_extensions:
+  - md
   - mdown
   - markdown
   - markdn
-  - md
 scope: text.html.markdown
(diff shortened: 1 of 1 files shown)


djspiewak:
I almost exclusively use [MarkdownEditing](https://github.com/SublimeText-Markdown/MarkdownEditing), which appears to basically inherit this setting.  I *entirely* exclusively use `.md` as my markdown suffix, and I have literally never seen a single person voluntarily use `.mdown`.  Basically the only files I've ever seen with that suffix were created by a person in a hurry, using Sublime.

wbond:
I think this is a good change, thanks!

PR #756, opened 2016-12-23
[Scala] Various stateful bailout bugs

djspiewak:
Random best practice that's worth documenting somewhere…  When you have any sort of stateful context chain, make sure each link in the chain *ends* with the following:

```yaml
- match: (?=\S)
  pop: true
```

Anything you don't recognize, you should just bail on.

Another best practice: don't accidentally match on double word boundary by having a trailing pipe character.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -399,10 +399,12 @@ contexts:
       push: function-tparams-brackets
     - match: ':'
       set: function-return-type-definition
-    - match: '\n'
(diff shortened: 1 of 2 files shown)


FichteFoll:
> Another best practice: don't accidentally match on double word boundary by having a trailing pipe character.

I too have been guilty of this.

wbond:
I agree that some best practices would be good to add. I've opened #757 where you can write things down. I'll write them up and add to the docs at some point.

PR #761, opened 2016-12-30
[Scala] Fixed bug in val body lookahead

djspiewak:
`val` was using an older technique for operator lookaround, which has now been fixed.


wbond:
Thanks, as always!

PR #762, opened 2017-01-01
[Scala] Fixed incorrect scoping on qualified inheritence

djspiewak:
We were scoping `entity.other.inherited-class` without lookahead, which meant that any qualified identifiers ended up with the *leading* token scoped as the `inherited-class` while any trailing qualifiers (which were the actual class) were unscoped in master and outright broken with my other PR branches.

Leading qualifiers are now scoped as `support.constant`, which is basically accurate.  They may be a package name, but packages and objects in Scala are semantically almost identical, and objects are scoped as `support.constant`.  From the standpoint of the language specification, the only constraint is that the identifier is "stable" (which is to say, basically a constant).

As a random note, the scoping behavior changes when parentheses are introduced.  For example:

```scala
class Foo extends Bar
class Foo extends (Bar)
```

These are identical from Scala's perspective, but they will scope differently.  The reason for this is I will only attempt to apply the `inherited-class` scope at the top level of the type expression (and it is a *general* type expression, which cannot be evaluated without literally becoming scalac).  Most type expressions are restricted in `extends`/`with` clauses without parentheses, but once we're inside of the parentheses, all's fair.  For example:

```scala
class Foo extends (A ~> B)
```

In this case, the inherited class is `~>`!  There's no way we can figure that out in general.  So we just scope anything within parentheses as a type expression, just as we would within square brackets, following a colon or following a `type ... =` expression.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -574,6 +574,8 @@ contexts:
       set: class-inheritance-extends
 
   class-inheritance-extends-token:
+    - match: '{{id}}(?=\s*\.)'
+      scope: support.constant.scala
     - match: '{{id}}'
       scope: entity.other.inherited-class.scala
       set: class-inheritance-extends-token-after
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -575,7 +575,13 @@ contexts:
 
   class-inheritance-extends-token:
     - match: '{{id}}(?=\s*\.)'
-      scope: support.constant.scala
+      scope: entity.other.inherited-class.scala
+    - match: '(?=\.)'
+      push:
+        - meta_scope: entity.other.inherited-class.scala
+        - match: \.
+          scope: punctuation.accessor.scala
+          pop: true
     - match: '{{id}}'
       scope: entity.other.inherited-class.scala
       set: class-inheritance-extends-token-after
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -1441,3 +1441,9 @@ def foo()
 def foo():
    42
 // ^^ constant.numeric.integer.scala
+
+class Foo extends Bar.Baz with bin.Baz
+//                ^^^ support.constant.scala - entity.other
+//                    ^^^ entity.other.inherited-class.scala
+//                             ^^^ support.constant.scala - entity.other
+//                                 ^^^ entity.other.inherited-class.scala
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -1443,7 +1443,7 @@ def foo():
 // ^^ constant.numeric.integer.scala
 
 class Foo extends Bar.Baz with bin.Baz
-//                ^^^ support.constant.scala - entity.other
-//                    ^^^ entity.other.inherited-class.scala
-//                             ^^^ support.constant.scala - entity.other
-//                                 ^^^ entity.other.inherited-class.scala
+//                ^^^^^^^ entity.other.inherited-class.scala
+//                   ^ punctuation.accessor.scala
+//                             ^^^^^^^ entity.other.inherited-class.scala
+//                                ^ punctuation.accessor.scala


wbond:
Currently in C++, PHP, Python, Ruby and JavaScript the complete qualified identifier including namespace/module is scoped with `entity.other.inherited-class`.

In the case of JavaScript, the `punctuation.accessor` is not scoped with `entity.other`, resulting in a fragmented region of `entity.other`. All of the other syntaxes don't have the fragmentation, but do apply `punctuation.accessor` or `punctuation.separator` to the punctuation separating the qualifiers.

In my opinion, the JavaScript case should probably be fixed, and this PR should probably follow the convention of the other syntaxes. Mostly for backwards compatibility since this is a fairly common syntactic element and I think changing it significantly for such a large number of popular languages would be seen negatively by a large part of the community.

djspiewak:
@wbond That's directly inconsistent with how basically every (updated) mode scopes identifiers though.  Most recently, it's very inconsistent with the annotation scoping guidelines.  I think it's a justifiable convention in this case (and I still believe it *would* be justifiable in the annotation case), but it's certainly quite inconsistent.

In any case, I'll update the PR.

wbond:
Oh, I completely agree with you that it is inconsistent, but mostly because of the impact of making such a change.

With things like this I'm not opposed to change out mind in the future, but I think consistency between languages is of most importance since it allows color schemes to be reasonable and code to feel similar between syntaxes. Next on my priority list is not changing (much) constructs in heavy use, especially in mainstream languages. Some of the syntaxes were in really horrible shape and there was fallout from making them semi-reasonable – JavaScript specifically. After those comes internal conceptual consistency of how various scopes are applied for tokens in their respective syntactic contexts. Anyway, that is my current line of thinking.

I mean, we *are* dealing with one of the hardest parts of programming things here – naming things. :smile:

djspiewak:
I think that is a fair perspective, and I agree that cross-language scope consistency is of vital importance.  And frankly, there are much bigger fish to fry.  I feel more strongly about the annotation equivalent, but more generally, the over-use of `storage.type` and the inconsistent (between modes) application of `support` seem like more important windmills to tilt.

wbond:
Yes, I would like to address `storage.type`, `support.type` at some point, but I think for that to work, we may need a shim in the core to automatically add the color for those to whatever new scopes we use.

djspiewak:
Just imagine what would happen if suddenly `def` (and similar constructs) were no longer colored at all in major themes… :-P

Regarding `support`, I was more referring to the fact that some major modes (e.g. Scala) use the "scope all user constructs" approach, whereas other major modes (e.g. C++) barely use `support` at all.  The guidelines are even explicit about how this is effectively allowed to be an inconsistency, and it results in a markedly different experience between the modes.

But anyway, we're drifting a bit afield here.  :-)

wbond:
Yeah, that is what I was referring to about `support` also. I think we need a new scope to use there.

Other things that I think could make the transition easier:

 - Making it simpler/easier to customize color schemes
   - Improved docs
   - Simpler format (YAML)
   - Easily copy existing color scheme
 - Some sort of global versioning of color schemes, where ST could backfill new scopes based on old values from previous versions

But yes, we should continue these conversations on another issue or the forum.

FichteFoll:
>I mean, we are dealing with one of the hardest parts of programming here – naming things. 😄

Add maintaining backwards compatiblity (to a drgree) to that and we're talking.

Regarding a different color scheme format, I do still think that actual CSS (or something based on it, like my CSScheme format) would be easier to edit by hand than something serializable like Plist/YAML, but it does take away from manually automatically it and would make the lives of plugins that modify color schemes much harder.

djspiewak:
@wbond Made the change to scope the entire token with `entity.other`.  From a mode-structure standpoint, I kept the scoping specific to the tokens themselves, rather than setting it as a `meta_scope` over the whole predicted section.  I did this for three reasons: typing experience, avoiding a `clear_scopes` inside of the `(` subcontext, and future-proofing for if/when we decide to change this.  The only bizarre consequence of this was the need to do the lookahead + immediately-consume trick for `.` in order to double-scope it.

keith-hall:
@djspiewak why not just set two scopes at once: ?

    - match: \.
       scope: entity.other.inherited-class.scala punctuation.accessor.scala

like PHP does https://github.com/sublimehq/Packages/blob/77a5d112d5af05feb78f7d5c8759df9e241ce8f9/PHP/PHP%20Source.sublime-syntax#L89

djspiewak:
@keith-hall In my experience, that doesn't work reliably.  Most of the time you end up with only one of the scopes.

keith-hall:
@djspiewak oh, weird. I've not seen that happen. If you can replicate it, it might be worth reporting it on the core issues repo, because it should work.

wbond:
Thanks for the work on this!

PR #766, opened 2017-01-03
[C++] Add support for #include_next

MattDMo:
[`#include_next`](https://gcc.gnu.org/onlinedocs/gcc-6.2.0/cpp/Wrapper-Headers.html) is not often used, but while chasing down a compiler error I found it in GCC 6.2.0 (and going back to at least GCC 4, according to Stack Overflow), I found it is used in a couple of places in the C++ standard library - `cmath` and `cstdlib`, to be specific. 

I made a very simple modification to `C++/C++.sublime-syntax` to add support for `#include_next`. I also added tests for it, as well as `#include` (there were none, for some reason) in `C++/syntax_test_cpp.cpp` - they pass just fine.


wbond:
It seems this is applicable to C in addition to C++, right? https://gcc.gnu.org/onlinedocs/gcc-4.0.3/cpp/Wrapper-Headers.html

In addition to adding this to  C and C++, it will also need to be added to Objective-C and Objective-C++ and the tests for those.

MattDMo:
I just updated the C, Obj-C, and Obj-C++ syntaxes and tests. Everything should be ready to go now.

wbond:
Thanks!

PR #768, opened 2017-01-03
[Scala] Misc Fixes

djspiewak:
Started using sublime on a different sort of Scala codebase and ran into a surprising number of small issues.  Whacked them all.  Mostly.

@wbond I could use some help here.  Specifically, please look at `bf034bf` in this PR.  The test case hopefully makes it clear what scenario is being hit here, but basically the `for-braces-body` rule is doing lookahead from start-of-line to try to determine whether or not the `<-`, `←` or `=` tokens exist preceded only by something which *could* be a Scala pattern match.  Note that Scala pattern match syntax is context-free (it may have arbitrary nesting), so this is only an approximation, but I've never seen that part of it fail.

The problem here is that the token in question may be commented out, either because the line itself is commented out, or because it's part of a larger comment block.  Detecting this in lookahead appears to be non-trivial.  The commit I referenced attempts to fake it by looking for a line comment token `//` *or* a block comment start token `/*` prior to the token we're hunting in the lookahead (the arrow/equals).  If it sees those, it bounces out of the lookahead and assumes there's no generator.  This approximation has some issues though, such as incorrectly detecting a block comment which is fully contained *within* a pattern match to the left of a generator on one line (which is very much valid in Scala, and probably not an inconceivable use of syntax).

Anyway, I'm not sure exactly what would be a better way to go about this.  What I've done probably covers 95-75% of all valid Scala for-comprehensions involving comments, but the remaining ones are going to look a little strange.


FichteFoll:
What you ended up facing is the requirement of basically having to implement a fail-safe parser for an entire line in order to try not to screw up the context-free grammar by over-simplification. This *will* be hard, but there are ways to remain accurate in the huge majority of cases.

I only took a brief a look into the regular expression, but what you can do in cases like this is to match strings separately, e.g. `^(?=(\s+|"[^"]"|/\*([^*]|\*[^/])\*/|//.*$)(<-|\x{2190}|=))`. This does not consider `{{nonopchar}}` because I didn't look into what that variable represents. Using a multi-line match would greatly increase readability.

djspiewak:
@FichteFoll Yeah, multi-lining the regexp will help from an organization standpoint.  As for the tokens I want to match, it's literally anything which is accepted by the `pattern-match` context, followed by one of the generator tokens (the arrow/equals thing) such that the generator token is *not* followed by a valid operator character (this is something which is currently incorrectly handled) *and* such that the generator is not scoped as `comment` by some other rule.

FichteFoll:
Correction: `/\*([^*]|\*(?^/))\*/` is better. Just encountered this issue in my reviewer with some `.sublime-theme` file that used an even number of `*` before `/`: https://github.com/packagecontrol/package_reviewer/commit/9498ade4997ab0504b9a22fc865d4957ab208d24

djspiewak:
@FichteFoll LOL  Regular expressions are great…

Thanks for the review.  I'll incorporate your comments.

wbond:
@djspiewak Can you confirm my fix in 4577dfb971cce783ec8a9168c6e19ecd52859364 is correct?

djspiewak:
@wbond I had to do something equivalent in my local integration branch.  :+1:

PR #770, opened 2017-01-07
[C++] Fix none highlighted preprocessor definitions in extern C blocks.

deathaxe:
While reading some C++ headers (_cffi_include.h) I recently found some blocks
not being highlighted at all. Everything from the second `__cplusplus` to the
the final `}` keeps unhighlighted.

```C++
#include <Python.h>
#ifdef __cplusplus
extern "C" {
#endif

#ifdef __cplusplus       // THE LAST HIGHLIGHTED LINE
# ifndef _Bool
   typedef bool _Bool;   /* semi-hackish: C++ has no _Bool; bool is builtin */
# endif
#endif

#ifdef __cplusplus
}                        // HIGHLIGHTED AGAIN
#endif
```

With `#ifdef __cplusplus` within an `extern "C" {` block an empty scope is set
to stack causing this issue. This issue is fixed by removing this block at all.

So the `global` and `global-preprocessor` scopes are used within external C blocks
for highlighting.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -184,6 +184,16 @@ contexts:
           scope: punctuation.section.block.begin.c++
           set:
             - meta_scope: meta.extern-c.c++
+            - match: '^\s*(#\s*ifdef)\s*__cplusplus\s*'
+              scope: meta.preprocessor.c++
+              captures:
+                1: keyword.control.import.c++
+              set:
+                - match: '\}'
+                  scope: punctuation.section.block.end.c++
+                  pop: true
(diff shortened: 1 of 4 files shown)


wbond:
Related: 0e7352481ca760c8d8939f2962dcd3dde94333b2

deathaxe:
So lets try the other way round by adding some scopes to this block.

wbond:
Could you add these tweaks and tests to the Objective-C++ syntax also?

deathaxe:
Done.

wbond:
Thanks!

Issue #751, opened 2016-12-22
[CSS] highlighted known font-family part of unknown


stevenvachon:
```css
:root {
  --body-font: Raleway, Arial, Helvetica Neue, Helvetica, sans-serif;
}
```

produces:

![screenshot](https://cloud.githubusercontent.com/assets/170197/21429427/aacfcb96-c82c-11e6-8b03-6a5955212144.png)

It probably shouldn't highlight "Helvetica" in "Helvetica Neue".

MattDMo:
I just submitted PR #771 that will hopefully address this.

stevenvachon:
Thanks!

wbond:
This was fixed with 3889492cac950cb90b98db53d94ef47a6b0bd5fe

PR #771, opened 2017-01-09
[CSS] Fixes incorrect font highlighting in #751

MattDMo:
This addresses #751 - `Helvetica Neue` was getting highlighted as a generic font family while it shouldn't have been. I simply added a positive lookahead to ensure that after the indicated font family name is matched, it should be followed either by a comma `,` indicating another name is coming, or a semi-colon `;` indicating that the property's values are at an end.

If this isn't the best way to do this, I'm all ears.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -567,7 +567,7 @@ contexts:
             )\b
       scope: support.constant.property-value.css
       # Generic Font Families: https://www.w3.org/TR/CSS2/fonts.html
-    - match: \b(?i:arial|century|comic|courier|garamond|georgia|helvetica|impact|lucida|symbol|system|tahoma|times|trebuchet|utopia|verdana|webdings|sans-serif|serif|monospace|fantasy|cursive)\b
+    - match: \b((?i:arial|century|comic|courier|garamond|georgia|helvetica|impact|lucida|symbol|system|tahoma|times|trebuchet|utopia|verdana|webdings|sans-serif|serif|monospace|fantasy|cursive)(?=[;,]))\b
       scope: support.constant.font-name.css
 
   property-values:
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -567,7 +567,7 @@ contexts:
             )\b
       scope: support.constant.property-value.css
       # Generic Font Families: https://www.w3.org/TR/CSS2/fonts.html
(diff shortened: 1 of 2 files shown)


CSS/CSS.sublime-syntax
```diff
@@ -567,7 +567,7 @@ contexts:
             )\b
       scope: support.constant.property-value.css
       # Generic Font Families: https://www.w3.org/TR/CSS2/fonts.html
-    - match: \b(?i:arial|century|comic|courier|garamond|georgia|helvetica|impact|lucida|symbol|system|tahoma|times|trebuchet|utopia|verdana|webdings|sans-serif|serif|monospace|fantasy|cursive)\b
+    - match: \b(?i:arial|century|comic|courier|garamond|georgia|helvetica|impact|lucida|symbol|system|tahoma|times|trebuchet|utopia|verdana|webdings|sans-serif|serif|monospace|fantasy|cursive)(?=[;,])\b
```
wbond: We’ll want the lookahead to allow whitespace, and be followed by a `,`, `;`, `\n`, `}` or `)`. This will allow things like:

```css
.myclass { font-family: Helvetica Neue }
.myclass2 { font-family: Helvetica , Arial ; }
.myclass 3 {
    font-family: Arial
}
.myclass4 { font-family: var(—base-font, Arial) }
```

CSS/syntax_test_css.css
```diff
@@ -822,3 +822,9 @@
 /*^^^^^^^ meta.property-list */
 /*  ^^^ meta.property-name */
 }/*     ^ meta.property-value */
+
+:root {
+    --body-font: Raleway, Arial, Helvetica Neue, Helvetica, sans-serif;
+/*                               ^ -support.constant.font-name.css */
+/*                                               ^ support.constant.font-name.css */
+}
```
wbond: Some tests should be added for variations other than just `,`.

jrappen:
@MattDMo maybe you could add the requested changes here?

FichteFoll:
I would actually remove all fonts from this match and only match the generic font names `serif`, `sans-serif`, `cursive`, `fantasy` and `monospace`, as they are listed [here](https://www.w3.org/TR/css-fonts-3/#generic-font-families). All other font names should be generic strings.

wbond:
This should all be resolved by 3889492cac950cb90b98db53d94ef47a6b0bd5fe

PR #782, opened 2017-01-19
[Objective-C] Add support for Objective-C specific format specifier

### LINKED ISSUE #781: [Objective-C] Objective-C specific format specifier is not recognized
Along with the C/C++ format specifiers in format strings, Objective-C/C++ also supports the format specifier `%@` to indicate that the parameter is an object which should have its `description` method called to provide the text to insert. The specifier is only available in "object" strings (i.e. `@"Constant String Object"` but not `"Constant String"`).

More information is available in this [Apple article](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html) from the String Programming Guide.

The syntax in Sublime 3126 does not support this additional format code, which causes it to highlight as an error. From looking at the syntax files it appears that this is because the Objective-C syntax is using the `string_placeholder` scope directly from the C syntax file, which does not support the option.

A simple example of this is:
```objective-c
int main(int argc, char const *argv[])
{
    NSLog (@"This is a %@", @"Test");
    
    return 0;
}
```

This is rendered in Sublime as the following. This tends to draw your eye as you're editing code, even though it's not actually an error.

```c
int main(int argc, char const *argv[])
{
    NSLog (@"This is a %@", @"Test");
    
    return 0;
}
```

OdatNurd:
This request fixes #781 for Objective-C (and Objective-C++ by extension) by including an extra match for the relevant specifier while processing an Objective-C specific string.

I included two extra syntax tests in order to verify that this works only inside of a constant object string and not in a standard C string.

### DIFF
--- a/Objective-C/Objective-C.sublime-syntax
+++ b/Objective-C/Objective-C.sublime-syntax
@@ -110,6 +110,8 @@ contexts:
           scope: punctuation.definition.string.end.objc
           pop: true
         - include: scope:source.c#string_escaped_char
+        - match: '%@'
+          scope: constant.other.placeholder.objc
         - include: scope:source.c#string_placeholder
 
   unique-keywords:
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #786, opened 2017-01-20
[C#] fix IDisposable using  statements

### LINKED ISSUE #753: [C#] IDisposable `using` statements are (mis)interpreted as imports
These are roughly equivalent to `with` statements in python.

```cs
using(var reader = SomeCodeThatGetsAnIDisposable())
{
    for(var line in reader)
    {
        DoStuff(line);
    }
}
```

keith-hall:
fixes #753


wbond:
Thanks!

PR #788, opened 2017-01-24
[C#] fix lambdas breaking try/catch

### LINKED ISSUE #754: [C#] lambdas break try/catch highlighting
public class Test {
        Func<int, int> a;
        void TestMe () {
            a = (progress) => progress;

            try
            {
        
            }
            catch (InvalidOperationException ex)
            {
        
            }
            finally {
                
            }
        }
    }

`catch` and `finally` are scoped as `invalid.illegal`, but they are perfectly valid here.

keith-hall:
fixes #754

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -927,6 +927,8 @@ contexts:
         2: storage.type.function.lambda.cs
       push:
         - meta_scope: meta.function.anonymous.cs
+        - match: '(?=;)'
+          pop: true
         - include: line_of_code_in
     - match: \((?=[^)]*\)\s+=>)
       scope: meta.function.anonymous.cs meta.group.cs punctuation.section.group.begin.cs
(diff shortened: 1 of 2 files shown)


C#/C#.sublime-syntax
```diff
@@ -1541,7 +1546,7 @@ contexts:
 
   finally_block:
     - meta_scope: meta.block.trycatch.cs
-    - match: (finally)\b\s*
+    - match: finally\b
```
FichteFoll: Are you sure removing the `\b` anchor doesn't introduce a regression? I didn't check how the context flow goes.

keith-hall: I didn't remove it :) I only removed `\s*`, which meant that whitespace after `finally` was getting scoped as `keyword.control.trycatch.finally.cs`

FichteFoll: ... right, ignore me

wbond:
Thanks!

PR #798, opened 2017-02-05
[Rust] Patch for catastrophic backtracking

trishume:
Non-Sublime implementations of sublime-syntax that use Oniguruma take exponential time to parse `impl SomeIdentifier {` lines without this change. See https://github.com/trishume/syntect/issues/33.

I'm pretty sure syntect is the only non-Sublime project that uses these files so I understand if you'd prefer not to merge this PR. I'm just submitting this to ask if you are open to it. I notice some other syntaxes use possessive quantifiers for speed, but those are probably left over from before sregex.

Otherwise I'll probably have to add hacky patches to the version of sublimehq/Packages I bundle with syntect.

@wbond

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -624,7 +624,8 @@ contexts:
       pop: true
 
   impl-for:
-    - match: '(?=\s*(?:::|{{identifier}}|\$|<)+(<.*?>)?\s+for\s+)'
+    # inline {{identifier}} and use possesive quantifiers to avoid catastrophic backtracking in non-Sublime implementations
+    - match: '(?=\s*(?:::|[[:alpha:]][_[:alnum:]]*+|_[_[:alnum:]]++|\$|<)++(<.*?>)?\s+for\s+)'
       set:
         - meta_scope: meta.impl.rust
         - include: comments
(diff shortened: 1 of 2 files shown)


wbond:
Since sregex doesn't support possessive quantifiers, this would lead to a performance regression since it would force us to fall back to oniguruma. I'm not at a machine right now, but I'll look to see if it can be rewritten another way to avoid the issue.

trishume:
@wbond that should be fixed, there's also a syntax test now. @keith-hall figured it out.
I also rebased and fixed up the commit messages.

Rust/Rust.sublime-syntax
```diff
@@ -6,7 +6,7 @@ file_extensions:
   - rs
 scope: source.rust
 variables:
-  identifier: '(?:[[:alpha:]][_[:alnum:]]*|_[_[:alnum:]]+)'
+  identifier: '(?:(?:[[:alpha:]][_[:alnum:]]*|_[_[:alnum:]]+)(?![_[:alnum:]]))' # include a negative lookahead at the end to ensure all possible characters are consumed, to prevent catastrophic backtracking
```
michaelblyons: Is there a reason this is better than good old `\b`?

keith-hall: Good point, I just somehow didn't think of it, because I'm not used to the `[[:alpha:]]` character class ;)

PR #799, opened 2017-02-05
[HTML] Complete with simple tag: use $1 instead of $0

math2001:
Hi!

No need for a long message, here is the only change:

```diff
-    return (tag + '\tTag', tag + '>$0</' + tag + '>')
+    return (tag + '\tTag', tag + '>$1</' + tag + '>')
                                     ^
```

It's not much, but it keeps the completion of `kbd` consistent with the completion of `span.myclass`. (the last field is always at the end)

Matt

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -12,7 +12,7 @@ def match(rex, str):
 def make_completion(tag):
     # make it look like
     # ("table\tTag", "table>$1</table>"),
-    return (tag + '\tTag', tag + '>$0</' + tag + '>')
+    return (tag + '\tTag', tag + '>$1</' + tag + '>')
 
 def get_tag_to_attributes():
     """
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -448,8 +448,8 @@ class Unittest(unittest.TestCase):
         completion_list, flags = completor.get_completions(view, 'tab', [0], False)
 
         # should give a bunch of tags that starts with t
-        self.assertEqual(completion_list[0], ('table\tTag', '<table>$0</table>'))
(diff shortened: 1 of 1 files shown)


keith-hall:
You may need to update some of these unit tests to ensure they still pass: https://github.com/math2001/Packages/blob/53da5558b9937de8063c51ebc5cbd7e7fb2694a8/HTML/html_completions.py#L398

Also a small preference request: please could you include the name of the language(s) affected by the commit in the commit title - it makes it easier to understand which packages have been affected when looking at the commit history i.e. `[HTML]  Complete with simple tag: use $1 instead of $0` - thanks :)

math2001:
Oh, sorry... note sure I can `--amend` this commit, since I've created on github... Working on the tests.

FichteFoll:
Editing files on github creates normal commits on a selected branch. You can continue to work with these commits like any other (amend, squash, rebase, revert, etc.).

You can't amend from within github though, so you'll want to use the command line or a GUI client for that.

jrappen:
@math2001

Regarding your commit messages, you can reference commits by using `@` instead of `#`.

Compare:

* @53da555
* or sublimehq/Packages@53da555

wbond:
For consistency shouldn't all of the completions from line 281 to 301 use something other than `$0` as the completion between the beginning and end tags?

wbond:
After looking at this more, I'm not convinced this is the correct change. If anything, I am thinking `span.myclass` should complete to `<span class="myclass">$0</span>`.

However, it is probably not worth throwing off people's muscle memory for this type of thing. For all I know this was implemented this exact way to keep completions working the same as before, but then also to mirror Emmet's behavior with `span.myclass`.

wbond:
I don't think this is actionable, so I am going to close rather than let it linger open

Issue #785, opened 2017-01-20
[All - tmPreferences] RFC: wrap regular expressions in CDATA?


keith-hall:
Hi,

I've recently been working on a syntax definition for Plist files, so that scope selectors and regular expressions will be highlighted, making it much easier to work with indentation rules etc. (I'm hoping it would be gratefully received in the PackageDev package, if not here, on this repo.)

However, some `.tmPreferences` contain regular expressions with lookbehinds (for example), but the `<` is escaped for XML to `&lt;`, so it appears as an invalid regex. Rather than try to build the syntax definition to work with character escapes, I thought perhaps we could change the `tmPreferences` files in this repo to wrap the expressions in XML `<![CDATA[`...`]]>` segments where appropriate instead, to avoid the need to escape these characters. I'm happy to make the changes myself and submit a PR, but I wanted to check what people think about it, first.

![regex highlighting in tmpreferences](https://cloud.githubusercontent.com/assets/11882719/22142752/2091241c-df00-11e6-8457-45c4d59ee6d2.png)

I've tested and can confirm that ST correctly parses all CDATA segments inside Plist files, as does the Python `plistlib` module, so making these changes shouldn't adversely affect anything. Arguably, even without the syntax highlighting I am working on, the regexs would be easier to read without the XML character/entity escapes.

If people are not opposed to this idea, how much do you think we should wrap in CDATA segments? Should we wrap all indent patterns, for consistency? or only those that use the XML entities? or not just indent patterns, but everything with XML entities like `shellVariables`, `symbolTransformation`s and `scope` selectors? (I'm also working on scoping/highlighting symbol transformations / regex replacements, so the answer I am hoping for would be either "wrap everything" or "wrap everything that currently contains an XML entity [that would affect the highlighting] [and we'll try to wrap things in future when we introduce something that would otherwise need escaping]" 😉 )
Wrapping everything may simplify "Find in Files" searches, where one wants to find every location where a particular token is used, and not have to remember to search the escaped XML version of the token too. :)

Thanks in advance for your feedback! :)

keith-hall:
To be concrete, what I'm proposing is that this: https://github.com/sublimehq/Packages/blob/458e0764063a885cf8f7c2da3b41ac3d81fdfa57/Ruby/Miscellaneous.tmPreferences#L13-L46

would become:

    <string><![CDATA[(?x)^
        (\s*
            (
                module|class|def
                |unless|if|else|elsif
                |case|when
[...]

Issue #772, opened 2017-01-10
[XML] Question: does ST support highlightPairs in tmPreferences?


keith-hall:
I notice that the [XML tmPreferences file contains `highlightPairs`](https://github.com/sublimehq/Packages/blob/c55cf5662bb00020acdd5af32cf774d0217ec940/XML/Miscellaneous.tmPreferences#L35-L57), but in my experience they don't seem to do anything in ST.

If ST doesn't support `highlightPairs`, should we remove this part of the file to avoid confusion? It seems like ST instead just respects the configuration options `match_brackets`, `match_brackets_square`, `match_brackets_braces` and `match_brackets_angle`, such that matching the double quotes (`"`) isn't possible without a Python plugin anyway.

wbond:
Just for the record, I see no reference to `highlightPairs` in the Sublime Text 3 codebase.

PR #801, opened 2017-02-07
[All:tmPreferences] wrap strings in CDATA if they use XML character entities

keith-hall:
This addresses #785, by wrapping every `<string>` that currently contains XML entities in "CDATA" for all the `.tmPreferences` files.
This also addresses #772, and removes an unused `tmPreferences` file from the XML package, to clean it up a bit.


wbond:
This should help readability for humans too. Thanks!

PR #802, opened 2017-02-09
[Ruby] fix a highlight bug

hyrious:
bug before: 
default

### DIFF
--- a/Ruby/syntax_test_ruby.rb
+++ b/Ruby/syntax_test_ruby.rb
@@ -174,7 +174,7 @@ class ::MyModule::MyClass < MyModule::InheritedClass
 #                             ^ punctuation.definition.group.end
   end
 
-  def self.my_second_method *arg_without_parens
+  def self.my_second_method *arg_without_parens # comment.line.number-sign.ruby
 # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.function
 #     ^^^^ variable.language
 #         ^ punctuation.accessor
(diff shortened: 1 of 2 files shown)


Ruby/syntax_test_ruby.rb
```diff
@@ -174,7 +174,7 @@ def my_method(param1, param2)
 #                             ^ punctuation.definition.group.end
   end
 
-  def self.my_second_method *arg_without_parens
+  def self.my_second_method *arg_without_parens # comment.line.number-sign.ruby
```
wbond: Can you add some assertions here to make sure the comment you added is actually a comment? Right now, your additions don't actually check anything.

http://www.sublimetext.com/docs/3/syntax.html#testing has info about how to do that.

hyrious:
**Will it be merged? Actually it's a tiny patch.**
Just now I've found another ruby highlight issue that may be mentioned before. (But I'm not going to solve it here.)
Well we could know when it comes to highlighting ruby codes without parentheses, it usually will crash into a mess.
A practical solution is using Ripper(from Ruby's stdlib since 1.9.*), which allows to parse Ruby code and analyze their parse trees, and absolutely it can be used to mark highlight elements (to see how, just try this ruby code:
```ruby
require 'ripper'
require 'pp'
pp Ripper.lex('def meth arg, without, parentheses # bad highlight')
```
). The only problem is that it may be slower than *.sublime-syntax and difficult to be merged into subl.
Hope for your response.

FichteFoll:
Unfortunately I have to crush your hopes and say that this is not going to happen. The current syntax highlighting system is very optimized and cannot be interacted with from the outside.

This isn't official because I'm not affilicated with the project, but I can assure you that ST will not start to run something from Rubys stdlib in order to correctly highlight code – at least not within the next 5 years.

wbond:
Thanks!

Issue #266, opened 2016-03-30
[Markdown] A collection of issues


aziz:
- [x] List punctuations #264 
- [x] Underscore In the middle of words

``` md
The word `complicated` must be neither bold nor italic below:

perform_complicated_task
perform__complicated__task

But the first part below is italic and bold respectively:

_perform_complicated_task
__perform__complicated__task
```
- [x] Plain URLs

``` md
Normal plain link http://example.com
Complex url https://github.com/sublime/issues?page=1&state=closed
```
- [x] Links Refs with optional titles wrapped in single quotes 

``` md
[3]: http://example.com/  'Optional Title Here'
```
- ~~Escaped backticks inside inline code blocks~~ (seems invalid @wbond)

``` md
`raw \` more`
```
- [x] Ordered lists and unordered lists inside a block quotes

``` md
> Here is a quote block
> - list items
> - inside block
> - quote
> 
> 1. numbered list 
> 2. numbered list another
>
```
- [x] Detecting empty lines inside a block quote with just `>` symbol

``` md
> Here is a quote block
This quote continues on.  Line breaking is OK in markdown
> Here it is again
>
> The `>` in line above should be marked as punctuation but it's not
```
- [x] Lists immediately after paragraph

``` md
this is a paragraph
- and this
- is a list
```
- [x] Blockquote and code blocks inside a list

``` md
* this is a list

   > This is a blockquote.

 This is a paragraph still part of the 
 list item

* Lorem ipsum

        This is a code block  
* list continues
* list continues
```
### Extended _but common_ Markdown
- [x] Fenced Blocks

`````` md
    ```js
      for (var i = 0; i < 10; i++) {
         console.log(i);
      }
    ```
``````

[Here is test file for fenced blocks](https://github.com/aziz/PlainNotes/blob/master/Tests/md-test-fenced.md)
- [ ] Syntax highlighting inside fenced blocks
- [x] Tables
  [Here is a test file for tables](https://github.com/aziz/PlainNotes/blob/master/Tests/md-test-tables.md)
- [ ] Strikethrough

``` md
~~This text should be parsed as _strikethroughed_.~~

~~There may be __bold__ or _italic_ text inside strikethroughed text.~~

~~There may be ![Mou icon](mo.com) a keyboard `complicated` shortcut like [GFM][GFM] inside strikethroughed text.~~

__There may be ~~strikethroughed text~~ inside bold text.__
_There may be ~~strikethroughed text~~ inside italic text._

[...]

PR #803, opened 2017-02-09
[Markdown] fix list punctuation

### LINKED ISSUE #264: [Markdown] list punctuation
ST build 3019

List items in Markdown only appear to have the punctuation scope defined for the first item in a list.

e.g.

```
# Title

- unordered list item
- another list item
  - subitem

1. numbered list item
2. another one
```

markdown list punctuation

I would expect all the `-` and `.` characters to be highlighted in this example.

scope at the first `-` is: `text.html.markdown markup.list.unnumbered.markdown punctuation.definition.list_item.markdown`
scope at the second `-` is: `text.html.markdown markup.list.unnumbered.markdown meta.paragraph.list.markdown`

ideally, some syntax tests would be created as part of this package.

keith-hall:
Before this change, list punctuation was only scoped for the first item in the list. Now, all list punctuation will be correctly scoped.
Fixes #264


Markdown/Markdown.sublime-syntax
```diff
@@ -8,14 +8,25 @@ file_extensions:
   - markdown
   - markdn
 scope: text.html.markdown
+variables:
+    separator_line: |-
+        (?x)
```
FichteFoll: I suggest always wrapping variables in non-capturing groups and multi-line expressions in `(?x:...)` groups (unless they represent a set). That way you can include them anywhere without having to worry about the including pattern being in extended more or not.

For example, in `contexts.main[1].match` you are already using extended mode, but if you were not the result could be different and unexpected, since extended mode would be active after you included this variable.
However, in `context.separator[0].match` you do not initially use extended mode, but if you were to use any spaces after the inclusion of this variable it would be ignored.

The reason why I wrap everything in a group is because it allows to treat variables atomicly. You can add quantifiers after variables and they mean exactly what you think they would (what the variable matches can be repeated) instead of potentially only repeating the last token of the variable's contents. For example, `{{block_quote}}+` (which makes no sense, but for the example's sake) would expand to `[ ]{,3}.+` and not `(?:[ ]{,3}>.)+`.

keith-hall: that's great advice, thanks @FichteFoll - I'll update my variables :)

keith-hall:
I've fixed a few bugs mentioned at https://github.com/sublimehq/Packages/issues/266, added more tests, and started reworking the regex patterns that cause ST to use Oniguruma (https://github.com/sublimehq/Packages/issues/481). I think further improvements should be part of another PR now, as they start to get more complicated (the "italic" and "bold" regexes look intimidating, plus the lookbehinds and `\G`s might be more involved to remove).

wbond:
Awesome!

jrappen:
Thank you!

PR #805, opened 2017-02-11
[HTML] Support "text/paperscript" script tags

lehni:
[PaperScript](http://paperjs.org/tutorials/getting-started/working-with-paper-js/#what-is-paperscript) is a slightly enriched JavaScript, adding support for operator overloading and a few other features developed for http://paperjs.org/

Standard JavaScript syntax highlighting can be used for PaperScript.

[See here](https://www.reddit.com/r/webdev/comments/4o8tz6/paperscript_syntax_highlighting_within_html_file/d4akcej/) for the origin of the approach to add support for inlined PaperScript in Sublime Text 3.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -85,7 +85,7 @@ contexts:
           with_prototype:
             - match: (?i)(?=</style)
               pop: true
-    - match: '(<)((?i:script))\b(?![^>]*/>)(?![^>]*(?i:type.?=.?text/((?!javascript).*)))'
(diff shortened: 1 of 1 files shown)


wbond:
I don't think paperscript is mainstream enough to justify adding it to the default packages and maintaining it. Sorry!

lehni:
I understand that its use is not widespread enough for you, but there would be 0 maintenance involved because it is syntactically identical. 

What would be nice is a way to get Sublime Text apply the same syntax coloring and other editing magic to any nested scripts for which the editor has already registered a specific type through the bottom right pulldown menu.

PR #806, opened 2017-02-12
[Markdown] support lists in blockquotes, code fences, multiline inline code and multiline emphasis

keith-hall:
This PR simplifies a few contexts, removes a few incompatible regexs (mainly `\G`) and adds support for lists in blockquotes, plus includes a few more tests.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -8,6 +8,7 @@ file_extensions:
   - markdown
   - markdn
 scope: text.html.markdown
+comment: this definition aims to (eventually) meet CommonMark specifications http://spec.commonmark.org/
 variables:
     separator_line: |-
         (?x:
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -113,12 +113,11 @@ contexts:
             \s*                        # Leading whitespace
             (\[)(.+?)(\])(:)           # Reference name
             [ \t]*                     # Optional whitespace
-            (<?)(\S+?)(>?)             # The url
-            {{link_title}}             # Optional link title
-            \s*                        # Optional whitespace
-            $
+            (?:
+              (<)([^>]+)(>)            # The url
+            | (\S+)                    # The url
+            )
         )
-      scope: meta.link.reference.def.markdown
       captures:
         1: punctuation.definition.constant.begin.markdown
         2: constant.other.reference.link.markdown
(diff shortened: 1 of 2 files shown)


keith-hall:
I've got this to the state where the only incompatible regex patterns are two `\G`s :)
Note that I have scoped `.begin` and `.end` which could affect color schemes - I can remove these changes if needed.

Markdown/Markdown.sublime-syntax
```diff
@@ -643,13 +643,18 @@ contexts:
...
     - include: scope:text.html.basic
   raw:
-    - match: '(`+)'
-      scope: punctuation.definition.raw.begin.markdown
-      push:
-        - meta_scope: markup.raw.inline.markdown
-        - match: '\1'
-          scope: punctuation.definition.raw.end.markdown
-          pop: true
+    - match: '{{backticks}}'
```
FichteFoll: I suggest this instead:

```yaml
  raw:
    - match: '(`+)'
      scope: punctuation.definition.raw.begin.markdown
      push:
        - meta_scope: markup.raw.inline.markdown
        - match: '\1'
          scope: punctuation.definition.raw.end.markdown
          pop: true
        - include: raw
```


Should probably also exit on empty lines. See:

test `test
 
test`

FichteFoll: Actually, I just realized that this is problematic in case a backtick is used but a raw block is not intended. However, I do believe that it's better to call people out on their backticks and require them to escape them (`` \` ``) than not supporting multi-line inline raw segments.

keith-hall:
I've done some testing at http://spec.commonmark.org/dingus/ and made the backtick handling more consistent with how that works. This includes support for code fences (though it doesn't highlight the language contained inside the fence yet, I'll leave that for another PR, as I wonder if it would give us more `with_prototype` recursion limit warnings).

Markdown/Markdown.sublime-syntax
keith-hall: @FichteFoll thanks, I have just pushed a new commit with your proposal :)

keith-hall:
Performance comparison:

This PR:
> Syntax "Packages/Markdown/Markdown.sublime-syntax" took an average of 5.0ms over 10 runs

Before this PR:
> Syntax "Packages/Markdown/Markdown.sublime-syntax" took an average of 7.2ms over 10 runs

keith-hall:
Performance update & comparisons against some Markdown syntaxes on Package Control, for the syntax test file in this PR, which is now 40 KiB:

> Syntax "Packages/Markdown/Markdown.sublime-syntax" took an average of 10.4ms over 10 runs (this PR)

> Syntax "Packages/Markdown/Markdown.sublime-syntax" took an average of 11.2ms over 10 runs (before this PR - I guess the smaller difference now is due to all the added constructs it lexes in this PR)

> Syntax "Packages/Markdown Extended/Syntaxes/Markdown Extended.sublime-syntax" took an average of 12.3ms over 10 runs

> Syntax "Packages/MarkdownHighlighting/Markdown.sublime-syntax" took an average of 12.8ms over 10 runs

> Syntax "Packages/MarkdownEditing/Markdown (Standard).tmLanguage" took an average of 11.3ms over 10 runs

> Syntax "Packages/MarkdownEditing/Markdown.tmLanguage" took an average of 12.9ms over 10 runs

wbond:
Thank you very much for all of this work!

PR #808, opened 2017-02-14
[C] Add variable type: WPARAM

jfcherng:
Just find that currently `WPARAM` is not highlighted but `LPARAM` is.

image


wbond:
Thanks!

Issue #497, opened 2016-06-23
Bad Haskell syntax colors


csoroz:
The default syntax for Haskell shows wrong colors in some situations. Seems to be related with the pattern guards '|' in function definitions: 
[test.zip](https://github.com/sublimehq/Packages/files/329018/test.zip)

-- test.hs

```
max3 :: Int -> Int -> Int -> Int
max3 x y z                      -- wrong colors
    | (x <= z) && (y <= z) = z  -- wrong colors
    | (x <= y) && (z <= y) = y
    | (y <= x) && (z <= x) = z

test :: Int -> Int
test x               -- wrong colors
  | x < 0 = 0        -- wrong colors
  | otherwise = x

test :: Int -> Int
test x =  -- inserting a '=' fix colors
  | x < 0 = 0
  | otherwise = x
```

wbond:
This issue has been solved by a previous commit, hence it is being closed.

Issue #214, opened 2016-02-26
[Haskell] "&" does not get highlighted 


AllanZyne:
![d8476d38-f973-455c-8c73-dc79f7552718](https://cloud.githubusercontent.com/assets/3166380/13348208/13507fd8-dcad-11e5-9b25-d7b10a839a7c.png)

and 

![rtx](https://cloud.githubusercontent.com/assets/3166380/13348489/c03a68a2-dcae-11e5-9aff-9c8dca693c16.png)

scope `meta.function.type-declaration.haskell` is wrong

DrsExplorer:
I think we may match operactors in this way: ``[^\w\s_\(\)\[\]`"',:{}]+``

wbond:
Is the `&` considered part of the same token as `==>`?

For example 2, what scope should that token receive?

DrsExplorer:
Yes, I use code from this [tutorial](http://book.realworldhaskell.org/read/code-case-study-parsing-a-binary-data-format.html#id627853).

In example 2, I think that token could be "function parameter", but it can't be in "type-declaration".

jrappen:
@AllanZyne The Haskell syntax is currently being rewritten in #2225. Maybe test that and give feedback there?

PR #810, opened 2017-02-15
[Haskell] Unicode Name

AllanZyne:
I try to fix the issues at #214 and #497.
It's not perfect, but better than current one.

### DIFF
--- a/Haskell/Haskell.sublime-syntax
+++ b/Haskell/Haskell.sublime-syntax
@@ -6,22 +6,23 @@ file_extensions:
   - hs
 scope: source.haskell
 
+# Based on atom language-haskell
 variables:
-  identStartCharClass: '[\p{Ll}_\p{Lu}\p{Lt}]'
-  identContCharClass: '[\p{Ll}_\p{Lu}\p{Lt}'']'
-  identCharClass: '[\p{Ll}_\p{Lu}\p{Lt}\p{Nd}'']'
-  classNameOne: '[\p{Lu}\p{Lt}]{{identCharClass}}*'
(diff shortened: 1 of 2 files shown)


wbond:
Thanks for taking the time to contribute!

Could you add some tests for your changes? This way we can ensure that future changes don't undo work that you did.

From a style standpoint, we use `under_score` variable names instead of `camelCase`. Could you change those over?

wbond:
So it seems these changes remove a bunch of test assertions that were present before. Were those test assertions incorrect in some way?

AllanZyne:
Yes. I test it on Atom, these codes are syntax error!
(Sorry for my poor English!)

qq 20170216100339

wbond:
Just because Atom doesn't syntax highlight them properly, doesn't mean they are invalid.

In general, Atom and Sublime Text have syntaxes and originated from the same source, however Sublime Text has been quite a bit of work in fixing edge-case bugs since we've added a more powerful syntax highlighting engine.

@keith-hall You seem to have contributed these basic tests. Since I know nothing about Haskell, would you mind providing some info about these tests?

AllanZyne:
I'm sorry, then I test on ghci, and yes, you're right.

AllanZyne:
I try to fix it by this way, but I can't pop `meta.function.type-declaration` on comment line.
```yaml
push:
    - meta_scope: meta.function.type-declaration.haskell
    - include: type_signature
    - match: ^(?!\1[ \t]|[ \t]*$)
      pop: true
```
qq 20170216103650

Can I do this?

AllanZyne:
The patch on above is wrong! 
I'm sorry, I can't fix this bug.

PR #814, opened 2017-02-22
[Python] Fix single-quoted f-strings

FichteFoll:
The escape sequences includes went into the wrong match sub-context.


wbond:
Thanks!

PR #815, opened 2017-02-22
[ASP] fix small symbol transformation edge case

keith-hall:
This fixes a small symbol transformation edge case in the ASP package.
The edge case was a method definition without arguments & parens, where the method name was in square brackets and contained an open paren - before this fix it got cut off at the open paren.

i.e. `function [test_abc-def#123)_(ghgh]`


wbond:
Thanks!

PR #827, opened 2017-03-10
[Haskell] update for Haskell.sublime-syntax' `pragma` section

metaleap:
A small addition of all supported pragmas in Haskell, because I can never recall "was it `GHC_OPTIONS` or `OPTIONS_GHC`"..


wbond:
Thanks!

If you do find any other Haskell issues, we'd be grateful for input, bug reports or fixes!

PR #828, opened 2017-03-13
[*] fix file extensions

jrappen:
* uses info from code from @ihodev, @equinusocio, @ctf0, @Briles, @jamiewilson

### DIFF
--- a/Groovy/Groovy (Gradle).sublime-syntax
+++ b/Groovy/Groovy (Gradle).sublime-syntax
@@ -2,6 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: Groovy (Gradle)
+hidden: true
 file_extensions:
   - gradle
 scope: source.groovy.gradle
--- a/JavaScript/JSON (Babel).sublime-syntax
+++ b/JavaScript/JSON (Babel).sublime-syntax
@@ -2,6 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: JSON (Babel)
+hidden: true
 file_extensions:
   - .babelrc
 scope: source.json.babel
--- a/JavaScript/JSON (Bower).sublime-syntax
+++ b/JavaScript/JSON (Bower).sublime-syntax
@@ -2,6 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: JSON (Bower)
+hidden: true
 file_extensions:
   - bower.json
   - .bowerrc
--- a/JavaScript/JSON (Composer).sublime-syntax
+++ b/JavaScript/JSON (Composer).sublime-syntax
@@ -2,6 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: JSON (Composer)
+hidden: true
 file_extensions:
   - composer.json
   - composer.lock
--- a/JavaScript/JSON (ESLint).sublime-syntax
+++ b/JavaScript/JSON (ESLint).sublime-syntax
@@ -2,6 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: JSON (ESLint)
+hidden: true
 file_extensions:
   - .eslintrc
   - .eslintrc.json
--- a/JavaScript/JSON (NPM).sublime-syntax
+++ b/JavaScript/JSON (NPM).sublime-syntax
@@ -2,6 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: JSON (NPM)
+hidden: true
 file_extensions:
   - package.json
 scope: source.json.npm
--- a/JavaScript/JSON (PostCSS).sublime-syntax
+++ b/JavaScript/JSON (PostCSS).sublime-syntax
@@ -2,6 +2,7 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 name: JSON (PostCSS)
+hidden: true
 file_extensions:
   - .postcssrc
 scope: source.json.postcss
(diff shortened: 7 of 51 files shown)


keith-hall:
doesn't this mean that every user will have a zillion non-functional syntaxes show up in their list of available syntaxes?

michaelblyons:
I'm a little curious what this is for. I've seen something similar in the "A File Icon" package, but I'd be a little surprised if the purpose was solely for sidebar icons.

FichteFoll:
It does seem like it's mostly for file icons and while I can see the reason behind this, I don't like to have this sort of workaround in the default packages because it's a lot of noise.

 I'd rather have a different mechanism of providing file icons that can somehow take precedence over the current scope-based approach.

ihodev:
@michaelblyons Yep, that's hack for some sidebar icons. That's why it can be disabled via "A File Icon" options.

I completely agree with @FichteFoll I'd rather have a different mechanism of providing file icons.

jrappen:
So I should rather put missing extensions in the main files?

michaelblyons:
I'm kind of stretching to come up with other benefits to this change:
- You could do additional first-line targeting (`<snippet>` -> `XML (Sublime)` or something)
- You could set syntax-specific conventions for indentation (`*.tmPreferences` _really does_ want tabs)

In general, though, I share @FichteFoll's opinion that it's rather noisy.

jrappen:
@michaelblyons This PR is not related to targeting first lines or indentation. Not sure what you were looking at.

michaelblyons:
@jrappen Right, I know. I was trying to think of other reasons that splitting otherwise-identical syntaxes would be beneficial. I apologize if I did not express myself clearly.

wbond:
Jon and I have discussed file icons. I don't think we are going to use a solution like this.

Honestly, there aren't a lot of great options, from where I sit. Obviously themes need to provide the icons for them to look good with the theme. But then different users have different ideas about what kinds of file they want with the same icons. And then there are tons of different file types.

We discussed the idea of using file extensions, but then there would have to be hundreds of theme entries to ensure that the same icon was used for 20 related file extensions (such as with Ruby). We can obviously deal with file icon fallback in the case that a theme does not have an icon for a certain type of file.

We have an idea that we are going to try, but we also have to maintain backward compatibility with existing themes. I'll let you know (and you'll probably see in a dev release) once we have a solution we are happy with.

jrappen:
Going forward, I'll close this. Should I move the relevant extensions to the main language files, or not?

equinusocio:
VSC provide an awesome file icons api, how it work? can be helpful?

wbond:
@equinusocio Interesting, so they treat an icon theme as completely separate from the theme?

ihodev:
@wbond Yes, they separate ui themes (soon), syntax themes (Sublime's color scheme analog) and icon themes and I thinks that's a good solution, because the user can choose them independently. From the dev side we can provide these themes as separate packages or via all-in-one package.

equinusocio:
Right, and ST can provide a basic generict heme like now with the opened/closed folder and a generic file icon. Themes will do the rest. Also i like the VSC icon theme configuration through a json file like this:

```js
{
    "iconDefinitions": {
        "_file_markdown": {
            "iconPath": "./icons/markdown.svg"
        },
        "_file_css": {
            "iconPath": "./icons/css.svg"
        },
        "_folder_dark": {
            "iconPath": "./icons/folder.svg"
        },
        "_folder_light": {
            "iconPath": "./icons/folder-light.svg"
        },
        // ...
    },
    "fileExtensions": {
        "css": "_file_css",
        "postcss": "_file_css",
        "markdown": "_file_markdown",
        // ...
    },
    "fileNames": {
        "gruntfile.js": "_file_grunt",
        "bower.json": "_file_bower",
        // ...
    },
    "folderNames": {
        "node_modules": "_file_nodejs",
        ".git": "_file_git",
        ".github": "_file_github",
        ".gulp": "_file_gulp",
        "bower_components": "_file_bower"
    },
    "folderNamesExpanded": {
        "node_modules": "_file_nodejs",
        ".git": "_file_git",
        ".github": "_file_github",
        ".gulp": "_file_gulp",
        "bower_components": "_file_bower"
    },
    // ...
}
```

Where `iconDefinitions` object define the icon name key that can be used inside other objects like `fileExtensions` etc..

Also VSC support svg formats.. that's not a small thing when we have a lot of screen resolutions around

wbond:
> Also VSC support svg formats.. that's not a small thing when we have a lot of screen resolutions around

SVG has the downside of trading initial theme development for worse performance for every user. Having dealt with skia, having used SVGs on the web and having implemented minihtml, I am fairly sure that any implementation of SVG is going to be full of all sort of edge cases. PNG has the benefit of performance and simplicity.

wbond:
In terms of PNGs, you could check out using Affinity Designer. They have an excellent export persona for vector graphics where you can have it export your vectors as PNGs at regular, @2x, @3x, etc.

equinusocio:
This is an example of the vsc material theme `package.json`:

```js
"themes": [
  	{
        "label": "Material Theme Darker",
        "uiTheme": "vs-dark",
        "path": "./themes/Material-Theme-Darker.tmTheme"
    },
    {
        "label": "Material Theme Lighter",
        "uiTheme": "vs",
        "path": "./themes/Material-Theme-Lighter.tmTheme"
    }
],
"iconThemes": [
	{
	    "id": "material-theme-icons",
	    "label": "Material Theme Icons",
	    "path": "./material-theme-icons.json"
	}
]
```

ihodev:
I think that every thing is ok with PNG icons. That's not the problem to prepare @1x, @2x, @3x assets. However currently ST doesn't pick up @2x and @3x icons on windows and linux depending on dpi scale, only in macOS everything works fine (I think that's because of @2x, @3x is Apple convention of naming).

It would be nice to get "dpi_scale" value to use the right icons or enforce icon usage via ST internals.

wbond:
@ihodev Yes, I am aware of the issue that hi-dpi assets are only use on Windows and Linux when the dpi scale is 2.0. That is planned to be addressed in the next build.

jrappen:
The conversation went off-topic. Let me know if you'd like a new PR with the extension additions to the main syntaxes.

PR #834, opened 2017-03-22
[CSS] don't treat a pipe (|) char as a valid prop name for completions

### LINKED ISSUE #794: [CSS]: Completions not working in meta-scope
First, please consider the following snippet:

    <snippet>
        <content><![CDATA[
    /* this is a test */
    ]]></content>
        <tabTrigger>test</tabTrigger>
        <scope>source.css meta.property-list.css</scope>
    </snippet>

This will expands as expected at this cursor position (`|` marks the cursor):

    body {
        |
    }

Now, consider the following completions file:

    {
      "completions": [
        {
          "contents": "/* this is a test */", 
          "trigger": "test"
        }
      ], 
      "scope": "source.css meta.property-list.css"
    }

I would expect this to expand at the same cursor position, but it does not work at all. The same completion using `source.css` will expand only outside a CSS selector.

keith-hall:
This fixes #794, whereby `|` was being treated as a property name to be included in completions suggestions, when it shouldn't be.

### DIFF
--- a/CSS/css_completions.py
+++ b/CSS/css_completions.py
@@ -433,7 +433,7 @@ def parse_css_data():
         # Append values that are allowed for all properties
         allowed_values += ['all', 'inherit', 'initial', 'unset']
 
-        for name in names.split():
+        for name in names.split(' | '):
             props[name] = sorted(allowed_values)
 
     return props


wbond:
Excellent, thanks!

PR #836, opened 2017-03-22
[XML] first line match SOAP Envelope

keith-hall:
Quite often, when working with web services and web service logs, the XML declaration is omitted. Rather than add more file extensions, I thought it would be a good idea to try to match a [SOAP envelope](http://schemas.xmlsoap.org/soap/envelope/) on the first line of the file to know that it is an XML document, and make the out-of-box experience better for web service developers. :)

(We could, if we wanted, detect any valid XML element (with a namespace, to avoid false positives), but I think just detecting SOAP is enough for now.)

This also tweaks the XML declaration first line match slightly, because the following is valid, as is a tab immediately after `<?xml`:

```xml
<?xml
	version="1.0" encoding="UTF-8"?>
```

### DIFF
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -10,7 +10,7 @@ file_extensions:
   - rss
   - opml
   - svg
-first_line_match: '^<\?xml '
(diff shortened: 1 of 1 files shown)


wbond:
Perhaps this should be formatted into an extended, multi-line regex?

keith-hall:
great idea, done :)

wbond:
Great!

PR #837, opened 2017-03-22
[HTML] scope self closing HR tag punctuation

### LINKED ISSUE #800: [HTML] the punctuation in self closing hr tags is not scoped
<br /><hr /><br />

here, the `/` in the `br` tags gets scope `punctuation.definition.tag.end.html`, but the `/` in the `hr` tag only has the meta scope of the tag `meta.tag.block.any.html`.

keith-hall:
fixes #800


wbond:
Thanks!

PR #838, opened 2017-03-22
[ShellScript] slightly more lenient whitespace handling for increasing indentation

keith-hall:
This was originally suggested on the forums here: https://forum.sublimetext.com/t/bash-indent-to-sensitive-and-lacking/26798


wbond:
I added this in after #1057 in commit 73c4f86.

PR #839, opened 2017-03-22
[C#] Placeholders in multiline strings

michaelblyons:
Highlight `string.Format` replacements for `@"..."` strings, too.


wbond:
Thanks!

PR #840, opened 2017-03-23
[Ruby] Identify Fastlane scripts as Ruby scripts

outlyer:
Adds support for [Fastlane](https://fastlane.tools) build tools used by Android and iOS developers and continuous integration systems. This simply identifies the configuration files as Ruby since they are essentially Ruby scripts.


wbond:
Thanks!

PR #841, opened 2017-03-23
[HTML] merged batch indentation rules to normal rules and annotated them

keith-hall:
solves the problem with HTML unexpectedly unindenting as you type, and allows indentation to be increased and decreased as expected while typing HTML tags (previously one had to rely on pressing <kbd>Enter</kbd> inside an empty tag or using batch reindentation) - https://github.com/SublimeTextIssues/Core/issues/1583

### DIFF
--- a/HTML/Miscellaneous.tmPreferences
+++ b/HTML/Miscellaneous.tmPreferences
@@ -21,11 +21,12 @@
 		<string><![CDATA[(?x)
 			^\s*                                                                         # the beginning of the line followed by any amount of whitespace
 			(
-			    <\?php\s*\b(if|else|elseif|for|foreach|while)\b.*:(?!.*end\1)            # php control keywords that don't end themselves on the same line
+			    <\?(php)?\s*\b(?<php_control_word>if|else|elseif|for|foreach|while)\b.*: # php control keywords
+			    (?!.*end\g<php_control_word>)                                            # - that don't end themselves on the same line
 			|   \{[^}"']*$                                                               # open curly braces that don't have close braces or string punctuation after them
 			|   <!--(?!.*-->)                                                            # comments that don't close themselves on the same line
 			|   <(?!\?|area|base|br|col|frame|hr|html|img|input|link|meta|param|[^>]*/>) # skip self closing tags (tags that end with />, as well as known self closing tags
-			        ([A-Za-z0-9-]+)(?=\s|>)\b[^>]*>(?!.*</\1>)                           # valid tags that don't close themselves on the same line
+			        (?<html_tag>[A-Za-z0-9-]+)(?=\s|>)\b[^>]*>(?!.*</\g<html_tag>>)      # valid tags that don't close themselves on the same line
 			)
 		]]></string>
 		<key>bracketIndentNextLinePattern</key>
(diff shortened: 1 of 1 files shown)


wbond:
I’ll check this out and do some testing in the near future. Thanks for taking the time to figure it out!

jcberquist:
@keith-hall I was just looking at the current HTML indentation rules (because of https://github.com/jcberquist/sublimetext-cfml/issues/68) and your pull request showed up. I appreciate you cleaning this up, it is much clearer now. I do think the back references in the `increaseIndentPattern` might be wrong (possibly they were wrong before too)? There are two that refer to `\1` but I think those should refer to `\2` (for the php control keywords) and `\3` (for html tags that are closed on the same line) respectively?

keith-hall:
@jcberquist you're absolutely right, nice spot and thanks for pointing it out :) I'm just about to drive home but I'll update the PR when I can :)

keith-hall:
I've just pushed a commit that should address this, plus an inconsistency between the php indent and unindent patterns (regarding the `php` after `<?` being optional)

jcberquist:
Nice 👍 This also gives me a chance to thank you for your forum post on indentation. It was awesome, and super helpful to me.

keith-hall:
You're most welcome - thanks for the feedback :)

I've just nicked an idea you had about making HTML tags case insensitive ;)

michaelblyons:
@jcberquist @keith-hall Can you link the post? I went on a (brief) search for it and came up empty. 😉

keith-hall:
https://forum.sublimetext.com/t/everything-you-n-ever-wanted-to-know-about-indentation-in-st3/26207

HTML/Miscellaneous.tmPreferences
```diff
@@ -9,29 +9,25 @@
...
-			^\s*
-			<(?!\?|area|base|br|col|frame|hr|html|img|input|link|meta|param|[^>]*/>)
-			  ([A-Za-z0-9-]+)(?=\s|>)\b[^>]*>(?!.*</\1>)
-			|<!--(?!.*-->)
-			|<\?php.+?\b(if|else(?:if)?|for(?:each)?|while)\b.*:(?!.*end\1)
-			|\{[^}"']*$
+			^\s*                                                                              # the beginning of the line followed by any amount of whitespace
+			(
+			    <\?(php)?\s*\b(?<php_control_word>if|else|elseif|for|foreach|while)\b.*:      # php control keywords
+			    (?!.*end\g<php_control_word>)                                                 # - that don't end themselves on the same line
```
FichteFoll: I know it's been like this before, but using a backref here is quite naive because `elseif` is closed by `endif` and not `endelseif` (similar for `else`, `foreach`). One could batch the keywords that share the same end keyword in a named group and then resort to conditional expressions for the end block.

keith-hall: I've pushed a new commit to correct this. It seems that ST's Find panel works differently compared to the indentation rules (even though I thought they use the same Boost engine), so trying to use a conditional in these rules just causes ST to crash (https://github.com/SublimeTextIssues/Core/issues/1383) so I've used a slightly different approach. Maybe we can revise it after the next build is released (which has a fix for this crash bug), when we can see what the error actually is, but I'm happy with the approach I've used for now.

FichteFoll: Yes, that works too :+1:

wbond:
I am sure lots of people will be happy with this change!

PR #843, opened 2017-03-24
[JavaScript] many issue fixes

### LINKED ISSUE #86: [JavaScript] hotkey conflict
Does someone notice this problem?

In `JavaScript/Object-Method.sublime-snippet`, it uses `:f` as tab trigger. But once I type the letter `f`, other snippets (like `for`, `function`) are triggered.

I have to type `esc` first, then `tab`. Seems a little stupid.

### LINKED ISSUE #604: [JavaScript] Regular Expressions syntax: bug in group-definition?
I think there's a bug [here](https://github.com/sublimehq/Packages/blob/94f09eb/JavaScript/Regular%20Expressions%20%28JavaScript%29.sublime-syntax#L69-L75) in `JavaScript/Regular Expressions (JavaScript).sublime-syntax`:

``` yaml
group-definition:
  - match: (\()((\?:))?
    captures:
      1: punctuation.definition.group.regexp
      3: punctuation.definition.group.capture.regexp
      5: punctuation.definition.group.capture.regexp
      6: punctuation.definition.group.no-capture.regexp
```

Specifically, [non-capturing groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-non-capturing-parentheses) seem to be scoped as [capturing groups](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-capturing-parentheses).

For example:

```
/(?:alice)bob(charlie)/
  ^^
  punctuation.definition.group.capture.regexp
```

I think the `?:` should be scoped as `punctuation.definition.group.no-capture.regexp`.

### LINKED ISSUE #640: ES2015 (ES6) Syntax highlighting in template literals (i.e. instead of string concatenation)
Using ST3 build 3125 x64 in win10 x64.
For example (from [here](http://eslint.org/docs/rules/prefer-template))

```
var str = "Hello, " + name + "!";
var str = `Hello, ${name}!`;
```

Here is how it's highlighted in ST3: (see the `name` variable in line 2) 

And, for reference, here is in VSCode 1.5.3, i.e. correctly:

---

And another example:

```
let arr = [], temp;
$.each(arr, (index, value) => {

    arr[index].innerHTML = '(' + temp + ') ' + arr[index].innerHTML;
    arr[index].innerHTML = `(${temp} ) ${arr[index].innerHTML}`;
});
```

In ST3: (see both `temp` and `arr[index].innerHTML` in line 5)

and for reference, here is in VSCode:

<br>
Related issue: #213

### LINKED ISSUE #665: [JavaScript] A main object is not highlighted as other main objects
JavaScript has 3 main objects: `window`, `document` and `navigator`. However, the `navigator` object is not highlighted. This is the `document` object:
observacionsublime2 fw
And this is the `window` object:
observacionsublime3 fw
But this is the `navigator` object:
observacionsublime fw
The `navigator` object is not highlighted as the `document` and `window` objects. This is a few suggestion in JavaScript syntax.

### LINKED ISSUE #778: Syntax highlight issue in ternaries (JavaScript)
Using JavaScript syntax, special values like primitives, null, true, ... are not highlighted  below:

```js
var query = {
	type: type==undefined ? null : {$in: type.split(',')}
};
```
image

### LINKED ISSUE #780: Syntax hightlight issue with object shorthands (JavaScript)
the field is highlighted (in green in the image below) is a comment follows, for no specific reason
```js
var pop = 1;
var o = {
	pop // ok
}
```
image

### LINKED ISSUE #791: [JavaScript] Control word following construction without parentheses or semicolon is not highlighted
These two work as expected:
````javascript
const abc = new Set()
console.log('abc') //"console" is highlighted as a special object
````
and
````javascript
const abc = new Set;
if (true) console.log('abc') //"if" is highlighted as a control word
````
But neither of these work:
````javascript
const abc = new Set
console.log('abc') //"console" is not highlighted
````
````javascript
const abc = new Set
if (true) console.log('abc') //"if" is not highlighted
````
image
These are all valid ways to write the same thing.

### LINKED ISSUE #826: [JavaScript] Return, throw, and yield do not properly end at end of line.
Example A (correct):

    return;
        {a: 1};

Example B (incorrect):

    return
        {a: 1};

`Return` is a restricted production; therefore, it ends at the end of the first line. These two examples are semantically equivalent, and the second line is a block. However, in Example B, the second line is highlighted as an object literal as though it were part of the return statement.

The behavior is also incorrect for `throw` and `yield`.

keith-hall:
Fixes #826 and fixes #791 and fixes #780 and fixes #778 and fixes #640 and fixes #604 and fixes #665 and fixes #86

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -410,6 +410,7 @@ contexts:
     - match: '/(?=[^/*])'
       scope: punctuation.definition.string.begin.js
       push: regexp
+    - include: constants
     - match: '(?=[\S])'
       pop: true
 
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -710,6 +710,10 @@ contexts:
     - include: method-declaration
 
   keywords:
+    - match: ^\s*\b(?:(throw)|(yield|return))\b\s*$
+      captures: # these keywords are restricted productions and implicitly end at the end of the line, so we don't push the after-operator scope
+        1: keyword.control.trycatch.js
+        2: keyword.control.flow.js
     - match: \bthrow\b
       scope: keyword.control.trycatch.js
       push: after-operator
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1334,7 +1334,7 @@ contexts:
       scope: support.class.builtin.js
     - match: \b((Eval|Range|Reference|Syntax|Type|URI)?Error)\b
       scope: support.class.error.js
-    - match: \b(document|window)\b
+    - match: \b(document|window|navigator)\b
       scope: support.type.object.dom.js
     - match: \b(Buffer|EventEmitter|Server|Pipe|Socket|REPLServer|ReadStream|WriteStream|Stream|Inflate|Deflate|InflateRaw|DeflateRaw|GZip|GUnzip|Unzip|Zip)\b
       scope: support.class.node.js
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -577,7 +577,7 @@ contexts:
         - include: square-brackets
         - match: '(?=\()'
           set: constructor-params
-        - match: '(?=\S)'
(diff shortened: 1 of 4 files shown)


wbond:
Very nice!

PR #845, opened 2017-03-24
[PHP] indentation improvements

### LINKED ISSUE #140: [PHP] array indent
when reindenting php arrays i get the following:

```
public $helpers = [
'Profile',
'Session',
'Text',
'Time',
];

public $test = array(
    'Profile',
    'Session',
    'Text',
    'Time',
    );
```

i expect:

```
public $helpers = [
    'Profile',
    'Session',
    'Text',
    'Time',
];

public $test = array(
    'Profile',
    'Session',
    'Text',
    'Time',
);
```

does anyone know how to fix this?

keith-hall:
fixes #140 and makes PHP indentation consistent with HTML indentation when the PHP tag isn't closed before the end of the line.

### DIFF
--- a/PHP/Indentation Rules.tmPreferences
+++ b/PHP/Indentation Rules.tmPreferences
@@ -8,7 +8,20 @@
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string><![CDATA[(?x) ^ (.*\*/)? \s* \} .* $|<\?(php)?\s+(else(if)?|end(if|for(each)?|while))]]></string>
+		<string><![CDATA[(?x)
+			(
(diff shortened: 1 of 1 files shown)


michaelblyons:
If this is here... Why are there PHP indentation rules in the base HTML syntax?

keith-hall:
Because it all depends what scope is applied to the `\n` character at the end of the line - the HTML rules apply if the PHP tag is closed i.e. ` if true: ?>\n` so its the only way to do it at the moment, hardly ideal I know...
The indentation rules in the PHP package apply when the PHP tag isn't closed before the end of the line.

wbond:
Thanks!

PR #846, opened 2017-03-24
[PHP] support embedded syntaxes in nowdoc sections

### LINKED ISSUE #609: [PHP] embedded sections should work with nowdoc too
e.g.:

``` php
$selectSql = <<<'SQL'
    SELECT id, name
    FROM products
    WHERE price <= 1000
SQL
```

keith-hall:
fixes #609


wbond:
We should probably be able to support this now with the `embed` feature

Issue #715, opened 2016-11-30
Interpolated strings


FichteFoll:
I've compiled a few scope names for languages that have interpolation in strings (incomplete due to lack of time right now):

Bash: `$(fk yea)` -> `string.interpolated`; `"${apples}"` -> `variable.other.bracket`
C#: not implemented
Groovy: `source.groovy.embedded.source`
JavaScript: `string.template` into `source.js.embedded.template`
Perl: 
Php:
Python: see https://github.com/sublimehq/Packages/issues/715#issuecomment-289097254 (`meta.string.interpolated` for the entire string and `meta.interpolation source.python.embedded` for the interpolated parts)
Ruby: `source.ruby.embedded.source`
Scala:

(for examples, see https://en.wikipedia.org/wiki/String_interpolation)

All these add on top of the string scope and do not clear it. Punctuation scopes are inconsistent.
This raises the following items:

1) I remember that you, @wbond, considered using `clear_scopes` for interpolated strings to clear the string scope so that the interpolated code would not be using the string scoping. I like this a lot, and will most likely use it for Python's f-strings. We should apply that to the other syntaxes too, however.

2) I intend to also add `punctuation.definition.string.interpolation.begin` and `.end` respectively, so that color schemes could target `punctuation.definition.string` and would match all the string edge punctuation.

3) It seems like `source.<language>.embedded.source` is the most common scope for interpolated sections. Should this be the standard?

wbond:
So, my reaction is:

 1. Yes, I do believe `clear_scopes` is the correct thing to use here. The downside is now you can't select the whole string using `string.*` scopes. Thus I propose strings have a `meta.string.*` that will include all string parts, interpolated begin/end tokens and all embedded source.
 2. Since this is the beginning of a non-atomic element, I think the scope `punctuation.section.interpolation.*` should be used for begin and end tokens. These tokens should, IMO, *not* get `string.*` since they aren't technically string contents, but markers.
 3. Yes, I think that works for the scope. The `punctuation.section.interpolation.*` scopes should not be part of this, but should be just before and just after it.

wbond:
I suppose the `meta.string` only technically needs to be added to strings that can contain interpolated sections. Should it be included on *all* strings, for consistency? For example, single quoted strings in PHP.

FichteFoll:
[...]

PR #848, opened 2017-03-24
[Python] Add tests for interpolation punctuation

FichteFoll:
As I read through #715, I noticed I didn't add any tests for punctuation in interpolated sections, so that's what this PR is for.


wbond:
Thanks!

PR #850, opened 2017-03-27
[TCL] regexp improvements

### LINKED ISSUE #779: ST3 Tcl syntax highlighting not working with regexp
Here is another bug in the syntax highlighting of regexp command:

regexp -inline -all {%[a-zA-Z_]*%} "whatever"

the -inline option appears in green (string) instead of white (option).

With:

regexp -all -inline {%[a-zA-Z_]*%} "whatever"

the -all option appears in green (string) instead of white (option).

### LINKED ISSUE #783: ST3 Tcl syntax highlighting not working with regexp
In the syntax highlighting of following Tcl code in ST3, the right parenthesis is marked in red as being  wrong, which is not true:

set objRegExp {(.{0,200})}

image

### LINKED ISSUE #784: ST3 Tcl syntax highlighting not working with regexp
In the syntax highlighting of following Tcl code in ST3, the a lot of content is marked in red as being wrong, which is not true:

set objRegExp {(^[a-zA-Z]{2}[a-zA-Z0-9-]{2,12}$)}

image

keith-hall:
fixes #779, fixes #783, fixes #784

### DIFF
--- a/TCL/Tcl.sublime-syntax
+++ b/TCL/Tcl.sublime-syntax
@@ -122,10 +122,10 @@ contexts:
       set:
         - match: '(?={{end_chars}})'
           pop: true
-        # Consume the switches
-        - match: '\s+-(all|expanded|line|linestop|lineanchor|nocase|start|--)\b(?=\s)'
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #851, opened 2017-03-27
[Markdown] MultiMarkdown: better header support

keith-hall:
This PR adds better header support to the MultiMarkdown syntax, and adds a syntax test for it. The rules for the header were taken from http://fletcher.github.io/peg-multimarkdown/index
This fixes https://github.com/SublimeTextIssues/DefaultPackages/issues/162


wbond:
Thanks!

PR #856, opened 2017-03-27
[Scala] Reworked auto-indentation

djspiewak:
There's quite a bit of stuff here.  It leverages the new meta scoping which landed recently (thanks, @wbond!) together with some more idiomatic expression wrapping detection to produce auto-indentation which does a fairly decent job of closely approximating the standard structure of a Scala file.  Some notable features:

- Auto-indentation (indent *and* dedent!) for `case` blocks
  + Takes into account first vs non-first blocks
  + Correctly handles nested matches
- Support for standard "block" constructs
  + `if`
  + `else`
  + `for`
  + `do`
  + `while`
  + `yield`
- Parentheses/bracket indentation
- Definitions (e.g. `def`, `type`, etc) involving `=` which wrap definitions to the next line

Unfortunately, as with all sublime auto-indentation modes, it's relatively easy to break this.  For example, if you take the following text and reindent it, the results will be incorrect:

```scala
x match {
  case Foo =>
    "hey"
  case Bar =>
    "thar!"
}
```

Specifically, if you auto-reindent this, the results will be the following:

```scala
x match {
  case Foo =>
    "hey"
  case Bar =>
    "thar!"
  }
```

There is no way to solve this problem with Sublime as it currently stands, since it is impossible to reduce indentation by more than one level in the span of a single token (which is what needs to happen here).

Despite the flaws, this is vastly superior to the Scala auto-indentation in master.  I've been dogfooding this in various incarnations for about six months, and this latest version of it has been pretty seamless for me.  Seamless enough that I'm comfortable pull requesting it in its current form.


keith-hall:
Nice work @djspiewak  :) apologies if this is a silly question, as I don't use Scala, but, as the ideal indentation isn't possible, (why) would it not be preferable to have your `case` statements reindented as:

```scala
x match {
case Foo =>
  "hey"
case Bar =>
  "thar!"
}
```

(i.e. don't indent the top level case statements) so that the indentation of the rest of the file following the `case` statements remains correct?

also, you may want to consider adding some comments to the indentation rules regex patterns, so that they will be easier to revisit/maintain at a later date :)

djspiewak:
> (i.e. don't indent the top level case statements) so that the indentation of the rest of the file following the case statements remains correct?

I thought about it, but there are two reasons not to.  First off, as a matter of typing, because of sublime's hard-coded semantics regarding <kbd>Enter</kbd> within curly braces, you actually don't see the flaws with the current dedent unless you explicitly reindent that line, and in practice that doesn't happen very often.  The auto indenter is really just an approximation; idiomatic Scala indentation rules are far more complicated than anything sublime can currently express.  So there's literally never a use case where I would select an entire file and automatically reindent the whole thing, which is where this dedent semantic would be most harmful.

Using your semantic would mean that the typing experience would be worse, since users would have to manually add an extra indentation level to the `case` keyword as they type it, which seems like a poor experience.  Your semantic does allow for more localized harm in the case of a global reindent, but like I said, global reindentation is not a very common action while typing is.  :-)

In practice, I actually almost never notice this issue.  It comes up *on occasion* when I reindent some text that spans the closing brace for a `match`, but it's usually a matter of an instant to identify what happened and how to fix it.  The general, day-to-day experience is completely seamless, which again is due to the hard-coded curly brace behavior in sublime.

> also, you may want to consider adding some comments to the indentation rules regex patterns, so that they will be easier to revisit/maintain at a later date :)

Just using XML comment blocks?  I'm always leery about PLIST files, since I'm never sure exactly what is valid, especially within the data blocks within the file.

FichteFoll:
>Just using XML comment blocks? I'm always leery about PLIST files, since I'm never sure exactly what is valid, especially within the data blocks within the file.

While XML comments would work, enabling extended mode (`(?x)`) allows using `#` as a line comment marker within the regular expression, which is usually the preferred way.

wbond:
If you could annotate the regexes with comments, I think we'll be good to merge. Thanks for your work on this @djspiewak!

djspiewak:
@wbond Commented all the things!

Also, @FichteFoll, I'm reminded of just how great your PackageDev plugin is!  The `tmPreferences` highlighting is superb.

wbond:
Thanks @djspiewak!

PR #857, opened 2017-03-27
[Markdown] add table support

keith-hall:
this PR adds table support to the Markdown syntax, as per https://github.github.com/gfm/#tables-extension-


wbond:
Nice!

PR #858, opened 2017-03-28
[Java] handle dot access to fields named "class"

keith-hall:
to prevent highlighting breaking after `id.class` due to `class` being treated as a class definition, match the identifier if it isn't a function call (though it currently doesn't scope it, but this is existing behavior). This ensures that function calls are still handled correctly.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -390,8 +390,9 @@ contexts:
       scope: keyword.operator.arithmetic.java
     - match: (!|&&|\|\|)
       scope: keyword.operator.logical.java
-    - match: \.
-      scope: punctuation.accessor.dot.java
+    - match: (\.)({{id}}(?!(\s*{{within_generics}})?\s*\(|{{id}}))?
+      captures:
+        1: punctuation.accessor.dot.java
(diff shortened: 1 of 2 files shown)


wbond:
Hmm, this seems like it could have a side effect of changing more than just `.class`. What about a specific match for `.class` and a test? That way we can refactor in the future and ensure we don't break that case.

Alternatively, @djspiewak may have some input on potential Java syntax this may affect.

djspiewak:
I would rather have a specific match for `(\.)(class)\b` for the time being.  I have a partially-completed local branch which refactors some of this stuff to be more robust (e.g. correctly highlighting fully-qualified identifiers, allowing qualified identifiers to line-wrap, etc), but it's not ready yet.

djspiewak:
Oh, as a sidebar, shouldn't `(class)\b` in this match be given some sort of `keyword` scope?  It's a magic function; not a normal identifier.

keith-hall:
I've changed it to use a specific match, and scoped it the same way that Python scopes it's magic identifiers.

wbond:
Thanks guys!

PR #859, opened 2017-03-28
[Markdown] support GFM autolinks

keith-hall:
This enhances the Markdown syntax by adding support for autolinks as defined by the GitHub Formatted Markdown specification: https://github.github.com/gfm/#autolinks-extension-


wbond:
Thanks!

Issue #421, opened 2016-05-21
⟦Discussion⟧ Scope for keys in mappings/dictionaries (e.g. JSON, YAML)


FichteFoll:
**Terminology**:

- A *colored* scope is a scope name that most color schemes define a color for. Examples include `string", `entity` or `constant` and specifications of those.
- An *uncolored* scope is a scope name that is usually not colored, mainly `meta` and `puncutation` scopes.

### Introduction

I'd like to discuss the standardization of a specific scope name for key elements in dictionaries or mappings or objects or whatever you call them in your language.

Currently, the JSON definition doesn't have any specific highlighting for keys which it could certainly use and I indeded to add for a while (see also https://github.com/wbond/package_control_channel/pull/5496). There are different scopes for the key and value, but in a way that makes it impossible to specifically target the key with a selector.

In the YAML definition `entity.name.tag` is used, which is a "steal" from xml-like syntaxes where entity.name.tag should denote a tag name.

In most *programming* languages, no special highlighting is done.

Earlier this year, the JavaScript syntax was updated with a *colored* scope for keys in object literals, but many people seemed opposed of this change (hurr change is bad durr), thus causing the change to be reverted and an *uncolored* meta scope being used instead.

### Concluding questions

1. Should keys be highlighted with a "colored" scope and which one should it be? I personally like `entity.name.key`.
2. Should keys not be highlighted and get an "uncolored" meta scope instead?
3. Decide on a case-by-case basis? How complex would this be for color scheme creation?

keith-hall:
Personally, the first thing I do when I install ST on a new machine is modify the color scheme to highlight JSON keys. To me, it just looks wrong to have it not colored. Especially as punctuation is not colored, its harder to distinguish at a glance, where a key ends and the punctuation begins. Keys are strings and should be highlighted - simple as IMO ;)

As for scope name, I have an additional question - do we want any keys to show up in "Goto Anything"? I believe that at a minimum, "root" level keys should appear here for easier document traversal, but that's just my 2 cents. `entity.name.key` works for me, I like that the beginning is consistent with xml "keys" - `entity.name.tag`.

keith-hall:
[...]

PR #862, opened 2017-03-29
[JSON] Meta scopes for mapping and sequence

FichteFoll:
In direct response to #421.

Use `meta.key` and `meta.value` for key and value tokens respectively. In the process, generalize `meta.sequence` and `meta.mapping` (and "fix" their punctuation scopes).

Investigation on how this might affect third-party color schemes should likely be performed, for the latter change only, which I doubt had any special coloring applied.
The scopes changed in the first change have only been added in https://github.com/sublimehq/Packages/commit/5e24e2b77f6ee16e2fbc705558d393b0f1d58485 (https://github.com/sublimehq/Packages/pull/816), which hasn't been part of any build yet.

### DIFF
--- a/JavaScript/JSON.sublime-syntax
+++ b/JavaScript/JSON.sublime-syntax
@@ -96,13 +96,13 @@ contexts:
         - match: '"'
           scope: punctuation.definition.string.begin.json
           push:
-            - meta_scope: meta.key.json string.quoted.double.json
+            - meta_scope: meta.mapping.key.json string.quoted.double.json
             - include: inside-string
         - include: comments
         - match: ":"
           scope: punctuation.separator.mapping.key-value.json
           push:
-            - meta_scope: meta.value.json
+            - meta_scope: meta.mapping.value.json
             - match: '(,)|(?=\})'
               captures:
                 1: punctuation.separator.mapping.pair.json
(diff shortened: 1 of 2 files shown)


wbond:
Why `meta.mapping.json meta.key.json` and not specialize to `meta.mapping.key.json`?

I can see the appeal of `meta.value.json` since it can be used in a `mapping` and `sequence`.

wbond:
Actually, it doesn't appear that `meta.value` is being applied to sequence values.

So this is just changing the scope names, correct?

FichteFoll:
I didn't chose `meta.mapping.key` because the token would (most likely?) already have the `meta.mapping` scope applied to them. Furthermore, it allows to target keys individually outside of a mapping, although I don't think this is ever going to happen, since key-value storages are pretty much always mappings.
Thus, it probably makes sense to have "key" and "value" under `meta.mapping` to better indicate their purpose on their own.

I don't exactly see a merit in a `meta.sequence.value` scope, since `meta.sequence - punctuation.separator.sequence` should cover this already, pretty much, should it ever be needed. I could add it anyway, of course.

FichteFoll:
Any update on this?

FichteFoll:
Just a heads-up that I fixed the merge conflicts.

Issue #861, opened 2017-03-29
[JavaScript] String ES6 method definitions break syntax highlighting, method list


wearhere:
In Sublime Text 3, build 3126, on OS X 10.12.3 (fwiw), using an [ES6 method definition](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions) with a _string_ method name breaks ST's syntax highlighting and also related functionality like its "method lists" (what you see when you hit Cmd-R).

--------------

Example:

```js
const SalesforceContactsImportView = Backbone.View.extend({
  events: {
    'focusout'() { this.remove(); },
  },

  render() {
    return this;
  }
});
```
<img width="466" alt="screen shot 2017-03-28 at 10 06 57 pm" src="https://cloud.githubusercontent.com/assets/110000/24439440/1a73d5ec-1403-11e7-9ade-13c965072a74.png">
<img width="347" alt="screen shot 2017-03-28 at 10 07 06 pm" src="https://cloud.githubusercontent.com/assets/110000/24439444/213cc9a6-1403-11e7-92b2-0a86cd7a70b4.png">

--------------

If you use an ES6 method definition _without_ a string method name, you're fine:

```js
const SalesforceContactsImportView = Backbone.View.extend({
  events: {
    focusout() { this.remove(); },
  },

  render() {
    return this;
  }
});
```
<img width="494" alt="screen shot 2017-03-28 at 10 15 13 pm" src="https://cloud.githubusercontent.com/assets/110000/24439603/24a6eeae-1404-11e7-9a61-eaa04c50084c.png">
<img width="373" alt="screen shot 2017-03-28 at 10 15 20 pm" src="https://cloud.githubusercontent.com/assets/110000/24439606/276c764a-1404-11e7-840b-5bdf1cf7038a.png">

--------------

Or, if the method's name remains a string, but you don't use an ES6 method definition, you're also fine:

```js
const SalesforceContactsImportView = Backbone.View.extend({
  events: {
    'focusout': function() { this.remove(); },
  },

  render() {
    return this;
  }
});
```

<img width="534" alt="screen shot 2017-03-28 at 10 07 48 pm" src="https://cloud.githubusercontent.com/assets/110000/24439451/2653d2cc-1403-11e7-81fe-9b0b81ce1d98.png">
[...]

PR #863, opened 2017-03-29
[JavaScript] Add support for string ES6 method definitions

wearhere:
Which had previously been mistaken for object literal keys. The fix entails
adding a new form of method declaration, using a string method name; and
giving that precedence over object literal key matching.

Fixes https://github.com/sublimehq/Packages/issues/861.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1097,9 +1097,8 @@ contexts:
                 - include: either-function-declaration
                 - match: '(?=\S)'
                   pop: true
-        # An opening parenthesis is not allowed after a literal key; that makes
-        # it a method declaration instead, below.
-        - match: "(?=\"|')(?!.+\\()"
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -1097,7 +1097,9 @@ contexts:
                 - include: either-function-declaration
                 - match: '(?=\S)'
                   pop: true
-        - match: "(?=\"|')"
+        # An opening parenthesis is not allowed after a literal key; that makes
+        # it a method declaration instead, below.
+        - match: "(?=\"|')(?!.+\\()"
```
FichteFoll: This will also *not* match `x = { "key with an opening paren (": true }`.

wearhere: Great catch @FichteFoll! Take 2: 86c0837

wbond:
Thanks!

PR #865, opened 2017-03-30
[Markdown] improvements for lists in block quotes

keith-hall:
tweaks the scopes for lists inside block quotes to be consistent with the lists outside block quotes


wbond:
Thanks for catching that!

PR #866, opened 2017-03-30
[C#] @ escapes reserved words

michaelblyons:
You can use `@` to tell C# that a normally-reserved word is not to be used in its normal context. [MSDN reference.][msdn]

```cs
var group = MakeGroup();
// Current syntax gets this right.

var contents1 = group.GetContents();
// Current syntax marks `group` as 'keyword.other.sql.cs',
// which is not technically correct because the code compiles
// fine, but it doesn't bother me too much.

var contents2 = @group.GetContents();
// Current syntax marks `group` as 'keyword.other.sql.cs'
// and `@` as 'invalid.illegal.reserved-char.cs', which is wrong.
```

I notice that `@void` already does what it ought to, and added tests for those, too.

[msdn]: https://msdn.microsoft.com/en-us/library/x53a06bb.aspx
[l40]: https://github.com/sublimehq/Packages/blob/master/C%23/C%23.sublime-syntax#L40
[l1122]: https://github.com/sublimehq/Packages/blob/master/C%23/C%23.sublime-syntax#L1122

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -37,7 +37,7 @@ variables:
   type_suffix_capture: '(\?)?{{brackets_capture}}?(?:\s*(\*))?'
 
   reserved: '(?:abstract|as|base|break|case|catch|checked|class|const|continue|default|delegate|do|else|enum|event|explicit|extern|finally|fixed|for|foreach|goto|if|implicit|in|interface|internal|is|lock|nameof|namespace|new|null|operator|out|override|params|private|protected|public|readonly|ref|return|sealed|sizeof|stackalloc|static|string|struct|switch|this|throw|try|typeof|unchecked|unsafe|using|virtual|volatile|while)'
-  name: '(?:@{{reserved}}|@{{base_type}}|@var|{{name_normal}})'
+  name: '(?:@{{reserved}}|@{{base_type}}|@var|@?{{name_normal}})'
 
   start_char: '(?:{{unicode_char}}|[_\p{L}])'
   other_char: '(?:{{unicode_char}}|[_0-9\p{L}])'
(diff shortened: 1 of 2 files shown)


keith-hall:
This is great :) some projects I've used contain a variable called `from`, which would get highlighted as `keyword.other.sql.cs`, and your PR mostly fixes it, but there are a few scenarios where it still gets it wrong:

```cs
var from = "hello";
var to = "world";
var test = from + " " + to;
from = "foobar";
```

do you think it would make sense to also not scope it as a sql keyword if it is followed by a `=` or `+` operator?

michaelblyons:
Thanks! I'm not sure how I would do that without either negative lookbehinds (which are verboten in the new regex engine?) or basically building a C#-sql engine.

keith-hall:
My idea was just to change your `(?!\s*[.,;])` addition at https://github.com/sublimehq/Packages/pull/866/files#diff-9b5e53d4e14d87efc8f78017464d6b94R1127 to `(?!\s*[.,;+=])`

michaelblyons:
Excellent point. I don't think it can match a completely precise list of matches, but I can add some more.

keith-hall:
Thanks! :) Yeah, I'm not expecting preciseness here, and I think adding a whole C#-sql parser is out of scope for this PR ;)

for this reason, I deliberately didn't suggest adding `-` to that character class, because the following is perfectly valid:
```cs
(from c in Example
select -c.ExampleID)
```

and there is no easy way to detect that vs:
```cs
var test = from -2;
```
unless we use some complicated lookaheads, but I think your PR realistically covers most legitimate use cases now :)

wbond:
Excellent!

PR #867, opened 2017-03-31
[C#] use standard unclosed-string behavior

keith-hall:
This PR changes the unclosed string behavior in the C# syntax to match that of the other syntaxes, like Python, which scopes the new line character as invalid, which is less jarring than having the whole string appear as invalid.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1416,8 +1416,8 @@ contexts:
       scope: punctuation.definition.string.end.cs
       pop: true
     - include: string_placeholders
-    - match: '[^"]*$'
-      scope: invalid.string.newline.cs
+    - match: $\n?
+      scope: invalid.illegal.unclosed-string.cs
       pop: true
 
   format_string:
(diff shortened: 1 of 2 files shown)


michaelblyons:
Great! I _do not_ think this should get in the way of merging, but there are some rules about newlines in interpolation that we can work on at some point.

```cs
string good1 = $"bar { baz }";
string good2 = $@"bar { baz }";
string good3 = $@"bar {
    baz }";
string good4 = $@"bar {
    baz
    }";
string good5 = $@"bar
                { baz }";

string bad1 = $"bar {
    baz }"; // Interpolation must be single line without $@"..."
```

This would be pretty easy if you can `with_prototype` into an `include` instruction, but I suspect it's more likely that the `string_interpolation` construct will have to be broken up.

keith-hall:
I've improved it a tiny bit, but something like `string unclosed_interpolation = $"inner {new[] { 12, 24` won't be detected as invalid atm. I can't use `with_prototype` here because we hit this issue: https://github.com/SublimeTextIssues/Core/issues/1288

michaelblyons:
Excellent, this is much more elegant than I had expected. =)

wbond:
Thanks!

Issue #868, opened 2017-03-31
[C] Wrong auto-indentation level after a loop without body


ghost:
Example:

```c
int main(void){
  int a;

  for(a = 0; a < 1; printf("%d\n", a++));

  return 0;
}
```

If cursor is set to position at the end of the line containing for loop and enter is pressed, indentation level increases like we are in a new code block. But actually, this for loop has no body, so indentation level should remain the same.

OTOH in the following example while loop works correctly:

```c
int main(void){
  int a = 1;

  while(printf("%d\n", --a), a);

  return 0;
}
```

Here after while loop indentation level stays the same, which means that the problem is related only to for loops;

keith-hall:
You could change [this regex](https://github.com/sublimehq/Packages/blob/59d749c1d73198cd848357d7c89740c915ee8b6a/C%2B%2B/Indentation%20Rules.tmPreferences#L26) to `^ \s* \bfor\b \s* (?:[^;]++|;[ ]+(?!$))+ \s* $` to get it to ensure a `;` isn't at the end of the line before deciding to indent the next line.

WsDwA1022:
@keith-hall So, should I create a PR?

keith-hall:
A regex like this would seem to cover most scenarios, including comments (as long as the parens are balanced before the closing `for` paren):

```regex
^ \s* for\b \s* \(

(?<balanced_parens>
    [^()]++
    (?:\(\g<balanced_parens>\)[^()]*+)?
)

\)
(?<eol_or_comment>\s*(?:$|//|/\*(?:(?:(?!.*\*/).*$|.*?\*/\g<eol_or_comment>))))
```

PR #872, opened 2017-04-03
[C] Fix indentation level

WsDwA1022:
Fix indentation level after for loops without body (fix [#868](https://github.com/sublimehq/Packages/issues/868)).

### DIFF
--- a/C++/Indentation Rules.tmPreferences
+++ b/C++/Indentation Rules.tmPreferences
@@ -23,7 +23,7 @@
 		<key>bracketIndentNextLinePattern</key>
 		<string>(?x)
 		^ \s* \b(if|while|else)\b [^;]* $
-		| ^ \s* \b(for)\b .* $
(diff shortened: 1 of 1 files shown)


C++/Indentation Rules.tmPreferences
```diff
@@ -23,7 +23,7 @@
 		<key>bracketIndentNextLinePattern</key>
 		<string>(?x)
 		^ \s* \b(if|while|else)\b [^;]* $
-		| ^ \s* \b(for)\b .* $
+		|   ^ \s* \bfor\b (?:[^;]++|;[ ]+(?!$))+ \s* $
```
deathaxe: This pattern breaks indention after

```C
for(i=0;dd<5;i++)
```

The rule shouldn't require a whitespace after each `;` to make indention work properly.

deathaxe: This rule breaks intention, even if the `;` is part of the comment, but this seems the case for most such rules.

```c
for(i = 0; dd < 5; i++) // comment ;
// no indention applied here!!
```

PR #877, opened 2017-04-04
[Markdown] reduce backtracking possibilities

keith-hall:
this PR reduces the number of backtracking possibilities for regex engines like Oniguruma that use backtracking, and also explains the regex patterns better.

a simple Markdown snippet with a missing end backtick like

    ``he`lloworld`

will now fail in 73 steps rather than over 5000!

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -8,22 +8,22 @@ file_extensions:
   - markdown
   - markdn
 scope: text.html.markdown
-comment: this definition aims to (eventually) meet CommonMark specifications http://spec.commonmark.org/
+comment: this definition aims to (eventually) meet CommonMark specifications http://spec.commonmark.org/ with GitHub Formatted Markdown extensions https://github.github.com/gfm/
 variables:
     thematic_break: |-
         (?x:
-            [ ]{,3}
-            (?:
-                    [-](?:[ ]{,2}[-]){2,}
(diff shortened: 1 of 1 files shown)


FichteFoll:
I find the majority of these comments to be redundant. I can see myself that `[ ]{,3}` matches "between 0 to 3 spaces" and do not need a comment to explain that. This is exactly like adding a comment to `i += 1 # increase i by 1`. Comments should be used to convey information that cannot immediately be seen, such as `(?!`)      #   makes the {3,} possessive` (although it appears that this pattern specifically is redundant because the preceding match already ensures that there can't be any following backticks ([ref](https://github.com/sublimehq/Packages/pull/877/files#diff-bc7dff9190deda9f015e7ef8a4f3ceebR100)).

keith-hall:
I quite like the idea of being able to just read the comments to understand what a regex is doing, without having to look at the pattern itself at all, hence the "redundant" comments :) if you were to become the main maintainer of this package/syntax then I'm sure you could feel free to remove them ;)

I've updated the code fence regex you didn't like though :)

FichteFoll:
Of course, I was just voicing my opinion and giving an analogy to code.

I just spotted the code fence regex while trying to look for a comment that does not literally describe the regex and something seemed off, so that was pure coincidence. 😆

wbond:
Everything passes with the latest unreleased version of Sublime Text

PR #878, opened 2017-04-04
[*] update extensions

jrappen:


JavaScript/JSON.sublime-syntax
```diff
@@ -2,17 +2,24 @@
 ---
 name: JSON
 file_extensions:
+  - .babelrc
+  - .bowerrc
+  - .eslintrc
+  - .no-sublime-package
+  - .postcssrc
+  - .stylelintrc
   - json
```
FichteFoll: `json` must be the first extension because the first will be added as the default extension when saving a file under a new name with this syntax selected.

JavaScript/JSON.sublime-syntax
```diff
@@ -2,17 +2,24 @@
 ---
 name: JSON
 file_extensions:
+  - .babelrc
+  - .bowerrc
+  - .eslintrc
+  - .no-sublime-package
```
FichteFoll: This file is supposed to be empty. It doesn't deserve to be here.

ShellScript/Shell-Unix-Generic.sublime-syntax
```diff
@@ -13,6 +13,22 @@ file_extensions:
...
+  - .git
+  - .git-blame
+  - .git-commit
+  - .git-graph
+  - .gitattributes
+  - .gitblame
+  - .gitcommit
+  - .gitconfig
+  - .gitgraph
+  - .gitignore
```
FichteFoll: Most of these are *not* shell syntax and instead something custom (more akin to TOML or INI). I do not think they should be here.

JavaScript/JSON.sublime-syntax
FichteFoll: `json` must be the first extension because the first will be added as the default extension when saving a file under a new name with this syntax selected.

JavaScript/JSON.sublime-syntax
FichteFoll: This file is supposed to be empty. It doesn't deserve to be here.

FichteFoll:
I can't dismiss my review for some reason, but this looks fine now.

ShellScript/Shell-Unix-Generic.sublime-syntax
```diff
@@ -13,6 +13,10 @@ file_extensions:
   - .bash_profile
   - .bash_variables
   - .bashrc
+  - .eslintignore
+  - .npmignore
```
wbond: These don't appear to be shell scripts, but rather a glob-like syntax. I don't think we should include them.

PR #879, opened 2017-04-05
[Python] Match unpacking operators

### LINKED ISSUE #773: [Python] * and ** in method definitions are unscoped
https://docs.python.org/dev/tutorial/controlflow.html#keyword-arguments

    def example(test, *args, **kwargs): # <- scope my asterisks please!
        print(test)
        print(*args)
        print(list(range(*args)))
        print(*kwargs)
        blah = { 'def': 2, 'testing123': 'TEST' }
        combined = { **kwargs, **blah }
        print(combined)
    example('hello world', 3, 6, testing123='test', abc=1)

FichteFoll:
Follow-up on https://github.com/sublimehq/Packages/pull/743. Since this is also runs into https://github.com/SublimeTextIssues/Core/issues/1190, please confirm locally that the tests pass as I can't.

Fixes #773.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -633,19 +633,24 @@ contexts:
   lambda:
     - match: \b(lambda)(?=\s|:|$)
       scope: storage.type.function.inline.python
-      push:
-        - meta_scope: meta.function.inline.python
-        - meta_content_scope: meta.function.inline.parameters.python
-        - include: line-continuation-or-pop
-        - match: '\:'
-          scope: punctuation.section.function.begin.python
-          pop: true
-        - match: ','
-          scope: punctuation.separator.parameters.python
-        - include: keyword-arguments
-        - include: illegal-names
-        - match: '{{identifier}}'
-          scope: variable.parameter.python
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -395,6 +391,15 @@ contexts:
     - match: '@'
       scope: keyword.operator.matrix.python
 
+  allow-unpack-operators:
+    # Match unpacking operators, if present
+    - match: \*
+      scope: keyword.operator.unpacking.python
+    - match: \*\*
+      scope: keyword.operator.unpacking.python
```
keith-hall: This pattern will never be matched due to the single `*` pattern coming before it.

keith-hall: Currently, 3 or more stars in a row will be scoped as if valid, is that what we want?

keith-hall: As will stars with spaces between them, which I don't think is valid.

FichteFoll: Nice spot. All fixed.

wbond:
Tests passed, thanks!

Issue #882, opened 2017-04-05
[XML] xml-stylesheet processing instruction marked as invalid


keith-hall:
https://www.w3.org/TR/xml-stylesheet/#the-xml-stylesheet-processing-instruction

For example, on Windows, IIS' Failed Request Trace Logs are in XML format and start with:

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<?xml-stylesheet type='text/xsl' href='freb.xsl'?>
<!-- saved from url=(0014)about:internet -->
```

but ST currently scopes the `<` before `?xml-stylesheet` as `text.xml invalid.illegal.missing-entity.xml` and doesn't scope the rest of the processing instruction (though the rest of the document is correctly highlighted).

This was probably accidentally broken by https://github.com/sublimehq/Packages/pull/677

PR #883, opened 2017-04-05
[XML] correctly scope xml-stylesheet instructions

keith-hall:
fixes https://github.com/sublimehq/Packages/issues/882


wbond:
How about matching `xml-stylesheet` separately so we don’t inadvertantly add `tag-stuff` to other preprocessor elements?

keith-hall:
done :)

wbond:
Great!

PR #890, opened 2017-04-10
[Java] Lots of minor improvements

ismell:
* Fixed a lot of generic issues.
* Support multiline methods. This required removing the storage modifiers and
  return type from the meta.method scope.
* Display illegal trailing punctuation.

I've had a lot of these improvements locally for a while so the rebase was a little
painful. I had to remove/fix some strange tests.


wbond:
Those tests would definitely need to be preserved

ismell:
I added back the missing annotation changes I clobbered. I did make a few changes though.

wbond:
`variable.annotation` is the standard we decided upon in #709. Color schemes can choose to highlight it the same as `support.class` if they want – Monokai will styling it very similarly, but without italics.

ismell:
`variable.annotation` seems really strange in Java though. It is a `support.class` because it needs to be declared in the import section like any other `support.class`. `variable.` is something that the user declares that can be assigned.

Just pushed fixes for instantiation as well.

ismell:
if themes wanted to theme it they could do `meta.annotation support.class`.

wbond:
I realize that in certain syntaxes the names applied to various components won't be a 100% match. Instead, we are trying to be consistent for the sake of color scheme authors and end-users. A color scheme author can choose that an annotation name be styled the same way in Java as a class. For everyone else who wants annotation-specific styles, the new scope names make a more consistent experience out-of-the-box when working in multiple languages.

ismell:
That's fine I can change it.

    @com.java.FooBar
    //^^^^^^^^^^^^^^ variable.annotation.java
    //       ^ punctuation.accessor.dot.java

Does that look correct?

wbond:
That exact issue isn't currently resolved. #737 discusses the options here, but if I recall, there isn't too much of a consensus.

ismell:
I'll go with it. Makes it easier to extract the full class path. Fix pushed.

ismell:
Is there anything I can do to help this PR along? Also you can push commits to the branch if you want.

wbond:
Mostly I just need time to go through it – been working on bug fixes on core mostly to resolve issues from all the work that went into 3127

Java/Java.sublime-syntax
```diff
@@ -372,25 +330,14 @@ contexts:
...
-    - match: \b(try|catch|finally|throw)\b
-      scope: keyword.control.catch-exception.java
     - match: '\?|:'
       scope: keyword.control.java
-    - match: \b(return|break|case|continue|default|do|while|for|switch|if|else)\b
-      scope: keyword.control.java
     - match: \b(instanceof)\b
       scope: keyword.operator.java
-    - match: (==|!=|<=|>=|<>|<|>)
+    - match: (==|!=|<=|>=|<>|(?<!\.)<|>)
```
FichteFoll: This is a look-behind pattern and incompatible with the internal sregex engine, degrading performance by *a lot*. You should be able to find a way without needing a look-behind.

Java/Java.sublime-syntax
```diff
@@ -230,89 +180,98 @@ contexts:
...
-    - include: class
+    - include: constants-and-special-vars
+    - include: keywords
+    - include: assignment
+    - include: lambdas
+    - include: strings
+    - include: anonymous-classes-and-new
+    - include: all-types
+    - include: assertions
+    - match: (?<=\.)<
```
FichteFoll: Another look-behind.

Java/Java.sublime-syntax
```diff
@@ -598,20 +573,56 @@ contexts:
 
   parens:
     - match: \(
+      scope: punctuation.section.parens.begin.java
       push:
+        - meta_scope: meta.parens.java
```
FichteFoll: Afaik the recommendation is to use `punctuation.section.group`, according to the scope naming guidelines. (See also punctuation scopes.)

Java/Java.sublime-syntax
```diff
@@ -372,25 +330,14 @@ contexts:
...
       scope: keyword.operator.comparison.java
-    - match: (=)
-      scope: keyword.operator.assignment.java
-      push:
-        - meta_scope: meta.assignment.rhs.java
-        - match: (?=(;|\)))
-          pop: true
-        - include: code
+    # - match: (=)
+    #   scope: keyword.operator.assignment.java standalone
```
FichteFoll: I don't think we need these commented lines anymore.

Java/Java.sublime-syntax
ismell: Removed all the look behinds.

Java/Java.sublime-syntax
ismell: It says `based on appropriate semantics`. What's the difference between group and parens? Java doesn't have a concept of a `group` so parens seemed appropriate.

ismell: Thanks for the review

FichteFoll: I'm not quite sure, tbh. Python also uses `group`, where contents in parentheses are just evaluated before others, pretty much like Java.

Maybe @wbond can clarify.

Java/Java.sublime-syntax
```diff
@@ -230,89 +180,98 @@ contexts:
...
   class-body:
+    - include: annotations
     - include: class
+    - include: static-assignment
     - include: enums
+    - include: generic-declaration # for methods
     - include: methods
-    - include: annotations
+    # mark constants as constant
+    - match: '(\.)?\b(?!UUID\b|URI\b)([A-Z][A-Z0-9_]+)(?!<|\.class|\s*\w+\s*=)\b'
```
ismell: Yeah they are present in the standard java API. I should extract that out into a variable. I should also add tests. 
They break the standard class naming convention. Without this they get picked up as constants.

Java/syntax_test_java.java
```diff
@@ -201,61 +222,97 @@ public void test2(Type) abc
...
+//           ^ punctuation.definition.generic.end.java
+
+  new Foo<?>();
+//        ^ keyword.operator.wildcard.java
+
+  new Foo<? extends Bar, String>();
+//        ^ keyword.operator.wildcard.java
+//                  ^^^ support.class.java
+//          ^^^^^^^ keyword.declaration.extends.java
+//                     ^ punctuation.section.generic.separator.java
```
wbond: I think this should probably be `punctuation.separator` or some specialization of that.

Java/syntax_test_java.java
```diff
@@ -312,3 +371,502 @@ void test2() {
...
+  };
+
+  public abstract void doSomething();
+//                     ^^^^^^^^^^^^^ meta.method.java
+}
+
+public final class SomeClass<V extends OtherClass, T> extends BaseClass<V> {
+//                          ^ punctuation.definition.generic.begin.java
+//                                                  ^ punctuation.definition.generic.end.java
+//                                               ^ punctuation.section.generic.separator.java
```
wbond: `punctuation.separator`

Java/syntax_test_java.java
```diff
@@ -312,3 +371,502 @@ void test2() {
 //            ^ support.class
     }
 }
+
+public enum FooEnum {
+//     ^^^^ storage.type.java
+  FOO;
+//^^^ constant.other.enum
+}
+// <- meta.class.java meta.class.body.java punctuation.section.class.end.java
```
wbond: Generally we don't do things like `punctuation.section.class`, but instead `punctuation.section.block`. Then the color scheme author can target class braces via `meta.class punctuation.section.block`.

wbond: When possible in more recent syntax changes, we are trying not to "stack" scopes. So instead of having `meta.class.java meta.class.body.java` we have `public enum FooEnum ` as `meta.class.java` and `{…}` as `meta.class.body.java`. Does that make sense? It can take some extra context commands, but in many languages it can help prevent really deep scope stacks.

Java/syntax_test_java.java
```diff
@@ -201,61 +222,97 @@ public void test2(Type) abc
...
+     this.<B> foo();
+//            ^^^ variable.function.java
+//        ^^^ meta.generic.java
+//        ^ punctuation.definition.generic.begin.java
+//         ^ support.class.java
+
+     Function<String, Integer> func = a -> 42;
+//                                  ^^^^^^^^^ meta.assignment.rhs.java
+//                                    ^ variable.parameter.java
+//                                      ^^ storage.type.lambda.java
```
ismell: Done

Java/syntax_test_java.java
```diff
@@ -201,61 +222,97 @@ public void test2(Type) abc
...
 
-new Foo<int>();
-//      ^^^ invalid.illegal.primitive-instantiation.java
+     a -> { return 42; };
+//        ^^^^^^^^^^^^^^ meta.lambda.body.java
+
+     (a, b) -> 42;
+//    ^ variable.parameter.java
+//       ^ variable.parameter.java
+//          ^^ storage.type.lambda.java
```
ismell: Done

Java/syntax_test_java.java
```diff
@@ -201,61 +222,97 @@ public void test2(Type) abc
...
+//                                    ^ variable.parameter.java
+//                                      ^^ storage.type.lambda.java
+//                                         ^^ constant.numeric
+//                                           ^ punctuation.terminator.java
+     foo(a -> 42);
+//   ^^^^^^^^^^^^ meta.function-call.java
+//   ^^^ variable.function.java
+//      ^ punctuation.section.parens.begin.java
+//       ^ variable.parameter.java
+//         ^^ storage.type.lambda.java
```
ismell: Done

Java/syntax_test_java.java
```diff
@@ -201,61 +222,97 @@ public void test2(Type) abc
...
+
+     (int a, Foo<Integer>[] b) -> 42;
+//    ^^^ storage.type.primitive
+//        ^ variable.parameter.java
+//           ^^^ support.class.java
+//              ^ punctuation.definition.generic.begin.java
+//               ^^^^^^^ support.class.java
+//                      ^ punctuation.definition.generic.end.java
+//                          ^ variable.parameter.java
+//                             ^^ storage.type.lambda.java
```
ismell: Done

Java/syntax_test_java.java
ismell: Done

Java/syntax_test_java.java
```diff
@@ -687,7 +687,7 @@ public void someMethod(WithParam foo) throws Exception {
 //                                  ^ constant.numeric.java
 
     foo.forEach((k, v) -> {
-//                     ^ storage.type.lambda.java
+//                     ^ storage.type.function.lambda.java
```
rwols: storage.type.function.arrow seems to be common (e.g. Scala, Javascript)

Java/syntax_test_java.java
ismell: It makes sense, but I have a concern.
I have a plugin that takes the current cursor position and gives you the method and class name that the cursor is in. It does this by looking at all `meta.method` scopes and seeing if the cursor is in one of the scopes. It then looks for `entity.name.function` in the matching scope. It then does the same thing for the class and class name.

If `meta.method` only contains `entity.name.function` and not the `meta.method.body`, then I'm not sure how to find out the method name for a specific `meta.method.body`.

Java/syntax_test_java.java
ismell: Done

Java/syntax_test_java.java
ismell: wbond@ which do you prefer?

Java/Java.sublime-syntax
```diff
@@ -400,130 +346,175 @@ contexts:
...
+        2: punctuation.accessor.dot.java
       push:
-        - match: '<'
-          scope: punctuation.definition.generic.begin.java
-          set:
-            - meta_scope: meta.generic.java
-            - match: '>'
-              scope: punctuation.definition.generic.end.java
+        - match: \[
+          scope: keyword.operator.array.java
```
rwols: Should this not be a punctuation.section.brackets.begin ...

Java/Java.sublime-syntax
```diff
@@ -400,130 +346,175 @@ contexts:
...
+          push:
+            - meta_scope: meta.brackets.java
+            - match: \]
+              scope: keyword.operator.array.java
```
rwols: ... and this a Should this not be a punctuation.section.brackets.end?

Java/Java.sublime-syntax
```diff
@@ -400,130 +346,175 @@ contexts:
...
         - meta_scope: meta.generic.java
-        - match: '>'
-          scope: punctuation.definition.generic.end.java
-          pop: true
-        - include: generic-instantiation
-    - match: '\.\.\.'
-      scope: keyword.operator.varargs.java
-    - match: '(\[)\s*(\])'
+        - match: <
+          scope: punctuation.definition.generic.begin.java
```
rwols: Should this not be a punctuation.section.generic.begin ...

Java/Java.sublime-syntax
```diff
@@ -400,130 +346,175 @@ contexts:
...
-    - match: '{{id}}'
-      scope: variable.parameter.java
-      set: method-parameters
-    - match: (?=\S)
+        1: entity.other.inherited-class.java
+        2: punctuation.accessor.dot.java
+  object-type-generic:
+    - meta_scope: meta.generic.java
+    - match: ">"
+      scope: punctuation.definition.generic.end.java
```
rwols: ... and this a punctuation.section.generic.end ?

Java/Java.sublime-syntax
```diff
@@ -540,12 +531,7 @@ contexts:
   raw-parameter-after:
     - match: '<'
       scope: punctuation.definition.generic.begin.java
```
rwols: (see above comment)

Java/Java.sublime-syntax
```diff
@@ -598,20 +584,56 @@ contexts:
...
+  assignment:
+    - match: \=
+      scope: keyword.operator.assignment.java
+      push:
+        - meta_scope: meta.assignment.rhs.java
+        - match: '(?=;|\)|\}|,)'
+          pop: true
+        - include: code
+  static-code-block:
+    - match: "{"
```
rwols: Why is this `{` not scoped?

ismell: Done

Java/Java.sublime-syntax
ismell: Same

Java/Java.sublime-syntax
ismell: Same as above

Java/Java.sublime-syntax
ismell: Looks like the style guide recommends `punctuation.definition.generic.begin`

Java/Java.sublime-syntax
ismell: Done

Java/Java.sublime-syntax
ismell: Done

Java/Java.sublime-syntax
rwols: Interesting 👍 it's punctuation.section.generic.begin in C++, so, either C++ is wrong and should be updated, or the style guide is wrong.

Java/syntax_test_java.java
FichteFoll: `lambda` seems to be the more general term and isn't language specific. I favor going for that instead of "arrow", which relates to the syntax more than the actual construct.

rwols: I agree @FichteFoll. This construct (lambda notation) should be added to the style guide.

Java/syntax_test_java.java
```diff
@@ -88,15 +92,15 @@ public static void main(String... args) {
             System.out.println(i);
         }
     }
-
+//  ^ meta.method.java meta.method.body.java punctuation.definition.method.end.java
```
wbond: This should be testing for `punctuation.section.block.end.java`

Java/syntax_test_java.java
```diff
@@ -42,13 +45,14 @@
 //                                                     ^ punctuation.terminator.java
 
     public static void main(String... args) {
```
wbond: The `{` should be tested for `punctuation.section.block.begin.java`

Java/syntax_test_java.java
```diff
@@ -16,6 +16,9 @@
 //            ^ meta.import.java support.function.import.java
 //                               ^ meta.import.java punctuation.terminator.java
 
+import static a.b.Class.CONSTANT;
```
wbond: Let's add some tests here for `punctuation.accessor`

Java/syntax_test_java.java
```diff
@@ -42,13 +45,14 @@
...
-//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.method
+//                     ^^^^^^^^^^^^^^^^^^^^^^^ meta.method
 //  ^^^^^^ storage.modifier.java
 //         ^^^^^^ storage.modifier.java
-//                ^^^^ meta.method.return-type storage.type
+//                ^^^^ storage.type
 //                     ^^^^ meta.method.identifier.java entity.name.function.java
 //                         ^^^^^^^^^^^^^^^^ meta.method.parameters.java
 //                         ^ punctuation.section.parens.begin.java
+//                                ^^^ storage.modifier.variable-arity
```
wbond: In other syntaxes this is scoped as `keyword.operator.variadic` (C/C++/Obj-C/Obj-C++) or `keyword.operator` (Rust).

Java/Java.sublime-syntax
wbond: The style guide is correct. The idea being that `section` is used for a complex area that can have many elements and all sorts of nesting things. Like arrays/lists or blocks. `definition` is supposed to be for more atomic things. Generic types are sort of a grey area, but they feel like they should be treated as more of an atomic thing than a "section".

Java/Java.sublime-syntax
wbond: I don't feel strongly about this. There are so many syntaxes in the wild that use `parens` that we need to recommend they are handled by any decent color scheme. Some syntaxes parents are used where it definitely is not a "group". Similarly `brackets`.

I am leaving it up to individual syntax authors for the most part since I don't feel strongly and I'd rather spend my time on more meaningful things. 😄

Java/syntax_test_java.java
```diff
@@ -122,9 +126,10 @@ private static void anotherMethod() throws MyException<Abc> {
     }
 
     <T> void save(T obj);
-//  ^^^^^^^^^^^^^^^^^^^^ meta.method
+//           ^^^^^^^^^^^ meta.method
 //  ^^^ meta.generic
-//      ^^^^ meta.method.return-type
+//   ^  variable.parameter.type
```
wbond: This would be a new addition to our scope guidelines. I don't know if for C++, Rust, Java it is clear in what circumstances this scope would be applied. I am certainly no expert in the finer points of templating, so some one else may have more insight or feel more strongly about how this should be handled.

EDIT: Just saw a comment below from myself from back on Jun 8 about an example father down. 😄

Java/syntax_test_java.java
```diff
@@ -201,61 +222,97 @@ public void test2(Type) abc
...
+//       ^ variable.parameter.java
+//         ^^ storage.type.function.lambda.java
+//            ^^ constant.numeric
+//              ^ punctuation.section.parens.end.java
+//               ^ punctuation.terminator.java
 
-new Foo<int>();
-//      ^^^ invalid.illegal.primitive-instantiation.java
+     a -> { return 42; };
+//        ^^^^^^^^^^^^^^ meta.lambda.body.java
```
wbond: For consistency with other syntax, I think `meta.lambda` (and related scopes) should become `meta.function.anonymous`.

Java/syntax_test_java.java
```diff
@@ -201,61 +222,97 @@ public void test2(Type) abc
...
+
+     (a, b) -> 42;
+//    ^ variable.parameter.java
+//       ^ variable.parameter.java
+//          ^^ storage.type.function.lambda.java
```
wbond: This isn't a change needed here, but we do have some inconsistency in the default packages for this scope. I think Scala, Python and JavaScript should be updated to `storage.type.function.lambda` instead of `arrow`/`inline`.

FichteFoll: Why use `lambda` in `storage.type.function.lambda` when we use `anonymous` in `meta.function.anonymous`?

Java/Java.sublime-syntax
```diff
@@ -588,9 +601,22 @@ contexts:
...
+    - match: \(
+      scope: punctuation.section.parens.begin.java
+      set:
+        - meta_scope: meta.parens.java
+        - match: \)
+          scope: punctuation.section.parens.end.java
           pop: true
+        - include: illegal-open-block
         - include: code
+    - include: any_POP
```
rwols: Maybe call this context `else-pop` instead of `any_POP`?

Java/syntax_test_java.java
wbond: I am cool with `storage.type.function.anonymous`

wbond:
@ismell I'd love to get this included in one of the next dev builds since it has some great improvements. Do you think you'll have time to make some of the tweaks I suggested?

ismell:
I'll try to make some time this week or early next week.

Java/syntax_test_java.java
ismell: Went with keyword.operator.variadic

Java/syntax_test_java.java
ismell: Done

Java/syntax_test_java.java
ismell: Done

Java/syntax_test_java.java
ismell: Done

Java/syntax_test_java.java
ismell: Done

Java/syntax_test_java.java
ismell: Done. It only matches the last dot due to the limitations of the pattern.

ismell:
I pushed a lot of changes. Just have to say githubs code review UI is pretty bad.

wbond:
> Just have to say githubs code review UI is pretty bad.

Yes, when I used gerrit it was much more conducive to in-depth reviews. Unfortunately it wasn't very usable/discoverable. Maybe someday we'll have the best of both worlds!

ismell:
I should have squashed before you merged :)

Issue #889, opened 2017-04-10
[JavaScript] (PR provided) Cruft causing incorrect highlighting for standard JavaScript: Mustache templating, HTML comments, IE conditional compilation


Thom1729:
The following tests fail:

```js
{{ 42; }}
// ^^ meta.block.js meta.block.js constant.numeric.js

x --> y;
//^^ keyword.operator.arithmetic.js
//  ^ keyword.operator.relational.js

/*@if /*/
//     ^^ punctuation.definition.comment.js

z
// <- variable.other.readwrite.js
```

Each test is valid JavaScript, but the parsing is broken by code that handles non-JavaScript syntax:

- Highlighting for mustache templates conflicts with JavaScript blocks. This is fundamentally unfixable; the two languages conflict. If we want to maintain mustache embedding in JavaScript code for some reason, we can make a new syntax for it (simply including JavaScript with a `with_prototype`). Or, if it's not worth having a separate syntax for, it's probably not worth breaking the JavaScript syntax for.
- Highlighting for HTML comments conflicts with JavaScript operators. This could conceivably be patched, because `<!--` cannot conflict with real JavaScript at top level. But the implementation seems rather pointless to begin with, because it *can't* cover all uses of HTML comments (e.g. in strings) without breaking JavaScript and the HTML syntax should handle this anyway using `with_prototype`.
- Highlighting for ancient IE conditional compilation breaks JavaScript comments. This is an ill-advised and archaic practice that has long been abandoned by Microsoft, so it should be safe to just cut it.

Fixing these will help to clear the way for #885 and related issues.

Thom1729:
I'm closing this. The three bugs described here are addressed in more detail elsewhere.

Mustache templates: #927  
HTML comments: #928  
IE conditional compilation: TBD

PR #891, opened 2017-04-10
[JavaScript] Removed Mustache syntax, HTML comments, and IE conditional compilation.

Thom1729:
For #889. Replaces tests for removed features with tests for JavaScript syntax. All tests pass.

First GitHub pull request; please let me know if it is defective.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -21,6 +21,9 @@ contexts:
     - include: keywords-top-level
     - include: statements
 
+  prototype:
+    - include: comments
+
   keywords-top-level:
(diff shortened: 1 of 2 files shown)


wbond:
Removing mustache should be fine. I believe the HTML comments and IE conditional compilation are for the sake of when the JavaScript syntax is embedded in HTML.

We'll probably need to do some testing, and perhaps adds tests to the HTML syntax to make sure we don't lose the functionality. It may not be possible to fully extract these features from JavaScript and place them into HTML without resorting to `with_prototype`, which causes its own issues.

Thom1729:
Because I put this all in one PR, it's probably best that I close this one and redo it separately for each change.

PR #895, opened 2017-04-11
[C] Documentation blocks

rwols:
This is an attempt to add ~doxygen~ documentation blocks/lines to C ~and C++~. Consequently, documentation blocks are enabled for C++, Objective-C and Objective-C++.

Currently:

~Scopes `///`, `//!`, `///<` and `//!<` as `comment.line.documentation.doxygen`~

~Scopes `/** ... */` as `comment.block.documentation.doxygen`. Note the extra `*`.~

Assigns `comment.line.documentation` to `///`, `//!`, `///<` and `//!<`.
Assigns `comment.block.documentation` to `/** ... */`.

Note that *only if* a comment block starts with precisely two asterisks, then it's considered the start of a documentation block. I've added a test case for this in `syntax_test_doxygen.c`.

Anything of the form `@command`, `@command`, `@command[arg]` or `@command{arg}` is scoped as a `meta.function-call` within a documentation block (and various punctuation scopes). The backslash is interchangeable with the `@` symbol.

The special command `\f$` starts a `meta.environment.math.inline.dollar.latex` scope, ended by another `\f$`.

The special command `\f[` starts a `meta.environment.math.block.bracket.latex` scope, ended by `\f]`.

The special command `\f{environment}{` starts a `meta.environment.math.block.bracket.latex`, ended by `\f}`.

~For the special `\f` commands, it would be possible to include the `math-content` rule from LaTeX, but in practice I feel that this makes the comment block stand out too much from the actual source code. Perhaps color scheme authors can do something about this.~

[C#-style documentation constructs for Visual Studio](https://msdn.microsoft.com/en-us/library/ms177227.aspx) are also supported.


wbond:
I'm not super-familiar with automated doc generation in C/C++. Are there many alternatives to Doxygen? Would we be preventing adding support for another format by adding this?

For the time being I'd be hesitant to add changes to C that includes things like LaTeX, since that is a rather large syntax. We are starting to run into issues with syntaxes including other and the number of contexts exploding, resulting in very large cache files that take a long time to read back into memory.

rwols:
Thanks for the feedback.

> Are there many alternatives to Doxygen? Would we be preventing adding support for another format by adding this?

Sphinx: Extracts the documentation via a Doxygen plugin. [Here is an example of that happening](https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L250-L263). But note that the documentation block still starts with `/**` and ends with `*/`.

cldocs: Treats any ordinary comment blocks and lines as documentation, so Doxygen syntax doesn't stand in its way.

doc++: Uses a subset of Doxygen's syntax (namely `/** ... */` and `/// ...`)

I [do](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/cc/framework/ops.h#L29) [believe](https://clang.llvm.org/doxygen/SourceManager_8h_source.html) [that](https://github.com/tesseract-ocr/tesseract/blob/master/ccmain/control.h#L20-L23) [Doxygen](https://github.com/apple/swift/blob/master/include/swift/Basic/BlotSetVector.h#L54) [is](https://github.com/bitcoin/bitcoin/blob/master/src/chain.h#L431) [the](https://github.com/godotengine/godot/blob/master/core/dvector.h#L79-L81) [dominant](https://github.com/v8/v8/blob/master/include/v8-debug.h#L38-L40) docgenerator.

Some smaller players are [NaturalDocs](http://www.naturaldocs.org/documenting/walkthrough.html#OurFirstFunction), which uses ordinary comment blocks and lines, and [DoxyPress](http://www.copperspice.com/documentation-doxypress.html), which seems to be a fork of Doxygen, and uses Doxygen's syntax. [Here is a stack overflow thread about Doxygen alternatives](http://stackoverflow.com/questions/2323074/doxygen-alternative-for-c).

> For the time being I'd be hesitant to add changes to C that includes things like LaTeX, since that is a rather large syntax.

Agreed. I played around with including the latex scope, but that suddenly skyrocketed the compilation time of the syntax definition. Also, the colors are kind of overwhelming, so I think it's best to leave it out for now.

rwols:
OK, I did some more research and it looks like Visual Studio allows one to use C#-style documentation blocks. So, things like

```cpp
/// <summary>Does a thing.</summary>
/// <param name="d">A number</param>
/// <return>a thing.</return>
int doSomeThing(double d);
```
I think this is significant, and we'd be closing the door to users that use this XML-style if we commit to the current syntax. On the other hand, I think it's feasible to do both styles in one syntax definition. I'm going to research how it's done in the C# syntax def.

rwols:
This PR has gotten too large for its own good. I will push my documentation agenda in smaller, more digestable chunks.

PR #897, opened 2017-04-12
[Markdown] fix multiline list items with blank line separators

### LINKED ISSUE #896: [Markdown] Incorrectly scoped list items for lists with blank lines
When using the Markdown syntax from Dev build 3127, the text of list items that falls on lines after a list bullet appears to be incorrectly scoped as raw text if there are blank lines in the list.

As an example, given the following markdown:

```markdown
  * List Item 1
    Text under Item 1
  * List Item 2
    Text under Item 2

  * List Item 3
    Text under Item 3
```

Both "Text under Item 1" and "Text under Item 2" are scoped as `text.html.markdown markup.list.unnumbered.markdown meta.paragraph.list.markdown` while the text "Text under Item 3" is scoped as `text.html.markdown markup.list.unnumbered.markdown markup.raw.block.markdown` instead.

Removing the blank line between the items (which is scoped as `text.html.markdown markup.list.unnumbered.markdown`) resolves the problem.

keith-hall:
fixes #896

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -134,8 +134,6 @@ contexts:
         - meta_content_scope: markup.list.unnumbered.markdown
         - match: ^(?=\S)
           pop: true
-        - match: (?={{list_item}})
-          push: list-content
         - include: list-paragraph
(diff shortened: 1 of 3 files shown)


wbond:
Thanks!

Issue #899, opened 2017-04-13
[C++] static_assert syntax tests


rwols:
The following test fails, but is legal C++11.

```c++
template <typename AllocatorT = MallocAllocator, size_t SlabSize = 4096,
          size_t SizeThreshold = SlabSize>
class BumpPtrAllocatorImpl
    : public AllocatorBase<
          BumpPtrAllocatorImpl<AllocatorT, SlabSize, SizeThreshold>> {
                                                                  /* ^ punctuation.section.block.begin */
public:
  static_assert(SizeThreshold <= SlabSize,
            /* ^ punctuation.section.group.begin */
                "The SizeThreshold must be at most the SlabSize to ensure "
                "that objects larger than a slab go into their own memory "
                "allocation.");
             /* ^ string.quoted.double punctuation.definition.string.begin */
             /*             ^ string.quoted.double punctuation.definition.string.end */
             /*              ^ punctuation.section.group.end - invalid.illegal.stray-bracket-end */
};
/* <- punctuation.section.block.end - invalid.illegal */
```
![schermafbeelding 2017-04-13 om 12 22 21](https://cloud.githubusercontent.com/assets/2431823/25001055/e83449fc-2043-11e7-9904-3ca3d2697a10.png)

I'm on Sublime Build 3128.

Note: [from here](https://github.com/llvm-mirror/llvm/blob/master/include/llvm/Support/Allocator.h#L122-L371).

rwols:
I've brought it down to a minimal test case:

```c++
class A
{
public:
  static_assert(0 < 1, "");
};
/* <- punctuation.section.block.end - invalid.illegal */
```

It seems to trip on the `<` operator, because this works fine:

```c++
class A
{
public:
  static_assert(true, "");
};
/* <- punctuation.section.block.end - invalid.illegal */
```

rwols:
And here is another test case that happens to fail:

```c++
struct A { A() = default; };
                       /* ^ punctuation.section.block.end - invalid.illegal */
```

rwols:
This succeeds:
```c++
void f()
{
    static_assert(false, "asdf");
    /* ^ keyword.operator.word */
}
```
While this one fails:
```c++
void f()
{
    #ifdef X
    static_assert(false, "asdf");
    /* ^ keyword.operator.word */
    #endif
}
```

PR #905, opened 2017-04-17
[C++] Fix static assert decl

rwols:
This is an attempt to fix https://github.com/sublimehq/Packages/issues/899.

Currently, https://github.com/sublimehq/Packages/issues/899#issuecomment-294446880 still fails, but https://github.com/sublimehq/Packages/issues/899#issuecomment-293856070 is fixed.


rwols:
beebf3f fixes https://github.com/sublimehq/Packages/issues/899#issuecomment-294446880.

wbond:
It feels odd for `alignof|typeid|static_assert` to push into `data-structures-modifier`. Perhaps we need to clean this up into a new named scope. I can't imagine `friend` is valid inside of any of those.

Additionally, Objective-C++ will need the tests (and possibly changes) ported over.

rwols:
Good point about ObjC++, will have to check if nothing is breaking there.

wbond:
It looks like you missed porting the first test to Objective-C++ and the `keywords-parens` context in Objective-C++ wasn't modified with the changes.

wbond:
Thanks!

PR #906, opened 2017-04-17
[Rails] fix snippet scope miss-matches

tadast:
Looks like the [Ruby on Rails.sublime-syntax file](https://github.com/sublimehq/Packages/blob/6ed5bc862b5b12c628bfeb4b9a98c7535b91dc18/Rails/Ruby%20on%20Rails.sublime-syntax) defines scopes in the format of
`meta.<scope1>.<scope2>.rails` whereas most of the snippets were following
`meta.rails.<scope1>.<scope2>`, making them essentially useless as they would
never trigger.

PS.: I have a repo (and a package) with more up-to date Rails snippets https://github.com/tadast/sublime-rails-snippets. I'll try to port over the changes via small PRs when I have time. If anyone looking at this feels like helping, please do.


wbond:
Yes, it appears I broke these with the changes in fb406373119d9fe8b301e55b8369bfd7e0aef128. Thanks for the fix!

PR #910, opened 2017-04-18
[C#] fix symbol list

### LINKED ISSUE #909: [C#] Goto Definition broken for classes, missing method information
Hey Keith,
Creating a ticket for you here as you requested. It seems you've been able to repro:

"looking at the symbol list for the main C# syntax test file, it seems like classes don't appear in Goto Definition at all, and methods appear without the return type details"

https://forum.sublimetext.com/t/3129-build-go-to-definition-broken-for-c/27553/2

keith-hall:
fixes #909

- adds Goto Definition support for classes, interfaces and enums by adding them to the global symbol list
- as the return type or visibility of methods are not part of the `meta.method` scope, it shows only the method name in the symbol list, indented under the class/interface name. Parameters are shown for constructors, but only in the local symbol list, as this is where they are more useful.
- namespaces are shown in the symbol list, but are not added to the global symbol list, because many files can reference the same namespace, I don't think it's useful to be able to use Goto Definition on them - Goto Definition is more useful for classes and methods. If people want it, then we can consider changing this later.
- regions are shown in the symbol list, but are not added to the global symbol list, because they are never referenced from code - they are just useful for separating code, so having Goto Definition support on them would actually be counterproductive.


wbond:
Thank you, as usual!

robotron2084:
Thank you as well, Keith.

C#/Symbol List Classes.tmPreferences
```diff
@@ -4,11 +4,21 @@
...
 		<integer>1</integer>
+		<key>showInIndexedSymbolList</key>
+		<string>1</string>
+		<key>symbolTransformation</key>
+		<string>
+		s/^/class /;
+		</string>
+		<key>symbolIndexTransformation</key>
+		<string>
+		s/^/class /;
```
FichteFoll: This schema (`symbolIndexTransformation` adding something to the indexed symbol) will prevent Go To Definition from functioning if applied on a class/method/enum name. It's fine for the local symbol list opened with ctrl+r, but the project symbol list and go to definition both use the indexed symbols, as far as I'm aware.

keith-hall: It seems that you're right, thanks @FichteFoll - when I tested this I must have not waited for my project to be reindexed, because it seemed to work. I must admit that I thought the transformations were only used for display purposes, not lookup purposes as well. But I've proved it with `window.lookup_symbol_in_index` and at the moment, it doesn't find the class name alone, but does find `class ClassName`... I'll make another PR to remove the `symbolIndexTransformation`s.

PR #912, opened 2017-04-19
[Rails] Drop legacy snippets

tadast:
Delete RJS-related snippets – not a thing since Rails 3.1

Delete `find(:all)` and `find(:first)` – dropped in Rails 3.? in favour of .all and .first


wbond:
It seems rails 3.2 was EOL back in 2016, so we can probably drop these.

PR #913, opened 2017-04-19
[Rails] Update before/after validation on create/update

tadast:
This syntax has been dropped in Rails 3 (2010) and replaced by the one
suggested in this commit.

~~Removes these four methods from the `support.function.activerecord.rails` match as well.~~

https://apidock.com/rails/ActiveRecord/Callbacks/before_validation_on_create
https://apidock.com/rails/ActiveRecord/Callbacks/before_validation_on_update
https://apidock.com/rails/ActiveRecord/Callbacks/after_validation_on_update
https://apidock.com/rails/ActiveRecord/Callbacks/after_validation_on_create

~~Note: the suggested changes in snippets use the "new" hash syntax `on: :create` instead of `:on => :create`. The "new" syntax was introduced in ruby 1.9, [as of now the oldest supported version of ruby is 2.2](https://www.ruby-lang.org/en/).~~

Question to the maintainers: I could open another PR converting all the existing snippets to the new syntax if you'd like that.

Edits: I kept the matches unchanged for backwards-compatibility and didn't enforce the new ruby hash syntax in this change. Crossed out irrelevant parts of the PR description.

### DIFF
--- a/Rails/Snippets/after_validation_on_create.sublime-snippet
+++ b/Rails/Snippets/after_validation_on_create.sublime-snippet
@@ -1,6 +1,6 @@
 <snippet>
-	<content><![CDATA[after_validation_on_create ]]></content>
+	<content><![CDATA[after_validation :${1:attr}, :on => :create]]></content>
 	<tabTrigger>aftvoc</tabTrigger>
 	<scope>source.ruby.rails</scope>
-	<description>after_validation_on_create</description>
+	<description>after_validation on create</description>
 </snippet>
--- a/Rails/Snippets/after_validation_on_update.sublime-snippet
+++ b/Rails/Snippets/after_validation_on_update.sublime-snippet
@@ -1,5 +1,5 @@
 <snippet>
-	<content><![CDATA[after_validation_on_update ]]></content>
+	<content><![CDATA[after_validation :${1:attr}, :on => :update]]></content>
 	<tabTrigger>aftvou</tabTrigger>
 	<scope>source.ruby.rails</scope>
 	<description>after_validation_on_update</description>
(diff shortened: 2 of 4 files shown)


FichteFoll:
Syntax definitions should remain backwards-compatible for as long as possible. Old code bases will continue to exist and proper highlighting of them is always a good thing, even though it is deprecated.

For example, the Python syntax still highlights Python 2 perfectly fine (and will continue to do so for a long time, because it is still used in production).

tadast:
@FichteFoll thanks for your feedback. I've reverted the change to the syntax definition, it makes sense: ~5% of rails codebases could be affected [according to this](http://rails-hosting.com/2016/). 

The changes to snippets are still relevant imho, the autocomplete with this change would make it correct 95% of the cases v.s. 5% now. Note: the old `before/after_validation_on_*` syntax is not deprecated, it has been removed from Rails.

PS.: [off-topic] Do you know what would be the best way to proceed regarding the ruby hash syntax? I think ideally there could be a user setting (custom env var?) switching between the two, but that would make almost all of the Rails snippets a lot harder to maintain.

FichteFoll:
>Note: the old before/after_validation_on_* syntax is not deprecated, it has been removed from Rails.

Yeah, well, `print` and `exec` aren't keywords in Python 3 anymore, too, so they function very differently. It doesn't really matter for the point though.

>Do you know what would be the best way to proceed regarding the ruby hash syntax? I think ideally there could be a user setting (custom env var?) switching between the two, but that would make almost all of the Rails snippets a lot harder to maintain.

I can't comment on this because I don't really know Ruby, nor about its version distrubution (i.e. what people use). Normally, you do not want to have a snippet that suggests invalid syntax for your ruby version.

wbond:
Since this isn't about syntax highlighting, but snippets, I am going to merge. Considering how fast Rails moves, I think there are few users using these snippets on a Rails 2 codebase, and they can always put in place an override.

PR #917, opened 2017-04-20
[Java] meta.annotations with meta.method inline AND in method-parameters

yegeniy:
I must apologize, I have not read through #709, #737, or #890, all of which may also be covering this small tweak, but I hope that it's possible to introduce this small change without debilitating those, more comprehensive improvements.

----

I found no regressions in running `syntax_test_java.java`, nor in testing manually with my own code. 

No performance change was detected before and after the change, the result was the same:
> Syntax "Packages/Java/Java.sublime-syntax" took an average of 2.0ms over 10 runs

-----

Motivation: I first noticed a small syntax highlighting regression in some of my code, with the recent dev build of ST3:

```java
@Override public Object converse(final String utterance, final String id) {
```

### DIFF
--- a/Java/syntax_test_java.java
+++ b/Java/syntax_test_java.java
@@ -311,4 +311,17 @@ public class GrafoTest
 //      ^ storage.modifier
 //            ^ support.class
     }
+
+    @Partial @Mock(type=Grafo.class) DataLoader inline;
+//  ^^^^^^^^                          meta.annotation
+//           ^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation
+//                                   ^ support.class
+
+    @Override public int inline() {
+//  ^^^^^^^^^ meta.annotation
+//  ^ punctuation.definition.annotation
+//   ^^^^^^^^ variable.annotation
+//            ^ meta.method
+    }
+
 }
--- a/Java/syntax_test_java.java
+++ b/Java/syntax_test_java.java
@@ -324,4 +324,17 @@ public class GrafoTest
 //            ^ meta.method
     }
 
+    void annotatedArgs(@NonNull final String p1,
+//                     ^^^^^^^^ meta.annotation
+//                              ^ storage.modifier - meta.annotation
+        @Named(value = "") List<T> p2, @NonNull final String p3) {}
+//      ^^^^^^^^^^^^^^^^^^ meta.annotation
+//                         ^ support.class
+//                             ^ meta.generic punctuation.definition.generic.begin
+//                                 ^ variable.parameter
+//                                   ^ punctuation.separator
+//                                     ^^^^^^^^ meta.annotation
+//                                              ^ storage.modifier - meta.annotation
+//                                                    ^ support.class
+
 }
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -468,6 +468,7 @@ contexts:
       scope: entity.other.inherited-class.java
 
   method-parameters:
+    - include: annotations
     - match: \bfinal\b
       scope: storage.modifier.java
     - match: '{{primitives}}'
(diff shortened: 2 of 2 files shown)


Java/syntax_test_java.java
```diff
@@ -311,4 +311,17 @@ void test2() {
...
+//  ^^^^^^^^                          meta.annotation
+//           ^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation
+//                                   ^ support.class
+
+    @Override public int inline() {
+//  ^^^^^^^^^ meta.annotation
+//  ^ punctuation.definition.annotation
+//   ^^^^^^^^ variable.annotation
+//            ^ meta.method
+    }
```
yegeniy: examples based on Google Java Style Guide:

> A single parameterless annotation may instead appear together with the
> first line of the signature
> ...
> Annotations applying to a field also appear immediately after the
> documentation block, but in this case, multiple annotations (possibly
> parameterized) may be listed on the same line
-- https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations

yegeniy:
Please note that I added a second, related change to allow annotations of method arguments

Java/syntax_test_java.java
```diff
@@ -311,4 +311,30 @@ void test2() {
...
+//                              ^ storage.modifier - meta.annotation
+        @Named(value = "") List<T> p2, @NonNull final String p3) {}
+//      ^^^^^^^^^^^^^^^^^^ meta.annotation
+//                         ^ support.class
+//                             ^ meta.generic punctuation.definition.generic.begin
+//                                 ^ variable.parameter
+//                                   ^ punctuation.separator
+//                                     ^^^^^^^^ meta.annotation
+//                                              ^ storage.modifier - meta.annotation
+//                                                    ^ support.class
```
yegeniy: I also tested this manually, but I don't think it'll benefit to add it to the test suite, since the above test case is complicated enough as is:

```java
        @Named(foo = "", bar = BAZ) List<T> p2, @NonNull final String p3) {}
```

although not included here, I also manually verified that the punctuation in the method args list is detected properly even when there are commas in the annotation's parameter list.

ismell:
I think a lot of this is fixed in #890

yegeniy:
@ismell, that's true and I don't want this PR to derail that one. It seems to be a fairly comprehensive improvement; however, this changeset is two lines long to make annotations readable.

> Be as small and self-contained as is reasonable. Smaller changes are quicker to review.
https://github.com/sublimehq/Packages/blob/master/README.md#pull-requests

yegeniy:
Hi folks,  I've been using this locally, alongside #920, to some success for the past week..

Is there something I should do to get this turned down or up? Would it be better to close this out?

michaelblyons:
@yegeniy I think the Sublime guys have been busy with the Dev release. I can't speak on their behalf, but I think you should leave the PR open. I suspect someone will review it in the next two weeks.

wbond:
PRs can sometimes take hours to review, depending on how involved they are. And sometimes I am focused on edits to the core, so there may be a few weeks between reviews of substantial changes. This can be further compounded by not having used the language and needing feedback from someone who has, or learning it myself.

yegeniy:
Thank you both.

Extremely glad to hear you have a chance to focus on the core! Thanks for
the great product.

ismell:
Can you try now that #890 has been merged?

yegeniy:
Hi Raul, I think I can get the latest updates in the next release and apply the tests in
https://github.com/sublimehq/Packages/pull/917/files#diff-f7fcf52cd87d3ec05e5df7f8b58c5ba1, to make sure everything looks good. 

Is that what you had in mind?

wbond:
I added the tests from this to the test file in 6c161c6. Everything seems good, so I am closing this PR. Thanks for your contributions @yegeniy!

PR #919, opened 2017-04-21
[Diff] Handle non-existant files

FichteFoll:
Pretty self-explanatory. Happens a lot when opening resource files that don't exist on the file system at the location that ST maps them to.

Closes https://github.com/SublimeTextIssues/DefaultPackages/issues/96.

### DIFF
--- a/Diff/diff.py
+++ b/Diff/diff.py
@@ -38,7 +38,11 @@ class DiffFilesCommand(sublime_plugin.WindowCommand):
 class DiffChangesCommand(sublime_plugin.TextCommand):
     def run(self, edit):
 
-        fname = self.view.file_name();
+        fname = self.view.file_name()
+
+        if not fname or not os.path.exists(fname):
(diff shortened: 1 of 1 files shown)


Diff/diff.py
```diff
@@ -38,10 +38,15 @@ def is_visible(self, files):
 class DiffChangesCommand(sublime_plugin.TextCommand):
     def run(self, edit):
 
-        fname = self.view.file_name();
+        fname = self.view.file_name()
+
+        if not fname or not os.path.exists(fname):
+            sublime.status_message("The file does not exists")
```
keith-hall: Minor nitpick, the `s` at the end of `exist` is unnecessary :)
Maybe it would be better to say "unable to diff changes; the file does not exist", to make it clearer where the message comes from?

FichteFoll: It's not "unnecessary", though. It's just wrong. 😆

wbond:
Thanks!

PR #920, opened 2017-04-22
[Java] support multi-catch exception parameters

yegeniy:
Supports syntax highlighting with Java's multi-catch statements.

This change does not seem to introduce any regressions in
functionality or performance, aside from the fact that there is more code in the test itself now:

> HEAD^ (de4eaaa) without new test: 1.9ms
HEAD without new test: 2.0ms
HEAD^ (de4eaaa) with new test: 2.0ms
HEAD with new test: 2.0ms

---

I have a hunch that there's a way of making this implementation more elegant. 
Maybe it's possible to do something similar to how `method-parameters`
invokes `method-parameter-after`; and I bet a lot of the repeated matches,
such as `[,|] -> illegal` and `(?=[){}]) -> pop` could also be DRYed.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -374,8 +374,14 @@ contexts:
           pop: true
     - match: \b(goto|const)\b
       scope: invalid.illegal.java
-    - match: \b(try|catch|finally|throw)\b
+    - match: \b(try|finally|throw)\b
       scope: keyword.control.catch-exception.java
+    - match: \b(catch)\b
+      scope: keyword.control.catch-exception.java
+      push:
+        - match: (?=[){}])
+          pop: true
+        - include: exception-parameter-declaration
     - match: '\?|:'
       scope: keyword.control.java
     - match: \b(return|break|case|continue|default|do|while|for|switch|if|else)\b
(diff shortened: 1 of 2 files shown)


yegeniy:
Hi folks,  I've been using this locally, alongside #917, to some success for the past week..

Is there something I should do to get this turned down or up? Would it be better to close this out?

djspiewak:
:+1: from me.

Note that @wbond doesn't devote time to this repo every day, or sometimes even every week.  He usually merges PRs in bursts.  The best way to ensure things land expediently is to do exactly what you did: motivate, provide tests, and give a clean and self-contained implementation.  He'll get around to it eventually.

yegeniy:
Thanks @djspiewak - they gave me those details in #917. I should've updated the comment here.

btw, great work on the Java updates that came in Build 3127, btw! Glad to know there're people using ST for JVM development.

ismell:
Is this still valid after #890?

yegeniy:
Looking forward to trying out the work y'all did in 890.

I would add in the test to verify multi-catches to the work done in 890:

```java
} catch (final MyException | com.net.org.Foo.Bar |
    YourException ignore) {}
```

I didn't notice one there.

Best,
Eugene

wbond:
I've ported the tests from here and tweaked some scopes in d5204b8ae991639153b48a2e0288e157bf6a3f48.

PR #924, opened 2017-04-26
[CSS][Add] Scope for RGBA hexadecimal

Kristinita:
### 1. Settings

Part of my `SashaClassy.css` file:

```css
.SashaClass {
    color: #a1b2c3d4;
}

.SashaClass {
    color: #E5F6A7B8;
}
```

### 2. Behavior before pull request

Before

### 3. Behavior after pull request

After

### 4. Argumentation

[**RGBA Hexadecimal**](https://en.wikipedia.org/wiki/RGBA_color_space#RGBA_hexadecimal_.28word-order.29) support.

### 5. Tests

```python
Success: 2266 assertions in 1 files passed
[Finished]
```

Thanks.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -110,6 +110,10 @@ contexts:
       # Hex Color
     - match: '(#)(\h{3}|\h{6})\b'
       scope: constant.other.color.rgb-value.css
+      # RGBA Hexadecimal Colors
+      # https://en.wikipedia.org/wiki/RGBA_color_space#RGBA_hexadecimal_.28word-order.29
+    - match: '(#)(\h{8})\b'
+      scope: constant.other.color.rgba-value.css
       captures:
         1: punctuation.definition.constant.css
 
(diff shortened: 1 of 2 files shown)


Kristinita:
Sorry, I will try to fix the errors as quickly as possible.

CSS/CSS.sublime-syntax
```diff
@@ -110,6 +110,10 @@ contexts:
       # Hex Color
     - match: '(#)(\h{3}|\h{6})\b'
       scope: constant.other.color.rgb-value.css
+      # RGBA Hexadecimal Colors
```
keith-hall: it looks like this change introduces a regression in that capture group 1 (`#`) is no longer being scoped as `punctuation.definition.constant.css` for 3 or 6 hex digit colors.

Kristinita: @keith-hall , ✓ Fix by ae2aa9bb485fc6b99cd4409f06dab281a3dcc2c2. Thanks.

CSS/CSS.sublime-syntax
```diff
@@ -112,6 +112,12 @@ contexts:
       scope: constant.other.color.rgb-value.css
       captures:
         1: punctuation.definition.constant.css
+      # RGBA Hexadecimal Colors
+      # https://en.wikipedia.org/wiki/RGBA_color_space#RGBA_hexadecimal_.28word-order.29
+    - match: '(#)(\h{8})\b'
```
wbond: It is really a nitpick, but the second capture group is unused and not necessary.

wbond:
Thanks!

PR #926, opened 2017-04-26
[CSS] support nested parens in calc function call

### LINKED ISSUE #925: CSS: calc() highlighting incorrect beyond first parenthesis
It gets gradually worse for each parenthesis.

### LINKED ISSUE #987: CSS Grid & Custom Properties: var() inside repeat() breaks highlighting
Syntax highlighting for CSS breaks when combining CSS Grid and Custom Properties features as follows. Using any CSS language feature containing parenthesis (e.g. `var()`) within the CSS Grid `repeat()` function (I just call it that now, because I have no internet connection to verify terminology right now) will break the syntax highlighting after the statement.

I can only observe this behavior for `repeat()`. Other nested *function calls* in my code work (e.g. `var()` within `calc()`).

```
.sample {
  display: grid;
  --size: 50%;

  grid-template: var(--size) var(--size) / var(--size) var(--size);
  /* ↑ This highlights correctly */

  grid-template: repeat(2, 50%) / repeat(2, 50%);
  /* ↑ This highlights correctly as well */

  grid-template: repeat(2, var(--size)) / repeat(2, 50%);
  /* ↑ Highlighting breaks as soon as `()` appears within repeat */
}
```

P.S.: I’m not sure whether this is a duplicate of https://github.com/sublimehq/Packages/issues/925.

keith-hall:
fixes #925 and fixes #987

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -1751,6 +1751,7 @@ contexts:
             pop: true
           - include: comma-delimiter
           - include: color-adjuster-functions # must be included before `color-values`
+          - include: var-function
           - include: color-values
           - include: angle-type
           - include: number-type
@@ -1836,6 +1837,7 @@ contexts:
             scope: keyword.other.color-space.css
           - include: color-values
           - include: percentage-type
+          - include: var-function
 
   unicode-range:
     - match: |-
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #927, opened 2017-04-26
[JavaScript] Remove Mustache templating from JavaScript syntax.

Thom1729:
For #889. Fixes the following test:

```js
{{foo}}
// ^ meta.block meta.block variable.other.readwrite
```


wbond:
Thanks!

PR #928, opened 2017-04-26
[JavaScript][HTML] Removed HTML comments from JavaScript.

Thom1729:
For #889. Removes support for HTML comments in JavaScript, removes old tests, and adds new one.

#### Background

When JavaScript was first introduced, not all browsers supported `<script>` tags. If such a browser encountered a `<script>` tag, it would usually display its contents on the page. In order to avoid this, an accepted practice was to wrap the interior of the `<script>` tag with HTML comment markers `<!--` and `-->`.

Now, `<script>` tags cannot contain HTML comments -- anything inside them will be interpreted as JavaScript code. In browsers that support JavaScript (i.e. every browser today), these HTML comments are interpreted as either valid syntax (e.g. `a --> b`) or a syntax error (`<!--`). Fortunately (or unfortunately), browsers are lenient, so this workaround didn't break pages in practice.

There is no reason to use this old workaround anymore, because essentially every browser in use supports `<script>` tags. Supporting JavaScript itself isn't necessary, just correct parsing of HTML script tags. Of course, there's a lot of old code in the universe, and surely someone is using Sublime to view it.

#### Current state

Currently, JavaScript.sublime-syntax highlights these constructs as HTML comments even in pure JavaScript. This may sometimes result in the desired behavior, but it will also cause valid JavaScript to be highlighted incorrectly, as in the following failing test:

```js
x --> y;
//^^ keyword.operator.arithmetic.js
//  ^ keyword.operator.relational.js
```

It seems to be that correctly highlighting valid syntax wherever possible should be prioritized over highlighting invalid or variant syntax. If so, then the current implementation must be modified.

#### Options

The simplest fix -- implemented in this PR -- is to remove the HTML comment support from the syntax entirely. This should not affect any valid JavaScript code that is currently highlighted correctly. It will, however, affect the highlighting of invalid code and (by fixing the above bug) on valid code that is not currently highlighted correctly.

What then remains is ensuring that the highlighting of legacy HTML/JavaScript code is acceptable. After this PR, legacy code will be highlighted as follows:

```html
<script><!--
//      ^^^^ text.html.basic source.js.embedded.html keyword.operator.relational
   console.log("Hello, World!");
--></script>
```

This is invalid JavaScript, so there isn't a "correct" way to highlight it. This result doesn't bother me, and it shouldn't disrupt the surrounding HTML or JavaScript highlighting.

Alternatively, we could introduce special styling for this legacy workaround. I don't think it's necessary, but it wouldn't be particularly intrusive. The simplest option here would be to modify the HTML syntax so that an opening `<script>` tag will look for the characters `<!--` immediately afterward and highlight them as an HTML comment. Similarly, the closing tag could handle a preceding `-->`. This should not affect any valid JavaScript code, nor would it put any burden on the JavaScript.sublime-syntax. If there is a strong feeling in favor of this approach, I can implement it promptly.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -98,7 +98,7 @@ contexts:
         1: punctuation.definition.tag.begin.html
         2: entity.name.tag.script.html
       push:
-        - match: (?i)(-->)?(</)(script)(>)
+        - match: (?i)(-->)?\s*(</)(script)(>)
           captures:
             0: meta.tag.script.end.html
             1: comment.block.html punctuation.definition.comment.html
@@ -106,7 +106,7 @@ contexts:
             3: entity.name.tag.script.html
             4: punctuation.definition.tag.end.html
           pop: true
-        - match: '(>)(<!--)?'
+        - match: '(>)\s*(<!--)?'
           captures:
             1: meta.tag.script.begin.html punctuation.definition.tag.end.html
             2: comment.block.html punctuation.definition.comment.html
@@ -114,7 +114,7 @@ contexts:
             - meta_content_scope: source.js.embedded.html
             - include: 'scope:source.js'
           with_prototype:
-             - match: (?i)(?=(-->)?</script)
+             - match: (?i)(?=(-->)?\s*</script)
                pop: true
         - match: ''
           push:
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -98,20 +98,23 @@ contexts:
         1: punctuation.definition.tag.begin.html
         2: entity.name.tag.script.html
       push:
-        - match: (?i)(</)(script)(>)
+        - match: (?i)(-->)?(</)(script)(>)
           captures:
             0: meta.tag.script.end.html
-            1: punctuation.definition.tag.begin.html
-            2: entity.name.tag.script.html
-            3: punctuation.definition.tag.end.html
+            1: comment.block.html punctuation.definition.comment.html
+            2: punctuation.definition.tag.begin.html
+            3: entity.name.tag.script.html
+            4: punctuation.definition.tag.end.html
           pop: true
-        - match: '>'
-          scope: meta.tag.script.begin.html punctuation.definition.tag.end.html
+        - match: '(>)(<!--)?'
+          captures:
+            1: meta.tag.script.begin.html punctuation.definition.tag.end.html
+            2: comment.block.html punctuation.definition.comment.html
           push:
             - meta_content_scope: source.js.embedded.html
             - include: 'scope:source.js'
           with_prototype:
-             - match: (?i)(?=</script)
+             - match: (?i)(?=(-->)?</script)
                pop: true
         - match: ''
           push:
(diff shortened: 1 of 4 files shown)


wbond:
I think adding the HTML comment handling right inside of `<script>` tags in the HTML syntax is the way to go. I know from previous "best-practice" changes we've made that plenty of people still edit code written in a more legacy format.

Thom1729:
Here's an implementation. It's simplistic for now, only matching `<script...><!--` and `--></script>`. A more sophisticated approach might start by refactoring of that part of the HTML syntax.

keith-hall:
That's probably the way I would do it too :) - quick question though, do you think it could be quite common that legacy code had/has some whitespace between the HTML `script` tag and the HTML comment? i.e. `<script>   <!--` or anything?

HTML/HTML.sublime-syntax
```diff
@@ -98,20 +98,23 @@ contexts:
...
+        - match: '(>)(<!--)?'
+          captures:
+            1: meta.tag.script.begin.html punctuation.definition.tag.end.html
+            2: comment.block.html punctuation.definition.comment.html
           push:
             - meta_content_scope: source.js.embedded.html
             - include: 'scope:source.js'
           with_prototype:
-             - match: (?i)(?=</script)
+             - match: (?i)(?=(-->)?</script)
```
michaelblyons: Adding a `\s*` after `(-->)?` would allow whitespace if the `-->` is on the same line as the script closing tag, but if there is a line break, the `-->` would not be highlighted, correct? I ask because I have a similar problem with Graphviz.

keith-hall: Correct

Thom1729:
Yeah, that's simple enough to add. It won't work across multiple lines, though -- that's where we'd need a more substantial change.

Thom1729:
Also, unfortunately is is impossible to match `-->\n</script>` without breaking valid JavaScript.

wbond:
At some point I am hoping to explore adding some form of line lookahead, but that probably won't be in the near term, and I don't know how exactly it will end up being implemented.

I think a single line for now should be fine.

Thom1729:
@wbond Is the single-line limitation there to minimize reparsing when the text changes?

wbond:
I believe it is a combination of that and limiting how much Oniguruma can consume, thus preventing extensive backtracking. Jon and I have briefly discussed the topic and little, but nothing is currently planned. I think it may end up being limited to contexts where all patterns are compatible with sregex.

Thom1729:
Yeah, sregex matching on a given regex should be constant-space, linear-time-in-matched-length, right? I guess it depends on how the analysis works. When the correct match has capture groups, is the match performed in one pass or two? It seems like one pass might actually be slower in practice, not to mention the complexity of compiling together a whole context's worth of regexes including their capture groups. But two passes would require an extra state machine per match with captures.

wbond:
sregex is a non-backtracking engine, but the exact implementation details are beyond what I have taken the time to understand fully. I added support for a few simple regex operators, but Jon is the architect and primary implementor, and the only person who knows all of the details.

wbond:
Thank you for your work on this @Thom1729!

Issue #929, opened 2017-04-26
[JavaScript] Remove IE Conditional Compilation


Thom1729:
#### Background

Conditional compilation is an obscure feature introduced by Microsoft in IE 4 and removed in IE 11. It is similar to, but distinct from, Microsoft's HTML conditional comments.

Example ([source](http://www.javascriptkit.com/javatutors/conditionalcompile.shtml)):

```js
/*@cc_on
   /*@if (@_jscript_version >= 5)
        document.write("IE Browser that supports JScript 5+");
    @elif (@_jscript_version >= 4)
        document.write("IE Browser that supports JScript 4+");
    @else @*/
        document.write("Non IE Browser (one that doesn't support JScript)");
    /*@end
@*/
```

In most browsers, the contents of these comments would be ignored, but IE would execute the @directives and evaluate some of the comment contents.

This feature never caught on. HTML conditional comments can do the same thing in a more powerful and less confusing manner. Conditional compilation was removed entirely in IE 11 and no modern browser supports it. It is obscure even among web developers; I have never encountered it in the wild. Nonetheless, it's not impossible that *someone* might want to work with them in Sublime.

#### Current state

Right now, JavaScript.sublime-syntax does have special handling for conditional compilation. (Try the above example.) To the current implementations three objections may apply:

1.  It's buggy. The following test fails (see #889):

    ```js
    /*@if /*/
    //     ^^ punctuation.definition.comment.js
    ```

    This is valid JavaScript, and it's a nasty bug that messes up subsequent lines.

2.  It's inadequate. The contents are wrapped in a comment scope, which interferes with the attempt to highlight some terms correctly. It also doesn't (and probably can't) reliably integrate gracefully into the surrounding JavaScript. Overall, it seems tacked onto the syntax rather than robustly integrated into it. See, for instance:

    ```js
    /*@ This doesn't work */
    ```

3.  It's cruft. It's not real JavaScript, it's been obsolete for years, and it's quite rare in the wild. The core syntax can handle it gracefully without any special help. In other words, if they are stripped out entirely, then nothing of value will be lost.

    The consensus seems to be bearish toward proprietary extensions of JavaScript. See #860, which concerns an extension that would be useful to millions of developers. As this extension would be useful to next to no one, it doesn't seem like the sort of thing that needs to be integrated into the core bundle.

[...]

PR #930, opened 2017-04-26
[JavaScript] Removed IE conditional compilation. Added test to cover old bug.

Thom1729:
For #929. "Option 1": removing conditional compilation entirely.


wbond:
I think your arguments in #929 are compelling. Let's do it!

PR #932, opened 2017-04-27
[R] major update

randy3k:
It is a major update of the R language based on my package [R-Box](https://github.com/randy3k/R-Box). The default R syntax was derived from TextMate and it basically does nothing. The updated syntax does better in captures functions and function arguments. A set of built-in functions is also added as supported functions.

Once it is merged, the R-Extended syntax of R-Box will be deprecated.


wbond:
Is the `.` an accessor, or really part of the identifier?

wbond:
BTW, thanks for submitting this, especially with all of the tests!

randy3k:
There is a complicated story bebind the dot `.` in R language. However, for highlight purposes, `.` can be treated as a part of the identifier.

wbond:
Seems like it is probably a big improvement, and based on your work on LaTeX, I trust you have a pretty good idea what you are doing.

Thank you!

PR #936, opened 2017-04-27
[C#] fix catch with generic type and no variable name

### LINKED ISSUE #934: [C#] catch block with a generic type and no variable name breaks highlighting
Currently, a catch with a generic type and no variable name causes the highlighting to be messed up.

```cs
class Main {
    void Main () {
        try
        {
            // some code here to call a webservice
        }
        catch (FaultException<ServiceFault>) // the closing paren does not get scope punctuation.section.group.end.cs unless there is a variable name before it
        { // curly brace is unscoped
            throw; // throw gets scope variable.other.cs
        } // curly brace is unscoped
    } // curly brace is unscoped
} // curly brace is unscoped
```

adding a name like `e` before the `)` fixes it, but it isn't necessary in C#. Removing `<ServiceFault>` also fixes it.

generic in catch without name

keith-hall:
fixes #934

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1751,6 +1751,8 @@ contexts:
               captures:
                 1: variable.other.cs
               pop: true
+            - match: (?=\))
+              pop: true
         - include: type_argument
     - match: '({{base_type}}){{type_suffix_capture}}'
       captures:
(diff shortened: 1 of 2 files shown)


wbond:
Thank you!

PR #948, opened 2017-04-28
[C#] add support for composite string placeholder formatting

### LINKED ISSUE #947: [C#] composite string placeholder formatting currently unscoped
Currently, the C# syntax definition does not scope [string placeholders that have composite formatting](https://msdn.microsoft.com/en-us/library/txafckwd(v=vs.110).aspx) like:

```cs
string.Format("{0, -20} {1,5 :N1}", "Fred", 40.333m);
```

instead they are just highlighted as plain characters in the string. Although, [it does attempt to do so for interpolated strings](https://github.com/sublimehq/Packages/blob/9e8527ca1ae15403d30d70338ff7d1a472fba16b/C%23/C%23.sublime-syntax#L1484) like:
```cs
var test = 17.45;
Console.WriteLine($"{test,8:N1}");
```

Also, it would be handy if it highlighted invalid format specifiers like:

```cs
string.Format("{0G}", 40.333m); // missing a colon - should be string.Format("{0:G}", 40.333m)
string.Format("{0:G", 40.333m); // missing a closing brace
```

and still be sure that it still correctly highlights placeholders with escapes in the format string:

```cs
string.Format("{0:00.00000{{}}test} me", 40.333m); // outputs: 40.33300{}test me
string.Format("{0:00.00000{{}}test\\\"} me", 40.333m); // outputs: 40.33300{}test" me
string.Format(@"{0:00.00000{{}}test\""} me", 40.333m); // outputs: 40.33300{}test" me
```

keith-hall:
fixes #947

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1445,9 +1445,9 @@ contexts:
       push:
         - meta_scope: meta.string.interpolated.cs source.cs
         - clear_scopes: 2
-        - include: string_interpolation
         - match: $
           pop: true
+        - include: string_interpolation
     - match: $\n?
       scope: invalid.illegal.unclosed-string.cs
       pop: true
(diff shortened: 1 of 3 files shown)


wbond:
Thanks!

PR #956, opened 2017-04-29
[XML] support xml-model preprocessor instructions

keith-hall:
this adds support for the `xml-model` preprocessor instruction, just like we added for `xml-stylesheet`: https://github.com/sublimehq/Packages/issues/882

https://www.w3.org/TR/xml-model/#the-xml-model-processing-instruction


wbond:
Thank you

PR #960, opened 2017-05-03
[C#] make lambda parameter separator scope more specific

keith-hall:
use the same scope for parameter separators in lambdas as for normal methods, for consistent highlighting.
Fix bug with lambdas with multiple parameters and no curly braces, similar to https://github.com/sublimehq/Packages/pull/788


wbond:
Thanks!

PR #962, opened 2017-05-05
[CSS] Update missing values/properties

philipbordallo:
- Add `border-box`, `text-before-edge`, `text-after-edge`, `ruby-base`, `ruby-text`, `run-in`, `ruby-base-container`, and `ruby-text-container` to CSS syntax for correct highlighting, these already exist in _css_completions.py_
- Add `user-select` to completions. Although this still needs a prefix in most browsers (see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) and [Caniuse](http://caniuse.com/#search=user-select)), having this part of the completions would be great as most users are probably using Autoprefixer


wbond:
Thanks!

Issue #968, opened 2017-05-06
[Makefile] Wrong endif scope


evandrocoan:
It has the scope `source.makefile meta.function.prerequisites.makefile`, when the last immediate line has only a colon `:` as definition.

```makefile
# OS detecting makefile
# http://stackoverflow.com/questions/714100/os-detecting-makefile
#
ifeq ($(OS),Windows_NT)
    SEPARATOR = ;
else
    SEPARATOR = :
endif
```

![image](https://cloud.githubusercontent.com/assets/5332158/25775460/e5a4e846-327b-11e7-993f-1beb51f5fcbe.png)

rwols:
It looks like the makefile syntax def has trouble with colons in general. See also https://github.com/sublimehq/Packages/issues/903.

deathaxe:
This issue seems to be fixed.

Issue #903, opened 2017-04-16
makefile: built-in functions bug


pdknsk:
![1](https://cloud.githubusercontent.com/assets/1097609/25075022/b5f1ef98-230a-11e7-970b-f3501c56a2a7.png)

The `:` apparently opens a scope without closing it. Works with other functions too.
```Makefile
$(warning Bug: $(BUG))
```

rwols:
A naive fix would be to include the `variable` context in the `main` context, but I'm not sure if other things will start breaking in that case.

Just leaving this syntax test here for reference.

```makefile
$(warning he:llo)
# ^ meta.support.function support.function.builtin
#         ^ meta.support.function
#               ^ meta.support.function punctuation.definition.variable.end

all: deps
	$(warning he:llo)
	# ^ meta.function.body meta.support.function support.function.builtin
	#         ^ meta.function.body meta.support.function
	#               ^ meta.function.body meta.support.function punctuation.definition.variable.end
deps:
	$(warning he:llo)
```

deathaxe:
@wbond: This issue is closed by https://github.com/sublimehq/Packages/pull/974

wbond:
This issue has been solved by a previous commit, hence it is being closed.

PR #969, opened 2017-05-07
[Makefile] Fix colon problems

rwols:
Fixes both https://github.com/sublimehq/Packages/issues/968 and https://github.com/sublimehq/Packages/issues/903.

It also has some tiny enhancements like adding `punctuation.separator` scopes for a comma between function arguments (see the syntax tests).


rwols:
Disregard this PR, I'm working on a way better refactoring.

PR #971, opened 2017-05-10
[HTML] support self-closing other/custom tags

keith-hall:
fixes https://github.com/SublimeTextIssues/DefaultPackages/issues/176 (problems with autocompletion when typing `</` when a non-standard tag is used that is self-closing)


wbond:
Thank you!

PR #974, opened 2017-05-13
[Makefile] Big refactor

rwols:
# TL;DR
Fixes both https://github.com/sublimehq/Packages/issues/968 as well as https://github.com/sublimehq/Packages/issues/903. Also see https://github.com/sublimehq/Packages/issues/481.

# The Problem
The current syntax has a collection of problems:
1. It has various positive lookbehinds. This caused problems (https://github.com/sublimehq/Packages/issues/903).
2. It forces you to use tabs for indenting recipes.
3. It cannot detect wether something is the body of a recipe, or just regular indentation for an `ifeq/else/endif` block.
4. It has its own little subset of a shell-like syntax. This is used for highlighting recipe lines.

# Solutions
1. This refactoring has **zero look behinds**. Everything is done with look aheads.
2. You can use tabs or spaces for indenting recipe lines. It can detect recipes with mixed tabs/spaces and will scope the unexpected tabs or spaces as `invalid.illegal.expected.spaces.makefile` or `invalid.illegal.expected.tabs.makefile`.
3. I've done my best to try to detect wether a line in the file is the start of a new rule definition. If so, the syntax will push the `expect-rule` context onto the stack. After that, it'll go into the `recipe-junction-between-spaces-or-tabs` context. This context will detect wether you're starting a recipe with spaces or tabs. The rest should be straightforward. The nice result is that recipe lines are now scoped with `meta.function.body.makefile`.
4. The body of a recipe line is scoped with `scope:source.shell`, enabling the full power of the shell syntax.

# Other Improvements
- [Automatic variables](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html) are scoped as `variable.language.makefile`. This makes these important variables stand out.
- The wildcard symbol `*` inside a `$(wildcard ...)` function call is scoped as `variable.language.wildcard.makefile`.
- The percentage-symbol `%` is also scoped as `variable.language.makefile` in places where this is appropriate.
- The first argument of the `$(call ...)` function is actually scoped as `support.function` (like all other functions), instead of a first argument of a function. The word `call` is scoped as `constant.language`. This is because the `call` function is special in the sense that it's the only function that allows you to call your own written functions, and the word `call` is more of a keyword than an actual function name.
- The `source.shell` syntax is also pushed inside a `$(shell ...)` function call.
- The `source.shell` syntax is also pushed after a variable definition with the `!=` keyword.
- This PR has about ~400 syntax tests (the current one has about ~100).

I'll be pushing some more minor commits but the bulk of the work has been done I think.

### DIFF
--- a/Makefile/Makefile.sublime-settings
+++ b/Makefile/Makefile.sublime-settings
@@ -1,0 +1,4 @@
+{
+	"translate_tabs_to_spaces": false,
+	"detect_indentation": false
+}
--- a/Makefile/Makefile.sublime-settings
+++ b/Makefile/Makefile.sublime-settings
@@ -1,4 +1,0 @@
-{
-	"translate_tabs_to_spaces": false,
-	"detect_indentation": false
-}
--- a/Makefile/Makefile.sublime-settings
+++ b/Makefile/Makefile.sublime-settings
@@ -1,4 +1,4 @@
 {
 	"translate_tabs_to_spaces": false,
-	"detect_indentation": false
+	"detect_indentation": true
 }
--- a/Makefile/Comments.tmPreferences
+++ b/Makefile/Comments.tmPreferences
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+    <key>scope</key>
+    <string>source.makefile</string>
+    <key>settings</key>
+    <dict>
+        <key>shellVariables</key>
+        <array>
+            <dict>
+                <key>name</key>
+                <string>TM_COMMENT_START</string>
+                <key>value</key>
+                <string># </string>
+            </dict>
+        </array>
+    </dict>
+</dict>
+</plist>
--- a/Makefile/Makefile.sublime-syntax
+++ b/Makefile/Makefile.sublime-syntax
@@ -124,6 +124,7 @@ contexts:
       captures:
         1: keyword.control.conditional.makefile
       push:
+        - include: control-flow
         - include: comments
         - include: pop-on-line-end
     - match: ^\s*(endif)\b
@@ -281,6 +282,22 @@ contexts:
       scope: constant.character.escape.makefile
 
   comments:
+    # This hack is here in order to circumvent false-positives for the big
+    # rule lookahead. For example, if this match is not present, then things
+    # like
+    # 
+    #     # A comment with a : colon
+    #     
+    # will match the rule-lookahead.
+    - match: (?=^\s*\#)
+      push:
+        - match: \#
+          scope: punctuation.definition.comment.makefile
+          set:
+            - meta_scope: comment.line.number-sign.makefile
+            - include: pop-on-line-end
+    # This is the "normal" comment parsing logic, but not sufficient in every
+    # case (see above).
     - match: \#
       scope: punctuation.definition.comment.makefile
       push:
--- a/Makefile/Makefile.sublime-syntax
+++ b/Makefile/Makefile.sublime-syntax
@@ -5,6 +5,9 @@
 # Includes a few improvement like special-target, a bit more shell command,
 # better variable recognition, fix ifeq/ifneq missing highlight ...
 # Rewritten by Raoul Wols on May 2017.
+# 
+# All number references refer to the "Make manual" located at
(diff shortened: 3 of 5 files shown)


Makefile/Makefile.sublime-settings
```diff
@@ -1,4 +0,0 @@
-{
-	"translate_tabs_to_spaces": false,
```
FichteFoll: I suggest setting `translate_tabs_to_spaces` to `false` while still having `detect_indentation` as `true`. It will use the sane default while still being able to detect files that are indented with a single space (or any other character), in case anyone does use `.RECIPEPREFIX`.

rwols: OK, good advice, will change this later.

wbond:
This looks like a huge improvement with a ton of tests. Thank you!

The only thing I noticed in glancing over it is a style issue. You used CamelCase identifiers for variable names where the rest of the syntaxes (to my knowledge) use underscore_notation.

rwols:
I'm pretty satisfied with the progress now, but I still want to fix a tiny nitpick thing:

schermafbeelding 2017-05-25 om 15 30 24

The dark-green background color is for the `string.unquoted` scope. As you can see, it starts scoping it as `string.unquoted` immediately after the assignment token. In reality, Make first eats whitespace, and then when it finds the first non-whitespace character it considers that as the start of the value that you're assigning to the variable. I'll try to fix this.

rwols:
I should probably give an overview of what the last few commits did. I'm currently just browsing huge Makefiles on GitHub and scouting for potential bugs.

1. When somebody does a `sed` inside a `$(shell ...)` function-call, the syntax would trip over a `)` usually inside the regex for `sed`, since `)` is checked in the `with_prototype` context accompanying the pushed `scope:source.shell` syntax in order to exit the `$(shell ...)` function-call. This tripped up the syntax in major ways, because usually the closing double-quotes `"` of the argument to `sed` would be seen as the start of a `source.makefile` string, whereas it should be the end of a `source.shell` string. This is fixed by having our own `quoted-string` context in the `with_prototype` context. So, quoted strings in embedded `source.shell` parts no longer are `source.shell` strings but rather `source.makefile` strings. This is 43fe4ba. [Extra: and *only* for $(shell ...) calls; 9053e95]

2. When somebody invokes `$(error ...)` or `$(info ...)` or `$(warning ...)`, or any function, it is perfectly viable to write a `(` character; you can even balance that with an accompanying `)` character. Make parses this simply as text. For instance:
schermafbeelding 2017-05-29 om 20 21 17
Before, this would trip up the syntax because it thought that the second-to-last closing parenthesis was the closing parenthesis of the `$(call ...)` function. Commit bf2c8e6 fixes that.

3. Out in the wild I have encountered an insane rule definition that the rule-lookahead would not recognize (it didn't eat enough variable-substitutions).
schermafbeelding 2017-05-29 om 20 28 16
As you can see, it correctly recognizes it now as a rule as can be seen by the red colon somewhere in the middle of this expression. This is 	eb23df2.

4. Your own written functions are now scoped as `variable.function` and built-in functions are scoped as `support.function.builtin`. Textual substitutions are scoped as `variable.parameter`.

5. I was hesitant to include variable-substitutions in the main context. Now that we're at around 600 tests, I felt like it was time to add it in, because it feels more natural to the user when he starts typing `$(` that he'll immediately see it getting scoped as a variable-substitution. All tests pass, so I'm confident this will not screw anything up. This is a2f4cf7.

6. It turns out inline-recipes didn't push the `source.shell` syntax; they do now. This is 294d894. This feature is not used much in practice, so not a big deal.

wbond:
I looked through a bunch of this and didn't see anything that looked amiss, so I am going to merge it and we can gather feedback as part of the next dev build.

wbond:
Thanks for all of your work on this @rwols!

rwols:
Thanks mate! I should probably revisit this once the shell rewrite gets merged, because the make syntax now pushes that in certain places. I kept this in mind while doing the shell rewrite, so I *think* all should be good, but you never know.

wbond:
Unfortunately this is currently not usable, as-is because the regex cache is 12mb and causes a very noticeable hang when first loading a makefile. I am working to see if I can mitigate it some so we can get it in.

PR #976, opened 2017-05-15
Enhancements to C/C++ syntax to allow for indexing of enumeration types

educatedpork:
..  as discussed at https://forum.sublimetext.com/t/st3-goto-definition-and-c-c-enumerations/8669

The changes in this commit are from user christoph314 at https://forum.sublimetext.com/t/st3-goto-definition-and-c-c-enumerations/8669/13

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -209,6 +209,13 @@ contexts:
     - include: preprocessor-statements
     - include: scope:source.c#label
     - include: expressions
+    
+  statements-enum:
+    - include: preprocessor-statements
+    - include: scope:source.c#label
+    - match: '{{identifier}}'
+      scope: entity.name.enum.c++.constants
+    - include: expressions    
 
   expressions:
     - include: early-expressions
@@ -625,6 +632,9 @@ contexts:
         - match: \b(struct)\s+({{identifier}})\b
           captures:
             1: storage.type.c++
+        - match: \benum\b
+          scope: storage.type.c++
+          set: data-structures-enum-definition
         - include: expressions-minus-generic-type
 
   parens:
@@ -1011,7 +1021,7 @@ contexts:
         - match: '\}'
           scope: meta.enum.c++ meta.block.c++ punctuation.section.block.end.c++
           pop: true
-        - include: statements
+        - include: statements-enum
 
   data-structures-union-definition:
     - meta_scope: meta.union.c++
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -479,7 +479,7 @@ contexts:
         - match: '\}'
           scope: meta.enum.c meta.block.c punctuation.section.block.end.c
           pop: true
-        - include: data-structures-body
+        - include: data-structures-body-enum
 
   data-structures-union-definition:
     - meta_scope: meta.union.c
@@ -517,6 +517,14 @@ contexts:
     - match: '(?={{before_tag}})'
       push: data-structures
     - include: expressions
+    
+  data-structures-body-enum:
+    - include: preprocessor-data-structures
+    - match: '(?={{before_tag}})'
+      push: data-structures
+    - match: '{{identifier}}'
+      scope: entity.name.enum.constants
+    - include: expressions    
 
   block:
     - match: '\{'
@@ -624,6 +632,9 @@ contexts:
         - match: \b(struct)\s+({{identifier}})
           captures:
             1: storage.type.c
+        - match: '\benum\b'
+          scope: storage.type.c
+          set: data-structures-enum-definition
         - include: expressions
 
   function-call:


C++/C++.sublime-syntax
```diff
@@ -209,6 +209,13 @@ contexts:
     - include: preprocessor-statements
     - include: scope:source.c#label
     - include: expressions
+    
+  statements-enum:
+    - include: preprocessor-statements
+    - include: scope:source.c#label
+    - match: '{{identifier}}'
+      scope: entity.name.enum.c++.constants
```
wbond: This scope should be `entity.name.constant.c++` since that is what is indexed. The scope `entity.name.enum` is used for the enum type name. A note for future changes, the syntax scope (`.c++`) should always come last.

C++/C.sublime-syntax
```diff
@@ -517,6 +517,14 @@ contexts:
     - match: '(?={{before_tag}})'
       push: data-structures
     - include: expressions
+    
+  data-structures-body-enum:
+    - include: preprocessor-data-structures
+    - match: '(?={{before_tag}})'
+      push: data-structures
+    - match: '{{identifier}}'
+      scope: entity.name.enum.constants
```
wbond: This should be `entity.name.constant.c`

C++/C++.sublime-syntax
```diff
@@ -625,6 +632,9 @@ contexts:
         - match: \b(struct)\s+({{identifier}})\b
           captures:
             1: storage.type.c++
+        - match: \benum\b
+          scope: storage.type.c++
+          set: data-structures-enum-definition
```
wbond: I think this may require some more special casing. With typedef'ed struct, we don't scope the identifier as `entity.name.struct` since the purpose of the typedef is to expose the struct as the type's identifier. I think we'd probably want to do the same in this case, in which case we'd need to consume the identifier after `enum` and move right to a context where we can match the type's name, a comment, or `{`. Tests will definitely help to ensure we are handling the variations of this structure properly.

C++/C.sublime-syntax
```diff
@@ -624,6 +632,9 @@ contexts:
         - match: \b(struct)\s+({{identifier}})
           captures:
             1: storage.type.c
+        - match: '\benum\b'
+          scope: storage.type.c
+          set: data-structures-enum-definition
```
wbond: This will need the same treatment as typedefs for C++.

rwols:
@educatedpork, any updates on this?

educatedpork:
@rwols, unfortunately not. I've been too busy in my day-to-day work to look at it.

raxoft:
I can't believe this hasn't been solved yet. I gave ST3 a whirl now and hit this almost immediately. How can you guys work on a substantial project when you can't find enum values quickly? I am really glad that at least Find in Files works... Anyway, +1 for having this implemented. Thanks.

TheClams:
I'm not sure how to contribute to a PR on-going, but I tried to implement the fix / test requested by Will : https://github.com/TheClams/Packages/tree/index-c-enumeration-values

kurazsi:
any progress on this topic ?

kurazsi:
@wbond can this fix be pushed for the release pls. the content of the syntax check changed more then in the past, and now i cant hack it to enable this and it's very annoying having to search all the files in the project just to see the values of some enums.

wbond:
@kurazsi No one has provided a working fix, so I can't merge it.

marcjvandyke:
Can we change the definition of ready for release from completely working to mostly/somewhat working?
I get that what is mentioned below is the complete list. But how about the basics. C/C++, enum, typedef enum.
It is just a pain to use Find In Files to track down an enum type.
Thanks, Marc

    We'll need tests for a number of variations on enum, enum class, typedef enum identifier identifier, typedef enum identifier {} identifier, and typedef enum { } identifier
    The changes will need to be ported to the Objective-C and Objective-C++ syntaxes also. The differences in the languages prevent large-scale reuse, so most contexts are duplicated between C/Objective-C and C++/Objective

LinuxDevon:
I was going to take a stab at this... I am annoyed enough that I want to fix it. what is the best way to go about updating this? Do I need to checkout @educatedpork branch and push to that? 

I have the commit this is at and re-based with the conflicts but want to get a second opinion before updating your requested changes. Just not sure how to updating a PR I don't own or is it better to open a different one.

keith-hall:
I don't think you'll be able to push your changes to someone else's fork/branch. You can open a PR on their repo, and after merging that one, it would become part of this PR. But if the original author is unresponsive, I guess the way forward is to open a separate PR, and reference that it supersedes this one.

LinuxDevon:
Alright opened a PR on the original branch with the rebase. We will see if he responses / pulls it in.

LinuxDevon:
No comment from the original poster @educatedpork for a week now. Seeing as how it has been over 6 years I doubt there will be a response. I think this PR can be closed now as long as the other PR is sufficient.

The rebase is really ugly... so I started from master and pulled changes in manual. Then added some tests I opened a new PR for everyone's viewing pleasure. Hopefully this will get pulled in soon!

deathaxe:
Superseeded by #3299

PR #981, opened 2017-05-19
[Python] Do better at matching numbers

### LINKED ISSUE #931: [Python] the dot in the number `.999` is not highlighted correct.
It only happens with the updated Python syntax.

v3126
screen shot 2017-04-26 at 7 24 46 pm

The current master
screen shot 2017-04-26 at 7 24 28 pm

cc @FichteFoll 
I am cc'ing you because you were working on the Python syntax.

FichteFoll:
Fixes #931 and a couple other minor improvements.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -23,7 +23,7 @@ variables:
   # We support unicode here because Python 3 is the future
   identifier_continue: '[[:alnum:]_]'
   identifier: '\b[[:alpha:]_]{{identifier_continue}}*\b'
-  digitpart: (?:\d[\d_]*)
+  digitpart: (?:\d(?:_?\d)*)
   path: '({{identifier}} *\. *)*{{identifier}}'
   sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|WITH)\b
   illegal_names: (?:and|as|assert|break|class|continue|def|del|elif|else|except|finally|for|from|global|if|import|in|is|lambda|not|or|pass|raise|return|try|while|with|yield)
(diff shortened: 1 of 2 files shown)


FichteFoll:
Also added scopes for the decimal mark and suffixes for complex and long numbers (the latter being Python 2 syntax).

Python/Python.sublime-syntax
```diff
@@ -349,17 +349,31 @@ contexts:
     # complex
     - match: (?i){{digitpart}}?(\.){{digitpart}}(?:e[\-\+]?{{digitpart}})?j # mandatory fraction
       scope: constant.numeric.complex.python
+      captures:
+        1: punctuation.separator.decimal.python
     - match: \b(?i)(?:{{digitpart}}(?:(\.){{digitpart}}?)?|(\.){{digitpart}})(?:e[\-\+]?{{digitpart}})j # mandatory exponent
```
rwols: Same here.

Python/Python.sublime-syntax
```diff
@@ -328,29 +328,67 @@ contexts:
       scope: constant.language.python
 
   numbers:
-    - match: \b(?i)(0x\h*)L # py2
+    # https://docs.python.org/3/reference/lexical_analysis.html#numeric-literals
+    # hexadecimal
+    - match: \b(?i)(?:0x\h*)(L) # py2
```
FichteFoll: Why should I do that? What is the advantage?

rwols: There's some redundancy in having matches check for the leading `0x` constantly, but it's just a suggestion and if you feel that it isn't worth pursuing this idea then don't feel obliged.

FichteFoll: If you pay close attention, you can see that there are different scopes assigned for the two matches. The py2 legacy one had an additional `long` segment and this couldn't possibly be done by first pushing into a different context. 

It would also be a lot more verbose overall, imo.

wbond:
Thanks!

Issue #703, opened 2016-11-24
[Python] colon is not scoped after a control statement


keith-hall:
if True:
        pass

    while False:
        pass

the `:` characters in the above examples currently have no scope. Same with `try:`.
I think it will be useful to scope it, so color schemes can highlight it differently.

PR #982, opened 2017-05-19
[Python] Properly match all block statements

FichteFoll:
Assign meta and punctuation scopes.

Fixes https://github.com/sublimehq/Packages/issues/703.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -187,15 +187,46 @@ contexts:
         - match: '\b(as)\b'
           scope: keyword.control.flow.as.python
           set:
-            - meta_scope: meta.statement.except.python
+            - meta_content_scope: meta.statement.except.python
             - include: line-continuation-or-pop
             - match: ':'
-              scope: punctuation.section.block.except.python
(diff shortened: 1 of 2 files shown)


wbond:
Things like this cause issues when embedding the syntax in a markup language. I am going to hold off until I can look into the issue some more.

FichteFoll:
FWIW, the new keywords affected by this are `if` and `while`. `for`, `import` (and `from`) and `except` are already affected by this.

FichteFoll:
Will adding a pop pattern on any unmatched non-whitespace character suffice for getting this merged? Most situations where Python could be embedded are probably markdown and latex, which use `` ` `` and ` \ ` respectively - both invalid in Python (3). Backticks were used in 2 as a `repr()` alias, but I haven't seen them actually being used so far. In fact, they aren't even highlighted currently and probably have never been.

FichteFoll:
I believe with the recent addition of the `embed` feature, this can be looked at again.

wbond:
Let's try it out – thanks for the reminder.

PR #984, opened 2017-05-20
[Lua] Add support for multiline comments

p3lim:
Been using this for years, just never bothered with submitting it until now.

### DIFF
--- a/Lua/Comments.tmPreferences
+++ b/Lua/Comments.tmPreferences
@@ -15,6 +15,18 @@
 				<key>value</key>
 				<string>-- </string>
 			</dict>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_START_2</string>
+				<key>value</key>
+				<string>--[[</string>
+			</dict>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_END_2</string>
+				<key>value</key>
+				<string>--]]</string>
+			</dict>
 		</array>
 	</dict>
 </dict>


Lua/Comments.tmPreferences
```diff
@@ -15,6 +15,18 @@
...
+				<key>name</key>
+				<string>TM_COMMENT_START_2</string>
+				<key>value</key>
+				<string>--[[</string>
+			</dict>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_END_2</string>
+				<key>value</key>
+				<string>--]]</string>
```
FichteFoll: The leading `--` are optional here. Don't think I've seen code using them.

p3lim: One of the similarities between Lua and Python is their use of the same symbols for multiline comments as for multiline strings (`'''` and `[[]]`).

In Lua, you start a multiline string with `[[`, and a comment with `--[[`, but both of them can be ended with just `]]`. The reasoning for implementing this as `--]]` instead is just to add a distinguishing factor to it so you can easily tell it's a comment ending and not a string ending.  
(I'm not going to bother with the fact that you can infix up to multiple `=` symbols in either)

rwols: IMO it should be obvious from the color scheme and the scoping that a comment block ends.

p3lim: A good point, but there are people that for some reason don't run color schemes (a reasonably small percentage though).  
I'm not opposed to either version, it's just the one I'm used to.

p3lim: After some quick testing, Sublime's lexer is able to comment and uncomment when a multiline string is within the commented area with `--[[--]]`, but not with `--[[]]`.

OdatNurd: It's a somewhat common Lua idiom to use `--]]` to close a block comment like this because it allows you to easily toggle the state of the block comment by adding an extra `-` to the `--[[` token. It's mentioned in PiL for example: https://www.lua.org/pil/1.3.html

wbond:
From what I can tell, this appears to be good to include. Thanks!

PR #985, opened 2017-05-20
[XML] Add support for Document Type Definition (DTD)

deathaxe:
Hello Devs

The current _XML.sublime-syntax_ provides only basic support of internal DOCTYPE subsets.
The declarations don't highlight well if rules span multiple lines.

Reading some official DTD files from DITA/DOCBOOK etc. revealed support for external DTD files is missing at all.

Suggestion:

This PR intends to solve the highlighting issues related with all the declaration tags and adds support for DTD-files.

A _DTD.sublime-syntax_ is created which is imported by the XML.sublime-syntax. DTD is understood as part of XML and thus placed into the XML package. We may or may not create a dedicated package folder for it.

If you like this idea, the DOCTYPE definitions of HTML package could be updated the same way, too.

_Note: This PR includes #978 for technical reasons._

### DIFF
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -95,8 +95,6 @@ contexts:
         1: punctuation.definition.tag.begin.xml
         2: entity.name.tag.doctype.xml
       push: doctype-root-name
-      with_prototype:
-        - include: comment
 
   doctype-root-name:
     - meta_scope: meta.tag.declaration.doctype.xml
@@ -128,6 +126,7 @@ contexts:
         - include: scope:text.xml.dtd
 
   doctype-end:
+    - include: comment
     - match: '>'
       scope: meta.tag.declaration.doctype.xml punctuation.definition.tag.end.xml
       pop: true
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -47,18 +47,7 @@ contexts:
             4: punctuation.separator.key-value.xml
         - include: double-quoted-string
         - include: single-quoted-string
-    - match: '(<!)(DOCTYPE)(?:\s+({{name}}))?'
-      captures:
-        1: punctuation.definition.tag.begin.xml
-        2: keyword.doctype.xml
-        3: variable.documentroot.xml
-      push:
-        - meta_scope: meta.tag.sgml.doctype.xml
-        - match: \s*(>)
-          captures:
-            1: punctuation.definition.tag.end.xml
-          pop: true
-        - include: internal-subset
+    - include: doctype
     - include: comment
     - match: '(</?){{qualified_name}}([^/>\s]*)'
       captures:
@@ -113,11 +102,13 @@ contexts:
     - match: ']]>'
       scope: invalid.illegal.missing-entity.xml
     - include: should-be-entity
+
   should-be-entity:
     - match: '&'
       scope: invalid.illegal.bad-ampersand.xml
     - match: '<'
       scope: invalid.illegal.missing-entity.xml
+
   double-quoted-string:
     - match: '"'
       scope: punctuation.definition.string.begin.xml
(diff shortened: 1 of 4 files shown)


XML/syntax_test_xml.xml
```diff
@@ -17,9 +17,17 @@
 <!--                     ^^ punctuation.definition.tag.end -->
 <!--                       ^ - meta.tag.preprocessor -->
 
-
-     <?XML version="1.0" ?>
-<!-- ^ invalid.illegal.missing-entity - meta.tag.preprocessor -->
+     <?XML VERSION="1.0" ?>
+<!-- ^^^^^^^^^^^^^^^^^^^^^^ meta.tag.preprocessor -->
```
keith-hall: this is not valid in XML - try it on https://codebeautify.org/xmlvalidator or http://www.xmlvalidation.com/ or read the RFC. The behavior prior to this commit was correct.

> The processing instruction target matching "[xX][mM][lL]" is not allowed.

deathaxe: Wow, quick anwer. But the other tags like `<XML-STYLESHEET>` are case insensitive?

XML/XML.sublime-syntax
```diff
@@ -20,89 +21,62 @@ scope: text.xml
 variables:
   # This is the full XML Name production, but should not be used where namespaces
   # are possible. Those locations should use a qualified_name.
-  name: '[[:alpha:]:_][[:alnum:]:_.-]*'
+  name: '(?i)[[:alpha:]:_][[:alnum:]:_.-]*'
```
keith-hall: `:alnum:` and `:alpha:` already includes upper and lower case, so the `(?i)` is redundant, plus will affect the rest of any regex patterns that use these variables. http://www.regular-expressions.info/posixbrackets.html

XML/syntax_test_xml.xml
keith-hall: No, `<?xml-stylesheet` has to be all lowercase,

XML/syntax_test_xml.xml
```diff
@@ -201,6 +222,26 @@
...
 
+     <NS:TAGNAME XMLNS:NS="URI">
+<!-- ^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.tag -->
+<!--                            ^ - meta.tag -->
+<!-- ^ punctuation.definition.tag.begin -->
+<!--  ^^ entity.name.tag.namespace -->
+<!--    ^ punctuation.separator.namespace -->
+<!--     ^^^^^^^ entity.name.tag.localname -->
+<!--            ^ - entity -->
+<!--             ^^^^^ entity.other.attribute-name.namespace -->
```
keith-hall: Note that `xmlns` must be lowercase to be treated as a namespace declaration, so this is not defining a namespace with prefix `NS`, so this node is therefore only valid if the `NS` prefix has been defined previously in document order. As the syntax definition doesn't scope the attribute differently, there is actually nothing wrong with this syntax test, except potential confusion in future if the syntax definition gets changed to scope namespace declarations differently.

> Error on->  Line :2 Column :16
Message :Namespace prefix XMLNS for NS on TAGNAME is not defined

> Error on->  Line :2 Column :16
Message :Namespace prefix NS on TAGNAME is not defined

deathaxe: Just found that, too. XML-namespaces and some attributes must not be matched case insensitive at all. But this is the case prior to my most recent commit, too. 

But honestly - do you want to introduce a pattern matching all maybe allowed or not allowed case-insensitive tags/attributes? 

Me not.

deathaxe: A browser and most other XML tools like Oxygen XML don't care about it at all, too. They all just highlight those constructs case-sensitive

What you complain about is not syntax highlighting but linting!

This is to be handled by SublimeLinter or Exalt.

PR #989, opened 2017-05-22
[C] [C++] Greedily scope the "dot" as punctuation.accessor

rwols:
## Problem & Current Situation
For implementations of `on_query_completions` in event listeners, it'd be nice to check if the user wants to get completions for a member variable or member function. I think the most aesthetic way would be to do
```python
def on_query_completions(self, prefix, locations):
    point = locations[0]
    if self.view.match_selector(point - 1, 'punctuation.accessor'):
        # OK: user wants members of a namespace/struct/class
```
Now observe that:
* `->` is scoped greedily as an accessor, even when there's no identifier after it. 💯 
* `::` in C++ is scoped greedily as an accessor, even when there's no identifier after it. 💯 
* `.` in C and C++ is not scoped as an accessor when there's no identifier after it. 👎 

This forces plugin writers to use hacks:
```python
def on_query_completions(self, prefix, locations):
    point = locations[0]
    if self.view.match_selector(point - 1, 'punctuation.accessor'):
        # OK: user wants members of a namespace or pointer-to-struct/class
    elif is_dot_present_and_is_not_ellipsis_operator(point - 1): # hacky :-(
        # OK: user typed a single dot
```
## Solution
Scope the dot as `punctuation.accessor` even when there's not an identifier following it yet.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -613,10 +613,10 @@ contexts:
       captures:
         1: punctuation.accessor.c++
         2: variable.other.readwrite.member.c++
-    - match: (?<!\.)\.(?!\.)
-      scope: punctuation.accessor.c++
     - match: \.\.(?!\.)
       scope: invalid.illegal.syntax.c++
+    - match: \.(?!\.)
+      scope: punctuation.accessor.c++
 
   typedef:
     - match: \btypedef\b
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -609,10 +609,11 @@ contexts:
               scope: meta.method-call.c++ meta.group.c++ punctuation.section.group.end.c++
               pop: true
             - include: expressions
-    - match: '(\.|->)({{identifier}})(?!\s*\()'
+    - match: '(\.|->)({{identifier}})(::)?(?!\s*\()'
       captures:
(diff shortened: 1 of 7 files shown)


wbond:
These changes will also need to be made to Objective-C/C++

rwols:
I added another case where a greedy accessor might be useful, namely if you have a base class `Base` and a derived class `Derived`, and you want to access the base class' members explicitly, like so:

```
Derived derived;
derived.Base::
```
then the double colons will also be scoped preemptively as `punctuation.accessor` (this is https://github.com/sublimehq/Packages/pull/989/commits/74a6ccfe9547414a78f585d21603d6a23be12256).

At this point I think all possible cases for C/C++ are accounted for.

As for Objective-C, the language doesn't use characters for accessing members but rather it uses whitespace:

```
NSSstring* str = [[NSString alloc] init]
[str characterAtIndex:0]
//  ^ punctuation.accessor ?
```
I could maybe add this to Objective-C to make the PR really complete, but I don't have enough real-world experience with Objective-C to know if scoping whitespace as punctuation.accessor would be a good idea. Comments are welcome.

wbond:
I think leaving Objective-C spaces as-is is what most people would expect – no need to do anything there.

Thanks for your work on this!

PR #993, opened 2017-05-24
[LaTeX] Disable prototype inside the verb command

r-stein:
Comments (and other potential prototypes) should be disabled inside `\verb`-commands as done in this PR.

Fix https://github.com/SublimeText/LaTeXTools/issues/1135


wbond:
Should there be a test to ensure this doesn't regress?

r-stein:
I have added a test for this.

wbond:
Thanks!

PR #994, opened 2017-05-25
[C#] describe indentation rules and fix when a comma comes after a `}`

keith-hall:
This PR describes (with comments) the indentation rules used for the C# syntax, and fixes it so that the following gets reindented properly:

```cs
new [] {
    new Point {
        X = 5,
        Y = 10
    },
    new Point {
        X = 7,
        Y = 9
    },
}
```

(previously, the `},` wouldn't have it's indentation decreased...

I also removed a rule that said to decrease indentation on lines that only contain whitespace followed by `{}` because it doesn't make any sense.

### DIFF
--- a/C#/Indentation.tmPreferences
+++ b/C#/Indentation.tmPreferences
@@ -9,20 +9,34 @@
 	<dict>
 		<key>decreaseIndentPattern</key>
 		<string>(?x)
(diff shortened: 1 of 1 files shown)


wbond:
Thanks!

PR #999, opened 2017-05-26
[Python] Introduce meta.{mapping,sequence,set}

FichteFoll:
Now distinguishes between mappings and sets, allowing more fine-grained matching and meta scopes.

Also add meta.mapping.{key,value} scope to the appropriate sections of a dict literal.

This is in line with the proposed changes to JSON in https://github.com/sublimehq/Packages/pull/862.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -982,6 +982,7 @@ contexts:
     - include: comments
     - match: (?=\S)
       push:
+        - clear_scopes: 1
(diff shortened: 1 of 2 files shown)


FichteFoll:
I could probably make the `meta.mapping` scopes non-overlapping, so that we don't have `meta.mapping.python meta.mapping.key.python`. Is this desired?

FichteFoll:
Just a heads-up that I fixed the merge conflicts.

PR #1001, opened 2017-05-27
[Python] Language constants and string placeholder update and

FichteFoll:
Some changes that didn't fit anywhere else and were too small to warrant a separate PR.


wbond:
Thanks!

PR #1002, opened 2017-05-27
[YAML] Update of YAML boolean keywords

zoquda:
If my understanding is correct then `yes`, `no`, `on`, `off` and their various aliases were valid boolean keywords in YAML 1.1, but not anymore in the current YAML 1.2.

Relevant spec: [http://www.yaml.org/spec/1.2/spec.html#id2805071](http://www.yaml.org/spec/1.2/spec.html#id2805071)

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -113,9 +113,7 @@ variables:
   _type_null: (?:null|Null|NULL|~) # http://yaml.org/type/null.html
   _type_bool: |- # http://yaml.org/type/bool.html
     (?x:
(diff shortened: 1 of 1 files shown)


FichteFoll:
Please see what I wrote here: https://github.com/sublimehq/Packages/pull/913#issuecomment-296259419

YAML 1.1 is still widely used, for example in libyaml (and PyYAML). I do like that they were removed in 1.2 and it should've ideally been that way from the start, but removing support for highlighting would be a bit early.

zoquda:
Valid point. I hadn't looked at it from that side yet.

Issue #894, opened 2017-04-11
[JavaScript] (PR provided) End of regex sometimes interpreted as comment


Thom1729:
Tests:

```js
'// /* /**';
//^^^^^^^^^^ - comment

"// /* /**";
//^^^^^^^^^^ - comment

/test// 1;
// <- string.regexp.js
//    ^ keyword.operator.arithmetic.js

/test/* 1;
// <- string.regexp.js
//    ^ keyword.operator.arithmetic.js
// */

/test/** 1;
// <- string.regexp.js
//    ^^ keyword.operator.arithmetic.js
// */
```
In the syntax, the `comments` context is manually included in over twenty places. Presumably, this is a relic of the syntax's heritage as a `.tmLanguage`, which didn't support prototype contexts. Updating this is a step toward fixing #885, and this bug can be fixed in the process.

Issue #885, opened 2017-04-07
[JavaScript] Division incorrectly highlighted as regex. A general fix is known, but nontrivial and could use feedback.


Thom1729:
Examples (note that GitHub gets the second example wrong):

```js
1 /**/ / 2 / /**/ 3;

1
/ 2 /
3;

+function foo () {} /x/i;
```

The root cause seems to be the architecture of `expressions`. The `after-operator` and `after-identifier` contexts work to differentiate uses of the `/` characters in common cases, but they aren't really a general solution. The same architectural issue also holds back #324; the suggested solution there suffers from exactly the same bug.

The third example also suffers from a lack of distinction between statements and expressions.

The general solution I've worked out for this sort of issue is to have separate `expression-head` and `expression-tail` contexts. I've included a minimal example:

```yaml
%YAML 1.2
---
name: Example
scope: source.example

contexts:
  main:
    - match: (?=\S)
      push: [ expression-tail, expression-head ]

  prototype:
    - match: '/\*'
      scope: punctuation.comment.block.begin.example
      push:
        - meta_scope: comment.block.example
        - meta_include_prototype: false
        - match: '\*/'
          scope: punctuation.comment.block.end.example
          pop: true

  expression-tail:
    - match: '/'
      scope: keyword.operator.arithmetic.division.example
      push: expression-head

  expression-head:
    - match: '/'
      scope: punctuation.definition.regex.begin.example
      set:
        - meta_scope: string.regex.example
        - meta_include_prototype: false
        - match: '/'
          scope: punctuation.definition.regex.end.example
          pop: true

    - match: (?=\S)
      pop: true
```

The following are scoped correctly:
```
/abc/ /**/ / /ijk/ / /**/ /xyz/
```

```
/abc/
/ /ijk/ /
/xyz/
```

The `expression-head` context should:

- Match unary prefix operators and continue (do not push, pop, or set).
- Match complete expressions (such as literals or variables) and pop.
- Match opening punctuation (such as `(` or `/`) and set.
- Otherwise, pop.

The `expression-tail` context should:

- Match unary postfix operators and continue.
- Match binary operators and push `expression-head`.
- Otherwise, pop.

As written, this will match one expression and pop everything when the next character is not part of an expression. Slight modification will yield contexts that can parse a list of expressions or gracefully skip invalid input.

---

[...]

Issue #874, opened 2017-04-03
[JavaScript] Regex with escaped backslash is formatted like a comment


ghost:
Multiplying and dividing a regex with escaped backslash by a literal number results in wrong highlighting.

```js
console.log(/\//*1);
```

and

```js
console.log(/\///1);
```

aziz:
While these things are getting fix here, if you want the most correct syntax highlighting of JS try [Ecmascript-Sublime](https://packagecontrol.io/packages/Ecmascript%20Syntax)

It can also parse the other bug you reported correctly.

WsDwA1022:
@aziz Nice, I didn't know for that package. It seems to be the most detailed highlighting engine for JavaScript. However, it took me about a hour to find a [bug](https://github.com/bathos/Ecmascript-Sublime/issues/30) there ;)

PR #1003, opened 2017-05-28
[JavaScript] Use prototype for comments

Thom1729:
### Background

The default JavaScript syntax definition is descended from an old `.tmLanguage`. Since its transition to a `.sublime-syntax`, it has largely modernized, but signs remain of its history.

One such sign is that the syntax definition does not use the new `prototype` feature for comments. Instead, the `comments` context is included manually. There are many dozens of contexts in the definition and somewhere between many and most of them should have manually included `comments`. An unknown number of those did not, resulting in at least one bug (#885, first test case).

In addition, the tests did not include any negative tests to ensure that comments were *not* included where they should not have been. No related bugs are known.

### This commit

I have moved the `comments` include to the prototype and removed it from all other contexts. I added `meta_include_prototype: false` as appropriate to remove it from the few contexts that should not include it. Finally, I have added negative tests to verify that comments should not appear in those contexts and a test for the known bug that was fixed in the process.

### Discussion

The known bug could have been fixed by adding another manual `include`. So why make a larger change?

The primary reason is simplicity. Comments are a pervasive feature that applies to nearly every part of the language. Each `- include: comments` is a separate implementation of this feature. While not every context would need that include, it is necessary for every context to *consider* whether it is needed. In at least one known case, the include was mistakenly omitted (and there are probably more such cases).

Conversely, the parts of the language that do *not* need to support comments are few: strings, regexps, and comments themselves. These parts all stand out as special cases with their own unique microsyntaxes. Additionally, this enumeration is likely to be relatively stable as the language evolves. Every piece of new syntax may require a new context and explicit consideration of comments, whereas new types of strings don't come along very often.

Prototypes also lessen the testing requirements. In this commit, I added a test for each `meta_include_prototype: false` that covers all comment types. Without prototypes, there really ought to be such a test for every single context that *does* allow comments. Many, many tests would be required to catch bugs such as the example from #885. Such tests do not yet exist.

In short, I believe that it makes the most sense to mark each exception rather than each non-exception. Doing so will result in less code and require fewer tests.

### Remarks

- There seem to be three separate copies of single- and double-quoted strings. These could perhaps be consolidated.

- This change is one of the steps on the program of improvement I outlined in #885. I consider this PR to be an important part of the foundation for this program

- In addition, I anticipate that fixes for #874 and #894 will follow quickly. I nearly tossed them into this PR, but it's sufficiently separate that it's worth submitting separately.


Thom1729:
Will resubmit this with the merging done differently. (I'm still fairly new to git.)

Issue #831, opened 2017-03-17
[JavaScript] Should the comma operator be highlighted as an operator?


Thom1729:
In JavaScript, the comma is most often used as separator punctuation: in array and object literals, function formal parameters, function arguments, and variable declarations. However, the language also has a comma operator. It is described as an operator in the spec and it behaves like any other binary operator.

The built-in JavaScript syntax does not distinguish between the comma operator and the other uses of the comma. One consequence of this is that two very different constructions may be highlighted identically:

```js
myVar;
[1, 2, 3];
```
```js
myVar
[1, 2, 3];
```

In the former example, the commas are punctuation separating array elements. In the latter example, the commas are operators. There is an enormous semantic difference between these examples. In the latter case, because the commas are not highlighted as operators, someone reading the code is likely to misinterpret it as equivalent to the former case.

One might argue that the latter is probably a mistake by the programmer. I concur, but to me this makes it even more important to clearly distinguish the comma operator.

It seems to me that the comma operator should be marked as `keyword.operator.comma.js` (while other uses of the comma remain `punctuation.separator.comma.js`).

FichteFoll:
Could you explain what exactly the "comma operator" it is and when it is used? So far, it appears to be for slicing, but that's just a guess.

Thom1729:
The comma operator takes two arguments and evaluates to the latter. It does not short-circuit.

```js
var x;
x = (1,2,3);
console.log(x); // 3
```

JavaScript doesn't have a slice syntax. The statement `myVar[1, 2, 3];` is equivalent to `myVar[3];` (because the expressions `1` and `2` have no side effects). Deliberate use of the comma operator is fairly rare in written JavaScript, although I see it more often in minified code. *Accidental* use of the comma operator as a result of missing semicolons is probably more common. Distinguishing the comma operator from the comma punctuation that is used in function calls etc. would visually distinguish constructions with identical text but very different semantics.

wbond:
There are many little edge cases like these that come up when trying to write syntaxes for different languages, but sharing a set of scope name conventions.

In this case, I think most users would find having some instances of `,` scopes as an operator and others as a separator very odd, and would likely consider it a bug.

For the default JavaScript syntax I don't think we'll want to do this.

FichteFoll:
>In this case, I think most users would find having some instances of , scopes as an operator and others as a separator very odd, and would likely consider it a bug.

[...]

Issue #324, opened 2016-04-28
[JavaScript] possible enhancement: distinguish uses of square brackets


jcberquist:
With the addition of the `after-identifier` context, it looks to me like it is possible to distinguish the use of square brackets in a property accessor (or array index) and square brackets used to denote a literal array. Does that seem right? If so, is that something you would be interested in adding?

wbond:
Yes, that seems like a reasonable addition. I'll likely be adding such support to C/C++ and PHP soon also since those are now in better shape for more context-specific enhancements.

jcberquist:
I added a commit that implements this on a basic level: https://github.com/jcberquist/Packages/commit/3240bcb5ea4162e5acc2a4a8c0efbc93adde9088

I didn't create a pull request because I wasn't sure of the scopes you wanted to use, or if this was a satisfactory way to go about it (especially since you plan to do the same in C/C++ and PHP).

keith-hall:
The scoping for this is being discussed here: https://github.com/sublimehq/Packages/issues/790

Thom1729:
Counterexample to jcberquist@3240bcb:

```js
1 /**/ [a]
```
The `[a]` is incorrectly scoped `meta.array-literal.js`. The cause is the same as #885.

Issue #884, opened 2017-04-05
[RFC] More specific subscopes (e.g. meta.function.name)


Thom1729:
From the [scope naming guidelines](https://www.sublimetext.com/docs/3/scope_naming.html#meta):

> The entire scope of a function should be covered by one of the following scopes. Each variant should be applied to a specific part, and not stacked. For example, `meta.function.php meta.function.parameters.php` should never occur, but instead the scopes should alternate between `meta.function.php` then `meta.function.parameters.php` and back to `meta.function.php`.
>
> - `meta.function`
> - `meta.function.parameters`
> - `meta.function.return-type`

This is an excellent system, and I think that it can be extended gracefully to cover other parts of a function. The goal would be to provide standard subscope names so that syntaxes that exceed the minimum specificity requirements are compatible with each other.

An important point raised by wbond is that "the more specific it gets, the more difficult to standardize due to language semantics and differences." Any additions should be conservative; there's little point in standardizing a construct that only appears in a single common language. However, universality is not required; for instance, many languages will never use `meta.function.return-type`, but enough do that it is well worth having a standard for it.

The following sorts of constructs seem like good candidates for a standardized subscope:

-   `meta.function.name`. In many languages, this may always be marked with `entity.name.function`, which would seem to make this scope redundant. But this is not always true. Take JavaScript:

    ```js
    class Foo {
        myFunc() {}
        2e5() {} // The name of this function is 200000
        'another function'() {}
        [getFunctionName()]() {}
    }
    ```

    Or SQL:

    ```sql
    create function "foo" ...
    ```

-   `meta.function.body`. In some languages, such as C, this is delimited by punctuation like braces. In other languages, such as Python, it is not delimited by punctuation at all. In yet other languages, such as JavaScript, it varies:

    ```js
    (function() { return 42; });
    (() => { return 42; });
    (() => 42);
    ```

-   `meta.function.access`. In many languages, this may be more than one word, such as C#:

    ```c#
    class Foo {
        protected internal void Bar() {}
    }
    ```

-   `meta.function.type-parameters`. Many languages explicitly list type or template parameters, such as Java:

    ```java
    class Foo {
        void Bar<T>() {}
    }
    ```

[...]

PR #1009, opened 2017-05-29
[JavaScript] New structure for expressions

Thom1729:
### Background

In order to correctly parse JavaScript expressions, you must always know whether you are at the beginning or end of an expression. Examples:

```js
/abc/ /def/ /ghi/;
// <- string.regexp
//    ^ keyword.operator.arithmetic
//     ^^^ variable.other.readwrite
//          ^ keyword.operator.arithmetic
//          ^^^^^ string.regexp

x // expression continues
/abc/ /def/ /ghi/ y;
// <- keyword.operator.arithmetic
//^^ variable.other.readwrite
//  ^ keyword.operator.arithmetic
//    ^^^^^ string.regexp
//          ^ keyword.operator.arithmetic
//           ^^^ variable.other.readwrite
//              ^ keyword.operator.arithmetic

{} // semicolon not required
// <- meta.block

+ // expression continues
{};
// <- meta.object-literal

function foo() {} // semicolon not required after function declaration
/42/;
// <- string.regexp

+ // expression continues
function foo() {} // expression continues
/42/ z;
// <- keyword.operator.arithmetic
```

All of these cases can be handled exactly correctly by sublime's parsing engine. This is not a case like arrow functions where perfect parsing is impossible due to engine limitations. The premise for this work is that the JavaScript syntax therefore *should* handle these distinctions perfectly.

### Current state

The existing syntax tries to maintain these distinctions, but it does not do so systematically. For instance, the distinction between division and regexps is in many cases determined using a lookahead. This approach is buggy (#874, #885, #894).

In addition, the distinction is not made in some cases where it may be desirable. For instance, the syntax does not distinguish between array literals and bracketed property accesses (#324):

```jsx
[item];
// <- meta.brackets

x
[prop];
// <- meta.brackets
```

As a result, it is impossible to distinguish the rare comma operator from the more common comma separator punctuation (#831). An implementation of the distinction between arrays and property accesses has been attempted, but that implementation is vulnerable to the same sorts of bugs as the regex/division implementation and for the same reason (https://github.com/jcberquist/Packages/commit/3240bcb5ea4162e5acc2a4a8c0efbc93adde9088).

Since the current syntax was adopted for the core package, there have been a number of tweaks and patches to this behavior. Most notable is the addition of the `after-operator` and `after-identifier` contexts (https://github.com/sublimehq/Packages/commit/62d142a416d44d18785bf03b5a1ded1badcd8577). These contexts attempt to maintain the critical distinction between the beginning and end of an expression. They do so in a way that is relatively noninvasive. Unfortunately, because they are noninvasive, they do not and cannot solve the problem in the general case. While these contexts are a substantial improvement, bugs remain, and so do various patches that attempt to handle regexp corner cases.

### The solution

The core principle of the solution is that there must be separate states for the beginning and end of an expression. That is, when parsing `x / y;`, the parser must move back and forth between a state that expects constructs like numbers and regexps and a state that expects constructs like binary operators. When the parser is in one state, a slash will be interpreted as the beginning of a regexp, and when it is in the other state, it will be interpreted as the division operator.

The `after-operator` and `after-identifier` contexts represent such a system, but as implemented they are insufficient. The entire expression syntax must be written in this tick-tock fashion; it cannot be bolted on or it will not work in the general case. On the other hand, if expressions are designed in this way from the ground up, then these knotty syntax problems become simple and no patches or special cases will be needed.

After much trial and error, I have come up with a framework that handles this efficiently and effectively. To demonstrate this framework, here is a toy example:

```yaml
%YAML 1.2
---
name: Toy Syntax
scope: source.toy

contexts:
  main:
    - match: ';'
      scope: punctuation.terminator
    - match: (?=\S)
      push: [ expression-end, expression-begin ]

  expression-begin:
    - match: \d+
      scope: constant.numeric
      pop: true

    - match: /
      set:
        - meta_scope: string.regexp
        - match: /
          pop: true

  expression-end:
    - match: (?=;)
      pop: true
    - match: '[-+*/]'
      scope: keyword.operator.arithmetic
      push: expression-begin
```

This example has numeric literals, regexp literals, and the four common arithmetic operators. It will perfectly distinguish between division and regexps. (Try `/abc/ /def/ /ghi/;`, and add line breaks wherever you please.)

There are several key points that deserve explanation.

First, the expression contexts parse a single expression. The contexts are highly stack-oriented, pushing and popping after nearly every match. In the end, when a complete expression has been consumed, the contexts will both pop off the stack. This means that whatever parent context uses expressions must push them rather than including them.

Second, they parse a single *expression*. Terminating semicolons are not part of an expression. Nor are JavaScript constructs like `throw` and `return`. Nor, in JavaScript, are function declarations. This is critical: a function declaration is a complete statement that does not need to be terminated by a semicolon, but an expression does need termination. The examples at the top include an illustration of this distinction.

Third, we are pushing and popping rather than setting back and forth between `expression-begin` and `expression-end`. There are several reasons for this. One is that in this scheme, `expression-begin` does not have to refer explicitly to `expression-end`; it merely pops to get there. This is very important, because it allows us to have specialized versions of `expression-end`. In JavaScript, I use one of these for expressions that may not contain commas at top level (such as for function arguments, list items, or lambda concise body expressions). Another is used to implement the automatic semicolon insertion algorithm for expression statements. If `set` were used instead of `push` and `pop`, then we would need duplicate copies of `expression-begin`. Compare [Ecmascript Sublime](https://github.com/bathos/Ecmascript-Sublime), which uses many redundant almost-copies of contexts and in doing so exceeds 5,000 lines.

Fourth, every contexts that wants expressions does have to push both `expression-end` and `expression-begin` in the correct order. However, this can be easily shortcut for convenience:

```yaml
  main:
    - match: ';'
      scope: punctuation.terminator
    - match: (?=\S)
      push: expression

  expression:
    - match: (?=\S)
      set: [ expression-end, expression-begin ]
```

Fifth, we don't have any error handling. This is easily remedied in practice; each of `expression-begin` and `expression-end` should pop if no match succeeds. In practice, I have found pushing and popping in this manner to be far more resilient to invalid input than setting.

This toy example does little, but I hope that it can be seen how this framework can trivially distinguish blocks from objects, arrays from property accesses, function invocations from parenthesized expressions.

### Implementation

In order to implement the above approach in the existing JavaScript syntax, I made a number of changes.

Zeroth, I moved comment logic to a prototype (#1007). This cleaned up the other scopes and fixed at least one bug.

First, I separated `statements` from `expressions` by handling semicolons in `statements` and popping `expressions` whenever one was encountered.

Second, I moved non-expression constructs like variable declarations, `throw`, and so forth from `expressions` to `statements`. In the process, I implemented restricted productions in a general way and removed the existing implementations (that failed in some cases).

Third, I implemented "expression statements", which are basically just expressions that handle automatic semicolon insertion. I added class and function declarations to `statements`.

Fourth, I made the large change of implementing the two-level expression framework. I:

  - Created the `expression-begin` and `expression-end` contexts, and made `expressions` set them.
  - Split the former contents of `expressions` between them as appropriate.
  - Caused those contents to push or pop as appropriate.
  - Broke out `unary-operators` into `prefix-operators` and `postfix-operators`.
  - Broke out `round-brackets` into `round-brackets` and `parenthesized-expression`.
  - Broke out `square-brackets` into `square-brackets` and `bracketed-property-access`.
  - Added `expression-no-comma` for various uses.
  - Removed `after-identifier` and `after-operator`.
  - Changed various places that had included `expressions` to instead push it.
  - Fixed various tests that specified incorrect behavior.

Fifth, I renamed `expressions` to singular `expression`, reflecting its new behavior.

As a result of these changes, expressions universally use the two-context framework. All tests currently pass, including some that needed modification because the previous behavior was incorrect. Issues #874, #885, and #894 are fixed.

### Next steps

I write this PR early, as soon as the change is fundamentally implemented and before refactoring clutters the diffs. More work should be done to clean up the syntax and take advantage of the new expressions framework:

  - Most of the match anything/push expressions can and should be refactored to something cleaner, generally by using the stack more.
  - The match anything/pop behavior should probably be factored out and included.
  - Much of the lookahead logic for constructs like `x = function() {};` should be refactored for simplicity. I would not be surprised if bugs were hiding there after the big change.
  - Arrays should be rescoped (#324). This is a trivial change.
  - The line continuation lookahead probably has bugs and should be carefully reviewed.

I anticipate that these changes will decrease the size and complexity of the syntax as well as likely fixing bugs. In addition, I see opportunities for future improvement:

  - Better scoping for the inside of an object literal.
  - Specific-enough meta scopes that we may be able to provide lexically-aware completions (e.g. all variables in lexical scope).
  - Easy JSX support. (It would be simple enough to provide versions both with and without JSX. The key difference would be a single `include` in `expression-begin`.)
  - (#831) Marking the comma operator as an operator (leaving comma punctuation as punctuation).

For now, I would appreciate feedback on the general approach and the specific implementation. I am aware that this is a major change to the structure of the syntax and that we will certainly want to add more tests. I will be available to fix any bugs that may come out of it in the forseeable future.


wbond:
It seems in the process of merging your other PRs, this one got conflicted. Would you mind resolving the conflict for me?

Thom1729:
I'll try to resolve the conflicts and wrap this up by the end of Saturday. All tests pass (at least before the recent merges), but there are a couple of oddities in the test file that aren't covered.

Thom1729:
That's probably correct; I'll verify after lunch. The only new change is fixing a test that had a missing semicolon.

Thom1729:
That should be it, I think. I've fixed the remaining weirdness; it turns out there was a tiny regex typo that interacted badly with the new structure.

jcberquist:
@Thom1729 I just want to point out a small issue you might want to take a look at: 
https://github.com/Thom1729/Packages/blob/1bd37b28434a51a17c3f3c25ec5b521fb9e00e8f/JavaScript/syntax_test_js.js#L484 
If you examine the scopes on that line you will see that the colon on that line is now being matched as a ternary operator.

jcberquist:
@Thom1729 I am enjoying going through this :) I think I may have found another thing to look at:
https://github.com/Thom1729/Packages/blob/3de201f94eb7d430debfa70687bbaac8f1ee36f5/JavaScript/syntax_test_js.js#L446
I think the `var` there is not getting scoped as `storage.type.js`.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -454,19 +454,19 @@ contexts:
...
-        - match: (?=\S)
-          push: expression
+        - match: \b(const|let|var)\b
+          scope: storage.type.js
+
+        - include: expression-list
     - include: block-scope
 
   block-scope:
-    - match: '\}'
```
jcberquist: I am curious about this change - Removing these lines (462-464) means the `meta.for` scope (or anything that uses `parens-block-scope`) won't pop off immediately after the `}`, but rather when the next non-whitespace character is encountered.

jcberquist:
I have found a few other things (Using your latest `syntax_test_js.js` version).

[Line 54](https://github.com/Thom1729/Packages/blob/c91a31df3cb6c4efa8ecc4db8eb7f5339936248f/JavaScript/syntax_test_js.js#L54)
```javascript
export default function (a) { }
```
 - `meta.export` doesn't pop off after the `}`, while it does in the current syntax
 - `function` doesn't have the scope `meta.anonymous.function`, but it does in the current syntax
 - this is more of a question...both the original syntax and the rewrite scope the space between `)` and `{` with `meta.block` - that doesn't seem quite right to me?

[Line 198](https://github.com/Thom1729/Packages/blob/c91a31df3cb6c4efa8ecc4db8eb7f5339936248f/JavaScript/syntax_test_js.js#L198)
```javascript
 '// /* not a comment'() {},
```
 - the `()` gets scoped `meta.function.declaration.js` twice - see also [Line 286](https://github.com/Thom1729/Packages/blob/c91a31df3cb6c4efa8ecc4db8eb7f5339936248f/JavaScript/syntax_test_js.js#L286) for an example of doubled up scopes

[Line 268](https://github.com/Thom1729/Packages/blob/c91a31df3cb6c4efa8ecc4db8eb7f5339936248f/JavaScript/syntax_test_js.js#L268)
```javascript
    "key4": true,
```
 - only the opening `"` is geting the `meta.object-literal.key` scope

Thom1729:
> I am curious about this change - Removing these lines (462-464) means the meta.for scope (or anything that uses parens-block-scope) won't pop off immediately after the }, but rather when the next non-whitespace character is encountered.

Fixed -- although this is a hack. Probably the right thing to do is something like this:

```yaml
    - match: \bfinally\b
      scope: keyword.control.trycatch.js
      push:
        - - meta_scope: meta.finally.js
          - include: immediately-pop
        - block-scope
```

I'm not sure that this is the right time to fix this everywhere, though.

> meta.export doesn't pop off after the }, while it does in the current syntax

Fixed.

> function doesn't have the scope meta.anonymous.function, but it does in the current syntax

This is a weird case. I'm not sure what the point of the `meta.anonymous.function` scope is. It can't possibly be implemented correctly within the constraints of the sublime parser, because that would require lookahead across line boundaries. In addition, whether a function has a name or not is virtually never relevant.

In this case, we're not going through the usual function expression code because `export default function` is more like a function declaration than a function expression. For instance, `export default <expression>;` requires a terminating semicolon, but `export default function ...` does not. However, unlike an ordinary function declaration, a name is not required.

I can add in the same `meta.anonymous.function` logic as is used for function expressions, but personally I'd rather remove it from elsewhere instead. I wonder if anyone would care either way.

> this is more of a question...both the original syntax and the rewrite scope the space between ) and { with meta.block - that doesn't seem quite right to me?

Me neither. This might be a good thing to fix as part of a broader improvement of meta scopes.

> the () gets scoped meta.function.declaration.js twice - see also Line 286 for an example of doubled up scopes

Fixed.

> only the opening " is geting the meta.object-literal.key scope

Fixed.

jcberquist:
I think some of the function and class lookaheads are going to need word boundaries added. Consider the following:
```javascript
classes = true;
```
Without the boundary the `(?=class)` lookahead is matched. (I encountered this issue perusing the JQuery source.)

jcberquist:
Looking at [syntax_test_js.js#L786](https://github.com/Thom1729/Packages/blob/aaa3cc9f03df3210e5be5005e9fdfb248b65107c/JavaScript/syntax_test_js.js#L786)

I see `meta.instance.constructor.js` and `meta.function-call.constructor.js` duplicated at the `(` - I expect this is more of the same fix you did above.

jcberquist:
Looking at [syntax_test_js.js#L373](https://github.com/Thom1729/Packages/blob/aaa3cc9f03df3210e5be5005e9fdfb248b65107c/JavaScript/syntax_test_js.js#L373)

The `{` is getting the `meta.function.declaration.js` scope twice - should it be getting it at all?

Also, for the object literal method declarations the whole `{...}` block is getting the `meta.function.declaration.js` scope. (Possibly related to the above?)

Thom1729:
Fixed all of the above.

Some of the more complicated contexts (e.g. `object-literal`) could really use a complete rewrite in a more stack-based style with proper meta scopes and such. I'm a bit hesitant to do so unnecessarily at this point, though.

jcberquist:
It looks to me like the `meta.function.declaration.js` scope is still getting applied to some method blocks, and sometimes it extends beyond the `)` ending the parameters up to the opening `{`.  (To be clear the current syntax has some of these issues as well.)

You might be doing something like this already, but I am finding that a helpful way to review where a scope shows up is to do something like the following via the console:
```python
view.add_regions('scope_highlight', view.find_by_selector('meta.function.declaration'), 'source')
```

Thom1729:
Well, I think I'm convinced that I should go ahead and rewrite a couple of the more convoluted contexts. The way the syntax was originally set up, all of these meta scopes are specified repeatedly in contexts that set each other. I prefer to instead push a stack of contexts including a single-purpose context that provides a meta scope and immediately pops. I've been trying to work around the existing architecture, and I think I just need to do a bit more bulldozing. Should hopefully have a comprehensive fix tomorrow.

Thom1729:
I've rewritten the function declaration and object literal internals. Hopefully, there should be few-to-no duplicate contexts remaining.

The `meta.function.anonymous` context has been removed. If there is demand for it or something like it, it can be added. See #1071.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -914,83 +902,65 @@ contexts:
...
-        2: keyword.generator.asterisk.js
-    - match: '{{identifier}}'
-      scope: entity.name.function.js
-    - include: function-declaration-parameters
-    - match: '(?=\s*\{)'
-      set: function-block
+    - match: ''
+      set:
+        - - include: immediately-pop
+        - function-block
```
jcberquist: Since the `function-block` context currently has a `meta_scope` of `meta.block.js`, doesn't setting it into this list mean that everything set after it (so to speak) will get the `meta.block.js` scope? (e.g. if I have `function a() {}`, all of it will be scoped with `meta.block.js`. Perhaps the `function-block` scope should be reverted to applying the meta scope on the pushed context, and then on the ending `}` again?

Thom1729: Fixed. Incidentally, I think that the whole thing should be under a `meta.function`, including the block. Not sure why it hasn't been, but not sure if I should change it at this point.

jcberquist: Yeah, I would like that - but it isn't that way now so I agree it might be better to wait on that. I think you brought this up in https://github.com/sublimehq/Packages/issues/884 but as far as I know there isn't a definitive decision?

JavaScript/JavaScript.sublime-syntax
```diff
@@ -934,6 +933,10 @@ contexts:
     - match: '(?=\s*\{)'
       set: function-block
 
+  meta-function-declaration:
```
jcberquist: You added this and then decided not to use it?

JavaScript/JavaScript.sublime-syntax
```diff
@@ -1057,7 +1060,28 @@ contexts:
...
           pop: true
-        - include: comma-separator
+
+        - match: >-
+            (?x)(?=
+              {{method_name}}\s*:
+              (?: {{func_lookahead}} | {{arrow_func_lookahead}} )
+            )
+          push:
+            - - meta_scope: meta.function.declaration.js
```
jcberquist: Putting this meta scope here means the `meta.function.declaration.js` gets added to the function body as well - and causes duplicate scoping as well. This can be seen by examining the scopes in the following example:
```javascript
var test = {
    a: x => x * 2,
    b: function() {}
}
```

Thom1729: Yep; it needed to be swapped with the following context.

JavaScript/JavaScript.sublime-syntax
Thom1729: Yeah, because I kept running into https://github.com/SublimeTextIssues/Core/issues/1577. I should probably delete it.

jcberquist: I was wondering about that - I had just been looking at that issue.

JavaScript/JavaScript.sublime-syntax
jcberquist: Could you do the same for the `method-declaration` context?

jcberquist:
This looks really good to me. I would like to point out a few other minor things:
```javascript
var test = {
    a: true ? 1 : 0,
    b: something
}
```
Should the `ternary-operator` context use `expression-no-comma`? Right now the `b` above gets scoped as a variable.

```javascript
new function() {}();
new class Foo() {}();
```
The current syntax includes the scope `meta.function-call.constructor.js`  (on the `()`) but the rewrite does not.

These next examples are more to point out differences in the syntaxes:

```javascript
var instance = new foo()

```
After the `()` (including after the newline) the current syntax has the scope `source.js meta.instance.constructor.js` while the rewrite also includes `meta.function-call.constructor.js` (including after the newline).

```javascript
function() {}
```
I noted this before but I just want to mention it again - in the rewrite everything after `)` and before `{` gets the `meta.function.declaration.js` scope - the current syntax generally (I think) pops it off after the `)`.

Thom1729:
> Should the ternary-operator context use expression-no-comma?

Yes; fixed.

> The current syntax includes the scope meta.function-call.constructor.js (on the ()) but the rewrite does not.

> ...

> After the () (including after the newline) the current syntax has the scope source.js meta.instance.constructor.js while the rewrite also includes meta.function-call.constructor.js (including after the newline).

I've now completely rewritten `constructor`. It should behave sensibly now.

> I noted this before but I just want to mention it again - in the rewrite everything after ) and before { gets the meta.function.declaration.js scope - the current syntax generally (I think) pops it off after the ).

I've restored the old behavior, although I'm not convinced that it's correct. I think the `meta.function` scope should probably span the entire contstruct.

jcberquist:
The order of the contexts in `method-declaration` needs the same swap you did for the function declaration - right now `meta.function.declaration.js` is too extensive (see for example [syntax_test_js.js#L373](https://github.com/Thom1729/Packages/blob/602561afbdec16c089bc3041251fb0833621bac0/JavaScript/syntax_test_js.js#L373) where is applied to the block).

This is a completely minor point but I noticed that in the following example:
```javascript
var obj = {
    $key1: 0,
    $key2: function() {}
}
```
"$key1" has the scope `meta.object-literal.key.dollar.js` while "$key2" has both `meta.object-literal.key.dollar.js` and `meta.object-literal.key.js` (is this even something to care about?).

jcberquist:
I notice that `expression-end` still has `- include: literal-punctuation`, but you removed that context.

Thom1729:
> The order of the contexts in method-declaration needs the same swap you did for the function declaration

Fixed.

> "$key1" has the scope meta.object-literal.key.dollar.js while "$key2" has both meta.object-literal.key.dollar.js and meta.object-literal.key.js (is this even something to care about?).

It probably is, although it looks like handling it is going to add duplicate code. (I am *not* a fan of the weird dollar-sign special-casing, and this is one reason why.)

> I notice that expression-end still has - include: literal-punctuation, but you removed that context.

Fixed.

jcberquist:
In the `method-name` context should the identifier matches include `pop: true`? That would match what the string matches do since they set and then pop, correct?

Thom1729:
Yep. There used to be places where it wasn't safe to change that for fear of breaking existing code, but I suspect that that isn't one of them anymore.

Thom1729:
Semi-relatedly, there are a lot of places that look like this:

```yaml
parenthesized-expression:
  - match: \(
    scope: punctuation.section.group.js
    set:
      - meta_scope: meta.group.js
      - match: \)
        scope: punctuation.section.group.js
        pop: true
      - match: (?=\S)
        push: expression
```

That should really look like this:

```yaml
parenthesized-expression:
  - match: \(
    scope: punctuation.section.group.js
    set:
      - - meta_scope: meta.group.js
        - match: \)
          scope: punctuation.section.group.js
          pop: true
        - include: else-pop
      - expression
```

As-is, you could have odd behavior with some invalid syntax. Of course, I wish that I had my macros to work with so that I could write it like this:

```yaml
parenthesized-expression:
  - match: \(
    scope: punctuation.section.group.js
    set:
      - !meta meta.group.js
      - !expect [ ')', punctuation.section.group.js ]
      - expression
```

Anyway, I'll probably leave these alone for the time being. If anything around them starts to look buggy, then I'll go ahead and rewrite them.

jcberquist:
I am just starting to wrap my head around this approach; I think I see your point above. 

What you did with the `function-declaration` and `arrow-function-declaration` contexts a few commits back is really slick, I love it!

Thom1729:
Thanks. Although the changes are far more drastic now than I had originally hoped, I am very happy with the way it's turning out. But alas, the lack of macros! This:

```yaml
function-declaration:
  - match: ''
    set:
      - - match: (?=\S)
          set: function-block
        - include: else-pop
      - - meta_scope: meta.function.declaration.js
        - include: immediately-pop
      - - include: function-declaration-parameters
        - include: else-pop
      - - match: '{{identifier}}'
          scope: entity.name.function.js
          pop: true
        - include: else-pop
      - - match: \b(function)\b\s*(\*)?
          captures:
            1: storage.type.function.js
            2: keyword.generator.asterisk.js
          pop: true
        - include: else-pop
      - - match: '\basync\b'
          scope: storage.type.js
          pop: true
        - include: else-pop
```

Could be this:

```yaml
function-declaration: !stack
  - !stack function-block
  - !meta meta.function.declaration.js
  - - include: function-declaration-parameters
    - include: else-pop
  - !expect_identifier entity.name.function.js
  - !expect      [ '\*',     keyword.generator.asterisk.js ]
  - !expect_word [ function, storage.type.function.js ]
  - !expect_word [ async,    storage.type.js ]
```

jcberquist:
I think that the match then pop behavior of the `language-identifiers` and `well-known-identifiers` contexts is causing an issue with the way they are used in certain contexts:
```javascript
this.a = function() {}
```
The `.` above is not getting the `punctuation.accessor.js` scope.

jcberquist:
[JavaScript.sublime-syntax#L658](https://github.com/Thom1729/Packages/blob/1ebfdfe976e0200867e2e0945d19a23b869da93c/JavaScript/JavaScript.sublime-syntax#L658) is missing a wrapping group around the operators inside the `\b` markers:
```yaml
    - match: \bdelete|typeof|void\b
```
This means the `\b` markers are included with the first and last words and do not wrap each word.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -15,6 +15,22 @@ variables:
...
+      | \[ {{identifier}} (?:\.{{identifier}})* \]
+    )
+
+  line_continuation_lookahead: >-
+    (?x)
+    (?! \+\+ | -- )
+    (?=
+      != |
+      [ -+*/% ><= &|^ \[( ;,.:? ]
+    )
```
rwols: Could you explain why this `line_continuation_lookahead` exists? Why can't you check for a backslash followed by a newline character?

jcberquist: I think this is here to handle expressions split over multiple lines. e.g. if one has 
```javascript
x = 5
+ 3
```
that is one expression, but 
```javascript
x = 5
y
```
is two. 

So this lookahead determines whether the expression continues after the new line, or if it should be terminated (i.e. implicit semicolon).

rwols: Ah, excellent, thanks for the explanation.

Thom1729: It has to do with JavaScript's automatic semicolon insertion algorithm.

JavaScript expression statements end in a semicolon. However, the parser effectively makes semicolons optional in many cases by automatically inserting them under some circumstances. This means that a newline may represent the end of a statement depending on the contents of the following line. This lookahead examines the following line to see whether the expression continues.

Thom1729: Examples:

```js
x
y; // Two statement expressions: `x` and `y`

x
+
y; // One statement expression: `x + y`

x
++
y; // Two statement expressions: `x` and `++y`.

x
/y/g; // One expression: `(x / y) / g`.
```

jcberquist:
A couple of quick clean up suggestions:

- Do you want to remove the `meta-object-literal-key` context [(L1079-L1081)](https://github.com/Thom1729/Packages/blob/e10e465adc96d220b9fd1b978eb8ba2ab9d6adc3/JavaScript/JavaScript.sublime-syntax#L1079-L1081) since you didn't end up using it?

- The scope name `invalid.illegal.stray-bracket-end` on [L481](https://github.com/Thom1729/Packages/blob/e10e465adc96d220b9fd1b978eb8ba2ab9d6adc3/JavaScript/JavaScript.sublime-syntax#L481) is missing the trailing `.js`

rwols:
I decided to give this one a spin today.

Disclaimer: *I have never programmed in Javascript*.

That said, I've noticed a few quirks:

1. [These few lines](https://github.com/facebook/react/blob/master/src/renderers/dom/fiber/ReactDOMFiberComponent.js#L324-L329) seem to be a regression.
2. [Regression here too?](https://github.com/facebook/react/blob/master/src/renderers/dom/fiber/ReactDOMFiberComponent.js#L311) It looks like this is a function that specifies its return type explicitly, the return type is highlighted incorrectly.
3. [On this line](https://github.com/facebook/react/blob/master/src/renderers/dom/fiber/ReactDOMFiberComponent.js#L26), `DOCUMENT_NODE` is not highlighted, but on [this line](https://github.com/facebook/react/blob/master/src/renderers/dom/fiber/ReactDOMFiberComponent.js#L145), it *is* highlighted. On [this line](https://github.com/facebook/react/blob/master/src/renderers/dom/fiber/ReactDOMFiberComponent.js#L333) it's also highlighted.
4. [More of (1) and (2) it seems](https://github.com/facebook/react/blob/master/src/renderers/dom/fiber/ReactDOMFiberComponent.js#L398-L403).

JavaScript/JavaScript.sublime-syntax
```diff
@@ -25,205 +41,197 @@ contexts:
...
+
+  comma-separator:
+    - match: ','
+      scope: punctuation.separator.comma.js
+
   keywords-top-level:
     - match: \bimport\b
-      scope: meta.import.js keyword.control.import-export.js
-      push: import-extended
+      scope: keyword.control.import-export.js
```
rwols: Just keyword.control.import would do here ...

JavaScript/JavaScript.sublime-syntax
```diff
@@ -25,205 +41,197 @@ contexts:
...
+      push:
+        - import-meta
+        - import-export-final
+        - import-extended
+
     - match: \bexport\b
-      scope: meta.export.js keyword.control.import-export.js
-      push: export-extended
+      scope: keyword.control.import-export.js
```
rwols: ... and keyword.control.import.export here ...

JavaScript/JavaScript.sublime-syntax
```diff
@@ -25,205 +41,197 @@ contexts:
...
+      push:
+        - export-meta
+        - export-extended
+
     - match: \b(export|default|from|as)\b
       scope: keyword.control.import-export.js
```
rwols: ... and keyword.control.import.export here.

Thom1729:
That file isn't pure JavaScript; it uses Flow annotations to declare types. The default JavaScript syntax does not support Flow.

I do think it's worth discussing fallback for nonstandard syntax. Could there be a graceful way to "skip" Flow and TypeScript declarations in the new architecture? Maybe. Personally, I'm content to punt on nonstandard extensions issue to keep the core syntax clean.

I'm working on a new JavaScript package that leverages [YAML Macros](https://github.com/Thom1729/YAML-Macros) to extend this syntax with support for JSX and (eventually) Flow. In the meantime, [babel-sublime](https://github.com/babel/babel-sublime) already supports them.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -865,82 +900,62 @@ contexts:
...
+        - - include: function-declaration-parameters
+          - include: else-pop
+        - - match: '{{identifier}}'
+            scope: entity.name.function.js
+            pop: true
+          - include: else-pop
+        - - match: \b(function)\b\s*(\*)?
+            captures:
+              1: storage.type.function.js
+              2: keyword.generator.asterisk.js
```
rwols: keyword.generator doesn't seem to be in the style guide, maybe use keyword.operator.generator.js instead? Or punctuation.definition.keyword.generator.js? Or keyword.other.generator.js?

JavaScript/JavaScript.sublime-syntax
Thom1729: The `keyword.control.import-export` scope is inherited from the original syntax. I left it unchanged, but you could certainly make an argument for changing it now that we have finer control over the details.

JavaScript/JavaScript.sublime-syntax
Thom1729: I wouldn't like to say `import.export` because that to me implies that the keyword is of subtype `import` and sub-subtype `export`. The keyword may be `import` or `export`, but it is not both, and `export` is not truly subsidiary to `import`.

JavaScript/JavaScript.sublime-syntax
Thom1729: This is also inherited from the original. If we were to change it, then either `keyword.other` or `punctuation` might be reasonable. You could even make an argument for `storage`.

rwols:
> That file isn't pure JavaScript; it uses Flow annotations to declare types. The default JavaScript syntax does not support Flow.

Thanks for clearing that up, so strictly speaking one wouldn't have to worry about it.

> I do think it's worth discussing fallback for nonstandard syntax. Could there be a graceful way to "skip" Flow and TypeScript declarations in the new architecture? Maybe. Personally, I'm content to punt on nonstandard extensions issue to keep the core syntax clean.

Well, existing users that've been using the default Javascript syntax with flow annotations might be in for a surprise if this gets merged. I do agree with your stance on keeping the core syntax clean. I have no idea what percentage of Sublime users use babel-sublime for JSX and Flow. If it's a high percentage then my concerns are moot.

JavaScript/JavaScript.sublime-syntax
rwols: Now that I've read what `export` actually does in Javascript, it seems strange to me to put these two tokens at the same level. I think `storage.modifier.export` would be better for `export`. But as you said, this is all inherited from the current syntax so it's probably better to leave it as it is right now.

rwols:
I can't find a single thing that's wrong with these changes, so 👍 from me.

jcberquist:
@wbond: would it be possible to review this and give an update on its status? I am have been using it for a while and I am very happy with it. It _is_ a big change in the syntax file, which I know you are leery of given past history, but I don't think using it will cause issues with syntax highlighting changes. Maybe others could test it out and give feedback on this? @rwols brought up a good point regarding the differences when highlighting non vanilla JavaScript files.

At this point it does a really good job of preserving all the current tests - if you run it against the current syntax test file, you will see there are only a handful (10) of failed tests. Six of these are due to the removal of the `meta.function.anonymous` scope (see https://github.com/sublimehq/Packages/issues/1071). The others are due to the fact that the test file is missing a couple of terminating semicolons and so expressions actually continue that the current syntax does not correctly identify: see [L799](https://github.com/sublimehq/Packages/blob/master/JavaScript/syntax_test_js.js#L799) (fails due to the expression on [L795](https://github.com/sublimehq/Packages/blob/master/JavaScript/syntax_test_js.js#L795) continuing and not actually being terminated) and [L969](https://github.com/sublimehq/Packages/blob/master/JavaScript/syntax_test_js.js#L969) (fails for the same reason - [L965](https://github.com/sublimehq/Packages/blob/master/JavaScript/syntax_test_js.js#L965) is missing a terminating semicolon).

There are some differences that are not apparent from the tests. The updated syntax is more accurate with scoping - for instance the current syntax occasionally scopes spaces where it shouldn't (see the spaces after the `as` keyword in `import` and `export` statements; they are getting scoped as `variable.other.readwrite`). There are also several places the `meta.function.declaration` scope is duplicated and doesn't pop off as it should - see for example [L340](https://github.com/sublimehq/Packages/blob/master/JavaScript/syntax_test_js.js#L340). These things could of course be cleaned up. A more significant difference (albeit one that won't affect highlighting) is that in function declarations the `meta.function.declaration` scope is now applied all the way to the `meta.block` scope, whereas in the current syntax it tends to pop off after the closing `)` - a good example of this is on [L127-L137](https://github.com/sublimehq/Packages/blob/master/JavaScript/syntax_test_js.js#L127-L137). I prefer the new scoping, but it is different.

And the updated syntax (to highlight a few things) brings with it much more robust import/export highlighting, as well as the regex highlighting fixes, and it distinguishes array literals from bracket notation. It is also slightly faster in parsing files. 

In general I think that the expression matching setup used here is a great approach to using `sublime-syntax` (and really highlights its superiority to `tmLanguage`).

wbond:
@jcberquist My plan is to review this after 3.0 comes out. I am working towards that end right now, and this is such a big chunk of stuff to review, that it doesn't make sense to me to spend any time on it right now. Sorry to push it off for a bit longer.

Thom1729:
@wbond Any update on this now that 3.0 is out the door? I'm hesitant to add even more changes on top of this already-massive PR (e.g. #1102, #1237, #1238).

wbond:
It is on my list of things to check out. I dealt with some of the older large PRs first.

Most likely some of the YAML will need to be re-written to be less confusing to read. The nested anonymous context pushes are particularly hard to parse when scanning through the file.

rwols:
The multi-push with `else-pop` idiom is genius, I would be sad if it that got rewritten.

Thom1729:
It should be possible to decrease the nesting level by creating new named contexts. Are there any contexts in particular you're concerned about?

wbond:
I'm not concerned about the context transitions, but the syntax. Things like `- -` to create nested lists is hard to read.

Thom1729:
Ah. My usual approach to taming those nested lists is to use macros, but there may be something we can do without them. Some of them can be tamed by replacing anonymous contexts with named contexts, although there are way too many anonymous contexts to do this generally. I'm sure there are other simplifications that can be applied as well, but I doubt that it would be practical to eliminate all of the lists of anonymous contexts.

wbond:
Can you check and see what the filesize of the `{data}/Cache/JavaScript/JavaScript.sublime-syntax.rcache` file is on your machine?

wbond:
There appear to be 29 anonymous nested contexts. I don't think that is too much to maintain. I'd much rather that than layer a macro system over the whole thing.

Thom1729:
31 KB (versus 78 KB for the original). The new version also takes about 58 ms on my machine to parse the complete jQuery source, down from 79 ms.

Myself, I find the anonymous contexts to be easier to read, but then I've been staring at this file for an awfully long time and become accustomed to its quirks. If you think it looks better with named contexts instead, I can certainly do that.

Thom1729:
I've removed all nested lists from the syntax.

rwols:
> The new version also takes about 58 ms on my machine to parse the complete jQuery source, down from 79 ms.

> I've removed all nested lists from the syntax.

What can you report about the performance now?

wbond:
@rwols The change was strictly syntax, so there should be a 0% change in performance.

Thom1729:
@wbond Have you had a chance to take a look at the syntax now that the nested lists are gone? I'm anxious to get to work on #1269, but I don't want to pile yet more changes on top of an unmerged PR.

JavaScript/syntax_test_js.js
```diff
@@ -198,8 +234,12 @@ not_a_comment;
     '// /* not a comment'() {},
 //  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ -comment() {}
 
+    '// /* not a comment'() {},
+//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ -comment() {}
+
```
wbond: How is this different than the assertions right above?

Thom1729: It almost certainly isn't. Looks like it was added four months ago, probably by mistake.

wbond:
Thanks for all of your work on this @Thom1729, and thanks for the reviews @rwols and @jcberquist.

Issue #937, opened 2017-04-27
[LaTeX] Bad document rendering after dollar sign $


evandrocoan:
This document is rendered wrong after the dollar sign `$`, the scope `string.other.math.tex` is pushed into the stack, but should not:
```latex
\documentclass{article}
\usepackage{array}
\newcolumntype{$}{>{\global\let\currentrowstyle\relax}}
\newcolumntype{^}{>{\currentrowstyle}}
\newcommand{\rowstyle}[1]{\gdef\currentrowstyle{#1}%
  #1\ignorespaces
}

\begin{document}
\begin{tabular}{$l^c^r}
\rowstyle{\bfseries}
a & a & a \\
b & b & b \\
c & c & c \\
\end{tabular}
\end{document}
```

![image](https://cloud.githubusercontent.com/assets/5332158/25506494/52485510-2b7d-11e7-9fe0-4e47f5d9f1eb.png)

This code is from: 

1. https://tex.stackexchange.com/questions/4811/make-first-row-of-table-all-bold

FichteFoll:
Works for me on 3131. Do you perhaps have an override in place? (Check with [OverrideAudit](https://packagecontrol.io/packages/OverrideAudit))

![img](https://x0.at/-Ep.png)

evandrocoan:
Thanks for testing. I will check it.

Sorry, I forgot to mention, may be it could be because I am using the build 3114.

Issue #544, opened 2016-07-22
[LaTeX] Wrong highlighting of $ inside \path command


merlinthered:
Using a $ inside a \path command leads to the following text being highlighted as math mode. This is incorrect. \path takes special characters as-is, just like the \url command provided by the hyperref and url packages.

I have created a pull request #543 for a straight-forward fix. I would be interested if anyone is aware of any side effects this might have. Perhaps there are other packages that also define the \path command, which behave differently?

wbond:
#543 has been merged, closing

Issue #472, opened 2016-06-14
[LaTeX] lstinline should behave like verb


mwillsey:
`lstinline` (from the popular `listings` package for code listings) behaves like a verbatim environment that also highlights code. It should be highlighted like the `verb` macro, but more importantly, things inside the `lstinline` macro should not mess up the highlighting outside of it.

`verb{code with $ symbol} regular words` highlights as expected, with ` symbol} regular words` not put in math mode.

`lstinline{code with $ symbol} regular words` does not highlight this way; instead, the `$` puts you in math mode, breaking the highlighting for the rest of the file until another `$` (which may not exist).

I propose that `lstinline` should just behave identically to `verb`. I wrote a [pull request for this behavior in TextMate](https://github.com/textmate/latex.tmbundle/pull/151); it's a very simple change.

wbond:
Check out #370. There is a massive rewrite of LaTeX in progress.

randy3k:
It has been handled in #370.

wbond:
Alright, I'm going to close this then.

PR #1010, opened 2017-05-29
[LaTeX] Pop math environment on unmatched closing brace

r-stein:
This is not directly LaTeX syntax, but a suggestion to improve the handling of not directly supported behaviors by keeping the error local.

There have been a lot of issues of people opening math environments inside commands and not closing them, because it is not required with that specific command (but this specific command is not supported by the syntax definition). https://github.com/sublimehq/Packages/issues/544 https://github.com/sublimehq/Packages/issues/937 https://github.com/sublimehq/Packages/issues/472 https://github.com/SublimeText/LaTeXTools/issues/1124 https://github.com/SublimeText/LaTeXTools/issues/1104 https://github.com/SublimeText/LaTeXTools/issues/921 [tex.stackexchange#371105](https://tex.stackexchange.com/questions/371105/sublime-thinks-my-whole-text-is-an-equation)

The idea of the PR is to pop the math environment if a unmatched closing brace `}` occurs. This will end the math environment and keep the error local instead of messing up the whole document. Since closing a brace inside a math environment would be invalid anyway it should not break on existing documents.

I am open to feedback
@ig0774 @randy3k

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -486,6 +486,8 @@ contexts:
     - include: scope:text.tex#math-characters
     - include: scope:text.tex#math-numerics
     - include: general-constants
+    - match: (?=\})
+      pop: true
 
   inline-math:
     - match: \$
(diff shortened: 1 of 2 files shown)


randy3k:
Excellent.

wbond:
Thanks!

PR #1011, opened 2017-05-29
[Scala] Minor symbolic id bugs

djspiewak:
- Annotations can only contain alphabetical ids
- Tricky over-disambiguation issues on ids starting with `=` and containing a `>`


wbond:
Thanks for these Scala PRs!

PR #1014, opened 2017-05-29
[Scala] Out with old type lambdas, in with new

djspiewak:
This PR probably deserves a bit of discussion before merge.

As @jpz has pointed out, the highlighting of `λ`, `α` and such has been very ad-hoc.  And what's more does not correspond to the current way that the language is used.  Anyone doing type lambdas is using [kind-projector](https://github.com/non/kind-projector) or is still using the old structural types trick, but often without the Greek (e.g. [fs2](https://github.com/functional-streams-for-scala/fs2)).

So this PR does two things: it removes the old ad-hoc highlighting and adds support for highlighting `?` as well as `λ`/`Lambda` within types.


jpz:
I still disagree with what you are doing here, and while I disagreed, I respectfully stood down from further debating it at the time, as it appeared we were a long way apart, and you are a big contributor to this code. 😄 

The syntax of the language, symbols and keywords, is clearly defined in the language specification: http://scala-lang.org/files/archive/spec/2.11/01-lexical-syntax.html

What is not defined then are operators, as operators are actually arbitrary, and not specifically part of the grammar - which leads to making gut calls on what to add to the syntax highlighting.  "+" and "-" for instance are not a part of the grammar, but merely a function implemented on some Scala Standard Library classes.

So this is where the arbitrariness of syntax colouring enters the scene for Scala.

To resolve the question of what other arbitrary names should we colour, one place to start is with things which are commonly used prefix, postfix and infix operators (http://scala-lang.org/files/archive/spec/2.11/06-expressions.html#infix-operations)

Another place to consider is to highlight those operators which are defined in the Scala Standard Library (http://scala-lang.org/files/archive/spec/2.11/12-the-scala-standard-library.html) as functions upon classes such as Int, Double and Boolean.  

So I agree that some interpretation and judgement is needed when implementing the syntax highlighting.

The problem I have here in this implementation, is that "Lambda" and "λ" are getting arbitrarily getting special syntax colouring in Type Projection, even though they are neither keywords nor symbols of the language.  Type Projection is defined here in the language specification as T#x for arbitrary T, for specific values of T where T ∈ {"Lambda", "λ"} (http://scala-lang.org/files/archive/spec/2.11/03-types.html#type-projection)

It is my opinion that this implements syntax colouring according to a highly arbitrary (even if idiomatic amongst many) usage of Scala.  An analogy to this could be for instance someone deciding to add syntax colouring for the Akka DSL.  This would also be arbitrary, non-agnostic, non-language-spec feature to add to the syntax colouring.

While I dissent, I hope you regard it as respectful dissent!

djspiewak:
@jpz Your arguments did have an impact on me.  Which is part of why I've pulled out the arbitrary highlighting.  What is there now is literally just support for kind-projector, which is not a convention but an actual language feature generated by a plugin which is relatively standard.  The two variants of Lambda (as well as `?`) *are* keywords when in type position with that plugin.  So it's nearly the same as #1016  and similar.

jpz:
Hey, thanks for the response, that helps.  I read quickly and didn't catch the full meaning what you were trying to say in the initial PR text.  (I blame my lazy reading).  I see the specific keywords I questioned in fact defined at the top of the GitHub repo for kind-projector (https://github.com/non/kind-projector)

Your minimal support of these extensions seems fine to me, as I feel there is probably almost invisible impact for people that do not use them, and there obviously must be significant utility to yourself and others here.

The one aspect that may help would be to put a block-comment at the top of the syntax file to explain what specific things you've added which support for which are extensions, and your rationale, as this might assist future maintainers.  I see there are some parts of the file already which reference the spec via URL.

👍

FichteFoll:
+1 for adding comments about specific extensions the syntax definition supports.

djspiewak:
Agreed.  The mode needs more documentation in general.  The fact that the Scala spec itself is not really definitive on this stuff doesn't help.

PR #1015, opened 2017-05-29
[Scala] Better support for newline-wrapped imports

djspiewak:
Minor issue:

```scala
import foo._, Foo._  // was highlighted correctly

import foo._,
  Foo._   // was not highlighted correctly (also Github has the same bug, apparently)
```


FichteFoll:
I suppose you're looking for a `punctuation.separator.name-list` or similar for the comma. I don't think we have a standard scope name for this kind of token currently.

djspiewak:
@FichteFoll That seems like a reasonable scope.  I suspect that most schemes will just highlight `punctuation.separator`, or even just `punctuation`.

PR #1020, opened 2017-06-01
[Clojure] clean up syntax def to make it easier to work with in future

keith-hall:
this PR (re)moves duplicated anonymous contexts, adds variables and removes some lookbehinds, all to make the syntax easier to understand. More tests were added to help prove nothing was inadvertently broken.


keith-hall:
superseded by the Clojure rework at https://github.com/sublimehq/Packages/pull/1259

PR #1024, opened 2017-06-06
[Python] Format strings don't extend past strings

### LINKED ISSUE #1022: [Python] braces may disrupt highlight of strings
The following code

a=["aaaa{": "bbbb{"]

is incorrectly highlighted.
Can anyone confirm this?

FichteFoll:
The complex format string look-ahead disregarded string boundaries.

Fixes #1022.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1034,7 +1034,7 @@ contexts:
         2: storage.modifier.conversion.python
         3: constant.other.format-spec.python
         4: punctuation.definition.placeholder.end.python
-    - match: \{(?=[^\}]+\{) # complex (nested) form
+    - match: \{(?=[^\}"']+\{) # complex (nested) form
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -1034,7 +1034,7 @@ contexts:
         2: storage.modifier.conversion.python
         3: constant.other.format-spec.python
         4: punctuation.definition.placeholder.end.python
-    - match: \{(?=[^\}]+\{) # complex (nested) form
+    - match: \{(?=[^\}"']+\{) # complex (nested) form
```
michaelblyons: Does this break `"aaaa{'foo"`? I don't know why someone would do that, but I think we should add tests for it.

FichteFoll: `"aaaa{'foo"` is properly highlighted as a string.

wbond:
Thanks!

Issue #963, opened 2017-05-05
[TCL] foreach for can't highlighted in proc body


ZhiPing-S:
![default](https://cloud.githubusercontent.com/assets/16476209/25737512/d9e2f466-31aa-11e7-888d-336c527fd9e1.png)

keith-hall:
Please include the text used in the screenshot for easier reproduction via copy and paste.

joherv:
To clarify a bit on this. Sytax highlighting does not work for TCL inside while loops, for loops and foreach loops (maybe other bodies as well). I'm using build 3133, but have also tested with 3127 with the same results. Could this be a result of the fix for issue #645 ?

```
proc test {} {

	set saoid [svcmap_saoid sm]

	# Syntax highligting works inside an if statement:
	if {[defvar cpe_uplink_ifoid 0] == 0} {

        array set cpe_param [objectGetField -oid $cpeoid -fieldname parameters]
        if {[info exists cpe_param(model)]} {
            set cpe_model $cpe_param(model)
        } else {
            error "ERROR: Model not set"
        }
    }

    # but not within for example a foreach loop:
    foreach cpe_ifoid [objectTreeFind -oid $cpeoid -walkdown 0 -classmask interface] {
        set cpe_if_role [objectGetField -oid $cpe_ifoid -fieldname role]
        if {[string match "uplink.running" $cpe_if_role]} {
            set cpe_uplink_ifoid $cpe_ifoid
            set cpe_uplink_name [objectGetField -oid $cpe_uplink_ifoid -fieldname name]
            break
        }
    }
}
```

keith-hall:
I don't know TCL at all, but it seems one can fix it by changing [this `pop`](https://github.com/sublimehq/Packages/blob/03dff079bce4eb861d2b3a4d922ffb86bae84400/TCL/Tcl.sublime-syntax#L115) to

```yaml
      set:
        - match: '(?={{end_chars}})'
          pop: true
        - include: commands
```

joherv:
That didn't change anything. Not sure if I edited the correct file though. I run OSX and edited /Applications/Sublime Text.app/Contents/MacOS/Packages/TCL.sublime-package
I restarted ST3 after the change.

This works fine in 3114 where syntax file just has the single pop statement. Must have been something introduced in the fix for #645 that broke this.

keith-hall:
The fix for that issue doesn't seem to be related, you can try manually reverting it and see it makes no difference to the code snippet you posted here.

[...]

PR #1025, opened 2017-06-08
[TCL] fix bug where switch and foreach body contents were unscoped

keith-hall:
fixes https://github.com/sublimehq/Packages/issues/963


wbond:
Thanks!

PR #1033, opened 2017-06-15
Removed whitespace inside character class.

### LINKED ISSUE #1031: Javascript syntax highlighter issue
Javascript syntax highlighter regex issue:

'str'.match(/[" ]+/g);

a space and any type of quote in a collection breaks the highlighting

Operating system and version:
Windows10
Monitor:
Resolution 1080p
Sublime Text:
Build 3135
64bit

Thom1729:
I forgot that the `(?x)` flag doesn't ignore whitespace inside a character class. As a result, regexes containing spaces were not recognized by the syntax. Fixes #1031.


wbond:
Excellent, thanks for the quick fix! We should probably add a test or two to make sure we don't revert in the future.

PR #1035, opened 2017-06-15
[Scala] Fixed incorrect class decl state management

djspiewak:
I legit have no idea how this ever worked.  My only guess is that if you `pop: true` when there is only a single context on the stack, Sublime will treat it as a `set: main`.

This fixes https://github.com/SublimeText-Markdown/MarkdownEditing/issues/430, and probably some other issues that I didn't know about.  Wasn't quite sure how to write a test for it, unfortunately.


wbond:
Thanks!

PR #1038, opened 2017-06-16
[C#] more stray brace handling

keith-hall:
this makes it easier to find syntax errors in C# files by scoping stray closing braces as illegal, just like Python does. Previously the C# syntax did this, but only for curly braces and only at root level - i.e. when not inside a class or method etc.


wbond:
Thanks!

PR #1039, opened 2017-06-17
[Objective-C] Fix accidental *.c++ suffixes

rwols:
Some scopes ended in the c++ suffix instead of the objc++ suffix.


wbond:
Thank you!

PR #1043, opened 2017-06-19
[C#] more stray brace detection

keith-hall:
https://github.com/sublimehq/Packages/pull/1038 missed a case of stray braces in function calls. This corrects that, while limiting the amount of invalid highlighting, to help developers find the bug quickly without any extra distractions.


wbond:
Something with this was funky. I'm going to back it out for now, and you can do a new PR if you want.

Issue #1041, opened 2017-06-17
[C++] Syntax highlighting issue with __declspec(deprecated("msg"))


pgrit:
Having a set of parentheses within another set of parentheses after a known identifier starting with two underscores breaks syntax highlighting.

![screenshot from 2017-06-17 16-35-34](https://user-images.githubusercontent.com/12607894/27253778-023ccc7e-537b-11e7-848d-eb66d9d0037d.png)

```C++
_declspec(deprecated("** this is a deprecated function **")) void func2(int) {}

__declspec(deprecated("** this is a deprecated function **")) void func2(int) {}

__notdeclspec(deprecated("** this is a deprecated function **")) void func2(int) {}
```

PR #1045, opened 2017-06-19
[C++] [Objective-C] Fix declspec highlighting

rwols:
Fix for https://github.com/sublimehq/Packages/issues/1041.

Commit 3eabef5 is separate because Objective-C was missing a bunch of declspec handling. This made me wonder wether this was intentional or not. If it's intentional, then it's easy to revert that commit now.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -358,7 +358,7 @@ contexts:
         - match: '\)'
           scope: meta.group.c++ punctuation.section.group.end.c++
           pop: true
-        - match: '\b(align|allocate|code_seg|property|uuid)\b\s*(\()'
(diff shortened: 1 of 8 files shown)


wbond:
I am sort of out of my depth here, but I left the __declspec() stuff out of Objective-C since I thought it was an MSVC compiler feature, and thus Objective-C code on OS X wouldn't use it.

rwols:
I did some digging and it looks like Microsoft wants to support ObjC in some way too:
1. [clang added support for ObjC declspec](http://llvm.org/viewvc/llvm-project?view=revision&revision=275610)
2. [declspec present in Apple's ObjC header](https://opensource.apple.com/source/objc4/objc4-437.1/runtime/objc-api.h.auto.html) (see "TARGET_OS_WIN32")
3. [ObjC for Windows](https://github.com/Microsoft/WinObjC) (didn't know this existed)

PR #1047, opened 2017-06-20
[C#] add support for binary literals and digit separators

keith-hall:
https://github.com/dotnet/roslyn/pull/2950
http://www.davidhayden.me/blog/csharp-7-underscore-as-digit-separator-in-numeric-literals

EDIT: this PR also scopes the numbers more specifically, just like how Python does it.

PR #1049, opened 2017-06-21
[Python] scope numeric punctuation that marks which base to use

keith-hall:
based on a discussion on Discord, this allows plugins (and color schemes, if so desired) to tell where the number begins and ends, compared to the punctuation that isn't part of the number, but specifies what base the number is in.

(There is also a PR for C# that does the same: https://github.com/sublimehq/Packages/pull/1047)

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -333,28 +333,39 @@ contexts:
   numbers:
     # https://docs.python.org/3/reference/lexical_analysis.html#numeric-literals
     # hexadecimal
-    - match: \b(?i)(?:0x\h*)(L) # py2
+    - match: \b(?i)(0x)\h*(L) # py2
       scope: constant.numeric.integer.long.hexadecimal.python
       captures:
-        1: storage.type.numeric.long.python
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -333,28 +333,39 @@ contexts:
...
+      captures:
+        1: punctuation.definition.numeric.hexadecimal.python
     # octal
-    - match: \b(?i)0(?:o|[0-7])[0-7]*(L) # py2
+    - match: \b(?i)(0o?)(?=o|[0-7])[0-7]*(L) # py2
       scope: constant.numeric.integer.long.octal.python
       captures:
-        1: storage.type.numeric.long.python
-    - match: \b(?i)0[0-7]+ # py2
+        1: constant.numeric.integer.octal.python
```
wbond: Shouldn't this be `punctuation.definition…`?

keith-hall: yes, looks like I messed this one up, nice spot :)

wbond:
Thanks for these enhancements to the various syntaxes!

PR #1056, opened 2017-06-26
[CSS] Use \x{} for all custom_element_chars

robinst:
Some regex engines such as Oniguruma distinguish between `\x..` and
`\x{..}` in that the former matches bytes whereas the latter matches
Unicode code points, see here:

https://github.com/kkos/oniguruma/blob/master/doc/RE#L26

The CSS syntax matches e.g. the U+00B7 character. So for that to work in
Oniguruma, it needs to be written as `\x{B7}`. For readability, this
change uses `\x{00B7}` which is closer to how Unicode characters are
commonly denoted.

(CC @Briles and @keith-hall)


keith-hall:
It's common when making changes to the `.sublime-syntax` files that at least one `syntax_test_` file would also be modified to ensure the changes aren't regressed in future - I just want to make a note that [U+00B7](https://unicode-table.com/en/#00B7) is already included as part of an assertion in the syntax tests here: https://github.com/sublimehq/Packages/blob/303816f09fe5534487c010082bcd96c408823671/CSS/syntax_test_css.css#L882 so I think that in this case, no changes are required to the tests - they pass (in ST 3139) both before and after this change.

wbond:
Thanks!

Issue #626, opened 2016-09-21
shell highlighting broken if no space follows <<<


mvdan:
See https://github.com/rmarquis/pacaur/issues/557#issuecomment-248364847 and https://github.com/rmarquis/pacaur/issues/557#issuecomment-248599050.

oriSomething:
I just get this error no matter what
<img width="482" alt="screen shot 2016-09-22 at 08 16 33" src="https://cloud.githubusercontent.com/assets/2685242/18737664/f25b3ce8-809c-11e6-9f60-3ccdb9016c30.png">

neok-m4700:
Build 3124 breaks the shell syntax (25000 sanity limit) for me as well on Ubuntu 16.04

jfcherng:
Interesting... there is [no real change](https://github.com/sublimehq/Packages/commits/master/ShellScript) between 3124 and 3114 for shellscript.

wbond:
If you are hitting the recursion error it means you have a third party syntax installed that is creating an include loop. This is very likely an HTML syntax, but may be something like JavaScript.

FichteFoll:
Or some markdown flavor.

neok-m4700:
@jfcherng thanks, it's a fix for now ...

wbond:
What is almost certainly happening is that Shell Script allows embedded heredocs of the syntaxes Markdown, HTML, Python, Ruby, AppleScript and Textile. One of those syntaxes on your machine must be from a third-party package, and must have a scope that includes Shell Script directly (for embedding), or includes it via another include.
----

The one possible way to stem the tide of such recursion errors is for Sublime Text to add functionality that allows omitting a context when embedded. Then, we'd add an `embed:` directive like `include:` that would just skip contexts marked with `skip_embedding: true`. So, you could allow Shell Script to embed PHP, but it would include a version of PHP that does not highlight heredocs with embedded languages. This could also potentially make `with_prototype` more useful, since that can easily lead to include recursion.

This would almost certainly work fine in practice, since it would be kind of crazy for someone to embed PHP with a heredoc of a Shell Script inside of a Shell Script heredoc. And if they did, it would seem perfectly fine to not syntax highlight the innermost Shell Script code.

rwols:
[...]

Issue #190, opened 2016-02-19
Shell-Unix-Generic.sublime-syntax is matching fist line directories that start with "sh"


frogbyte:
The path to my python executable has a directory that starts with "sh"
The syntax file for Shell-Unix-Generic is catching it. This could be fixed by checking for a word boundary after `(bash|zsh|sh|tcsh)`.

current:
`first_line_match: '^#!.*\b(bash|zsh|sh|tcsh)|^#\s*-\*-[^*]*mode:\s*shell-script[^*]*-\*-'`

proposed:
`first_line_match: '^#!.*\b(bash|zsh|sh|tcsh)\b|^#\s*-\*-[^*]*mode:\s*shell-script[^*]*-\*-'`

Issue #46, opened 2015-07-02
ShellScript syntax is marking everything as keyword!


aziz:
Now that I'm using the new ShellScript syntax my bash files are full of keywords. 
Not everything is a keyword. I want to differentiate between `shell keyword`, `shell builtin` and `external commands`

jfcherng:
Would you please try [this one](https://github.com/jfcherng/Packages/blob/b2d194c22f6b91105d36e08a6ead1238d456e1c1/ShellScript/Shell-Unix-Generic.sublime-syntax)?  ([diff](https://github.com/sublimehq/Packages/compare/master...jfcherng:b2d194c))

But I have some question on doing that.

For example, what is the category that `pwd` belongs to?
It's now classified as an `builtin` but on my machine it's in `/usr/bin`.
It should be `keyword`, `builtin` or `external commands`?
And how about `cat`?

The same code which looks like the following screenshot now.
![2015-07-02_16-08-16](https://cloud.githubusercontent.com/assets/6594915/8472539/b2108202-20d4-11e5-9560-4eed353231dd.png)

aziz:
> what is the category that pwd belongs to?

I would personally use `type`

``` bash
$ type pwd
pwd is a shell builtin
```
So it's a `shell builtin`

>  To go further, should zip and cat be the same category or not?

Using the same tool to judge, I would say both of them are `external commands`.

jfcherng:
Oh, I get your the point. 
I used `$ which pwd` to do the test and get `/usr/bin/pwd`.
`$ type pwd` gives a `builtin`.

jfcherng:
[Updated](https://github.com/jfcherng/Packages/blob/a72778cd7d0579038441d1a2caef66606ba6fe6e/ShellScript/Shell-Unix-Generic.sublime-syntax). The followings are considered as `builtin`s.

```
bind, builtin, command, declare, enable, export, getopts, 
hash, help, let, local, printf, pwd, read, readarray, readonly, 
return, shopt, test, times, trap, type, ulimit
```

But I just found that `g++` won't be highlighted and I don't know why.

aziz:
:+1: much better now
I still don't like that all those external commands are marked as `keyword`. keywords should be reserved words of the language. I like them to be marked by the syntax but I argue that their scope should not be `keyword`. I think `support` is better

> support — things provided by a framework or library should be below support.

[...]

PR #1057, opened 2017-06-26
[ShellScript] Rewrite

rwols:
Issue links
-----------
Intends to close https://github.com/sublimehq/Packages/issues/434
Fixes https://github.com/sublimehq/Packages/issues/626
Fixes https://github.com/sublimehq/Packages/issues/190
Fixes https://github.com/sublimehq/Packages/issues/46

Goodies
-------

* Proper context sensitive command invocations
* Variable expansion as from ShellScriptImproved
* Backtick handling as from ShellScriptImproved
* No with_prototype contexts
* Generic options handling
* Builtin-commands know their valid options
* Renamed Shell-Unix-Generic to bash, as it's really just bash (and dash)

About 1300 syntax tests, tested out in the wild with various bash scripts.

Deficiencies
------------

The heredocs handling is still not perfect, but it gets the job done if you
don't have a function definition, inline if-block or compound command on
the same line as the start of the heredoc token.

* It's easy to bring back the embedded language highlighting for particular
  heredoc tokens, but this is omitted for now.
* It's easy to embed the regex language for regex matching in if-blocks at
  this point.

Acknowledgements
----------------

* @jfcherng for starting ShellScriptImproved and the huge amount of syntax
  tests,
* @okdana for creating syntax tests,
* @fichtefoll for pointing me to the Ruby syntax for heredoc handling,
* @keith-hall, @wbond for their suggestions about scope names.


ShellScript/ShellScript.py
```diff
@@ -0,0 +1,36 @@
...
+        file_name = view.file_name()
+
+        # Determine the shell to run
+        firstline = view.substr(view.line(0))
+        match = re.match(r"\s*#!(.+)", firstline)
+        if match:
+            # Note: we split on whitespace and turn it
+            # into a list, otherwise things like
+            # "/usr/bin/env bash" won't work
+            shell = str(match.group(1)).split(" ")
```
FichteFoll: You probably want to use `shlex.split` here.

okdana:
Hmm, i decided to try this today, and it seems to have quite a few issues for me that i don't have with ShellScript Improved.

* Function/utility names and parameters (like `cat`, and `cat -n`) are under the `variable` scope — specifically `source.shell.bash` > `meta.function-call.shell` > `variable.function.shell`. I could just change my highlight colours to account for this, but it doesn't seem appropriate.

* If you have a 'bare' variable assignment (`foo=bar`) in column 1, the first letter in the variable name is unhighlighted.

* Tab-indented closing heredoc delimiters (which are supported with `<<-`) are not recognised, causing the heredoc to never be closed.

* Contents of literal heredocs (where the opening delimiter is surrounded by single-quotes) aren't treated as literal — it continues to highlight interpolations, &c.

* Function definitions are not highlighted correctly if the function name has non-POSIX-compatible punctuation characters in it. (Notably, `foo:bar()` is a valid function definition in bash, and i use colons like this as a sort of fake name-spacing.) I believe that bash actually allows all printable characters in function names except for these: ``/ $ ` = | & ; ( ) < >``

* Process substitution (`foo < <( bar )`) isn't highlighted correctly.

* File redirects (the `bar` in `foo > bar`) are highlighted as function arguments, which doesn't seem right.

* There are lots of issues with parameter expansion. Strings aren't highlighted as such (e.g., `bar` in `${foo:=bar}` is highlighted as a variable). The `//` in `${foo//a/b}` isn't highlighted correctly, and the `b` (but not the `a`) is highlighted as a variable. The operators/indexes in `${foo:0:8}` aren't highlighted correctly.

* Unquoted `case` strings aren't highlighted as strings.

* `break` isn't recognised as a shell key-word.

* Pattern meta-characters like `*` aren't recognised as such in many cases, including `[[ "${foo}" == bar*baz ]]`

That's what a brief glance shows anyway. This is with Sublime 3126.

rwols:
Thanks for the feedback!

> Function/utility names and parameters (like cat, and cat -n) are under the variable scope — specifically source.shell.bash > meta.function-call.shell > variable.function.shell. I could just change my highlight colours to account for this, but it doesn't seem appropriate.

This is according to the [scope naming guidelines](https://www.sublimetext.com/docs/3/scope_naming.html): `variable.function` for functions, `support.function` for builtin functions. I'm aware that older color schemes seem to not pick up `variable.function`. Function parameters are given the scope `variable.parameter`. I won't change this unless more people want this different.

> If you have a 'bare' variable assignment (foo=bar) in column 1, the first letter in the variable name is unhighlighted.

Could you go into more detail as to what the problem is here? I cannot reproduce this:
```bash
 foo=bar
#^^^ variable.other.readwrite
#   ^ keyword.operator
#    ^^^ string.unquoted
```

> Tab-indented closing heredoc delimiters (which are supported with <<-) are not recognised, causing the heredoc to never be closed.

Thanks, will write a fix for this.

> Contents of literal heredocs (where the opening delimiter is surrounded by single-quotes) aren't treated as literal — it continues to highlight interpolations, &c.

Thanks, should be easy to fix.

> Function definitions are not highlighted correctly if the function name has non-POSIX-compatible punctuation characters in it.

Thanks, will write a fix for this.

> Process substitution (foo < <( bar )) isn't highlighted correctly.

Thanks, will write a fix for this.

> File redirects (the bar in foo > bar) are highlighted as function arguments, which doesn't seem right.

This is a meta scope so it doesn't affect the highlighting. You're right in that, to be absolutely correct, they shouldn't be part of the function arguments. I want to fix this, but it's not a high priority.

> Unquoted case strings aren't highlighted as strings.

This is a design issue and it's easy to highlight them as `string.unquoted`. I'm not sure if that's helpful.

> bar in ${foo:=bar} is highlighted as a variable

Everything in ${...} receives the scope `variable.other.readwrite`. I could change the syntax so that only `foo` gets the `variable.other.readwrite` scope. Do you suggest this?

```bash
: ${foo:=bar}
# ^ keyword - variable.other
#  ^ punctuation - variable.other
#   ^^^ variable.other
#      ^^ keyword - variable.other
#        ^^^ string.unquoted - variable.other
#           ^ punctuation - variable.other
```

I would like to discuss this more before I change something about it.

> The // in ${foo//a/b} isn't highlighted correctly

schermafbeelding 2017-07-08 om 19 49 39

This may have to do with your color scheme. The second `/` in `//` has a different meaning then the first `/`, so I gave the second `/` in `//` the `punctuation.definition` scope. I could also give the second one the same `keyword.operator` base scope but a different subscope just so that older color schemes pick it up. It's up for debate what should happen here and I'm fine with any scope.

> the b (but not the a) is highlighted as a variable.

It's an error that `a` is not highlighted as `variable.other.readwrite` actually. Will write a fix for this.

> The operators/indexes in ${foo:0:8} aren't highlighted correctly.

Yes they are. They're punctuation scopes.

schermafbeelding 2017-07-08 om 19 48 49

It's again up for debate what should happen here and how the colons should be scoped. It's trivial to give them a different scope, and I'm fine with anything.

> break isn't recognised as a shell key-word.

Yup, forgot that one. Will write something for this.

> Pattern meta-characters like * aren't recognised as such in many cases, including [[ "${foo}" == bar*baz ]]

It's hard to decide exactly when `*` should be highlighted... I'll look at ShellScriptImproved to see how @jfcherng does it. You can see that GitHub gives up and *always* highlights `*`.

okdana:
>This is according to the scope naming guidelines

Oh, i guess i'd just never seen that before. Seems strange to me but if it's the standard i guess i should just update my shit. Sorry about that.

>Could you go into more detail as to what the problem is here? I cannot reproduce this

Weird. This is what i get:

assignment

If it's indented at all, or if it's got a 'storage modifier' like `local` in front of it, it looks fine.

>Do you suggest this?

I *think* that's what i'm suggesting, yeah. In ShellScript Improved `bar` has the following scope:

```
source.shell
string.quoted.double.shell
variable.other.bracket.shell
variable.other.bracket.shell
meta.reset.color
```

Which is not super accurate either (again, should be `string` IMO), but at least it distinguishes between the variable name and the rest of it.

>This may have to do with your color scheme. The second / in // has a different meaning then the first /, so I gave the second / in // the punctuation.definition scope.

Ah, i see that now. I'm not sure what else uses that scope, but that might be sufficient.

>Yes they are. They're punctuation scopes.

Hm, i don't know. The first index operator is under the `variable` scope whilst the second is under the inherited `string` scope. And it seems like the indexes themselves should be highlighted as `constant.numeric`, or at least something to distinguish them from regular strings (ShellScript Improved uses `meta.reset` again here).

>This is a design issue and it's easy to highlight them as string.unquoted. I'm not sure if that's helpful. 

It seems consistent with the way unquoted strings are highlighted in variable assignments and parameter expansion, at least.

>It's hard to decide exactly when * should be highlighted

Re: `case` stuff and pattern meta-characters: ShellScript Improved correctly highlights meta-characters in both `[[ ... ]]` string comparisons and `case` clauses. It does *not* highlight unquoted strings as strings in either case (nor in variable assignments — again it uses `meta.reset`), though it *does* give `case` clauses `meta.scope.case-clause.shell meta.scope.case-pattern.shell`.

Anyway, thanks for your work!

keith-hall:
Most likely the first letter bug is due to https://github.com/SublimeTextIssues/Core/issues/1190 being fixed  in 3127, so this syntax won't work correctly in 3126. This repo is aimed at the latest dev build after all.

rwols:
* The "first letter bug" should be gone for build 3126 too (63632a0)
* Tab-indented closing heredocs with `<<-` should work (4e652b6)
* Literal heredocs don't highlight expansions anymore (4e652b6)
* Allow non-posix function names (ff757b2)
* Process substitution syntax should work (f22fcb7)
* Pattern expansions (things like `*` and `?` and `[[:upper:]]`) are highlighted (7f8e4ec)
* break keyword is highlighted (f2914d1)

I went with `keyword.operator` for the colons in a substring expansion `${foo:0:8}` instead of `punctuation.accessor`.

I changed the variable parameter expansions such that only the actual word (e.g. "foo" in ${foo//bar/baz}) gets the `variable.other.readwrite` scope. Everything inside ${...} gets the scope `meta.group.expansion`.

I didn't go for highlighting the arguments to a command as `string.unquoted` -- too much green (or yellow, whatever your color scheme does), nor do I think is it necessary for case items.

It is perhaps strange then that assignment of a variable does get `string.unquoted`, maybe I should change that. On the other hand, it's very clear now when variables get assigned versus command invocations.

I didn't touch the way redirections also get a `meta.function-call.arguments` scope, because you can give arguments *after* a redirection too (example `ls>foo.txt -l`)

I hope you find the time and willingness to put it under scrutiny once again 👍

michaelblyons:
>I changed the variable parameter expansions such that only the actual word (e.g. "foo" in ${foo//bar/baz}) gets the `variable.other.readwrite` scope. Everything inside ${...} gets the scope `meta.group.expansion`.

I am conflicted about this. On one hand, it is different from the behavior for string interpolation in other syntaxes, which feel very much like the same pattern as `${}` expansion. On the other hand, it could allow you to separately highlight additional variables inside the expansion, like `${PWD/#$HOME/~}`.

michaelblyons:
I'm using this branch now, and I love it. There are a very few oddities I'm running into, though.

1. Should `return` be `support.function`? I'm pretty used to `keyword.control` for it.

2. Similar question for `exit` as `support.function`, although I wouldn't know what it ought to be instead. In Python, it's `variable.function`, which I don't really like either. This one I'm less opinionated about, but I do think it's also a special case.

3. I like the support for kebab-cased functions and arguments, but I think function arguments that have `=` should scope it:
    ```sh
    $ python foo.py --option=value --other-option
                            ^ keyword.operator.assignment
    $ git log --format="%h git has this pattern, too"
    ```

rwols:
About `return` and `exit`: [They're both builtin-functions](https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#Bourne-Shell-Builtins). I haven't put much thought into it beyond that 😄. It's probably better if they're keyword.control. I'll do that tomorrow.

okdana:
Sorry i didn't get back to this very quickly. The last comment reminded me.

After fixing up my colour scheme to account for the new standards i think this looks pretty good.

I do understand the concern with `return`. ShellScript Improved treats it as a control key word, which makes sense because it's usually a key word or language construct in other languages like C and PHP. However, in bash, it is *technically* just a built-in, the same as `printf` or `getopt` or anything else. So even though i'm not used to it, i think giving it the same basic scope as `printf` makes sense. If people want it coloured like a key word they can add a colour thing for `support.function.return`. I don't feel strongly either way tho.

That said, i did notice some potential inconsistencies (?) with built-ins and key words:

* `declare`, `local`, `readonly`, and `typeset` are all built-ins just like `return` and `printf`, but they aren't treated as function calls (no `meta.function-call`). That means that it's not possible to customise arguments to these built-ins the same way you can with the others.

* `alias` and `let` don't get `meta.function-call` either.

* The `foo` in `alias foo=bar` is highlighted as a variable, same as `local foo=bar`. It might be appropriate to scope it as a variable i guess, but maybe `entity.name.alias` would be more appropriate (to match the `entity.name.function` used for function definitions). Seems like it should be distinguishable from a regular variable *somehow*, though.

* `[[` is scoped as an alias for the built-in `test` — that's not quite correct. `[` is an alias for `test`, but `[[` is a full key word. In practice i'm not sure how much it matters, but it is inaccurate, and it seems to come with other inaccuracies like the closing `]]` being scoped as an argument. It *does* recognise the difference between `[ -n foo && -n bar ]` (which is invalid) and `[[ -n foo && -n bar ]]` (which is not), though, so it's probably smarter than the scope name would suggest i guess.

Other feedback:

* Should `case`-clause terminators (`;;`) be scoped as operators? I'm not sure how bash sees these internally, but they seem analogous to `;`, which is scoped as an operator.

* Should the `-` in `2>&-` (close fd `2`) be scoped as a redirection operator? It's a special case so i think so.

* There seem to be some unintended (?) differences between `$(( ... ))` and `(( ... ))`. The latter doesn't have a `meta.group` scope, and it incorrectly highlights the `/` and `~` operators.

* Is it possible to highlight 'bare' variables within arithmetic expressions? For example, the `i` in `(( i++ ))`.

* The number-base 'operator' `#` in `(( 8#123 ))` should be scoped as an operator i think.

* Hexadecimal numbers aren't highlighted correctly in arithmetic expressions: `(( 0x1f ))`

* `*` in arithmetic expressions is scoped as a regex quantifier.

* Speaking of which, `*`/`?`/`[` are scoped as regex meta-characters even when they're actually part of a glob. Not really a big deal, it's the same basic concept, but again the scope name is slightly inaccurate, and it looks like you are in fact able to distinguish globs from regex, so i thought i'd mention it for your consideration.

* Dollar-single-quoted strings are highlighted correctly, but not (the admittedly much less common) dollar-double-quoted strings: `$"foo"`

* Lastly, it's again pretty nit-picky, but i think giving command-line options (`-a` in `foo -a`) the scope name 'parameter' is inaccurate/ambiguous — that term has a specific and entirely different meaning in the context of bash scripting than the way it's used here. The POSIX spec describes arguments like `-a` as 'options' (preferred) or 'flags' (historical), so i would suggest one of those.

Bike-shedding aside, i wanted to say that i'm actually super impressed with this now that i've spent some more time with it. I was trying to find weird things to break its assumptions, like line-continuation in unusual places, and `foo -- -a`, and the aforementioned `[[ -n foo && -n bar ]]`, but it surprised me. It's very complete, good job!

rwols:
Excellent feedback, I'll work on this in the coming days.

rwols:
Quite a bombshell of commits again, I didn't want to bore all the watchers of this repository with my tiny commits.

* "return" is now `keyword.control.return` instead of `support.function.return` (e2ed77a)
I did not scope "exit" as a keyword, to mirror the Python syntax.
* If an equals-sign follows an option/switch immediately, it'll be scoped as an operator (6f62098)
* declare, local, readonly, typeset, export, alias and let all have `meta.function-call` (bc34bbe) (c59b9d8)
* The "foo" in `alias foo=bar` now gets `entity.name.function.alias` (b419c42)
* `[[ ... ]]` already got distinguished from `[ ... ]` syntactically, it's just that their scopes were the same. They now have different scopes (305034b)
* The "close filedescriptor" token should be highlighted as `punctuation.terminator` (305d039)
* Inconsistencies between `$(( ... ))` and `(( ... ))` should be gone (eed2403)
* Fix highlighting of various operator symbols like `/` and `~` in arithmetic expressions (a9f93a0)
* Numbers are treated a lot better now (octal number punctuation, hex prefix punctuation, number base operator is punctuation) (96bbd66) and (95f8b67)
* The `*` in arithmetic expressions should not be scoped as a regex/glob quantifier anymore (ac68cd8)
* Locale strings (`$"foo"`) are highlighted (93e741b)
* For command options/switches, we used to have `variable.parameter`. This is now `variable.parameter.option` in order to better reflect the idea that it's an option to a command (2ac4d43)

> Should case-clause terminators (;;) be scoped as operators? I'm not sure how bash sees these internally, but they seem analogous to ;, which is scoped as an operator.

No, I think `punctuation.separator` is a good scope for them. They separate the case items. This is one of the few cases where I use `punctuation.separator`.

> Is it possible to highlight 'bare' variables within arithmetic expressions? For example, the i in (( i++ )).

Maybe... But I don't think that's a high priority. For instance the C syntax attempts no such feature. I'll think about it.

> [[ is scoped as an alias for the built-in test — that's not quite correct. [ is an alias for test, but [[ is a full key word. In practice i'm not sure how much it matters, but it is inaccurate, and it seems to come with other inaccuracies like the closing ]] being scoped as an argument.

Could you go into more detail here? I thought I nailed that problem of `]]` being seen as an argument.

okdana:
Sorry for the delay again!

>Could you go into more detail here? I thought I nailed that problem of ]] being seen as an argument.

If it was an issue before, it's not now. Either i was confused or you fixed it.

I think it looks great. Not a huge amount of time to play with it now, but i spent 10 minutes or so on it and couldn't find anything wrong, with two exceptions which are purely in the 'would be nice' category:

* Grouping parentheses in arithmetic constructs aren't highlighted
* Bare variables in arithmetic constructs aren't highlighted (as previously mentioned)

I suppose you might also consider scoping integer arguments to `break`, `return`, and `exit` as numbers, if you think that's useful.

Other than that, :+1: :+1: :+1:

okdana:
Oops! I spoke to soon, i did find some other things:

* **Bug:** Positional parameters (`$1`) are highlighted as integer constants unless they're wrapped in braces

* **Bug:** Options to `typeset` are highlighted as variables — compare `declare -f` vs `typeset -f`

* **Enhancement:** Highlight *any* option to 'storage modifiers' as such, even if it's not valid in bash (`local -U foo`). This seems like it would be a low-effort way to make the syntax file work better with zsh, as well as probably future versions of bash

* **Enhancement:** Highlight parameter names provided to `read` as variables (`read -r foo`)

You might be interested to know that the highlighting actually doesn't break too badly in zsh. It doesn't like zsh's fancy parameter expansion (`"${(j. .)${(q-)@}}"`), but it doesn't ruin the document or anything.

okdana:
Hey again! FYI, some weirdness i noticed with `declare` and friends (build 3143):

screen shot 2017-09-13 at 11 23 43

Not sure what the scope should be for array values like the ones shown. I guess it could be `string.unquoted.shell` like it is for scalars? (Quoted values are correctly highlighted as such.)

michaelblyons:
This is a (naive?) enhancement I included in `string-quoted-double-common` to mark [ANSI escape sequences:][1]

```yaml
  string-quoted-ansi-escape:
    - match: \\(e|x1[bB])\[
      push:
        - meta_scope: constant.character.escape.shell
        - match: m
          pop: true
        - match: ;
          scope: punctuation.separator.sequence.shell
        - match: \d{4,}
          scope: invalid.illegal.shell
        - match: \d{1,3}
          scope: constant.numeric.integer.ansi-escape.shell
        - match: (.|\n)+
          scope: invalid.illegal.shell
```

[1]: https://misc.flogisoft.com/bash/tip_colors_and_formatting

rwols:
Thanks @okdana, @michaelblyons, will get to work on this.

rwols:
> Hey again! FYI, some weirdness i noticed with declare and friends (build 3143):

Should be fixed in b40ee55.

> This is a (naive?) enhancement I included in string-quoted-double-common to mark ANSI escape sequences:

I played around with this, but what if we're typing this:

```
echo -e "\x1B[41|"
```
(the `|` is the cursor). The accompanying double-quote gets inserted automatically by sublime when we type the first `"`. Then we type `\x1B` because we want to type an ansi escape sequence. Then we type `[`, then `41`, then `m` to close it. But when we're at `41`, the context thinks the following `"` is invalid so it'll scope it as invalid.illegal. Thoughts?

michaelblyons:
@rwols Good point. Dropping the last `match` will stop what you describe, but I'm not sure if that's what we ought to do. Maybe change the regex to `[^"]+`?

okdana:
>Should be fixed in b40ee55.

Hmm, not quite:

screen shot 2017-09-19 at 15 09 01

(Admittedly these are super contrived examples that you probably wouldn't encounter in the real world, but they're valid and affected by the same class of error)

wbond:
I was just looking over this to merge and I have two things I'd like to see done:

 1. Renamed `bash.sublime-syntax` to `Bash.sublime-syntax` for consistency with other syntax names
 2. Change `Shell-Unix-Generic.sublime-syntax` to just include Bash, as not to break the syntax highlighting of open views when users restart after an upgrade. We can hide Shell-Unix-Generic so that user's don't select it in the future, but I'd prefer not to trigger alert boxes about the syntax not being available.

wbond:
I'm going to merge this in since it fixes regex cache size issues for the revamped Makefile syntax. We can work through remaining bugs in smaller individual issues.

wbond:
I should say, thank you very much for the work on this @rwols, and also obviously @jfcherng and @okdana!

rwols:
@okdana I'm still working on your obscure bug, don't worry!

PR #1060, opened 2017-06-29
[C#] fix accessor brackets incorrectly being seen as a type

### LINKED ISSUE #1059: [C#] Incorrect 'Stray Brace' detection in certain cases
Using the syntax for C# as shipped with Sublime 3139, the following code is not correctly highlighted:

```c#
public class Test
{
    static void Main (string[] args)
    {
        int[] items = new int[4];
        items[0]  = 12;
        items[ 1] = 34;
        items[2 ] = 56;
        items[ 3 ] = 78;
        Console.WriteLine("{0}, {1}, {2}, {3}", 
            items[0], items[ 1], items[2 ], items[ 3 ]);
    }
}
```

In particular, in the assignments of `items[ 1]` and `item[ 3 ]` the closing square brace is scoped  as `invalid.illegal.stray.brace.cs`, while it is correctly highlighted for `items[0]` and `items[2 ]`, presumably due to the presence of leading white space.

This only occurs in the assignment; the same code as used in the `WriteLine` highlights correctly in all cases here.

keith-hall:
fixes #1059

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -901,17 +901,10 @@ contexts:
         3: keyword.operator.assignment.variable.cs
       set: line_of_code_in
     - include: lambdas
-    - match: '\b({{base_type}})\s*(?=[\[\],*?]*\s+\w)'
+    - match: '\b(?:({{base_type}})|({{name}}))(?=[,*?]*(\[\])?\s+{{start_char}})'
       captures:
         1: storage.type.cs
-      set: [variables_declaration, type_no_space]
-    - match: '\b({{base_type}})\s+'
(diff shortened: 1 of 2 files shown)


FichteFoll:
Do brackets not denote an array type? Imo a storage.type.array seems appropriate.

wbond:
In this context, it was array access, not a variable declaration. The variable name was getting `storage.type`.

FichteFoll:
I only looked at the tests, which just have a meta scope tested (https://github.com/sublimehq/Packages/pull/1060/files#diff-10f512797a60d78052ee1f1c69fb409bR1034).

Issue #964, opened 2017-05-05
CSS: variables bug(s)


pdknsk:
```CSS
--x1: url(data:image/png;base64,PNG);
--x2: url(data:image/png;base64,PNG);
background-image: -webkit-image-set(var(--x1) 1x, var(--x2) 2x);

--1x: url(data:image/png;base64,PNG);
--2x: url(data:image/png;base64,PNG);
background-image: -webkit-image-set(var(--1x) 1x, var(--2x) 2x);
```
![1](https://cloud.githubusercontent.com/assets/1097609/25751291/c509e7de-31b4-11e7-9544-cf9ef3acbadd.png)

Chrome considers the second valid, which I think may be wrong. I'm not quite sure how to read the definition [diagram](https://www.w3.org/TR/css-syntax-3/#ident-token-diagram), linked from [here](https://www.w3.org/TR/css-variables/#typedef-custom-property-name). It's a bug, but where is the question.

keith-hall:
That `ident-token` regex railroad diagram shows that the identifier can't start with a number, so it seems that ST's highlighting/scoping is correct there.

pdknsk:
I've been informed by a Googler that Chrome is correct. This is addressed in an [updated version](https://drafts.csswg.org/css-syntax-3/#ident-token-diagram) of the diagram.

PR #1064, opened 2017-07-04
[CSS] improve (variable) identifier matching

keith-hall:
fixes https://github.com/sublimehq/Packages/issues/964


wbond:
Thanks!

Issue #992, opened 2017-05-23
CSS Grid: grid-template-rows highlighting incomplete


kleinfreund:
The property `grid-template-rows` is highlighted incompletely. Specifically, the `rows` portions is not treated as `support.type.property-name.css`.

Oddly enough, the sister property `grid-template-columns` is highlighted correctly.

```
.rule {
    grid-template-rows: none;
    /*            ^^^^ not highlighted */
    grid-template-columns: none;
}
```

pdknsk:
> Oddly enough, the sister property `grid-template-columns` is highlighted correctly.

Almost.
```CSS
grid-row-gap:;
        ^^^^
grid-template-areas:;
             ^^^^^^
grid-template-columns:;
             ^
grid-template-rows:;
             ^^^^^
```

PR #1065, opened 2017-07-04
[CSS] reorder some property matches to fix bugs

keith-hall:
fixes https://github.com/sublimehq/Packages/issues/992 - some property names are shorter versions of other valid property names, and were being matched first - thus preventing the longer names from matching. These have now been reordered to fix this.


wbond:
Thanks for the fix!

PR #1075, opened 2017-07-15
Recognize Cython file exts as Python

jbrockmendel:


FichteFoll:
Strictly speaking, Cython uses a slightly different syntax, but it shouldn't hurt to highlight the files as Python until a proper Cython syntax definition exists (which I currently doubt it will).

wbond:
Alright, let's try it!

sweettyler:
I would prefer having little more (i.e., minimal, not full) support in the default Python syntax definition instead of just adding the file extensions. See in https://github.com/sublimehq/sublime_text/issues/4006 how the keyword "as" is highlighted.

Thom1729:
@sweettyler This is an old PR that's already been merged, and proper Cython support isn't really within its scope. You'll get much more useful feedback if you create a new issue.

PR #1077, opened 2017-07-15
Add build-system for ActionScript

idleberg:
While ActionScript might no longer have much of an significance in 2017, I thought it should at least come with a build-system.


keith-hall:
Please (for future reference) can you make your commit titles begin with the package name in square brackets followed by a short description of the change - it makes it much easier to see what has changed when looking at the commit history - "first commit" doesn't really tell us much, "[ActionScript] Add build system" would be much more helpful :)

wbond:
Thanks for this!

PR #1079, opened 2017-07-20
[C#] individually scoped chars are not included in autocompletion

keith-hall:
...so scope the whole word at a time instead to workaround it

originally reported/discovered at https://forum.sublimetext.com/t/gwenzek-all-autocomplete-broken-in-st3-dev-for-c-mode/29925


wbond:
Thank you!

PR #1083, opened 2017-07-22
[C++] [Objective-C] Fix friend business

### LINKED ISSUE #1082: [C++] The scope of friend function incorrectly
``` cpp
class Example {
public:
  friend void example();
  friend bool operator==(const Example &e1, const Example& e2);
};
```

The scope of `example` should be 
```
source.c++ meta.class.c++ meta.block.c++ meta.method.c++ entity.name.function.c++
```
 but it is actually
```
source.c++ meta.class.c++ meta.block.c++
```
and the scope of `operator==` has the same problem too.

rwols:
fixes #1082

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -1078,14 +1078,26 @@ contexts:
     - match: \bfriend\b
       scope: storage.modifier.c++
       set:
-        - match: (?=\s*(?:{{before_tag}}\b)?\s*{{path_lookahead}}\s*;)
+        # "friend class SomeClass;"
+        # - match: (?=\s*{{before_tag}}\s+{{path_lookahead}}\s*;)
+        #   set:
+        #     # Not a super precise meta scope, but I don't think anyone will be
+        #     # bothered by this. e.g. if you type "friend class Foo;", then this
+        #     # will also get this meta scope.
+        #     - meta_scope: entity.name.struct.forward-decl.c++
+        #     - match: '{{before_tag}}'
+        #       scope: storage.type.c++
+        #     - match: (?=\s*;)
+        #       set:
+        #         - match: ;
+        #           scope: punctuation.terminator.c++
+        #           pop: true
+        #     - include: comments
+        #     - match: '::'
+        #       scope: punctuation.accessor.double-colon.c++
+        # "friend SomeClass;" (without struct/class/union)
+        - match: (?=\s*{{path_lookahead}}\s*;)
           set:
-            # Not a super precise meta scope, but I don't think anyone will be
-            # bothered by this. e.g. if you type "friend class Foo;", then this
-            # will also get this meta scope.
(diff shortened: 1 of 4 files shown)


wbond:
Friend functions shouldn't get `entity.name.function` since it isn't the definition of them, just indicating they are friends and can access non-public data. We don't want every friend definition to show up in Goto Definition.

rwols:
Let's look at what [cppreference](http://en.cppreference.com/w/cpp/language/friend) has to say about friends.
Translated into "sublime-syntax" language, when you encounter `friend` in a class/struct/union, there are four branches that can (must) follow:

1. A declaration of another class/struct/union with the keyword class/struct/union before it. .e.g.
```cpp
struct X {
  friend struct Y;
};
```
2. A declaration of another class/struct/union without the keyword class/struct/union before it. e.g.
```cpp
struct X {
  friend Y;
};
```
3. A declaration of a function. e.g.
```cpp
struct X {
  friend void f();
};
```
4. A declaration and a definition of a function. e.g.
```cpp
struct X {
  friend void f() {
    std::cout << "I am f\n";
  }
};
```
So, I agree with you for (1), (2) and (3). But case (4) basically forces us to scope friend functions as entity.name.function, otherwise (4) will never show up in the indexer.

In fact, I realize now I handled the forward-declaring stuff wrong. cppreference says a class/struct/union is only forward-declared when it has the keyword class/struct/union before it.

For (1):
> Designates the class, struct, or union named by the elaborated-class-name (see elaborated type specifier) as a friend of this class. This means that the friend's member declarations and definitions can access private and protected members of this class and also that the nested classes of the friend can inherit from private and protected members of this class. The friend itself can also inherit from private and protected members of this class. (since C++11) **If the name of the class that is used in the friend declaration is not yet declared, it is forward declared on the spot**.

For (2):
> Designates the type named by the simple-type-specifier or typename-specifier as a friend of this class if that type is a (possibly cv-qualified) class, struct, or union; otherwise the friend declaration is ignored. **This declaration will not forward declare new type**.

wbond:
But does anyone actually use syntax 4? I mean, why not just put a method on the class/struct?

rwols:
Usually case 4 happens for `operator <<`, for example:

```c++
class Foo {
 private:
  int hidden_data;
 public:
  Foo(int data) : hidden_data(data) {}
  friend std::ostream& operator << (std::ostream& os, const Foo& foo) {
    return os << "A Foo object with value: " << foo.hidden_data;
  }
};
int main() {
  Foo foo(42);
  std::cout << foo << '\n';
};
```
You have to "friend" the operator otherwise it cannot access the hidden_data.

wbond:
That's a very good example.

If possible, I'd like to write a lookahead to not highlight friends as `entity.name` if we can definitely see a semicolon at the end of the line. That should remove most false positives that are possible to catch.

rwols:
I'll have to work on this PR a bit more, yeah.

Issue #18, opened 2015-06-19
Missing packages


jrappen:
* Is there a reason, why `Color Scheme - Default`, `Default`, `Language - English`, `Theme - Default`, `Vintage` are missing from this repo?
* Will you accept people adding new language packages to the current ones?
* Can we add new packages that include executables such as linters, which require not being packed into a `*.sublime-package`? If so, do we mark them with a `.no-sublime-package` file as required by PackageControl?

FichteFoll:
`.no-sublime-package` is a Package Control-only thing.
I would advise against bundling secondary executables, especially if they depend on other runtimes like php or node.

I am, however, curious as to what kind of criteria a package needs to fulfill in order to become a default package. Personally, I haven't used a fair share of the default syntaxes that ST has shipped since version 2.

sublimehq:
I've expanded README.md to cover some of these questions now.

PR #1084, opened 2017-07-24
[YAML] Use non-ambiguous syntax for xdigit class

robinst:
While `\p{XDigit}` works, alternate regex engines might interpret it
according to [UTS #18](http://www.unicode.org/reports/tr18/#Compatibility_Properties), which contains more than just the ASCII
characters, or not even implement it at all.

The YAML spec only contains the ASCII characters, see here:

http://yaml.org/spec/1.2/spec.html#id2775468

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -51,11 +51,11 @@ variables:
   c_indicator: '[-?:,\[\]{}#&*!|>''"%@`]'
   c_flow_indicator: '[\[\]{},]'
   ns_word_char: '[0-9A-Za-z\-]'
-  ns_uri_char: '(?x: %\p{XDigit}{2} | [0-9A-Za-z\-#;/?:@&=+$,_.!~*''()\[\]] )'
+  ns_uri_char: '(?x: %[0-9A-Fa-f]{2} | [0-9A-Za-z\-#;/?:@&=+$,_.!~*''()\[\]] )'
 
   # Tag stuff
   c_tag_handle: (?:!(?:{{ns_word_char}}*!)?)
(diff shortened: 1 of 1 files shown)


YAML/YAML.sublime-syntax
```diff
@@ -51,11 +51,11 @@ variables:
   c_indicator: '[-?:,\[\]{}#&*!|>''"%@`]'
   c_flow_indicator: '[\[\]{},]'
   ns_word_char: '[0-9A-Za-z\-]'
-  ns_uri_char: '(?x: %\p{XDigit}{2} | [0-9A-Za-z\-#;/?:@&=+$,_.!~*''()\[\]] )'
+  ns_uri_char: '(?x: %[0-9A-Fa-f]{2} | [0-9A-Za-z\-#;/?:@&=+$,_.!~*''()\[\]] )'
```
rwols: > alternate regex engines

What alternative regex engines?

keith-hall: @rwols there are some non-ST projects that make use of ST's syntax definitions, and they may/will use different regex engines, although they kinda have to emulate Oniguruma.

rwols: Does `\h` have the same ambiguity?

keith-hall: Yes, I believe in .NET world it represents horizontal white space characters, though, as I said, anything using these syntax definitions should expect things to break unless they interpret the expressions the same way Onigurama does.
http://www.regular-expressions.info/refshorthand.html

wbond:
Thanks!

YAML/YAML.sublime-syntax
robinst: This change was motivated by Rust's engine not having `\p{XDigit}`. Java's engine for example does have `\p{XDigit}`, but it includes Unicode full-width and other digits (which does not correspond to YAML spec).

`\h` is also tricky because it is valid syntax in some engines but means horizontal whitespace. But given that it's used in other grammars already, we could have use that too, yeah.

PR #1085, opened 2017-07-24
[Ruby] inline comments in multiline parameter list

smanolloff:
Occasionally, there are method definitions like these:

```ruby
def bar(arg1,         # some comment
        arg2,         # some other comment
        opt1: 'foo',  # some other comment
        opt2: 'bar')
```

This PR prevents sublime from confusing inline comments with method arguments/values.

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -745,12 +745,12 @@ contexts:
 
   method-parameters:
     - meta_content_scope: meta.function.parameters.ruby
-    - include: comments
     - match: '\)'
       scope: meta.function.parameters.ruby punctuation.definition.group.end.ruby
       pop: true
     - match: '{{identifier}}'
       scope: variable.parameter.ruby
+    - include: comments
     - match: ','
       scope: punctuation.separator.ruby
     - match: '\*'
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -748,6 +748,12 @@ contexts:
     - match: '\)'
       scope: meta.function.parameters.ruby punctuation.definition.group.end.ruby
       pop: true
+    - match: '(?=#)'
+      scope: comment.line.number-sign.ruby
+      push: comments
+      with_prototype:
+        - match: '(?=$)'
+          pop: true
     - match: '{{identifier}}'
       scope: variable.parameter.ruby
     - match: ','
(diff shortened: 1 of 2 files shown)


rwols:
I notice there is no prototype context. Maybe put `comments` in a new `prototype` context and remove `-include: comments` everywhere? (but use `- meta_include_prototype: false` in heredocs and strings)

smanolloff:
@rwols your observation is valid, but the changes  you suggest (refactoring) are way beyond the scope of this PR (fix). It would be nice if you could address them in a separate PR ;)

rwols:
I have more than enough to do at the moment, so nope 😉

Ruby/Ruby.sublime-syntax
```diff
@@ -748,6 +748,12 @@ contexts:
     - match: '\)'
       scope: meta.function.parameters.ruby punctuation.definition.group.end.ruby
       pop: true
+    - match: '(?=#)'
+      scope: comment.line.number-sign.ruby
```
keith-hall: why not just `- include: comments` as opposed to having a `scope` that isn't used, a `push` and an unnecessary `with_prototype`?

Ruby/Ruby.sublime-syntax
```diff
@@ -745,6 +745,7 @@ contexts:
 
   method-parameters:
     - meta_content_scope: meta.function.parameters.ruby
+    - include: comments
```
rwols: You can probably get away with putting this at line 796.

smanolloff: nah, its not working as expected. I have updated the test accordingly :)

rwols: Ah, because of the comma on line 754 I think. If you put it above there, it'll work I think.

smanolloff: I am not sure what exactly are you suggesting. It is already above line 754.

rwols: The match rules are tested in the order that they appear. So, try to put match rules that pop at the top of the context, and include stuff at the bottom of the context. It doesn't technically matter at this point, because `#` is clearly a different character than `(`, but it's good to have safety nets just in case. I suggest to include the comments context between lines 753 and 754 because of this reasoning. In any case I think this PR is good to merge at this point.

wbond:
Thanks for the fixes and reviews!

PR #1087, opened 2017-07-25
[SQL] Allow [square brackets] as part of schema/tablename

winthropgutmann:
**Addressing**:
  1. Syntax highlighting did not account for square brackets as part of SQL `create` naming conventions
  2. Syntax highlighting did not account for `hash` and `loop` join hints

In SQL reserve words can be used as part of a schema, table, procedure. or function names but only when wrapped by square brackets. It is also a norm for these names to be wrapped in square brackets even when reserve words are not part of the naming convention.
**Examples**: 
`source` is a SQL reserve word
```
CREATE TABLE [dbo].[source]...
```
Normal table definitions can still contain square brackets
```
CREATE TABLE [dbo].[orders]...
```


wbond:
A few high-level comments:

 1. Your additions to the tests didn't add any assertions, so they don't actually cause anything to be tested
 2. You added `\[\w+\]`, however I presume that `[]` quoting is often used to quote identifiers other than word chars, correct?

What SQL engines use `[` and `]` for quoting? Ideally I'd like to start annotation the syntax and test with SQL dialects. Hopefully in the future we'll be moving to a system where there are different syntaxes for the different dialects. This is because it is currently impossible to write a syntax that supports all of them at the same time. A `\` is an escape in MySQL, but is valid in other RDBMS's, resulting in it being impossible to properly detect strings.

michaelblyons:
`[]`-quoting is a Microsoft SQL Server (TSQL?) thing.

> Ideally I'd like to start annotation the syntax and test with SQL dialects. Hopefully in the future we'll be moving to a system where there are different syntaxes for the different dialects.

I am curious about the plans.  The current syntax is certainly helpful but not exactly thorough, even for older versions of ANSI SQL.

winthropgutmann:
@wbond 
  1. Sorry about the testing aspect. First time doing anything Sublime Text related so I'm a bit ignorant to some of the standardizations. 
  2. As @michaelblyons was eluding to, bracket quoting is a t-sql "thing", [MSDN link](https://msdn.microsoft.com/en-us/library/ms176027.aspx). It's used for object names and specifically for:
    - When reserved words are used for object names or parts of object names.
    - When you are using characters that are not listed as qualified identifiers.
    - Spaces in object names (It's not an efficient coding standard but nonetheless, it is supported.)


Allowed Characters in Object Names |
-- |
tilde (~) |
exclamation point (!) | 
percent (%) | 
caret (^) | 
ampersand (&) | 
left parenthesis (() | 
right parenthesis ()) | 
hyphen (-) |
left brace ({) |
right brace (}) |
apostrophe (') |
period (.) |
backslash (\) |
accent grave (`) |

_It isn't uncommon_ for (Microsoft) SQL shops to use object quoting all the time even when the three cases above do not apply, it's actually a good coding practice.
Additionally, [QUOTENAME()](https://docs.microsoft.com/en-us/sql/t-sql/functions/quotename-transact-sql) is used frequently for quoting a character string to create a valid SQL Server delimited identifier for T-SQL; the output being quoted in `[ ]`

So I suppose the `\[\w+\]` isn't entirely correct. Hooray for code review.

PR #1088, opened 2017-07-27
[Regular Expressions] add GNU word bounaries \< and \>

keith-hall:
add GNU word bounaries `\<` and `\>` as these are used by Boost, which ST uses in the Find panel, which is highlighted using this syntax.

see also http://www.regular-expressions.info/wordboundaries.html#gnu


wbond:
And I thought I had a fairly good handle on regex syntax!

Issue #1078, opened 2017-07-16
[C] [C++] Line continuation inside function parameter list inside macro definition is not highlighted


rwols:
this fails:
```c
#define FOO(x,y) \
int f(int x,     \
/*               ^ punctuation.separator.continuation  */ \
      int y)     \
{                \
    return x + y;\
}
```
(the other backslashes are highlighted correctly)

FichteFoll:
Note that this was closed because of #1089 being merged despite it saying it wasn't entirely fixed. It should probably be reopened.

PR #1089, opened 2017-07-27
[C++] [Objective-C] Allow line-continuations inside function parameter lists

rwols:
This PR does NOT fix https://github.com/sublimehq/Packages/issues/1078, but it is a step towards the fix.


wbond:
Thanks!

PR #1094, opened 2017-07-30
[C++] [Objective-C] Auto-pair "<" and ">" inside meta.preprocessor.include

rwols:
This has annoyed me for several years, but I never got around to fixing it. I also assumed someone else would do this :-). Anyway, this does what it says on the tin: auto-pairing of "<>" inside meta.preprocessor.include scopes.

For instance, if you type this now:
```c
#include <|
```
(the `|` represents the cursor), Then it'll insert the `>` automatically now:
```c
#include <|>
```
When you then type `>`, the cursor will just move over the existing `>`:
```c
#include <>|
```
This functionality is already present in the default keybindings and well-known by everyone for e.g. double-quotes. I just copied that and adjusted it accordingly.

### DIFF
--- a/C++/Default.sublime-keymap
+++ b/C++/Default.sublime-keymap
@@ -1,0 +1,41 @@
+[
+    // Auto-pair less-than and greater-than symbols in include statements
+    { "keys": ["<"], "command": "insert_snippet", "args": {"contents": "<$0>"}, "context":
+        [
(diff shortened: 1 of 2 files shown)


wbond:
Alright, let's try it!

FichteFoll:
Wouldn't  this be interesting for generics too?

PR #1096, opened 2017-08-02
[ShellScript] Fix if/then indentation

Lastique:
This makes `then` keyword to have the same indentation level as the corresponding `if` and also makes the code following `then` be indented one level deeper.

See this topic: https://forum.sublimetext.com/t/incorrect-auto-indentation-for-bash-if-then/

### DIFF
--- a/ShellScript/Indentation.tmPreferences
+++ b/ShellScript/Indentation.tmPreferences
@@ -11,11 +11,16 @@
 		<string>^\s*(\}|(elif|else|fi|esac|done)\b)</string>
 		<key>increaseIndentPattern</key>
 		<string>(?x)
-			^\s*(if|elif|else|case)\b
+			^\s*(elif|else|case)\b
 		|	^.*(\{|\b(do)\b)\s*$
+		|	^.*\bthen\b\s*$
 		</string>
 		<key>indentNextLinePattern</key>
 		<string>^.*[^\\]\\$</string>
+		<key>disableIndentNextLinePattern</key>
+		<string>^\s*then\s*$</string>
+		<key>bracketIndentNextLinePattern</key>
+		<string>^\s*if\b.*(?!\bthen\b)$</string>
 	</dict>
 </dict>
 </plist>


rwols:
I predicted this would break, could you find the time to fix the merge conflicts? I haven't really touched the tmPreferences, only renamed the file, so the resolve should be straightforward.

Lastique:
Done. The new content of `Indentation.tmPreferences` is not the same as it was originally, when the patch was created. I don't really understand the contents of this file, and I simply replaced the new content with the proposed one - I don't know if anything essential is removed with this.

wbond:
It would be preferable if we preserved the recently introduced multi-line formatting so that the regex is easier to read and maintain in the future.

Lastique:
Is there any documentation for this syntax?

keith-hall:
http://docs.sublimetext.info/en/latest/reference/metadata.html#indentation-options-children-of-settings

Lastique:
According to that document, the text in the `<string>` tag should be a regex. However, if I break that text into multiple lines the XML parser will introduce spaces within the regex that were not intended and could affect parsing. So e.g. `^\s*(elif|else|case)\b|^.*(\{|\b(do)\b)$|^.*\bthen\b\s*$` would become `^\s*(elif|else|case)\b | ^.*(\{|\b(do)\b)$ | ^.*\bthen\b\s*$`. This is why I'm asking about syntax - it seems that the multiline variant is not equivalent to the original. Am I missing something?

keith-hall:
That's correct, when splitting the regex over multiple lines, one generally has to enable free-spacing aka extended mode by using `(?x)` immediately after `<string>` - there's plenty of examples of this in this repo e.g. https://github.com/sublimehq/Packages/blob/54340707ecbf7de99ab5b02b9414258bd27ed403/PHP/Indentation%20Rules.tmPreferences#L39

Lastique:
I see. I didn't know about the free-spacing mode, thanks for clarifying. I've updated the PR.

wbond:
Thanks for this @Lastique!

PR #1105, opened 2017-08-09
[Erlang] Fix issue 1104, and cosmetics, and start on syntax file

### LINKED ISSUE #1104: [Erlang] Syntax highlight of Erlang dose not work when come across “~”
Erlang source code file: https://github.com/erlang/otp/blob/master/lib/stdlib/src/proc_lib.erl
OS: Linux Mint 18.2
Sublime Text version 3126
The syntax highlight of erlang does not work when come across "~", please refer the following screenshot:

image

rwols:
Fixes #1104 (commit 62f0be2).

Also:
- Use punctuation.section for parens, brackets and braces,
- Use variable.function for function calls instead of entity.name.function,
- Remove redundant single quotes around various regexes,
- Use "scope: ..." instead of a single "captures: 1: ...",
- Add syntax test file.

### DIFF
--- a/Erlang/Erlang.sublime-syntax
+++ b/Erlang/Erlang.sublime-syntax
@@ -716,8 +716,6 @@ contexts:
           captures:
             1: punctuation.definition.placeholder.erlang
             2: punctuation.separator.placeholder-parts.erlang
-        - match: ~.?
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #1106, opened 2017-08-09
[Go] Remove snake_case from placeholder text in Go func snippets

y0ssar1an:
[Go uses camelCase, not snake_case](https://golang.org/doc/effective_go.html#mixed-caps). In the `func` and `tfunc` snippets, the placeholder text for the function name was `func_name`. I replaced this with just `name`. The fact that it's a function name is implied.


wbond:
Thanks!

PR #1108, opened 2017-08-11
[C++] [Objective-C] Add standard library include completions

rwols:
What this does: it presents the standard libraries in the completion widget when you're inside of an `#include <...>`.

I had to adjust the way the ObjC syntax scopes "#import ..." for this. It used to assign a meta scope of `meta.preprocessor.include`, but it now assigns a meta scope of `meta.preprocessor.import`. The "#include ..." statements still receive `meta.preprocessor.include` for their meta scope. This is done so that the completions only trigger for an #include and not for an #import.

This works great in conjunction with https://github.com/sublimehq/Packages/pull/1094.

### DIFF
--- a/C++/C Standard Includes.sublime-completions
+++ b/C++/C Standard Includes.sublime-completions
@@ -1,0 +1,38 @@
+{
+    "scope": "(source.c | source.objc) & (meta.preprocessor.include string.quoted.other)",
+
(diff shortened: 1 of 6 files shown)


wbond:
I definitely think this is work trying out, thanks!

PR #1109, opened 2017-08-11
[YAML] Fix timestamp highlighting with space separated timezone

deathaxe:
The example.yaml of the PyYAML package and the example on http://yaml.org/type/timestamp.html both show a timestamp with space separated timezone:

  2001-12-14 21:59:43.10 -5

It is not matched by the regexp displayed on the same page which is used by YAML.sublime-syntax to do highlighting. But as a canonical timestamp like

  2001-12-15T02:59:43.1 Z

is considered valid, there is no reason to not match the first one, too.


FichteFoll:
It appears the regexp listed on the http://yaml.org/type/timestamp.html page does only account for spaces before a trailing `Z` but not before a timestamp offset. I mainly inherited the regexps from the type pages, which is why this was carried over.

I'll raise this issue in #yaml on freenode and report back on the consensus, because currently the example and the regexp are conflicting.

deathaxe:
The regex does, but the examples provided with the package and displayed on the web site do not. So one could argue about the examples or the regex to be considered correct.

Not sure what the standard says about it. I just found ST to highlight the first example as invalid in the example.yaml provided with PyYAML package and considered it a bug.

I am honestly confused about what's correct. Maybe someone has more information about it?

FichteFoll:
I'm debating this currently with the current yaml spec maintainers, more or less. We haven't reached a consensus yet, however. Personally, I'm in favor of adjusting the reference regexp and thus also adjusting it in the syntax definition because it seems to match the intent better.

keith-hall:
Did you get anywhere with the yaml spec team @FichteFoll?

FichteFoll:
Not really. The hunch was basically "sure, we can talk about it. We'll open-source the website/spec/idk soon and you can file a pull request then, which we may discuss", but other than that I didn't get a lot smarter.

I still agree that we should support what the implementations do in this situation, so I agree with merging this.

wbond:
Thanks for sorting this out!

PR #1115, opened 2017-08-23
[Rust] Fix handling of `#[attribute = "value"]` attributes

robinst:
The string was not recognized before, so with something like:

    #[test = "foo ] bar"]

The first `]` would close the annotation, and the code/lines following
the second " would be recognized as a string.

Screenshot of the problem from Sublime Text 3:

screen shot 2017-08-23 at 16 42 14

Here's the documentation about attributes in Rust: https://doc.rust-lang.org/reference/attributes.html

CC @keith-hall and @trishume

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -233,14 +233,14 @@ contexts:
 
   attributes:
     - match: '(#!?)(\[)'
-      scope: meta.annotation.rust
       captures:
         1: punctuation.definition.annotation.rust
         2: punctuation.section.group.begin.rust
       push:
+        - meta_scope: meta.annotation.rust
         - match: '({{identifier}})(\()'
           captures:
-            1: meta.annotation.rust variable.annotation.rust
+            1: variable.annotation.rust
             2: meta.annotation.parameters.rust meta.group.rust punctuation.section.group.begin.rust
           push:
             - meta_content_scope: meta.annotation.parameters.rust meta.group.rust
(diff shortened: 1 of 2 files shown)


Rust/Rust.sublime-syntax
```diff
@@ -238,9 +238,10 @@ contexts:
         1: punctuation.definition.annotation.rust
         2: punctuation.section.group.begin.rust
       push:
+        - meta_scope: meta.annotation.rust
```
robinst: Not quite sure about this, but without it the string doesn't get the `meta.annotation.rust` scope.

rwols: Doesn't this give a double meta.annotation scope? See line 236 above here. If that's the case, maybe use meta_content_scope?

FichteFoll: Remove line 236, I say.

robinst: Indeed, thanks! Done.

(I'm not quite clear why `punctuation.definition.annotation.rust` from the captures still also gets the `meta.annotation.rust` scope. Do the scopes in `captures` automatically get the `meta_scope` as well?)

rwols: Yes, meta_scope gets applied to all of the matched text that causes the push, and it also gets applied to all of the matched text that causes the pop. In contrast, meta_content_scope doesn't have this property.

wbond:
Thanks!

PR #1117, opened 2017-08-23
Revert broken change in StartDocBlock snippet.

taylorotwell:
Recently, a change was made to not trigger snippets in comments; however, this breaks the StartDocBlock snippet because it by-definition exists within a comment. Reverted that change on this snippet.

### DIFF
--- a/PHP/Snippets/Start-Docblock.sublime-snippet
+++ b/PHP/Snippets/Start-Docblock.sublime-snippet
@@ -3,6 +3,6 @@
  * $0
  */]]></content>
 	<tabTrigger>/**</tabTrigger>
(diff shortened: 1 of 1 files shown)


rwols:
Would it be logical to *only* trigger this snippet in a comment? If so, consider setting the scope to `source.php comment` to prevent it from being triggered outside comments.

taylorotwell:
I guess? The snippet is by definition a comment so I don't see how it could ever be triggered outside of one.

taylorotwell:
Done. Tested that it works.

wbond:
Thanks @taylorotwell and @rwols!

PR #1119, opened 2017-08-24
[ShellScript] Add scopes to #!/usr/bin/env snippet

### LINKED ISSUE #1118: [ShellScript]#!/usr/bin/env snippet is missing a scope
I don't know if this is intentional or not, but [the `#!/usr/bin/env` snippet](https://github.com/sublimehq/Packages/blob/master/ShellScript/Snippets/%23!-usr-bin-env-(!env).sublime-snippet#L5) is missing a scope.
```xml
<snippet>
	<content><![CDATA[#!/usr/bin/env ${1:${TM_SCOPE/(?:source|.*)\.(?:(shell)|(\w+)).*/(?1:bash:$2)/}}
]]></content>
	<tabTrigger>!env</tabTrigger>
	<scope></scope>
	<description>#!/usr/bin/env</description>
</snippet>
```

unexpected_snippet
I noticed this snippet in the command palette while I was editing a plain text file. I'll submit a PR to scope it to `source.shell`. You can close it if the missing scope is intentional. If it is intentional, adding a comment probably wouldn't hurt.

y0ssar1an:
Fix #1118. The `#!/usr/bin/env` snippet didn't have a scope, so it would always be offered, even if the current syntax wasn't `Shell Script (Bash)`.

### DIFF
--- a/ShellScript/Snippets/#!-usr-bin-env-(!env).sublime-snippet
+++ b/ShellScript/Snippets/#!-usr-bin-env-(!env).sublime-snippet
@@ -2,6 +2,6 @@
 	<content><![CDATA[#!/usr/bin/env ${1:${TM_SCOPE/(?:source|.*)\.(?:(shell)|(\w+)).*/(?1:bash:$2)/}}
 ]]></content>
 	<tabTrigger>!env</tabTrigger>
-	<scope></scope>
+	<scope>source.actionscript.2, source.applescript, source.clojure, source.cs, source.d, source.dart, source.fsharp, source.groovy, source.haskell, source.java, source.js, source.julia, source.lisp, source.lua, source.makefile, source.ocaml, source.perl, source.php, source.python, source.r, source.ruby, source.rust, source.scala, source.shell, source.swift, source.tcl, source.ts</scope>
 	<description>#!/usr/bin/env</description>
 </snippet>


FichteFoll:
In contrast, shebang lines aren't shell-specific and used for almost any interpreted language  (rust, node, python, Lua etc).

It makes sense to have a snippet like this active in all languages, imo.

FichteFoll:
The conditional could be expanded, however.

keith-hall:
related to this snippet is https://github.com/SublimeTextIssues/Core/issues/1852

y0ssar1an:
Ahh, I get it now. In that case, I think we should explicitly list all the interpreted languages.

Some cases where you don't want to see the shebang:
1. JSON
1. YAML
1. SQL
1. CSS
1. Go
1. TOML
1. C
1. most compiled languages
1. basically anything declarative

y0ssar1an:
I added this long list of scopes to the shebang snippet. That should cover most popular languages. Did I miss anything?
```
source.actionscript.2
source.applescript
source.clojure
source.cs
source.d
source.dart
source.fsharp
source.groovy
source.haskell
source.java
source.js
source.julia
source.lisp
source.lua
source.makefile
source.ocaml
source.perl
source.php
source.python
source.r
source.ruby
source.rust
source.scala
source.shell
source.swift
source.tcl
source.ts
```

rwols:
Some javascript interpreters also accept a shebang line if it's the first line of the file, not sure if it's worthwhile.

FichteFoll:
`source.js` is already included. 

Now, if only $TM_SCOPE would actually work :wink:

wbond:
TextMate has scopes for things like beginning of file, and uses those to limit this to the first char in the file. Obviously we don't have that scope, so this is better than nothing.

wbond:
Thanks!

wbond:
Just for reference, `TM_SCOPE` was added in 3154

PR #1122, opened 2017-08-24
Add .clisp file extension for 'lisp' syntax

Azeirah:
`.clisp` is a common file extension for common-lisp files.


wbond:
Thanks!

Issue #915, opened 2017-04-20
[HTML] Indent Issue


jgbishop:
I'm seeing an issue in Build 3129 that I wasn't seeing prior to updating from beta build 3126. I have attached a small GIF showing the issue (which I will then describe):

![st3_bug](https://cloud.githubusercontent.com/assets/4210087/25210793/9751d232-254f-11e7-96ea-445862f954d6.gif)

Here's the order of operations:

1. I insert a new list item by typing li then pressing Tab to expand the tag.
2. After typing some text, I press the Enter key to continue on the next line.
3. As soon as I type a single character, the second line is magically indented, presumably to match the indention level from the previous line.
4. Proceeding to the next line does not un-indent things so that all my li elements are aligned. In fact, I can keep increasing the indent as shown in the screenshot.

Note how the indent / unindent jumps around on the second line of item 2. I type the letter a, followed by a backspace, then b followed by another backspace, etc. I'm not sure if there's a plugin at work here or not, but I know this behavior wasn't present in the beta build.

Below are my relevant indent preferences, in case it helps. I'm seeing this in Windows, HTML syntax, tab size of 4. I also have Emmet installed, which may have a bearing on this.

* detect_indentation: true
* auto_indent: true
* smart_indent: true
* indent_to_bracket: false
* trim_automatic_white_space: true
* word_wrap: "auto"
* wrap_width: 0
* indent_subsequent_lines: true

OdatNurd:
I've noticed an indent issue with HTML as well, which I assume is related. As noted here, when closing a tag, the indent jumps backwards an indentation level even if the cursor is already at the appropriate position. 

![sublime_text_3128_indent](https://cloud.githubusercontent.com/assets/13212876/25211605/c92911a4-253a-11e7-9186-377e104b9d44.gif)

As a contrast, in build 3126 closing a tag does not adjust the indent level in this case. 

![sublime_text_3126_indent](https://cloud.githubusercontent.com/assets/13212876/25211745/9d9ff2f4-253b-11e7-8eda-14175f10eb11.gif)

The settings in use here are the same as the OP, with theexception of `wrap_width` being set to `110`.

keith-hall:
It looks like the problem here is that the rules that increase indentation and the rules that decrease indentation don't currently sync up nicely:
- Indentation is currently always increased when an open tag starts the line (ignoring whitespace) and isn't followed by a corresponding closing tag on the same line. It doesn't currently cope with an open tag that follows a close tag.
[...]

PR #1124, opened 2017-08-27
[Ruby] unindent on closing curly brace or square bracket at bol

### LINKED ISSUE #1123: ruby indenting {} within block not 'closing' the indent
version 'Stable Channel, Builld 3126'
Mac OS 10.12.6

within a block (do end, or def xxx end) code within {} is creating an indent, and not 'closing' it.

after edit/line/reindent, the following code within file.rb is wrongly indented

```
require 'rails_helper'

RSpec.describe StaticController, type: :controller do

	expect{
		post :create, xhr: true
		}.to change(Job, :count).by(1)

	end

	def thing

		expect{
			post :create, xhr: true
			}.to change(Job, :count).by(1)

		end
```

where it should be:

```
require 'rails_helper'

RSpec.describe StaticController, type: :controller do

	expect{
		post :create, xhr: true
	}.to change(Job, :count).by(1)

end

def thing

	expect{
		post :create, xhr: true
		}.to change(Job, :count).by(1)

end
```

removing the {} section fixes the indentation problem

keith-hall:
fixes #1123

the biggest problem with indentation and unindentation rules are that they need to mirror/complement each other, otherwise you end up in exactly this situation where for e.g. `{` indents the line, but the corresponding `}` doesn't unindent it, or more rarely that a line gets unindented when nothing triggered an indentation previously. (Tweaks were also made to HTML to this effect a while ago: https://github.com/sublimehq/Packages/issues/915)

So for Ruby, a `{` or `[` anywhere on the line not followed by the respective closing token would cause indentation to be increased, but it would only then be decreased again for `}` or `]` if it was the only non-whitespace token on the line - this PR changes that to decrease the indentation when it is the first non-whitespace token on the line, regardless of what comes afterwards, so that the rules align with each other better.

### DIFF
--- a/Ruby/Miscellaneous.tmPreferences
+++ b/Ruby/Miscellaneous.tmPreferences
@@ -8,7 +8,7 @@
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string>^\s*([}\]]\s*$|(end|rescue|ensure|else|elsif|when)\b)</string>
+		<string>^\s*([}\]]|(end|rescue|ensure|else|elsif|when)\b)</string>
 		<key>increaseIndentPattern</key>
 		<string><![CDATA[(?x)^
 		(\s*


wbond:
Thank you, indentation master!

ConfusedVorlon:
excellent - thank you

PR #1127, opened 2017-08-28
[Markdown] following headers appear incorrect in goto

braver:
This follows from a report on the forum. If two headers appear directly after each other, the first will eat the next and it will appear incorrectly in the symbols list:

https://forum.sublimetext.com/t/a-bug-of-markdowns-ctrl-r-panel/30872

af57522df708ca08ded63dfa55b2ee2f7b584120

This solves that by capturing the newline character and removing it from the symbol.

Questions:

- Originally the match was for `\n?`but I can't seem to get Sublime to capture the `\n` if I also use the `?`quantifier. Perhaps I'm missing something?
- Is `meta.whitespace.newline.markdown` a good scope? I couldn't find any examples.

Possible future improvements:
- Change it to lookahead for the newline and not capture it. I think it's a bit odd to capture it, but ¯\_(ツ)_/¯ . It would simplify things, but perhaps it breaks something somewhere. (something always breaks).

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -412,7 +412,7 @@ contexts:
     - match: '{{escape}}'
       scope: constant.character.escape.markdown
   atx-heading-terminator:
-    - match: '[ ]*(#*)[ ]*(?=$)($\n?)' # \n is optional so ## is matched as end punctuation in new document (at eof)
+    - match: '[ ]*(#*)[ ]*($\n?)' # \n is optional so ## is matched as end punctuation in new document (at eof)
       captures:
         1: punctuation.definition.heading.end.markdown
         2: meta.whitespace.newline.markdown
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -418,9 +418,10 @@ contexts:
       push:
         - meta_scope: markup.heading.1.markdown
         - meta_content_scope: entity.name.section.markdown
-        - match: \s*(#*)$\n?
+        - match: \s*(#*)$(\n)
           captures:
             1: punctuation.definition.heading.end.markdown
+            2: meta.whitespace.newline.markdown
           pop: true
         - include: inline-bold-italic
     - match: '(##)(?!#)\s*(?=\S)'
@@ -429,9 +430,10 @@ contexts:
       push:
         - meta_scope: markup.heading.2.markdown
         - meta_content_scope: entity.name.section.markdown
-        - match: \s*(#*)$\n?
+        - match: \s*(#*)$(\n)
           captures:
             1: punctuation.definition.heading.end.markdown
+            2: meta.whitespace.newline.markdown
           pop: true
         - include: inline-bold-italic
     - match: '(#{3,6})(?!#)\s*(?=\S)'
@@ -440,9 +442,10 @@ contexts:
       push:
         - meta_scope: markup.heading.markdown
         - meta_content_scope: entity.name.section.markdown
-        - match: \s*(#*)$\n?
+        - match: \s*(#*)$(\n)
           captures:
             1: punctuation.definition.heading.end.markdown
+            2: meta.whitespace.newline.markdown
           pop: true
         - include: inline-bold-italic
   image-inline:
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -411,6 +411,12 @@ contexts:
   escape:
     - match: '{{escape}}'
       scope: constant.character.escape.markdown
+  atx-heading-terminator:
+    - match: '[ ]*(#*)[ ]*(?=$)($\n?)'
+      captures:
+        1: punctuation.definition.heading.end.markdown
+        2: meta.whitespace.newline.markdown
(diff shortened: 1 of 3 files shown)


kleinfreund:
I’m not sure, but don’t things like headings require an additional newline before it?

Markdown/Markdown.sublime-syntax
```diff
@@ -418,9 +418,10 @@ contexts:
       push:
         - meta_scope: markup.heading.1.markdown
         - meta_content_scope: entity.name.section.markdown
-        - match: \s*(#*)$\n?
+        - match: \s*(#*)$(\n)
```
keith-hall: hmm, looks like this is failing [example 43 (trailing whitespace)](http://spec.commonmark.org/0.27/#example-43) atm, any chance you could fix for it and a test assertion while you are here please? :)

braver:
> I’m not sure, but don’t things like headings require an additional newline before it?

@kleinfreund I guess it depends on what spec/dialect you're talking about. John Gruber doesn't say anything about it, neither does the atx spec he borrowed it from. GFM seems to take it just fine without the newline.

Markdown/Markdown.sublime-syntax
braver: Looking into it!

keith-hall:
@kleinfreund it is quite common for block level items in Markdown to require a blank line to preceed them, but as @braver says, [ATX headings](http://spec.commonmark.org/0.27/#atx-heading) don't seem to have this rule - [proof](http://spec.commonmark.org/dingus/?text=%23%20Test%0A%23%23%20Try%20CommonMark%0A%0AYou%20can%20try%20CommonMark%20here.%20This%20dingus%20is%20powered%20by%0Ahttps%3A%2F%2Fgithub.com%2Fjgm%2Fcommonmark.js%2C%20the%0AJavaScript%20reference%20implementation.%0A%23%20test%0A%0A1.%20item%20one%0A2.%20item%20two%0A%20-%20sublist%0A%20-%20sublist%0A%0A)

Markdown/Markdown.sublime-syntax
keith-hall: Regarding the `\n` not being optional, I'm not sure why it wouldn't work, but have you tried having the `?` outside the capture group like `(\n)?`? I guess it is not that important if it is no longer optional as a header is extremely unlikely to be on the last line of a document...

I'm half thinking of suggesting moving the whole

```yaml
       - match: \s*(#*)$(\n)?
          captures:
            1: punctuation.definition.heading.end.markdown
            2: meta.whitespace.newline.markdown
          pop: true
        - include: inline-bold-italic
```

block to a separate context (and including it from the current locations), to avoid repetition, but I'm not sure if it's worth it...

braver: > have you tried having the ? outside the capture group like (\n)??

jup

> I'm half thinking of suggesting moving the whole

probably not a bad idea... it's already not-so-easy-as-I-thought, so I don't mind putting in some more effort. 

Am I understanding correctly that before this PR it's also failing example 43?

braver: >  a header is extremely unlikely to be on the last line of a document...

Well, it kinda matters if you're writing a new document and the first line is a h1. And that's actually very likely.

There are less elegant ways to make the \n optional, so I think I got something. Commit coming up.

keith-hall: Yes it was/is failing example 43 before this PR, sorry for not being clearer

braver: @keith-hall I'm having more luck with `[ ]` as apposed to `\s`, with regards to making the `\n` optional. The spec also allows only spaces, not tabs or any other whitespace. Not sure how you feel about such a change?

Markdown/Markdown.sublime-syntax
```diff
@@ -411,42 +411,36 @@ contexts:
   escape:
     - match: '{{escape}}'
       scope: constant.character.escape.markdown
+  atx-heading-terminator:
+    - match: '[ ]*(#*)[ ]*(?=$)($\n?)' # \n is optional so ## is matched as end punctuation in new document (at eof)
```
FichteFoll: The `(?=$)` lookahead is redundant here.

braver: Good catch, thanks. Leftover from earlier iterations.

Markdown/Symbol List - Heading.tmPreferences
```diff
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Symbol List: Heading</string>
 	<key>scope</key>
-	<string>text.html.markdown markup.heading</string>
+	<string>text.html.markdown markup.heading - (meta.whitespace.newline.markdown)</string>
```
rwols: Nice idea with the meta.whitespace scope. Why the parentheses around meta.whitespace.newline.markdown?

braver: It's @keith-hall's idea really. I only used excludes like this a few times before, never would've thought to use it like this. And apparently it's fine without the parentheses.

wbond:
Thanks for the work on this everyone!

PR #1129, opened 2017-08-30
[Python] Recognize escape sequences in docstrings

### LINKED ISSUE #1128: [Python] Bug in string highlighter and error reporting for triple-quoted strings
If you type the following docstring into a Python file in the latest Sublime, it will incorrectly mark the final `"` as the start of another string, with an error complaining about there being no terminating end-quote.

```python
"""Testing: \"Nothing to Be Afraid Of?\""""
```
What it should be doing is treating the second `\"` as part of the docstring contents, rather than part of the terminating `"""`. This also happens with single-quote docstrings:
```python
'''Testing: \'Nothing to Be Afraid Of?\''''
```

FichteFoll:
Fixes #1128.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -968,7 +968,18 @@ contexts:
   docstrings:
     - match: ^\s*(?=(?i)(ur|ru|u|r)?("""|'''))
       push:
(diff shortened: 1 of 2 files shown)


wbond:
Thank you!

PR #1132, opened 2017-09-05
[C#] Fix character escape sequences and add tests

robinst:
* `\u` needs to be followed by 4 hex digits, `\U` by 8
* `\x` needs at least one hex digit
* Add missing escapes `\a` and `\v`

See spec: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#character-literals

(CC @trishume and @keith-hall)


wbond:
Thanks!

PR #104, opened 2015-10-19
[HTML] Remove obsolete or deprecated elements and attributes

philipbordallo:
Bringing completions up to HTML5 standards by removing obsolete and/or deprecated elements and attributes (#102). 

Elements are not being removed from `HTML.sublime-syntax` so backwards compatibility with older version of HTML will still work but older elements and attributes will not autocomplete.

### Elements Removed

[`<applet>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/applet) element is obsolete

[`<basefont>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/basefont) element is obsolete

[`<center>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center) element is deprecated

[`<dir>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dir) element is obsolete

[`<font>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font) element is obsolete

[`<frame>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/frame) element is deprecated

[`<frameset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/frameset) element is deprecated

[`<isindex>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/isindex) element is deprecated

[`<keygen>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/keygen) element is deprecated

[`<menu>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu) element is [deprecated](http://www.w3.org/TR/html4/struct/lists.html#h-10.4) for HTML4 and is experimental its newest reincarnation with [limited support](http://caniuse.com/#feat=menu) in HTML5

[`<tt>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt) element is obsolete

#### [`<a>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a)

[`charset`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-charset) attribute is obsolete

[`coords`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-coords) attribute is obsolete

[`name`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-name) attribute is obsolete

[...]

PR #1135, opened 2017-09-12
[Erlang] Remove query as a keyword

bitnitdit:
In Erlang, query was no longer a keyword since the commit
https://github.com/erlang/otp/commit/
0dc3a29744bed0b7f483ad72e19773dc0982ea50, 2012-11-20

### DIFF
--- a/Erlang/Erlang.sublime-syntax
+++ b/Erlang/Erlang.sublime-syntax
@@ -239,16 +239,6 @@ contexts:
           pop: true
         - include: internal-expression-punctuation
         - include: everything-else
-    - match: \b(query)\b
-      captures:
-        1: keyword.control.query.erlang
-      push:
-        - meta_scope: meta.expression.query.erlang
-        - match: \b(end)\b
-          captures:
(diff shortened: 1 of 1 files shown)


FichteFoll:
See https://github.com/sublimehq/Packages/pull/104 and https://github.com/sublimehq/Packages/pull/913 regarding the removal of highlighting (in default packages).

bitnitdit:
Each year, Erlang will release a new version. Deprecated features will only be kept in the next two versions providing two years period to developers to refactor the code base, then the 3rd new version Erlang compiler will report the deprecated features as errors. 

'query' as a keyword was deleted in 2012, but it was deprecated two years before 2012 and was considered to it should be deprecated several years before 2010. Although there is no official statistics, I think very few Erlang projects maintained under the Erlang compiler of 2011 version. Even so, the maintainers will likely to use sublime text 2 to maintain the old Erlang code base. 

Adding "some advanced matches" will add some unnecessary complexity at the same time.

PR #1137, opened 2017-09-13
[JavaScript] String template snippet adapted from existing implementation for Ruby

cottonflop:
Changes:
* Adds a snippet for `${}` when entering `$` in a JavaScript template literal context- adapted from the existing Ruby handling.

### DIFF
--- a/JavaScript/Default.sublime-keymap
+++ b/JavaScript/Default.sublime-keymap
@@ -1,0 +1,14 @@
+[
+	{ "keys": ["$"], "command": "insert_snippet", "args": {"contents": "\\${${1:$SELECTION}}$0"}, "context":
+		[
+			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{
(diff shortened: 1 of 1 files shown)


FichteFoll:
How does ruby handle the case where you want to insert a literal `$`? Does it just require you to delete the braces?

cottonflop:
Yes, but with the way braces are handled (by default, at least), a single backspace after the dollar is sufficient to clear it.

I'd imagine there might be some way to have it work around using a backslash to change scope, but then you're still using a second keypress.

I got so used to the way it's handled for Ruby that it became jarring when I work in JS- for me it's definitely been worth the tradeoff for those few times I need an actual dollar sign inside a template literal.

FichteFoll:
Considering that template literals use a different syntax than normal string, it is probably more likely you'd want the dollar sign not to be a literal and instead start a placeholder.

cottonflop:
Thanks! ❤️

PR #1139, opened 2017-09-13
[HTML] fix invalid comment syntax highlighting bug

blunket:
For comments with an odd number of hyphens on the closing tag such as `<!-- test ----->`, the hyphens are marked invalid but everything in the entire file after the closing tag becomes highlighted as a comment.

This update fixes that bug, and also improves how invalid comments are marked, to match exactly in accordance with the W3 validator.

Preview of bug and invalid comment marks before this fix:
image

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -29,9 +29,11 @@ contexts:
       scope: punctuation.definition.comment.html
       push:
         - meta_scope: comment.block.html
-        - match: '--\s*>'
+        - match: '(-*)--\s*>'
+          captures:
+            1: invalid.illegal.bad-comments-or-CDATA.html
           pop: true
-        - match: '-{2,}(?!-?\s*>)'
+        - match: -{2,}
           scope: invalid.illegal.bad-comments-or-CDATA.html
     - match: <!
       scope: punctuation.definition.tag.html
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -31,7 +31,7 @@ contexts:
         - meta_scope: comment.block.html
         - match: '--\s*>'
           pop: true
-        - match: '--(?!(>|->))'
+        - match: '-{2,}(?!-?\s*>)'
           scope: invalid.illegal.bad-comments-or-CDATA.html
     - match: <!
       scope: punctuation.definition.tag.html
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -31,7 +31,7 @@ contexts:
         - meta_scope: comment.block.html
         - match: '--\s*>'
           pop: true
-        - match: "--"
+        - match: '--(?!(>|->))'
           scope: invalid.illegal.bad-comments-or-CDATA.html
     - match: <!
       scope: punctuation.definition.tag.html
--- a/HTML/syntax_test_html.html
+++ b/HTML/syntax_test_html.html
@@ -51,6 +51,33 @@
         <!-- Comment -->
         ## ^ comment.block.html
 
+        <!----- hyphens - -- --- -->
+        ## ^^^^^^^^^^^^^^^^^^^^^^^^^ comment.block.html
+        ##^^ - invalid.illegal.bad-comments-or-CDATA.html
+        ##  ^^^ invalid.illegal.bad-comments-or-CDATA.html
+        ##                ^^ invalid.illegal.bad-comments-or-CDATA.html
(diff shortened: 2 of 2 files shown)


michaelblyons:
You clearly found a bug, but there are some problems with the regex in your PR. You'll notice that the trailing hyphens in odd-number sets are unhighlighted in your second screenshot (like GitHub's below?).

```html
<!------- test single - hyphens inside are okay. Doubles -- are not, --- nor triples --------->
```

Instead, might I suggest using this regex:
```yaml
        - match: -{2,}(?!-?\s*>)
          scope: invalid.illegal.bad-comments-or-CDATA.html
```

And add a syntax test so that no one breaks it in the future. 😉

blunket:
@michaelblyons Thank you very much, that is a much better regex. It is matching all of the invalid hyphens, exactly and perfectly.

I am having serious trouble testing the syntax file. Despite following [the documentation here](http://www.sublimetext.com/docs/3/syntax.html) I cannot get it to work:

image

Then again I am not sure I fully understand how the testing syntax works. I tried this -- it could be way off -- could you (or anyone) help me add the test for this?
```
<!----- - -- --- ---- -----  >
##      ^ comment.block.html
##  ^^^   ^^ ^^^ ^^^^ ^^^ invalid.illegal.bad-comments-or-CDATA.html
```

Any help or feedback would be much appreciated. I apologize for my limited experience. Thanks again.

michaelblyons:
I remember having that error when I first got started, but I don't exactly remember what I did to fix it. I do know that the first line (`## SYNTAX TEST "Packages/HTML/HTML.sublime-syntax"`) really wants to match the location (usually with symlinking) of the syntax file you're trying to test, relative to the `%APPDATA%/Sublime Text 3` folder.

Here's a stack of tests, though:

```html
<!----- hyphens - -- --- -->
## ^^^^^^^^^^^^^^^^^^^^^^^^^ comment.block.html
##^^ - invalid.illegal.bad-comments-or-CDATA.html
##  ^^^ invalid.illegal.bad-comments-or-CDATA.html
##                ^^ invalid.illegal.bad-comments-or-CDATA.html
##                   ^^^ invalid.illegal.bad-comments-or-CDATA.html
##                       ^^ - invalid.illegal.bad-comments-or-CDATA.html
##                          ^ - comment.block.html
<!---   hyphens - -- --- -- >
## ^^^^^^^^^^^^^^^^^^^^^^^^^^ comment.block.html
##^^ - invalid.illegal.bad-comments-or-CDATA.html
##                ^^ invalid.illegal.bad-comments-or-CDATA.html
##                   ^^^ invalid.illegal.bad-comments-or-CDATA.html
##                       ^^^ - invalid.illegal.bad-comments-or-CDATA.html
##                           ^ - comment.block.html
<!----- hyphens - -- --- ---->
## ^^^^^^^^^^^^^^^^^^^^^^^^^^^ comment.block.html
##^^ - invalid.illegal.bad-comments-or-CDATA.html
##  ^^^ invalid.illegal.bad-comments-or-CDATA.html
##                ^^ invalid.illegal.bad-comments-or-CDATA.html
##                   ^^^ invalid.illegal.bad-comments-or-CDATA.html
##                       ^^ invalid.illegal.bad-comments-or-CDATA.html
##                         ^^ - invalid.illegal.bad-comments-or-CDATA.html
##                            ^ - comment.block.html
```

Also, I'm not sure whether the regex I posted also has a bug. Notice that the middle test has three hyphens. I don't know whether its valid or not, but my regex lets it go. Same with three hyphens at the closing tag.

blunket:
@michaelblyons Thank you very much. Turns out I followed this repo's README a little too mindlessly (the command for Windows sets up the Python folder but not the HTML one).

I got the tests working now.

The [validator](https://validator.w3.org/nu/#textarea) doesn't have any issue with opening a comment like `<!---`.
However, it appears to have an issue with closing with ` --->`... Which the new regex doesn't catch.

michaelblyons:
Glad you got it working.

I think this should do the right thing, per how you describe the validator. I just shunted the burden onto the closing tag.

```yaml
    - match: <!--
      scope: punctuation.definition.comment.html
      push:
        - meta_scope: comment.block.html
        - match: '(-*)--\s*>'
          captures:
            1: invalid.illegal.bad-comments-or-CDATA.html
          pop: true
        - match: -{2,}
          scope: invalid.illegal.bad-comments-or-CDATA.html
```

blunket:
Thank you so much for all of your help @michaelblyons -- this seems to be working perfectly now. Added one final test and your new code there. Hope this can be merged.

wbond:
Thanks @blunket and @michaelblyons for this!

PR #1142, opened 2017-09-14
[Ruby] match /<regex>/ syntax explictly

### LINKED ISSUE #923: [Ruby] Incorrect parsing of division as regex when using instance variables
Sublime Text Build 3131

The following gets incorrectly marked as a regular expression:

```ruby
foo = 2 / @d
#         ^^ source.ruby string.regexp.classic.ruby
```

Removing the `@` fixes it:

```ruby
foo = 2 / d
#         ^ source.ruby
```

Issue also happens when not using assignment:

```ruby
2 / @d
#   ^^ source.ruby string.regexp.classic.ruby
```

### LINKED ISSUE #933: [Ruby] incorrect highlighting with slash and string
Using build 3131, I'm noticing a regression in parsing code like this:

    a = b / 'c'
    d = e

It seems the combination of the slash and the quoted string (whether single or double quoted) is causing the entirety of the next few lines lines to be interpreted as a string.

(If you're curious, the above syntax is common when using the Path gem, and concatenating pathnames together.)

### LINKED ISSUE #1006: [Ruby] Fail to highlight Ruby code with variable name 'index' correctly
image

This issue is similar to <https://bitbucket.org/birkenfeld/pygments-main/issues/1063/fail-to-highlight-ruby-code-with-index>.
#923 seems to be a similar issue.

### LINKED ISSUE #1136: [Ruby] `/ @` breaks syntax highlighting
Just upgraded to version 3.0 (congrats) but noticed it's introducing a syntax highlighting issue with ruby when you have a `/` followed by a `@`:

image

Removing either the `/` or the `@` stops the syntax highlighting from breaking.

I've replicated this across multiple OSX machines, with multiple color schemes.

randy3k:
close #923 
close #933 
close #1006 
close #1136 

The issue was looking ahead pattern cannot be popped immediately, see https://github.com/SublimeTextIssues/Core/issues/1906#issuecomment-329536967

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -1055,9 +1055,7 @@ contexts:
   try-regex:
     # Generally for multiline regexes, one of the %r forms below will be used,
     # so we bail out if we can't find a second / on the current line
-    - match: '(?=\s*/[^/]*$)'
-      pop: true
-    - match: '\s*(/)(?![*+{}?])'
+    - match: '\s*(/)(?![*+{}?])(?=.*/)'
       captures:
         1: string.regexp.classic.ruby punctuation.definition.string.ruby
       push:
--- a/Ruby/syntax_test_ruby.rb
+++ b/Ruby/syntax_test_ruby.rb
@@ -442,3 +442,13 @@ end
 
 p huh?("H", 1) + huh?("u") + huh?("gs")
 )
+
+
+# issue #923
+foo = 2 / @d
+#         ^^ - string.regexp.classic
+#         ^ variable.other.readwrite.instance punctuation.definition.variable
+
+# issue #933
+a = b / "c"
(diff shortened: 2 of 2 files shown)


wbond:
Excellent, thanks!

michaelblyons:
This needs a scope test for the `/` itself. As noted in #1150, it is missing `keyword.operator.arithmetic.ruby`.

varyform:
It's better than https://github.com/sublimehq/Packages/issues/1006#event-1250909843
But the highlighting is still broken:
screen shot 2017-10-10 at 17 34 17

gregblass:
I'm experiencing this in my current Sublime Text 3.

When I hit 'Check for updates', it says everything is up to date. I'm guessing this must be part of a package that I have to update?

Is there a recommended process to get this fix?

gregblass:
I figured it out. Just needed to read the README. Imagine that!

PR #1143, opened 2017-09-14
[Java] Fix methods ending in reserved words

### LINKED ISSUE #1141: [Java] Syntax highlighting failure if method name ends with 'class'
If the sequence of letters in the word 'class' appears at the *end* of a method name, syntax colouring is broken from that point on.

As per this example, which compiles and runs:

screen shot 2017-09-14 at 16 04 09

If you change any letter in the 'class' part of 'asSubclass', or add any letters to the end, eg: to '.asSublass(' or '.asSubclasse(' it all goes right. But of course, that's not the name of the method in java.lang.Class, so it doesn't compile...

(I've seen other problems with Java highlighting. This is just the first I've been able to isolate.)

Herewith pasted the code as well so you can try it...

```java
import java.text.*;

public class Test {
		try {
			Number n = NumberFormat.getInstance().parse("1234.56");
			System.out.println("Value "+n+" is a "+n.getClass().getName());
			Class<? extends Double> cdbl = n.getClass().asSubclass(Double.class);
			System.out.println("If we got here it's a Double as expected.");
			System.out.println("But nothing is syntax coloured any more");
			System.out.println("Nothing. It's all blank. Except the word 'class'");
			System.out.println("But remove one letter from the 'class' in 'asSubclass' and it all goes right.");
			System.out.println("(Except of course it doesn't compile because that's not the name of the method.)");
		} catch (ParseException pe) {
			// it's gone wrong here too hasn't it...
			System.err.println("Give us a number!");
		} catch (ClassCastException cce) {
			System.err.println("It wasn't a Double?");
		}
	}
}
```

michaelblyons:
Fix #1141

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -231,13 +231,13 @@ contexts:
           scope: keyword.operator.assert.expression-separator.java
         - include: code
   class:
-    - match: '(?=\w?[\w\s]*(?:class|(?:@)?interface|enum)(\s+\w+|\b))'
+    - match: '(?=[\w\s]*(?:\bclass|@?\binterface|\benum)\b)'
       push: class-declaration
 
   class-declaration:
     - meta_content_scope: meta.class.java
     - include: storage-modifiers
-    - match: (\bclass|(?:@)?\binterface|\benum)(?:\s+|\b)
+    - match: (\bclass|@?\binterface|\benum)(?:\s+|\b)
       captures:
         1: storage.type.java
       set:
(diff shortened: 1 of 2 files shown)


keith-hall:
This regex (before any changes even) seems a bit weird to me:

    (?=\w?[\w\s]*\b(?:class|(?:@)?interface|enum)(\s+\w+|\b))

- the first `\w?` seems pointless
- the wrapping of the `@` in a non capture group seems pointless
- the `(\s+\w+|\b)` doesn't need to be in a capture group, and just `\b` would have the same effect

FichteFoll:
Also, a test for this should be added imo.

michaelblyons:
I modified it, added a test, and moved the `\b` inside the parens so that `@interface` isn't skipped.

wbond:
Is this still relevant with all of the recent changes to Java by @ismell?

michaelblyons:
I think this is superseded by @ismell's work.

PR #1155, opened 2017-09-18
[LaTeX] extract the included languages to separate syntaxes

randy3k:
As discussed with @FichteFoll and @r-stein, the included languages are extracted to `LaTeX (minted)` and `LaTeX (listings)` syntaxes. In the new syntaxes, we import almost everything from `LaTeX.sublime-syntax` excepts

1. global-braces
2. lists
3. beamer

since they contain `include: main` which need to be redefined in the new files.

PS, there are actually some leftover patterns also containing `include: main`, but they are things like `\text{}`, `\box{}` which usually do not contain the minted or listings env.


FichteFoll:
Halving the file length of `LaTeX.sublime-syntax` is already a big win by itself, imo.

wbond:
Was this for performance reasons? If so, do we still need it now with build 3153?

randy3k:
Yap, it was because of performance. The opening speed of latex files has been much improved in the latest dev version. Great job.

I think we can close it now.

PR #1158, opened 2017-09-18
[C#] fix goto case bug

### LINKED ISSUE #1157: [C#] goto case breaks case highlighting
```cs
class Test {
    public void Main()
    {
        var test = "abc";
        var result = 0;
        switch (test[0])
        {
            case 'a':
                result += 4;
                goto case 'b';
            case 'b':
                result += 6;
                break;
            case 'c':
                result += 8;
                break;
        }
        result.Dump();
    }
}
```

in ST 3143, the `case` before `'b':` gets scoped as `invalid.illegal.unexpected.keyword.cs`, but this code compiles fine and the output is 10.

image

keith-hall:
fix #1157

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -835,7 +835,7 @@ contexts:
       captures:
         1: keyword.control.flow.goto.cs
         2: keyword.control.switch.case.cs
-      push: line_of_code_in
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #1159, opened 2017-09-18
Support #1152

### LINKED ISSUE #1152: [Ruby] Incorrect Indentation when private is on the same line as method definition
### Summary

When parsing ruby, sublime text doesn't correctly recognize method definitions when they occur on the same line as `private` modifiers.

### Steps to Reproduce

1. Type this snippet.
```ruby
private def foo
  a = 1
  b = 2
end
private
def var
  a = 1
  b = 2
end
```
2. Choose `Indentation: Reindent Lines` from the command palette.

### Expected behavior

3. Expected formatting:
```ruby
private def foo
  a = 1
  b = 2
end
private
def var
  a = 1
  b = 2
end
```

### Actual behavior
3. Actual formatting:
```ruby
private def foo
a = 1
b = 2
end
private
def var
  a = 1
  b = 2
end
```

## Second Repro Case

This also comes up while typing new code:

### Repro Steps

1. Type this snippet.
```ruby
private def foo[enter]
```

### Expected behaviour
```ruby
private def foo
  # Leading spaces.
```

### Actual behaviour
```ruby
private def foo
# No leading spaces.
```

## Other interesting cases:
All these cases seem to be un-indented.
```ruby
private def foo(bar)
  x = 1;
end
public def foo(bar)
  x = 1;
end
protected def foo(bar)
  x = 1;
end
```

### Environment

* Sublime Text Build 3143

Routed here from https://github.com/SublimeTextIssues/Core/issues/1923

shalecraig:
Expand method matching to match visibility modifiers before `def` statements to resolve #1152

This matches all 6 combinations of `private/public/protected` `_class_method`(/instance method).

Technically `protected_class_method` isn't a valid keyword, so I can modify this to not support that indentation if you prefer.

### DIFF
--- a/Ruby/Miscellaneous.tmPreferences
+++ b/Ruby/Miscellaneous.tmPreferences
@@ -13,7 +13,7 @@
 		<string><![CDATA[(?x)^
 		(\s*
 			(
-				module|class|(private(_class_method)?\s*)?def
+				module|class|((private|public|protected)(_class_method)?\s*)?def
 				|unless|if|else|elsif
 				|case|when
 				|begin|rescue|ensure
--- a/Ruby/Miscellaneous.tmPreferences
+++ b/Ruby/Miscellaneous.tmPreferences
@@ -13,7 +13,7 @@
 		<string><![CDATA[(?x)^
 		(\s*
 			(
-				module|class|def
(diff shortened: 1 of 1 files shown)


wbond:
This was resolved #1154.

PR #1160, opened 2017-09-19
[HTML] fix reindentation of style tag

keith-hall:
Prior to this PR, the following HTML gets reindented incorrectly:

- new tab
- set syntax to HTML
- `html` <kbd>Tab</kbd>
- type `<sty` <kbd>Tab</kbd><kbd>Enter</kbd>
- looks fine, but reindent it =

```html
<!DOCTYPE html>
<html>
<head>
    <title></title>
    <style type="text/css">
    
</style>
</head>
<body>

</body>
</html>
```

this PR fixes that by capturing the whitespace (specifically the newline) after the `<style>` tag, so that it doesn't get a CSS (`source.css`) scope, which was causing the CSS indentation rules take over, and thus not indenting after `<style>\n`. Which was a problem, because it unindents after `</style>` (due to the newline after that having the html scope), and thus the indentation was getting out of kilter.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -80,8 +80,9 @@ contexts:
             2: entity.name.tag.style.html
             3: punctuation.definition.tag.end.html
           pop: true
-        - match: '>'
-          scope: meta.tag.style.begin.html punctuation.definition.tag.end.html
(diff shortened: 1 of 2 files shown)


wbond:
Awesome!

PR #1161, opened 2017-09-19
[Markdown] scope link definitions with an entity.name scope

### LINKED ISSUE #1133: [Markdown] Link definitions should use an `entity.name` scope
Currently, identifiers of link definitions get the scope `constant.other.reference.link.markdown`, but an `entity.name` scope would make much more sense. It might make sense to index these as well (locally).

img

```
[JSON]: http://en.wikipedia.org/wiki/JSON
```

keith-hall:
...and show them in the local symbol index - fixes #1133

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -174,7 +174,7 @@ contexts:
         )
       captures:
         1: punctuation.definition.constant.begin.markdown
-        2: constant.other.reference.link.markdown
+        2: entity.name.reference.link.markdown
         3: punctuation.definition.constant.end.markdown
         4: punctuation.separator.key-value.markdown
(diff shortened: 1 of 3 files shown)


FichteFoll:
Since this is only the local symbol list (and not indexed), maybe links should be distinguished from headings by wrapping them in `[]`? I knpw headings have leading `#` but the difference is kind of subtle imo.

See also https://github.com/SublimeTextIssues/Core/issues/1676.

keith-hall:
I've opted to add the whole link to the local symbol list, meaning the link definitions all start with `[`... I feel it is more useful that way - one can filter by link description, url or name/number. Headings don't show the leading hashes in the symbol list, just indentation.

image

FichteFoll:
Oh that works too. For some reason I didn't even check the tmPreferences's selector 😅

Issue #1100, opened 2017-08-04
[CSS] Parentheses in parentheses still breaks highlighting; Property in animation looks strange.


PaperFlu:
While using CSS function in (),the ) ended prematurely by the function. There are two issues talking about this, #925 and #987 . Their problems have been solved but mine.

![Nested parens](https://user-images.githubusercontent.com/22674396/28977207-b1a659aa-7973-11e7-9f40-6d4db7a4e616.png)

The examples in the two issues work well, (117 - 119). The others, not mentioned in the two issues, ended prematurely and lose highlight, (120, 121). It seems that it's just a workaround for some known nested parens only. Maybe the core code in CSS still has this problem. Freely written 'top: a(());' will prove it.

Code:
```css
.Test {
	top: calc(100% - (1 * 10px) / 1 - (1 * 10px) / 1 - (1 * 10px) / 1);
	background-color: color(var(--background) blend(var(--forground) 80%));
	grid-template: repeat(2, var(--size)) / repeat(2, 50%);
	transform: translate(var(--center, 0)) scale(var(--ripple-scale, 1));
	top: a(());
	animation: 83ms opacity-in forwards;
}
```

Another puzzling thing, 'opacity' in animation is highlighting because it's a CSS property. By default, animation-name in animation is gray as '-in'. Maybe you want me to notice that it's a CSS property. But it looks a little strange and it's useless, right?

rwols:
What is `opacity-in` supposed to do and for which property is it a short-hand value for? (animation-name, animation-duration, animation-timing-function, animation-delay, animation-iteration-count, animation-direction, animation-fill-mode, animation-play-state).

You're right about `var` I believe.

PaperFlu:
@rwols 
Unbelievable my bad English is understandable!! I was sad because no one was leave a reply here.

The `opacity-in` was copied from Material Components by Google when I was reading its CSS file and trying to make a ripple component like its works by myself.

```css
@keyframes mdc-ripple-fg-opacity-in {
  from {
    opacity: 0;
    animation-timing-function: linear; }
  to {
    opacity: 1; } }
/* ...... */
.mdc-theme--dark .mdc-button.mdc-ripple-upgraded--foreground-activation::after {
  animation: 300ms mdc-ripple-fg-radius-in forwards, 83ms mdc-ripple-fg-opacity-in forwards; }
```

Thanks, I will name it `show` in mine. I was just thinking if a company be very strict to their coder's code.

[...]

PR #1162, opened 2017-09-19
[CSS] support 0 lengths i.e. for translateZ function and var function in other transform functions

### LINKED ISSUE #1066: (CSS) mini-bug: translateZ
```css
transform: translateZ(0);
                      ^
```
The value is not highlighted as a number, whereas it is for ```translateX``` and ```translateY```.

keith-hall:
fix #1066 and https://github.com/sublimehq/Packages/issues/1100


wbond:
Thank you!

PR #1167, opened 2017-09-22
[R] include comments in lambda function

randy3k:
issue from R-Box https://github.com/randy3k/R-Box/issues/113


wbond:
Thank you!

PR #1171, opened 2017-09-24
Git merge conflicts syntax mixin for other syntaxes

phts:
Introduce a mixin for git conflicts which can be included to other syntaxes via:

```yaml
contexts:
  prototype:
    - include: Packages/Git Conflicts/git-conflicts.sublime-syntax
```

This syntax makes git conflicts more noticeable in the code.

In this PR I also fixed CSS, HTML, JavaScript, Ruby, YAML syntaxes to enable highlighting of git conflicts.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -72,6 +72,9 @@ variables:
     )
 
 contexts:
+  prototype:
+    - include: Packages/Git Conflicts/git-conflicts.sublime-syntax
+
   main:
     - include: comment-block
     - include: selector
--- a/CSS/syntax_test_css_git_conflicts.css
+++ b/CSS/syntax_test_css_git_conflicts.css
@@ -1,0 +1,10 @@
+/* SYNTAX TEST "Packages/CSS/CSS.sublime-syntax" */
+
+<<<<<<< HEAD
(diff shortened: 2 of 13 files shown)


rwols:
https://forum.sublimetext.com/t/syntax-highlighting-and-merge-conflicts/30234

phts:
@rwols what did you mean by this forum post? We still don't have any working solution in ST, do we?

FichteFoll:
It includes some discussion about why something like this is problematic, since it messes with proper context stacking due to invalid syntax and nesting across these artificial tokens, which can be inserted *anywhere*.

wbond:
This does not seem like a likely candidate for inclusion, based on the previous discussion. I do, however, appreciate you efforts in trying to contribute. I hope you'll stick around and contribute to syntaxes for languages you use on a regular basis.

michaelblyons:
I'd also like to point you to [this existing community package][1] that seems relevant. (I, for one, have it installed.) If it could benefit from your attention, I'm sure a PR would be welcome there.

[1]: https://github.com/Zeeker/sublime-GitConflictResolver

michaelblyons:
FWIW, something very similar is included in the default packages now, beginning with https://github.com/sublimehq/Packages/pull/4047.

PR #1176, opened 2017-09-26
[JavaScript] Add CSS inspired keybindings

deathaxe:
The CSS package contains some nice key bindings to avoid double adding `;` or `,` and adds some tiny snippets.

This PR adds those helpers to the JSON file format as well.


FichteFoll:
`Auto-add : ,` might be controversial tbh, especially when the JS package is the only one receiving this with other languages being left out, e.g. Python. I like trailing commas in my Python source for example, but I know for certain that there are people who detest those and would rather not have to delete them every time.

(Disregard that this would be quite hard to implement in Python due to colons being used in dict comprehensions and to start a block. I'm just making an example here.)

PR #1177, opened 2017-09-26
[PHP] smarter indentation rules to ignore colons in strings

### LINKED ISSUE #1175: Auto-indent is indenting too far after inline if-statements in PHP
### Summary
After an inline if-statement **containing a string with a colon somewhere on the line**, ST3 is inserting an extra indent after pressing ENTER.

### Expected behavior
The indent should match the above line, not be further indented.

### Actual behavior
There is too much indentation.

### Steps to reproduce
1. Create a blank PHP file and open it in ST3.
2. Paste the following code:
```php
<?php
if ('a:b') return;
if ('a:b') { return; }
if (true) echo('a:b');
?>
```
3. Set the cursor after any line 2-4 and press ENTER.

### Environment
* Operating system and version: Windows 10 64-bit
* Monitor:
  * Resolution 1920x1080
* Sublime Text:
  * Version 3.0, Build 3143

keith-hall:
fixes #1175

### DIFF
--- a/HTML/Miscellaneous.tmPreferences
+++ b/HTML/Miscellaneous.tmPreferences
@@ -11,8 +11,7 @@
 		<string><![CDATA[(?x)
 			^\s*+                                                                                   # the beginning of the line followed by any amount of whitespace
 			(
-			    <\?(php)?\s+(else(if)?|end(if|for(each)?|while))\b                                  # ending PHP control keywords
-			|   [^<>]*+</(?!html)[A-Za-z0-9-]+\b[^>]*>                                              # optionally, anything that's not an angle bracket, followed by any valid HTML close tag except "html"
+			    [^<>]*+</(?!html)[A-Za-z0-9-]+\b[^>]*>                                              # optionally, anything that's not an angle bracket, followed by any valid HTML close tag except "html"
 			|   (<!\[endif\])?-->                                                                   # closing comment punctuation, optionally preceded by an end "comment selector"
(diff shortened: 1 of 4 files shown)


keith-hall:
Note that I've moved the PHP indentation rules out of the HTML package, to avoid duplication.
This could have a couple of small side effects for lines that have an open (unclosed) HTML tag or a close HTML tag before the PHP block i.e.:

```php
<? if (true): ?>
    <div><?
    // this line not indented
```

```php
<? if (true): ?>
    <div>
        // indented
        </div><? // this line not unindented
```

How often people write code like this, I don't know - but if we need to, I can see a couple of possibilites:

- undo this deduplication change, leaving PHP indentation rules in the HTML package (IMO bad because surely it would/should then become bloated with rules from ASP, JSP etc.)
- duplicate the HTML indentation rules into the PHP package (could consider whether the HTML is inside a string or before the `<?`) (IMO better than the current situation, despite the additional duplication, because it would make more sense for each package that has an embedded HTML syntax to have a copy of the HTML indentation rules)
- try to mitigate this, like in the examples above where only comments/whitespace come after the `<?` by scoping the `\n` as HTML, kinda like I did for `?>\n` in this PR. But this would get tricky because the syntax would need to check, do we want to defer to the HTML indentation rules or not (i.e. was there a control flow statement or a curly brace inside the PHP code on this line?)...

wbond:
Sounds good. I think option 2 is probably the direction we'd want to go. Thanks for this fix.

PR #1181, opened 2017-10-01
[Markdown] HTML blocks fix - follow CommonMark spec

### LINKED ISSUE #1180: [Markdown] HTML blocks should be able to interrupt paragraphs
At least according to the CommonMark spec (<http://spec.commonmark.org/0.28/#html-blocks>) in the following example the markdown syntax should treat everything after (and including) the `<p>` as raw HTML, so only the first `a` should be italicized.
```
*a*
<p>*a*</p>
*a*
```
Currently though the syntax only enters raw HTML mode (`meta.disable-markdown`) when the `<p>` tag is preceded by a new line:
```
*a*

<p>*a*</p>
*a*
```
But even in this case, the raw HTML mode should not be exited until an empty line is encountered, so the last `*a*` should not be italicized, as it currently is by the syntax.

keith-hall:
fixes #1180

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -201,16 +201,23 @@ contexts:
         - match: ^(?=\S)
           pop: true
         - include: list-paragraph
-    - match: '^[ ]{0,3}(?=<{{html_tag_block_end_at_close_tag}})'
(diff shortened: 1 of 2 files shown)


wbond:
Thank you

PR #1182, opened 2017-10-02
[HTML] correctly scope comment punctuation

keith-hall:
Prior to this PR, the HTML comment blocks had no scope at all applied to the closing comment punctuation (`-->`), and the opening comment punctuation (`<!--`) was missing the `.begin` that is common to all comment blocks. This PR fixes that, and brings it in line with all the other packages, which will allow color schemes to color said punctuation properly, and plugins to make proper use of it too.


wbond:
Excellent!

PR #1183, opened 2017-10-02
[PHP] Add scope to the nullable types symbol

jfcherng:
image

Add a scope (`storage.modifier.nullable.php`) to `question marks` in `function parameters` and the `return type hinting`.

```php
<?php

function testReturn(?string $str): ?string
{
    return $str;
}
```

Ref: http://php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types


jfcherng:
Wrong branch.

PR #1184, opened 2017-10-02
[PHP] Add a scope to the nullable types symbol

jfcherng:
image

Add a scope (`storage.type.nullable.php`) to `question marks` in `function parameters` and the `return type hinting`.

```php
<?php

function testReturn(?string $str): ?string
{
    return $str;
}
```

Ref: http://php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -586,6 +586,8 @@ contexts:
     # Class name type hint
     - match: '{{identifier}}'
       scope: meta.path.php support.class.php
+    - match: '\?'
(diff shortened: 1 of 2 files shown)


keith-hall:
C# uses `storage.type.nullable.cs` in this case, because it's part of the type - and I think it'd make sense to keep the definitions consistent.
Simple C# example `int? a;`

jfcherng:
@keith-hall Thanks for that information. Revised.

jfcherng:
merged to https://github.com/sublimehq/Packages/pull/1198

PR #1186, opened 2017-10-03
[Markdown] cope with stray underscores/asterisks in bold markup

### LINKED ISSUE #1163: [Markdown] Scope break if an non-letter is followed by underscore within bold markers
With the following text:
```
__test!_test__
```
The syntax leaves a scope `invalid.illegal.non-terminated.bold-italic.markdown`.
It actually appears when the middle '_' has a non alphanumerical character in front.

I guess the regex for detecting bold/italic end marker should be improved, but I didn't check.

keith-hall:
fixes #1163


wbond:
Thanks!

PR #1192, opened 2017-10-06
Add Java Completions

ismell:


jrappen:
To use completions containing a dot (`a.b`) you can "make" those using Python. Compare the `CSS` subfolder on how to.

wbond:
Thanks for adding these

PR #1193, opened 2017-10-06
Java: Add constant entities to symbol list

ismell:
Now that the java syntax has been updated we can add the constants to the symbol list.

### DIFF
--- a/Java/Symbol List%3A Constants.tmPreferences
+++ b/Java/Symbol List%3A Constants.tmPreferences
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+  <key>name</key>
+  <string>Symbol List: Constants</string>
+  <key>scope</key>
+  <string>source.java entity.name.constant.java</string>
+  <key>settings</key>
+  <dict>
+    <key>showInIndexedSymbolList</key>
+    <integer>1</integer>
+
+    <key>showInSymbolList</key>
+    <integer>1</integer>
+  </dict>
+</dict>
+</plist>


Java/Symbol List%3A Constants.tmPreferences
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>name</key>
+	<string>Symbol List: Constants</string>
+	<key>scope</key>
+	<string>(source.java) &amp; (entity.name.constant)</string>
```
wbond: This should be `source.java entity.name.constant`

Java/Symbol List%3A Constants.tmPreferences
```diff
@@ -0,0 +1,14 @@
...
+	<key>settings</key>
+	<dict>
+		<key>showInIndexedSymbolList</key>
```
wbond: Do you only want these to be indexed, not in the symbol list? If they should be in the symbol list, we'll need to add `showInSymbolList` with a value in `1`.

ismell:
lol, I feel dumb for not knowing how to reply to your comment or even find it anymore. I rebased the change and now your comment is gone...

Anyway, I added `showInSymbolList`. I would ideally like to omit `showInIndexedSymbolList` but then go to symbol doesn't work as expected.

It would be nice when performing a `go to symbol` if the indexed symbol list sorted the matching symbols declared in the current file first. That way I don't accidentally move to a different file with the same symbol.

FichteFoll:
Imo constant definitions should be added to the index and maybe even done by the Default package, as done for a couple other common entity scopes.

wbond:
@FichteFoll I'll discuss indexing constants by default with Jon as some point

PR #1194, opened 2017-10-06
[PHP] Fix ligature not working on comparison operators

jfcherng:
https://forum.sublimetext.com/t/dev-build-3145/32340/27

`!` operator has a higher priority than `!==` hence `!==` is cut into `!` and `==` causing ligature not working.

PR #1198, opened 2017-10-06
[PHP] Add some PHP 7.x features

jfcherng:
# General

- Fix ligature not working on comparison operators (#1194)

# PHP 7.0

- [spaceship operator](http://php.net/manual/en/migration70.new-features.php#migration70.new-features.spaceship-op)

# PHP 7.1

- [nullable types](http://php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types)
- [multi catch exception handling](http://php.net/manual/en/migration71.new-features.php#migration71.new-features.mulit-catch-exception-handling)


wbond:
Thank you!

PR #1199, opened 2017-10-07
[ShellScript] Fix not popping backtick after end-of-options token

rwols:
If we're inside backticks, we should look ahead for a backtick and then pop. Forgot to account for this after an end-of-options token. [See this forum post for reference](https://forum.sublimetext.com/t/bug-broken-highlighting-with-bash/32402).


wbond:
Thanks!

PR #1202, opened 2017-10-07
[C#] support case ... when syntax

### LINKED ISSUE #1201: C# switch case when not highlighted properly
C#7 allows one to add a `when` condition to a `case` inside a `switch` statement - this is currently not scoped properly.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/switch#the-case-statement-and-the-when-clause

```cs
switch (sh) {
    case Shape shape when sh.Area == 0:
```

currently ST scopes `Shape shape when` all as `variable.other`, but here `Shape` is a type and `when` is a keyword.

keith-hall:
fixes #1201

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1767,6 +1767,10 @@ contexts:
           push:
             - match: '{{name}}(?=\s*:)'
               scope: constant.other.cs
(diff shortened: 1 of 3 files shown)


wbond:
Thanks!

PR #1210, opened 2017-10-09
[C++] [Objective-C] Fix too greedy template bracket matchings

rwols:
- Use the generic_lookahead variable to scout ahead for a closing ">" when
  encountering an opening "<" for member variables/methods. [From here](https://github.com/pybind/pybind11/blob/1e6172d4055d654df466f32b8327de7145bc9ea3/include/pybind11/cast.h#L292).
- Inside of a template parameter list, make sure we match "<=" first, before
  we push another template scope. [From here](https://github.com/pybind/pybind11/blob/1e6172d4055d654df466f32b8327de7145bc9ea3/include/pybind11/cast.h#L928).

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -228,6 +228,10 @@ contexts:
     - include: keywords-parens
     - include: keywords
     - include: numbers
+    # Prevent a '<' from getting scoped as the start of another template
+    # parameter list, if in reality a less-than-or-equals sign is meant.
+    - match: <=
+      scope: keyword.operator.comparison.c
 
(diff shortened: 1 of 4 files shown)


wbond:
I had just fixed this for the next dev build, but I like you implementation better since it is more complete. Would you mind resolving the conflict and adding it to Objective-C++ also?

rwols:
I'm stuck at work at the moment but will update this this evening (GMT+1).

rwols:
Should be good now! By the way, that cast.h file is one of the worst files I've come across for highlighting.

wbond:
Thanks for these tweaks!

PR #1213, opened 2017-10-10
[C#] better C#7 support

keith-hall:
This PR adds much better C#7 support:

- [`out` variables](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#out-variables)
- [Tuples](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#tuples)
- [`ref` locals and returns](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#ref-locals-and-returns)
- [local functions](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#local-functions) (prior to this PR, it worked for functions that didn't return generic types)

NOTE: [numeric literal improvements](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#numeric-literal-syntax-improvements) remain in https://github.com/sublimehq/Packages/pull/1047

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -527,6 +527,17 @@ contexts:
         6: punctuation.section.brackets.end.cs
         7: keyword.operator.pointer.cs
       push: method_name
+    - match: \(
+      scope: punctuation.section.group.begin.cs
+      push:
+        - meta_scope: meta.group.cs
+        - match: \)
+          scope: punctuation.section.group.end.cs
+          set: method_name
+        - match: ','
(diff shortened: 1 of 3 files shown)


wbond:
Thanks!

keith-hall:
@michaelblyons you're absolutely right - apologies! not sure how I missed that, I guess I expected there to be a test for it already... this should be resolved by https://github.com/sublimehq/Packages/pull/1278

Issue #1095, opened 2017-07-31
[PHP] `require` is wrongly highlighted inside parentheses


jfcherng:
ST version: 3142 dev

![image](https://user-images.githubusercontent.com/6594915/28771544-e0d9efc2-7615-11e7-80a0-231213d02d4f.png)
```php
(require 'foo.php');
```

It's scoped as `constant.other.php` right now.
But it should be `keyword.control.import.include.php`.

`include` and `require` would return a value which could be then further used.

PR #1222, opened 2017-10-13
[PHP] Fix include/require could be used as an expression

jfcherng:
Fix https://github.com/sublimehq/Packages/issues/1095

Ref: http://php.net/manual/en/function.include.php#example-125


wbond:
Thanks!

PR #1224, opened 2017-10-13
[Python] strftime in format() strings

### LINKED ISSUE #1209: (Python) datetime formatting with format()
It skips fields highlighting and falls back to `%`.
```Python
'{0:%Y%m%d}'.format(datetime.date.today())
'{0:%Y}{0:%m}{0:%d}'.format(datetime.date.today())
```

michaelblyons:
Fixes #1209 

Note that this doesn't allow arbitrary strings as `strftime` formats. We can't tell the difference between a `datetime` argument and something else.

```python
'{0:%Y%m%d}'.format(datetime.date.today()) # legal, newly highlighted
'{0:%Y-%m-%d}'.format(datetime.date.today()) # legal, not highlighted

'{0:%Y%m%d}'.format('foo') # not legal, newly highlighted
'{0:%Y-%m-%d}'.format('foo') # not legal, not highlighted
```

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -44,6 +44,7 @@ variables:
       (?:\.\d+)?         # precision
       [bcdeEfFgGnosxX%]? # type
     )
+  strftime_spec: '(?:%(?:[aAwdbBGmyYHIpMSfzZjuUVWcxX%]|-[dmHIMSj]))'
 contexts:
   main:
     - include: statements
@@ -1034,7 +1035,7 @@ contexts:
       scope: constant.other.placeholder.python
       captures:
         2: variable.other.placeholder.python
-    - match: '%(?:[aAwdbBGmyYHIpMSfzZjuUVWcxX%]|-[dmHIMSj])'
+    - match: '{{strftime_spec}}'
       scope: constant.other.placeholder.python
     - match: '\{\{|\}\}'
       scope: constant.character.escape.python
@@ -1045,9 +1046,11 @@ contexts:
     - match: |- # simple form
         (?x)
         (\{)
-          (?: [\w.\[\]]+)?         # field_name
-          (   ! [ars])?            # conversion
-          (   : {{format_spec}} )? # format_spec
+          (?: [\w.\[\]]+)?             # field_name
+          (   ! [ars])?                # conversion
+          (   : (?:{{format_spec}}|    # format_spec OR
+                   {{strftime_spec}}+) # strftime_spec
+          )?
         (\})
       scope: constant.other.placeholder.python
       captures:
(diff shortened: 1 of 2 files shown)


FichteFoll:
Maybe we could make the match less restrictive and accept anything that's not a curly brace? The first part of the regex should be restrictive enough to not yield too many false positives, if any.

wbond:
@FichteFoll I'm not quite sure from you comment what you are suggesting, or trying to improve.

FichteFoll:
I meant to highlight something like `'{0:%Y-%m-%d}`, since basically nobody would use `{0:%Y%m%d}` (or at least very rarely). With this PR, only multiple %-sequences are recognized, as outlined in the OP.

It might be sensible not to put too many restrictions onto the allowed format string when trying to match for a pattern and instead just require a closing brace to be present. Then highlight any `%.` sequence, since in theory each class can implement their own `__format__`. That, or just stick to the strftime placeholders since those will be the most common. Or just highlight anything in the formatting part as `constant.other.format-spec`, which is what we do for f-strings and nested (complex) format placeholders already.

wbond:
@michaelblyons Could you update this to highlight the formatting part as `constant.other.format-spec`, or alternatively any `%` sequence? Perhaps add some tests for form such as `%Y-%m-%d`?

Python/Python.sublime-syntax
```diff
@@ -1045,9 +1046,11 @@ contexts:
     - match: |- # simple form
         (?x)
         (\{)
-          (?: [\w.\[\]]+)?         # field_name
-          (   ! [ars])?            # conversion
-          (   : {{format_spec}} )? # format_spec
+          (?: [\w.\[\]]+)?             # field_name
+          (   ! [ars])?                # conversion
+          (   : (?:{{format_spec}}|    # format_spec OR
+                   (?:[^}\n]*?%.)+?)   # any format-like string
```
FichteFoll: The problem with this regex is that it will only match when the `%.` pattern matches at the end. I suggest adding another `[^}]*` after it. `\n` is redundant because then the terminating `\}` would cause the match to fail anyway as there can't be any more characters after a line break. The quantifier after the group can be dropped as well then as the intend is to match at least one `%.` pattern, not every. The new 'format-like string' pattern should thus be:

```regex
[^}%]*%.[^}]*
```

The following code/test case should be able to show the issue:
```py
'{0:%Y-%m-%dT}'
```

michaelblyons: Good points, certainly. It doesn't (and shouldn't?) highlight the technically-legal but rather boring `'{0:T}'`.

wbond:
Thanks!

PR #1225, opened 2017-10-14
[ShellScript] Disable brace and pattern expansion in heredocs

### LINKED ISSUE #1221: [Bash] Brace expansion in heredocs screws up the syntax
```
sed 's/^	//' << EOF >$dummy.c
#ifdef __cplusplus
#include <stdio.h>  /* for printf() prototype */
	int main (int argc, char *argv[]) {
#else
	int main (argc, argv) int argc; char *argv[]; {
#endif
	#if defined (host_mips) && defined (MIPSEB)
	#if defined (SYSTYPE_SYSV)
	  printf ("mips-mips-riscos%ssysv\n", argv[1]); exit (0);
	#endif
	#if defined (SYSTYPE_SVR4)
	  printf ("mips-mips-riscos%ssvr4\n", argv[1]); exit (0);
	#endif
	#if defined (SYSTYPE_BSD43) || defined(SYSTYPE_BSD)
	  printf ("mips-mips-riscos%sbsd\n", argv[1]); exit (0);
	#endif
	#endif
	  exit (-1);
	}
EOF
# <- EOF not recognized due to non-closing } due to brace expansion
```
Guess I have something to do over the weekend.

rwols:
fix #1221

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -607,9 +607,15 @@ contexts:
       pop: true
 
   heredocs-body-common-with-expansion:
+    # [Bash] 3.6.6: all lines of the here-document are subjected to parameter
+    # expansion, command substitution, and arithmetic expansion, the character
+    # sequence \newline is ignored, and ‘\’ must be used to quote the
+    # characters ‘\’, ‘$’, and ‘`’.
     - match: \\[`$"\\]
       scope: constant.character.escape.backtick.shell
-    - include: expansion
+    - include: expansion-parameter
+    - include: expansion-arithmetic
+    - include: expansion-command
 
   heredocs-body-no-expansion:
     - meta_include_prototype: false
--- a/ShellScript/test/syntax_test_bash.sh
+++ b/ShellScript/test/syntax_test_bash.sh
@@ -1879,6 +1879,35 @@ cat << \FOO
 FOO
 # <- keyword.control.heredoc-token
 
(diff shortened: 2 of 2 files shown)


wbond:
I would think a better solution would be to come up with a lookahead that can detect if a curly brace appears to be part of an expansion, right?

rwols:
I've consulted the Bash bible.

> 3.6.6 Here Documents
> 
> This type of redirection instructs the shell to read input from the current source until a line containing only word (with no trailing blanks) is seen. All of the lines read up to that point are then used as the standard input (or file descriptor n if n is specified) for a command.
> 
> The format of here-documents is:
> 
>     [n]<<[-]word
>             here-document
>     delimiter
> No parameter and variable expansion, command substitution, arithmetic expansion, or filename expansion is performed on word. If any part of word is quoted, the delimiter is the result of quote removal on word, and the lines in the here-document are not expanded. **If word is unquoted, all lines of the here-document are subjected to parameter expansion, command substitution, and arithmetic expansion**, the character sequence \newline is ignored, and ‘\’ must be used to quote the characters ‘\’, ‘$’, and ‘`’.
> 
> If the redirection operator is ‘<<-’, then all leading tab characters are stripped from input lines and the line containing delimiter. This allows here-documents within shell scripts to be indented in a natural fashion.

That means:
- No brace expansion
- No pattern expansion
- No tilde expansion
- No job expansion

rwols:
Rebased onto latest upstream master, and I've squashed my two commits.

PR #1226, opened 2017-10-15
[ShellScript] Split up compound statement into backticks and non-backticks version

### LINKED ISSUE #1220: [Bash] command list in backticks buggo
```
foo=`(uname -r) 2>/dev/null`
#                          ^ punctuation.section.group.end - punctuation.section.group.begin
```

rwols:
fix #1220 

After merging this PR and https://github.com/sublimehq/Packages/pull/1225, [this file](https://github.com/llvm-mirror/llvm/blob/master/cmake/config.guess) should highlight correctly. The problematic regions are [here](https://github.com/llvm-mirror/llvm/blob/0446db2b0e822bb1abb094bf3ff3eda824ddc457/cmake/config.guess#L138-L141) and [here](https://github.com/llvm-mirror/llvm/blob/0446db2b0e822bb1abb094bf3ff3eda824ddc457/cmake/config.guess#L438-L458).

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -250,7 +250,7 @@ contexts:
       set:
         - meta_scope: meta.group.for.shell
         - include: cmd-args-boilerplate
-        - include: compound
+        - include: arithmetic
         - match: \bin\b
           scope: keyword.control.in.shell
 
@@ -769,6 +769,13 @@ contexts:
 
   cmd:
     - include: cmd-common
+    - match: \(
+      scope: punctuation.definition.compound.begin.shell
+      push:
+        - match: \)
+          scope: punctuation.definition.compound.end.shell
+          set: [cmd-post, cmd-args]
+        - include: main
     - include: scope:commands.builtin.shell.bash#main
     - match: \blet\b
       scope: support.function.let.bash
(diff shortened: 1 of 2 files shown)


FichteFoll:
Am I right to assume that you can nest backticks when they occur within parenthesis?

rwols:
Fun question! Let's see for bash 4.4...

```
#!/usr/bin/env bash
echo `(uname `echo "-a"`)`
```
results in:
```
/Users/raoulwols/dev/test.sh: command substitution: line 3: syntax error: unexpected end of file
/Users/raoulwols/dev/test.sh: command substitution: line 2: syntax error near unexpected token `)'
/Users/raoulwols/dev/test.sh: command substitution: line 2: `)'
echo -a
[Finished in 0.0s]
```
while this:
```
#!/usr/bin/env bash
echo `(uname $(echo "-a"))`
```
results in:
```
Darwin MBP-van-Raoul 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64
[Finished in 0.0s]
```

And this is allowed too:
```
#!/usr/bin/env bash
(uname `echo "-a"`)
```
And this lovely syntax works too:
```
#!/usr/bin/env bash
echo `(uname \`echo "-a"\`)`
```
So I'm guessing the same rules apply for backticks-inside-backticks: you have to escape the backticks if you want the backticks-in-backticks to be backticks.

All I could find about it in the reference manual was section 3.5.4:
> Command substitutions may be nested. To nest when using the backquoted form, escape the inner backquotes with backslashes.

FichteFoll:
So... You can only nest two layers I presume?

PR #1227, opened 2017-10-15
[PHP] Add constant PREG_JIT_STACKLIMIT_ERROR

jfcherng:
Unreadable syntax regex. I guess you dont even know what I add if I dont tell...

http://php.net/manual/en/function.preg-last-error.php#refsect1-function.preg-last-error-returnvalues


FichteFoll:
Maybe you should split that into multiple lines? Would certainly make it and diffs more readable.

PR #1229, opened 2017-10-16
[PHP] always keep the closing heredoc keyword unindented

keith-hall:
Previously, having code like

```php
<?php
function test () {
    echo <<<HTML
    <table class="table table-bordered table-striped sup1em" id="tableResponsive">
        <thead>
            <tr>
                <th class="text-center linhaStatus">ID Pedido</th>
                <th class="text-center">Cliente / Destino</th>
                <th class="text-center">Pedido</th>
                <th class="text-center linhaStatus">Gerenciar</th>
            </tr>
        </thead>
    </table>
HTML;
}
?>
```

and then doing a full "Reindent" would cause the `HTML;` to be indented and thus no longer close the HEREDOC. This PR fixes that by:

- scoping the semi-colon and newline after the heredoc (if present) with a meta scope
- targeting that meta scope with some indentation rules (this required the specificity of the main PHP indentation rules to be tweaked slightly) to:
  - ignore lines ending with this meta scope when calculating indentation
  - preserve the indentation (i.e. none) of lines ending with this meta scope

### DIFF
--- a/PHP/Indentation Rules - heredoc end.tmPreferences
+++ b/PHP/Indentation Rules - heredoc end.tmPreferences
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>name</key>
+	<string>Indentation Rules - HEREDOC end</string>
+	<key>scope</key>
+	<string>source.php meta.heredoc-end.php</string>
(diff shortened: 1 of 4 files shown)


wbond:
At some point we'll have to see if we can come up with a solution for indentation tests :-)

Issue #860, opened 2017-03-29
Add JSX Syntax


wbond:
We'd like to support JSX out of the box to provide a high-quality JS Goto Definition experience for users of React

Thom1729:
Is the goal to add JSX to the existing JavaScript package, or as a separate package? Or to replace the JavaScript syntax with a more full-featured one? I ask because, frustrated with the limitations of the built-in package and some issues with babel-sublime, I've nearly completed a new JS + JSX + Flow syntax.

wbond:
It would be a separate syntax, but if possible will include contexts from the existing JavaScript syntax. I am not sure if it would make sense to include in the current JavaScript package, or a new package.

We won't be replacing the existing JavaScript syntax, but instead fixing bugs and improving it over time. It got a big rewrite in 2016, and we wouldn't want to do another breaking change since the new syntax follows the scope naming guidelines for the most part.

It seems @ccampbell (who contributed a bunch towards the JavaScript syntax in 2016) has recently added a JSX package to Package Control that is based on top of the existing JavaScript syntax. I haven't looked at it yet, but that would probably be a reasonable place to start from. https://packagecontrol.io/packages/Simple%20JSX

In my opinion the biggest thing will be coming up with a good set of syntax tests to prevent regressions as we develop it.

aziz:
Please take a look at [Sublime Ecmascript](https://github.com/bathos/Ecmascript-Sublime/blob/master/ecmascript.sublime-syntax#L4892) syntax as well. It has a really good support for JSX. 

The only downside of it might be that it fails too early and marks a whole bunch of the view as invalid. For me it's not a big deal but I've seen others complaining about a similar thing here. 

In terms of completeness, correctness and granularity of scopes, [Sublime Ecmascript](https://github.com/bathos/Ecmascript-Sublime/blob/master/ecmascript.sublime-syntax#L4892) is phenomenal. For example using it you can differentiate `{}` when used for normal JS statements like `if` and `switch` from the ones that define an object or the ones that are used for functions or classes, the same thing is true for `()`. These things can not be done with the built-in JS syntax or any another popular JS syntax definition on package control (like babel-sublime). Unfortunately the author is not abiding to the new scope naming rules we're trying to promote here and I don't see it being merged to the default packages repo but it could be used as a perfect reference. 

Note that Sublime Ecmascript syntax is 5038 lines of code compared to only 1476 lines for the official one here.

ccampbell:
[...]

PR #1236, opened 2017-10-17
JavaScript with JSX and ES7 decorators

Kronuz:
This adds basic JSX and ES7 decorators to the JavaScript

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -359,6 +359,8 @@ contexts:
       pop: true
 
   expressions:
+    - include: literal-jsx
+    - include: es7-decorators
     - include: regexp-complete
     - include: literal-string
     - include: literal-string-template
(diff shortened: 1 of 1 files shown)


rwols:
This will most definitely result in a merge conflict if https://github.com/sublimehq/Packages/pull/1009 gets merged, perhaps branch off of that and see where you land?

FichteFoll:
I'm also not sure whether it was decided to have jsx be a separate syntax or to include it in the Javascript syntax. See the separate issue for details.

Thom1729:
#1009 is merged, and it does look as though the changes conflict. If it helps, I intend to implement decorators as part of #1269 within the next few weeks. (I'm waiting on #1292 to lessen the risk of conflicts, but that's a much smaller change than #1009 was.)

Per #860, the current plan is to add a separate JSX syntax. There are some unanswered questions about the details of maintaining two syntaxes like that. In any case, it would probably be best to finish up #1292 first.

michaelblyons:
Is this obsolete, now that #1483 is merged?

Thom1729:
It is.

wbond:
Decorators are now part of JavaScript, and if we include support for JSX, it will be a separate syntax.

PR #1239, opened 2017-10-17
[Python] Fix line continuation not popping properly

FichteFoll:
Discovered randomly while investigating some code base a few weeks ago.

A string on the continued line would cause the pop pattern not being
matched, which in return meant the line continuation context wasn't
exited and whitespaces until the first non-string character were
consumed, which in return caused def and class statements to match as
invalid, since those require to be written on their own line.

(What a nice sentence.)

Previously failing test:


wbond:
Thanks!

PR #1241, opened 2017-10-17
[Python] Exclude string type from string scope

FichteFoll:
The string prefixes are not part of the actual string, so they shouldn't be scoped as such. Should ease the life of BracketHighlighter, for example (cc @facelessuser).


wbond:
Thanks!

PR #1253, opened 2017-10-23
[RST] Support multi-line comments

akrabat:
In reStructured Text, comments are multiline if they are indented after the initial `..`. Hence we need to capture up to, but not including the first non-whitespace character on a subsequent line.

See http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#comments.

### DIFF
--- a/RestructuredText/reStructuredText.sublime-syntax
+++ b/RestructuredText/reStructuredText.sublime-syntax
@@ -196,5 +196,5 @@ contexts:
         1: punctuation.definition.comment.restructuredtext
       push:
         - meta_scope: comment.line.double-dot.restructuredtext
-        - match: $\n?
+        - match: '^(?=\S)'
           pop: true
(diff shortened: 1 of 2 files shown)


keith-hall:
Any chance you'd like to take a stab at creating a syntax test file for reStructured Text, to help ensure your change doesn't get regressed in the future? :)

If so, docs are at http://www.sublimetext.com/docs/3/syntax.html#testing, and I highly recommend installing https://packagecontrol.io/packages/PackageDev to help with building the test assertions

akrabat:
@keith-hall Done. The multi-line comment test fails without the change in this PR.

I've also put in a few tests for paragraph, heading, italic and bold to make it easier for others to see how to add to it in the future.

FichteFoll:
The default reST syntax definitely needs some love, so thanks for taking a shot at it and including some basic tests.

wbond:
Thank you @akrabat!

PR #1255, opened 2017-10-23
[Python] fix nested string placeholder detection being too eager

### LINKED ISSUE #1252: Python | Highlighting breaks out of string
# Summary
This issue causes any code beyond a string including a character between 2 open braces to be incorrectly highlighted.
# Replicate issue
Set Sublime syntax to Python. Paste either of the following examples: 
### Example 1
```python
# Version 3.0, Build 3143
# Syntax highlighting error
foo = "{text{" # Comment
bar = "}}" # Comment

print(foo),
print(bar)
```
As long as there is a character between the two braces, highlighting will continue out of the string.
### Example 2
```python
sampleString = "{ {" # Incorrectly highlighted comment "}", return still incorrect "in a comment" 
```
_Python 2.7.x and 3.6.x runs both examples correctly without any issues._

keith-hall:
Fix #1252

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1055,7 +1055,7 @@ contexts:
         2: storage.modifier.conversion.python
         3: constant.other.format-spec.python
         4: punctuation.definition.placeholder.end.python
-    - match: \{(?=[^\}"']+\{) # complex (nested) form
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

Issue #921, opened 2017-04-22
[Clojure] Include support for multiline comments?


Qwerp-Derp:
In Clojure, there are multi-line comments:

    #_(
      This
      is
      a multiline
      comment)

Anything in between the `#_(` and `)` are ignored by the Clojure compiler - I was wondering if there could be support for this?

keith-hall:
I'm not a Clojure user but I was interested, so I looked this up in [the Clojure docs](https://clojure.org/reference/reader#_dispatch):

> *Ignore next form (`#_`)*
The form following `#_` is completely skipped by the reader. (This is a more complete removal than the comment macro which yields nil).

The question I have is, should we also update the [Clojure `Comment.tmPreferences` file](https://github.com/sublimehq/Packages/blob/61bfd68ea07a3aede68cd6a65c80be702b3d6db8/Clojure/Comment.tmPreferences) with block comment support?

keith-hall:
Can you link to some docs where it mentions `#+` is a multiline comment for Common Lisp please? http://www.gigamonkeys.com/book/syntax-and-semantics.html doesn't mention it, and [the answers on this SO question](http://stackoverflow.com/questions/6365334/lisp-commenting-convention) mention `#|`...`|#`, which is already catered for, and `#+(or)`, which is quite different.

Qwerp-Derp:
@keith-hall How is `#+(or)` different? Pretty sure the code inside the `#+` thing are ignored, so they're essentially comments.

keith-hall:
I think my point was that `#+or` [isn't technically a comment, even if it can be used to disable code](https://stackoverflow.com/a/8653748/4473405). Other syntaxes that offer similar features don't scope it as comments atm (i.e. C# preprocessor directives like `#if false`), and some people might want the disabled code to be displayed normally, and not as a comment, so some other solution would be ideal. Maybe a meta scope like `meta.disabled` or `meta.code-unreachable`, but color schemes aren't supposed to target those directly...

Qwerp-Derp:
Hmmm... well, I'll edit my starting message, then.

Also, I made a pull request for Clojure (#996) to include support for missing keywords and support for colon keywords (e.g. `:foo`). Can you have a look?

mitranim:
[...]

Issue #807, opened 2017-02-13
[Clojure] Punctuation is highlighted differently than rest of word


hanyuone:
![beautiful](https://cloud.githubusercontent.com/assets/14177193/22876007/a4fa27a0-f222-11e6-9e10-4ff29057082e.PNG)

### Code:

```clojure
(ns experiments.core
  (:gen-class)) 

(defn bubble-sort [array]
  (loop [ind 0
         sort-arr array
         swaps? 0]
    (if (= ind (- (count array) 1))
      (if (not= swaps? 0)
        (bubble-sort sort-arr))
      (if (> (nth sort-arr ind) (nth sort-arr (+ ind 1)))
        (let
          [temp-arr
            (vec
              (concat
                (subvec sort-arr 0 ind)
                [(nth sort-arr (+ ind 1))]
                [(nth sort-arr ind)]
                (subvec sort-arr (+ ind 2))))]
          (do
            (println temp-arr)
            (recur (+ ind 1) temp-arr (+ swaps? 1))))
        (recur (+ ind 1) sort-arr swaps?)))))

(defn -main [& args]
  (bubble-sort [6 5 3 1 8 7 2 4]))
```

I'm not sure if this is intended, but the punctuation (e.g. the `?` in `swaps?` or `-` in `-main`) is highlighted as punctuation, and is separate from the actual word itself.

wbond:
This issue has been solved by a previous commit, hence it is being closed.

Issue #795, opened 2017-02-01
[Clojure] Add support for keywords


hanyuone:
The syntax highlighter for Clojure doesn't support the valid highlighting of keywords, which are basically strings with a colon at the front:

```clojure
:foo ; This is a keyword
```

Currently, though, Sublime highlights keywords like so (in Monokai):
![clojure](https://cloud.githubusercontent.com/assets/14177193/22503763/5a32711e-e8c7-11e6-93f7-52e6d07cd9d3.PNG)

Code:

```clojure
(defn change-disp []
  [:div.change-disp
   [:input {:type "button"
            :value "Display stuff"
            :on-click #(swap! display? not)}]])
```

The colons are highlighted (presumably because it's punctuation), but the rest isn't highlighted.

djspiewak:
For context, clojure keywords should be scoped the same as Ruby/Scala symbols, since their usage is almost equivalent.

Qwerp-Derp:
Also, functions with slashes in them (functions from other Clojure files) aren't highlighted at all - only the `/` is highlighted:

![clojure2](https://cloud.githubusercontent.com/assets/14177193/22544082/71f39cb8-e987-11e6-9d5b-5f6967682af3.png)

(This also demonstrates the first issue.)

FichteFoll:
Please include the text used in screenshots for easier reproduction via copy and paste.

Qwerp-Derp:
@FichteFoll I included that.

Qwerp-Derp:
Bump: anything useful yet?

keith-hall:
Seems like you could just modify your color scheme to highlight `source.symbol.clojure`, no?

Qwerp-Derp:
@keith-hall That did (partially) solve the issue for the highlighter, but the `:` was still being highlighted incorrectly as punctuation. I believe this might be a "bug" (maybe not) with the actual Clojure syntax thing itself.

Picture and code for reference:

![clojure](https://cloud.githubusercontent.com/assets/14177193/23158507/903fcc0a-f873-11e6-9dbc-a9a2c233fc63.PNG)

```clojure
(defproject experiments "0.1.0-SNAPSHOT"
  :description "FIXME: write description"
  :url "http://example.com/FIXME"
  :license {:name "Eclipse Public License"
            :url "http://www.eclipse.org/legal/epl-v10.html"}
            
  :dependencies [[org.clojure/clojure "1.8.0"]]
  :main experiments.core/-main)
```

[...]

Issue #582, opened 2016-08-15
Discussion: Scopes for keywords affecting flow and type (e.g. `function` and `class`)


FichteFoll:
Currently, there appears to be a discrepancy with scope names for keywords that indicate a function or class body (or similar), where some get a `storage.type` scope name and some a `keyword` scope.

Quoting myself from #550:

> In Python, `def` and `class` _are_ keywords as well, but they not only indicate a different type but also have different effects on the following block (and what goes into the parens following the identifier). To me, this is a pretty sure indicator that scoping both `def` and `class` as `storage.type` is wrong since they actually have keyword-like effects on code. `storage.type` should most likely only be used for staticly typed languages like C and Java.
> 
> The exception to this would be the `async` "keyword", which certainly should get a `storage` scope, although `storage.modifier`. Furthermore, the `global` and `nonlocal` keywords are semantically tied to referring to storage, so it makes sense for those to get a storage scope too. `storage.scope` or similar?
> 
> In JavaScript we have `var`,  `let` and `const`. I would most likely mark those as `storage.type` and give `class` and `function` a `keyword.` scope for the same reason as for Python.

---

Carrying this over from #550, which is where this was raised initially. I highly suggest reading that issue in full.

wbond:
My current implementations and the official Sublime Text docs are currently based on the [TextMate docs](https://manual.macromates.com/en/language_grammars), which explicitly state that keywords such as `class` and `function` should be `storage.type`.

It does seem in popular usage that scoping has split where some use `storage.type` and others use `keyword`. In one sense, `keyword` is sort of broad, and can be argued for almost anything. Additionally, `storage.type` does not strictly mean a "type" as in an `integer`. Instead, I think about it more as "classification", or something like that. Considering `storage.modifier`, most of those are "keywords" also, but there is value is distinguishing them. 

I think one of the major downsides of changing `storage.type` to `keyword` for `class`, `function`, `trait`, `enum`, `impl`, etc is the color shift among users. This will affect a broad number of users, so I don't want to take the decision too lightly.

djspiewak:
[...]

PR #1259, opened 2017-10-24
[Clojure] syntax rework

mitranim:
This PR is a complete rework of the Clojure syntax.

Improvements:

  * The syntax structure is modeled after the AST and the Clojure reader. Should
    handle any formatting, regardless of whitespace and indentation quirks.

  * No unnecessary special cases. Simple and consistent rules. User-defined
    syntax should look and feel the same as standard syntax.

  * All syntactic forms are correctly scoped in file root

  * Much better at handling literals such as numbers, keywords, etc

  * Works for custom defs, e.g. `defroute`

  * Correctly declares `definterface`, `defprotocol`, `deftype`, `defrecord` types

  * Correctly highlights methods in `definterface`, `defprotocol`, `deftype`, `defrecord`

  * Correctly declares functions defined by `defprotocol`

  * Correctly declares names in the presence of metadata, whitespace, comments

  * Correctly handles `defmulti` and `defmethod`

  * Doesn't declare `declare` (noise)

  * Highlights mismatched brackets, helping with balancing

  * Consistent highlighting of function calls

  * ≈ 4 times smaller

  * ≈ 6 times faster

Doesn't detect function parameters and local bindings. I'm not sold on supporting 8+ special assignment forms, each with their own quirks, in a language with macros. It immediately breaks when a user introduces their own variation ([example](https://gist.github.com/Mitranim/148f0a0b51f36ef6a2cd3a0996dc7104) of a slightly different `let` I use in practice). Suggestions are welcome.

Would appreciate review.

**Edit:** while the PR is incubating, the syntax is available as standalone at https://github.com/Mitranim/sublime-clojure


jrappen:
## 1)

⚠️ There are no tests in your test files...?

* https://www.sublimetext.com/docs/3/syntax.html
* https://www.sublimetext.com/docs/3/syntax.html#testing
* https://www.sublimetext.com/docs/3/scope_naming.html

Compare the test for Clojure:

* [`./Clojure/syntax_test_clojure.clj`](https://github.com/sublimehq/Packages/blob/928a7a6/Clojure/syntax_test_clojure.clj)

Also, your package structure:

* One test file:
  ```text
  /Clojure
      Clojure.sublime-syntax
      syntax_test_clojure.clj
  ```
* several test files:
  ```text
  /Clojure
      Clojure.sublime-syntax
      /tests
          syntax_test_clojure.clj
          syntax_test_clojure_benchmark.clj
          ...
  ```

## 2)

Using whitespace for a vertical separation of blocks is nice to the eyes (at least mine...). For the syntax files as well as the test(s).

Compare the Python subfolder:

[`./Python`](https://github.com/sublimehq/Packages/tree/928a7a6/Python)

also great, LaTeX:

[`./LaTeX`](https://github.com/sublimehq/Packages/tree/928a7a6/LaTeX)

Clojure/Clojure.sublime-syntax
```diff
@@ -1,1123 +1,269 @@
...
+    [-+]?
+    (?:
+      \d+N?                             # int
+      |0[Xx][0-9A-Fa-f]+N?              # hex int
+      |[1-9]\d*[Rr][0-9A-z]+            # n-radix int
+      |\d+/\d+                          # ratio
+      |\d+(?:\.\d+)?(?:[Ee][-+]?\d+)?M? # decimal with scientific notation
+    )
+    (?=[{{non_number_chars}}])
+  reader_macro_scope: constant.other.symbol.reader-macro.clojure
```
rwols: You cannot assign variables to scopes. Remove this variable.

Clojure/Clojure.sublime-syntax
```diff
@@ -1,1123 +1,269 @@
...
-              pop: true
-            - include: binding
-    - match: '[^\s]'
-      push:
-        - meta_scope: invalid.illegal.bindings.clojure
-        - match: '[^\]]'
-          pop: true
-  bindings_form:
+    - include: match_expr
+  match_expr:
```
rwols: Style: use-dashes-for-contexts, and use_underscores_for_variables.

Clojure/Clojure.sublime-syntax
```diff
@@ -1,1123 +1,269 @@
...
+    - include: match_noise
+    - match: '[)\]}]'
+      scope: invalid.illegal.clojure
+    - match: '\('
```
rwols: You can omit the redundant single quotes around this regex, and various other regexes. Download PackageDev and enable the "Sublime Syntax" syntax from that package to better see what is possible and what is not possible within regexes. I highly recommend that package.

Clojure/Clojure.sublime-syntax
```diff
@@ -1,1123 +1,269 @@
...
+      scope: punctuation.section.parens.end.clojure
+      pop: true
+    - match: '\('
+      scope: punctuation.section.parens.begin.clojure
+      push: pop_fn_list_tail
+    - include: match_expr
+  pop_fn_list_tail:
+    - include: match_constant_set_normal_list_tail
+    - match: '{{symbol}}'
+      scope: entity.name.fn.clojure
```
rwols: entity.name.function?

Clojure/syntax_test_clojure.clj
```diff
@@ -1,107 +1,666 @@
...
-;;                              ^ invalid.illegal.unexpected-quantifier.regexp
-;;                                       ^^ keyword.operator.quantifier.regexp
-;;                                                 ^ keyword.operator.quantifier.regexp
-
-  (def hello (fn [] #"(?=unclosed_paren"))
-;;                  ^^^^^^^^^^^^^^^^^^^^ string.regexp.clojure
-;;                  ^^ punctuation.definition.string.begin.clojure
-;;                                     ^ punctuation.definition.string.end.clojure
-;;                                      ^ - string.regexp.clojure
+; SYNTAX TEST "Packages/Clojure/Clojure.sublime-syntax"
```
rwols: The point of these syntax test files is that you *test* the syntax. For example,
```
"blah \" blah"
; <- punctuation.section
; ^ string.quoted.double
```
Put your build system on "Automatic" and hit "Build" to run these two tests.

wbond:
The existing tests should remain. If scope names need to change, a brief explanation of why should be sufficient.

As @rwols mentioned, it appears your additions to the test file doesn't actually add any assertions. From this I presume you weren't running the Syntax Tests build system when developing this, correct? If so, you should have gotten an indication that there were 0 assertions.

Mitranim:
Hey guys, thanks for the reviews and instructions!

@jrappen Added whitespace for clarity, moved tests under.

@rwols

Thanks for suggesting PackageDev. It's really cool!

Addressed the cosmetics: junk scope variables, quotes, context naming.

Why `entity.name.fn` over `entity.name.function` for lambdas: personally, I prefer keeping lambdas out of the symbol index. Since tail recursion uses `recur`, `fn` names are useful only for better stacktraces, and you never use that name outside the function body. It just interferes with search. Personal preference! Receptive to opinions.

@rwols, @wbond

Tests: I focused on covering as many edge cases as possible, since existing tests were wildly insufficient. Is there a way to automatically add missing assertions to the test file? (If not, this would be really cool to have in PackageDev.)

On a tangential note: what does `Symbol List` do?

jrappen:
@Mitranim

* How are you running the tests? You can run different tests on your syntax files. Open the syntax file, then from the menu choose "Tools > Build with ...". You can see that you can run:
  * Syntax tests
    * what @rwols and me are talking about, see <http://www.sublimetext.com/docs/3/syntax.html#testing>
  * Syntax tests - All Syntaxes
  * Syntax tests - Performance
    * what you are talking about
  * Syntax tests - Regex compatibility
* When looking at [#1259`/files`](https://github.com/sublimehq/Packages/pull/1259/files) it shows me that you removed the syntax file completely. There is a file in your master branch though, not sure what's happening here. I tested on various machines and browers.
* Regarding `Symbol List`: Open a file, then type <kbd>Ctrl</kbd><kbd>R</kbd> or <kbd>Ctrl</kbd><kbd>P</kbd> followed by an <kbd>@</kbd>. For how to use them compare the metadata files in this repo (`*.tmPreferences`) for various syntaxes. There are three available settings:
  * showInSymbolList
  * showInIndexedSymbolList
  * symbolTransformation

Mitranim:
@jrappen 

For now, those tests are for 'manual' reading. I understand how to run the syntax test, but prioritised covering all edge cases first, since it's more useful than half a page of assertions that cover like 5%. Now the question is: is there a tool to automatically add missing assertions to the test? That would be just perfect.

Pretty sure the syntax file is there. Github collapses the diff because it's huge.

Thanks for explaining the symbol lists. Looks like the Clojure symbol list references a scope (`entity.global.clojure`) that's unused in both versions of the syntax. Maybe we should just delete it for now?

jrappen:
@Mitranim I forgot to mention `showInIndexedReferenceList` (Build 3145 or later).

wbond:
@Mitranim No, there is no tool to automatically generate the assertions. From a logical standpoint they wouldn't be testing much since they would just be using what you've already codified as the scope names in the syntax.

When I have some time, I'll go through some of the scope naming. Hopefully you read http://www.sublimetext.com/docs/3/scope_naming.html before writing the syntax.

In terms of `entity.name.fn` and the function list: the default syntaxes aren't a place to customize things to your liking. Plus, the correct way to not have certain scopes indexed or in the symbol list is to create a `.tmPreferences` file in your `Packages/User/` folder that turns off indexing/symbol list for those scopes. We'll definitely want to use the standardized scopes in this syntax, especially since so much work went into it.

Unless there is a good reason to delete the old tests, I'd really prefer they be put back. You should be able to leave them as-is and add your own tests afterwards.

The final question I have for now is: did you run the Syntax Tests: Regex Compatibility build system to see if any incompatible regex features were used?

Mitranim:
Thanks for all the comments folks! Pushed an update following your suggestions, plus extra corrections:

* added autotests
* corrected lambda scopes, banned them from indexing using the "symbol list" file (@jrappen, @wbond, thanks for the hints)
* restored regex support
* added `reify` and `proxy` support
* restored the old test file, but without the meta scope assertions

@wbond  Yep, it passes the regex compliance test.

Please let me know if you feel the giant test suite is too verbose for maintenance. I felt like comprehensiveness was a priority, on the grounds that the previous tests didn't keep the syntax from being buggy. Most of the rules are there because I broke them at some point while writing the syntax, and future changes could do the same.

Kept the "human-only" suite since it's easier to review on a glance. Maybe should remove that.

Haven't added meta scopes yet. I've seen quite a few syntaxes with flat scopes (1-2 levels everywhere), still unsure.

There might be other lurking issues. Please feel free to make suggestions or edits. I'm at the end of my wits here, might need a few days' break. 😅

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -534,7 +534,7 @@
 ;          ^^^^^^ string.clojure
 ;                ^ punctuation.section.parens.end.clojure
 ;                  ^ punctuation.section.parens.begin.clojure
-;                    ^^^^^^
+;                    ^^^^^^ string.clojure
```
rwols: maybe use `string.quoted.double.clojure`?

Mitranim: renamed `string.clojure` → `string.quoted.double.clojure` everywhere?

jrappen:
* The `uuid` in `*.tmPreferences` files is redundant if not used (... by Python files to access them or whatever).
* While at it, maybe you could fix some of the open clojure issues?

Mitranim:
@jrappen 

Removed unused `uuid`.

Looked into the open Clojure issues:

* #921 looks reasonable; turned out that this syntax uses the stack in a way that makes it difficult to add meta scopes, so this will take some time to figure out; I'll try to look into it later
* #807 is fixed automatically
* #795 is fixed automatically
* #481 is fixed automatically
* not sure if we can do #582, seems like it needs a big buy-in from color scheme developers and users

Mitranim:
Additional improvement: disabled apostrophe auto-pairing outside of comments and strings.

Mitranim:
Corrected detection of n-radix ints. (They no longer accidentally grab non-alphanumeric symbols.)

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -0,0 +1,2007 @@
...
+;                ^- constant
+;                 ^^^^^^^^^^^^^^^ constant.numeric.clojure
+;                                 ^^^^^^^^^^^^^^^ constant.numeric.clojure
+;                                                 ^^^^^^^^^^^^^^^ constant.numeric.clojure
+
+; ## Breaks
+
+  10,20,30
+; ^^ constant.numeric.clojure
+;   ^ comment.punctuation.comma.clojure
```
wbond: `comment.punctuation` seems like a funny scope name. Are the commas really comments and not some sort of separator?

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -0,0 +1,2007 @@
...
+; ^^^^ constant.language.clojure
+;     ^ comment.punctuation.comma.clojure
+  true;false;nil
+; ^^^^ constant.language.clojure
+;     ^ comment.line.clojure punctuation.definition.comment
+
+; ## Unaffected
+
+  'nil (true) (nil)
+; ^ constant.other.symbol.reader-macro.clojure
```
wbond: Can you explain a little bit about what sort of semantics these `constant.other.symbol.reader-macro` symbols are? It seems like a funny scope name to me, however I am not familiar with Clojure, and only slightly familiar with lisp-style languages.

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -0,0 +1,2007 @@
...
+
+  1234 1234N +1234 +1234N -1234 -1234N
+; ^^^^ constant.numeric.clojure
+;     ^- constant
+;      ^^^^^ constant.numeric.clojure
+;            ^^^^^ constant.numeric.clojure
+;                  ^^^^^^ constant.numeric.clojure
+;                         ^^^^^ constant.numeric.clojure
+;                               ^^^^^^ constant.numeric.clojure
+  0x1234af 0x1234afN 0X1234AF 0X1234AFN
```
wbond: What do the various `af` and `n` suffixes mean? We have some conventions we've been applying to recent work on syntaxes where we apply scopes to target storage/type info in the numeric literals. My hunch is they may be applicable here also. See https://github.com/sublimehq/Packages/blob/master/Python/syntax_test_python.py#L935-L940 for an example.

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -0,0 +1,2007 @@
...
+;                                                  ^^^^^^^^^^ constant.numeric.clojure
+  -2r1010 -16r1234af -32r1234az -2R1010 -16R1234AF -32R1234AZ
+; ^^^^^^^ constant.numeric.clojure
+;        ^- constant
+;         ^^^^^^^^^^ constant.numeric.clojure
+;                    ^^^^^^^^^^ constant.numeric.clojure
+;                               ^^^^^^^ constant.numeric.clojure
+;                                       ^^^^^^^^^^ constant.numeric.clojure
+;                                                  ^^^^^^^^^^ constant.numeric.clojure
+  0/10 10/20 30/0
```
wbond: What does the `/` represent?

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -0,0 +1,2007 @@
...
+;               ^ punctuation.section.brackets.end.clojure
+;                ^ punctuation.section.parens.end.clojure
+  ([10/20 30/40])
+; ^ punctuation.section.parens.begin.clojure
+;  ^ punctuation.section.brackets.begin.clojure
+;   ^^^^^ constant.numeric.clojure
+;         ^^^^^ constant.numeric.clojure
+;              ^ punctuation.section.brackets.end.clojure
+;               ^ punctuation.section.parens.end.clojure
+  ([100.100 200.200])
```
wbond: Are these floats, or do the `.` represent something else?

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -0,0 +1,2007 @@
...
+;                               ^^ constant.language.keyword.clojure
+;                                  ^^ constant.language.keyword.clojure
+;                                     ^^ constant.language.keyword.clojure
+;                                        ^^ constant.language.keyword.clojure
+;                                           ^^ constant.language.keyword.clojure
+  :++ :--
+; ^^^ constant.language.keyword.clojure
+;    ^- constant
+;     ^^^ constant.language.keyword.clojure
+  :blah
```
wbond: Is `:blah` really a built in keyword?

Clojure/tests/syntax_test_clojure.clj
```diff
@@ -0,0 +1,2007 @@
...
+  blah:
+  blah::blah
+  /blah
+  blah/
+
+
+
+; # Keywords
+
+  :! :$ :% :& :* :- :_ := :+ :| :< :> :. :/ :?
```
wbond: Are these keywords, or more like operators?

Clojure/tests/syntax_test_clojure.clj
Mitranim: In Clojure/EDN, comments are parsed as whitespace. They disappear from the AST after parsing, so yeah, they're comments.

Clojure/tests/syntax_test_clojure.clj
Mitranim: `0x_______` is a hex integer literal, which may include `[A-Fa-f]`.

Clojure/tests/syntax_test_clojure.clj
Mitranim: In Clojure, `N/M`, where N and M are integers, is a ratio literal. Ratios are a special numeric data type.

Clojure/tests/syntax_test_clojure.clj
Mitranim: These are floats. Like many other tests, this one looks funny because it tests for funny mistakes I made when developing the syntax.

Clojure/tests/syntax_test_clojure.clj
Mitranim: Clojure doesn't have any built-in syntactic keywords as seen in other languages, since it's defined in terms of EDN data structures. The keyword data type is the same as Erlang or Ruby atoms. They're literals similar to strings, interned for performance and used as map keys.

Clojure/tests/syntax_test_clojure.clj
Mitranim: There are actual valid keywords. On a side note, Clojure doesn't have any operators; things like `+` and `=` are regular identifiers.

Clojure/tests/syntax_test_clojure.clj
Mitranim: Come to think of, this scope name is probably wrong. Reader macros are basically unary prefix operators that modify the next piece of AST as it's being read. I didn't want to mark them as regular operators since they work at parse time, not runtime, and it's nice if the color scheme can make a distinction, hence the current name. In many syntaxes, `constant.other.symbol` receives special highlighting. But we should probably change this.

Mitranim:
One issue that I see with the current implementation is that it's written in a way that makes it difficult to add meta scopes. Can address this if the need arises, but it just felt like a low priority.

Clojure/tests/syntax_test_clojure.clj
Mitranim: Ooh I see about storage. By that logic, the `N` at the end should be `storage.type.numeric.bigint.clojure`. I think it should still be part of the `constant.numeric` scope, since it's part of the number literal, so it needs a double scope.

FichteFoll: Yes, that's also how it is in Python (and other syntaxes). Also note the `punctuation.definition.numeric.hexadecimal.python` scopes.

Clojure/tests/syntax_test_clojure.clj
FichteFoll: How about adding a `punctuation.separator.ratio-literal.clojure` or similar scope to the slash?

Clojure/tests/syntax_test_clojure.clj
FichteFoll: There is a `keyword` scope. Why did you not choose a descendant of that?

Clojure/tests/syntax_test_clojure.clj
wbond: In that case they should be `constant.other`, not `constant.language`: https://www.sublimetext.com/docs/3/scope_naming.html#constant.

Clojure/tests/syntax_test_clojure.clj
wbond: So it would see to me from your description that the semantics of `:+` and `:blah` are the same, right? They should either be scoped the same, or if by convention they are used as "operators" we could scope them as such. I don't really see a reason they should be keywords, however.

FichteFoll: Currently they both have the `constant.language.keyword.clojure` scope.

I don't see any tests against a punctuation scope here, so scoping the colon as `punctuation.definition.keyword.clojure` sounds like a good addition.

Alternatively, if these are more similar to Ruby's symbol, maybe draw some inspiration for scope names from that?

Mitranim: AFAICT, Ruby symbols are exactly like Clojure keywords. Looks like we all agree to follow the example and use `constant.other.keyword.clojure` for the entire constant and `punctuation.definition.keyword.clojure` for the colon.

Clojure/tests/syntax_test_clojure.clj
Mitranim: Sure. `punctuation.separator.ratio.clojure` sounds good to me.

Clojure/tests/syntax_test_clojure.clj
Mitranim: Not sure if `keyword.macro.clojure` or `keyword.operator.macro.clojure` makes sense? Looks like these fall under `keyword` variety, and I want a scope that's different from `keyword.operator` but not too Clojure-specific. What would you suggest?

Mitranim:
Addressed the comments above.

Mitranim:
Pretty satisfied with how it works. Not strictly "done" (see the above comments on meta scopes), but haven't felt the need for more changes in the past few months. Might be stable enough to move forward with the PR for the sake of other Sublime/Clojure users, all 3 of them (?).

wbond:
I am going to hold off merging any significant work right now since the dev cycle is almost complete. We'll get this merged in at the beginning of the next dev cycle.

My only concern, is based on comments, it sounds as if there may be somewhat of a regression around function definitions. Is it just parameters that aren't highlighted, or the whole function definitions themselves? I think it is worth extra effort to support Goto Definition if this currently doesn't track the definitions of functions. Please excuse my lack of memory in regards to this PR.

Mitranim:
Yeah I just don't know the process, was wondering.

Symbol indexing works perfectly. I just banned inline lambdas from being indexed, since you never want to "goto" them anyway.

Clojure/Clojure.sublime-settings
```diff
@@ -0,0 +1,7 @@
+{
+  "word_separators": "/\\()[]{}\"'`,:;@#^~.",
+  "smart_indent": false,
+  "detect_indentation": false,
+  "tab_size": 1,
```
wbond: This seems funny to me. Why would the tab size be hardcoded to 1 for all Clojure files?

Clojure/Clojure.sublime-settings
```diff
@@ -0,0 +1,7 @@
...
+  "draw_indent_guides": false
```
wbond: Is this related to the tab size of 1?

Clojure/Default.sublime-keymap
```diff
@@ -0,0 +1,26 @@
+[
```
wbond: This file should use tabs for indentation

Clojure/Clojure.sublime-settings
Mitranim: Most Clojure code follows the same indentation convention, which sometimes involves 1-space indents. The [Lispindent](https://packagecontrol.io/packages/lispindent) package automates this style, but when refactoring, you often need to shift a block by 1 space, and it's needlessly painful if the indentation is not set to 1.

Example from the test file: [[1]](https://github.com/Mitranim/Packages/blob/c035c0a72b273069eba8dafb0962c8f0434e8fd9/Clojure/tests/syntax_test_clojure.clj#L2050-L2051).

Clojure/Clojure.sublime-settings
Mitranim: Yeah. There's a reason for these settings.

* `smart_indent` doesn't play well with Lisp code
* `detect_indentation` can get confused by traditional Clojure indentation, and there's very rarely any reason for indent other than 1
* `draw_indent_guides` creates needless noise with indent 1

Clojure/Default.sublime-keymap
Mitranim: Fixed.

Clojure/Clojure.sublime-settings
Mitranim: The most common reason for indent 1 is when writing nested data structures, such as Hiccup-style markup:

```clj
(def page []
  [:div.class0
   [:div.class1
    [:div.class2
     ...]]])
```

FichteFoll: If most code is formatted that way, the automatic detection would be able to catch this, wouldn't it? I'd be very vary of overriding indentation settings for all users, unless the format basically requires you to use it (kind of like YAML).

FichteFoll:
If most code is formatted that way, the automatic detection would be able to catch this, wouldn't it? I'd be very wary of overriding indentation settings for all users, unless the format basically requires you to use it (kind of like YAML).

However, if you believe in good faith that no sane user should ever use any other settings than the ones you provided (because they are very troublesome to work with or outright don't work at all), I certainly do see reasons for having them.

Mitranim:
Automatic indent detection is based on the first indented lines. Most Clojure files start with this:

```clj
(ns my-domain.my-namespace
  "Namespace docstring"
  (:require
    [namespace0]
    [namespace1])
  (:import
    [class0]
    [class1]))
```

Conventionally, the next line inside `()` is intended by 2 spaces, whereas the next line inside `[]` or `{}` is indented by 1 space. Automatic detection would always pick 2 spaces, which would be incorrect.

The main reason why I feel 1 space is more convenient is refactoring. Sometimes you need to shift a block of code by an uneven number of spaces due to the aforementioned conventions. If the block doesn't have any newlines in it, multi-cursor editing works fine. But if the block does have newlines and you're shifting to the left, you can't just multi-cursor: it would delete a newline and then delete something extra.

This is just my experience. If someone finds this default inconvenient, they're welcome to complain and we'll change it.

wbond:
I'm keen on getting this in so that it gets testing from other Clojure users during the next dev cycle. That said, I'm not convinced about the indentation settings tweak. Could you remove those and propose them in another PR, and hopefully we can get some feedback from other ST Clojure users in the future?

Mitranim:
Glad to hear this is getting progress.

No problem with removing the tweak, but are you sure going 2 → 1 is the right order? I assume that for most users, if 2 is inconvenient, they won't provide feedback saying 1 is better; nobody will even try 1 because it's so radical. Going 1 → 2, if folks find 1 inconvenient, they're somewhat more likely to complain, and we're more likely to receive feedback.

ItamarShDev:
I will be glad to help feedback-wise.
I use clojure daily at work :)

rwols:
@ItamarShDev have you tried this diff on your codebase?

PR #1261, opened 2017-10-24
[Groovy] Add word bounds to type in function definition

### LINKED ISSUE #1260: [Groovy] Function calls with names starting in "def" or a primitive type are mis-parsed
### Summary

The Groovy syntax highlighter has a rule which applies to functions defined with `def name(...)`. However, there's no word bound on `def` in this rule, so it also gets incorrectly applied to function calls whose names happen to begin with those three letters.

A similar rule, with similar bounding issues, also exists for primitive types (`int`, `double`, `char`, etc.) and `void`.

### Expected behavior

Function calls whose names start with `def` or a primitive type name will be highlighted correctly.

### Actual behavior

Function calls whose names start with `def` or a primitive type name are incorrectly highlighted as if they were function definitions, which in turn causes their arguments to be mis-parsed.

### Steps to reproduce

1. Create a Groovy file in Sublime Text with the following contents:

```groovy
void defectiveSyntaxDemo(thing) {
    println thing
}

defectiveSyntaxDemo("this highlighting is /*not*/ right")
interpretResults(new File("foo.bar"))
chartHistogram([1,1,2,3,5,8])
```

2. Note the very wrong highlighting in the last three lines: `def`, and the primitive type names that are substrings of the method names, are in a different color than the remainder of the method name. Even worse, the comment delimiters within the string argument are parsed!

### Environment

    * Windows 10 build 15063.674
    * macOS 10.13
* Sublime Text:
    * Build 3143

codeman38:
This should fix #1260.

### DIFF
--- a/Groovy/Groovy.sublime-syntax
+++ b/Groovy/Groovy.sublime-syntax
@@ -280,9 +280,9 @@ contexts:
             (?: # or modifier and optional type
               (?:(?:\b(public|private|protected)|(static)|(final)|(native|synchronized|abstract|threadsafe|transient))\b\s+)+\s* # modifier
               (?:\b
-                (void)
+                (void\b)
                 |
-                (boolean|byte|char|short|int|float|long|double) # primitive
+                ((?:boolean|byte|char|short|int|float|long|double)\b) # primitive
                 |
                 ( # or class type
                   (?:\w+\.)*[A-Z]\w+\b # Class name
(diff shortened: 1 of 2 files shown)


keith-hall:
Hi, it's great to have someone improving one of the seemingly lesser used syntaxes in this repo - welcome! :) Is there any chance you'd like to have a go at creating a syntax test file for Groovy, to help ensure your change doesn't get regressed in the future? :)

If so, docs are at http://www.sublimetext.com/docs/3/syntax.html#testing, and I highly recommend installing https://packagecontrol.io/packages/PackageDev to help with building the test assertions

codeman38:
Syntax test has been added.

PR #1262, opened 2017-10-26
[PHP] Fix the static accessor (::) is not scoped

jfcherng:
image

```php
<?php

$foo::BAZ;
$foo->bar::BAZ;
$foo->bar['x']::BAZ;
```

It look like GitHub's code highlight doesn't highlight it well either.

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -1005,6 +1005,11 @@ contexts:
         2: variable.other.member.php
         3: punctuation.definition.variable.php
       push: after-identifier
+    - match: '(::)({{identifier}})?'
+      captures:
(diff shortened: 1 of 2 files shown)


PHP/PHP Source.sublime-syntax
```diff
@@ -1005,6 +1005,11 @@ contexts:
         2: variable.other.member.php
         3: punctuation.definition.variable.php
       push: after-identifier
+    - match: '(::)({{identifier}})?'
+      captures:
+        1: punctuation.accessor.php
```
FichteFoll: Specify this to `punctuation.accessor.double-colon.php`?

wbond:
Thanks!

PR #1263, opened 2017-10-26
[Lua] Fix syntax highlighting for anonymous functions

Resike:

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -17,6 +17,13 @@ contexts:
         4: punctuation.definition.parameters.begin.lua
         5: variable.parameter.function.lua
         6: punctuation.definition.parameters.end.lua
+    - match: '\b(function)\s*(\()([^)]*)(\))'
+      scope: meta.function.lua
+      captures:
+        1: keyword.control.lua
+        2: punctuation.definition.parameters.begin.lua
+        3: variable.parameter.function.lua
+        4: punctuation.definition.parameters.end.lua
     - match: '(?<![\d.])\b0[xX][a-fA-F\d\.]+([pP][\-\+]?\d+)?|\b\d+(\.\d+)?([eE]-?\d+)?|\.\d+([eE]-?\d+)?'
       scope: constant.numeric.lua
     - match: "'"
--- a/Lua/syntax_test_lua
+++ b/Lua/syntax_test_lua
@@ -1,0 +1,14 @@
+-- SYNTAX TEST "Lua.sublime-syntax"
+
+local func = function(arg1, arg2)
+-- ^ source.lua
+        -- ^ source.lua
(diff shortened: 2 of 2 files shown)


Lua/syntax_test_lua
```diff
@@ -0,0 +1,14 @@
+-- SYNTAX TEST "Lua.sublime-syntax"
+
+local func = function(arg1, arg2)
```
rwols: You cleverly avoided the comma in-between the parentheses. The `,` also has the `variable.parameter` scope, which it shouldn't have.

Lua/Lua.sublime-syntax
```diff
@@ -17,6 +17,13 @@ contexts:
         4: punctuation.definition.parameters.begin.lua
         5: variable.parameter.function.lua
         6: punctuation.definition.parameters.end.lua
+    - match: '\b(function)\s*(\()([^)]*)(\))'
```
rwols: Would it be possible to match `\b(function)\s*(\()`, and then `push` to handle the rest?

Lua/syntax_test_lua
Resike: Well it currently works like this with every other function arguments by the default syntax. All the parameters and the , are colored the same way under the same scope.

Lua/Lua.sublime-syntax
Resike: What do you mean by push?

rwols: I took a closer look at your changes. Would changing the regex on line 13 to

```
'\b(function)\b\s*([a-zA-Z_.:]+[.:])?([a-zA-Z_]\w*)?\s*(\()([^)]*)(\))'
```

give us the same desired effect? I think we can drop these additions from line 20 to 26 then.

Lua/syntax_test_lua
rwols: Fair enough. You see, [the whole syntax needs a facelift](https://github.com/sublimehq/Packages/issues/481). So I'm trying to figure out what this PR means. Is it a patch for this old crufty syntax, or is it the beginning of a series of changes that will eventually bring Lua up to standards? The choice is yours ;)

Lua/Lua.sublime-syntax
Resike: I guess, however it would worth to fix the function names too then:

```'\b(function)\b\s*([a-zA-Z_]\w*[.:])?([a-zA-Z_]\w*)?\s*(\()([^)]*)(\))'```

Only one . or :

And Lua accepts numbers in function scopes too, after the first more strict character.

Lua/syntax_test_lua
Resike: Well that's a good question, i use a custom Lua syntax that i made, and i just wanted to fork some of the changes/fixes i made into the official one. The full rework probably needs a lot more work.

PR #1270, opened 2017-11-01
[Java] Better handling of numbers

lastsecondsave:
New regex for numbers. Old one was definitely C-inspired as Java doesn't have `U` modifier.

New one handles underscores in numbers (like `123_45_6`) and binary numbers (`0b00110101`).

### DIFF
--- a/Java/syntax_test_java.java
+++ b/Java/syntax_test_java.java
@@ -746,6 +746,81 @@ public class Foo {
   }
 //^ meta.method.java meta.method.body.java punctuation.section.block.end.java
 
+  int numbers() {
+    a = 0 + 0L;
+//      ^ constant.numeric
+//        ^ keyword.operator
+//          ^^ constant.numeric
+
+    a = 0xABCD + 0xAB_CD;
+//      ^^^^^^ constant.numeric
+//             ^ keyword.operator
+//               ^^^^^^ constant.numeric
+
+    a = 0xAB_CD_ - 0x_AB_CD - 0_xAB_CD;
+//      ^^^^^^^^ constant.numeric
+//                 ^^^^^^^^ -constant.numeric
+//                            ^^^^^^^^ -constant.numeric
+
+    a = 07 + 0_7;
+//      ^^ constant.numeric
+//         ^ keyword.operator
+//           ^^^ constant.numeric
+
+    a = 07_ - 09;
+//      ^^^ constant.numeric
+//            ^^ -constant.numeric
+
+    a = 0b101101 + 0b10_11_01;
+//      ^^^^^^^^ constant.numeric
+//               ^ keyword.operator
+//                 ^^^^^^^^^^ constant.numeric
+
+    a = 0b_101101;
+//      ^^^^^^^^^ -constant.numeric
+
+    a = 12345 + 12_34_5 + 1_____5;
+//      ^^^^^ constant.numeric
+//              ^^^^^^^ constant.numeric
+//                        ^^^^^^^ constant.numeric
+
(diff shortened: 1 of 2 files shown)


Java/syntax_test_java.java
```diff
@@ -746,6 +746,67 @@ int operators() {
...
+    a = 0 + 0L;
+//      ^ constant.numeric
+//          ^^ constant.numeric
+
+    a = 0xABCD + 0xAB_CD;
+//      ^^^^^^ constant.numeric
+//               ^^^^^^ constant.numeric
+
+    a = 0x_AB_CD - 0xAB_CD_;
+//      ^^^^^^^^ -constant.numeric
```
ismell: Instead of not applying constant.numeric can you mark the _ as invalid.illegal or mark the whole thing as invalid

Java/syntax_test_java.java
```diff
@@ -746,6 +746,67 @@ int operators() {
   }
 //^ meta.method.java meta.method.body.java punctuation.section.block.end.java
 
+  int numbers() {
+    a = 0 + 0L;
```
ismell: Can you check for the operator. Just to ensure the regex doesn't gobble it up.

Java/syntax_test_java.java
```diff
@@ -746,6 +746,67 @@ int operators() {
...
+//              ^^^^^^^ constant.numeric
+//                        ^^^^^^^ constant.numeric
+
+    a = 12345l + 12345L + 123_45d + 12_3245F
+//      ^^^^^^ constant.numeric
+//               ^^^^^^ constant.numeric
+//                        ^^^^^^^ constant.numeric
+//                                  ^^^^^^^^ constant.numeric
+
+    a = 12_34_5_ - _12_34_5 - 12_D - 12_L;
```
ismell: Can you add a test case where there is no space in between the operator.

lastsecondsave: Added.

Java/syntax_test_java.java
keith-hall: I would say go for a compromise - scope everything except the trailing underscore as a number, and only scope the trailing underscore as invalid if followed immediately by a semicolon - that should avoid flickering colors.

ismell:
Or just scope the trailing underscore as part of the number and let the
compiler deal with the error

On Sat, Nov 4, 2017, at 09:40 AM, Keith Hall wrote:
> *@keith-hall* commented on this pull request.


> 
> In Java/syntax_test_java.java[1]:


> > @@ -746,6 +746,67 @@ int operators() { } //^ meta.method.java
> > meta.method.body.java punctuation.section.block.end.java  +  int
> > numbers() { +    a = 0 + 0L; +//      ^ constant.numeric +//
> > ^^ constant.numeric + +    a = 0xABCD + 0xAB_CD; +//      ^^^^^^
> > constant.numeric +//               ^^^^^^ constant.numeric + +    a
> > = 0x_AB_CD - 0xAB_CD_; +//      ^^^^^^^^ -constant.numeric
>> I would say go for a compromise - scope everything except the trailing
> underscore as a number, and only scope the trailing underscore as
> invalid if followed immediately by a semicolon - that should avoid
> flickering colors.> — You are receiving this because you commented. Reply to this email
> directly, view it on GitHub[2], or mute the thread[3].> 

Links:

  1. https://github.com/sublimehq/Packages/pull/1270#discussion_r148933985
  2. https://github.com/sublimehq/Packages/pull/1270#discussion_r148933985

lastsecondsave:
I've tweaked regex to allow trailing underscores.

PR #1272, opened 2017-11-02
[Graphviz] use new embed action for embedded HTML and tidy up comments

keith-hall:
This PR updates the Graphviz syntax to use the new `embed`/`escape` action for embedded HTML and updates the `comments` context to use the preferred style of pushing into a context and pushing at the end of the line (instead of matching the comment start token and the rest of the line in one pattern).


wbond:
Thanks!

PR #1274, opened 2017-11-02
[Markdown] add syntax highlighting in code fences for a few languages

keith-hall:
This PR uses the new `embed` action to add syntax highlighting inside Markdown code fences for XML, SQL, Python and Graphviz.


Markdown/Markdown.sublime-syntax
```diff
@@ -855,6 +855,58 @@ contexts:
...
+        2: markup.raw.code-fence.sql.markdown constant.other.language-name.markdown
+      embed: scope:source.sql
+      embed_scope: markup.raw.code-fence.sql.markdown
+      escape: '^[ ]{,3}\1\s*$'
+      escape_captures:
+        0: markup.raw.code-fence.sql.markdown punctuation.definition.raw.code-fence.end.markdown
+    - match: |-
+         (?x)
+          {{fenced_code_block_start}}
+          ((?i:python))
```
FichteFoll: Also add the 'py' variant, which I use all the time
 😉

Markdown/Markdown.sublime-syntax
```diff
@@ -855,6 +855,58 @@ contexts:
...
+          ((?i:xml))
+          \s*$          # any whitespace until eol
+      captures:
+        1: markup.raw.code-fence.xml.markdown punctuation.definition.raw.code-fence.begin.markdown
+        2: markup.raw.code-fence.xml.markdown constant.other.language-name.markdown
+      embed: scope:text.xml
+      embed_scope: markup.raw.code-fence.xml.markdown
+      escape: '^[ ]{,3}\1\s*$'
+      escape_captures:
+        0: markup.raw.code-fence.xml.markdown punctuation.definition.raw.code-fence.end.markdown
```
FichteFoll: Only the fences should get the punctuation scope, not any whitespace surrounding it.

keith-hall:
Adjustments made - I deliberately didn't add too many languages to start with in case tweaks were needed, as unfortunately variables are not supported in the `escape` pattern, and I was lazy to make lots of syntax tests... About things like C, C++, Obj-C - I wonder if there is some canonical reference somewhere of what the "language identifier" in the info string should be?

Markdown/Markdown.sublime-syntax
```diff
@@ -856,66 +863,113 @@ contexts:
...
-      escape: '^[ ]{,3}\1\s*$'
+      escape: '^[ ]{,3}(\1)\s*$'
       escape_captures:
-        0: markup.raw.code-fence.graphviz.markdown punctuation.definition.raw.code-fence.end.markdown
+        0: markup.raw.code-fence.graphviz.markdown
+        1: punctuation.definition.raw.code-fence.end.markdown
+    - match: |-
+         (?x)
+          {{fenced_code_block_start}}
+          ((?i:javascript))
```
FichteFoll: `js` also

FichteFoll:
Following [github's GFM docs](https://help.github.com/articles/creating-and-highlighting-code-blocks/) I found this: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml. Looks like it's a case-insensitive match of the name and also supports `aliases`, although Python doesn't have a `py` alias 🤔

wbond:
I would expect shell is probably one of the most popular fenced syntaxes. According Tiobe, things like C, PHP, Ruby, Obj-C, and Go would probably be higher on the priority list than Rust.

randy3k:
I vote for the R syntax.

keith-hall:
I vote for `embed` to be changed to work more like `include` in that missing syntaxes are just warnings, so I can keep languages I will never use in my `ignored_packages` list without a load of errors appearing and the syntax highlighting not working at all, that's the main thing stopping me from adding more languages atm, apart from the extra syntax test and maintenance headache

wbond:
@keith-hall That is an interesting idea. Make embedding sort of optional. The escape would still work, but if the embedded syntax isn't present, just print a warning saying it couldn't be found.

keith-hall:
It would then potentially be possible (though I'm not sure if we want to) to add code fence support for some popular third party syntaxes too :)

wbond:
Well, if it printed a warning when they were missing, that would cause the CI tests to fail

keith-hall:
Added C, C++, Go, ObjC, ObjC++, R, Ruby

michaelblyons:
Crosslinking to #266, where code fences and strikethrough are the last two issues.

rwols:
This won't work correctly until https://github.com/SublimeTextIssues/Core/issues/1062 is fixed. For example, comments, line-continuations and character escapes are not rendered for Bash.

wbond:
I spent a bit of time on https://github.com/SublimeTextIssues/Core/issues/1062 before 3157 went out, but wasn't able to come up with an implementation that didn't break existing default syntaxes in funny ways.

I think this is largely due to implementation details, but also this would end up being a change in behavior, so I may just end up documenting the edge case.

Markdown/Markdown.sublime-syntax
```diff
@@ -898,18 +915,285 @@ contexts:
         - match: $
           pop: true
   fenced-code-block:
+    - match: |-
+         (?x)
+          {{fenced_code_block_start}}
+          ((?i:xml))
+          \s*$          # any whitespace until EOL
+      captures:
+        2: markup.raw.code-fence.xml.markdown-gfm punctuation.definition.raw.code-fence.begin.markdown
```
FichteFoll: I don't think these captures should be getting the `markup.raw` scope. They aren't part of the "raw" segment since they are not rendered verbatim. (Also, I have dimmed background for `markup.raw` and it would be nice if that started on the first line of the raw block.)

Markdown/Markdown.sublime-syntax
```diff
@@ -898,18 +915,285 @@ contexts:
...
           ([\w-]*)     # any number of word characters or dashes
-          .*$          # all characters until eol
+          .*$          # all characters until EOL
       captures:
-        1: punctuation.definition.raw.code-fence.begin.markdown
-        2: constant.other.language-name.markdown
+        2: punctuation.definition.raw.code-fence.begin.markdown
+        3: constant.other.language-name.markdown
       push:
         - meta_scope: markup.raw.code-fence.markdown-gfm
```
FichteFoll: For the same reason, imo this should be `meta_content_scope`. And if it would otherwise match on the newline, the match probably needs to include `\n` also, but I'm not sure about that.

Markdown/Markdown.sublime-syntax
FichteFoll: And the `escape_captures` for `0` should be removed.

deathaxe:
How about code fence highlighting for JSON as a subset of JavaScript?

keith-hall:
@rwols can you give some examples of some Bash code that doesn't highlight correctly in the Markdown code fence please - I've added some test assertions for comments and character escapes and it seems to be working in build 3161, but maybe I'm missing something?

rwols:
I checked out the `STealthy-and-haSTy:markdown_codefences` locally and it indeed seems to work fine. So I take back what I said :)

keith-hall:
maybe let's get this merged then :smiley: :fireworks:

randy3k:
@keith-hall 
I just noticed that this PR broke the regex compatibility because of the use of backreference.
https://github.com/sublimehq/Packages/pull/1274/files#diff-bc7dff9190deda9f015e7ef8a4f3ceebR159

keith-hall:
It is indeed only used from `pop` and `escape`, but ST is indeed marking it as incompatible for some reason. It's referring directly to the variable so I'd guess its a bug / overzealous warning in ST.

> Packages/Markdown/Markdown.sublime-syntax:123:11: Back references are unsupported
FAILED: 1 pattern in "Packages/Markdown/Markdown.sublime-syntax" are incompatible with the new regex engine
[Finished]

(at the current state of the `master` branch, this is https://github.com/sublimehq/Packages/blob/289782ff2e4cb58de171579c7fc86fe00d280619/Markdown/Markdown.sublime-syntax#L123)

randy3k:
[R Markdown](https://github.com/REditorSupport/sublime-ide-r) is based on the built in Markdown syntax and I recently received an [issue](https://github.com/REditorSupport/sublime-ide-r/issues/6) and that seems to be related to this bug. Particular, that is resolved by reverting this PR.

PR #1275, opened 2017-11-02
[Java] Declaration statement for try-with-resources

lastsecondsave:
This code is perfectly legal:

screen shot 2017-11-02 at 14 11 53

But current syntax finds invalids in it. Also the first `final` is not highlighted.


Java/Java.sublime-syntax
```diff
@@ -247,6 +247,7 @@ contexts:
     - include: static-code-block
   code:
     - include: constants-and-special-vars
+    - include: final-modifier
```
ismell: Instead of adding the final-modifier here, can you add it on line 583?

Java/Java.sublime-syntax
```diff
@@ -568,8 +572,7 @@ contexts:
           pop: true
         - include: illegal-parens-terminators
         - include: code
-  for-loop-parens:
-    - meta_scope: for-loop
+  declaration-statement-parens:
     - match: \(
       scope: punctuation.section.parens.begin.java
       set:
```
ismell: This is really a comment for line 584 but I can't seem to add one there.

You could change `include: code` to `include: code-block` and avoid declaring a final block.

lastsecondsave: Yes, this seems to be a more correct solution. Fixed.

Issue #1114, opened 2017-08-21
Need to add double quote " to legal matlab String syntax


apassy:
I tried, but can't get it to work.  It's legal as of R2017a.

FichteFoll:
How does this issue surface and how can we reproduce it? Please explain it for someone who is not experienced with Matlab, since I'm not sure one of the regular contributors is.

apassy:
Matlab used to have one syntax for strings 
a='lksjdfl' which would create one kind of string.
" was illegal, and the syntax definitions flag that as illegal.
as of r2017a (Feb of this year) " is legal.
b = "new string" is a new type of string, so needs to be flagged as a string.

TheClams:
I think this can be closed now, it is part of the latest ST3 dev

PR #1276, opened 2017-11-02
[Matlab] Regex Compatibility & minor fixes

TheClams:
- Remove all lookbehind regex => now pass the compatibility test
 - Properly scope the function keyword and args:
  * function was storage.type instead of keyword
  * Input and output arguments now scoped as variable.parameter
 - Add support for double quote string (fix issue #1114)
 - Fix the block comment not inserting  new line
 - Add missing scope for ~ operator
 - Add some more test to ensure nothing was broken with the removal of lookbehind regex


wbond:
Awesome, thanks!

PR #1277, opened 2017-11-03
[Python] Fix New-Property snippet not being suggested

### LINKED ISSUE #1264: [Python] Property snippet is not working
**To reproduce:**

1. Open a new file
2. Set python syntax
3. Create a class
4. Start to add a property

**What should happen:**

The completion menu should appear including the property snippet.

**What actually happens:**

No completion menu appears at all.

**Other notes:**

- Running the latest development build on macOS 10.13.
- There might be a scope mismatch: the snippet specifies meta.function.decorator which doesn't show up in syntax scope test file.

deathaxe:
fixes #1264

Due to scope changes the "New-Property" snippet was no longer available to end users.

Therefore the snippet scope needs to be changed to `meta.annotation`, too.

### DIFF
--- a/Python/Snippets/New-Property.sublime-snippet
+++ b/Python/Snippets/New-Property.sublime-snippet
@@ -4,6 +4,6 @@ def ${1:foo}(self):
 	return self.${2:_$1}
 $0]]></content>
 	<tabTrigger>property</tabTrigger>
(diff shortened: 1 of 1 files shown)


wbond:
Thanks!

PR #1285, opened 2017-11-07
[PHP] Fix PHP static method call scope

jfcherng:
```php
echo $this->pro1::bar();
//                ^^^ variable.function
```


wbond:
Thanks!

PR #1289, opened 2017-11-08
[Lua] Fix auto-indenting for functions

p3lim:
image

In the first and the third case in that image, the content of the function is auto-indented, but not for the second variant (which is commonly used for recursion).

This commit fixes that.


p3lim:
This fix doesn't include support for non-ascii characters though, which are valid in variable names.

Thom1729:
The `∆` character in your comment is U+2006 "Increment", from the Mathematical Operators block. Unlike the visually similar `Δ` U+0394 "Greek Capital Letter Delta", the increment operator is not a word character.

wbond:
This ended up being handled with some other fixes in #1462, as far as I understand. I apologize this sat for so long!

Issue #1269, opened 2017-10-31
[JavaScript] ECMAScript Omnibus


Thom1729:
I'm creating this issue to catalog missing features in the JavaScript bundle and discuss their implementation. Once #1009 is merged it should be easy to rip through most of these in short order.

Some of these features are still moving through the standardization process and may change or even be cancelled. Implementing them exposes us to the risk of a breaking change. But countless developers already use experimental features via transpilers like Babel, and failing to support them would mean leaving them out in the cold.

In my opinion, we should add an experimental feature when it:

- Is on the TC39 standards track.
- Is available for practical use by developers.
- Has a stable *syntax* (even if its semantics are likely to change).

These criteria are necessarily fuzzy. A stage-3 proposal is a much safer bet than a stage-0 proposal. "Available for practical use" will usually mean "implemented in Babel". Stability might best be determined by taking the sense of the TC39 meeting notes.

If we do implement an experimental feature, and that feature is cancelled or substantially altered, then we must take responsibility for updating or removing our implementation of that feature. An "expired" experimental feature is nothing but cruft.

---

The following features are not implemented as of May 8, 2018. Examples are largely taken from [Babel](https://github.com/babel/proposals) or from the proposal spec.

### `function.sent`

```js
function* generator() {
    console.log("Sent", function.sent);
    console.log("Yield", yield);
}
```

Status: Stage 2
Importance: Moderate (Babel implementation)
Risk: Low.

### Optional Chaining

```js
a?.b = 42;
```

Status: Stage 1
Importance: Moderate (Babel implementation)
Risk: High.

---

### Class Fields (implemented in #1292)

```js
class Bork {
  static a = 'foo'; // static property
  x = 'bar'; // public instance property
  #p = 'baz'; // private instance property
}
```

### Numeric Separators and BigInt (implemented in #1294)

```js
1_000_000_000
0b1010_0001_1000_0101
0xA0_B0_C0
50000n + 60n;
```

### Asynchronous Iteration (implemented in #1341)

```js
for await (const line of readLines(filePath)) {
  console.log(line);
}
```

### RegExp Improvements (implemented in #1416)

```js
/(?<=foo)$/; // Lookbehind
/\p{Script=Greek}/u; // Unicode property escapes
/(?<year>\d{4})-(?<month>\d{2})-(?<day>\d{2})/u; // Named capture groups
/./s; // dotAll flag
```

### Function parameter destructuring (implemented in #1423)

```js
[...]

PR #1292, opened 2017-11-08
[JavaScript] Add support for class fields

### LINKED ISSUE #876: [JavaScript] Missing ES7 features
I don’t know if you would prefer these to be in separate tickets, but here are a couple things missing in the native JS syntax (that I am aware of).

- Async methods inside of a class

    ```javascript
    class SomeClass {
        async instanceMethod() {
        }

        static async someStaticMethod() {
        }
    }
    ```
- Class static and instance properties (taken from [Babel site](https://babeljs.io/docs/plugins/transform-class-properties/))

    ```javascript
    class Bork {
        //Property initializer syntax
        instanceProperty = 'bork';
        boundFunction = () => {
            return this.instanceProperty;
        }

        //Static class properties
        static staticProperty = 'babelIsCool';
        static staticFunction = function() {
            return Bork.staticProperty;
        }
    }
    ```

I am not sure what scopes they should be. The `async` keyword highlights as `entity.name.function` right now. The static and instance properties currently cause the highlighting to break:

kqyb

### LINKED ISSUE #1238: JavaScript: unsupported class properties
Support class properties.

```javascript
class Foo {
  static fooProp = {};

  // scope here no longer has meta.class.js mets.block.js
  fooMethod() {
  }
}
```

Thom1729:
Implements the [class fields proposal](https://github.com/tc39/proposal-class-fields). See #1269 for more discussion. Also resolves #1238 and partially resolves #876.

Outline of changes:

- No longer recognize the `static` keyword in object literals. It should be interpreted as an identifier.
- In classes, rescope `static` as `storage.modifier`. It was previously scoped as `storage.type`.
- Recognize class property declarations and initializers in class bodies.
- Recognize private class properties in expressions.

Relevant to https://github.com/tc39/proposal-class-fields/issues/57.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -785,6 +785,23 @@ contexts:
     - match: '\}'
       scope: punctuation.section.block.js
       pop: true
+
+    - match: \bconstructor\b
+      scope: entity.name.function.constructor.js
+      push:
+        - function-declaration-expect-body
+        - function-declaration-meta
+        - function-declaration-expect-parameters
+
+    - match: \bstatic\b
+      scope: storage.modifier.js
+      push:
+        - include: class-field
+        - include: else-pop
+
+    - include: class-field
+
+  class-field:
     - include: method-declaration
 
   constants:
@@ -1172,7 +1189,7 @@ contexts:
   method-declaration:
     - match: |-
         (?x)(?=
-          \b(?: get|set|async|static )\b
+          \b(?: get|set|async )\b
           | \*
           | {{method_name}} \s* \(
         )
@@ -1188,8 +1205,6 @@ contexts:
       scope: keyword.generator.asterisk.js
     - match: \b(get|set)\b(?!\s*\()
       scope: storage.type.accessor.js
-    - match: \bstatic\b
-      scope: storage.type.js
     - include: else-pop
 
   parenthesized-expression:
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -806,25 +806,46 @@ contexts:
 
     - match: \bstatic\b
       scope: storage.modifier.js
-      push:
-        - include: class-field
-        - include: else-pop
-
-    - include: class-field
-
-  class-field:
-    - include: method-declaration
+      push: class-field
 
     - match: (?={{property_name}})
+      push: class-field
+
+  class-field:
+    - match: '{{method_lookahead}}'
+      set: method-declaration
+      
+    - match: (?={{property_name}})
+      set:
+        - field-initializer-or-method-declaration
+        - field-name
+
+    - include: else-pop
+
+  class-field-rest:
+    - match: ','
+      scope: punctuation.separator.js
       push:
         - field-initializer
         - field-name
+    - include: else-pop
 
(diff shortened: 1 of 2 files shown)


wbond:
I very much appreciate all of the work you've been doing on the JavaScript syntax recently. I know lots of users are benefitting from your hard work.

I think that #1269 outlines that this currently has no implementations. I feel like once Babel or a browsers supports it, then I'd be more keen on merging it. Mostly because every feature we add increases the complexity, and having things in heavy flux implemented here are just going to make things more complicated in the future. Contributors will have to dig through history to understand why the implementation is as it is.

Perhaps we can merge in some of the foundational stuff in now and leave the proposals in a PR for once there is someone implementing them?

Thom1729:
Public class fields are implemented in Babel and are widely used. Private fields are implemented in Babel's parser, but the transform is not yet finished.

In addition, [the entire spec (including private fields) is at Stage 3](https://github.com/tc39/proposal-class-fields), which means that [it should be very stable](https://tc39.github.io/process-document/). It probably won't move to Stage 4 until after it's actually shipping in major browsers. (Stage 4 requires "Two compatible implementations which pass the acceptance tests", although one of them may be a transpiler like Babel.) Expected changes are "Limited: only those deemed critical based on implementation experience". In the unlikely event that there are any changes to the proposal, it is even more unlikely that the syntax would be affected; late-stage changes tend to be esoteric tweaks to obscure semantics.

On the other hand, in my experience there are a lot of developers actively using class fields, including my own company. Without explicit support in the `sublime-syntax`, highlighting of JavaScript with class fields fails quite ungracefully. This severely impacts the usability of the JavaScript syntax when using class fields. Because class field declarations may contain arbitrary expressions, I don't believe that there is any way of ignoring them without fully and properly parsing them, at which point we may as well give them the appropriate scopes as well.

If we want to be maximally conservative, what about commenting out private field support until Babel supports it?

Thom1729:
Just to add on to this: I would suggest that a conservative standard might be that we should implement "experimental" features when:

- The feature is at Stage 3 status; OR
- The feature is at Stage 2, AND there is a working implementation, AND the syntax risk is Low.

Stage 3 is the gold standard indicating that a feature is ready for implementation. It must have a complete formal specification and sign-offs from various reviewers and editors. In addition, "The solution is complete and no further work is possible without implementation experience, significant usage and external feedback." This is the point at which most major browser vendors will begin shipping the feature, and indeed it cannot advance to Stage 4 without "[t]wo compatible implementations which pass the acceptance tests" and "[s]ignificant in-the-field experience with shipping implementations".

Stage 2 means that a feature is largely complete, but not totally finalized. It is required that "all major semantics, syntax and API are covered", and "[t]he committee expects the feature to be developed and eventually included in the standard". It is expected that features in Stage 2 will be implemented by transpilers and other tools and often by major browsers (usually behind dev flags). Having a working implementation means that actual developers will be using the feature in production code. Although it is possible for a feature's syntax to change at this stage, it is very unusual. The standards committee actively tries to avoid making changes that would break Babel, and from reading their meeting notes you can easily get a sense of whether there is any significant syntax risk.

Thom1729:
@wbond Any thoughts on the standard described above?

wbond:
@Thom1729 I think those standards for inclusion sound reasonable.

PR #1296, opened 2017-11-09
[PHP] Fix scopes for scientific number notation

jfcherng:
```php
$var2 = -123e10;
//       ^^^^^^ constant.numeric.float.decimal
```


wbond:
Thanks for improving this!

PR #1299, opened 2017-11-10
[LaTeX] Correct scope for \citeyearpar cite command

jlegewie:
The LaTeX syntax does not assign the correct scope to the cite function `\citeyearpar`.` \citeyearpar` has the scope `support.function.general.latex` but should be `support.function.cite.latex` (like all other cite function).


wbond:
Thanks!

PR #1303, opened 2017-11-11
[Python] Fix `from . import` relative imports

### LINKED ISSUE #1300: [Python] from . import does not highlight import keyword
The commit 5ad641c4298d828648d36f317db55362a4139a39 somehow broke highlighting of relative imports from current package like:

```python
from . import module
```

The import keyword is scoped as module path instead of keyword.

FichteFoll:
Also highlight invalid names in the `from` section.

Fixes #1300.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -160,11 +160,11 @@ contexts:
     - match: (?={{path}})
       push:
         - meta_scope: meta.import-path.python
-        - match: '{{illegal_names}}'
+        - match: '{{illegal_names}}\b'
           scope: invalid.illegal.name.python
         - match: '{{identifier}}'
           scope: meta.import-name.python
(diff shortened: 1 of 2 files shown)


deathaxe:
The primary issue is fixed by this commit 👍 

Question about illegal scoping:

While in

```python
from .module.else import utils
```

the else is scoped as illegal, it is not in the following statement.

```python
from . else import utils
```

I like the idea of catching such common cases and would ask for the second statement to scope `else` (and all the other keywords) as illegal, too.

deathaxe:
Works fine.

deathaxe:
@FichteFoll

Hmm, just found this:

screenshot

Guess there is a `\b` missing somewhere?

FichteFoll:
Nice catch. Fixed.

wbond:
Thanks!

Issue #1305, opened 2017-11-13
[HTML] Whitespace breaks tags and attributes.


Thom1729:
```
<div title
    ="foo"></div>
//  ^ punctuation.separator.key-value
//   ^^^^^ string

<
    div></div>
//  ^^^^ meta.tag

<script
    type="text/notjavascript">
        
    42
//  ^^ - source.js
</script>
```

I ran into these bugs while trying to write an extended HTML syntax for Vue.js (https://github.com/vuejs/vue-syntax-highlight/issues/108). Fixing them will make that task much easier.

The solution to these is probably to use the stack more. The old:

```yaml
tag-generic-attribute:
  - match: '(?:^|\s+)(([a-zA-Z0-9:_.-]+)\s*(=)\s*)'
    captures:
      1: meta.attribute-with-value.html
      2: entity.other.attribute-name.html
      3: punctuation.separator.key-value.html
    push: # <snip>
  - match: '([a-zA-Z0-9:\-_.]+)'
    captures:
      1: entity.other.attribute-name.html
```

The new:

```yaml
tag-generic-attribute:
  - match: '\b[a-zA-Z0-9:_.-]+\b'
    scope: entity.other.attribute-name.html
    push:
      - meta-attribute
      - expect-attribute-equals

meta-attribute:
  - meta_scope: meta.attribute.html
  - match: ''
    pop: true

expect-attribute-equals:
  - match: '='
    scope: punctuation.separator.key-value.html
    set: expect-attribute-value
  - match: (?=\s*(?!=))
    pop: true

expect-attribute-value: # <snip>
```

Thom1729:
#1379 will mostly wrap this up.

PR #1306, opened 2017-11-14
[HTML] Fixed most problems with whitespace in attributes.

Thom1729:
For #1305. Fixes the below cases.

```html
<div
title = "description"></div>
## <- entity.other.attribute-name
##    ^ punctuation.separator.key-value.html
##      ^^^^^^^^^^^^^ string

<div title=
    "description"></div>
##  ^^^^^^^^^^^^^ string

<div title
    ="description"></div>
##  ^ punctuation.separator.key-value
##   ^^^^^^^^^^^^^ string
```

This is a bulkier change than I'd like because it wasn't possible to reuse most of the contexts due to minor scope differences. I'm not convinced of the utility of scopes like `meta.attribute-with-value.event.html`, but I left them in.

All attributes will now get the `meta.attribute-with-value` scope. In general, it is not possible to tell whether an attribute has a value without looking over line boundaries. If we really want to maintain the distinction for some reason, it should be possible to make some good guesses at the expense of considerable complexity. I'm generally skeptical about this sort of thing. If we don't want to maintain the distinction, we should change the scope name to `meta.attribute`.

There is still a fair bit of cleanup to do in this syntax. The biggest remaining challenge is fixing `<script>` tag language detection. Right now, all manner of things can screw it up, including whitespace. In addition to this, a lot of the scope names need to be brought in line with more modern practices (e.g. `meta.toc-list.id.html`).

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -282,7 +282,7 @@ contexts:
     - include: immediately-pop
 
   tag-generic-attribute-equals:
-    - match: =
+    - match: '='
       scope: punctuation.separator.key-value.html
       set: tag-generic-attribute-value
     - match: '{{not_equals_lookahead}}'
@@ -321,7 +321,7 @@ contexts:
     - include: immediately-pop
 
   tag-class-attribute-equals:
-    - match: =
+    - match: '='
       scope: punctuation.separator.key-value.html
       set: tag-class-attribute-value
     - match: '{{not_equals_lookahead}}'
@@ -362,7 +362,7 @@ contexts:
     - include: immediately-pop
 
   tag-id-attribute-equals:
-    - match: =
+    - match: '='
       scope: punctuation.separator.key-value.html
       set: tag-id-attribute-value
     - match: '{{not_equals_lookahead}}'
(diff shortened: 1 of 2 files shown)


keith-hall:
How much does this change affect performance, out of interest?

I like the event scopes, I can give them a different background color to make them visually distinctive.

Thom1729:
About 10-15% slower to parse this page's source. Well under 20 ms overall. Presumably a result of the additional stack usage, which is unavoidable when fixing the bugs. If we care, it might be possible to shave a few milliseconds by eliminating the separate meta contexts.

If all attribute values containing JavaScript code shared a scope, would that satisfy your highlighting preferences?

keith-hall:
I don't think a change of less than 20 ms will matter :+1:

I believe that would indeed satisfy my highlighting preferences :)

wbond:
With something as ubiquitous as the HTML syntax, lots of users and third-party packages have built upon it. Because of that, we need to be even more cautious of changes to scope names. I think the right move is leaving `meta.attribute-with-value` and `meta.toc-list`.

I'm not crazy about all of the extra contexts and complexity, but I think it is worth it if we can provide better extensibility for other syntaxes to build upon. I'll merge this in and see if we haven't many issues in the dev channel.

wbond:
Unfortunate this seems to have messed up the ASP syntax. I'll see if I can get that sorted.

With the next build I should be able to get the syntax test runner working for PRs, so we won't run into this in the future.

PR #1311, opened 2017-11-16
[Rails] Add transactional callbacks highlight

louim:
Added the following callbacks:
- after_commit
- after_create_commit
- after_destroy_commit
- after_rollback
- after_update_commit

[See the doc here](http://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html).

The diff might be a little hard to visualize, I added:
```
after_rollback|after_(create_|destroy_|update_)?commit|
```


wbond:
This is great, thanks!

If you are up for it, would you mind adding a commit to reformat the regexes into extended form (where whitespace doesn't matter) so we can read them a little easier?

If you are up for even more, adding a couple of syntax tests for these new callbacks would be a really nice addition. The current tests are at https://github.com/sublimehq/Packages/blob/master/Rails/syntax_test_rails.rb and you can read more about the syntax test format at http://www.sublimetext.com/docs/3/syntax.html#testing.

louim:
Hey thanks for merging this, and sorry for not doing the reformat. I must have missed the notification back then 😅.

PR #1314, opened 2017-11-18
[CSS] Add a hint to each completion item

deathaxe:
Most auto-completion packages provide basic information about their suggestions as hints right next to each item, which distinguishes auto-completion items very clearly from word-completions.

CSS disables word-completions, but it is not obvious to the end user.

Therefore this commit adds a hint to each returned completion item to allow the end user to clearly see it is a valid value from a database but not something ST might have caught from anywhere in the text.

### DIFF
--- a/CSS/css_completions.py
+++ b/CSS/css_completions.py
@@ -490,7 +490,7 @@ class CSSCompletions(sublime_plugin.EventListener):
                     add_semi_colon = view.substr(sublime.Region(loc, loc + 1)) != ';'
 
                     for value in values:
-                        desc = value
+                        desc = value + "\t" + prop_name
                         snippet = value
 
                         if add_semi_colon:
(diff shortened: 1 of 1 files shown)


CSS/css_completions.py
```diff
@@ -509,8 +509,8 @@ def on_query_completions(self, view, prefix, locations):
 
             for prop in self.props:
                 if add_colon:
-                    l.append((prop, prop + ": "))
+                    l.append((prop + "\tcss", prop + ": "))
```
FichteFoll: Shouldn't this rather get `"\tproperty"`?

deathaxe: May other plugins provide completions within a CSS scope if the CSS is embedded into another syntax, which may or may not be meant for CSS but are displayed anyway and therefore might cause confusion?

With CSS in mind as a standalone syntax "property" was my first choice, too.

FichteFoll: I'm inclined to blame those other plugins for that. I don't use any third-party package for standard css or HTML, so I'm not running into any of this, but that doesn't mean we shouldn't consider third-party packages at all.

wbond:
This seems like a nice enhancement, thanks!

PR #1315, opened 2017-11-18
[CSS] Avoid calling find

deathaxe:
The `find` seems a bit more expensive than using `x in list`. After this commit collecting completions for "padding" values was reduced from 70ms to 50ms on each first run. The time savings vary very much but seem smaller in most cases.


wbond:
Thanks!

PR #1316, opened 2017-11-18
[Java] 'invalid.illegal' for strings without closing quote

lastsecondsave:
Other syntax files already have this, why not add it to Java.

Match rules are copied from js syntax.


wbond:
Awesome, thank you!

PR #1317, opened 2017-11-19
[CSS] Fix @counter-style not scoped correctly if name is missing

deathaxe:
Prior to this commit the globally valid keyword @counter-style was highlighted as `entity.name.tag.custom` if no style-name is provided between it and the next `{`.

Even though the style-name is required the keyword must/should be scoped correctly if it is missing in order to avoid confusion.

Therefore this commit marks the style-name as optional to fix scoping of incomplete counter-style rules.


wbond:
Yes, I think this sort of tweak is the right call. Thanks!

PR #1318, opened 2017-11-20
[Java] Annotation-related fixes

lastsecondsave:
1. If a default value in annotation's definition is a string it is not highlighted:

2. Annotation's definition only works after an access modifier:

3. Annotations are not highlighted when used as parameters to other annotations:


wbond:
Thanks for all of your work on the Java syntax recently @lastsecondsave!

Issue #1228, opened 2017-10-15
[RFC] Add new scopes to "jumping" flow control keywords/statements.


FHTheron:
## Intro
First let me define what I mean with "jump" keywords. Things like `if` and `else` are of course also flow control keywords and they do result in the CPU executing branches, but I'm more specifically considering the jump in terms of visual code flow. For example, a subtle `goto` that radically changes the execution flow isn't always as visible as the effect it has.

## Current scopes
In `C` these currently get the scope `keyword.control.c`, but this scope is also used (correctly) for `if`, `while`, and other actual flow control keywords.

In `Python` the scope is `keyword.control.flow.python` and includes loop keywords as well as `return` and `raise`.

## New scopes
I suggest that some of these flow control keywords get scoped as two new flow control types - one type jumps within the current frame/call stack the other jumps out of the current call stack.

### `C`
* Jump "around"
    * `goto`
    * `break`
    * `continue`
* Jump "out"
    * `return`
    * Some standard functions that also result in (possibly) dramatic call stack jumps and these should be scoped the same as the keywords.
        * `exit()`
        * `assert()`

### `C++`
* Jump "around"
    * Same as `C`
* Jump "out"
    * Based on `C`
    * `throw`

### `Python`
* Jump "around"
    * `break`
    * `continue`
    * **Not** `pass`.
* Jump "out"
    * `return`  (already scoped with `keyword.control.flow.return.python`)
    * `raise`  (already scoped with `keyword.control.flow.raise.python`)
    * `assert`
    * Perhaps `yield` ?

And many more languages with similar keywords/statements.

## Alternative
The same colouring can be achieved if all languages define unique scopes for all flow control keywords. This will avoid having to define what each keyword's jump type is. `Python` has this for 2 keywords, `Shell-Unix-Generic` seems to have `keyword.control.*.shell` for all keywords. A colour scheme can then add each of these to colour them as desired. It does mean that the colour scheme now defines what jump type a jump keyword is, which may or may not matter.

## Example screenshots

Current:
![current](https://user-images.githubusercontent.com/16208294/31589922-4f4ca894-b209-11e7-8054-5ad6ed4559eb.png)

Proposed, using new scopes with illustrative colours:
![proposed](https://user-images.githubusercontent.com/16208294/31589925-523de856-b209-11e7-8aa4-fed2b2b950a6.png)

FichteFoll:
I like the rationale. I also used to do something similar when I had to use eclipse for some Java code.

[...]

PR #1319, opened 2017-11-20
[Java] Fix 'throws' keyword has a wrong scope

lastsecondsave:
It was `storage.modifier` for some reason.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -586,7 +586,7 @@ contexts:
     - include: invalid-generic-tags
   throws:
     - match: \bthrows\b
-      scope: storage.modifier.java
+      scope: keyword.declaration.throws.java
       push:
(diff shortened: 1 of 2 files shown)


Java/Java.sublime-syntax
```diff
@@ -586,7 +586,7 @@ contexts:
     - include: invalid-generic-tags
   throws:
     - match: \bthrows\b
-      scope: storage.modifier.java
+      scope: keyword.declaration.throws.java
```
FichteFoll: Pretty sure `keyword.control.flow.throws` makes more sense here.

`keyword.declaration` is only used for `extends` basically: https://github.com/sublimehq/Packages/search?utf8=%E2%9C%93&q=%22keyword.declaration%22&type=.

See also https://github.com/sublimehq/Packages/issues/1228.

wbond:
Thanks!

PR #1323, opened 2017-11-21
[JavaScript] Support `export function|class` form

### LINKED ISSUE #1322: [JavaScript] `export class Foo {}` not highlighting
In my javascript files I often use the ES6 export syntax `export class Foo { ... }`. I'm pretty sure that this used to be highlighted correctly, but as of build 3154 the `class` keyword has scope `variable.other.readwrite` instead of `storage.type.class`, and the subsequent class definition doesn't have the proper scope either. This messes up "go to symbol", since neither the class name nor any of its methods are treated as symbols.

Digging into the sublime syntax file, it looks like the problem is in the `export-extended` definition. If I comment out `push: ... - export-extended` within `keywords-top-level` the problem goes away.

Thom1729:
You should now be able to directly export function and class declarations. Added tests to verify.

Fixes #1322.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -175,16 +175,23 @@ contexts:
   export-extended:
     - include: variable-declaration
 
+    - match: (?=\bclass\b)
+      set: class
+
+    - match: (?={{func_lookahead}})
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #1330, opened 2017-11-28
[Batch File] fix set quoted string bug

### LINKED ISSUE #1329: [Batch] Open parentheses not getting matched to their closing counterpart in strings
Following the Sep. 13, 2017 3.0 release of Sublime Text 3 (build 3143), I noticed that parentheses pairs were no longer being matched inside of quoted-strings, resulting in a red error-bar (top image is build 3143, bottom is the release immediately before, 3126)

download - sublime text_2017-11-27 000047

The language definition appears to be incorrectly interpreting the closing quote as a new opening quote. Following the closing quote with another quote clears the red bar (closing the "new string").

keith-hall:
fixes #1329

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -84,6 +84,15 @@ contexts:
           - match: '(?=$\n|{{command_terminators}})'
             pop: true
           - include: variables
+          - match: '"'
+            scope: punctuation.definition.string.begin.dosbatch
+            push:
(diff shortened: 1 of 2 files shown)


FichteFoll:
What about escape sequences? Not sure if they are valid in variable assignments but I think so.

keith-hall:
support for escape chars added :)

wbond:
Thanks!

PR #1331, opened 2017-11-30
[C#] support readonly structs

keith-hall:
https://blogs.msdn.microsoft.com/mazhou/2017/11/21/c-7-series-part-6-read-only-structs/


wbond:
Thanks!

PR #1335, opened 2017-12-01
Add Travis CI config to run syntax tests

wbond:


wbond:
There is a bug in the current `syntax_tests` executable that causes the return code to be a random int instead of `0` on a successful run. Once 3156 comes out, this can be merged.

jrappen:
3156 is out.

FichteFoll:
3157 with the fixed return code is out, so Travis CI can probably be enabled for the repo now.

Issue #1337, opened 2017-12-03
[JavaScript] [RFC] Removing special handling of dollar signs.


Thom1729:
JavaScript allows dollar signs in identifiers. Dollar signs have no special meaning; they're just like any other identifier character. The popular jQuery framework names its global “namespace” variable `$`. An old version of this syntax marked identifiers consisting of a single dollar sign `keyword.other.jquery`. A later update corrected this, marking that identifier normally. This proved controversial, as some users liked the old highlighting (#141). Consequently, the syntax was modified to do both — leading dollar signs in identifiers are marked as `punctuation.dollar`, and identifiers beginning with dollar signs are marked as `variable.other.dollar` or `variable.other.dollar.only`.

Now, this may be convenient for users of a specific framework, but as implemented it's problematic for several reasons:

- It's “cruft”; building in special-case logic for a specific third-party framework. We have been moving away from that, and for good reason.
- The `$` is marked as punctuation, but this is spurious and arguably incorrect; the language makes no such distinction.
- It interferes with other `variable` subscopes. For example, as it stands all variables are marked `variable.other.readwrite` *unless* they begin with a `$`, in which case they're marked `variable.other.dollar`.
- It bloats the syntax definition. The hassle it introduces when editing the syntax is totally out of proportion to its usefulness. At present, I count *eleven contexts* that implement special-case logic to handle dollar signs in identifiers, not counting contexts that merely depend upon those. Removing the jQuery-specific logic would trim dozens of lines of dubious code.
- It's not self-consistent. There are many places that highlight identifiers that do not special-case dollar signs.

There are several potential solutions to this:

1. Preserve the behavior as-is. Clearly, I'm not a fan.
2. Preserve the behavior, make it consistent thoughout the syntax, and maintain it forever. Blech.
3. Nuke the behavior entirely. Elegant, though possibly overkill.
4. Nuke the behavior in general, but highlight `$` as `support.function.jquery`. This is in line with what we do native DOM and node.js constants. We'd still be implementing nonstandard third-party functionality in the core syntax, but it's not so bad as these things go.
5. Nuke the behavior in general, but preserve it as-is only for variables in expressions. This is inelegant and inconsistent, but it would preserve the existing behavior in common cases without metastasizing throughout the syntax.

[...]

Issue #1336, opened 2017-12-03
[JavaScript] [RFC] Removing `readwrite` from variable scopes.


Thom1729:
Originally, JavaScript did not support constants; all variables could be reassigned. Now, with the adoption of `const`, variables are not necessarily writable. Unfortunately, it is impossible to make this distinction at the time of use. In the statement `console.log(myVar);`, it is not possible for the syntax to determine whether `myVar` is a read/write variable or a read-only variable. In other words, the JavaScript syntax does not distinguish read/write variables from read-only variables.

Currently, all variables are scoped as `variable.other.readwrite` (except in some special cases). This used to be trivially correct, because all JavaScript variables were read/write. Now, it is often incorrect. There is absolutely no way to correctly mark a variable as read/write or readonly, because fundamentally this is not a syntactic distinction.

In order to fix this, there is no option but to strip the `readwrite` subscope from the common case and mark variables simply as `variable.other`. We could follow the example of some languages and mark variables beginning with uppercase letters as constants because of common convention, although the language does not enforce this and there may often be exceptions. We can't really mark all-lowercase variables as readwrite because it is very common to have lowercase constants.

It would be possible to mark variables as readwrite/readonly when they are *declared*. However, because this is really a syntactic feature of the entire declaration than of the name itself, I think it would be more appropriate to mark the entire declaration with `meta.declaration.variable.readwrite` or something similar. (This would also mean substantially less code.)

Thoughts, objections, alternative suggestions?

FichteFoll:
From my understanding, the `readwrite` name is mostly a convention for user-defined variables and used in several default syntax definitions since the initial JS rewrite in early 2016. I've never been too fond of the choice, however, since imo it would be better to name it something like `variable.user` to reflect that more accurately. (Basically, get rid of all the "other" stuff. The mere existence of that is odd.)
Additionally, "constant" variables could be called `variable.constant`, like function variables are called `variable.function`.

A similar thing was discussed while looking at scope names for annotations: https://github.com/sublimehq/Packages/issues/709#issuecomment-266835130.

FichteFoll:
Updating on that, the `variable` scope in general is intended for more or less user-controlled names for things. Basically what variables are, mostly. Technically speaking, the "readwrite" (or "default") variable type could get a scope of just `variable` and then the syntax suffix. However, I strongly believe that we need to differentiate these types of variables from the others, so it needs to get at least a second token to subclass these generic variables. Maybe `variable.generic`?

[...]

Issue #1102, opened 2017-08-08
[JavaScript] ES6: Destructuring assignment in parentheses in functions support


PaperFlu:
### Destructuring assignment in parentheses in functions

The special color shows that ST hasn't support this..or it's the right color whereas I just want you to noticed if you haven't know about this.

![Destructuring Assignment](https://user-images.githubusercontent.com/22674396/28993435-e1e17a9e-79e8-11e7-9c0a-699147995479.png)

Thanks!

Thom1729:
This should probably wait on #1009, because the syntax is largely restructured. Once those changes are merged, I should be able to implement this in short order.

wbond:
This should be resolved with the latest changes in master.

PR #1338, opened 2017-12-03
[JavaScript] Overhaul binding declarations. Feedback requested.

Thom1729:
Modern JavaScript offers destructuring assignment:

```js
const [foo, bar] = [1, 2];
console.log(foo); // 1
console.log(bar); // 1
```

Although this was not explicitly implemented in the syntax definition, it mostly worked in many common cases. I'm implementing it properly to work in all cases; then, I will implement destructuring for function parameters (#1102). This will involve a fair bit of new code as well as changes to existing code.

Before I continue, I'm looking for feedback on a couple of points:

- [Removing `readwrite` from variable scopes.](https://github.com/sublimehq/Packages/issues/1336)
- [Removing special handling of dollar signs.](https://github.com/sublimehq/Packages/issues/1337)


Thom1729:
For the time being, I'll leave in both dollar identifiers and the `readwrite` scopes.

JavaScript/tests/syntax_test_js_bindings.js
```diff
@@ -0,0 +1,38 @@
+// SYNTAX TEST "Packages/JavaScript/JavaScript.sublime-syntax"
+
+const x = value;
+//    ^ meta.binding.name variable.other.readwrite
+
+
+const [ x, y, ...z, ] = value;
+//    ^^^^^^^^^^^ meta.binding.destructuring.sequence
```
wbond: These assertions should probably continue for another 4 characters, right?

JavaScript/tests/syntax_test_js_bindings.js
```diff
@@ -0,0 +1,38 @@
...
+//       ^ punctuation.separator.comma
+//         ^ meta.binding.name variable.other.readwrite
+//          ^ punctuation.separator.comma
+//            ^^^ keyword.operator.spread
+//               ^ meta.binding.name variable.other.readwrite
+//                ^ punctuation.separator.comma
+
+const [ x, [a, b], z] = value;
+//    ^^^^^^^^^^^^^^^ meta.binding.destructuring.sequence
+//         ^^^^^^ meta.binding.destructuring.sequence meta.binding.destructuring.sequence
```
wbond: This should probably be `meta.binding.destructuring.sequence meta.binding.destructoring.sequence` to ensure it is doubled up, right?

JavaScript/tests/syntax_test_js_bindings.js
Thom1729: Yes.

JavaScript/tests/syntax_test_js_bindings.js
Thom1729: Is there a typo? I'm not seeing it.

wbond: Sorry, it must have been wrapped on my screen last night and I didn't notice the second scope name.

PR #1339, opened 2017-12-03
[JavaScript] Stripped dollar sign handling. Added new jquery constant.

Thom1729:
See #1337 for discussion.


wbond:
Unfortunately this very likely won't be possible. There was a huge backslash when we did significant changes to the JavaScript syntax, and a huge number of people complained about not being able to target the `$` as they wanted. Changing these scopes will almost certainly break a whole bunch of color schemes, which isn't something we want to do.

Thom1729:
Because this will not be accepted in its current form and it is likely to conflict with future changes to the syntax, I'm closing this PR. Any future changes along the same lines (subject to the discussion in #1337) will have a new PR.

PR #1340, opened 2017-12-04
[JavaScript] `yield` and `await` expressions

Thom1729:
`yield` and `await` now work inside expressions:

```js
    x = yield;
//      ^^^^^ keyword.control.flow

    x = yield * 42;
//      ^^^^^ keyword.control.flow
//            ^ keyword.generator.asterisk

    x = yield
    function f() {}
    [];
//  ^^ meta.sequence - meta.brackets


    x = yield*
    function f() {}
    [];
//  ^^ meta.brackets - meta.sequence

    y = await 42;
//      ^^^^^ keyword.control.flow
```

This does introduce a bug in that `yield` and `await` should be valid identifiers outside generator and async functions, respectively. The ECMAScript formal syntax represents this by using weird EBNF with parameterized productions. The equivalent implementation for JavaScript would be to have four near-complete copies of the syntax, one for async generators, one for non-async generators, one for async non-generators, and one for everything else. Alternatively, people could not use `yield` and `await` as identifiers, and that would be cool too.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -272,7 +272,7 @@ contexts:
     - match: \b(break|continue|goto)\b
       scope: keyword.control.loop.js
 
-    - match: \b(await|return)\b
+    - match: \breturn\b
       scope: keyword.control.flow.js
       push: restricted-production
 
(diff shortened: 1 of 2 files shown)


wbond:
I think it is fine to not handle those keywords as identifiers in other contexts.

Thanks!

Issue #1297, opened 2017-11-09
[JavaScript] [RFC] Refactor syntax tests


Thom1729:
JavaScript is a syntactically complex language, and its complexity continues to increase (#1269). As a result, the syntax test file is over a thousand lines. Similar tests are often grouped together, but there's no overall organizational scheme. As a result, it can be hard to find tests pertaining to a certain feature, and it's even harder to verify that the tests are comprehensive. (I suspect that there are a lot of holes, and of course adding more tests could exacerbate these issues.)

A natural solution might be to split the tests across several files. Each file could cover a major aspect of the language in a single, manageable chunk. The downside is that there would then be multiple files to test.

Currently, the Python and C bundles each have two syntax test files per language. As far as I am aware, no core bundles have more than that. How many ways is it reasonable to split a syntax test? For JavaScript, I might imagine separate files for core syntax, strings & regular expressions, and ES6 classes. Each of these files could be individually longer than most languages' syntax tests.

Thoughts?

keith-hall:
I guess you've not looked at the C# package :) https://github.com/sublimehq/Packages/tree/9e83b95ae096816005c69af1d675f065ed079a17/C%23/tests

Thom1729:
I saw the `syntax_test_c#.cs` at the bottom of the package and somehow missed the entire `tests` directory. With that in mind, I suppose there's no reason not to split the JavaScript tests accordingly.

FichteFoll:
The YAML package also has several test files inside a subfolder. I personally prefer using multiple files for organization of tests covering various areas of a language, but @wbond was more fond of single test files, which would add to the overall test file complexity and make it more unlikely likely to notice regressions.

FichteFoll:
Imo, closing each file with a test assertion against any sort of meta scope (see also the python test) is sufficient to check against potential context leaking.

wbond:
There have been multiple times where changes to the C syntax have been caught by the interaction of different contexts, which would not have been caught if the tests were split out into different files. A test at the end of the file for a meta scope would catch certain errors, however the syntax may be in an incorrect context that doesn't have a meta scope. The error may not show up until you try and parse further syntax.

I am not against organizing files into sections, but I think there is a benefit to having the tests in a single file.

Thom1729:
[...]

PR #1343, opened 2017-12-04
[JavaScript] Moved tests to own directory.

Thom1729:
In #1297, we discuss splitting the massive-yet-woefully-incomplete JavaScript test file into several smaller, more manageable files. Here, we move the existing tests to a dedicated subdirectory in preparation for splitting and expanding them.

The first new test file will probably cover regular expressions (in preparation for adding the new regexp features described in #1269).


wbond:
Since this was conflicted, I just made a commit with the same actions - 3c2c7f1

PR #1344, opened 2017-12-04
[JavaScript] Added spread operator and made computed property names consistent.

Thom1729:
Implemented object literal [spread properties](https://github.com/tc39/proposal-object-rest-spread). (Stage 3, widely used, very stable.)

Also fixed inconsistencies with computed property names:

The square brackets were sometimes `punctuation.section.brackets` and sometimes `punctuation.definition.symbol`. The latter seems clearly wrong, so I standardized on the former.

When the value of a property is a function expression (determined by lookahead), we generally try to scope the property name as `entity.name.function`. This works fine with literal property names, mostly fine with string and number property names, and not at all for computed property names. Consider the following:

```js
({
    [foo.bar]: function() {}
})
```

The string `foo.bar` was scoped `entity.name.function`. But this is incorrect; we have no way of knowing the name of the function (insofar as that's meaningful in this case), but it certainly isn't "foo.bar". Rather, `foo.bar` is an expression with an identifier `foo`, punctuation `.`, and a property name `bar`. Now, it is highlighted accordingly.

The same applies to computed method names and class field names. These have been fixed as well.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1180,6 +1180,10 @@ contexts:
           scope: punctuation.section.block.js
           pop: true
 
+        - match: \.\.\.
+          scope: keyword.operator.spread.js
+          push: literal-variable
+
         - match: >-
             (?x)(?=
               {{property_name}}\s*:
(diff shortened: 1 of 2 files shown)


FichteFoll:
The same is also useful for Lua, although that doesn't have symbol highlighting for functions in tables at all currently.

wbond:
The questions I have in regards to computed function names is: how should be show these in the symbol list. Clearly we can't include them in the index since we don't know the value, but I feel like they should show up in some form in the symbol list.

Thom1729:
The JavaScript symbol list includes `meta.function.declaration` rather than `entity.name.function`. As a result, this change will not prevent methods with computed names from appearing in the symbol list.

Example:

```js
class Foo {
    bar(){}

    [baz]() {}
}
```

The symbol list contains `Foo`, `bar()`, and `[baz]()`.

wbond:
Sounds good!

PR #1345, opened 2017-12-04
Upate D.sublime-syntax for new function contracts.

veelo:
Since recently (http://forum.dlang.org/post/esjclvbebspqkekjmwki@forum.dlang.org) the `do` keyword can be used in place of the `body` keyword in function declarations with contracts.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -261,7 +261,7 @@ contexts:
             - include: $top_level_main
         - match: (in|out)
           scope: keyword.control.statement.d
-        - match: (?=body)
+        - match: (?=body|do)
           pop: true
     - match: '(?<=\})'
       pop: true
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -159,7 +159,7 @@ contexts:
       scope: keyword.control.branch.d
     - match: \b(while|for|do|foreach(_reverse)?)\b
       scope: keyword.control.repeat.d
-    - match: \b(version|return|with|invariant|body|scope|in|out|inout|asm|mixin|function|delegate)\b
+    - match: \b(version|return|with|invariant|body|do|scope|in|out|inout|asm|mixin|function|delegate)\b
       scope: keyword.control.statement.d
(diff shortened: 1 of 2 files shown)


veelo:
There you go, thanks for the friendly guidance.

veelo:
Is there anything else that prevents a merge?

FichteFoll:
ST developers are currently working on the rendering engine because of some performance bugs in certain situations on macOX. Pull requests will be reviewed again once that is done, when nearing the next dev build.

wbond:
Sorry for the delay in reviewing this.

I definitely appreciate someone familiar with D contributing. If you are inclined to improve the syntax some more, we'd definitely be happy to accept your contributions, assuming you are fairly familiar with D.

If you do want to contribute some more, it may make sense to read up on some of our [standardized scope naming guidelines](http://www.sublimetext.com/docs/3/scope_naming.html) and see if we can make the tests more robust, thus allowing more improvements to the D syntax with fewer regressions.

If you are looking for some other syntax definitions to learn some from, I'd check out Python, JavaScript, Rust, and maybe C (although it has some warts due to dealing with the C preprocessor).

veelo:
Thanks. I'll keep this in mind.

The D compiler front-end is being made available as a library (alpha-stage currently) which will enable all kinds of interesting applications including syntax analysis. Are there, or could there be in the future, alternative ways to do syntax highlighting in ST using external lexers/parsers? This would make it easier for the highlighter to grow with the language definition and eliminate the need to replicate the grammar  in syntax files and make the highlighter more robust. It may even be able to do a better job at lexing all the crazy stuff that is possible in D. The compiler already has syntax highlighting built in that it uses in colored error messages.

wbond:
@veelo Unfortunately Sublime Text likely won't ever be able to support external syntax highlighting engines, just due to performance issues. We have all sorts of optimizations in place to make syntax highlighting fast, including a custom regex engine.

However, functionality like what you describe sounds like it would fit well with the Language Server Protocol for adding other functionality to editors such as Sublime Text.

PR #1346, opened 2017-12-05
[PHP] Add new type hinting "object" for PHP 7.2

kinoute:
PHP 7.2 introduces the "object" type hint: https://wiki.php.net/rfc/object-typehint

Without this modification, it looks actually like this:

capture d ecran 2017-12-05 a 13 47 36

After this PR, it will correctly use the same colors of the other type hints:

capture d ecran 2017-12-05 a 13 46 58

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -586,7 +586,7 @@ contexts:
         - match: (?=\S)
           pop: true
     - include: comments
-    - match: \b(array|callable|int|string|bool|float)\b
+    - match: \b(array|callable|int|string|bool|float|object)\b
       scope: storage.type.php
     # Class name type hint
     - match: '{{identifier}}'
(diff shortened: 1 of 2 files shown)


PHP/PHP Source.sublime-syntax
```diff
@@ -643,7 +643,7 @@ contexts:
       scope: punctuation.separator.php
     - match: '\?'
       scope: storage.type.nullable.php
-    - match: \b(array|bool|int|float|string)\b
+    - match: \b(array|bool|int|float|string|object)\b
```
jfcherng: - add the missing `callable` as well?
  - make these 2 changed lines exactly the same.
    - make the following into a new context so we don't have to change this everywhere in the future?

```yaml
- match: \b(array|callable|int|string|bool|float|object)\b
  scope: storage.type.php
```

kinoute:
I added "callable" and use the same line for both situations. Not sure I'm qualified for your last point though, sorry

wbond:
Sorry for the delay in getting this reviewed.

@kinoute Would you mind adding a couple of tests for this in https://github.com/sublimehq/Packages/blob/master/PHP/syntax_test_php.php?

wbond:
Thanks!

PR #1347, opened 2017-12-05
[Ruby] Add Snippet map(&:e)

stoivo:


Ruby/Snippets/map-(&:e)-(map).sublime-snippet
```diff
@@ -0,0 +1,6 @@
+<snippet>
+	<content><![CDATA[map($:${1:e})]]></content>
+	<tabTrigger>map&</tabTrigger>
```
FichteFoll: Are you sure you want to be using `&` here? Regardless of this being an XML file and `&amp;` the properly escaped variant, you use `$` in the actual snippet and description. But then you use `&` in the file name and PR title.

Also note that special characters are a little wonky currently and it is probably not a good idea to use them until this is resolved: https://github.com/SublimeTextIssues/Core/issues/1061

stoivo: I did not think of `&amp;`. I'm not sure what a good tab trigger would be? it is calling the same method map.

I don't know why I wrote `$` in the content I should be `&`. So I updated the PR.

stoivo: Thanks for commenting so quick 🎉

stoivo:
What are we waiting for?

stoivo:
hello?

wbond:
Generally, I don't think we'll be adding new snippets to the default packages. They generally tend to be somewhat user-specific due to formatting, and community-maintained snippet packages are a better fit.

Issue #1216, opened 2017-10-10
[Java] Javadoc can 'escape' its bounds


lastsecondsave:
From JDK's `HashMap`:

![sketch 1](https://user-images.githubusercontent.com/6729879/31411174-70eb7514-ae19-11e7-9e73-b9d46c311088.png)

Second angle from `<<` opens tag and it never closes. Here is that comment in text:
```
/**
 * The maximum capacity, used if a higher value is implicitly specified
 * by either of the constructors with arguments.
 * MUST be a power of two <= 1<<30.
 */
static final int MAXIMUM_CAPACITY = 1 << 30;
```

wbond:
This was fixed with the Javadoc refactor

Issue #966, opened 2017-05-05
[Java] Javadoc closing tags highlighted, but not opening tags


djspiewak:
```java
/**
 * <p> This class represents either a table of bits or a set of non-negative
 *     numbers.</p>
 */
```

![](http://i.imgur.com/MOxanrt.png)

wbond:
This was fixed in #1351

PR #1351, opened 2017-12-09
[Java] Refactor javadoc

lastsecondsave:
Javadoc's syntax hasn't been changed since tmLanguage era and there are so many things wrong with it so I just redid it from the scratch.

'Strategic' changes:

1. Syntax is now hidden. It is part of Java, by itself it's pointless.
2. It is not included anymore, it now relies on new `embed`/`escape` functionality. This fixes issue #1216.
3. URL's highlighting was removed. It isn't part of the javadoc, links are defined with html's `<a>` tag (which is already highlighted).

Fixes/improvements:

1. One-line javadocs now work.
2. Leading html tag doesn't turn off the highlighting.
3. Lines without leading asterisk are not treated as invalid anymore. The asterisk is optional since 2002.
4. All inline tags can now be multiline. They do not re-highlight the leading asterisk.
5. 'code' tag now supports brackets' balancing.
6. 'link' tag has reference and label scoped properly. Label can now contain html tags.
7. 'param' tag has its parameters highlighted. Type parameter is not recognized as html tag anymore.
8. 'see', 'throws' and other tags with reference now have this reference recognized.
9. Added new tags from Java 8 and 9.

Before:

After:

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -295,8 +295,16 @@ contexts:
   comments:
     - match: /\*\*/
       scope: comment.block.empty.java punctuation.definition.comment.java
-    - include: scope:text.html.javadoc
+    - include: javadoc
     - include: comments-inline
+  javadoc:
+    - match: /\*\*
+      scope: comment.block.documentation.javadoc punctuation.definition.comment.begin.javadoc
+      embed: scope:text.html.javadoc
+      embed_scope: comment.block.documentation.javadoc
+      escape: \*/
+      escape_captures:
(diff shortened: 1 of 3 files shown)


lastsecondsave:
BTW this also fixes issue #966

FichteFoll:
Always love seeing screenshot comparisons of PR changes.

wbond:
Awesome, thanks for this!

rwols:
Perhaps we can use this in C/C++/ObjC/ObjC++ too?

wbond:
@rwols Do people use Javadoc in C languages?

rwols:
No, but it's so similar to Doxygen, I wonder if it'll "just work".

Issue #1146, opened 2017-09-15
Java Syntax Highlighting


Palidino:
Version 3126: https://imgur.com/RZGvEIP
Version 3143: https://imgur.com/bYjHcvm

EDIT: Monokai, 3126 next to 3143: https://imgur.com/ciJ4NJN

Final/capitalized variables are no longer being colored correctly. Method return types are no longer highlighting, and periods (though this may be intentional) no longer use the color previously associated with them. This issue happens with the default and adaptive theme, and also a custom theme I use.

Here's a piece of code demonstrating the issues displayed in the image. These issues occur with any of the default themes and color schemes with #3143, but were fine in #3126.

```
public class ClassDemo implements Runnable {
    public static final Object COLOR_INCORRECT = new Object();
    public static final Object COLOR_INCORRECT_2 = new Object(), COLLOR_CORRECT = new Object();

    @Override
    public void run() {  }

    public Object returnTypeColorIncorrect() {
        //The 'Object' return type is not the right color (should be the same
        //as void return type up above).
        //Periods using current return type color, unsure if this is correct. It
        //previously used the colors associated with things such as the if,
        //else, new, this, super, etc keywords.
        return COLLOR_CORRECT.toString();
    }
}
```

wbond:
> This issue happens with the default and adaptive theme, and also a custom theme I use.

Theme's don't control the color of the code. Color schemes do. Try Mariana and see if you get a different experience.

> Annotations and final/capitalized variables are no longer being colored correctly.

The scope names for annotations have changed, so unless you've kept up with the scope naming standards, they will look different. This is on purpose since the old scope names for Java annotations were poor or incorrect.

Unfortunately your first link is broken, so I am unsure as to the differences in final/capitalized variables.

Palidino:
Sorry I updated my post, and I had meant to say that I did use the color schemes that come with sublime and the issues still remain, I also copied the images from the forum post over incorrectly. Outside of the annotation color, the other issues did happen with the other color schemes.

Mariana color scheme, you can see the inconsistency between the final keywords at the top of the class and the return type keywords (void and Object) being different.
https://imgur.com/lPmds3G

wbond:
The `final` and `return` keywords look highlighted properly to me in 3143 and (especially `final`) broken in 3126.

It appears the Java syntax used to give some sort of scope to uppercase identifiers, which it no longer does. Does Java treat uppercase identifiers differently in some way, or is it more of a style thing?

[...]

PR #1352, opened 2017-12-10
[Java] Use same color for constants

lastsecondsave:
See issue #1146.

For default color schemes makes all constants to appear with the same color.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -391,7 +391,7 @@ contexts:
         1: punctuation.accessor.dot.java
   constant-entity:
     - match: '(\.)?(?!{{classes_named_as_constants}})\b([A-Z][A-Z0-9_]+)(?!<|\.class|\s*\w+\s*=)\b'
-      scope: entity.name.constant.java
+      scope: entity.name.constant.java constant.other.java
       captures:
         1: punctuation.accessor.dot.java
   enums:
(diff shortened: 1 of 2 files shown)


FichteFoll:
I am strongly opposed to this change since `entity.name.constant` is the more appropriate scope name, yet is receiving a weaker binding.

This should be acceptable if it was the other way around, though.

lastsecondsave:
I see these alternatives:

1. Use same color in color schemes for `constant.other` and `entity.name.constant.other`.
2. Change `entity.name.constant` to `meta.entity.name.constant` and update symbols' list to use this scope. It is not any better than solution implemented but at least this way we don't stack two unrelated scopes together.

lastsecondsave:
To clarify, by 1 I meant to actually update color schemes bundled with Sublime. I too as @Palidino see no reason for uppercase identifiers to have 2 colors, as a Java dev you see all of them as the same thing. Can't say about other languages though.

FichteFoll:
The upper-case variable coloring is based on conventions since there are no hard rules for how constants should be named in any language that I know. For the most part, this does a reasonably good job since many languages have these conventions, and I quite like it. This is currently done in at least Python, Java and Rust.
Again, if people do not want special highlighting, they can easily target the scopes and color them in the usual color, but you can't easily do the opposite.

See also: https://github.com/sublimehq/Packages/pull/1242, https://github.com/sublimehq/Packages/pull/1193

lastsecondsave:
OK, now I agree. Let's deal with color inequality with color schemes. I'll close this PR (previous time I missed the button).

Thank you!

PR #1354, opened 2017-12-13
[CSS] allow url function to be used in at-namespace

keith-hall:
as used by Firefox chrome customizations - https://support.mozilla.org/en-US/questions/1178962#answer-1015047


wbond:
Thanks!

PR #1360, opened 2017-12-19
[Rails] Rails partial snippet

alassiter:
I would like to default to the new hash syntax


wbond:
Looks good to me

wbond:
Thanks!

PR #1362, opened 2017-12-20
[Rust] fix nested array template arguments

### LINKED ISSUE #995: [Rust] Incorrect error highlighting with nested array template arguments
I'm using Sublime Text 3, Dev Build 3131 on macOS Sierra. The second semicolon gets erroneously highlighted as error in the last two cases:
incorrect-highlighting

keith-hall:
fixes #995


wbond:
Thanks!

Issue #736, opened 2016-12-15
[Java] Correctly scope fully-qualified identifiers


djspiewak:
The Java mode currently uses a `qualified_id` regex to scope identifiers.  This has two problems.  First, it doesn't allow fully-qualified identifiers to split across lines (which is valid in Java).  Second, it incorrectly manages both the meta scopes and the punctuation scopes (since the `.` will be scoped as part of the identifier).

This is non-trivial to correct, due to the stateful contexts which use `qualified_id`, but still worth doing.

ismell:
How do we see this working?

foo.bar.MyClass.SubClass

what scopes should be applied to `foo`, `bar`, `MyClass` and `SubClass? Is there a meta scope wrapping the entire thing?

Do we have a scope for foo.bar that denotes it as a class path?

djspiewak:
> what scopes should be applied to foo, bar, MyClass and `SubClass?

I would have to look up exactly what is correct here, but off the top of my head:

```java
   foo.bar.MyClass.SubClass
// ^^^ variable.package.java
//    ^ punctuation.accessor.java
//     ^^^ variable.package.java
//        ^ punctuation.accessor.java
//         ^^^^^^^ support.class.java
//                ^ punctuation.accessor.java
//                 ^^^^^^^^ support.class.java
// ^^^^^^^^^^^^^^^^^^^^^^^^ meta.identifier.java
```

Something to that effect, anyway.

ismell:
if we had foo.bar.MyClass.SubClass.myStaticMethod(). I would assume myStaticMethod would not be inside the meta.identifier.

This gets interesting when referencing variables:

    class Node {
        Parent parent;
    }

    Node node = ...;

    node.parent.parent = new Node();
    
    node.parent // <- Can't tell here if this is a meta.identifier, or just a variable with an accessor
        .parent = new Node();

rwols:
Might want to look at how Python handles qualified paths.

FichteFoll:
Python handles them well in general, but the scope names are quite unconventional since I just needed *something* to test against. They will likely change in the future to something suggested here or similarly.

lastsecondsave:
Is anyone here working on it right now? I'm experimenting with scoping packages in Java (I want all dots be the same color), maybe publish it in a week or two. Don't want to clash with someone's work.

ismell:
Not working on it. Let me know how it goes!

PR #1366, opened 2017-12-23
[Java] Fully Qualified Name

lastsecondsave:
Fully qualified name (FQN) is mentioned in issue #736.

Here is the way FQN is scoped now:

```
com.example.component.ClassName
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ support.class
                     ^ punctuation.accessor.dot
```

My problem with that is that only last dot is scoped and has its own color. Instead it should be scoped like this:

```
com.example.component.ClassName
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.path
^^^ support.type.package
   ^ punctuation.accessor.dot
    ^^^^^^^ support.type.package
    ...
                     ^ punctuation.accessor.dot
                      ^^^^^^^^^ support.class
```

To achieve this I completely redid name's matching. I also heavily modified contexts that are mostly something around an FQN (like generics and import statements) and tweaked some others. Unfortunately the deeper you dig the more broken stuff you find.

Fixes/amendments:

1. Packages and dots in fully qualified names are scoped. All the dots (finally) have the same color. For the most cases such names can span multiple lines.
2. Package of the class (the one in the beginning of every Java file) is scoped as `entity.name.namespace`. It kind of deserves it.
3. `static` in `import static` statement is scoped as a `keyword`.
4. Asterisks in imports are recognized.
5. Constants in static imports are correctly recognized.
6. All constants' declarations are recognized, ever for constants declared in one statement.
7. Constants can be of any language, not only English.
8. Uppercase class names are correctly recognized in generics, fields and methods' parameters.
9. Generics' multiple bounds operator (`&`) is recognized.
10. No more generic/array related false-positive invalids.
11. Diamond operator (`<>`) is drawn with a ligature.
12. `void` is not considered as primitive type anymore.
13. `threadsafe` is not considered as storage modifier anymore (how did it get here? There is no such thing in Java).
14. All curly braces are scoped as `punctuation.section.block/braces`.

About FQN and line-breaks: in scopes where we know the class name is to follow the line-breaks are allowed. Like in import/throws/new statements, fields, parameters, inside of generics, technically almost all of the scopes with class names. The exception is method's body, here we can't say what `a.b.c` is. It maybe a package, maybe a chain of variables. Here only one-line FQNs are recognized.

Before:

After:

### DIFF
--- a/Java/syntax_test_java.java
+++ b/Java/syntax_test_java.java
@@ -1,25 +1,106 @@
 // SYNTAX TEST "Packages/Java/Java.sublime-syntax"
 
 package apple;
-// <- source.java meta.package.java keyword.other.package.java
-//      ^ meta.package.java support.other.package.java
-//           ^ meta.package.java punctuation.terminator.java
+// <- source.java meta.package-declaration.java keyword.other.package.java
+//      ^^^^^ entity.name.namespace.java
+//           ^ punctuation.terminator.java
+
+package com.example.apple;
+//^^^^^^^^^^^^^^^^^^^^^^^ meta.package-declaration.java
+//      ^^^^^^^^^^^^^^^^^ entity.name.namespace.java
+//         ^ punctuation.accessor.dot.java
+//                 ^ punctuation.accessor.dot.java
+//                       ^ punctuation.terminator.java
 
 import a.b.Class;
-// <- meta.import.java keyword.other.import.java
-//     ^ meta.import.java support.class.import.java
+// <- meta.import.java keyword.control.import.java
+//     ^^^^^^^^^ meta.path.java
+//     ^ support.type.package.java
+//      ^ punctuation.accessor.dot.java
+//       ^ support.type.package.java
 //        ^ punctuation.accessor.dot.java
-//              ^ meta.import.java punctuation.terminator.java
+//         ^^^^^ support.class.import.java
+//              ^ punctuation.terminator.java
+
+import a.b.Class.SubClass;
+//^^^^^^^^^^^^^^^^^^^^^^^ meta.import.java
+//              ^ punctuation.accessor.dot.java
+//               ^^^^^^^^ support.class.import.java
+
+import a.b.Class.*;
+//^^^^^^^^^^^^^^^^ meta.import.java
+//              ^ punctuation.accessor.dot.java
+//               ^ keyword.operator.wildcard.asterisk.java
+
+import com.google
+//     ^^^^^^^^^^ meta.import.java meta.path.java
+//        ^ punctuation.accessor.dot.java
+//         ^^^^^^ support.type.package.java
+  .common.collect
+//^ punctuation.accessor.dot.java
+//       ^ punctuation.accessor.dot.java
+//        ^^^^^^^ support.type.package.java
+//^^^^^^^^^^^^^^^ meta.import.java meta.path.java
+  .ListMultimap;
+//^ punctuation.accessor.dot.java
+// ^^^^^^^^^^^^ support.class.import.java
+//^^^^^^^^^^^^^ meta.import.java meta.path.java
+//             ^ punctuation.terminator.java
+
+import no.terminator
+// <- meta.import.java keyword.control.import.java
+
+import static no.terminator
+// <- meta.import.java keyword.control.import.java
+
+import
+// <- meta.import.java keyword.control.import.java
+
+import static
+// <- meta.import.java keyword.control.import.java
+
+import java.net.URL;
+// <- meta.import.java keyword.control.import.java
+//^^^^^^^^^^^^^^^^^ meta.import.java
+//     ^^^^^^^^^^^^ meta.path.java
+//              ^^^ support.class.import.java
+//     ^^^^ support.type.package.java
+//         ^ punctuation.accessor.dot.java
+//          ^^^ support.type.package.java
+//             ^ punctuation.accessor.dot.java
+//                 ^ punctuation.terminator.java
+
+import java.util.*;
+//^^^^^^^^^^^^^^^^ meta.import.java
+//     ^^^^^^^^^^^ meta.path.java
+//     ^^^^ support.type.package.java
+//         ^ punctuation.accessor.dot.java
+//          ^^^^ support.type.package.java
+//              ^ punctuation.accessor.dot.java
+//               ^ keyword.operator.wildcard.asterisk.java
 
 import static a.b.Class.fooMethod;
-// <- meta.import.java keyword.other.import.java
-//     ^  meta.import.java storage.modifier.static.java
-//            ^ meta.import.java support.function.import.java
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.import.java
+// <- meta.import.java keyword.control.import.java
+//     ^^^^^^ meta.import.java keyword.control.import.static.java
+//             ^ punctuation.accessor.dot.java
+//            ^^^^^^^^^ meta.path.java
+//                 ^^^^ support.class.import.java
+//               ^ punctuation.accessor.dot.java
 //                     ^ punctuation.accessor.dot.java
-//                               ^ meta.import.java punctuation.terminator.java
+//                      ^^^^^^^^^ meta.import.java support.function.import.java
+//                               ^ punctuation.terminator.java
 
 import static a.b.Class.CONSTANT;
-/*                      ^ constant.other.java */
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.import.java
+//            ^^^^^^^^^ meta.path.java
+//                 ^^^^ support.class.import.java
+//                     ^ punctuation.accessor.dot.java
+//                      ^^^^^^^^ constant.other.import.java
+
+import static a.b.Class.*;
+//                     ^ punctuation.accessor.dot.java
+//                      ^ keyword.operator.wildcard.asterisk.java
 
 public class SyntaxTest {
 //^^^^^^^^^^^^^^^^^^^^^^^ meta.class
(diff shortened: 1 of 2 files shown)


Java/syntax_test_java.java
```diff
@@ -398,6 +475,75 @@ void anonymousFunctions() {
...
+    List<URI> x;
+//      ^^^^^ meta.generic.java
+//       ^^^ support.class.java
+
+    List<java.net.URI> x;
+//      ^^^^^^^^^^^^^^ meta.generic.java
+//       ^^^^^^^^^^^^ meta.path.java
+//                ^^^ support.class.java
+
+    List<int[]> x;
```
ismell: Can you add a test for List<int> and make sure the primitive is marked as invalid.

Java/syntax_test_java.java
```diff
@@ -420,13 +566,41 @@ void anonymousFunctions() {
 //          ^^^^^ keyword.declaration.super.java
 
   new Foo<int>();
-//        ^^^ invalid.illegal.primitive-instantiation.java
+//        ^^^ -storage.type.primitive
```
ismell: This shouldn't be allowed. It would require new Foo<Integer>

ismell:
This is awesome. I'll try and do a deeper code review later.

Java/syntax_test_java.java
lastsecondsave: This construction is legal, you can use array as a type argument.

ismell: Sorry I should have said, can you add a fest for List<int> x; as the int is invalid.

lastsecondsave: Well, about primitives in generics. I actually opted out of using `invalid` scope in favor of just not highlighting them. This lonely lint feels out of place, especially in a language in which you won't go very far without a compiler. And it activates when you type an array, the behavior I find a bit unpleasant.

I'd really like to hear another opinions on this topic. I personally think that the syntax file should mostly care about the correct stuff. The compiler will handle incorrect one much better.

rwols: > The compiler will handle incorrect one much better.

Yes. Don't use the `invalid` scope unless you're **absolutely, provable, 100% certain** that something is invalid. In practice this means you can scope unmatched parentheses/brackets/braces as invalid, but that's about it. Don't be too clever.

ismell: Ah, I didn't think about the scopes applying white typing out the array.

Java/syntax_test_java.java
```diff
@@ -1,25 +1,96 @@
 // SYNTAX TEST "Packages/Java/Java.sublime-syntax"
 
 package apple;
-// <- source.java meta.package.java keyword.other.package.java
-//      ^ meta.package.java support.other.package.java
-//           ^ meta.package.java punctuation.terminator.java
+// <- source.java meta.package-declaration.java keyword.other.package.java
+//      ^^^^^ entity.name.package.java
```
ismell: entity.name.namespace.java

Java/syntax_test_java.java
```diff
@@ -1,25 +1,96 @@
...
 // <- meta.import.java keyword.other.import.java
-//     ^ meta.import.java support.class.import.java
+//     ^^^^^^^^^ meta.path.java
+//     ^ support.type.package.java
+//      ^ punctuation.accessor.dot.java
+//       ^ support.type.package.java
+//        ^ punctuation.accessor.dot.java
+//         ^^^^^ support.class.import.java
+//              ^ punctuation.terminator.java
+
```
ismell: Can you add a test for:

    import a.b.Class.SubClass;

SubClass gets drops out of the meta.import and meta.path scopes.

Java/syntax_test_java.java
```diff
@@ -1,25 +1,96 @@
...
+//^^^^^^^^^^^^^^^^ meta.import.java
+//     ^^^^^^^^^^^ meta.path.java
+//     ^^^^ support.type.package.java
+//         ^ punctuation.accessor.dot.java
+//          ^^^^ support.type.package.java
+//              ^ punctuation.accessor.dot.java
+//               ^ keyword.operator.wildcard.asterisk.java
 
 import static a.b.Class.fooMethod;
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.import.java
```
ismell: not sure if it's possible to structure the meta classes to make it easy for a script to pull out all the static and non static imports:

    view.find_by_selector("meta.import meta.path") # doesn't get static methods.

Java/syntax_test_java.java
```diff
@@ -451,7 +625,9 @@ void test2() {
 // <- meta.annotation punctuation.definition.annotation
```
ismell: This rule should have a `- meta.annotation.identifier` at the end.

Java/syntax_test_java.java
```diff
@@ -451,7 +625,9 @@ void test2() {
...
  // <- meta.annotation.identifier
 //^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation
-//^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation.identifier
```
ismell: meta.annotation.identifier should contain the entire classpath

Java/syntax_test_java.java
```diff
@@ -603,7 +779,28 @@ void annotatedArgs(@NonNull final String p1,
 // <- meta.annotation.java meta.annotation.parameters.java punctuation.section.parens.end.java
 @fully.qualified.Annotation
 // <- punctuation.definition.annotation.java
-//^^^^^^^^^^^^^^^^^^^^^^^^^ variable.annotation.java
+//^^^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation.java meta.annotation.identifier.java meta.path.java
+//^^^^ variable.annotation.package.java
```
ismell: @wbond `variable.annotation` is really strange for Java. All annotations must be interface references.

Ideally I would like to see `variable.annotation.package.java` => `support.type.package.java` and `variable.annotation.java` => `support.class.java`. Having two ways of referencing the same thing feels strange. If a theme wants to target the annotations they can use `meta.annotation.identifier`.

FichteFoll: See https://github.com/sublimehq/Packages/issues/709 and https://github.com/sublimehq/Packages/issues/737 for context.

Java/syntax_test_java.java
lastsecondsave: I'll fix `meta.import`, but `meta.path` was intentional. Nested classes are members of the outer class so technically they can't be fully qualified. They just come along with the id of the outer class which can either be simple of fully qualified. Same logic is used in the other contexts as well.

Java/syntax_test_java.java
lastsecondsave: I definitely won't get methods (and constants) as `meta.path` can only be part of the class id. We may add something like `meta.imported` for everything that comes after the keywords.

lastsecondsave:
Changes:
- `entity.name.package` -> `entity.name.namespace` (should reread scope naming guide again).
- `keyword.other.import` -> `keyword.control.import`.
- Fix subclass is dropped out of `meta.import` scope.
- Fix `punctuation.definition.annotation` was inside of `meta.annotation.identifier`.

Java/syntax_test_java.java
```diff
@@ -1314,9 +1574,11 @@ public V apply(V input) {
 // <- storage.type.java
 
 public class Generic<T> implements fully.qualified.Other<T> {
```
ismell: fully.qualified.Other should be in `meta.path`.

Java/syntax_test_java.java
```diff
@@ -603,7 +788,28 @@ void annotatedArgs(@NonNull final String p1,
...
 @fully.qualified.Annotation
 // <- punctuation.definition.annotation.java
-//^^^^^^^^^^^^^^^^^^^^^^^^^ variable.annotation.java
+//^^^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation.java meta.annotation.identifier.java meta.path.java
+//^^^^ variable.annotation.package.java
+//    ^ punctuation.accessor.dot.java
+//     ^^^^^^^^^ variable.annotation.package.java
+//              ^ punctuation.accessor.dot.java
+//                ^^^^^^^^^ variable.annotation.java
+@fully.qualified.ParentClass.InnerAnnotation
```
ismell: Another comment regarding the one above.

If I understand what #737 says about annotations, only the last part of the qualifier should be marked with `variable.annotation.java`. In this case `InnerAnnotation`.

ismell:
I didn't review the actual sublime-syntax, but the tests look good. 

LGTM

rwols:
With such a big change, someone should probably check it out locally and test if it works on a few large Java projects.

ismell:
I've been running it locally for a few days now and haven't noticed any regressions.

Java/syntax_test_java.java
lastsecondsave: It is, it's the test that is missing.

Java/syntax_test_java.java
lastsecondsave: The idea behind the funny scope of the annotation's package is to keep the complete identifier the same color. Same with `entity.other.inherited-class`. Both of these kind of stand out from the rest of the code and with one color this behavior is more 'emphasized'. But that's my personal view, maybe I'm missing something.

Java/syntax_test_java.java
lastsecondsave: Added test.

FichteFoll:
Unfortunately the force-pushed commits make it quite hard to monitor the actual changes as nobody can remember the exact changes of the previous commit.

wbond:
This seems like a significant improvement, and based on the feedback so far, I think we'll merge it in to gather more feedback from other users.

Thanks for the work @lastsecondsave and everyone who has reviewed it.

PR #1368, opened 2017-12-26
[Groovy] Improve string syntax highlighting

JoshuaWebb:
- Highlight triple quoted strings
- Highlight "dollar slashy strings"
- Highlight unclosed strings
- Highlight erroneous newlines in single line strings
- Highlight invalid identifiers in interpolated strings
- Highlight interpolation in slashy strings ("regexp")
- Enable full language highlighting inside interpolation
- Only include the proper unicode ranges in identifiers (variables)
  inside string interpolation.
- Only match valid escape sequences (instead of any character at all);
  note that the different kinds of strings don't share all of the same
  escape sequences.
- Add a whole bunch of tests for strings

Note: Technically "slashy strings" (scoped as "regexp" here) allow
newlines as well, but I think more work needs to be done before
removing the lookahead expression on the opening '/'.

For instance, the following is already highlighted incorrectly and it
would probably lead to even more troubles if these were set free from
their current single line containment (without additional work).

```groovy
// Current (incorrect) highlighting
def a = 1 / 2 + / 3 /
//      ^ constant.numeric.groovy
//        ^^^^^^^ string.regexp.groovy
//                ^ constant.numeric.groovy
//                  ^ keyword.operator.arithmetic.groovy

// Correct highlighting
def a = 1 / 2 + / 3 /
//      ^ constant.numeric.groovy
//        ^ keyword.operator.arithmetic.groovy
//          ^ constant.numeric.groovy
//            ^ keyword.operator.arithmetic.groovy
//              ^^^^^ string.regexp.groovy
```

### DIFF
--- a/Groovy/Groovy.sublime-syntax
+++ b/Groovy/Groovy.sublime-syntax
@@ -6,6 +6,23 @@ file_extensions:
   - groovy
   - gvy
   - gradle
+variables:
+  unicode_letter: |-
+    (?:(?xi)
+      # Valid unicode letters according to:
+      # http://groovy-lang.org/syntax.html#_normal_identifiers
+      #   Literal Unicode         Escaped Unicode
+          [\x{00C0}-\x{00D6}]  |  \\u00C[0-9A-F] | \\u00D[0-6]
+        | [\x{00D8}-\x{00F6}]  |  \\u00D[89A-F]  | \\u00E[0-9A-F] | \\u00F[0-6]
+        | [\x{00F8}-\x{00FF}]  |  \\u00F[89A-F]
+        | [\x{0100}-\x{FFFE}]  |  \\u0[1-9A-F][0-9A-F]{2} | \\u(?!FFFF)[1-9A-F][0-9A-F]{3}
+    )
+
+  # dollars aren't allowed in the single dollar interpolated identifiers
+  # (dotted expressions), but they are supposed to be valid characters
+  # in identifiers in other contexts
+  # e.g. `"$$a"` is invalid, but `"${$a}"` is fine.
+  single_dollar_interpolation_identifier: (?:{{unicode_letter}}|[a-zA-Z_])(?:{{unicode_letter}}|[a-zA-Z0-9_])*
 scope: source.groovy
 contexts:
   main:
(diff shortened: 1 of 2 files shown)


JoshuaWebb:
Just did a cheeky rebase as I accidentally used `-`s instead of `_`s in a variable name in a last minute cleanup and didn't re-run my tests... silly me.

Groovy/Groovy.sublime-syntax
```diff
@@ -386,6 +421,32 @@ contexts:
       scope: storage.type.def.groovy
     - match: '\b(boolean|byte|char|short|int|float|long|double)(?:\[\s*\])*\b'
       scope: storage.type.primitive.groovy
+  single-dollar-string-interpolation:
+    - match: \${{single_dollar_interpolation_identifier}}
+      scope: variable.other.interpolated.groovy
+      push:
+        - match: \.(?={{single_dollar_interpolation_identifier}})
+          scope: keyword.operator.navigation.groovy
```
keith-hall: I think the other syntaxes tend to scope this as `punctuation.accessor.dot` as opposed to a `keyword`.

JoshuaWebb: I'm happy to update it, I've mostly been guessing/copying when it comes to scopes; in this case I copied it from https://github.com/sublimehq/Packages/blob/ce7af4d6177d0340230893e49742070ae4143246/Groovy/Groovy.sublime-syntax#L164

keith-hall: yeah, the Groovy syntax hasn't been updated for a while so probably contains a few existing scopes that are not standardized; it'd be great if you could "update" that line too, please. :) The `punctuation.accessor` scope can be found in the official scope naming documentation at https://www.sublimetext.com/docs/3/scope_naming.html.

wbond:
Thanks for these improvements! I would be great to have more input from you in terms of fixes, test or reviewing others work on the Groovy syntax in the future.

The only minor thing I noticed was using `captures: 0: ` instead of `scope:`. The former is only necessary when using the new `embed` action, since it doesn't have separate `scope` and `captures` for the `escape` pattern. I'll just change those over in a subsequent commit.

Issue #1358, opened 2017-12-17
Misc. bash/shell issues, some RFEs


okdana:
Hey there. I've been collecting some cases of erroneous or inconsistent behaviour in the new shell syntax definition. I've also included some RFEs and zsh-compatibility issues near the bottom, if you're interested. (I'd understand if zsh compatibility is not a major concern — most of the stuff i mention here is low-hanging fruit, though, for whatever that's worth.)

Edit: I've broken the zsh-specific stuff out into a separate block so it's easier to distinguish it from stuff that affects bash too/only.

```sh
#!/usr/bin/env bash

# Unescaped hash inside parameter expansion breaks file
: ${foo# #} # <- second # interpreted as comment
} # fix highlight

# Open-bracket in glob/regex character class breaks file
[[ a == [abc[]* ]] ]] # <- remove extra brackets to break file
: ${foo//[abc[]/x} ]] } # <- remove extra brackets/brace to break file

# Escaped space in glob/regex character class breaks file
if [[ ' foobar' == [\ ]foo* ]]; then ]] # <- remove extra brackets to break file
  :
fi

# `!` operator only highlighted immediately after `if`
if ! foo || ! bar; then :; fi
while ! foo; do :; done
! false && foo

# Comment not highlighted as such following redirect to fd
foo >&2 # comment

# Closing sub-shell parenthesis not highlighted as such when immediately
# following a variable assignment
(foo=bar)

# Function definitions are not highlighted as such inside command substitutions,
# and the closing `)` is interpreted as terminating the substitution
func() {
} # OK
foo=$(
  func() {
  } # broken
)

# Redirects not highlighted properly after variable assignment (legal in both
# bash and zsh)
LC_ALL=C 2> /dev/null

# Compound commands after `coproc` key word not highlighted as such (braces
# highlighted as brace expansion)
coproc { foo bar }
coproc ( foo bar )

# Unquoted array and assocation values not highlighted consistently with scalar
# values during assignment
scalar=foo
array=( foo bar )

# Association keys and punctuation not highlighted consistently during
# assignment
assoc[foo]=bar
assoc=( [foo]=bar )

# Unquoted array and association members highlighted only as 'function call'
# when assignment is performed as part of a call to `declare`, &al.
declare array=( foo bar baz )
declare assoc=( [foo]=bar )

# Variable names given to `unset` are not highlighted as such (note that the
# presence of `-f` signifies that the names belong to functions rather than
[...]

PR #1369, opened 2017-12-27
[ShellScript] Refactor redirection handling

rwols:
- Redirections may occur *before* a command. For example: `2>&1 echo foo`.
- Redirections may occur *after* a variable assignment. For example: `LC_ALL=C 2> /dev/null`.

Related (but doesn't close): https://github.com/sublimehq/Packages/issues/1358


wbond:
Thanks!

PR #1373, opened 2017-12-29
Handle markdown footnotes

Manishearth:
They currently render with the red background (as an error) because footnotes with any more than one word in them are not valid "link references".

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -222,6 +222,18 @@ contexts:
     - match: |-
         (?x:
             \s*                        # Leading whitespace
+            (\[)\^(.+?)(\])(:)         # Reference name
+            (.*)                       # The footnote
+        )
+      captures:
+        1: punctuation.definition.constant.begin.markdown
+        2: entity.name.reference.link.markdown
+        3: punctuation.definition.constant.end.markdown
+        4: punctuation.separator.key-value.markdown
+        5: markup.underline.footnote.markdown
+    - match: |-
+        (?x:
+            \s*                        # Leading whitespace
             (\[)(.+?)(\])(:)           # Reference name
             [ \t]*                     # Optional whitespace
             (?:


keith-hall:
Please can you add some syntax tests for this - details at http://www.sublimetext.com/docs/3/syntax.html#testing.

Manishearth:
Don't really have time for that right now -- this is the patch I used to fix it locally and felt I should share it.

I may have time to fix this up later, but in the meantime feel free to take over.

keith-hall:
Which Markdown flavor are you building the footnote support for? Because it's not a part of CommonMark or the GFM extensions, which this syntax definition is based on. https://stackoverflow.com/q/25579868/4473405
I ask because I want to see how it defines what footnotes should look like.

Manishearth:
It's the one used by Octopress, which is https://github.com/davidfstr/rdiscount I think.

The syntax is basically that `[^something] ` followed by an arbitrary amount of text (before a newline) is a footnote. It's pretty common across markdown usage.

Given that it's nonstandard I don't mind it not being highlighted, but right now it gets highlighted as an error with the distracting pink-background-white-foreground text which really shouldn't happen.

Markdown/Markdown.sublime-syntax
```diff
@@ -219,6 +219,18 @@ contexts:
     - match: '^[ ]{0,3}(?=(?:{{html_tag_open_commonmark}}|{{html_tag_close_commonmark}})\s*$|<{{html_tag_block_end_at_blank_line}})'
       comment: Markdown formatting is disabled inside block level tags and if a complete HTML tag is the only thing on the line.
       push: [disabled-markdown-pop-at-eol, disable-markdown-pop-at-blank-line]
+    - match: |-
+        (?x:
+            \s*                        # Leading whitespace
+            (\[)\^(.+?)(\])(:)         # Reference name
+            (.*)                       # The footnote
```
keith-hall: it looks like footnotes don't have to end at eol, and can span multiple lines and contain block-level elements: https://michelf.ca/projects/php-markdown/extra/#footnotes
also, the `(.+?)` should probably also be changed to match a valid HTML `id` attribute as per the same docs

wbond:
Closing this as @keith-hall added an implementation in #1375 with tests

PR #1374, opened 2017-12-31
[Java] Refactor Java Properties

lastsecondsave:
Another remnant of the TextMate era. The format is really simple but has its own 'hidden features'.

Changes:

1. Property key and value are now scoped as `entity.name.key` and `string.unquoted` respectively. Key's scope is added to symbols' list.
2. Line continuation and escapes are recognized.
3. `tmPreferences` with settings for line comments.
4. Block comments were removed as there is no such thing.


wbond:
Sorry that I didn't end up having time to get this into 3157, but thanks for the work on it!

PR #1375, opened 2018-01-02
[Markdown] support footnotes from PHP Markdown Extra

keith-hall:
supersedes https://github.com/sublimehq/Packages/pull/1373


wbond:
Looks good, thanks!

Issue #1376, opened 2018-01-02
(feature request) highlight LESS in HTML


anovsiradj:
allow using LESS (and another CSS preprocessor) syntax hightight if `<style>` tag attribute `type` value is not `text/css`. for example:
```html
<style type="text/x-less">
...
body {
  #content {
    ...
  }
}
...
</style>
```
thanks!

*(sory for my english)*

Thom1729:
I'm ambivalent about this. It's not a standard HTML feature, and we don't generally like adding nonstandard features to core syntaxes. In this case, there's no LESS syntax in core, so the feature wouldn't even work without a third-party syntax. I think that's a compelling reason not to implement it in core.

On the other hand, I can definitely see that it would be nice if third-party syntaxes could more easily provide extensions like this. I've been thinking about improving the language-determination functionality for `<script>` and `<style>` tags in the core HTML syntax. Improving that mechanism might also make it easier for third-party syntaxes to "hook into" it.

michaelblyons:
Sounds like a job for a third-party package with `extends`.

michaelblyons:
Like #2722, this would require linking an external syntax. Any unit tests would fail. I recommend closure.

deathaxe:
It seems LESS and SCSS are only small extensions compared to CSS. Inheriting them from CSS would be even more straigth forward than the aproache SASS package does at the moment. Maybe we can add them at some point, because they seem to have matured to some kind of defacto standard in web development.

michaelblyons:
> It seems LESS and SCSS are only small extensions compared to CSS. Inheriting them from CSS would be even more straigth forward than the aproache SASS package does at the moment. Maybe we can add them at some point, because they seem to have matured to some kind of defacto standard in web development.

Good points. I'm willing to concede.

PR #1379, opened 2018-01-06
[HTML] Improved language detection in `<script>` and `<style>` tags

Thom1729:
In HTML, `<script>` tags usually contain JavaScript code and `<style>` tags usually contain CSS. However, if the tag's `type` attribute is set to a value other than `text/javascript` or `text/css` (respectively), the tag's contents will not be processed. The default HTML syntax handles this with a lookahead. However, there are several cases where this lookahead will fail:

```html
<script type='other'>
    var x = 0;
</script>

<script
type='text/other'>
    var x = 0;
</script>

<script type=text/javascriptFOO>
    var x = 0;
</script>
```

This PR adds a robust system for detecting attributes that determine the language. As implemented, this should handle all of the corner cases for JavaScript and CSS. In addition, third parties could use a system like YAML Macros to easily extend the core syntax with support for new languages. For instance (#1376), you could add support for [LESS](http://lesscss.org/) with the following YAML Macros patch:

```yaml
!merge
contexts:

  style-type-decider: !prepend
    - match: (?i)(?=text/x-less(?!{{unquoted_attribute_value}})|'text/x-less'|"text/x-less")
      set:
        - style-less
        - tag-generic-attribute-meta
        - tag-generic-attribute-value

  style-less:
    - meta_content_scope: meta.tag.style.begin.html
    - include: style-common
    - match: '>'
      set:
        - include: style-close-tag
        - match: ''
          embed: scope:source.less
          embed_scope: source.less.embedded.html
          escape: (?i)(?=</style)
```

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -83,11 +83,13 @@ contexts:
         - include: tag-attributes
     - match: (<)((?i:style))\b
       captures:
+        0: meta.tag.style.begin.html
         1: punctuation.definition.tag.begin.html
         2: entity.name.tag.style.html
       push: style-css
     - match: '(<)((?i:script))\b'
       captures:
+        0: meta.tag.script.begin.html
         1: punctuation.definition.tag.begin.html
         2: entity.name.tag.script.html
       push: script-javascript
@@ -196,7 +198,7 @@ contexts:
       scope: invalid.illegal.incomplete.html
 
   style-css:
-    - meta_scope: meta.tag.style.begin.html
+    - meta_content_scope: meta.tag.style.begin.html
     - include: style-common
     - match: '>'
       scope: punctuation.definition.tag.end.html
@@ -208,7 +210,7 @@ contexts:
           escape: (?i)(?=(?:-->\s*)?</style)
 
   style-other:
-    - meta_scope: meta.tag.style.begin.html
+    - meta_content_scope: meta.tag.style.begin.html
     - include: style-common
     - match: '>'
       scope: punctuation.definition.tag.end.html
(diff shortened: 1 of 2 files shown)


keith-hall:
if multiple packages wanted to add their own language syntax to HTML using YAMLMacros, how do you envision that working, @Thom1729? Would each package overwrite the `Packages/HTML/HTML.sublime-syntax` file so they could build on top of each other? But then they have to detect their own changes so as not to apply them twice, and it'd be hard to keep it up to date when the syntax changes in this repo I guess... But if they keep a local copy of the HTML syntax in their package, with their changes applied, users would have to choose which syntax to use - the one with LESS or the one with SCSS, for example, with no possibility to have both active at once? Sorry if this is a bit off topic here, please feel free to move this discussion to a YAMLMacros issue if it would be more appropriate...

Slightly more on topic, sorry for being picky bit please could you include the package (`[HTML]`) prefix in the commit titles in future, it makes it easier to see what was changed when looking purely at the commit history. :)

Thom1729:
That's a good question, and one to which I do not have a complete answer. [JS Custom](https://github.com/Thom1729/Sublime-JS-Custom/) is in some sense a partial answer — it's a general engine for selectively patching a syntax based on user preferences, and it just happens that the provided syntax and patches are for JavaScript. There's not yet a mechanism for users or third-party packages to provide their own patches, but I think that's a no-brainer feature to add once there's a stable public release.

In the case of HTML, the same problem could be solved in the same way. An “HTML Custom” package could allow the user to specify any number of custom languages to embed. Third-party packages could offer their own patches for the engine to use.

The reason I feel this answer is incomplete is that such an implementation would require an additional package per language. Sure, much of the code could be factored out into a common dependency, but there would still be substantial overhead. It's possible that a single package could provide this behavior for all languages with extensibility hooks, but there are some challenges there.

For instance, patching a syntax outside your control (such as a core syntax) means tying your patch to internal implementation details that may change. JS Custom addresses this by keeping a private copy of the core JavaScript syntax. Whenever the core JavaScript syntax is updated, I will pull it into JS Custom and release a new version myself. This is perfectly manageable for one syntax. For *every* syntax it would be impractical. For a single generic package that offered to patch all of the core syntaxes, I'm not sure where you'd start.

Perhaps if one day some sort of macro system were incorporated into core, a better approach might present itself. For now, I'm content to get JS Custom out there and get some real experience with the system as it stands. Once it's been out in the wild for a couple of months I'll have a better sense of what parts work well and what parts could be improved.

---

I should probably squash down a lot of these commits; I could put a package prefix on them at that time.

wbond:
LGTM, thanks!

Issue #1377, opened 2018-01-02
[Python] Comprehensions split by newline


michaelblyons:
```py
[foo for foo in bar]
#            ^^ keyword.control.flow.for.in.python
[foo for foo
 in bar]
#^^ keyword.operator.logical.python
```
The second `in` has the wrong scope. Also, the `meta.expression.generator` terminates early.

PR #1380, opened 2018-01-06
[Python] Fixed a couple of newline-related bugs.

### LINKED ISSUE #1367: [Python] parameter unpacking token not recognized on separate line
```
def foo(a, b, *c):
    #         ^ keyword.operator.unpacking
    pass

def bar(a,
        b,
        *c):
    #   ^ missing keyword.operator.unpacking
    pass
```

Thom1729:
Should fix #1367 and #1377. Tests added.

The Python syntax should probably have comments in the prototype rather than including it manually. I ran into a comment bug while working on this, and there are probably more lurking. It's much easier to put `meta_include_prototype: false` in a handful of places than `include: comments` in dozens of places.

There was a bit of invalid Python in a test, and fixing this broke the following test. I changed the first test to avoid the problem. It's possible that there's some special error behavior that the first test is trying to trigger; someone more familiar with the Python syntax definition might take a look.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -515,7 +515,7 @@ contexts:
     - match: \*
       scope: keyword.operator.unpacking.sequence.python
       pop: true
-    - match: (?!\s|#)
+    - match: (?=\S)
       pop: true
 
   classes:
@@ -1765,6 +1765,7 @@ contexts:
         1: storage.modifier.async.python
         2: keyword.control.flow.for.generator.python
       push:
+        - include: comments
         - meta_scope: meta.expression.generator.python
         - match: \b(in)\b
           scope: keyword.control.flow.for.in.python
@@ -1782,7 +1783,6 @@ contexts:
       scope: keyword.control.flow.else.inline.python
 
   target-list:
-    - include: line-continuation-or-pop
     - match: ','
       scope: punctuation.separator.target-list.python
     - match: \(
--- a/Python/syntax_test_python.py
+++ b/Python/syntax_test_python.py
@@ -266,7 +266,7 @@ def _():
 #                       ^ invalid.illegal.expected-parameter.python
 #                            ^ invalid.illegal.expected-parameter.python
 
-    lambda
+    lambda x
 #   ^^^^^^ storage.type.function.inline
 
     ( 3 - 6 \
@@ -578,6 +578,16 @@ def func(*args, other_arg=2**10, **kwargs):
 #                                ^^ keyword.operator.unpacking.mapping.python
     pass
 
+def func(
+    *args,
+#   ^ keyword.operator.unpacking.sequence
+    other_arg=2**10,
+#              ^^ keyword.operator.arithmetic
+    **kwargs
+#   ^^ keyword.operator.unpacking.mapping
+):
+    pass
+
 
 ##################
 # Class definitions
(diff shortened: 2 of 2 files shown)


Python/syntax_test_python.py
```diff
@@ -266,7 +266,7 @@ def _():
 #                       ^ invalid.illegal.expected-parameter.python
 #                            ^ invalid.illegal.expected-parameter.python
 
-    lambda
+    lambda x
```
FichteFoll: I suppose you were talking about this invalid line breaking the tests? It's still invalid, but the bare keyword is used specifically for bail-out during typing. The lambda context should be exited if a newline without a backslash follows.

Python/Python.sublime-syntax
```diff
@@ -1782,7 +1783,6 @@ contexts:
       scope: keyword.control.flow.else.inline.python
 
   target-list:
-    - include: line-continuation-or-pop
```
FichteFoll: And this is likely the change that cause the test breakage.

It would be better to change the usage of it in the inline for context because removing the include here will break other cases (that aren't being tested currently, it seems).

https://github.com/Thom1729/Packages/blob/4a1543f14949593fbde7553a7361543edc92f0a1/Python/Python.sublime-syntax#L1777

Proposal: outsource all items from `target-list` but this one exclude into a `target-list-inline` context and include that from the inline for context.

Thom1729: It looks to me like the cases that need `line-continuation-or-pop` (in `block-statements`) are including it manually. The only place that doesn't is `inline-for`, which doesn't want it. I believe the lookahead fix in `allow-unpack-operators` led to the test failure.

FichteFoll: Looks like it indeed. This change is fine then.

Python/syntax_test_python.py
Thom1729: I agree that the current failure mode is pretty poor. We'd run into a similar problem here:

```python
lambda foo,
42
```

But we can't simply bail on a newline because the lambda could be inside parens:

```python
(lambda
foo
:0)
```

In order to handle errors gracefully, we might need duplicate implementations of lambdas inside and outside parens. I hope there's a better solution than that.

FichteFoll: Huh, I didn't even know this was possible.

It's pretty much an anti-pattern though. If you end up in a position where you'd look into multi-lining the arguments to a `lambda` function, you probably should be using `def` already. I've never seen this before and I don't think it's likely to pop up, so … I guess you're free to implement two lambda versions, but I doubt I'll bother.

Thom1729: I had to try it out to be sure. Python grammar is hard (and, in principle, impossible to parse in a Sublime syntax).

Anyway, now that I think about it, when would you ever write a lambda outside parens in the first place? As long as the behavior in such a case is technically correct, it probably isn't a big deal if it's a bit ungraceful (certainly next to the hassle of duplicate code).

FichteFoll: Good point. Now that PEP8 (and tooling) actively discourages you from assigning lambdas (iirc it didn't do that a few years ago), you hardly ever write lambdas outside of parens and when you do, you probably don't care about 100% accutate highlighting either.

I guess we can safely ignore the line-splitting case for incorrect grammar and just bail out as soon as we find an unrecognized token. Should put a comment with this decision somewhere around the lambda match.

FichteFoll:
Regarding prototype contexts: I usually prefer the explicit way of stating exactly where and when I expect a comment to be matched, although this is mostly a principal thing and doesn't matter in reality aside from performance as the situations where you will never expect a comment are also those that you never end up being in as the comment would have been matched at a different place already. In reality, I'm quite indifferent and just inherit whatever was used by the syntax before, which is why the Python is still this way.

If you end up adding a protoype context in a different PR, I'd be interested in benchmarks though.

Thom1729:
I haven't tested this, but I'm pretty sure that prototypes should be perfectly performance-neutral (compared to manually including contexts). They're baked in when the syntax is compiled to a giant state machine, so it shouldn't make any difference. I'll be sure to test this if I try such a change for the Python syntax.

PR #1383, opened 2018-01-07
[ShellScript] Fix: regex set handling

rwols:
Related: https://github.com/sublimehq/Packages/issues/1358

These constructs should not break highlighting anymore:
```bash
[[ a == [abc[]* ]]
: ${foo//[abc[]/x}

if [[ ' foobar' == [\ ]foo* ]]; then
  :
fi
```


wbond:
Thanks for all of these ShellScript fixes!

PR #1384, opened 2018-01-08
[Rust] add punctuation accessor dot scope

keith-hall:


FichteFoll:
Don't you think "as" should get a storage scope?

wbond:
@FichteFoll The `as` keyword already gets `keyword.operator` in another place in the syntax, so this is just a consistency thing, as far as I can tell.

FichteFoll:
Fair enough. What are your options on it though? It's a cast of primitives, fwiw.

PR #1386, opened 2018-01-11
[Regular Expressions] enable toggle block comment support

keith-hall:
This PR adds the required `tmPreferences` metadata file to enable support for the "toggle block comment" functionality in ST. Maybe in future we will want to get fancy and allow line comments to be toggled when in extended mode, but this PR simply uses the `(?#...)` block style comments in all situations.


wbond:
Sounds nifty, thanks!

PR #1389, opened 2018-01-16
[HTML] Updated comment errors for HTML5.

Thom1729:
As noted [on the forum](https://forum.sublimetext.com/t/html-comment-false-highlighting/34605), the HTML syntax definition currently marks strings of hyphens inside comments as errors. This is congruent with the [HTML4](https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.4) spec. However, the [HTML5](https://www.w3.org/TR/html5/syntax.html#comments) spec is more lenient. This PR updates the HTML syntax definition and tests to use the HTML5 rules.


wbond:
Thanks!

PR #1391, opened 2018-01-16
[HTML] Improved scoping for entities and fixed uppercase hex entities.

Thom1729:
With this change, the leading `#` and/or `x` will be scoped as punctuation. This conforms with the way numeric character escapes work in other syntaxes. In addition, hex entities may now use an uppercase `X`.

Tests are updated, including a new test ensuring that named escapes only contain ASCII characters (as [specified](https://html.spec.whatwg.org/#character-references)).


HTML/HTML.sublime-syntax
```diff
@@ -237,8 +237,18 @@ contexts:
...
+      scope: constant.character.entity.hexadecimal.html
+      captures:
+        1: punctuation.definition.entity.html
+        3: punctuation.definition.entity.html
+    - match: (?i)(&#)([0-9]+)(;)
+      scope: constant.character.entity.decimal.html
+      captures:
+        1: punctuation.definition.entity.html
+        3: punctuation.definition.entity.html
+    - match: (?i)(&)([a-zA-Z0-9]+)(;)
```
michaelblyons: If this is case-insensitive, why put both upper and lower A-Z?

Thom1729: I went back and forth on this a couple of times, and I decided to strip the `(?i)` flags. The new expressions are a bit more explicit.

Thom1729:
I've pushed another change. In HTML5, loose ampersands are no longer illegal. Ampersands are only illegal when part of an invalid named entity and even then only in some contexts.

The only way to know whether a named entity is invalid is with a [giant list](https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references).

Issue #1394, opened 2018-01-21
[Lua] match "end" keyword BUG


RobinCodeX:
![123](https://user-images.githubusercontent.com/8408783/35192450-6fd7f6ac-fecd-11e7-9e13-d5c284c8bd34.gif)

This `Spend()` is matched to `end`

keith-hall:
The problem is with the `increaseIndentPattern` regex [here](https://github.com/sublimehq/Packages/blob/61bfd68ea07a3aede68cd6a65c80be702b3d6db8/Lua/Indent.tmPreferences#L13). PHP used to have similar problems: https://github.com/sublimehq/Packages/issues/1175

PR #1395, opened 2018-01-22
[Lua] fix increaseIndentPattern with end keyword

RobinCodeX:
Fix my issue #1394 
This is my first pull request and I hope it can help you😄

### DIFF
--- a/Lua/Indent.tmPreferences
+++ b/Lua/Indent.tmPreferences
@@ -10,7 +10,7 @@
 		<key>decreaseIndentPattern</key>
 		<string>^\s*(elseif|else|end|\})\s*$</string>
 		<key>increaseIndentPattern</key>
-		<string>^\s*(else|elseif|for|(local\s+)?function|if|repeat|until|while)\b((?!end).)*$|\{\s*$</string>
+		<string>^\s*(else|elseif|for|(local\s+)?function|if|repeat|until|while)\b((?!\send).)*$|\{\s*$</string>
 	</dict>
 </dict>
 </plist>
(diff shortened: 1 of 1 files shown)


Lua/Indent.tmPreferences
```diff
@@ -10,7 +10,7 @@
 		<key>decreaseIndentPattern</key>
 		<string>^\s*(elseif|else|end|\})\s*$</string>
 		<key>increaseIndentPattern</key>
-		<string>^\s*(else|elseif|for|(local\s+)?function|if|repeat|until|while)\b((?!end).)*$|\{\s*$</string>
+		<string>^\s*(else|elseif|for|(local\s+)?function|if|repeat|until|while)\b((?!\send).)*$|\{\s*$</string>
```
FichteFoll: I suggest `(?!.*\bend\b)` instead of `((?!\send).)*$`. It is more performant and checks for non-word symbols on both ends of the keyword.

Note that it will still errorneously match inside strings, but this should get rid of all cases not involving strings.

FichteFoll:
Well, this version is even less efficient than the original.

RobinCodeX:
@FichteFoll 😄 What do you think of  `(?!\bend\b)`

FichteFoll:
That fixes the issue at least. I'm not sure why you didn't like my proposed change, but maybe it doesn't work, though I don't see why it shouldn't. Did you try it?

RobinCodeX:
@FichteFoll Yes, I've tried all.
I'm sorry, I misunderstand your meaning, I changed back. 😂

FichteFoll:
It seems you're misreading my proposal. Please compare exactly what I mentioned to replace and what to replace it with.

PR #1396, opened 2018-01-23
[Java] Fix multiple const assignmnet

ismell:
Fixes part of #1146


lastsecondsave:
This will conflict with #1366, I have similar fix as well.

PR #1401, opened 2018-01-29
[R] fix bug in detecting function parameters

randy3k:
In some cases, parameters are wrongly detected as `entity.name`


wbond:
Thanks!

PR #1402, opened 2018-01-29
[C#] support value tuples in generic types

keith-hall:
this PR adds support for `ValueTuples` in generic type declarations - prior to this, tuples were scoped correctly as part of normal method declarations but not when the method is declared as a generic `Func`.
More info on ValueTuples can be found at https://blogs.msdn.microsoft.com/mazhou/2017/05/26/c-7-series-part-1-value-tuples/.


wbond:
Thank you, as always!

PR #1407, opened 2018-02-02
[HTML] Add the onauxclick event attribute

KwanEsq:
auxclick is new event, equivalent to click but for non-primary mouse buttons. [Spec](https://w3c.github.io/uievents/#event-type-auxclick).
Implemented in at least Blink and Gecko.

I didn't add a syntax test for this on the basis that it wouldn't add anything over the [existing event attribute test](https://github.com/sublimehq/Packages/blob/e0759e08f54b51a436f589c51b836ac3ae5894b1/HTML/syntax_test_html.html#L213), but if you disagree I'm more than happy to write one.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -429,7 +429,7 @@ contexts:
   tag-event-attribute:
     - match: |-
         (?x)\bon(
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #1417, opened 2018-02-10
[PHP] fix docblock comments that transition to normal comments

### LINKED ISSUE #1378: [PHP] Syntax Highlight issue for multi-line asterisk comments
When creating a comment in PHP with this syntax, the highlighter "breaks" rest of the file.

***Not working***
```<?php
/**
 * PHP comment
 *
 * @see
 * @author
 * @package
 * @version
 NOTE: Modified */
```  

The issue lies in the last sentence **NOTE: Modified** which does not begins with a asterisk (*) and thus breaks the syntax highlight.

***Working***
```<?php
/**
 * PHP comment
 *
 * @see
 * @author
 * @package
 * @version
 * NOTE: Modified */
```

keith-hall:
fixes #1378

### DIFF
--- a/PHP/syntax_test_php.php
+++ b/PHP/syntax_test_php.php
@@ -386,6 +386,18 @@ $var->meth()[10];
 //  ^ keyword.other.phpdoc
  */
 
+/**
+ * PHP comment from issue #1378
+ *
+ * @see
+ * @author
+ * @package
+ * @version
+ NOTE: Modified */
+//^^^^^^^^^^^^^^^^ comment.block - comment.block.documentation
+//              ^^ punctuation.definition.comment.end
+//                ^ - comment
+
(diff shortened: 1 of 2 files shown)


PHP/syntax_test_php.php
```diff
@@ -386,6 +386,18 @@ function i(
...
  */
 
+/**
+ * PHP comment from issue #1378
+ *
+ * @see
+ * @author
+ * @package
+ * @version
+ NOTE: Modified */
```
rwols: Funny how GitHub fails to highlight this correctly too.

wbond: This is actually somewhat common as both GitHub and Sublime started from the same source syntax definitions from TextMate.

PR #1418, opened 2018-02-10
[Python] Fix stray brackets in with statements

### LINKED ISSUE #1397: [Python] meta.statement.with.python invalid.illegal.stray.brace.round.python
That parenthesis should not have the scope `meta.statement.with.python invalid.illegal.stray.brace.round.python`:

image

```python
    def test_function_name(self):
        with captured_output() as (out, err):
```

Sublime Text build 3156

deathaxe:
Fixes #1397

### DIFF
--- a/Python/syntax_test_python.py
+++ b/Python/syntax_test_python.py
@@ -424,6 +424,57 @@ def _():
 #                           ^ punctuation.separator.with-resources
 #                             ^^^^ support.function
 #                                    ^^ keyword.control.flow.with.as
+
+    with captured() as (out, err):
+#   ^^^^ keyword.control.flow.with
+#        ^^^^^^^^ variable.function
+#                ^ punctuation.section.arguments.begin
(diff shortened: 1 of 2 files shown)


Python/syntax_test_python.py
```diff
@@ -424,6 +424,57 @@ def _():
...
+        out,
+#       ^^^ meta.generic-name
+#          ^ punctuation.separator.tuple
+        err
+#       ^^^ meta.generic-name
+    ):
+#   ^ punctuation.section.group.end
+#    ^ punctuation.section.block.with
+
+    with captured() as [out, err]:
```
FichteFoll: Is this even valid? I've never seen a destructoring assignment with brackets.

deathaxe:
I didn't see that too, but python doesn't complain about it. Works the same way as with using tuples. I guess it's because python doesn't differentiate between tuples and lists in many cases.

wbond:
Thanks!

PR #1420, opened 2018-02-10
[ShellScript] Fix: empty array assignment regression

rwols:
Some interplay between the last few merges caused a regression where an empty array assignment is incorrectly scoped as the start of a function definition. The fix is one extra byte in the `is_function` lookahead.

```bash
#!/bin/bash
my_array=()
```


wbond:
Thanks!

PR #1421, opened 2018-02-10
[HTML] Should handle all HTML5 attribute names.

### LINKED ISSUE #1419: Angular event bindings in .html files
Angular event bindings may give wrong syntax coloring. This normally happens when there is an equal sign nested in the event bind. 

Example:
`<ngb-alert (close)="message=null" >{{ message }}</ngb-alert>`

Everything after `null` will have string syntax coloring.

screen shot 2018-02-10 at 11 20 46 am

Thom1729:
Fixes #1419. Also removes an unreachable rule.

Spec: https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -13,7 +13,7 @@ first_line_match: (?i)<(!DOCTYPE\s*)?html
 scope: text.html.basic
 
 variables:
-  attribute_name: (?:[^ "'>/=]+)
+  attribute_name: (?:[^ "'>/=\x00-\x1f\x7f-\x9f]+)
   unquoted_attribute_value: (?:[^\s<>/''"]|/(?!>))+
   not_equals_lookahead: (?=\s*[^\s=])
 
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -13,6 +13,7 @@ first_line_match: (?i)<(!DOCTYPE\s*)?html
 scope: text.html.basic
 
 variables:
+  attribute_name: (?:[^ "'>/=]+)
   unquoted_attribute_value: (?:[^\s<>/''"]|/(?!>))+
   not_equals_lookahead: (?=\s*[^\s=])
 
@@ -268,14 +269,11 @@ contexts:
         - include: entities
 
   tag-generic-attribute:
-    - match: '[a-zA-Z0-9:\-_.]+'
+    - match: '{{attribute_name}}'
       scope: entity.other.attribute-name.html
       push:
         - tag-generic-attribute-meta
         - tag-generic-attribute-equals
-
-    - match: '[a-zA-Z0-9:\-_.]+'
-      scope: entity.other.attribute-name.html
 
   tag-generic-attribute-meta:
     - meta_scope: meta.attribute-with-value.html
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -305,6 +305,7 @@ contexts:
         - include: attribute-entities
     - match: '{{unquoted_attribute_value}}'
       scope: string.unquoted.html
+      pop: true
     - include: else-pop
 
   tag-class-attribute:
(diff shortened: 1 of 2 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -13,6 +13,7 @@ first_line_match: (?i)<(!DOCTYPE\s*)?html
 scope: text.html.basic
 
 variables:
+  attribute_name: (?:[^ "'>/=]+)
```
keith-hall: The HTML5 validators I've tried:

- https://validator.w3.org/nu/#textarea
- https://html5.validator.nu/

seem to suggest that we should add `\t` to this because:

```html
<!DOCTYPE html>
<html>
<head>
<title>Test</title>
</head>
<body>
<p	class	id	=	hi i	d	=	"ok"></p>
<!-- all the whitespace inside the above `p` tag are tab characters -->
</body>
</html>
```

results in:

```
Error: Attribute i not allowed on element p at this point.
Error: Attribute d not allowed on element p at this point.
```

meaning that it doesn't see an attribute called `class\tid`, it sees `class` with no value and `id` with value `hi`, so scoping the tab character as `entity.other.attribute-name.html` seems wrong. Related to this, it should pop [when an unquoted value is encountered](https://github.com/sublimehq/Packages/blob/b85a691be63ca8daf4873a404f7d16d255fe5613/HTML/HTML.sublime-syntax#L308-L309), otherwise it incorrectly marks the rest of the tag contents as an unquoted string or at least until it reaches a quoted string.

Thom1729: You're right — the tab, line feed, and form feed characters are listed in the [parsing algorithm](https://html.spec.whatwg.org/multipage/parsing.html#before-attribute-name-state). No idea why they're omitted in the other part of the spec.

willrowe:
This has caused a template syntax I have to stop working when a template tag is at the end of an HTML attribute like so: `<input {% if not is_enabled %} disabled{% endif %} name="user_name" type="text" />`
After this PR, it applies the scope `meta.attribute-with-value.html` and `entity.other.attribute-name.html` overriding the scopes that should applied to the `{% endif %} tag.

wbond:
@willrowe If you are using a templating syntax, I wouldn't expect the HTML syntax to handle that. I would recommend looking at Package Control for a syntax built for the templating language you are using.

willrowe:
This is a syntax that I have built which leverages embeds, so it uses the HTML syntax unless it encounters the template tags. It is not handling this correctly when next to or part of an attribute name anymore though. This does not work either `<div data-{{ custom_attribute_suffix }}></div>`.

wbond:
@willrowe It sound like you'll have to work on changing how your syntax is implemented since the HTML syntax has changed. Most of @Thom1729's work on HTML and JavaScript seems to be motivated by making it more flexible, since he has some tools he is working on that allows people to write macros to extend syntaxes.

If you want advice on your syntax, I recommend going to the pinned topic on the forum and following the link to Discord. Most of the most active community members hang out there and would likely be happy to give some advice or help out.

willrowe:
I'm not sure that I would be able to implement it differently in order for it to work as expected. No matter what I do, the regex for the attribute name would still be there and match the template tag syntax, preventing any syntax definition I create from then matching against it. Since it matches the entire attribute name using the regex instead of pushing and then popping at the end of the attribute name, it has no chance check for the template tag regexes I have in my syntax definition.

I expect this is also affecting other template syntaxes that are embedding using the core HTML syntax. I do not believe that the [Jinja example](https://www.sublimetext.com/docs/3/syntax.html#include-syntax) given in the docs would work either in this case.

wbond:
I'm not going to continue commenting on an a closed PR. You can move this to the forum if you wan to discuss some more, or ask questions on Discord.

Thom1729:
@willrowe The problem here is that `disabled{%` is a perfectly valid HTML5 attribute name. In a previous version of the core HTML syntax, it would not have been recognized. This PR corrected that bug. What you (or the author of the syntax you're using) probably need to do is create a modified version of the core HTML syntax with more stringent rules for attribute names. I can give you a hand with this if you message me on Discord. (I would also prefer not to have this conversation on a closed issue.)

PR #1425, opened 2018-02-12
[C#] Index properties

michaelblyons:
Add class properties to <kbd>cmd</kbd>+<kbd>R</kbd>, and add `(…)` to functions like Python does.

### DIFF
--- a/C#/Symbol List - Inner Function.tmPreferences
+++ b/C#/Symbol List - Inner Function.tmPreferences
@@ -2,7 +2,7 @@
 <plist version="1.0">
 <dict>
 	<key>scope</key>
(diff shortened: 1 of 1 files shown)


FichteFoll:
Related: https://github.com/SublimeTextIssues/Core/issues/1676

wbond:
I think we should drop the `(...)` addition for functions and then this should be good to merge. We've got some things in the queue for disambiguating symbols in the symbol list.

michaelblyons:
Since the two files are basically the same now, I merged them. I'm presuming here that the disambig stuff you have in the backend only cares about the scope names, not the files that they are indicated by.

If it does matter, LMK and I'll keep both files.

michaelblyons:
After poking around the new hotness for a bit, I think we want to keep them separate. 😉

wbond:
Thanks for this!

Issue #1324, opened 2017-11-23
[Ruby] Invalid String#match? highlighting


ilnurnasyrov:
![image](https://user-images.githubusercontent.com/3808127/33668958-3209f90e-dab2-11e7-987f-f59863c700d1.png)

pRdm:
I believe the issue comes from line `269` in `Ruby.sublime-syntax`:

```yaml
- match: '\?(\\(x\h{1,2}\b|0[0-7]{0,2}\b|[^x0MC]\b)|(\\[MC]-)+\w\b|[a-zA-Z0-9_]\b(?!\s*:)|[^a-zA-Z0-9_\s\\])'
```

adding a negative lookahead `(?!\()` fixes the issue but not sure if it introduces any other bugs:
```yaml
- match: '\?(?!\()(\\(x\h{1,2}\b|0[0-7]{0,2}\b|[^x0MC]\b)|(\\[MC]-)+\w\b|[a-zA-Z0-9_]\b(?!\s*:)|[^a-zA-Z0-9_\s\\])'
```

ilnurnasyrov:
I believe problem is there:
```
    # Methods that may be followed by a regex
    - match: |-
        (?x:
          \b(
            gsub!|
            sub!
          )(?=\s)
          |
          \b(
            assert_match|
            gsub|
            index|
            match|
            scan|
            sub
          )\b
        )
```
I just don't known how to solves problem with invalid highlighting. Removing `match|` solves this problem, but `match?` is not highlighted in this case. Replacing `match|` with `match\?|` doesn't highlight `match?` correctly.

michaelblyons:
@ilnurnasyrov The `\?` messes with the `\b` at the bottom of the group. Maybe this?

``` yaml
    # Methods that may be followed by a regex
    - match: |-
        (?x:
          \b(
            gsub!|
            sub!
          )(?=\s)
          |
          \b(
            assert_match|
            gsub|
            index|
            match|
            scan|
            sub
          )\b\??    # put the optional `?` here?
        )
```

I don't use Ruby, so I don't know what the `?` is for, or if it needs to be scoped separately.


Also, here's the snippet, for future tests:

``` ruby
# correct highighting
def qwe; end
'qweqweqwe'.matches?(/qwe/)
def qwe; end

# incorrect highlighting
def qwe; end
'qweqweqwe'.match?(/qwe/)
def qwe; end
```

FichteFoll:
[...]

PR #1427, opened 2018-02-14
[JavaScript] Improved identifier and number lexing.

Thom1729:
Implemented the identifier rules from the spec.

The syntax previously used `\b` in a lot of rules. This is not quite right, and it leads to some bugs:

```js
function$NotAFunction
```

Where `\b` followed identifier characters, I replaced it with a negative lookahead. (Testing showed that there was no effect on performance.) Where `\b` *preceded* identifier characters, we would need a lookbehind, which the fast regexp engine does not support. However, we don't actually need assertions at the beginning of matches if we maintain the invariant that the "cursor" should never pause with identifier characters on both sides. This is guaranteed if every match that ends with an identifier character ends with the negative lookahead.

Implementing this correctly required changing the implementation of numeric literals. I added more tests and improved the behavior of numbers in some corner cases.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -886,7 +886,7 @@ contexts:
       scope: keyword.operator.arithmetic.js
     - match: new{{identifier_break}}
       scope: keyword.operator.word.new.js
-    - match: \b(?:delete|typeof|void)\b
+    - match: (?:delete|typeof|void){{identifier_break}}
       scope: keyword.operator.js
 
   binary-operators:
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -8,12 +8,16 @@ file_extensions:
 first_line_match: ^#!\s*/.*\b(node|js)\b
 scope: source.js
 variables:
-  identifier: '[_$[:alpha:]][_$[:alnum:]]*'
-  constant_identifier: '[[:upper:]][_$[:digit:][:upper:]]*\b'
-  dollar_only_identifier: '\$(?![_$[:alnum:]])'
-  dollar_identifier: '(\$)[_$[:alnum:]]+'
-  func_lookahead: '\s*\b(async\s+)?function\b'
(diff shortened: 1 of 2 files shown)


michaelblyons:
Ruby [may need](https://github.com/sublimehq/Packages/issues/1324#issuecomment-369667473) this sort of thing, too.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -8,12 +8,16 @@ file_extensions:
...
-  func_lookahead: '\s*\b(async\s+)?function\b'
-  arrow_func_lookahead: '\s*(\basync\s*)?([_$[:alpha:]][_$[:alnum:]]*|\(([^()]|\([^()]*\))*\))\s*=>'
+  identifier_start: '[_$\p{L}\p{Nl}]'
+  identifier_part: '[_$\p{L}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}\x{200C}\x{200D}]'
+  identifier_break: (?!{{identifier_part}})
+
+  identifier: '{{identifier_start}}{{identifier_part}}*{{identifier_break}}'
+  constant_identifier: '[[:upper:]]{{identifier_part}}*{{identifier_break}}'
+  dollar_only_identifier: '\${{identifier_break}}'
+  dollar_identifier: '(\$){{identifier_part}}*{{identifier_break}}+'
```
sharkdp: @Thom1729 This `dollar_identifier` regular expression seems to cause errors when the JavaScript language definition is used within [`syntect`](https://github.com/trishume/syntect/).

In particular, it uses `+` on `{{identifier_break}}`, which is a negative-lookahead group: `(?!...)+`. Is this valid regex syntax? Could/should the `+` be removed?

Thom1729: It's valid but redundant. I don't know why syntect doesn't handle this, but if you open up a PR to remove the `+` I'll endorse it.

robinst: (for anyone arriving here, @sharkdp raised a PR: https://github.com/sublimehq/Packages/pull/1525)

Issue #1248, opened 2017-10-18
[Shell] Clear string scope in "interpolated subcommands"


FichteFoll:
The new shell script syntax uses `string` even within interpolated subcommands, i.e. `$(xsel -ob)`. It would be nice if it cleared the string scope for the contents of these. Same for backticks, probably.

See also the corresponding RFC in #715 with an example from how Python does it.

cc @rwols

FichteFoll:
It would also be nice if it didn't use `punctuation.definition.variable` for the `$(` and `)` and instead one of the interpolation punctuation scopes (I have a selector for that in my color scheme 😉).

rwols:
I have my mind on bugfix-mode for the time being but feel free to make the changes.

If you use `clear_scopes` during an interpolation, how would you handle:
```bash
ECHO=echo
$ECHO 'hello, world!'
```
Here, the `$ECHO` still gets a `variable.function` scope, but it would disappear when using `clear_scopes`.

FichteFoll:
`clear_scopes` would only be used to clear the last `string` scope from the stack. Within the interpolation expression everything would then be colored as normal.

I haven't looked at the bash syntax yet to check how involved this change would be, but it works really great in Python and JavaScript.

michaelblyons:
All you need is a `clear_scopes: 1` in the two `push`es in `expansion-command`.

You might also want to add a `meta_content_scope` with a non-`meta.` scope, so that color schemes can "officially" target it. 😉  (In my test, I'm re-using the root scope of `source.shell.bash` to trigger the `source source` coloring.)

rwols:
> All you need is a clear_scopes: 1 in the two pushes in expansion-command.

OK, I can see how this can be beneficial. Feel free to make the PR 👍

michaelblyons:
Now I'm actually going to walk that back a little. There are indeed some context-dependent weirdnesses.

``` sh
# works good
foo=$(echo "bar")
echo "$(echo "bar")"

# accidentally pops `meta.function-call.arguments`
echo $(echo "bar")
```

I also noticed a minor bug while I was testing:

``` sh
foo=bar
#   ^^^ string.unquoted
foi="baz"
#   ^^^^^ string.unquoted string.quoted.double
```

rwols:
Your last example is intended. Though I guess it may look a bit weird. The syntax uses a `meta_scope: string.unquoted` after the equals-sign. Because e.g.

```
foo="b"'a'z
```
[...]

PR #1432, opened 2018-02-16
[C#] meta.string scope for interpolated verbatim strings

michaelblyons:
To match scopes for normal interpolated strings

Needed for completeness on facelessuser/BracketHighlighter#468

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1590,7 +1590,7 @@ contexts:
 
   long_format_string:
     - meta_include_prototype: false
-    - meta_scope: string.quoted.double.interpolated.cs
+    - meta_scope: meta.string.interpolated.cs string.quoted.double.raw.cs
     - match: '""'
       scope: constant.character.escape.cs
     - match: '"'
--- a/C#/tests/syntax_test_Strings.cs
+++ b/C#/tests/syntax_test_Strings.cs
@@ -26,6 +26,30 @@
 ///                                        ^^ constant.character.escape.cs
 ///                                           ^^ constant.character.escape.cs
 
+var literal = "foo";
+///           ^^^^^ string.quoted.double
+var interpolated_none = $"foo";
+///                     ^^^^^^ meta.string.interpolated.cs string.quoted.double.cs
+var interpolated_yes = $"foo {bar} foo";
+///                    ^^^^^^^^^^^^^^^^ meta.string.interpolated.cs
+var verbatim_singleline = @"foo";
+///                       ^^^^^^ string.quoted.double.raw.cs
+var verbatim_singleline_interpolated_none = $@"foo bar";
+///                                         ^^^^^^^^^^^ meta.string.interpolated.cs string.quoted.double.raw.cs
+var verbatim_singleline_interpolated_yes = $@"foo {bar} foo";
+///                                        ^^^^^^^ string.quoted.double.raw.cs
+///                                        ^^^^^^^^^^^^^^^^^ meta.string.interpolated.cs
+var verbatim_multiline = @"foo bar
+///                      ^^^^^^^^^^ string.quoted.double.raw.cs
+baz";
+var verbatim_multiline_interpolated_none = $@"foo bar
+///                                        ^^^^^^^^^^^ meta.string.interpolated.cs string.quoted.double.raw.cs
+baz";
+var verbatim_multiline_interpolated_yes = $@"foo {bar}
+///                                       ^^^^^^ string.quoted.double.raw.cs
+///                                       ^^^^^^^^^^^^ meta.string.interpolated.cs
+baz";
+
     "{32F31D43-81CC-4C15-9DE6-3FC5453562B6}"
 /// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ string.quoted.double
 ///  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ constant.other.guid


michaelblyons:
If I were to do this for Javascript backtick `string.templates`, is there a particular scope I should invent to use? Here's a list of the `meta.string` scopes in this repo.

* meta.string-constant.braces.bibtex
* meta.string-constant.parenthesis.bibtex
* meta.string-contents.quoted.double.php
* meta.string-contents.quoted.single.php
* meta.string.python
* meta.string.interpolated.python
* meta.string.interpolated.cs (this PR)

keith-hall:
For JS template literals, rather than invent a scope, I would go for `meta.string.interpolated.js`, as they support string interpolation features (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

FichteFoll:
Note that for unification, all strings should get a `meta.string` scope, probably.

Thom1729:
I wish that we had a single consistent recommendation for handling string interpolation. It seems like every language scopes them a little differently:

- JavaScript templates use `string.template`, `meta.template.expression`, and `punctuation.template-expression.begin/end`.
- Python f-strings use `meta.string.interpolated string.quoted.single/double`, `meta.interpolation`, and `punctuation.section.interpolation.begin/end`.
- Ruby double-quoted strings use `string.quoted.double` and `punctuation.section.embedded`.

Each one of these uses `source.<whatever>.embedded`; Ruby includes the punctuation in that scope while the others don't. In addition, Ruby keeps the `string` scope inside the interpolation while the others don't.

Depending on the language, all strings might be eligible for interpolation, a generic type (e.g. double-quoted strings in Ruby), or a language-specific type (f-strings in Python or templates in JavaScript). We have to respect broader conventions, like using `string.quoted.double` for Ruby interpolated strings instead of `string.interpolated` or something else. Beyond that, more specific scopes are only useful if they can be standardized across many languages.

It seems to me that there is/ought to be consensus on the following:

- The entirety of an interpolated string should be scoped `meta.string`.
- The string should be scoped `string`, excluding interpolations (contra Ruby).
- Interpolations, including punctuation, should receive a `meta` scope.
- Punctuation defining an interpolation should be scoped `punctuation.definition` or `punctuation.section`. (We should standardize on one or the other.)
- Code inside the interpolation should be scoped `source.<whatever>.embedded`.

Then, open questions are:

- What subscopes should `meta.string` have? What information should we try to encode there?
- Is there any useful way to further specialize `string`, or should we stick with the standard `string.quoted.single/double/triple/other`?
- What meta scope should we use for an interpolation? Hopefully something generic.
- Should `${}` and its ilk be `punctuation.definition` or `punctuation.section`? (Honestly, I don't feel like I have an intuitive grasp of the distinction.)

FichteFoll:
There is already an RFC issue for this topic (#715), but since you asked here and I cross-referenced it already anyway, I'll just give my opinion on your points and questions here.

Disclaimer: I implemented Python's interpolated string highlighting and am highly in favor of following that model, as I would've done it differently otherwise.

>- Punctuation defining an interpolation should be scoped `punctuation.definition` or `punctuation.section`. (We should standardize on one or the other.)

@wbond had some guideline for these two, where the former, `definition`, would describe a more or less atomic syntax element (e.g. a string that will always represent a string) and the latter a potentially larger section, such as a braced block following an `if` keyword in C-like languages. Applying this here would hint at using a `section` scope for interpolation sections *within* strings, but `definition` for the string itself.
Do not ask me about things such as generics though. I'm glad I haven't had to deal with this so far. We certainly need a separate RFC for the punctiation differentiation.

> - What subscopes should `meta.string` have? What information should we try to encode there?

I think originally this was implemented for BracketHighlighter to get a grasp on an entire string (with all its punctuation) and with clearing the `string` scope that information was lost for interpolated strings. Of course, this was just a hint that we were generally losing information, so adding a `meta.string` scope was a natural consequence. I don't know what other information this meta scope could entail besides a string either being interpolated or not, since information about the string's punctuation can already be found in the `string` scope itself.

>- Is there any useful way to further specialize string, or should we stick with the standard `string.quoted.single/double/triple/other`?

I don't think we can change any of the segments mean at this point. We could probably add more options for the second and third segment segment, but I haven't found a need for that yet.

>- What meta scope should we use for an interpolation? Hopefully something generic.

I'm not sure I understand the context for this one. Do interpolations ever occur outside of strings? is `meta.string.interpolated` not generic enough or do you mean this in the sense that JavaScripts calls them "templates"?

Regarding section vs definition, see my paragraph at the beginning of this post.

michaelblyons:
@FichteFoll Thanks for cross-linking. I checked the forum for guidance, but didn't check GitHub.

My thoughts run thusly:

- On #715, there is conversation about `source.<lang>.embedded` bit with `source` removed. Are the proponents meaning that we'd have a scope like `python.embedded`? It seems odd to me to have a scope with the language name first. If we were going to go so far, I'd prefer to reverse them, and here's why:
    + If I have `embedded.python`, `embedded.css`, `embedded.javascript`, etc., a color scheme designer can make a scope to target `embedded` and give it (for instance) a subtly different background color to show that the language has changed. And s/he can do this without messing with each language separately.
- I'd _like_ to rename the Javascript `template` scopes to match `interpolated` like Python/C#.
- Ruby should also be brought in line.
- I haven't looked at Bash/PHP, but maybe there's work to do there, too.
- Once this is hashed out, it'd be great to update the [official docs.](https://www.sublimetext.com/docs/3/scope_naming.html)

FichteFoll:
>a color scheme designer can make a scope to target embedded and give it (for instance) a subtly different background color

You can do that with `source source, text source, source text` already. That's what I do at least, and I love it. I like the thinking behind unifying that to just `embedded`, but then you can't find *any* source of Python with just `source.python` anymore and need `source.python, embedded.python`. This may or may not have advantages in other areas, however.

Bash definitely needs work currently, since the `string` scope isn't cleared (see #1248). It also uses different scope names overall, or at least those of my color scheme that I use for Python (generically) don't work.

michaelblyons:
@FichteFoll Good point on the `source source`, etc. If that's what you're doing, I think _keeping_ `source.<lang>.embedded` is better than either what I said above or dropping the `source` prefix.

wbond:
From reading over these comments, @FichteFoll seems to have highlighted my thoughts on this.

I need to get this, the guidance for section/definition and decorator scoping into the official docs.

In terms of `string.quoted`, @FichteFoll, TCL now has `string.quoted.brace`. Because TCL is weird. :-)

michaelblyons:
Rebased onto master.

wbond:
Thanks @michaelblyons!

Issue #1433, opened 2018-02-16
[C++] Templates and operator overloading throws off grammar


TheGag96:
I was writing up some homework the other day and I ran into a syntax highlighting issue.

![](https://i.imgur.com/D9bjph4.png)

`template` should be highlighted, but it isn't after the first one there. Removing both the firs `<T>` the `<` in `operator<` will make them highlight again:

![](https://i.imgur.com/Li0Z4vc.png)

![](https://i.imgur.com/WLhcyld.png)

Also, shouldn't `polarNum<T>::funcName` highlight regardless of whether the `<T>` is there?

rwols:
Could you post the faulty snippet as text instead of a picture?

rwols:
Never mind, I think I have a fix for this.

PR #1435, opened 2018-02-16
[C++/ObjC] Fix: operator definitions combined with template classes

rwols:
fixes https://github.com/sublimehq/Packages/issues/1433

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -845,7 +845,7 @@ contexts:
     # All uppercase identifier just before a newline is most likely a macro
     - match: '[[:upper:][:digit:]_]+\s*$'
     # Operator overloading
-    - match: '(?=({{path_lookahead}}\s*::\s*)?{{operator_method_name}}\s*(\(|$))'
+    - match: '(?=({{path_lookahead}}\s*(?:{{generic_lookahead}})?::\s*)?{{operator_method_name}}\s*(\(|$))'
       set:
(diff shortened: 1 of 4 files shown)


C++/C++.sublime-syntax
```diff
@@ -845,7 +845,7 @@ contexts:
     # All uppercase identifier just before a newline is most likely a macro
     - match: '[[:upper:][:digit:]_]+\s*$'
     # Operator overloading
-    - match: '(?=({{path_lookahead}}\s*::\s*)?{{operator_method_name}}\s*(\(|$))'
+    - match: '(?=({{path_lookahead}}\s*(?:{{generic_lookahead}})?::\s*)?{{operator_method_name}}\s*(\(|$))'
```
FichteFoll: If the generic look ahead variable is a... Look ahead and matches, the following two colons will likely not match. I only see the difference here and doing more review on mobile is tedious, so please excuse me if I missed something here.

FichteFoll: I mean, you added Tests for this and they pass, but it just seems weird. Maybe the variable terminology is misleading?

rwols: I'm not entirely sure what you mean.

Yes, this still doesn't highlight correctly:
```
template <class T> bool A::B<T>::C::operator<(const C& c) { return false; }
```
But this works now:
```
template <class T> bool A::B::C<T>::operator<(const C& c) { return false; }
```

PR #1436, opened 2018-02-17
[Rust] OSX: don't override loaded path and build & run a single file

Uupis:
5b29174: Sublime Text loads environment variables on startup (`zsh -l`, in case of zsh), so the hardcoded path in Rust.sublime-build file overrides the loaded path. On top of that, the hardcoded path doesn't include `~/.cargo/bin` anyway, so the build does not work.

2552ed0: I didn't initially intend to pull-request it, but I did not know GitHub would tack it onto the pull request on push. In any case, it adds a build variant to build and run the Rust file. It's been convenient while learning Rust to just have one-off files here and there to play with. I won't be at all surprised if you don't accept this one, though.

### DIFF
--- a/Rust/Rust.sublime-build
+++ b/Rust/Rust.sublime-build
@@ -2,10 +2,6 @@
     "cmd": ["rustc", "$file"],
     "selector": "source.rust",
     "file_regex": "^(.*?):([0-9]+):([0-9]+):\\s[0-9]+:[0-9]+\\s(.*)$",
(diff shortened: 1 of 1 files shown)


Uupis:
Thank you for the feedback, @FichteFoll! I've made the changes in 5ca5880.

wbond:
These seem like reasonable changes, thanks for taking the time to make a PR!

PR #1447, opened 2018-03-04
[C++/Python/ShellScript] Flow control keywords - more specific scopes for "jumping" keywords.

FHTheron:
Relating to RFC #1228, this adds/refines scopes for jumping type flow control keywords. In this context jumping refers to more arbitrary changes in flow than the branch from an `if` or such.

This only covers **C, C++, Python and Bash** and does not attempt to do any standardization across languages and does not replace existing scopes. It also does not change anything for support functions like `exit` or `assert`, but this can be reviewed in future if necessary and most of these already have fairly specific scopes.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -87,6 +87,16 @@ contexts:
   unique-keywords:
     - match: \busing\b
       scope: keyword.control.c++
+    - match: \bbreak\b
+      scope: keyword.control.flow.break.c++
+    - match: \bcontinue\b
+      scope: keyword.control.flow.continue.c++
+    - match: \bgoto\b
+      scope: keyword.control.flow.goto.c++
+    - match: \breturn\b
+      scope: keyword.control.flow.return.c++
+    - match: \bthrow\b
+      scope: keyword.control.flow.throw.c++
     - match: \b({{control_keywords}})\b
       scope: keyword.control.c++
     - match: '\bdelete\b(\s*\[\])?|\bnew\b(?!])'
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -111,6 +111,14 @@ contexts:
       scope: constant.other.placeholder.c
 
   keywords:
+    - match: \bbreak\b
+      scope: keyword.control.flow.break.c
+    - match: \bcontinue\b
+      scope: keyword.control.flow.continue.c
+    - match: \bgoto\b
+      scope: keyword.control.flow.goto.c
+    - match: \breturn\b
+      scope: keyword.control.flow.return.c
     - match: \b({{control_keywords}})\b
       scope: keyword.control.c
     - match: \bsizeof\b
--- a/C++/syntax_test_c.c
+++ b/C++/syntax_test_c.c
@@ -584,46 +584,51 @@ scanf("%ms %as %*[, ]", &buf);
 // Control Keywords
 /////////////////////////////////////////////
 
-if (x < 5)
-/* <- keyword.control */
-{}
-else
-/* <- keyword.control */
-{}
-
-switch (x)
-/* <- keyword.control */
+int control_keywords()
 {
-case 1:
-/* <- keyword.control */
-    break;
-    /* <- keyword.control.flow.break */
-default:
-/* <- keyword.control */
-    break;
-    /* <- keyword.control.flow.break */
+  if (x < 5)
+  /* <- keyword.control */
+  {}
+  else
+  /* <- keyword.control */
+  {}
+
+  switch (x)
+  /* <- keyword.control */
+  {
+  case 1:
+  /* <- keyword.control */
+      break;
+      /* <- keyword.control.flow.break */
+  default:
+  /* <- keyword.control */
+      break;
+      /* <- keyword.control.flow.break */
+  }
+
+  do
+  /* <- keyword.control */
+  {
+      if (y == 3)
+          continue;
+          /* <- keyword.control.flow.continue */
+  } while (y < x);
+  /*^ keyword.control */
+
+  switch (a) {
+      case 1: break;
+  /*        ^ punctuation.separator */
+      case 100 - 10: break;
+  /*               ^ punctuation.separator */
+      default: break;
+  /*         ^ punctuation.separator */
+  }
+
+  goto label;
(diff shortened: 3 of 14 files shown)


Python/Python.sublime-syntax
```diff
@@ -84,9 +84,13 @@ contexts:
     - match: \b(exec)\b(?! *($|[,.()\]}]))
       scope: keyword.other.exec.python
     - match: \b(return)\b
-      scope: keyword.control.flow.return.python
-    - match: \b(break|continue|pass)\b
-      scope: keyword.control.flow.python
+      scope: keyword.control.flow.python keyword.control.flow.return.python
```
keith-hall: It seems a little odd to duplicate the `keyword` scope to remove a "scope atom" before the "scope suffix". Do your color schemes really have to target `keyword.control.flow.python`? Using a selector like `source.python keyword.control.flow` would have the same effect and remove the need for the duplication changes here, if you really must use language specific colors.

FHTheron: This would be specific to the `return` right? I tried to keep a pattern for `return`, `break`, `pass` and `continue`. For the break/pass/continue scopes I had to keep the `flow.python` scope and thus add the new scopes. For `return` this is not strictly necessary, but it seemed good to have the pattern. 

I don't mind removing it for `return`, but for the others the duplication is necessary unless we address the language-suffix topic for the whole file which I did not want to do just for this small addition. I see the suffix pattern in many other syntaxes and I would certainly prefer to not have the language suffix, since that would have made things simpler and the duplication completely unnecessary.

C++/C++.sublime-syntax
```diff
@@ -87,6 +87,16 @@ contexts:
...
+      scope: keyword.control.flow.break.c++
+    - match: \bcontinue\b
+      scope: keyword.control.flow.continue.c++
+    - match: \bgoto\b
+      scope: keyword.control.flow.goto.c++
+    - match: \breturn\b
+      scope: keyword.control.flow.return.c++
+    - match: \bthrow\b
+      scope: keyword.control.flow.throw.c++
     - match: \b({{control_keywords}})\b
```
rwols: Could the `control_keywords` variable be made simpler now? It might be used elsewhere for different purposes, so I don't know.

C++/C.sublime-syntax
```diff
@@ -111,6 +111,14 @@ contexts:
...
 
   keywords:
+    - match: \bbreak\b
+      scope: keyword.control.flow.break.c
+    - match: \bcontinue\b
+      scope: keyword.control.flow.continue.c
+    - match: \bgoto\b
+      scope: keyword.control.flow.goto.c
+    - match: \breturn\b
+      scope: keyword.control.flow.return.c
```
rwols: Don't forget to mirror your changes in Objective-C and Objective-C++, too.

ShellScript/commands-builtin-shell-bash.yml
```diff
@@ -92,6 +92,7 @@ pwd:
 return:
   allow-end-of-options-token: false
   scope: keyword.control.return.shell
+         keyword.control.flow.return.shell
```
rwols: Fun stuff. Did you run the Makefile in the ShellScript directory?

ShellScript/Bash.sublime-syntax
```diff
@@ -182,9 +182,9 @@ contexts:
       scope: keyword.control.case.begin.shell
       set: [case-body, case-item, case-item-first-character, case-preamble]
     - match: \bcontinue{{keyword_boundary_end}}
-      scope: keyword.control.continue.shell
+      scope: keyword.control.continue.shell keyword.control.flow.continue.shell
     - match: \bbreak{{keyword_boundary_end}}
-      scope: keyword.control.break.shell
+      scope: keyword.control.break.shell keyword.control.flow.break.shell
```
rwols: I agree with @keith-hall in that it seems strange to have these two scopes. I think we can make do with just your new `keyword.control.flow.*` scopes, no?

C++/syntax_test_c.c
```diff
@@ -578,3 +578,52 @@ scanf("%ms %as %*[, ]", &buf);
...
+/*^ keyword.control */
+
+switch (a) {
+    case 1: break;
+/*        ^ punctuation.separator */
+    case 100 - 10: break;
+/*               ^ punctuation.separator */
+    default: break;
+/*         ^ punctuation.separator */
+}
```
rwols: Could you please put these statements and expressions in a function definition (like `main`)? These added tests might screw up future tests because there's this global-local idea throughout the syntax definition. And strictly speaking, this is not valid C if it's at global scope.

ShellScript/commands-builtin-shell-bash.yml
FHTheron: Yes I did. Took a moment to figure out where things are coming from.

C++/C++.sublime-syntax
FHTheron: How do you mean? The `control_keywords` variable is indeed used elsewhere so these flow keywords are still in it, it's just in this one context that they get some special pre-`control_keywords` matching.

ShellScript/Bash.sublime-syntax
FHTheron: If a minor breakage is acceptable then yes, but I would think that we should then look for similar scopes that can be changed accordingly at the same time and also decide if `keyword.control` implies flow control like `if` and `while` or if those should also move down to `keyword.control.flow.*` for consistency.

Again, my goal was to add some extra specific scopes without really affecting anything else rather than make any kind of decision.

Python/Python.sublime-syntax
FichteFoll: Usually, you just append the suffix at the end of whatever scope you have at that point. Scope names may change, and they do so rather frequently, so I don't think we need a backwards compat for this change. Suffixes in selectors are to be used when you need to target something *very specific* and may need to be updated occasionally.

C++/C.sublime-syntax
FHTheron: It seems like the ObjC/C++ syntaxes already have these because of `- include: scope:source.c#keywords` so I updated the syntax tests (but excluded the language suffix from the check)

wbond:
Thanks, and sorry for the delay in going over these

Issue #1361, opened 2017-12-20
[Go] Test file uses tabs for indentation


FichteFoll:
Noticed this via https://github.com/SublimeText/PackageDev/issues/177.

Using tabs is not recommended because it will break your tests if you find yourself with mixed indentation between code and test lines or when they don't start at the same indent level, which isn't that hard to run into.

deathaxe:
@FichteFoll @wbond  fixed by the Go Rework

FichteFoll:
:tada: https://github.com/sublimehq/Packages/commit/14a35f641f348b4ee0ead558e5ee474f7255b50f

PR #1448, opened 2018-03-04
[repo] Add .editorconfig to repo root (http://editorconfig.org)

FHTheron:
To use this an editor needs to support it. For SublimeText there
is a package called EditorConfig (I'd argue this should become
part of ST Core in future). For other editors, see editorconfig.org
for a list of editors that have built-in support and ones that have
plugins.

This commit only provides a first draft as starting point based on
what a first time contributor thinks the settings should be. This
file was prompted after the user-specific setting
"trim_trailing_white_space_on_save" caused a syntax test to break
(it had specific trailing whitespace for a test).


.editorconfig
```diff
@@ -0,0 +1,35 @@
...
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 2
```
FichteFoll: I don't think the indent size can be set generally for all files, notably test files, as many will probably be using four spaces (or whatever their syntax prefers). You unset them later for syntax test files, but why set it indentation settings globally at all then? What is the purpose of this specification?

.editorconfig
```diff
@@ -0,0 +1,35 @@
...
+indent_size = 4
+
+# Ensure specific settings for YAML (and similar) in case defaults change
+[{*.yml,*.sublime-syntax}]
+indent_style = space
+indent_size = 2
+
+[{*.tmPreferences,*.sublime-snippet}]
+indent_style = tab
+indent_size = 2
```
FichteFoll: I think having a tab rendered with width 4 is most common nowadays (and also the default for Sublime Text), so I believe this shouldn't be changed. It should be using tabs, but there is no need to specify the indent size.

.editorconfig
FHTheron: The global settings are perhaps force of habit. Normally I'd say the purpose is consistency across a code base, but in this particular case I suppose the primary purpose is actually to ensure that mutative settings are disabled for the files that do have strict requirements. This repo is also not a typical collection of files and contains many very different languages.

So the global/default settings can be removed, or perhaps just reduced to `charset`.

.editorconfig
FHTheron: `indent_size` can be removed across the board then, except for `{*.yml,*.sublime-syntax}`. The Makefile section should perhaps also not be there, since there is only one Makefile in the repo apart from the whole Makefile package.

If we go for the minimum, I guess it will be:
```
root = true

[{*.yml,*.sublime-syntax}]
indent_style = space
indent_size = 2

[{*.tmPreferences,*.sublime-snippet}]
indent_style = tab

# Syntax test files should use spaces to work correctly.
[syntax_test_*]
trim_trailing_whitespace = false
indent_style = space
```

FichteFoll: This looks better :+1:

We might even recommend able to add the newline at eof. I'd like to check if any file doesn't do that currently though.

FichteFoll: s/recommend/be/

FHTheron: Seems there are 7 files that do not end with a new line:
```
% git ls-files | while read f; do tail -c1 "$f" | wc -l | grep -q "1" || echo "$f"; done                                                                     ActionScript/ActionScript.sublime-build
JavaScript/JavaScript.sublime-syntax
Lua/Lua.sublime-build
Ruby/Default.sublime-keymap
Scala/info.plist
Scala/syntax_test_scala.scala
Text/Snippets/lorem.sublime-snippet
```

FichteFoll: >`Scala/info.plist`

Wow, this does nothing and can be removed.

To all other files a newline could easily be added at the end. This would be a safe change, but it should be made together with the change to standardize the editorconfig configuration.

FHTheron:
I kept `trim_trailing_whitespace = false` for a global setting. Without it my default causes a lot of whitespace changes every time I save a file, but I want to keep it set for my daily workflow and I imagine others might too.

Looking at EOLs briefly, it looks like there is only one file using CRLF. We could add a global `eol` setting to editorconfig or we could add a .gitattributes file to let git handle it (assuming random EOLs bother more than one person). This is getting very pedantic though and my primary goal is to prevent syntax test files from being affected by user settings.

FHTheron:
Just adding a reference to #1361 - .editorconfig would have helped there.

wbond:
Since we don't support EditorConfig out of the box, it doesn't really make sense to make this part of the repo.

PR #1450, opened 2018-03-06
[JavaScript] Simplified function arguments.

Thom1729:
There were two separate code paths handling function arguments, and they worked slightly differently. As a result, some comma separators were being parsed as comma operators. This is not currently visible in the core syntax (though see #831), but JS Custom offers an option to distinguish them. Because the solution is a straight simplification, I thought it reasonable to handle this upstream.

Now, all function call arguments are handled with one context. I renamed this context `function-call-arguments` (from `function-call`).

I also added semicolons to some related syntax tests. They worked either way, but without the semicolons they were a bit misleading.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -608,7 +608,9 @@ contexts:
     - include: ternary-operator
 
     - include: property-access
-    - include: function-call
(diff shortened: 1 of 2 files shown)


wbond:
This seems to have some conflicts with master

PR #1461, opened 2018-03-14
[Lua] function scope fixes

Resike:


keith-hall:
You may want to add some [syntax tests](http://www.sublimetext.com/docs/3/syntax.html#testing) to ensure these changes aren't regressed in the future.

Resike:
I'm on it.

Resike:
Okay, how can i add the targeted syntax_test file?

Resike:
Added a test file, which should properly cover every possible function scopes at #1463.

Lua/Lua.sublime-syntax
```diff
@@ -8,18 +8,16 @@ file_extensions:
 scope: source.lua
 contexts:
   main:
-    - match: '\b(function)\b\s*([a-zA-Z_]\w*)?\s*(([.]\s*[a-zA-Z_]\w*\s*)*)\s*(([.:])?\s*[a-zA-Z_]\w*)?\s*(\()([^)]*)(\))?'
+    - match: '\b(function)\b\s*(([a-zA-Z_]\w*\s*[.]\s*)*)?\s*([a-zA-Z_]\w*\s*[:])?\s*([a-zA-Z_]\w*)?\s*(\()([^)]*)(\))?'
```
FichteFoll: I think this regex would benefit from being split up and using variables (for the identifiers). It's not really readable at the moment.

Lua/Lua.sublime-syntax
```diff
@@ -8,18 +8,16 @@ file_extensions:
...
       scope: meta.function.lua
       captures:
         1: keyword.control.lua
         2: entity.name.function.scope.lua
-        3: entity.name.function.scope.lua
```
FichteFoll: If you don't want to assign a scope to a group, you should make it non-capturing with `(?:)`.

Resike: But i want it to capture:
```
function foo.bar:fooo(...)

end
```
`bar:` here is only a scope, and not a name.

Resike:
I can break down the [regex ](https://regex101.com/) for you:

'\b(function)\b\s*(([a-zA-Z_]\w*\s*[.]\s*)*)?\s*([a-zA-Z_]\w*\s*[:])?\s*([a-zA-Z_]\w*)?\s*(\()([^)]*)(\))?'


`\b(function)\b`
function

`\s*`
Whitespaces.
Why `*` instead of `+`? Because `self = function(arg1, arg2, ...) end` is also a valid lua function.

`(([a-zA-Z_]\w*\s*[.]\s*)*)?`
Matches `scope.` as many times as possible.
First character can be only `[a-zA-Z_]` after then `\w*`.
Optional.

`\s*`
Whitespaces.

`([a-zA-Z_]\w*\s*[:])?`
Matches the last scope before the `:` without this this scope would be called name instead.
First character can be only `[a-zA-Z_]` after then `\w*`.
Optional.

\s*
Whitespaces.

([a-zA-Z_]\w*)?
Function name.
First character can be only `[a-zA-Z_]` after then `\w*`.
Optional.

`\s*`
Whitespaces.

`(\()`
Staring parentheses.

`([^)]*)`
Everything until the next closing parentheses.

`(\))?`
Closing parentheses.
Optional, to allow function closure in the next line.

Resike:
I couldn't make the punctuation coloring to work with a single regex match (you guys can work on that), but all the goto definitions should work properly now.

Here are all the valid Lua functions that should get covered:

image

FichteFoll:
Can you add tests for this?

Resike:
Thom and I will rewrite the whole Lua syntax, so this is will be obsolete.

PR #1463, opened 2018-03-14
[Lua] adding syntax test file

Resike:


wbond:
I believe this is now redundant as of #1811. Sorry for not getting to it sooner.

PR #1465, opened 2018-03-15
[Ruby] Fixed keybind for interpolation and improved scopes.

### LINKED ISSUE #1458: [Ruby] Autocomplete within string interpolation doesn't complete word
When interpolating an existing variable into a string via `"#{}"`, the autocomplete dropdown shows up, but pressing `Tab` doesn't autocomplete the word. Instead, the cursor jumps to the outside of the right bracket.

Gif to demonstrate:

string_interp

Thom1729:
Fixes #1458. Uses interpolation scopes based on Python. See #1432 for further discussion.

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -1259,7 +1259,7 @@ contexts:
         - meta_scope: meta.interpolation
         - meta_content_scope: source.ruby.embedded
         - match: '\}'
-          scope: punctuation.definition.interpolation.end.ruby
+          scope: punctuation.section.interpolation.end.ruby
           pop: true
         - include: nest-curly-expressions
         - include: expressions
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -1252,17 +1252,14 @@ contexts:
       scope: constant.character.escape.ruby
 
   interpolated-ruby:
-    - match: '(#\{)(\})'
-      scope: punctuation.section.embedded.ruby
-      captures:
-        1: source.ruby.embedded.source
-        2: source.ruby.embedded.source.empty
     - match: '#\{'
-      scope: punctuation.section.embedded.ruby
+      scope: punctuation.definition.interpolation.begin.ruby
       push:
-        - meta_scope: source.ruby.embedded.source
+        - clear_scopes: 1
+        - meta_scope: meta.interpolation
+        - meta_content_scope: source.ruby.embedded
         - match: '\}'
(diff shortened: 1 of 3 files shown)


Ruby/Ruby.sublime-syntax
```diff
@@ -1252,17 +1252,14 @@ contexts:
...
 
   interpolated-ruby:
-    - match: '(#\{)(\})'
-      scope: punctuation.section.embedded.ruby
-      captures:
-        1: source.ruby.embedded.source
-        2: source.ruby.embedded.source.empty
     - match: '#\{'
-      scope: punctuation.section.embedded.ruby
+      scope: punctuation.definition.interpolation.begin.ruby
```
FichteFoll: Via https://github.com/sublimehq/Packages/pull/1432#issuecomment-366750733, I believe `punctuation.section.interpolation.*` should be used.

If this is not currently the case with Python, it should be adjusted imo.

wbond: @Thom1729 this should be

```suggestion
      scope: punctuation.section.interpolation.begin.ruby
```

Ruby/Ruby.sublime-syntax
```diff
@@ -1252,17 +1252,14 @@ contexts:
...
       push:
-        - meta_scope: source.ruby.embedded.source
+        - clear_scopes: 1
+        - meta_scope: meta.interpolation
+        - meta_content_scope: source.ruby.embedded
         - match: '\}'
-          scope: punctuation.section.embedded.ruby
+          scope: punctuation.definition.interpolation.end.ruby
```
wbond: This should be

```suggestion
          scope: punctuation.section.interpolation.end.ruby
```

Ruby/syntax_test_ruby.rb
```diff
@@ -20,7 +20,8 @@
 
 class_eval <<-RUBY, __FILE__, __LINE__ + 1
   def #{sym}(*args, &block)
-#     ^^ punctuation.section.embedded
+#     ^^ punctuation.definition.interpolation.begin
```
wbond: This should be

```suggestion
#     ^^ punctuation.section.interpolation.begin
```

Ruby/syntax_test_ruby.rb
```diff
@@ -20,7 +20,8 @@
...
+#          ^ punctuation.definition.interpolation.end
```
wbond: This should be

```suggestion
#          ^ punctuation.section.interpolation.end
```

Thom1729:
All done.

PR #1469, opened 2018-03-21
[JavaScript] Refactor statements.

Thom1729:
This PR does for statements what #1009 did for expressions. Fortunately, this is a much, much simpler change.

This change:
- Allows for the removal of some code that worked slightly differently for statements and expressions. This was especially noticeable in the case of import/export statements.
- Properly treats semicolons as part of the statements they terminate. An immediately observable consequence is that the semicolon ending an import/export statement will be scoped `meta.import|export`.
- Scopes the semicolon of an empty statement `punctuation.terminator.statement.empty`. (These semicolons are usually unwanted.) This will make it much easier to write sensible tests for statements that do not require a semicolon.
- Fixes the `of`/`in` keyword in `for` loops, which was apparently broken. There are some corner cases that aren't handled perfectly, and in fact distinguishing the two types of for loops perfectly is impossible in Sublime (like distinguishing arrow functions from parenthesized expressions). The new code for this should do the right thing in reasonable cases and fail gracefully elsewhere.
- Untangles some of the conditional logic for constructs like `switch` and `do-while`.
- Renames some scopes that have drifted from their original purpose.

In addition, it fixes a parsing problem with commas in exports that was visible in JS Custom. The core syntax scopes the rare JavaScript comma *operator* as punctuation, so the problem was not observable in the core syntax. However, this fix lets us implement #831 if we want to.

As a result of this change, we do have a lot of anonymous contexts providing meta scopes:

```yaml
    - match: switch{{identifier_break}}
      scope: keyword.control.switch.js
      set:
        - - meta_scope: meta.switch.js
          - include: immediately-pop
        - switch-block
        - parenthesized-expression
```

These contexts are not reusable, and in my opinion they're easier to understand "inline" rather than factored out. Opinions may vary on this.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -106,14 +106,14 @@ contexts:
   import-export:
     - match: import{{identifier_break}}
       scope: keyword.control.import-export.js
-      push:
+      set:
         - import-meta
         - expect-semicolon
         - import-string-or-items
 
     - match: export{{identifier_break}}
       scope: keyword.control.import-export.js
-      push:
+      set:
         - export-meta
         - export-extended
 
@@ -245,9 +245,6 @@ contexts:
       scope: invalid.illegal.stray-bracket-end.js
       pop: true
 
-    - match: \;
-      scope: punctuation.terminator.statement.js
-
     - match: (?=\S)
       push: statement
 
@@ -285,7 +282,7 @@ contexts:
 
   expect-semicolon:
     - match: \;
-      scope: punctuation.terminator.statement.empty.js
+      scope: punctuation.terminator.statement.js
       pop: true
     - include: else-pop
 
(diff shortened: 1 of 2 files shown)


Thom1729:
@wbond FYI, this does fix a fairly visible bug with `for ... in` loops, so it would be nice to get it into the next dev build if possible. I may have buried the lede in the original comment. The bug was probably introduced by one of the changes implementing destructured bindings.

wbond:
I am going to do another pass through the packages before the next dev build, so I should be able to get this in

wbond:
I'm definitely not keen on the `- -` nested structures. For someone who reads YAML pretty regularly, it isn't that big a deal, but it gets harder to determine what indentation level different parts are.

Thom1729:
@wbond I've refactored the meta contexts in #1479 to avoid the nested lists.

PR #1470, opened 2018-03-22
[PHP] Add highlight for PHPUnit annotations

jfcherng:
PHPUnit is the most popular (or say, dominant) unit test framework for PHP.

For example, this PR highlights the `@dataProvider` in the following code.

```php
/**
 * Test Parser::parse().
 *
 * @dataProvider parserParseDataProvider
 *
 * @param string $input    The input
 * @param array  $expected The expected output
 */
public function testParse(string $input, array $expected): void
{
    $this->assertEquals($expected, Parser::parse($input));
}
```

Ref: 

- PHPUnit annotations: https://phpunit.readthedocs.io/en/latest/annotations.html


wbond:
Thanks!

Issue #1475, opened 2018-03-24
Batch file syntax color issues in build 3157 


rivy:
With build 3157 (Win10), I'm getting some color syntax issues with Windows batch files (.bat/.cmd):

1. `set "X="` seems not to correctly end syntax scope, causing future lines (eg, `:: NOTE: ...` and `set ...`) to be highlighted incorrectly.
2. `set /p OUTPUT="( ... ) ... "` incorrectly sees the internal `)` as an error.

![batch-syntax-errors](https://user-images.githubusercontent.com/80132/37868204-e4d156a0-2f70-11e8-82ac-25ecab5fa43d.png)

keith-hall:
It looks like this was partially fixed by https://github.com/sublimehq/Packages/pull/1412, although quoted strings after `set /p` then had a `string.unquoted` and a `string.quoted` scope. All of these issues have been resolved in https://github.com/sublimehq/Packages/pull/1476.

rivy:
Thanks!

PR #1476, opened 2018-03-26
[Batch File] fix some `set` string related bugs

keith-hall:
Fixes https://github.com/sublimehq/Packages/issues/1475


deathaxe:
Looks good so far.

keith-hall:
It may make sense to do the `set /a` fixes in a separate PR, as it's looking pretty complicated to get right :tm: 

i.e. because this is valid `set /a abc"*="var2+="%var3%"`

PR #1479, opened 2018-03-27
[JavaScript] Un-nest statement meta contexts.

Thom1729:
Adds explicit meta contexts to prevent nested lists. See discussion in #1469.


wbond:
I realize you weren't a fan of this change, but I appreciate your willingness to do it anyway - thank you!

PR #1487, opened 2018-03-30
[JavaScript] Extend `meta.function` scope to entire function.

Thom1729:
Functions should be enclosed in a `meta.function` scope. Currently, the JavaScript syntax does not do this. However, it does enclose part of the function in a `meta.function.declaration` scope.

This PR encloses the entire function in `meta.function`. The old `meta.function.declaration` behavior is preserved exactly. The scope changes between `meta.function.js` and `meta.function.declaration.js`; it does not use both at once.

### DIFF
--- a/JavaScript/tests/syntax_test_js.js
+++ b/JavaScript/tests/syntax_test_js.js
@@ -196,35 +196,31 @@ someFunction({
 //                ^ variable.other.readwrite
 });
 
-function foo() {
-// <- meta.function.declaration
- // <- meta.function.declaration
-  // <- meta.function.declaration
-// ^^^^^^^^^^^ meta.function.declaration - meta.function.declaration meta.function.declaration
-// ^ storage.type.function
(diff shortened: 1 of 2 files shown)


JavaScript/tests/syntax_test_js.js
```diff
@@ -896,14 +896,19 @@ Bar {}
 class Foo extends getSomeClass() {}
 //                ^^^^^^^^^^^^ meta.function-call variable.function - entity.other.inherited-class
 
-() => {}
-// <- meta.function.declaration punctuation.section.group
- // <- meta.function.declaration punctuation.section.group
-//^^^ meta.function.declaration
-//    ^^ meta.block punctuation.section.block
+    () => {}
+//  ^^^^^^^^ meta.function
```
FichteFoll: Perhaps this should include a test for `- meta.function meta.function`?

PR #1491, opened 2018-04-03
gitmodules are git config

braver:
Applies the Git Config syntax to .gitmodules files.

> The .gitmodules file, located in the top-level directory of a Git working tree, is a text file with a syntax matching the requirements of git-config[1].

https://www.git-scm.com/docs/gitmodules#_description

### DIFF
--- a/Git Formats/Git Config.sublime-syntax
+++ b/Git Formats/Git Config.sublime-syntax
@@ -7,7 +7,6 @@ name: Git Config
 file_extensions:
   - gitconfig                # /etc/gitconfig
   - .gitconfig               # ~/.gitconfig
-  - gitmodules               # /etc/gitmodules
   - .gitmodules              # ~/.gitmodules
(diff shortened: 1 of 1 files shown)


Git Formats/Git Config.sublime-syntax
```diff
@@ -7,6 +7,8 @@ name: Git Config
 file_extensions:
   - gitconfig                # /etc/gitconfig
   - .gitconfig               # ~/.gitconfig
+  - gitmodules               # /etc/gitmodules
```
deathaxe: As your comment states, the .gitmodules file exists in the top-level of a working tree. Therefore a globally used file in a directory like `/etc/gitmodules` will most likely not exist, I think.

braver: Oh, no that doesn't make sense.

wbond:
Thanks!

Issue #1492, opened 2018-04-05
HTML Syntax highlighting doesn't work in `script` tag


IPWright83:
I'm using the HTML syntax on a file with an `.hbs` extension. I ran an update on launch so I'm now on 3161 and it seems that there have been changes to syntax highlighting (which are certainly detrimental to me). I'm wondering if this is a bug?

In this particular case I've got `HTML` within a `<script type="text/html">` tag which therefore should be using HTML (rather than JavaScript) highlighting. Currently it looks like this:

<img src="https://forum.sublimetext.com/uploads/default/optimized/3X/d/3/d309085982ce74622c4d18aaf89b390ae71f70db_1_690x240.png" width="690" height="240">

However it used to look like this:

<img src="https://forum.sublimetext.com/uploads/default/optimized/3X/5/1/51b70cff3b097b761160abb26c43a1f2fe2e25b2_1_690x306.png" width="690" height="306">

Thom1729:
I'm not sure why this would have changed in 3161; however, the new highlighting is correct. The core HTML syntax does not support nonstandard scripting languages, including HTML. Currently, the contents of a `<script>` tag will be highlighted as JavaScript if there is a JavaScript-compatible `type` attribute or no `type` attribute, and not highlighted at all otherwise. (See [PR #1379](https://github.com/sublimehq/Packages/pull/1379) for more details.)

As discussed in that PR, the current system is designed to be extensible in core (directly) or in a third-party package (using [YAML Macros](https://github.com/Thom1729/YAML-Macros)). If we do want to add support for other languages inside `<script>` tags, then it should be easy to do.

The question, then, is whether this is a feature that ought to be added to the core syntax. What is the example accomplishing? Is there a third-party JavaScript library looking at the contents of that `<script>` tag or something? Or have I missed a new web standard?

IPWright83:
@Thom1729 Thanks for the response - it may not have been the 3161 build directly that did this, I had 2 machines update and the one exhibiting this behavior took much longer to download, which potentially indicates several updates.

In this example I'm using KnockoutJS [templates](http://knockoutjs.com/documentation/template-binding.html) which are switched out. The reason they go into `<script>` tags as per the Knockout documentation is to prevent them being rendered on initial load.

wbond:
It would seem that we would have to add special-casing back to support `type="text/html"` to keep backwards compatibility.

IPWright83:
[...]

PR #1493, opened 2018-04-06
[HTML] Implemented <script type="text/html">.

Thom1729:
For #1492.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -310,6 +310,18 @@ contexts:
           embed_scope: source.js.embedded.html
           escape: (?i)(?=(?:-->\s*)?</script)
 
+  script-html:
+    - meta_content_scope: meta.tag.script.begin.html
+    - include: script-common
+    - match: '>'
+      scope: punctuation.definition.tag.end.html
+      set:
+        - include: script-close-tag
+        - match: (?=\S)
+          embed: scope:text.html.basic
+          embed_scope: text.html.embedded.html
+          escape: (?i)(?=(?:-->\s*)?</script)
+
   script-other:
     - meta_content_scope: meta.tag.script.begin.html
     - include: script-common
@@ -360,6 +372,11 @@ contexts:
     - match: (?i)(?=module(?!{{unquoted_attribute_value}})|'module'|"module")
       set:
         - script-javascript
+        - tag-generic-attribute-meta
+        - tag-generic-attribute-value
+    - match: (?i)(?=text/html(?!{{unquoted_attribute_value}})|'text/html'|"text/html")
+      set:
+        - script-html
         - tag-generic-attribute-meta
         - tag-generic-attribute-value
     - match: (?i)(?=>|''|"")
--- a/HTML/syntax_test_html.html
+++ b/HTML/syntax_test_html.html
@@ -36,6 +36,14 @@
         <script
         type
         =
+        text/html>
+            <div></div>
+        ##  ^^^^^^^^^^^ text.html.embedded
+        </script>
+
+        <script
+        type
+        =
         'other'
         >
             var foo = 100;


HTML/HTML.sublime-syntax
```diff
@@ -310,6 +310,18 @@ contexts:
...
+    - meta_content_scope: meta.tag.script.begin.html
+    - include: script-common
+    - match: '>'
+      scope: punctuation.definition.tag.end.html
+      set:
+        - include: script-close-tag
+        - match: (?=\S)
+          embed: scope:text.html.basic
+          embed_scope: text.html.embedded.html
+          escape: (?i)(?=(?:-->\s*)?</script)
```
wbond: Do we need the close comment in the lookahead?

Thom1729: Myself, I don't think we should have it at all.

It's there for code written to support *extremely* old browsers that didn't know about the `<script>` tag. It's not valid, and I can count on one hand the number times I've seen comments used like that. I'd just as soon change all of the `escape` lookaheads to `(?i)</script)` (which is what every browser from the past twenty years will actually look for) and remove the comment handling in `script-close-tag`.

wbond: I was thinking in this situation, since this is a lookahead, and the embedded content is also HTML, the `-->` can be handled by the embedded HTML instead of being part of the lookahead.

Does that make sense?

wbond: In other situations (such as JavaScript) we need to pop before the close comment, so it isn't detected as JS.

deathaxe: 1. Why does a language embed itself? 
2. What happens if the embedded html contains another `<script>...</script>` section? I guess it will trigger the `escape` and break highlighting?

keith-hall: I was also wondering, why not just scope the `script` tag and not bother with the embed action if the type is `text/html`, but I guess the closing script tag would then need special handling because otherwise nothing would scope it correctly?

FichteFoll: >What happens if the embedded html contains another `<script>...</script>` section? I guess it will trigger the `escape` and break highlighting?

I guess this largely depends on how the templating engine is implemented. The HTML5 spec isn't exactly clear on that (to me):
https://www.w3.org/TR/html50/scripting-1.html#restrictions-for-contents-of-script-elements

Thom1729: After thoroughly reviewing the canonical HTML5, parsing algorithm, I have reached the conclusion that it's a mess and my head hurts.

It won't be possible to exactly implement the correct algorithm; we'd have to maintain two stacks.

deathaxe: > I guess this largely depends on how the templating engine is implemented

As the `escape` pattern is matched before any of the embedded rules the underlying implementation does not influence it in any way, I think. The very first `</script>` will cause the embedded syntax to be popped off from stack, without any chance for it to avoid that.

Just replace the added test by the following lines to see the new test to fail upon the embedded `</script>` tag.

```
        <script
        type
        =
        text/html>
            <div></div>
        ##  ^^^^^^^^^^^ text.html.embedded
            <script type=text/javascript>
        ##  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ text.html.embedded.html
                function test() {}
        ##      ^^^^^^^^^^^^^^^^^^ source.js.embedded.html
            </script>
        ##  ^^^^^^^^^ text.html.embedded.html
        </script>
```

FichteFoll:
>As the escape pattern is matched before any of the embedded rules the underlying implementation does not influence it in any way, I think.

What I was getting at is whether the "other" implementation supports nested script tags or handles them like the current pr does. 

I don't think this is going to be a big issue, personally, so I'd say we should go for the solution with the least significant side effects.

wbond:
So perhaps we just treat `<script type="text/html">` like any other tag?

PR #1501, opened 2018-04-16
[Python] Ensure variable.annotation/function

FichteFoll:
Allow special names to override `variable.annotation` and `variable.function` and ensure that these scopes are *still* being set before the special ones.

This is noticeable when you set italics rendering on these two scopes and have foreground color overrides on the special names like `support`.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -729,17 +729,16 @@ contexts:
 
   function-call-wrapper:
     - meta_scope: meta.function-call.python
-    - match: \)
-      scope: punctuation.section.arguments.end.python
-      set: after-expression
-    - match: \(
-      scope: meta.function-call.python punctuation.section.arguments.begin.python
-      push:
-        - clear_scopes: 1
-        - meta_content_scope: meta.function-call.arguments.python
-        - match: (?=\))
-          pop: true
-        - include: arguments
+    - match: (?=\()  # need to remove meta.function-call.python from opening parens
+      set:
+        - match: \(
+          scope: punctuation.section.arguments.begin.python
+          set:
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -700,18 +729,34 @@ contexts:
...
-      scope: meta.qualified-name.python variable.function.python
     - match: \)
       scope: punctuation.section.arguments.end.python
       set: after-expression
     - match: \(
-      scope: punctuation.section.arguments.begin.python
+      scope: meta.function-call.python punctuation.section.arguments.begin.python
       push:
+        - clear_scopes: 1
         - meta_content_scope: meta.function-call.arguments.python
```
deathaxe: Most `meta.`s like `meta.group`, `meta.brackets`, ... or even `string.quoted` scopes include the opening and closing punctuation. The C++ syntax as an example uses `meta.group` to scope the function-call arguments. 

I am just wondering why `meta.function-call.arguments` is handled differently (not including the parens)?

deathaxe:
According to the second commit line 338 also contains stacked `meta.function-call` and `meta.function-call.arguments`.

Python/Python.sublime-syntax
wbond: Any thoughts on including the parens in the `.arguments` specialization?

FichteFoll: I probably did it because of `set` semantics and that requiring me to push into a temporary context to get rid of the previous `meta_scope`. See 5be7b83dc88c0dc0db1fe0292e1756616a780938 for details.

wbond:
Thanks!

PR #1502, opened 2018-04-16
[Python] Add special bindings for auto-pairing

### LINKED ISSUE #1441: [Python] Poor invalid scopes while typing
I've been noticing lots of invalid scopes flashing the Python syntax as I've been refactoring a codebase over the past few days. These should be left to a linter, since it makes the editing experience kind of jaring.

Wrapping a value in a method call (cursor after `.`):

```python
a = foo.'bar'
```

Adding a parameter to a function/method call (cursor after `.`):

```python
foo.bar(baz., True)
```

Need to be improved by better pairing (cursor after `'`):

```python
regex = r'
```

I think there are a few others, but I don't recall off of the top of my head.

FichteFoll:
Auto-pairing was disabled when following the string modifiers.

Closes #1441.

### DIFF
--- a/Python/Default.sublime-keymap
+++ b/Python/Default.sublime-keymap
@@ -7,6 +7,7 @@
             { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
             { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "^|[^'a-zA-Z0-9_]$|(?i)\b[bfru]+$", "match_all": true },
+            { "key": "selector", "operator": "equal", "operand": "source.python" },
             { "key": "eol_selector", "operator": "not_equal", "operand": "string.quoted.double - punctuation.definition.string.end", "match_all": true }
         ]
     },
@@ -16,6 +17,7 @@
             { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
             { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "^|[^'a-zA-Z0-9_]$|(?i)\b[bfru]+$", "match_all": true },
+            { "key": "selector", "operator": "equal", "operand": "source.python" },
             { "key": "eol_selector", "operator": "not_equal", "operand": "string.quoted.single - punctuation.definition.string.end", "match_all": true }
         ]
     },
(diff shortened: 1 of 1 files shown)


FichteFoll:
Doesn''t shouldn't happen regardless through the preceding text context, but Tee selector is a good point. I just forgot actually adding it.

Not sure what you mean with the last sentence.

deathaxe:
> Doesn''t shouldn't happen

Actually it did here while writing some comments in python files using the keybindings of this PR. After adding ` - string - comment` to the selector,  it no longer happened.

Last sentence maybe off-topic, but all the auto-pair features include a `move` command, to avoid accidentally adding duplicated closing tags. Depending on coding habits this feature seems to eat brackets sometimes as the keybindings' contexts are not able to check for balanced brackets.

see: https://forum.sublimetext.com/t/st3-eats-right-parentheses-brackets-quotes/36339

FichteFoll:
>Actually it did here while writing some comments in python files using the keybindings of this PR.

Ah well, I borked the `preceding_text` regex. Fixed via 80d9429.

Python/Default.sublime-keymap
```diff
@@ -0,0 +1,24 @@
+[
+    // Auto-pair quotes even after string modifiers.
+    // Copied over from the default bindings with moficiations to `preceding_text`.
+    { "keys": ["\""], "command": "insert_snippet", "args": {"contents": "\"$0\""}, "context":
+        [
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "^$|[^\"a-zA-Z0-9_]$|(?i)\b[bfru]+$", "match_all": true },
```
deathaxe: Looks like the `\b` needs to be a `\\b` here to be applied as word-boundary. Furthermore I am wondering which function `^$` has? Works without it as well.

```
{ "key": "preceding_text", "operator": "regex_contains", "operand": "[^\"a-zA-Z0-9_]$|(?i)\\b[bfru]+$", "match_all": true },
```

FichteFoll: :man_facepalming: JSON gets me every time.

The `^$` is needed for the beginning of the line because there's no character to be matched in the second pattern.

deathaxe: Works fine without `^$` at beginning of a line or even at the very beginning of a file. Therefore I ask.

FichteFoll: So, i can't test this right now, but do you have any idea why? I'm not seeing a reason for it to work without ^$.

deathaxe: Looks like ST's default key binding applies at the beginning of a line with or without `^$` in Python's context.

FichteFoll: Right, I only need to apply the special bindings for the special case as the default bindings will still be active.

Fixed with 5951a1a8a9a9880424e4d586a72e810d9dd7e3cc.

deathaxe:
Works.

Issue #1498, opened 2018-04-12
[Bash] `readonly` not highlighted when prefixed with "local"


zx8:
`readonly` is not highlighted in the following code (notice GitHub _does_ highlight correctly):

```sh
local readonly a=b
```

Whereas Sublime doesn't:

![image](https://user-images.githubusercontent.com/8544900/38672832-46478884-3e47-11e8-9608-9577b9065f83.png)

zx8:
False alarm!

As [pointed out here](https://github.com/koalaman/shellcheck/issues/1178#issuecomment-383418057], I just created a local variable named `readonly`. The highlighting is in fact correct.

Confirmed by executing the following:

```sh
#!/bin/bash

set -o nounset

#!/bin/bash

function foo() {
  local readonly bar=baz
  echo "$bar $readonly"
}

foo
```

Yields:
```sh
foo.sh: line 9: readonly: unbound variable
```

rwols:
Woops! Thanks for pointing this out. I just checked if bash parsed it.

PR #1507, opened 2018-04-18
[ShellScript] Fix: readonly after local keyword

rwols:
closes https://github.com/sublimehq/Packages/issues/1498


zx8:
This highlighting is actually invalid! I was mistaken in my original issue, see https://github.com/sublimehq/Packages/issues/1498

PR #1511, opened 2018-04-22
[PHP] Fix anonymous class with arguments

jfcherng:
Anonymous classes can pass arguments through to their constructors.

Ref: http://php.net/manual/en/language.oop5.anonymous.php


wbond:
Thanks!

Issue #1516, opened 2018-04-25
[HTML] Consider not allowing '<' in attribute names


jcberquist:
Since https://github.com/sublimehq/Packages/pull/1421 and its looser rules for matching attribute names in HTML, syntaxes that use the HTML syntax such as ASP and PHP can no longer highlight code like the following correctly:
```php
<option<?php if($condition): ?> selected<?php endif; ?>></option>
```
(since `selected<?php` is now matched as an attribute name) 

A simple fix would be to add `<` to the `attribute_name` exclusion set at https://github.com/sublimehq/Packages/blob/master/HTML/HTML.sublime-syntax#L16

FichteFoll:
I suspect this will also be an issue for template languages using curlies.

Thom1729:
Hmm. The HTML5 spec is clear: `selected<?php` is a perfectly valid attribute name. I'd rather not regress that if we can avoid it.

The problem is that PHP et al want to interrupt mid-token. This doesn't work because we're parsing the entire attribute name using a regexp. We can use a context instead that parses the names character-by-character, and that should work exactly the same for HTML5 while giving syntaxes that embed HTML the chance to inject their own behavior in the middle of an otherwise-valid attribute name.

Thom1729:
I guess you could `clear_scopes`.

wbond:
I added 463d994ff3c27ea80f3cdf355c53b9a475afbd27 as a test to make sure we don't regress on PHP tags in the middle of HTML tags.

I don't think it is a huge deal if the PHP get the attribute scope. Technically it is correct, and allows for somewhat more correct highlighting. See the tests for an example where PHP is echoing part of an attribute name.

@jcberquist Do you want to check and see if this resolve the issue you had in your syntax?

jcberquist:
@wbond Yes, it does. Thanks!

FichteFoll:
>Technically it is correct

True, haven't looked at it like that. It might look weird for templating punctuation though since it's a colored scope that we're leaving here.

willrowe:
I can confirm this is working. It also fixes the issue I mentioned in [#1421](https://github.com/sublimehq/Packages/pull/1421#issuecomment-379346962).

jcberquist:
This is fixed, and 3166 is out, so I am closing this. I trust that is ok. Thanks for the fix @Thom1729!

PR #1517, opened 2018-04-26
[HTML] Parse attribute names character-by-character

Thom1729:
For #1516.


willrowe:
I'm not sure if this quite works, I seem to be losing the `entity.other.attribute-name.html` scope on the attribute name itself.

willrowe:
Sorry, I didn't have right version of the code. I just tried out the new release and it is working as expected.

PR #1519, opened 2018-04-27
[Python] Fix tuple argument highlighting

### LINKED ISSUE #990: Python initiate object with tuple - Showing syntax error
screen shot 2017-05-22 at 6 20 25 pm

deathaxe:
Fixes #990

Python 2 supports function definitions with tuble style parameters.

By PEP-3113 this style of function definitions was removed, but as python 2.7 is still a popular version being in active use, ST should still support that kind of function declaration.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -610,10 +610,42 @@ contexts:
             - match: '(?=[,)=])'
               set: function-parameters
             - include: expressions
+    - include: function-parameters-tuble
     - include: illegal-names
     - match: '{{identifier}}'
       scope: variable.parameter.python
     - include: line-continuation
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -610,11 +610,43 @@ contexts:
             - match: '(?=[,)=])'
               set: function-parameters
             - include: expressions
+    - include: function-parameters-tuble
     - include: illegal-names
     - match: '{{identifier}}'
       scope: variable.parameter.python
     - include: line-continuation
 
+  function-parameters-tuble:
```
FichteFoll: I think you mean 'tuple' here :wink:

PR #1525, opened 2018-04-30
Remove superfluous '+' in JavaScript syntax

sharkdp:
Removes a `+` on a negative-lookahead group that causes problems with some regex implementations.

See this short discussion with @Thom1729: https://github.com/sublimehq/Packages/pull/1427#discussion_r184985538


Thom1729:
The `+` does nothing, so removing it is a Good Thing. This change should not affect Sublime's highlighting, but it will avoid a bug in syntect.

PR #1528, opened 2018-05-05
[CSS] Add `will-change` to CSS completions

philipbordallo:
This will add completions for [`will-change `](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) as it exists in the syntax file but its completions were missing.


wbond:
Should `contents` be added to the list? https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

wbond:
Thanks!

Issue #1527, opened 2018-05-05
[C++] operator overload with template return type break syntax highlight


jasjuang:
I am on the latest sublime 3 dev build.

```
class Sample
{
 public:
 	
  T operator[](int a) 
  {
     // do something
  }
};
```

`operator` is highlighted as if it's a type and not a function name.

I did some tests and the below scenarios highlight the `operator` correctly.

```
class Sample
{
 public:

  int operator[](int a) 
  {
     // do something
  }
};
```
```
class Sample
{
 public:

  T func(int a) 
  {
     // do something
  }
};
```

Weirdly enough this is also correct too

```
class Sample
{
 public:
 	
  T &operator[](int a) 
  {
     // do something
  }
};
```

It seems like it's the combination of template return type and operator overload that breaks the syntax highlight, and without the reference. 

This is preventing plugins like doxydoxygen working properly https://github.com/20Tauri/DoxyDoxygen/issues/110 . It would be great to see this fixed!

PR #1530, opened 2018-05-05
[C++/ObjC] Fix: macro_identifier should be at least two characters long

rwols:
Closes https://github.com/sublimehq/Packages/issues/1527.

The `macro_identifier` variable is too greedy right now. It may match tokens like `T` or `A` or `B`. Defining single-character macros is extremely unwise and frowned upon. Moreover, the syntax may incorrectly parse an operator overload like this one:

```
template <class T> class Sample {
  T operator[](const int x);
};
```
Here, the `T` is assumed to be a macro, while it should be a proper return type (namely the template type `T`). This PR changes the `macro_identifier` variable so that only identifiers with at least two characters get parsed as a macro.

It is still possible to trip up the syntax with two-character template parameters like `T1` or `T2`.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -22,7 +22,7 @@ scope: source.c++
 variables:
   identifier: '\b[[:alpha:]_][[:alnum:]_]*\b'
   path_lookahead: '(?:::\s*)?(?:{{identifier}}\s*::\s*)*(?:template\s+)?{{identifier}}'
-  macro_identifier: '\b[[:upper:]_][[:upper:][:digit:]_]*\b'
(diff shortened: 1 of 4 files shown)


jasjuang:
Can we change it so it's at least three characters long? `T1` or `T2` is still very possible to be used whereas I believe double-character macros is also extremely unwise and frowned upon.

rwols:
@jasjuang, while working on a fix for issue https://github.com/sublimehq/Packages/issues/1537, it turns out that PR https://github.com/sublimehq/Packages/pull/1559 also fixes this issue. In fact it's even better because like you said, I made the macro_identifier variable at least 3 chars there (inherited from the C syntax). I'll keep this open for now because PR https://github.com/sublimehq/Packages/pull/1559 does not include the syntax tests I wrote here.

PR #1533, opened 2018-05-05
[C++] Remove redundant meta.group.c++ scope from brace

### LINKED ISSUE #1406: [C++] Redundant meta.group scope for methods without parameters in class definition
In the following case, it seems that the closing round brace in the method's signature gets an extra `meta.group.c++` scope, which seems wrong. I discovered it while trying to create a color scheme that draws pairs of nested braces with different colors:

<table>
<tr><td>
Code Sample
</td></tr>
<tr><td>
<pre lang="c++">
struct Lemonade : public MassiveDropAttack
{
    // See how the closing brace gets TWO meta.group.c++ scopes instead of one
    virtual int PureFunc() override
    {
        return 10;
    }
};
</pre></td></tr>
</table>

<table>
<tr><td>
Screenshot
</td></tr>
<tr><td>

Screenshot demonstrating the bug

</td></tr>
</table>

<table>
<tr><td>
Relevant part in the color scheme
</td></tr>
<tr><td>
<pre lang="XML">
&lt;dict&gt;
    &lt;key&gt;name&lt;/key&gt;
    &lt;string&gt;Braces Level 1&lt;/string&gt;
    &lt;key&gt;scope&lt;/key&gt;
    &lt;string&gt;meta.group punctuation.section.group&lt;/string&gt;
    &lt;key&gt;settings&lt;/key&gt;
    &lt;dict&gt;
        &lt;key&gt;foreground&lt;/key&gt;
        &lt;string&gt;#00D93D&lt;/string&gt;
    &lt;/dict&gt;
&lt;/dict&gt;
&lt;dict&gt;
    &lt;key&gt;name&lt;/key&gt;
    &lt;string&gt;Braces Level 2&lt;/string&gt;
    &lt;key&gt;scope&lt;/key&gt;
    &lt;string&gt;meta.group meta.group punctuation.section.group&lt;/string&gt;
    &lt;key&gt;settings&lt;/key&gt;
    &lt;dict&gt;
        &lt;key&gt;foreground&lt;/key&gt;
        &lt;string&gt;#EAFF00&lt;/string&gt;
    &lt;/dict&gt;
&lt;/dict&gt;
</pre></td></tr>
</table>

What do you think? Is this really a bug, or is it the right syntax behaviour, and I'm just using it wrong in my color scheme?

Thank you for your time,
Have a lovely weekend :smile:

### LINKED ISSUE #1500: [C++] Open/Closing braces within class definition get different scopes
With the newest C++ syntax highlighting available, when coloring the braces based on the number of the number of `meta.group.c++` scopes they contain, the braces of method declaration seems to get different scopes:

screen

That seems like a bug in the syntax file..

papadokolos:
I found that a closing brace found inside a class definition gets a redundant `meta.group.c++` which separates it from the openining brace. This should not happen, since the opening and closing braces should have the same amount of `meta.group.c++` scopes.

Fix #1406 
Fix #1500 

I struggle to find the appropriate way to write syntax tests for this. Would you mind helping me?

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -1341,7 +1341,7 @@ contexts:
           set:
             - meta_content_scope: meta.method.parameters.c++ meta.group.c++
             - match : \)
-              scope: meta.group.c++ punctuation.section.group.end.c++
+              scope: punctuation.section.group.end.c++
               set: method-definition-continue
             - match: '\bvoid\b'
               scope: storage.type.c++
--- a/C++/syntax_test_cpp.cpp
+++ b/C++/syntax_test_cpp.cpp
@@ -148,6 +148,12 @@ struct FOO1 FOO2 FOO3 Test {
   /*                               ^ storage.modifier */
 }
 
+struct X {
+  X();
+   /* <- meta.group */
+  /*^ meta.group - meta.group meta.group */
+};
+
 #define DEPRECATED(msg) [[deprecated(msg)]]
 
 struct Test {
--- a/Objective-C/syntax_test_objc++.mm
+++ b/Objective-C/syntax_test_objc++.mm
@@ -147,6 +147,12 @@ struct FOO1 FOO2 FOO3 Test {
   /*                               ^ storage.modifier */
 }
 
+struct X {
+  X();
+   /* <- meta.group */
+  /*^ meta.group - meta.group meta.group */
+};
+
 #define DEPRECATED(msg) [[deprecated(msg)]]
 
 struct Test {


rwols:
You should check if this bug is present in the C, Objective-C and Objective-C++ syntaxes, too.

papadokolos:
@rwols you're right! From a brief check it seems to come up in several other places as well. I identify the errors by simply searching for this (wrong) combination:

```YAML
- match : \)
  ... scope: meta.group.c++ ...
```

wbond:
@papadokolos Thanks for your submission here. Did you ever get a chance to fix the other occurrences of this you found?

papadokolos:
@wbond I gave it a shot, and "blindly" removed all occurrences of the "erroneous" pattern from the syntax files. Unfortunately, it messed up the syntax highlighting...

I'm certain that this pattern should never occur - a closing paren should never get that extra `meta.group` scope. It think it's illogical.

This means that there is a deeper problem in the syntax file/s regarding `meta.group` scopes, which requires the work of a more proficient contributer than I.

wbond:
@papadokolos Thanks for this fix!

PR #1539, opened 2018-05-07
[HTML] Accept all recognized JavaScript MIME types in <script> tags.

### LINKED ISSUE #1536: [HTML] Syntax highlighting broken for certain values of the script's type attribute
Just upgraded to build 3170 and noticed that JavaScript inside a `<script>` tag where the type is `application/javascript` is no longer rendered as JavaScript syntax but instead is just plain text. This doesn't conform to HTML spec (https://html.spec.whatwg.org/multipage/scripting.html#the-script-element) which states any JavaScript essence MIME type match (https://mimesniff.spec.whatwg.org/#javascript-mime-type) should be considered JavaScript.

I noticed the same issue with JSON embedded inside scripts (specifically with the type `application/ld+json`).

Thom1729:
Fixes #1536.

[Reference](https://mimesniff.spec.whatwg.org/#javascript-mime-type).

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -32,6 +32,15 @@ variables:
       button|datalist|input|label|legend|meter|optgroup|option|output|progress|select|template|textarea
     )\b
 
+  javascript_mime_type: |-
+    (?ix)(?:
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

Issue #1535, opened 2018-05-07
Bash scripts display error when variable declared in string


Mono-Forums:
Hello -

I have been using Sublime Text for a while for bash scripts and its been great.
The new build 3170 is giving some errors, however.
...
When I have a statement of the format:
printFunction "$variableString1" "$(declare -p variableArray)"

Where the situation above is:
-- The printFunction is my user-defined function.
-- And there is a declare statement in the print string, here passing an array to the printFunction.

Everything below the line appears in yellow -- like a string. Even though the double quote ends after $variableArray
...
Please note this error does not occur if, there are two strings being passed to the function.
printFunction "$variableString1" "$variableString2"
Its only where a variable is declared as part of the call to the printFunction.
...
Or with one string echo command, ie a line ending with a quoted string.
echo "------------------------"
...
Hope this helps in clearly stating the issue.
I hope this will be fixed soon as it is making all my bash scripts unreadable.
Really appreciate the great product Sublime has been for bash.
Please feel free to reach me with any questions.

Thank you.

Mono

PR #1541, opened 2018-05-07
[ShellScript] Fix vardef-name context not popping

rwols:
For example, in
```
printFunction "$variableString1" "$(declare -p variableArray)"
```
the token `variableArray` is scoped with the vardef-name context, and it should
pop when encountering the closing parenthesis. We solve this by looking ahead
for a metacharacter.

In general more gratuitous popping in the vardef-* contexts is applied in order
to not get stuck in an unwanted context.

Closes https://github.com/sublimehq/Packages/issues/1535.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -380,6 +380,8 @@ contexts:
 
   vardef-maybe-more:
     - meta_include_prototype: false
+    - match: (?=`)
+      pop: true
     - match: (?=\s*#)
       pop: true
     - include: cmd-args-boilerplate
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

rwols:
Hi @Mono-Forums, it should become available in the next stable build. In meantime as a work-around, you can copy [this file](https://github.com/sublimehq/Packages/blob/master/ShellScript/Bash.sublime-syntax) to your Packages/User folder (call it something like Bash-Temp.sublime-syntax). You might also want to change the `name: ` key in that file to something else, so you can distinguish it in the command palette. You should delete that file once a new stable build is available. Alternatively, you can clone this repository and make a symlink to the ShellScript folder in your Packages folder, as explained in the readme.

Mono-Forums:
Appreciate it Raoul. Will try it out. Will update here on how it goes.

Thank you.

Mono-Forums:
Thanks for the update Will and Raoul. Version 3.1.1/ B 3176 fixes this issue.

Much appreciated.

Mono

PR #1544, opened 2018-05-08
[C#] fix bug when "dynamic" is used as the return type of a method

### LINKED ISSUE #1534: [csharp] Incorrect c# highlighting inside markdown
C# highlighting has long standing problems for me. I'm encountering them mostly in markdown files. I'm using MarkdownEditing package for markdown. The author said that it's not a problem with markdown-plugin itself but with C# highlight package - 
https://github.com/SublimeText-Markdown/MarkdownEditing/issues/510

Sublime 3.1 (Build 3170)

An example
image

keith-hall:
fixes #1534


wbond:
Thanks!

PR #1548, opened 2018-05-09
[PHP] fix bug whereby indentation isn't decreased with `<? }`

keith-hall:
as reported at https://forum.sublimetext.com/t/sublime-text-3-1/36591/34


wbond:
Thanks!

PR #1549, opened 2018-05-09
[PHP] Added functions password_* for PHP >= 5.5

exploitmik:


wbond:
Thanks for adding these!

PR #1550, opened 2018-05-09
[HTML] The .inc (and some others) extension could be generally any language

jfcherng:
`inc` means `include`, which is not a language-specific term.
`tpl` and `tmpl` which mean `template` are the same.

But I am **NOT** sure about this change. Maybe people use these file extensions in HTML more? Feel free to reject this PR.

Related: https://forum.sublimetext.com/t/first-line-match-mode-shell-script/36742/3

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -6,7 +6,6 @@ file_extensions:
   - htm
   - shtml
   - xhtml
-  - inc
   - tmpl
   - tpl
 first_line_match: (?i)<(!DOCTYPE\s*)?html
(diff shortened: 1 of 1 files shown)


Thom1729:
I dimly remember using it in some ancient server-side include system back in the stone age. But all sorts of systems in all sorts of languages could use `.inc` files; I don't think that HTML has any particular claim to that file extension. Removing it is perfectly reasonable. If anyone still needs it, they can use `Open all with current extension as...` to restore the old behavior.

I wonder if the same applies to `.tpl` and `.tmpl`. What are these even for? Has anyone run into one of these in the wild?

jfcherng:
@Thom1729 I did see people use `.tpl` (template) in some old PHP files. But nowadays PHPer hardly do that.

So, that's do more here if this PR is accepted.

PR #1553, opened 2018-05-09
[Perl] regular expressions /r option added in perl 5.14.

Um9iIEdvcmRvbg:


Um9iIEdvcmRvbg:
Looks like TheClams [commit](https://github.com/sublimehq/Packages/pull/1556) is way better than this one. This is my first pull request so I’m not sure how to cancel this pull request.

Issue #1495, opened 2018-04-08
Perl. Syntax highlight. // operator as a comment


axlewdev:
After // operator all lines is highlighted as comment.
Version 3.0 3143

alexqfredrickson:
+1 as I am facing this issue as well in Sublime Text 3.0 (Build 3143).

This also happens with `split / /`, for example:

    use XML::XPath;

    $file_handler = XML::XPath->new(filename => "$filename");
    my @config_info = split / /, get_configuration( "//foo/config", $file_handler );

    sub get_configuration {
        my $param1 = shift;
        my $param2 = shift;
        return "configuration";
    }

deathaxe:
Should be fixed by the merged PR #1564

KES777:
Is there instruction which steps I should take to be updated to latest version?

rwols:
@KES777:

1) clone this repository like any other repository, say /path/to/Packages
2) make a symlink to /path/to/Packages/Perl from
  - $XDG_CONFIG_HOME/sublime-text-3/Packages/Perl (linux), or
  - ~/Library/"Application Support"/"Sublime Text 3"/Packages/Perl (osx), or
  - %APPDATA%\Sublime Text 3\Packages\Perl (windows)

This is just for Perl, you can make more symlinks for other languages.

KES777:
1. I did that. Now code is pink =) May you advice some theme editor so I can fix colors?

2. The same part of code may be highlighted differently:

![delete1](https://user-images.githubusercontent.com/799021/41648896-d674d06c-7482-11e8-97db-c03fd6a41901.jpg)

3. Regular expressions are not highlighted correct

![delete2](https://user-images.githubusercontent.com/799021/41648924-e5a60c22-7482-11e8-98a2-e69dd3b1500e.jpg)
as you can see third and forth lines are broken

KES777:
The text of target sub:
```
sub _is_email {
  state $email_rfc5322_re = do {
    my $atom           = qr;[a-zA-Z0-9_!#\$\%&'*+/=?\^`{}~|\-]+;o;
    my $quoted_string  = qr/"(?:\\[^\r\n]|[^\\"])*"/o;
    my $domain_literal = qr/\[(?:\\[\x01-\x09\x0B-\x0c\x0e-\x7f]|[\x21-\x5a\x5e-\x7e])*\]/o;
    my $dot_atom       = qr/$atom(?:[.]$atom)*/o;
[...]

Issue #1326, opened 2017-11-24
Perl variable highlighting with variables that start with an _ (underscore) is broken


gullevek:
Variables that are named $foo are ok, but if they are named $_foo they do not highlight correctly. This is also the case for @ or % arrays/hashes.

compare screenshot between vim and Sublime Text 3 built 3143

![screen shot 2017-11-24 at 14 30 26](https://user-images.githubusercontent.com/855861/33196673-0eff3132-d124-11e7-91df-ac82095188c0.png)
![screen shot 2017-11-24 at 14 30 31](https://user-images.githubusercontent.com/855861/33196674-0f29af3e-d124-11e7-8d71-a898943d8772.png)

michaelblyons:
Looks good now.

``` perl
#!/usr/bin/perl

$foo = 'bar';
$_foo = 'bar';
@bar = ();
@_bar = ();
%foobar = ();
%_foobar = ();

${%_foobar} = '';
%{$foobar} = 12;
%{$_foobar} = 13;
%{$foo{'bar'}{'bar'}} = 'excl';
```

![Screen Shot 2019-07-30 at 1 51 27 PM](https://user-images.githubusercontent.com/2701605/62152733-262e7a80-b2d1-11e9-91ed-ae0ef8f54992.png)

Issue #1325, opened 2017-11-24
Perl highlighting issue in regexp with newer (perl 5.20) modifiers


gullevek:
It seems that a lot of the new modifiers in the perl regex are not supported. Like /p, /xx, /a, /d, /l, /u, /n, /r

When using one of those in a perl regex it will break any code highlighting after this regex.

Also simple match =~ // highlighting seems to be a bit broken

Below compare shot between vim and Sublime Text 3 built 3143

![screen shot 2017-11-24 at 14 28 05](https://user-images.githubusercontent.com/855861/33196629-bf054c48-d123-11e7-8fc7-9caa8350b8fc.png)
![screen shot 2017-11-24 at 14 28 14](https://user-images.githubusercontent.com/855861/33196630-bf399840-d123-11e7-8fff-574c6e31d2bc.png)

Issue #541, opened 2016-07-21
Bad coloring for Perl regexes beginning with a space


zort:
Perl regexes beginning with a space (or any whitespace) don't get colored properly. I'm using build 3114.

Good:
![1](https://cloud.githubusercontent.com/assets/535865/17036390/1cd045aa-4f5a-11e6-9cb8-7be0889119d1.PNG)

Bad:
![2](https://cloud.githubusercontent.com/assets/535865/17036396/20990ab4-4f5a-11e6-8e35-3f8081deae4e.PNG)

Worse if you have a tilde before:
![3](https://cloud.githubusercontent.com/assets/535865/17036410/2e41f4b4-4f5a-11e6-89d3-5ae3aac6ee92.PNG)

I think the culprit is this regexp: https://github.com/sublimehq/Packages/blob/a493e90787e876d4f221e2bf621b47cb3505b99a/Perl/Perl.sublime-syntax#L342

To anyone else who runs into this, I solved it by installing https://packagecontrol.io/packages/ModernPerl and doing View > Syntax > Open all with current extension as ... > ModernPerl.

wbond:
Yes, regex literals can be kind of a pain. We fixed them in JavaScript by tracking the state more explicitly, so a `/` following an operator or other non-identifier is considered the start of a regex, and never a division operator.

The rules are slightly more complicated than that to deal with them being passed as parameters, and allowing division to happen on function call return values, but you get the idea.

PR #1556, opened 2018-05-10
[Perl] Fix a few regexp related issues and variable with underscore

TheClams:
Fix issue #541 with  space at the beginnign of a regexp
Fix issue #1495 and #1594 where // can break the syntax highlighting
Fix issue #1325 and #1029 related to regexp modifier

Fix issue #1326 with variable starting with underscore

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -244,16 +244,16 @@ contexts:
             - include: escaped_char
             - include: variable
         - match: \s+
-        - match: '((([egimosx]*)))(?=(\s+\S|\s*[;\,\#\{\}\)\]>]|$))'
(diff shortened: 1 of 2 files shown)


deathaxe:
Nice, job.

As Perl syntax still "suffers" from Oniguruma I decided to refactor it a week ago, to let it use the latest features and best possible performance.

It works very well right now fixing all the mentioned issues as well and even much more which I found while reading though a couple of available Perl scripts.

Once I am done with the preparations, I look forward for willing testers to find the remaining issues.

TheClams:
Yes it definitely needed a rework, but I could not find the courage to do it ;)

wbond:
@TheClams Have you looked at the work in #1564? If so, I presume it would be safe to close this?

wbond:
Upon further investigation, it was mentioned in #1564 that it covers this, so I am closing. I do appreciate your work, though, @TheClams.

PR #1558, opened 2018-05-10
[C++/ObjC] Fix: '-->' character sequence was not recognized as '--' and '>'

### LINKED ISSUE #1557: [C++] `-->` is not highlighted correctly
For the code:

```
int main(){
	int a=5,b=0;
	while(a-->0)++b;
}
```

the `-->` part is not highlighted correctly (the first `-` is red and the `->` is gray), and the last `}` is colored in red.

Image

As explained [here](https://stackoverflow.com/questions/1642028/what-is-the-operator-in-c), that is 2 operators `--` and `>` (that is, `while((a--)>0)`).

rwols:
fixes #1557

### DIFF
--- a/Objective-C/syntax_test_objc.m
+++ b/Objective-C/syntax_test_objc.m
@@ -2,7 +2,7 @@
 
 int main(){
     int a=5,b=0;
-    while(a-->0)++bπ;
+    while(a-->0)++b;
     /*     ^^ keyword.operator.arithmetic */
     /*       ^ keyword.operator.comparison */
     /*        ^ constant.numeric */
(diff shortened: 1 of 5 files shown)


Objective-C/syntax_test_objc.m
```diff
@@ -1,5 +1,14 @@
 /* SYNTAX TEST "Packages/Objective-C/Objective-C.sublime-syntax" */
 
+int main(){
+    int a=5,b=0;
+    while(a-->0)++bπ;
```
rwols: A pie got in here somehow...

PR #1560, opened 2018-05-10
[JavaScript] Regexp cleanup

### LINKED ISSUE #1554: [JavaScript] Some identifiers are being scoped as separate tokens
### Summary

While reviewing some JavaScript code, I noticed that some keywords were being highlighted in sections with no special meaning. Later I noticed that the same issue was happening in other parts of the view, not just the editor, the "Find…" form, for example. Notice that this only happens with the Celeste UI color scheme, I tried other options and the issue is not reproducible.

### Expected behavior

* Keyword `foo.id` should be colored all in blue.
* Text in search form should not be highlighted.

### Actual behavior

* Keyword `foo.id` is highlighted in blue, but the letter `i (eye)` contains a different color.
* Text in search form is being highlighted with different shades of blue and purple.

aaa

### Steps to reproduce

1. Create a new _(empty)_ tab.
2. Copy and paste the code below.
3. Select menu "Find > Find…"
4. Search the word `history`

```
// Code to showcase the bug.
function foobar(foo) {
    return Foobar.fetch(uuid, foo.id);
}
```

### Environment

* Mac OS 10.13.4 _(x86_64)_
* Sublime Text 3.1 Build 3170

Thom1729:
Fixes #1554.

Made a number of small tweaks and fixes:

- Removed unused regexp capture groups, either by removing them or making them non-capturing. Because the engine breaks tokens at capture group boundaries, these spurious groups were interfering with hash-based coloring.
- Expanded out a few dense regexps that were hard to read as one-liners.
- Wrapped regexp variables in non-capturing groups, just in case.
- Fixed a bug in string newline escaping; added a test.
- Simplified a few rules in strings.
- Improved the `new.target` implementation to use the new identifier lexing from #1427.

Is there an existing `sublime-syntax` linter? Now that token boundaries are more important, it would be nice to have a rule preventing match expressions from containing unused capture groups.

Also, the `support` names in the syntax are a mess. It's trying to match way too many built-in functions, properties, and types, including many that are specific to browsers or to Node. This is probably not what the core syntax should be doing. Marking all `id` properties `support.constant.dom` is going to result in a *lot* of false positives, and the principle applies generally. What's more, keeping those lists up to date on one hand is a hopeless challenge and on the other would exacerbate the problem — what common name *isn't* the name of some property or other on some built-in object?

I think that we should remove most of the `support` names, leaving only the functions, types, and constants built into the language and excluding property names. (We can make a few key exceptions, like `window` and `module`.) If a user wants to mark a bunch of extra environment-specific names, then a third-party package can accommodate them. (Yes, I'm volunteering.)

The above does not apply to the dollar sign (addressed in #1339), which is not generally marked `support`.


Thom1729:
@wbond At the rate that the minor JS fixes are piling up, I'm concerned about possible merge conflicts. If this issue can be merged early, then I can make sure that everything else is in good shape before the next build.

wbond:
I think the suggestion about highlighting the `support` scopes probably makes sense, but let's open an issue for that where other people can weigh in, if an issue doesn't already exist.

Thom1729:
#1563

Issue #1565, opened 2018-05-11
[HTML] Nested angle brackets <> not highlighted correctly inside DOCTYPE tags


evandrocoan:
`text.html.basic meta.tag.sgml.html meta.tag.sgml.doctype.html entity.name.tag.doctype.html `

![image](https://user-images.githubusercontent.com/5332158/39949413-0da90478-5551-11e8-879b-c0f48d79a0b0.png)

```html
<!DOCTYPE html [
    <!ENTITY nbsp "&#160;">
]>
```

deathaxe:
Neither XML nor HTML support any kind of DTD within the doctypes. PR #985 attempted to add support for it to the XML language without conclusion. It could easily be extended to be used by HTML as well if someone was interested in.

wbond:
This issue has been solved by a previous commit, hence it is being closed.

PR #1566, opened 2018-05-12
[HTML] Fix DOCTYPE and improve CDATA highlighting

deathaxe:
This commit introduces/proposes the following changes:

1. A fix for issue #1565 which is caused by not handling all the DOCTYPE internal syntax elements.
2. Highlight the CDATA tag as self closing xml/html tag.
3. Don't handle unknown `<! ... >` tags leaving them to special purpose syntaxes, which include this general purpose html syntax.

Point 2. and 3. are not directly related to #1565 but were part of handling opening/closing brackets within special tags. CDATA was the only remaining one within the `<!` context, which somehow made no sense but putting it into another PR would have caused conflicts.

For details about the changes please refer to the commit messages as this PR is split for better readability.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -41,6 +41,8 @@ variables:
       | text/livescript
     )
 
+  tag_name: '[[:alpha:]_][[:alnum:]:_.-]*'
+
 contexts:
   immediately-pop:
     - match: ''
@@ -52,6 +54,7 @@ contexts:
 
   main:
     - include: comment
+    - include: doctype
     - match: (<\?)(xml)
       captures:
         1: punctuation.definition.tag.begin.html
@@ -71,14 +74,6 @@ contexts:
         - match: ">"
           scope: punctuation.definition.tag.html
           pop: true
-        - match: (?i:DOCTYPE)
-          scope: entity.name.tag.doctype.html
-          push:
-            - meta_scope: meta.tag.sgml.doctype.html
-            - match: (?=>)
-              pop: true
-            - match: '"[^">]*"'
-              scope: string.quoted.double.doctype.identifiers-and-DTDs.html
         - match: '\[CDATA\['
           push:
             - meta_scope: constant.other.inline-data.html
@@ -243,6 +238,48 @@ contexts:
           pop: true
         - match: '<!--(?!-?>)|--!>'
           scope: invalid.illegal.bad-comments-or-CDATA.html
+
+  doctype:
+    - match: (<!)(DOCTYPE)
+      captures:
+        1: punctuation.definition.tag.begin.html
+        2: keyword.declaration.doctype.html
+      push:
+        - doctype-meta
+        - doctype-content
+        - doctype-content-type
+        - doctype-name
+
+  doctype-meta:
+    - meta_scope: meta.tag.sgml.doctype.html
+    - match: '>'
+      scope: punctuation.definition.tag.end.html
+      pop: true
+
+  doctype-name:
+    - match: '{{tag_name}}'
+      scope: variable.other.doctype.html
+      pop: true
+    - include: else-pop
+
+  doctype-content-type:
+    - match: (?:PUBLIC|SYSTEM)\b
+      scope: keyword.content.external.html
+      pop: true
+    - include: else-pop
+
+  doctype-content:
+    - match: \[
+      scope: punctuation.section.brackets.begin.html
+      set:
+        - meta_scope: meta.brackets.html meta.internal-subset.xml.html
+        - match: \]
+          scope: punctuation.section.brackets.end.html
+          pop: true
+        - include: comment
+    - include: string-double-quoted
+    - include: string-single-quoted
+    - include: else-pop
 
   style-css:
     - meta_content_scope: meta.tag.style.begin.html
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -261,7 +261,7 @@ contexts:
 
   doctype-name:
     - match: '{{tag_name}}'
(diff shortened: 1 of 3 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -63,34 +64,13 @@ contexts:
...
-        - match: '<!--(?!-?>)|--!>'
-          scope: invalid.illegal.bad-comments-or-CDATA.html
     - match: <!
       scope: punctuation.definition.tag.html
       push:
         - meta_scope: meta.tag.sgml.html
         - match: ">"
           scope: punctuation.definition.tag.html
           pop: true
-        - match: (?i:DOCTYPE)
```
deathaxe: Remarks: This block should have been removed by the next commit.

pittAndrews:
Am from testing changes made on HTML concerning the highlighting issue using:
$ git clone https://github.com/sublimehq/Packages.git
$ ln -s `pwd`/Packages/HTML ~/.config/sublime-text-3/Packages/

Still does not solve the issue, still breaks color when commenting within scripts of type text/html

deathaxe:
Yes, because this PR does not address comment highlighting within scripts.

The commit be2905b just enables a fix for it as it creates a dedicated context for `comment`s which could be included wherever needed, even though the `<script type="text/html">` might not even need it.

deathaxe:
#### `<!DOCTYPE ... >`

1. Recently found that `meta.tag.sgml` is already addressed by a couple of color schemes such as "ayu". From that point of view, renaming it might not be the best idea in the end. 

2. To The `variable.documentroot` is renamed to `constant.language.doctype` by this PR at the moment. To follow the naming guidelines and express the meaning of that token best, maybe `variable.other.documentroot` or `variable.language.documentroot` might be better solution?

Thoughts?

#### `<![CDATA[...]]>`, ...

All dtd declaration  tags (`<!...`) use `keyword` rather then `entity.name.tag` in the xml syntax. So I wonder, whether

1. rename all those `keyword` to `entity.name.tag` or
2. keep using `keyword` but rename them to `keyword.declaration` as it is suggested by PackageDev as "convention"

The later solution might result in CDATA to be tagged as `keyword` instead of `entity` as well for consisteny.

Thoughts?

deathaxe:
I've reworked the commits of this PR to stay a little closer to the original tag names and fix some smaller regressions. 

The DOCTYPE now pushes multiple contexts to the stack in order to enforce the real order of the keywords.

According to the XML syntax `CDATA` and `DOCTYPE` are scoped as `keyword.declaration`.

HTML/HTML.sublime-syntax
```diff
@@ -241,6 +211,78 @@ contexts:
...
+      pop: true
+
+  doctype-name:
+    - match: '{{tag_name}}'
+      scope: variable.other.doctype.html
+      pop: true
+    - include: else-pop
+
+  doctype-content-type:
+    - match: (?:PUBLIC|SYSTEM)\b
```
Thom1729: Is `\b` sufficient here? Could `PUBLIC$` be a valid token, for instance?

HTML/HTML.sublime-syntax
```diff
@@ -41,6 +41,8 @@ variables:
       | text/livescript
     )
 
+  tag_name: '[[:alpha:]_][[:alnum:]:_.-]*'
```
Thom1729: https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state

It might be worthwhile to mimic the standard parsing algorithm, which allows almost anything after the first character of a tag name.

deathaxe: Took that entity from the existing XML syntax. Are there differences?

HTML/HTML.sublime-syntax
deathaxe: Those keywords must normally be followed by whitespace. So `\b` is sufficient here.

HTML/HTML.sublime-syntax
Thom1729: Yes. The HTML5 spec defines different HTML and XML parsing modes.

deathaxe: But looks like HTML is more permissive with DOCTYPE names than XML.

https://html.spec.whatwg.org/multipage/parsing.html#doctype-name-state

wbond:
`constant.language.doctype` feels more correct than `variable.*` to me.

I'd probably vote to keep `CDATA` as some specialization of `keyword`.

I would not couple HTML to XML.

Otherwise are you happy with this PR @deathaxe?

deathaxe:
> constant.language.doctype feels more correct than variable.* to me.

I chose `variable.` because ...

1. it is used in all declaration tags of the XML syntax right after the opening tag name. 

```
    <!ENTITY name ... >
             ^^^^ variable.entity
```

   so the doctype root would follow that.
   
```
    <!DOCTYPE html ...>
              ^^^^ variable.
```

2. The `html` specifies the root xml tag being used as valid entry point for the xml tree. Is this really a language constant? Basically one could specify anything there followed by a valid FPI and URL to define the entry point of an XML.

On the other hand `constant.language` ...

1. would ensure it to be colored differently than normal variables,
2. is the only valid value in HTML files and
3. was my first choice, too. 

So I would be happy with it, too.


> I'd probably vote to keep CDATA as some specialization of keyword.

Agree. As `entity.name.tag` it would disappear the same way it did as `string`.


> I would not couple HTML to XML.

Agree. It was just an idea of modularization, but I've learned XML to still require much more strict naming guidelines than HTML, which makes coupling impossible. It would also increase complexity as such kinds of dependencies between languages are hard to track and may cause regressions in the future.

deathaxe:
Changed the scope to `constant.language.doctype`.

PR #1567, opened 2018-05-13
 [HTML] More precise html-end-tag scoping

deathaxe:
Whitespace is no punctuation and should therefore not scoped as such.

This commit changes the `tag-end-maybe-self-closing` to no longer include the preceding space into the `punctuation.tag.end.html`.

The reason for it to be included here is not obvious anyway as none, one or any number of spaces would cause the `/?>` to be detected as `end of tag` anyway.

_Note:_ In order to reduce duplication and increase code reuse dedicated contexts for the 3 different tag-end punctuation rules were created before fixing the issue.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -623,6 +623,6 @@ contexts:
       pop: true
 
   tag-end-maybe-self-closing:
-    - match: '(?: ?/)?>'
+    - match: '/?>'
       scope: punctuation.definition.tag.end.html
       pop: true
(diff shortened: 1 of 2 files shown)


keith-hall:
It looks like my [fix for `style` reindentation](https://github.com/sublimehq/Packages/pull/1160) was broken in https://github.com/sublimehq/Packages/pull/1379. In hindsight, I should have added some comments or something explaining why the `\s*` was there, but anyway, it seems relevant to this PR, maybe we can include a fix here - the question is, what would be the best way to work around this indentation problem so that the code can stay de-duplicated but also work as desired in all situations?

Maybe just reapplying my change to https://github.com/sublimehq/Packages/pull/1567/files#diff-5fe623186009322c73eeb06682f1b128R616 won't cause any problems elsewhere?

Thom1729:
I've fixed the indentation issue in #1591.

Issue #1569, opened 2018-05-13
[RFC] [JavaScript] Remove some nonstandard scopes


Thom1729:
While looking into #1563, I've run into a number of nonstandard scopes that don't seem to be serving any purpose.

`variable.other.object`. Every JavaScript value is an object. (Internally, primitive values are autoboxed, but the distinction is not syntactically significant.) However, the syntax definition will sometimes scopes a name `variable.other.object` if the name is followed by a period. This distinction is at best vacuous and at worst misleading. In addition, the `variable.other.object` scope:

- Is not standard.
- Is not used in any other default syntax.
- Is not used in any default color scheme.
- Is not applied consistently.
- *Cannot* be applied consistency because of lookahead limitations.
- Is implemented in a way that interferes with some `support` scopes that should be applied.
- Does not appear to be used by any popular themes.

I would replace it with `variable.other.readwrite`, which is used generally for JavaScript names.

The `meta.prototype.declaration` and `meta.prototype.access` scopes are used to mark some assignment expressions involving the `prototype` property. This is not a real syntactic distinction; I suspect that it might have been used for the symbol list a long time ago. (It isn't now.) The same problems apply to these scopes: they're nonstandard, nothing else uses them, they can't possibly be made to work reliably, and they interfere with `support` scopes.

The `meta.instance.constructor` scope is used to scope expressions that are the target of the `new` operator. The same objections apply.

I would simply eliminate these three scopes, saving a few dozen lines of code and improving highlighting of `support` scopes.

If anyone knows of any reason why these scopes shouldn't be eliminated, please let me know.

Issue #1563, opened 2018-05-11
[RFC] [JavaScript] Reorganize/update `support` scopes.


Thom1729:
The JavaScript syntax definition contains special handling for a large number of built-in types, object, functions, methods, and properties, generally marked with `support` scopes. This is useful and important in many cases, but the current implementation (mostly predating the `sublime-syntax` transition) suffers from several problems:

1. It's outdated. Many modern items are missing.
2. There's no clear policy. It's a mix of language builtins, environment-specific functions and classes, and nonstandard miscellany.
3. The lists are poorly organized, often golfed, and generally hard to read.
4. The giant lists of property names are overeager, wrongly highlighting common property names like `id` in irrelevant contexts.
5. In many cases, recognition relies too much on regexp captures, so the implementation is unnecessarily unreliable.
6. Test coverage is nearly nonexistent.

To fix these problems, we should start by establishing a coherent policy.

1. From what source should we draw names?
2. What kinds of name should we try to support?
3. In ambiguous cases, should we err on the side of marking or not marking?

For instance, we could:

1. Support:
    - Names from the ECMAScript standard, such as `undefined`, `Math.sqrt`, and `decodeURI`.
    - A small fixed set of browser- or Node-specific names, such as `window` and `__filename`.
2. Mark:
    - Built-in classes, objects, and functions
    - Static properties of built-in classes and objects.
    - Properties of `Object.prototype`.
3. Avoid ambiguity by not marking other properties (like `id` or `slice`).

Once we choose a sensible policy, I can implement it immediately. Ideally, we would implement these in a way that made it easy for [third parties](https://github.com/Thom1729/Sublime-JS-Custom) to extend them to provide more detailed environment-specific highlighting.

It's worth noting that the new policy would almost certainly exclude some names that are currently highlighted. In some theoretical sense, this is a breaking change. However, it is extremely unlikely that any user is depending on `NO_MODIFICATION_ALLOWED_ERR` being highlighted. On the other hand, it is extremely common to use property names like `id`, and marking them as `support` creates far more false positives than true ones.

Years ago, JavaScript was a language that run almost exclusively in web browsers. Today, alternate environments like Node are just as important. Sublime's core JavaScript syntax should provide a common, standardized baseline that will work well for any standard JavaScript environment. For users who want more specialized highlighting for a particular environment or framework, third-party syntaxes can extend the core to provide more detailed `support` scopes tailored to that environment.

Thoughts?

Thom1729:
[...]

PR #1571, opened 2018-05-14
[JavaScript] Remove meta.prototype scopes.

Thom1729:
See #1569. As far as I can tell, these scopes are not useful for anything. Maybe they once were, in the `tmLanguage` days. The benefit to removing them is simplicity: they will no longer interfere with the `support` scopes that need updating (#1563).

Examples of formerly scoped constructs:

```js
Class3.prototype = function() {};
// ^^^^^^^^^^^^^ meta.prototype.declaration

Proto.prototype.attr
// ^^^^^^^^^^^^ meta.prototype.access
```

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -777,8 +777,6 @@ contexts:
   expression-begin:
     - include: expression-break
 
-    - include: literal-prototype
-
     - include: regexp-complete
     - include: literal-string
     - include: tagged-template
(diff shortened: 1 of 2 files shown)


deathaxe:
> these scopes are not useful for anything

I agree with you.

In general we should keep in mind that primary function of syntax definitions is syntax highlighting and drive basic goto... features. Not sure whether it should be desired goal to try to create full featured static code analysis tools using the syntax definitions. This is what I feel some developments head to.

We should leave that to language servers or any other kind of external tools. 

If more than 2/3rd of code is required to drive meta scopes, which no one will see, something goes wrong. Not only performance suffers from that, but also more and more edge cases with certain/uncommon but valid code constructs are introduced.

Therefore I vote for your change.

PR #1572, opened 2018-05-14
[JavaScript] Simplify parsing of new operator.

Thom1729:
Removed the `meta.instance.constructor` scope and simplified the parsing of the `new` operator. The `new` operator itself is no longer scoped `meta.function-call.constructor`; only the call itself is. This PR de-kludges the changes made in #1424.

### DIFF
--- a/JavaScript/tests/syntax_test_js.js
+++ b/JavaScript/tests/syntax_test_js.js
@@ -1149,9 +1149,11 @@ var Constructor = function() {
 var abc = new ABC(
 //        ^^^ keyword.operator.word.new
 //            ^^^^ meta.function-call.constructor
+//        ^^^^^^^^ - meta.instance.constructor
     'my-name-is-abc',
     new (function () {
 //  ^^^ keyword.operator.word.new
+//  ^^^^^^^^^^^^^^^^^^ - meta.instance.constructor
 //      ^^^^^^^^^^^^^^ meta.function-call.constructor meta.function-call.constructor meta.group
         var foo = 1;
 //      ^^^^^^^^^^^^ meta.function-call.constructor meta.function-call.constructor meta.group meta.block
@@ -1174,6 +1176,7 @@ function f() {
 new Date().getTime()
 //  ^^^^^^ meta.function-call.constructor
 //  ^^^^ support.class.builtin
+//^^^^^^^^^^^^^^^^^^ - meta.instance.constructor
 
 new $();
 //  ^ variable.type.dollar.only punctuation.dollar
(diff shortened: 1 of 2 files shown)


JavaScript/tests/syntax_test_js.js
```diff
@@ -1146,15 +1147,14 @@ var Constructor = function() {
...
-//         ^ meta.instance.constructor keyword.operator.word.new
-//               ^ meta.instance.constructor meta.function-call.constructor
-//               ^ - meta.instance.constructor meta.instance.constructor
+//        ^^^ keyword.operator.word.new
+//            ^^^^ meta.function-call.constructor
     'my-name-is-abc',
     new (function () {
-//  ^ meta.instance.constructor meta.function-call.constructor meta.instance.constructor keyword.operator.word.new
-//      ^ meta.instance.constructor meta.function-call.constructor meta.instance.constructor meta.function-call.constructor meta.group
+//  ^^^ keyword.operator.word.new
```
FichteFoll: Maybe you could add a negative assertion for the `new` keyword and the constructor meta scope?

JavaScript/tests/syntax_test_js.js
```diff
@@ -1172,10 +1172,8 @@ function f() {
 }
 
 new Date().getTime()
-// ^^^^^^^ meta.instance.constructor
 //  ^^^^^^ meta.function-call.constructor
 //  ^^^^ support.class.builtin
-//        ^^^^^^^^^^ - meta.instance.constructor
```
FichteFoll: Why remove this assertion?

JavaScript/tests/syntax_test_js.js
```diff
@@ -1146,15 +1147,14 @@ var Constructor = function() {
 // Tests to ensure the new keyword is highlighted properly even when the
 // following element is not an identifier
 var abc = new ABC(
-//         ^ meta.instance.constructor keyword.operator.word.new
-//               ^ meta.instance.constructor meta.function-call.constructor
-//               ^ - meta.instance.constructor meta.instance.constructor
+//        ^^^ keyword.operator.word.new
```
FichteFoll: Imo it would make sense to add a negative assertion for the `new` keyword and the constructor meta scope.

JavaScript/tests/syntax_test_js.js
Thom1729: Added.

JavaScript/tests/syntax_test_js.js
Thom1729: Added.

JavaScript/tests/syntax_test_js.js
Thom1729: I've added it back.

PR #1574, opened 2018-05-14
[JavaScript] Refactor function calls, simplify identifiers.

Thom1729:
The JavaScript syntax tries to enclose function calls in `meta.function-call`. This is impossible to do perfectly, so the implementation is necessarily fuzzy. This PR provides a new implementation that improves upon the old one in several ways.

The key goal of this change is to consolidate as much identifier logic as possible. Identifiers may be scoped differently depending on where they occur and based on their content. The old function call code substantially duplicated some of this; this new implementation keeps as much of it as possible in one place. In addition, it will now handle unlimited chains of dotted property accesses ending in a method call. This can be observed in the new test file.

This, along with other current PRs (#1560, #1571, #1572), will serve as the basis for overhauling the `support` scopes (#1563). Consolidating identifier handling is an important prerequisite for that work.


wbond:
@Thom1729 I think that in some of my merging this got into a conflicted state. Would you mind taking a moment and checking over it?

Thom1729:
Done. (The CI failure seems to be Python-related.)

PR #1577, opened 2018-05-16
[Python] Allow digits in CONSTANTS_123

gerardroche:


FichteFoll:
From your experiences witht my arbitrary heiristic to detect actual constant names, how do you feel the minimum of 3 leading alphabetic characters fares? I noticed that constants like `NO` and `OK` are missed and those seem frequent enough. Should we lower the requirement to 2 charcters or remove it altogether?

gerardroche:
You beat me to it. 

I think the requirement could probably be lowered it to two, then see how it goes from there. 

>  Edit: Maybe change the heuristic to have at least three consecutive alpha chars in the name rather than only at the beginning? That should should cover names like NO_ONLINE_HELP for example.

This might be better, yes.

deathaxe:
A short look into `sublime.py` with its `OP_...` and `UI_...` constants speaks for is own.

wbond:
So is this PR going to be updated to make the change to 2 chars, or will that be separate?

gerardroche:
Sorry I haven't had time to address it. If you can sort it @FichteFoll that would be great.

FichteFoll:
See #1642 for my proposal.

megadr01d:
This was merged in, right?
Using `startup, version: 3180 osx x64 channel: dev`, I still get highlight issues on 2 chars.

screencap-1540039736976
screencap-1540039756528

jfcherng:
@megadr01d No. The last commit for this repo is on Jul 10, 2018 as shown in https://github.com/sublimehq/Packages/commits/master.

megadr01d:
@jfcherng  OK, [committed](https://github.com/sublimehq/Packages/commit/6a9fa63eba4f9f891695a1d2777ff19f496c3846) but not [release](https://github.com/sublimehq/Packages/releases) yet. Thanks for the pointer.

PR #1583, opened 2018-05-17
[ShellScript] Add .zshrc as a valid extension

michaelmior:
This just adds `.zshrc` as a valid extension for shell scripts in the same vein as `.bashrc`.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -24,6 +24,7 @@ file_extensions:
   - .bashrc
   - .profile
   - .textmate_init
+  - .zshrc
 
 first_line_match: |
   (?x)                     # ignore whitespace in this regex


wbond:
Perhaps @rwols can comment on the ramifications of this?

It seems to me that zsh has a bit different syntax than bash, but I don't know to what extent.

michaelmior:
It's true that the syntax is somewhat different, but that seems tangential to this issue since files with `.zsh` extensions already use this shell script highlighting.

rwols:
@michaelmior this syntax promises correct bash syntax highlighting. It does not promise correct zsh syntax highlighting.

Some zsh constructs may trip up the syntax. A good overview can be found here: https://github.com/sublimehq/Packages/issues/1358

I believe the best way forward for a zsh syntax is to make this syntax compatible with it. I don't think major changes are required to make that work.

However, at this point in time, I don't think it is wise to make the promise to users that this syntax will highlight zsh correctly.

rwols:
I guess since `zsh` is in the file extension list already we could go ahead and add `.zshrc`, but please note that some parameter expansion features are not highlighted correctly right now.

michaelmior:
@rwols  Yes, I agree that there will be some incorrect highlighting in some cases but my point about the `.zsh` file extension is that the claim is already made that these files will be highlighted correctly so it might as well include `.zshrc` for consistency. Otherwise it would probably make more sense to remove zsh support entirely until it's properly fixed. But I think some highlighting that will occasionally be incorrect is better than none.

FichteFoll:
I've been using the syntax to highlight zsh code all this time and haven't run into any issues so far.

wbond:
Awesome, thanks for this improvement and the feedback from everyone!

michaelmior:
Thanks @wbond for the merge!

Issue #1586, opened 2018-05-17
[php] Syntax colouring suddenly gives up


louisl:
I think this is Sublimes fault it's only started occurring since recent updates.

<img width="1185" alt="screenshot 2018-05-17 19 37 46" src="https://user-images.githubusercontent.com/1140394/40197061-3bb2b25c-5a0a-11e8-82ee-88cc63e46be9.png">

```php
<?php
   /**
     * Delete child (DELETE)
     *
     * @param Request $request     Data
     * @param integer $id          Parent primary key
     * @param string  $childMethod Child method
     * @param integer $childId     Child primary key
     *
     * @return \Illuminate\Http\JsonResponse
     */
    public function destroyChild(Request $request, int $id, string $childMethod, int $childId)
    {
        try {
            if ($request->input('destroy') === 'true') {
            	$child = $this->currentModel::findOrFail($id)->{$childMethod}->find($childId);
            	$this->currentModel::findOrFail($id)->{$childMethod}()->detach($childId);
            	$child->forceDelete();
            } else {
                $child = $this->currentModel::findOrFail($id)->{$childMethod}->find($childId);
                $child->delete();
            }
        } catch (Exception $e) {
            return $this->respondWithBadRequestError($e);
        }

        return $this->respondDeleted();
    }
```

louisl:
I'm still getting this in 3.3.1, Build 3176. ( Remove [FROM HERE ON IS WHITE] )

```php
    /**
     * Delete child (DELETE)
     *
     * @param Request $request     Data
     * @param integer $id          Parent primary key
     * @param string  $childMethod Child method
     * @param integer $childId     Child primary key
     *
     * @return \Illuminate\Http\JsonResponse
     */
    public function deleteChild(Request $request, int $id, string $childMethod, int $childId)
    {
        try {
            if ($request->input('destroy') === 'true') {
                $child = $this->currentModel::findOrFail($id)->{to_method_case($childMethod)}->find($childId);
                $this->currentModel::findOrFail($id)->{to_method_case($childMethod)}()->detach($childId);
                $result = $child->forceDelete();
            } else {
                $child = $this->currentModel::findOrFail($id)->[FROM HERE ON IS WHITE]{to_method_case($childMethod)}->find($childId);
                $result = $child->delete();
            }
        } catch (Exception $e) {
            return $this->respondWithBadRequestError($e);
        }

        return $this->respondDeleted();
    }
```

jfcherng:
[...]

PR #1588, opened 2018-05-17
[PHP] Fix chained static function/method call

jfcherng:
Fix https://github.com/sublimehq/Packages/issues/1586.


louisl:
I'm still getting this in 3.3.1, Build 3176. ( Remove [FROM HERE ON IS WHITE] )

```php
    /**
     * Delete child (DELETE)
     *
     * @param Request $request     Data
     * @param integer $id          Parent primary key
     * @param string  $childMethod Child method
     * @param integer $childId     Child primary key
     *
     * @return \Illuminate\Http\JsonResponse
     */
    public function deleteChild(Request $request, int $id, string $childMethod, int $childId)
    {
        try {
            if ($request->input('destroy') === 'true') {
                $child = $this->currentModel::findOrFail($id)->{to_method_case($childMethod)}->find($childId);
                $this->currentModel::findOrFail($id)->{to_method_case($childMethod)}()->detach($childId);
                $result = $child->forceDelete();
            } else {
                $child = $this->currentModel::findOrFail($id)->[FROM HERE ON IS WHITE]{to_method_case($childMethod)}->find($childId);
                $result = $child->delete();
            }
        } catch (Exception $e) {
            return $this->respondWithBadRequestError($e);
        }

        return $this->respondDeleted();
    }
```

Issue #1603, opened 2018-05-23
[ShellScript] Indented comments break syntax highlighting


rolfbjarne:
And quotes in the comment make things even worse

![output_rkjqwi](https://user-images.githubusercontent.com/249268/40422306-425996f0-5e8f-11e8-8ae1-af0bf838747f.gif)

v3.1.1, build 3176

rwols:
I see you have your syntax set to "Shell-Unix-Generic". If you set it to "Bourne Again Shell (bash)", comments should work. See also: https://forum.sublimetext.com/t/problem-with-shell-syntax-for-comments/37087

rolfbjarne:
Yes, hacking around ApplySyntax' decisions made it work. Thanks!

PR #1604, opened 2018-05-23
[ShellScript] Fix: forgot prototype context in old Shell-Unix-Generic syntax

rwols:
fixes https://github.com/sublimehq/Packages/issues/1603


FichteFoll:
Random question, but how long are we going to keep the old syntax?

PR #1606, opened 2018-05-23
[JavaScript] Fix ending of blocks.

### LINKED ISSUE #1605: Select .. case coloring in JS
Hi,

Sublime text has been asking me to update for a few weeks, finally did it this morning. The last update seems to have broken the code coloring in Javascript for switch cases when you use brackets in the cases.

For example:

```
switch ($a) {
    case 1:
        functionA();
        break;
    case 2:
        functionB();
        break;
}
```
is okay, but:

```
switch ($a) {
    case 1: {
        functionA();
        break;
    }
    case 2: {
        functionB();
        break;
   }
}
```

Breaks the code coloring. I always, ALWAYS use brackets. To quote my team lead: Life is way too short not to put your brackets. This is a small issue that is very frustrating for me... It leads me to think I have a syntax error, which I tried to find for 10 minutes until I figured it's a problem with ST, and not my code.

I've been using ST for 2 years now and I do love it thoroughly... but this bug makes my life more difficult. I've installed Visual Studio Code for now, please fix this fast before I get used to it! XD

Thanks!

Thom1729:
Fixes #1605.

### DIFF
--- a/JavaScript/tests/syntax_test_js.js
+++ b/JavaScript/tests/syntax_test_js.js
@@ -702,6 +702,10 @@ switch ($foo) {
     default$
 //  ^^^^^^^^ - keyword
     ;
+
+    case 0: {}
+    case 1:
+//  keyword.control.switch
 }
 // <- meta.block punctuation.section.block.end
 
--- a/JavaScript/tests/syntax_test_js.js
+++ b/JavaScript/tests/syntax_test_js.js
@@ -705,7 +705,7 @@ switch ($foo) {
 
     case 0: {}
     case 1:
-//  keyword.control.switch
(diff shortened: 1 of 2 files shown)


JavaScript/tests/syntax_test_js.js
```diff
@@ -702,6 +702,10 @@ switch ($foo) {
     default$
 //  ^^^^^^^^ - keyword
     ;
+
+    case 0: {}
+    case 1:
+//  keyword.control.switch
```
keith-hall: this appears to just be a comment, not a syntax test assertion, it is missing the `^` chars

PR #1608, opened 2018-05-24
[C++] Teach the C++ comments settings about doxygen comment prefixes.

chandlerc:
Without this, tools like Wrap Lines Plus won't be able to correctly wrap doxygen comments.


wbond:
Thanks for this contribution @chandlerc!

PR #1616, opened 2018-05-26
[Markdown] Allow whitespace between opening fence and language

### LINKED ISSUE #1615: [Markdown] Allow whitespace between opening fence and info string
The following Markdown code [seems to be valid](https://github.github.com/gfm/#example-112) in (GitHub-flavored) Markdown:

    ``` ruby
    def foo(x)
      return 3
    end
    ```
(notice the space between the opening fence <code>```</code> and `ruby`). However, the code-block is currently not syntax-highlighted properly (the Ruby part).

Compare:
image

image

sharkdp:
Allow for code blocks like

    ``` ruby
    def foo(x)
      return 3
    end
    ```

with whitespace between the opening fence and the 'ruby' keyword.

closes #1615

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -98,10 +98,10 @@ variables:
           [ ]*
           ([ ]{0,3})   # up to 3 spaces
           (
-            `{3,}      #   3 or more backticks
+            (`){3,}    #   3 or more backticks
             (?![^`]*`) #   not followed by any more backticks on the same line
           |            # or
-            ~{3,}      #   3 or more tildas
+            (~){3,}    #   3 or more tildas
             (?![^~]*~) #   not followed by any more tildas on the same line
           )
           \s*          # allow for whitespace between code block start and info string
@@ -117,12 +117,13 @@ variables:
         )
     code_fence_escape: |-
       (?x:
-        ^         # the beginning of the line
-        #\1        # whitespace previously captured
-        #[ ]{0,3}  # up to 3 spaces - doesn't have to be the same as the opening fence, but within 3 spaces
+        ^             # the beginning of the line
         [ ]*
-        (\2)      # the backtick/tilde combination that opened the code fence
-        \s*$      # any amount of whitespace until EOL
+        (
+          \2          # the backtick/tilde combination that opened the code fence
+          (?:\3|\4)*  # plus optional additional closing characters
+        )
+        \s*$          # any amount of whitespace until EOL
       )
     html_tag_open_commonmark: |-
       (?xi:
@@ -939,7 +940,7 @@ contexts:
       captures:
         0: meta.code-fence.definition.begin.xml.markdown-gfm
         2: punctuation.definition.raw.code-fence.begin.markdown
-        3: constant.other.language-name.markdown
+        5: constant.other.language-name.markdown
       embed: scope:text.xml
       embed_scope: markup.raw.code-fence.xml.markdown-gfm
       escape: '{{code_fence_escape}}'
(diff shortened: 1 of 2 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -924,6 +924,7 @@ contexts:
     - match: |-
          (?x)
           {{fenced_code_block_start}}
+          \s*
```
keith-hall: you could put the `\s*` into the `fenced_code_block_start` variable instead to avoid needing to repeat it for each language's code fence.

Markdown/Markdown.sublime-syntax
```diff
@@ -1134,6 +1148,7 @@ contexts:
...
           ((?i:ruby|rb|rbx))
           \s*$\n?          # any whitespace until EOL
```
keith-hall: it may also make sense to move the `\s*$\n?          # any whitespace until EOL` into a variable so we can modify that to get GFM example 112 working with `~~~~    ruby startline=3 $%@#$`.

Markdown/syntax_test_markdown.md
```diff
@@ -1927,7 +1927,7 @@ var_dump(expression);
 |^^^ meta.paragraph meta.code-fence.definition.end.regexp - markup
 |^^ punctuation.definition.raw.code-fence.end
 
-```bash
+``` bash
```
keith-hall: it would probably be a good idea to add an assertion that the whitespace isn't scoped as `punctuation` or `constant.other.language-name`, and that "`bash`" is still scoped correctly as `constant.other.language-name`

Markdown/Markdown.sublime-syntax
sharkdp: Sure, I'll fix this. I wasn't sure if that would be possible as `{{denced_code_block_start}}` is used in a few other places as well.

Markdown/Markdown.sublime-syntax
sharkdp: I moved this to a new variable and adapted it to get [GFM example 112](https://github.github.com/gfm/#example-112) working, except for the part where the code block is closed by `~~~~~~~` (i.e., more `~` than the start of this code block `~~~~`). I also added a test for this.

Markdown/syntax_test_markdown.md
sharkdp: Good suggestion. I moved this to a new test. Not sure if the assertion for the whitespace-part is okay.

sharkdp:
This PR now includes a few things

* Allow for whitespace in between the code block start (e.g. <code>```</code>) and the language (e.g. `ruby`).
* Allow for trailing characters in the info string (after the language definition) as shown in [this example](https://github.github.com/gfm/#example-112):

<pre>
~~~~    ruby startline=3 $%@#$
def foo(x)
  return 3
end
~~~~~~~
</pre>

* Allow code blocks to be closed by more `~` or <code>`</code> characters than it started with, see above example or [this](https://github.github.com/gfm/#example-93):

<pre>
````
aaa
```
``````
</pre>

File: Markdown/Markdown.sublime-syntax [hunk 4]
```diff
@@ -98,21 +98,32 @@ variables:
           [ ]*
           ([ ]{0,3})   # up to 3 spaces
           (
-            `{3,}      #   3 or more backticks
+            (`){3,}    #   3 or more backticks
```
FichteFoll: Why did you add capture groups here? It's usually not a good idea to have those in variables because they affect the index for the captures map, and indeed you had to increase the offset of many captures from 3  to 5.

sharkdp: I need these new capture groups in the `code_fence_escape` part because markdown blocks that are opened by *n* backticks/tildes can be closed by *n+x* respective closing characters, e.g:
```
~~~~    ruby startline=3 $%@#$
def foo(x)
  return 3
end
~~~~~~~
```

FichteFoll: Okay, but wasn't that functionality there beforehand? In this pr you just added capture groups but you didn't add any references to them (besides the scope association). Besides, the added groups only match a single character and not the while opening sequence.

wbond:
Thanks!

PR #1618, opened 2018-05-27
[Markdown] Fix reference to Javascript Syntax

trishume:
The Javascript syntax uses `source.js`. This caused failures in syntect where a broken reference is an error, which perhaps it shouldn't be if it isn't in Sublime.

cc @keith-hall


deathaxe:
Can you rename the PR to start with the package name in brackets, please - just to quickly see which package is addressed.

[Markdown] Fix reference to Javascript Syntax

keith-hall:
maybe it'd make sense to add a (loose) test assertion on some of the JS code inside the fence - if I had done this we would have caught it sooner ;)

trishume:
@keith-hall done.

PR #1621, opened 2018-05-27
[XML] Create contexts and fix a couple of highlighting issues.

deathaxe:
This quite big PR ...

1. splits all rules into several contexts which could be reused by other syntaxes 
    _(done in the first commits)_
2. fixes most highlighting issues caused, if tags/attributes span several lines.

   Example
    ```
    <tag
      attr=
      "value"
    >

3. tries to be as permissive as possible with incomplete tags or expressions in order not to break the whole document's highlighting because of incomplete tags.
4. invalid tag/attribute names  like `na$%me` are highlighted as one token rather than matching the first part as valid name.
5. it applies a couple of fixes/enhancements which were introduced to the HTML syntax recently, such as more detailed entity scoping, etc.
6. improves DOCTYPE declarations and internal subsets.
7. more permissive <`?xml... ?>` preprocessing tags that support more than just `xml-stylesheet` and `xml-model` as the official spec reserves any lowercase tag name beginning with `xml` for future use.


#### DOCTYPE DECLARATION

The code is inspired by a former PR but heavily reduced to the least required stuff needed for highlighting. This means much simpler `meta.` scopes in general. 

All valid DTD tags are fully supported.

Instead of creating a separate syntax definition the already existing contexts in the XML.sublime-syntax are reused, renamed and extended to fully support the DTD syntax even if tags span over multiple lines. 

Adding a DTD.sublime-syntax (if wanted) was as easy as adding a file with one `- include: text.xml#dtd` statement.

The declaration scope names were slightly changed to hopefully follow naming guidelines more closely, while the top-level scopes `keyword` are unchanged to distinguish those tags from normal xml. The existing `meta.tag.sgml` is used for all declarations.

#### CDATA

The `<![CDATA[ string ]]>` tag is scoped the same way the DTD tags are to let it stand out from ordinary tags and distinguish the enclosing name/punctuation from its content.

### ~~NOTE ABOUT SYNTAX TESTS~~

~~Why do syntax tests of CI fail?~~

~~In order to be able to add as much tests as possible for any detail, it was required to introduce a dedicated syntax test comment style, that does not interfere with the XML syntax in any way.~~

~~Inspired by HTML the `## ...` was choosen.~~

~~In order to make the syntax tests pass, the `prototype` context in the XML.sublime-syntax needs to be commented in as those comments are not meant to be scoped as valid in the wild.~~


keith-hall:
As it is a big PR, please can you provide performance stats of before and after these changes, as per the readme on this repo - ideally with a fairly large XML file.

deathaxe:
Benchmarks were performed during development with two XML files of about 100k lines of code with a size of about 3MB. The first one is a "simple" XML file and the second one is a Word document exported as OpenDocumentFormat heavily using namespaces. Both files heavily use attributes.

_Timings of the complete syntax:_

 file         | 1 |  2
------|------|------
before | 267ms  |  440
after   |  300ms |  503
diff    |   +12%  |  + 14%

The overall performance after all the rework is about 10..15% slower.

It is caused by the more complicated approach of matching attribute names and additional context switches required to support attributes spanning multiple lines.

The old syntax didn't need any lookahead to identify an attribute name as it used `({{qualified_name}})\s*(=)` which fails, if the `=` is located at the next line. Even though this might be uncommon it is still valid XML.

Now the name is qualified by using lookaheads checking for any control character which indicates an end of attribute name followed by a context switch in order to check for `=`. This is just a bit more expensive.

If there were no attributes, the performance would even be about 7% better.

_Timings with `tag-stuff` and `tag-attributes` commented out._

 file         | 1 |  2
------|------|------
before | 259 |  320
after | 240 |  301
diff  |  -7% |  - 6%

wbond:
I don't quite understand the issue as to why the tests must fail, but the tests won't be useful if they are always failing, or have to be manually enabled sometimes.

deathaxe:
The test rules must be excluded from normal parsing somehow in order to not interfere with the syntax itself. This is achieved by using ordinary comments in most languages as they are normally allowed anywhere.

XML comments like `<!-- -->` are not valid within tags. Something like `<tag <!-- comment --> >` is therefore not allowed. 

In the following example everything beginning from `bar`  would no longer be part of the tag.

```
     <foo
<!--  ^ - meta.tag -->
<!--   ^^^^^ meta.tag -->
<!--   ^ punctuation.definition.tag.begin -->
<!--    ^^^ entity.name.tag.localname -->
      bar
<!--    ^^^ entity.other.attribute-name.localname -->
<!--^^^^^^^^ meta.tag -->
      =
<!--^^^^^^ meta.tag -->
<!--    ^ punctuation.separator.key-value -->
      "baz"
<!--^^^^^^^^^^ meta.tag -->
<!--   ^ - string.quoted.double -->
<!--    ^^^^^ string.quoted.double -->
<!--         ^ - string.quoted.double -->
      />
```

In order to follow the official specs and (more important) to avoid highlighting large parts of a document illegal due to incomplete tags/attributes, a `<` will always cause a tag context to be popped off in order to gracefully continue with the next tag (highlighting the last whitespace in front of it as illegal).

The HTML syntax therefore uses the `##` to mark syntax test cases. But those are not valid in multi line tags, too.

In the following example the `"baz"` wouldn't be recognized as attribute value because the XML spec requires the `=` to be followed by quoted string. 

```
     <foo
##  ^ - meta.tag
##   ^^^^^ meta.tag
##   ^ punctuation.definition.tag.begin
##    ^^^ entity.name.tag.localname
      bar
##    ^^^ entity.other.attribute-name.localname
##^^^^^^^^ meta.tag
      =
##^^^^^^ meta.tag
##    ^ punctuation.separator.key-value
      "baz"
##^^^^^^^^^^ meta.tag
##   ^ - string.quoted.double
##    ^^^^^ string.quoted.double
##         ^ - string.quoted.double
      />
```

To make that test work I'd need to remove the test for `=`.

This is actually what HTML does. It avoids tests in places, in which they would break the syntax.

So I see 4 possible solutions:

1. I could follow HTML and remove all test cases, which break the syntax. Feels like this will remove most ones.
2. The `prototype` context could remain active hopefully not causing issues with real world XML, but definitely causing a small performance impact.
3. Add a rule to the main context to check for SYNTAX TEST line to enable those comments only in the syntax_test_ files_. This additionally required push/set might interfere with languages embedding/including the XML syntax some day.
4. A custom way to specify syntax test rules which do not interfere with the syntax itself. Maybe an entry in a `tmPreference` to define the style of syntax test lines, which then are excluded from parsing at all?

I'd be more than happy about an idea how to keep all tests and make them always work. I honestly don't like this solution very much, too.

keith-hall:
related: https://github.com/SublimeTextIssues/Core/issues/1152

deathaxe:
That's the point.

wbond:
I doubt we are going to add the code paths in the (already complicated) tokenizer to try and skip syntax highlighting tokens. That would add a bunch of complexity for very little benefit.

For now I think this will probably have to be some sort of commented-out tests and rules that can be enabled if someone wants to double-check the work.

FichteFoll:
In the python tests, I just repeated multi-line tests for multi-line statements (like `with`) and with each test I added another line which I then ran assertions on. The idea is that until the line break, the syntax wouldn't know whether the following required token is present, so it needs to be fail-safe either way. You cannot have a line continuation after a line comment in Python.

deathaxe:
@wbond @FichteFoll @Thom1729 @keith-hall 

Any comments, wishes ... on that PR?

wbond:
I have not done an extensive line-by-line review, however:

 1. The existing tests seem to be modified just to change the prefix tokens (good!)
 2. The syntax has a vastly increased number of tests handling a bunch of invalid contexts and more complex cases
 3. The structure of the syntax seems to be more well-organized and modular
 4. This will be included early in a dev cycle, so there will be a fair bit of time to identify any major breaks

Thanks for your work on this @deathaxe and the time put into writing tests!

Issue #1503, opened 2018-04-17
[Bash] Terminating keyword of control structures incorrectly highlighted when not followed by space/newline


zx8:
> **Actual**
> ![image](https://user-images.githubusercontent.com/8544900/38861748-34365bf6-422b-11e8-8950-028dad8d8777.png)
>
> vs.
>
> **Expected** (without newline/space)
> ![image](https://user-images.githubusercontent.com/8544900/38861774-40735d60-422b-11e8-9346-45b28a4236e4.png)

This applies to `if [...]; then [...] fi`,  `while [...]; do [...] done` etc. as well.

PR #1623, opened 2018-05-28
[ShellScript] Fix: scope keywords correctly while typing

rwols:
fixes https://github.com/sublimehq/Packages/issues/1503.

When typing keywords like "done" or "fi", previously these got scoped as a
function call when the cursor was at the EOF.


ShellScript/Bash.sublime-syntax
```diff
@@ -57,7 +57,7 @@ variables:
   is_path_component: (?=[^\s/]*/)
   is_start_of_arguments: '[`=|&;()<>\s]'
   is_variable: (?=\s*{{nbc}}(?:[({]{{nbc}}[)}])?{{nbc}}=)
-  keyword_boundary_end: (?!=)(?=[^\w_-])
+  keyword_boundary_end: (?!=)(?:(?=[^\w_-])|$)
```
FichteFoll: You could just add the `|$` to the look-ahead.

PR #1634, opened 2018-06-08
[PHP] Support PHP 7.3 Flexible Heredoc and Nowdoc Syntaxes

jfcherng:
Ref: https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes

PHP 7.3 alpha 1 just released today :+1:

### DIFF
--- a/PHP/Indentation Rules - heredoc end.tmPreferences
+++ b/PHP/Indentation Rules - heredoc end.tmPreferences
@@ -7,10 +7,15 @@
 	<string>source.php meta.heredoc-end.php</string>
 	<key>settings</key>
 	<dict>
-		<key>unIndentedLinePattern</key>
+		<!--
+			If you are writing codes for PHP less then 7.3, you may want to
+			uncomment the the following section to ensure that "heredoc ending
+			token must be at BOL".
+		-->
+		<!-- <key>unIndentedLinePattern</key>
 		<string>.</string>
 		<key>preserveIndent</key>
-		<true />
+		<true /> -->
 	</dict>
 </dict>
 </plist>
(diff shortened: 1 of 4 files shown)


keith-hall:
you may wish to undo the indentation rules that prevent the closing heredoc identifier from being indented: https://github.com/sublimehq/Packages/pull/1229

jfcherng:
@keith-hall I have no experience in using ST's reindent (I format PHP codes with php-cs-fixer and my hands manually). Could you help on that?

keith-hall:
thinking about this in terms of how to get indentation to be more natural for users of PHP 7.3 alpha 1 or newer, and allow users of older versions to keep the old "heredoc end token must be at BOL" behavior, I suggest commenting out the

```xml
<key>unIndentedLinePattern</key>
<string>.</string>
<key>preserveIndent</key>
<true />
```

bit in https://github.com/sublimehq/Packages/pull/1229/files#diff-04cd5e4d59074cb0298a7bdab122b81e, so that people can uncomment it if they want this behavior.

I think it should be safe to remove ` - meta.heredoc-end` from https://github.com/sublimehq/Packages/pull/1229/files#diff-f88b509255abd76ff01f58edcd261848R7 as the scope selector scoring algorithm and `tmPreferences` precedence rules should then keep everything working as expected for both user bases.

PHP/Indentation Rules - heredoc end.tmPreferences
```diff
@@ -7,10 +7,10 @@
 	<string>source.php meta.heredoc-end.php</string>
 	<key>settings</key>
 	<dict>
-		<key>unIndentedLinePattern</key>
+		<!-- <key>unIndentedLinePattern</key>
```
wbond: There should probably be a comment here explaining why this is commented-out.

jfcherng: @wbond Added.

jfcherng:
@wbond I would like to see this PR be merged. And also another quite simple one https://github.com/sublimehq/Packages/pull/1740.

PR #1635, opened 2018-06-11
[SQL] NULL-safe equal to operator ligature.

sowizz:
Problem described at [Sublime forum](https://forum.sublimetext.com/t/wrong-ligature-for-null-safe-equal-to-operator-in-sql-syntax/37407/4).

### DIFF
--- a/SQL/syntax_test_sql.sql
+++ b/SQL/syntax_test_sql.sql
@@ -30,5 +30,5 @@ multiline comment
 select
 
 
-<=>  
+   <=>  
 -- ^^^ keyword.operator.comparison.sql
--- a/SQL/syntax_test_sql.sql
+++ b/SQL/syntax_test_sql.sql
@@ -28,3 +28,6 @@ multiline comment
 -- ^ source.sql comment.block.c
 */
 select
+
+
+<=>  -- <- keyword.operator.star.sql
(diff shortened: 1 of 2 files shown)


SQL/syntax_test_sql.sql
```diff
@@ -28,3 +28,6 @@ multiline comment
 -- ^ source.sql comment.block.c
 */
 select
+
+
+<=>  -- <- keyword.operator.comparison.sql
```
FichteFoll: This is not a proper test line. Test lines should be specified in the lines *after* the line you want to test, so there should be a newline before `--`.

djspiewak: Oh yeah, this is just testing the beginning token.  I'm pretty sure the following will be comprehensive though:

```sql
   <=>  
-- ^^^ keyword.operator.comparison.sql
```

sowizz: Sorry, I'm new to this. Didn't write a single test case in my life. Should I do another commit?

djspiewak: Yes. Just change your test line above into the two lines that I wrote.

djspiewak:
@sowizz Note how your test failed. You can run tests in Sublime (use the build command on the syntax mode file) to see this for yourself. You need to preserve the spacing that I used in my comment.

wbond:
Thanks!

PR #1637, opened 2018-06-11
[HTML] Remove isolated options in variables.

Thom1729:
See [core issue #2354](https://github.com/SublimeTextIssues/Core/issues/2354).

Modified the test to make sure that the ignore-case flag is working.


FichteFoll:
Tbf, isolated options should **never** be used in variables.

PR #1640, opened 2018-06-18
[JavaScript] Remove isolated regexp flags from variables.

Thom1729:
Because of https://github.com/SublimeTextIssues/Core/issues/2354, and because it's generally better anyway.


FichteFoll:
If I make a linter for this, I'll definitely make sure to flag these things.

PR #1641, opened 2018-06-19
[JavaScript] Reorganize support scopes

Thom1729:
See discussion in #1563.

- Reorganize most of the support scopes into the following categories:
  - ECMAScript builtins.
  - DOM/browser.
  - Node.js.
  - Console API.
- Eliminate items with no obvious source: `super`, `GLOBAL`, `root`, and `self`.
- Add missing top-level ECMAScript builtins.
- Moved `Math` from `support.class` to `support.constant` (along with its new friends `Atomics`, `JSON`, and `Reflect`).
- Add properties of `Object.prototype` (e.g. `toString`).
- Removed all other `support` scopes for object properties (e.g. `id`, `slice`).
- Scope items from ECMAScript Annex B as deprecated (notably `__proto__`).
- Added tests for all of the ECMAScript builtins.

With this reorganization in place, [third party packages](https://github.com/Thom1729/Sublime-JS-Custom/) will be able to easily add environment- or library-specific items (so the core doesn't have to).

Next steps:

- Implement standard second-level properties (e.g. `Object.keys`).
- Trim down the DOM and Node support items to the bare essentials (e.g. probably not the name of every popular Node module).
- Implement full-featured `support` for browsers and Node in aforementioned third-party package.
- Maybe: move the jQuery dollar sign special-casing to third party packages (and out of core).
- Maybe: eliminate the meaningless `variable.other.object` distinction.


Thom1729:
I've added highlighting for builtin methods of top-level objects. This is a fair bit of boilerplate-y code, but these methods are basic language features, many of them are used very frequently, and I think it makes sense to highlight them. Some popular examples:

- `Array.isArray`
- `Date.now`
- `JSON.stringify`
- `Math.abs`
- `Object.is`
- `Promise.all`
- `String.raw`

This commit highlights all standard properties of the built-in objects. I considered highlighting only the most commonly-used properties, but decided that that would be arbitrary and likely confusing.

Complete test coverage is provided.

Thom1729:
I've fixed up the Node second-level properties (e.g. `module.exports`) to work like the builtins now do. I also removed the arbitrary list of DOM constants. Tests have been added for all of the remaining support scopes.

In addition, I've removed all environment-specific tests from the base syntax test file (adding them to other test files as necessary). This is so that [a third-party syntax based on the core syntax](https://github.com/Thom1729/Sublime-JS-Custom) could remove all of the environment-specific support scopes and still pass all of the non-environment-specific tests. This would allow such a syntax to selectively add in environment-specific support scopes and (only) the appropriate tests.

This is about as far as we should probably go for now. When this PR is merged, third-party packages can start providing more detailed environment support and we can make final determinations about what items should be highlighted in core without leaving anyone out in the cold.

Thom1729:
@wbond If possible, it would be very helpful if this issue were merged soon-ish even if there is no release imminent. When it's merged, I can update JS Custom accordingly and start implementing environment-specific feature sets. In addition, the scoping of DOM-specific property names (#1700, https://github.com/Thom1729/Sublime-JS-Custom/issues/30) is my last major point of concern for babel-sublime users migrating to core or to JS Custom (or to a future JS Custom-based version of babel-sublime).

wbond:
Thanks for this, and sorry for the delay!

PR #1642, opened 2018-06-20
[Python] Update constant identifier regex

### LINKED ISSUE #1607: [Python] All-caps identifiers mis-highlighted if they contain digits
This is with `3175`.

bad_highlight

My theory is that `THIS_IS_FINE` is properly highlighted but `THIS_IS_FINE8` is not. Perhaps it should be the other way around; I'd just like to make sure they're consistent. 

(Secondarily, perhaps this is all fine and there's simply something I don't know about python that makes combining all caps identifiers with digits un-cool?)

FichteFoll:
Now only requires two consecutive upper-case letters and allows digits.

Supersedes/Closes #1577.
Fixes #1607.


deathaxe:
Looks good.

wbond:
Thanks!

PR #1643, opened 2018-06-20
[Python] Python 3.7 support

FichteFoll:
Not many changes for 3.7, which is currently in rc1 and the scheduled release date is at the [end of this month](https://www.python.org/dev/peps/pep-0537/#schedule), in exactly one week.

Syntax changes were the deprecation of `async` and `await` as non-keywords, which we cannot address in the syntax definition, and addition of a `__future__` import to prevent type annotation evaluation at load time, which also doesn't matter for us.

A single built-in function was added and three new magic methods from features in the stdlib.

See https://docs.python.org/3.7/whatsnew/3.7.html for the complete list.

Also no tests because …

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -869,7 +869,7 @@ contexts:
   builtin-functions:
     - match: |-
         (?x)\b(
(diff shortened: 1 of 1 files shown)


wbond:
Thanks!

PR #1644, opened 2018-06-20
[Perl] Small whitespace related improvements

deathaxe:
This PR applies a change inspired by @keith-hall to avoid possibly breaking the `break` variable if it is used in extended regex patterns. It includes surrounding the `::` by tabs as well.

The second commit turns an `(?x)` into an `(?x: ...` as the latter one was used in another context the same way like.


wbond:
Could we add a test for this?

deathaxe:
Basically there are already some tests about accessors surrounded by whitespace, but a second look revealed some pragma keywords like `warnings :: register` did not yet support that.

wbond:
Thanks!

PR #1646, opened 2018-07-02
Enhance modular structure

jrappen:
* move completions, snippets, tests to subdirs
* add `.gitattributes` file


wbond:
The problem with moving a bunch of files around is overrides. Moving tests probably isn't a big deal, but moving completions and snippets would affect user overrides, so I'm not inclined to do this.

YAML/tests/syntax_test_preview.yaml
```diff
@@ -1,3 +1,4 @@
+# SYNTAX TEST "Packages/YAML/YAML.sublime-syntax"
```
FichteFoll: This file is not a syntax test file.

deathaxe:
While creating a sub directory for a bunch of tests may make sense somehow even though no normal user will care about it, I can't see a reason to create such sub directories with only one single file in it - which seems the case with many test files. Even 2 test files are not problem to distinguish them from the rest.

Same with completions. The only syntax providing more than 2 completion files is the Git Format package, which already organizes them in a sub directory.

YAML/tests/syntax_test_preview.yaml
jrappen: Not automatic, no.

jrappen:
Should I change anything or can I close this?

rwols:
You should probably split up the two commits into separate PRs, because adding a gitattributes file (which I'm all for) is a different topic than restructuring unit tests.

PR #1659, opened 2018-07-12
[C++] Add possibility for paths in front of class/struct identifiers

mjendruk:
When a struct or class has been forward declared in a namespace or other class, the class or struct can be defined using an identifier with a path in front of it. I added this possibility.

**Example**
```C++
namespace Namespace
{
  class BaseClass;
}

class Namespace::BaseClass {};
```

The test file was broken and still is. I fixed a part of it. The keywords `break`, `continue`, etc. should have had dedicated scopes according to the test file even though they don’t. The other part is related to the expression `a-->0`, which isn’t recognized as `a-- > 0` but as `a - ->`, i.e. an arrow accessor. I don’t know if this is an error in the test file or syntax definition.

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -12,7 +12,7 @@ first_line_match: ^#!.*\bperl\b
 scope: source.perl
 
 variables:
-  break: (?!\w| *::)
+  break: (?!\w|\s*::)
   identifier: '\b[_[:alpha:]]\w*{{break}}'
   module: '\b[_[:upper:]]\w*\b'
   member: '\b[_[:lower:]]\w*\b'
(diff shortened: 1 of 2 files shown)


FichteFoll:
Please add the language a pull request, issue or commit is about in the title (and commit message) similar to how other commits or pull requests are formatted. There are many c-like languages and reading your comment doesn't make it clear what language this affects.

wbond:
Tests on master seem fine (https://travis-ci.org/sublimehq/Packages/builds/401952969). Most likely you are in some diverged state locally.

The `::` should not get `entity.name.class`, neither should the namespace. The namespace name should be something like `support.namespace`.

All changed to C++ also need to be applied to Objective-C++.

rwols:
> The other part is related to the expression a-->0, which isn’t recognized as a-- > 0 but as a - ->

Should have been fixed [here](https://github.com/sublimehq/Packages/pull/1558). Have you pulled from a recent master branch?

mjendruk:
Hi! Thank you all for your feedback. 

> @wbond The :: should not get entity.name.class, neither should the namespace. The namespace name should be something like support.namespace.

Yes that’s what I thought too and then I looked at how scopes got applied to function definitions and function calls with namespace paths and realised that `entity.name.function` and `variable.function` where applied to the entire path including the namespace. 

With my addition, indexing works as expected as the namespace gets stripped away in `Symbol Index.tmPreferences`.

PR #1660, opened 2018-07-13
[C#] support namespaced types in catch clauses

keith-hall:
I noticed that before this PR, a catch containing a type with a namespace only scopes up to the dot as `support.type`. The rest of the namespace/type was unscoped. This fixes that.


C#/C#.sublime-syntax
```diff
@@ -1674,13 +1674,21 @@ contexts:
...
         2: meta.group.cs punctuation.section.group.begin.cs
-      push: [catch_block, var_declaration_explicit]
+      push: [catch_block, var_declaration_explicit, catch_namespace]
     - match: 'catch'
       scope: keyword.control.trycatch.catch.cs
       push: trycatch_block
     - match: (?=\S)
       pop: true
 
+  catch_namespace:
```
FichteFoll: I'm not versed in C#, but wouldn't it make sense to perform this kind of capture in the `var_declaration_explicit` context? Are type declarations with namespaces not valid?

keith-hall: good point, I somehow hadn't noticed that the syntax was missing support for things like `System.String blah = "test";`

C#/C#.sublime-syntax
```diff
@@ -432,14 +433,14 @@ contexts:
...
         - include: method_declaration
-        - match: '\{'
-          scope: punctuation.section.block.begin.cs
-          push:
-            - meta_scope: meta.block.cs
-            - match: '\}'
-              scope: punctuation.section.block.end.cs
-              pop: true
-            - include: data_type_body
+        # - match: '\{'
```
wbond: Can these commented-out lines be ditched now?

keith-hall: Yep, done :)

PR #1661, opened 2018-07-14
[C++/ObjC] Add support for paths in front of struct/class identifiers

mjendruk:
I'm sorry for the new pull request. The old one was automatically closed for some reason.

## _Copied from old PR_:

When a struct or class has been forward declared in a namespace or other class, the class or struct can be defined using an identifier with a path in front of it. I added this possibility.

**Example**
```C++
namespace Namespace
{
  class BaseClass;
}

class Namespace::BaseClass {};
```

## _Last reply_:

Hi! Thank you all for your feedback. 

> @wbond The :: should not get entity.name.class, neither should the namespace. The namespace name should be something like support.namespace.

Yes that’s what I thought too and then I looked at how scopes got applied to function definitions and function calls with namespace paths and realised that `entity.name.function` and `variable.function` where applied to the entire path including the namespace. 

With my addition, indexing works as expected as the namespace gets stripped away in `Symbol Index.tmPreferences`.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -1018,8 +1018,9 @@ contexts:
       set: data-structures-class-definition-after-identifier
     - match: '(?={{path_lookahead}})'
       set:
-        - meta_scope: meta.class.c++ entity.name.class.c++
+        - meta_scope: meta.class.c++
         - match: '{{identifier}}(?!\s*::)'
+          scope: entity.name.class.c++
           set: data-structures-class-definition-after-identifier
         - include: identifiers
         - match: '(?=[^\w\s])'
@@ -1051,8 +1052,9 @@ contexts:
       set: data-structures-struct-definition-after-identifier
     - match: '(?={{path_lookahead}})'
       set:
-        - meta_scope: meta.struct.c++ entity.name.struct.c++
+        - meta_scope: meta.struct.c++
         - match: '{{identifier}}(?!\s*::)'
+          scope: entity.name.struct.c++
           set: data-structures-struct-definition-after-identifier
         - include: identifiers
         - match: '(?=[^\w\s])'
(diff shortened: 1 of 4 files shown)


C++/C++.sublime-syntax
```diff
@@ -1044,6 +1050,12 @@ contexts:
     - match: '{{identifier}}(?={{data_structures_forward_decl_lookahead}})'
       scope: entity.name.struct.forward-decl.c++
       set: data-structures-struct-definition-after-identifier
+    - match: '(?={{path_lookahead}}\s*::\s*{{identifier}})'
+      set:
+        - meta_scope: entity.name.struct.c++
+        - include: identifiers
+        - match: '(?=[^\w\s])'
+          set: data-structures-struct-definition-after-identifier
```
rwols: I tested this out locally and it seems to work fine. Nice job! Only thing missing IMO are unions:

```
namespace ns {
class Bar;
struct Foo;
union Qux;
} // namespace ns

class ns::Bar {
public:
  Bar() = default;
};

struct ns::Foo {
  Foo() = default;
};

union ns::Qux {
  Qux() = default;
};
```

mjendruk: Okay the latest commit added support for both enums and unions. Should be good to go now.

mjendruk:
Correct handling of macros is not yet working

```c++
class Namespace::MyClass Macro
{
};
```

mjendruk:
There is still something weird though. In the code snippet below, both `Namespace` and `MyEnum` are recognized as symbols for `enum class Namespace::MyEnum` but for the other definition only `MyClass` is a symbol (as it should be).

```c++
class Namespace::MyClass
{
};

enum class Namespace::MyEnum
{
};
```

wbond:
This will have to be adapted to that the `entity.name` is not applied to the namespace name, nor the `::`.

If somewhere else you are seeing such a pattern, I would consider that a bug that we need to fix also.

mjendruk:
@wbond, yes I agree. My newer PRs #1819 and #1818 address this. This one needs to be updated.

mjendruk:
Fixed it!

wbond:
Thanks for these fixes, they look like solid enhancements!

PR #1662, opened 2018-07-14
[Go] syntax rework

mitranim:
This PR is a complete rework of the Go syntax.

Improvements:

  * Contextual types. Types are scoped where types are expected. There's no whitelist of "known" types.

  * Better understands the language and its syntax. Better at handling whitespace and comments inside complex forms. Better at handling other nuances. Unlikely to be confused by unusual code style.

  * Vars and consts in the root scope are added to the symbol index and searchable by ⌘R, ⌘⇪R, and `goto_definition`.

  * Supports parenthesized `type (...)` groups, not just `const (...)` and `var (...)`.

  * The first argument to `new` and `make` is scoped as a type.

  * Methods declared inside an `interface` are added to the symbol index, reflecting the fact that they actually exist as functions.

  * Supports embedded structs.

  * Supports inherited interfaces.

  * Supports imaginary number literals.

  * Variable declarations are scoped differently from other variable occurrences.

  * `iota` is scoped only in constant initialization expressions.

  * Symbols in ⌘R are annotated with their origin: `var`, `const`, `type` or `func`.

  * Auto-pairing of backticks.

  * Much more comprehensive tests.

  * All keywords and predeclared identifiers are added to auto-completions by default.

  * Better snippets that don't mess with auto-completion.

  * Slightly better indentation rules that more closely match `gofmt`.

Current shortcomings (?):

  * Fewer meta scopes: needs feedback.

  * No support for block labels: needs feedback.

While this syntax is ready-to-use and fairly complete, I expect the current maintainers, particularly @wbond who's responsible for the current version of the Go syntax, to have many qualms and request many changes. Arguably, this version has some "regressions", particularly where meta scopes are concerned. This needs feedback and discussion, as I don't fully understand what role many of those meta scopes were intended to serve.

While the PR is pending, the syntax is available standalone: https://github.com/Mitranim/sublime-gox

Screenshot before-after:

screen shot 2018-07-28 at 10 54 50

Screenshot of symbol index before-after:

screen shot 2018-07-28 at 10 54 55

(Many other differences didn't fit into the screenshots. This uses my Cloud color scheme, available at https://github.com/Mitranim/sublime-themes.)


Go/Go.sublime-settings
```diff
@@ -0,0 +1,11 @@
+{
+  // Prefer tabs because gofmt enforces tabs.
+  "translate_tabs_to_spaces": false,
```
FichteFoll: This is certainly debatable, but I don't have insight on whether people write Go without gofmt.

Go/Go.sublime-settings
```diff
@@ -0,0 +1,11 @@
...
+  // Trigger autocompletion on dot access.
+  "auto_complete_triggers": [
```
FichteFoll: Many packages use dots for attribute access, however none of them defines `auto_complete_triggers`. I'm not sure whether they should or should not, but this is something we should discuss, imo.

Go/Symbols/GoConstSymbolIndex.tmPreferences
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+  <key>scope</key>
+  <string>
+    variable.other.constant.declaration.go - meta.block.go
```
FichteFoll: `entity.name.constant.go` would be more appropriate, assuming the syntax definition is updated as well.

Go/Symbols/GoConstSymbols.tmPreferences
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
FichteFoll: You can define both, the indexed and non-indexed symbols, in the same file, assuming their scope selectors are equal. This applies to the other files as well.

Go/Default.sublime-keymap
```diff
@@ -0,0 +1,45 @@
...
+    [
+      { "key": "selector", "operator": "equal", "operand": "source.go", },
+      { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+      { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+      { "key": "following_text", "operator": "regex_contains", "operand": "^`", "match_all": true },
+    ]
+  },
+  { "keys": ["backspace"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Delete Left Right.sublime-macro"}, "context":
+    [
+      { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
```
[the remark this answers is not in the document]
Mitranim: Thanks, added the missing selector.

Go/Default.sublime-keymap
```diff
@@ -0,0 +1,45 @@
...
+      { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+      { "key": "preceding_text", "operator": "regex_contains", "operand": "`$", "match_all": true },
+      { "key": "following_text", "operator": "regex_contains", "operand": "^`", "match_all": true },
+    ]
+  },
+
+  { "keys": ["enter"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Add Line in Braces.sublime-macro"}, "context":
+    [
+      { "key": "setting.auto_indent", "operator": "equal", "operand": true },
```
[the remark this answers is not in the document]
Mitranim: Thanks, added the missing selector.

Go/Go.sublime-completions
```diff
@@ -0,0 +1,71 @@
+{
+  "scope": "source.go",
+
+  "completions": [
+    // https://golang.org/ref/spec#Keywords
+    "break",
```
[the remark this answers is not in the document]
Mitranim: Hmm the docs don't mention any descriptions.

Apparently, some completions in the standard packages use tab-separated triggers: `word\tdescription`. Tried this; it looks cool in the completion list, but suppresses the normal buffer completions, which is _exactly_ what you don't want, _especially_ in a standard package. If we could add descriptions without messing things up, I'd love to.

Go/Go.sublime-settings
Mitranim: This certainly could be removed. I'll run without this setting for a few days to get a feel.

Go/Symbols/GoConstSymbolIndex.tmPreferences
Mitranim: Agreed. The problem is that applying different const/var scopes depending on whether we're inside a block seems to require _lots_ of duplication in the syntax; not sure it's worth it.

Go/Symbols/GoConstSymbols.tmPreferences
Mitranim: These are defined in separate files to keep `symbolTransformation` from applying to the indexed symbols and messing up "goto_definition". I'd love to be able to merge the files, but it doesn't seem possible while keeping the current rule logic.

Go/Go.sublime-completions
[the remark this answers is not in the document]
Mitranim: Played with generating `sublime-snippet` files for these completions. Snippets have descriptions and don't suppress normal buffer completions. They also don't appear when you just hit `^Space` or equivalent to bring up the completion list. The downside is that they clog up the Command Palette.

Really the problem is that even _one_ `sublime-completion` with a tab-separated description completely suppresses buffer completion. It's ridiculous and wrong.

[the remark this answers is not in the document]
FichteFoll: Meh, sounds like https://github.com/SublimeTextIssues/Core/issues/1061 is striking again. I thought tab triggers wouldn't affect it.

Or maybe it is actually a different issue, but that would be surprising.

Go/Symbols/GoConstSymbols.tmPreferences
FichteFoll: Can you confirm that `symbolTransformation` does indeed affect the symbol that is used for the index if `showInIndexedSymbolList` is set? Previous experience doesn't indicate so (and neither to [the docs](http://docs.sublimetext.info/en/latest/reference/symbols.html) derived from that).

Even with two different files, since the scope selectors are the same it would have the exact same effect as all the meta settings will be set on any token that is matched.

Mitranim: Interesting! I double checked, and indeed, it looks like `symbolTransformation` doesn't affect indexing. Must have mixed it up with something else when I was writing this.

Merged some of the symbol files. Thanks!

Go/Go.sublime-settings
Mitranim: Removed the trigger setting for now.

FichteFoll:
Something I did for my YAML rework was to have a sample file containing lots of examples with bad highlighting in the original and then providing a Screenshot of that file being rendered with and without the pr to give a better before-after comparison.

Mitranim:
@FichteFoll Nice idea. Added comparison screenshots.

cskujbus:
Hi,
I'd like to ask why **type**, **struct**, **func**, **chan** etc. don't have keyword scope? They have storage.type which means they are styled differently than keywords.

Mitranim:
@cskujbus An excellent point. I followed the established convention of scoping type _keywords_ the same as _type names_, but frankly, it makes no sense whatsoever. Perhaps we should break with the convention, and ask the rest of the syntax developers to follow suit.

@wbond @keith-hall

FichteFoll:
https://github.com/sublimehq/Packages/issues/582

Mitranim:
As an experiment, I've scoped these as keywords, and keyword coloring started "popping out" in the middle of type signatures. It feels weird when you're used to the "old" way. I've come to associate keyword coloring with control flow.

image
image

Thoughts?

wbond:
I’m not inclined to revamp the scope names for function and class keywords. Most languages that use those have the ability to treat such as first class citizens, so I think it makes sense the keywords are scoped as `storage.type`. The keyword is describing the type of the identifier, in this case a function, class, etc.

FichteFoll:
For now, I suppose the best action forward is to scope them as `storage.type.class` and `storage.type.function` (and not `func`!) and `storage.type.struct`, so that people *may* colorize them differently if they wish.

Mitranim:
My inner language-author/syntax-author/Lisper feels that all language keywords should have a unified scope, so that a simplified color scheme can consistently mark them as keywords, without special-casing the various `storage.` variants and without affecting the actual types names. However, the screenshots above convince me that for _type keywords_ that may occur inside type signatures, giving them a scope and a coloring similar or identical to _type names_ is a nice default. This is particularly useful for languages with anonymous inline types, such as Go, where a type signature may mix type names with keywords.

Satisfying both use cases requires the scope on keywords like `struct` or `interface` to be distinct from `storage.type` so we can distinguish _type keywords_ from _type names_. The current convention is to scope them identically, which is really dubious.

I would propose `keyword.storage` if we had the luxury of wide breaking changes. Otherwise, `storage.keyword` appears to be the most sensible option. Probably not `storage.modifier` since it more commonly refers to mode of mutation rather than the data layout itself.

Long story short: what if we scope all type keywords, such as `func`, `struct`, `interface`, as `storage.keyword`?

Go/Go.sublime-syntax
```diff
@@ -1,619 +1,1047 @@
...
+  # Note: this matches ALL valid identifiers, including predeclared constants,
+  # functions and types.
+  ident: \b(?!{{keyword}})[[:alpha:]_][[:alnum:]_]*\b
+
+  # Single line only
+  inline_comment: /[*](?:[^*]|[*](?!/))*[*]/
+
+  # Whitespace and general comments on a single line.
+  # This should only be used for lookahead, not for capturing / scoping.
+  noise: (?:\s*|{{inline_comment}})*
```
deathaxe: The first asterisk after `\s` can be omitted here.

```suggestion
  noise: (?:\s|{{inline_comment}})*
```

wbond:
I really wanted to merge this for the next dev build, but it is extensively using `storage.keyword`, which is not part of the guidelines, and has chosen to change them from scopes that are currently the standard.

I don't want to break a bunch of syntax coloring for users, and thus am not inclined to merge this since it takes scope names in a new direction that is different (in a major way) from all of the syntaxes that have been working on over the past couple of years. For consistency sake, I want new syntax work to be consistent with other actively maintained syntaxes. So to merge this, we would have to revamp scope names in a bunch of syntaxes, change the standard and revamp color schemes.

mitranim:
Glad to see you back in action, @wbond!

I've updated the scopes you mentioned to to `storage.type.keyword.X.go`, along with other minor changes. Let me know if there are any other issues.

FichteFoll:
I suppose replacing storage.type.keyword.function with storage.type.function would be appropriate, which is used across multiple syntaxes iirc.

Not sure about interface and const, however.

wbond:
I am more inclined for that format @FichteFoll. The impression I get is that @mitranim is trying to make it possible to color `storage.type.keyword` the same as `keyword`, and so having to enumerate all of the different possible `storage.type.` specializations would not be viable.

deathaxe:
`Time` would need to be scoped using `support.type.go` if it is part of a "basic" framework library. `storage.type.go` is meant for language types only.

(This is what I know from other syntaxes and is how I understand the scope names, which may be wrong)

mitranim:
Thanks for asking. This is an important difference between my syntaxes and most other ST built-ins. I've been meaning to raise this topic for a while now.

For the language user, there's no functional difference between predefined and user-defined types. Standard library types like `time.Time` are user-defined as far as the language is concerned. A type is just a type.

The idea of treating types differently is so weird, I can't even comprehend where it's coming from. Maybe it's inertia from syntaxes that wanted but _couldn't_ scope types properly?

Modern general-purpose languages are explicitly designed to be extensible. Complicated programs tend to develop their own DSLs, speaking mostly in their own "language", often supplanting built-in types and idioms. Treating all types equally, exactly the way the language does, supports this notion.

A syntax's job is to reflect the language. When the language treats all types equally, so does the syntax.

Yes, many built-in ST syntaxes have these weird ideas about "built-in" and "support" types. This doesn't make them any less wrong. I plan to convince other syntax developers to revise this notion across all standard packages.

deathaxe:
Not sure about the origin, but this is just the existing way to achieve @ccampbell 's request.

I honestly find it helpful to see language defined functions (`support.function`) in another color than user defined (`variable.function`) ones. Same might be helpful with basic/builtin/atomic language types like ints/floats/... and complex types like classes/structs/... which are a collection of the atomic types extended by functions/methods/... .

You are right with `support.` describing basic framework functions/types being less suitable for use with user defined types. This is indeed weird. It's even not possible to distinguish them in most languages.

But the possibility to distinguish between atomic and complex types should be applied anyway. Maybe something like `storage.type.go` for simple types and `storage.class.go` for classes (instead of `support.class.go`)  or all in the 3rd level (`storage.type.simple.go` vs.  `storage.type.class.go`)?

Thom1729:
I'm in favor of scoping builtin types and functions.

It must be admitted that scoping builtins is not a truly syntactic concern. In most languages, a built-in function or type is parsed exactly the same as any other type. You can have a completely accurate syntax definition that doesn't scope any builtins at all.

On the other hand, to the extent that builtins are defined by the language itself, I think that it's useful to use scopes to distinguish them. One reason is that builtins generally don't have lexical declarations, so when I see something highlighted as a builtin, I know not to look for one. In the absence of full language-specific intellisense, builtin highlighting also serves as a first line of defense against typos.

I think that support scopes should be purely additive. That is, a builtin function should be scoped just like any other function, plus whatever extra scope builtins get. The core syntaxes don't all do this, and I think they should. For one thing, this makes it easier for users who don't want builtin highlighting to modify their color scheme to disable it.

What builtins deserve highlighting? The answer surely depends on the language. In JavaScript, there is a fixed set of global types and functions that do not need to be included and are available in any environment. These, I think, should be scoped. There are also many other global builtins available in other environments, such as web browsers. I'm not convinced that these should be scoped in the core language. For JavaScript, I'm planning to write an extension for [JS Custom](https://github.com/Thom1729/Sublime-JS-Custom) to allow users to choose between sets of builtins (like browser/node) and to add their own.

FichteFoll:
This is becoming somewhat off-topic and approaching RFC levels. I suggest to summarize a few options and formulate those into a proposal in a several issue. I also have a thing or two to say about this, later.

After all, this doesn't just affect Go.

PR #1667, opened 2018-07-18
[C++/ObjC++] Add highlighting and indexing support for using aliases

### LINKED ISSUE #1578: [C++] Aliases defined with keyword `using` not indexed
Currently, aliases that are defined with `using` are not indexed as symbols whereas those that are defined with `typedef` are. 

### Examples
```C++
class A {};
typedef A B; // B is indexed
using C = A; // C is not indexed
```

mjendruk:
This PR adds indexing support for using aliases, just like typedefs.

fixes #1578

```c++
typedef Foo MyAlias;  
using MyAlias = Foo; 
```

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -435,7 +435,7 @@ contexts:
         - include: expressions
 
   using-namespace:
-    - match: '\b(using)\s+(namespace)\s+(?={{path_lookahead}})'
+    - match: '\b(using)\s+(namespace)'
       captures:
         1: keyword.control.c++
         2: keyword.control.c++
@@ -699,8 +699,6 @@ contexts:
       scope: punctuation.accessor.arrow.c++
       push: members-after-accessor-junction
 
-  
-
   using-alias:
     # consume keyword if followed by typename
     - match: '\b(using)\b(?=\s+typename\b)'
@@ -710,7 +708,7 @@ contexts:
       captures:
         1: storage.type.c++
         2: entity.name.type.using.c++
-        
+
   typedef:
     - match: \btypedef\b
       scope: storage.type.c++
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -234,6 +234,7 @@ contexts:
     - include: comments
     - include: case-default
     - include: typedef
+    - include: using-alias
     - include: keywords-angle-brackets
     - include: keywords-parens
     - include: keywords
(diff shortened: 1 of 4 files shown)


rwols:
Checked this out on my company's codebase, and it seems to work great. Nice job!

mjendruk:
Hi! Sorry I forgot to make some changes for ObjC too.

C++/C++.sublime-syntax
```diff
@@ -694,6 +699,18 @@ contexts:
       scope: punctuation.accessor.arrow.c++
       push: members-after-accessor-junction
 
+
```
wbond: Could we remove the extra blank lines here?

C++/C++.sublime-syntax
```diff
@@ -431,7 +434,7 @@ contexts:
           pop: true
         - include: expressions
 
-  namespace:
+  using-namespace:
     - match: '\b(using)\s+(namespace)\s+(?={{path_lookahead}})'
```
wbond: We need to add handling for the instance when the user has typed just `using namespace`. Currently the `using` and `namespace` end up getting highlighted with `namespace` is a type definition. That shouldn't ever be the case.

On a high level, the issue is that it is preferable for fully-typed keywords to not shift around after a following token has been typed. Obviously this isn't possible in all cases, but we'd like to try and keep the scope changes minimal as the user types.

C++/C++.sublime-syntax
```diff
@@ -694,6 +699,18 @@ contexts:
...
+  
+
+  using-alias:
+    # consume keyword if followed by typename
+    - match: '\b(using)\b(?=\s+typename\b)'
+      captures:
+        1: keyword.control.c++
+    - match: '\b(using)\b\s+({{identifier}})(?!\s*(<|::))'
+      captures:
+        1: storage.type.c++
```
wbond: This should consistently be `keyword.control.c++`. I don't think the scope should change to `storage.type`.

mjendruk: The reasoning behind this was that `typedef`, `class`, `enum`, etc. all have the keyword `storage.type` and I wanted it to be consistent. On the other hand, `namespace` when used for namespace aliases – as in `namespace foo = bar;` – doesn't change its scope either and sticks with `keyword.control`. For me both choices seem reasonable so I will leave it up to you.

wbond:
Thanks for these tweaks!

PR #1668, opened 2018-07-20
allow transpose for curly brackets

rmcdermo:
In Matlab, you can run into situations where the transpose of a cell array H behaves differently if you take H' or H{:}' (see example below).   We have a project that makes use of the H{:}' construct.  This PR fixes the highlighting for that case and does not seem to adversely affect the syntax_test_matlab.m file.

```
A = 'a,b,c,d'
H = textscan(A,'%q','delimiter',',')
h = H'
h2 = H{:}'
```

results in
```
A =

    'a,b,c,d'


H =

  1x1 cell array

    {4x1 cell}


h =

  1x1 cell array

    {4x1 cell}


h2 =

  1x4 cell array

    {'a'}    {'b'}    {'c'}    {'d'}
```

### DIFF
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -134,7 +134,7 @@ contexts:
         - meta_content_scope: meta.brackets.curly.matlab
         - match: '\}'
           scope: meta.brackets.curly.matlab
-          pop: true
+          set: transpose_post_parens
         - include: allofem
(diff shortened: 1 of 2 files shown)


wbond:
@rmcdermo Sorry for the delay in getting to this. It appears I already merged another PR that fixed this same issue (#1692). This fix will be part of the next dev build.

PR #1675, opened 2018-08-01
[PHP] added functions apcu_*

sptndc:


PHP/PHP.sublime-completions
```diff
@@ -48,6 +48,18 @@
 		{ "trigger": "apc_load_constants", "contents": "apc_load_constants(${1:key})" },
 		{ "trigger": "apc_sma_info", "contents": "apc_sma_info()" },
 		{ "trigger": "apc_store", "contents": "apc_store(${1:key}, ${2:var})" },
+		{ "trigger": "apcu_add", "contents": "apcu_add(${1:key}, ${2:var})" },
+		{ "trigger": "apcu_cache_info", "contents": "apc_cache_info()" },
```
michaelblyons: I don't know anything about PHP, but it looks like this is missing a `u` in the contents **or** this line should be removed as a duplicate of line 37.

wbond:
Thanks!

PR #1680, opened 2018-08-06
[JavaScript] Implemented unicode escapes in identifiers.

### LINKED ISSUE #1679: [JavaScript] \u1234 not recognized as function name!
Hello!

Take a look at this:

```js
+function \u1234(){}/1//*

console.log('THIS IS NOT A COMMENT!');

//*/
```

the line containing console.log is highlighted like a comment, while it isn't. The problem, is, I think, that st doesn't recognize \u1234 as a function name and then it probably thinks that `{}` is an object and therefore it thinks that `/1/` is a regular expression, and then it thinks that `/*` is start of a multiline comment, and so on, until it reaches the second line and highlights actual code like a comment and messes everything up!

I did my best to format this code snippet like it is formated in the table in issue #1600, I hope I did it correctly. Couldn't attach screenshot atm.

This must be fixed, I think. Any opinions?

Thom1729:
Fixes #1679.

Implements [Unicode escape sequences in JavaScript identifiers](http://www.ecma-international.org/ecma-262/#sec-names-and-keywords).


JavaScript/JavaScript.sublime-syntax
```diff
@@ -8,8 +8,9 @@ file_extensions:
 first_line_match: ^#!\s*/.*\b(node|js)\b
 scope: source.js
 variables:
-  identifier_start: '[_$\p{L}\p{Nl}]'
-  identifier_part: '[_$\p{L}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}\x{200C}\x{200D}]'
+  identifier_escape: (?:\\u(?:\h{4}|\{\h+\}))
+  identifier_start: (?:[_$\p{L}\p{Nl}]|{{identifier_escape}})
+  identifier_part: (?:[_$\p{L}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}\x{200C}\x{200D}]|{{identifier_escape}})
```
bounc3-paradise-on-e: :+1:

JavaScript/tests/syntax_test_js.js
```diff
@@ -321,6 +321,15 @@ not_a_comment;
...
 //      ^^^ variable.other.readwrite
 
+    function \u123fxyz () {};
+//           ^^^^^^^^^ entity.name.function
+
+    function xyz\u123fxyz () {};
+//           ^^^^^^^^^^^^ entity.name.function
+
+    function xyz\u{deadbeef42}xyz () {};
+//           ^^^^^^^^^^^^^^^^^^^^ entity.name.function
```
bounc3-paradise-on-e: `\u{deadbeef42}` is not a valid escape sequence. It is ok to support superset of valid identifiers, but why is invalid syntax included in test for valid syntax? How about `function xyz\u{123f}xyz () {};` instead?

Thom1729: It's not semantically valid (the code point is undefined) but it is *syntactically* valid and should be parsed accordingly. The [CodePoint production](https://www.ecma-international.org/ecma-262/#prod-CodePoint) consists of one or more hex digits with no other limitation. By contrast, with no brackets, the grammar explicitly requires four hex digits.

bounc3-paradise-on-e: No, it is not syntactically valid. In the spec it says (section 10.1.1):

```
1. Assert: 0 ≤ cp ≤ 0x10FFFF
```

where `cp` is escaped code point. The value `0xdeadbeef42` is not a valid UTF-16 code point, and thus it is reported as an EarlyError. Every code that yields an EarlyError is considered as syntactically invalid.

For example `0xfffe` is not a valid unicode character, but JS code `'\u{fffe}'` is *syntactically* valid, while `0x0xdeadbeef42` is outside UTF-16 range.

> The CodePoint production consists of one or more hex digits **with no other limitation**

It says *"The definitions of the nonterminal UnicodeEscapeSequence is given in 11.8.4."*
In 11.8.4. it says *"The definition of the nonterminal HexDigit is given in 11.8.3. SourceCharacter is defined in 10.1."*
And in 10.1. it says that if `cp` is larger than `0x10FFFF` it is considered as an EarlyError.

Until new spec is released (or until UTF-16 is extended), this is invalid syntax. All intepreters reject this code.

---

I am ok with ST supporting arbitrary number of hex digits. My point was that including invalid syntax in tests as demonstration of a valid code is not appropriate. But if others are ok with that, I'll not complain.

Thom1729: The governing rule is [11.6.1.1](https://www.ecma-international.org/ecma-262/#sec-identifier-names-static-semantics-early-errors). It is an early error if the value of the escape sequence is greater than 10ffff, but also if it is a valid unicode code point that is not allowed at that position in the identifier. In general, we cannot detect this (without an absurdly complicated regexp).

If such an escape sequence is used, then a conforming parser will parse it according to the ECMAScript grammar, then report an early error. I see Sublime's highlighting as corresponding to the first part of that process. With rare exceptions, it is impractical to try to detect syntax that conforms to the grammar but would cause an early error.

This is to say that the question of whether a given construct is valid ECMAScript syntax has some subtleties. The text `let \u0030` (for instance) is parsed by the grammar, but it is not a valid program; yet, `let \u0050` is a valid program, as is `let a\u0030`. Furthermore, `let \u0063onst` is invalid, but `let \u0062onst` and `let \u0063onst1` are valid. The grammar allows all of these, but the invalid programs trigger early errors that result in `SyntaxError`s.

All that said, it does make sense for the code in the Sublime syntax tests to be completely valid, unimpeachable JavaScript code that you or I could run in a browser or other spec-compliant JavaScript environment. I'll write up a change tomorrow to fix `\u{deadbeef42}`. (I almost wrote that I was going to fix `\u123f`, except that, by pure chance, the “Ethiopic syllable schwa” is a valid JavaScript identifier. Nevertheless, this could probably be made more obviously correct.)

wbond:
@Thom1729 We are just waiting on a tweak here from you, right?

Thom1729:
Ah, yes, I forgot about this. Just a brief update of the test. I'll have it in a jiffy.

PR #1690, opened 2018-08-19
[Scala] More robust unit constant and other misc

djspiewak:
I wasn't quite sure how to group these things and I didn't want to open a million PRs because it creates conflicts in the tests. So, here you go:

- Unit constant is now highlighted in patterns
- Unit constant is *not* highlighted in dispatch position following primitive types
- Varargs operator allows spaces (e.g. `_ *`), which I legitimately didn't know
- `ne` is an operator
- I screwed up the original version of parentheses nesting in val patterns; they now nest arbitrarily
- Lambda lookahead now supports bare types with spaces (bad style, but still valid)


djspiewak:
I pushed some new commits which address a few minor regressions I introduced with the unit handling fixes.

PR #1698, opened 2018-08-29
[MATLAB] add parfor as a keyword

monty5811:
Add parfor as a keyword in MATLAB syntax.

(https://uk.mathworks.com/help/distcomp/parfor.html)

### DIFF
--- a/Matlab/syntax_test_matlab.m
+++ b/Matlab/syntax_test_matlab.m
@@ -201,3 +201,9 @@ s1="00:06:57";
 %   ^^^^^^^^ string.quoted.double.matlab
 %           ^ punctuation.definition.string.end.matlab
 
+%---------------------------------------------
+% parfor keyword test
+% parfor
+parfor x = 1:10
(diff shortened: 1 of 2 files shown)


monty5811:
Of course. I _think_ I have added a test.

Matlab/syntax_test_matlab.m
```diff
@@ -201,3 +201,9 @@ function foo(bar)
 %   ^^^^^^^^ string.quoted.double.matlab
 %           ^ punctuation.definition.string.end.matlab
 
+%---------------------------------------------
+% parfor keyword test
+% parfor
+parfor x = 1:10
+^ keyword.control.matlab
```
FichteFoll: Syntax test lines need to begin with a certain token that is defined in the first line of the file (and is usually the line comment punctuation, I. E. `%` here).

monty5811: Thanks! Fixed now I think.

PR #1703, opened 2018-09-02
[ShellScript] Update Bash.sublime-syntax: add support for 'ash' shell

sc0ttj:
Add support for Ash shell too, so these files are auto-highlighted, just like sh, bash, zsh, etc ..

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -32,7 +32,7 @@ first_line_match: |
     ^                          # assert start of the line
     \#!                        # shebang token
     .*                         # any characters
-    \b(bash|zsh|sh|tcsh|ash)\b # "bash" or "zsh" or "sh" or "tcsh"
+    \b(bash|zsh|sh|tcsh|ash)\b # "bash" or "zsh" or "sh", "tcsh" or "ash"
     |                          # ... or ...
     ^                          # assert start of the line
     \#                         # literal "#" character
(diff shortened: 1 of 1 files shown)


ShellScript/Bash.sublime-syntax
```diff
@@ -27,22 +28,22 @@ file_extensions:
...
-    mode:                  # literal "mode:" string
-    \s*                    # any whitespace
-    shell-script           # literal "shell-script" string
-    [^*]*                  # any characters except "*"
-    -\*-                   # literal "-*-" string
+  (?x)                         # ignore whitespace in this regex
+    ^                          # assert start of the line
+    \#!                        # shebang token
+    .*                         # any characters
+    \b(bash|zsh|sh|tcsh|ash)\b # "bash" or "zsh" or "sh", "tcsh" or "ash"
```
rwols: I think you could have just changed `bash` into `b?ash`

FichteFoll: I prefer the explicit version here. It's more readable that way.

FichteFoll:
By the way, the way this regex is commented is atrocious, as indicated by every line having to be changed in order to align the comments, while explaining step by step each regex token used, more or less. Not every regex should include an introduction to regular expressions themselves, so I'd rather write this as follows:

```yaml
first_line_match: |-
  (?x)
    ^\#! .* \b(bash|zsh|sh|tcsh|ash)\b
  | ^\# \s* -\*- [^*]* mode: \s* shell-script [^*]* -\*-
```

Feel free to change your PR to make this change as well. I don't think anyone will complain.

wbond:
I pulled the changes in via 54f0cf6af385dd10de44b767dbf3a1d6a21342f2, as this PR was conflicted. I also grabbed your reformatting of the regex @FichteFoll.

Thanks everyone!

PR #1705, opened 2018-09-04
Add bazel files for python

coltonmorris:


wbond:
Are bazel files really Python? I looked at the Bazel site some, and it seemed to imply it was a custom DSL.

laurentlb:
It's a Python dialect (described here: https://github.com/bazelbuild/starlark). Ideally there should be a separate syntax for it. But maybe using Python syntax is an improvement in the meantime?

FichteFoll:
From the example in the referenced Readme, this really is very python-like (as it is valid python code and uses quite a few of its features). I suppose this is an okay choice to make.

PR #1706, opened 2018-09-05
[Ruby] Line regular expression works w/ 1.8 - 2.5

lzap:
Hello, the current regexp does not work correctly:

    "file_regex": "^(...*?):([0-9]*):?([0-9]*)"

There are multiple backtrace formatting in Ruby. In versions 1.8 - 2.4 the
format is:

    test.rb:7:in `partridges': Shouldn't this be recursive? (RuntimeError)
      from test.rb:3:in `turtle_doves'
      from test.rb:10:in `<main>'

Starting from Ruby 2.5 it's reversed with index numbers:

      2: from test.rb:10:in `<main>'
      1: from test.rb:3:in `turtle_doves'
    test.rb:7:in `partridges': Shouldn't this be recursive? (RuntimeError)

The regular expression in Sublime Text 3 (build from summer 2018) does only 
capture line 7 and completly ignores "from" lines. But there's more, exceptions 
are often printed to console via `backtrace` call from `Exception` class from
Ruby standard library. An example from our application:

    E, [2017-01-16T12:15:59.678835 #32142] ERROR -- : undefined method `join' for #<String:0x007f1a587784a8> (NoMethodError)
    /usr/share/sinatra-1.3.5/lib/sinatra/showexceptions.rb:37:in `rescue in call'
    /usr/share/sinatra-1.3.5/lib/sinatra/showexceptions.rb:21:in `call'
    /usr/share/sinatra-1.3.5/lib/sinatra/base.rb:124:in `call'
    ...

Testing frameworks does the same often prepending whitespace (tab usually) for 
each line:

    NoMethodError: undefined method `render' for "test":String
        app/models/host/base.rb:351:in `render_template'
        /home/lzap/work/foreman_discovery/lock_templates.rb:14:in `block in lock_templates'
        /home/lzap/work/foreman_discovery/lock_templates.rb:10:in `each'
        /home/lzap/work/foreman_discovery/lock_templates.rb:10:in `lock_templates'

Sublime Text needs a regular expression which captures them all. And this is the purpose
of my PR. Relevant reading:

https://engineering.freeagent.com/2017/12/21/ruby-2-5-how-is-that-backtrace/

### DIFF
--- a/Ruby/Ruby.sublime-build
+++ b/Ruby/Ruby.sublime-build
@@ -1,5 +1,11 @@
 {
 	"shell_cmd": "ruby \"$file\"",
-	"file_regex": "^(...*?):([0-9]*):?([0-9]*)",
-	"selector": "source.ruby"
+	"file_regex": "^\\s*([^:]+):(\\d+):in() `[^']+':(.*)$",
+	"selector": "source.ruby",
+	"variants": [
+		{
+			"name": "Ruby (backtrace)",
+			"file_regex": "^\\s*(?:\\d:)?\\s*(?:from )?([^:]+):(\\d+):in() `[^']+':?(.*)$"
+		}
+	]
 }


FichteFoll:
Just for confirmation, do you actually *want* to have errors (jump to next error and phantoms) for the "from" traceback lines?

lzap:
@FichteFoll well I wish I could define regexps to match errors and traceback lines separately, but documentation does not appear to support that. I guess the build feature is more focused on building obviously then errors from interpreted languages. I have to admit it looks ugly but it lets me to click on all lines from stacktraces.

However, I pushed new version which contains two builder definitions - one is "correct" and it only captures lines with errors (including the error). And then there is new one called "Ruby (backtrace)" which captures them all. Error is set to empty string, so there is still a small empty "bubble" rendered in the text pane. But this looks better and it's opt-in.

I've created an idea to have a separate regexp tho:
https://forum.sublimetext.com/t/regexp-patter-to-match-runtime-stacktraces-in-build-window/38721

wbond:
Can you post a screenshot of how this works with `show_errors_inline` and the traceback mode? I am imagining it won't look pretty.

We have a plan for drastically improving the build output capturing in a future dev cycle, so for now perhaps we just fix the breakage of current functionality with Ruby 2.5?

lzap:
Sure, here it is:

2019-02-13-155833_1276x1401_scrot

wbond:
It would seem to me this isn't compatible with inline errors, so I doubt we'll merge it as-is.

lzap:
Feel free to close then! Cheers.

wbond:
I'm going to close this for now. In a future release I'm hoping to have a more robust solution for error matching that will give the power of something like this, but without the erroneous markers for inline errors.

PR #1708, opened 2018-09-07
[Scala] Support for raw strings

djspiewak:
I'm not 100% sure why anyone uses this feature in Scala, but it's there:

```scala
raw"foo\nbar"
```

That's equivalent to:

```scala
"""foo\nbar"""
```

Which is to say, the string `"foo\nbar"`, exactly as written. It now highlights accordingly.


Scala/Scala.sublime-syntax
```diff
@@ -990,6 +990,19 @@ contexts:
           pop: true
         - match: \n
           scope: invalid.string.newline.scala
+    - match: '\b(raw)(")'
+      captures:
+        1: support.function.scala
```
FichteFoll: In Python I decided to use `storage.modifier` for this, but it's currently unclear whether this is the desired scope for string modifiers such as raw strings. I'm unsure where this was discussed, however.

Scala/Scala.sublime-syntax
```diff
@@ -990,6 +990,19 @@ contexts:
...
+        2: punctuation.definition.string.begin.scala
+      push:
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.raw.interpolated.scala
+        - match: '"'
+          scope: punctuation.definition.string.end.scala
+          pop: true
+        - match: \n
+          scope: invalid.string.newline.scala
```
FichteFoll: You should most likely pop here.

PR #1714, opened 2018-09-10
[Scala] Encoded even/odd newline semantics in single type expressions

### LINKED ISSUE #1054: [Scala] Incorrect highlighting for modifier preceded by abstract method
If a `lazy val`, `override def`, etc. line is immediately preceded by an abstract `def` with a return type, the highlighting is incorrect

```scala
trait A {
  def a: Int
  lazy val b: Int = 123 // "lazy val b" is incorrect

  def c: String
  override def d = "blah" // also incorrect
}
```

The highlighting seems to affected specifically if the preceding line is an abstract `def`

(I'm using build 3139 on Ubuntu)

### LINKED ISSUE #1691: [Scala] Paren-less `new` and `type` eats trailing identifiers
```scala
   type S = Map
   evalNT
// ^^^^^^ - support.type.scala

   new Foo
   evalNT
// ^^^^^^ - support.type.scala
```

These tests should pass. They do not at present.

djspiewak:
There are good motivating examples in the linked issues.

Fixes #1691 
Fixes #1054


Scala/Scala.sublime-syntax
```diff
@@ -1376,6 +1376,11 @@ contexts:
       set: single-type-expression-tail
     - match: (?=@{{plainid}})
       set: single-type-expression-tail
+    - match: \(
+      set:
+        - match: \)
```
rwols: Shouldn't this be scoped with some variation of a `punctuation` scope?

djspiewak: Yes!

djspiewak:
Note that there's an outstanding issue with this PR related to instantiations of the following form:

```scala
new Foo[A].bar
```

The `bar` is not part of the type and should not be scoped as such. Right now, we don't consider square brackets as a stop on the initialization.

djspiewak:
Okay I fixed the outstanding issues, though at a small cost: I had to remove literal value type scoping (SIP-23). That kind of sucks. I know how to restore it, I just have to copy/paste a ton of stuff, so I'm punting on it for now.

djspiewak:
FYI, the build failures are spurious. It looks like the infrastructure just wasn't working when I pushed the changes to this PR

wbond:
@djspiewak I got most of your PRs resolved and merged. Mind taking a look and fixing this one up?

djspiewak:
@wbond Yep! No problem.

djspiewak:
Sorry for close. Apparently Cmd-Shift-Return closes and comments?

djspiewak:
@wbond Done

djspiewak:
@wbond Those should in theory be there ultimately, but they can't be right now since they break scoping (due to the way they handle lookahead and popping). Since we can't have parameterized contexts, it's not really possible to include their functionality without copy/paste duplicating everything, but that in turn multiplies with the fact that the type lookahead stuff *already* requires copy/paste duplicating everything in order to achieve the even/odd semantics. It's kind of a mess.

`sublime-syntax` as a format doesn't provide a good way out of this knot. I've been considering looking into YAML references as away forward. In the short term, I simply disabled the functionality. So in a sense this is a regression, but only with [SIP-23](https://docs.scala-lang.org/sips/42.type.html) behavior. This is something that I think is important to re-enable, but it doesn't regress anything in the current mainline Scala source code (only feature-flagged functionality in future Scala versions), so I'm okay with it for now.

Issue #1713, opened 2018-09-09
[C++] syntax highlight problem (bug?)


imfl:
This template function doesn't seem to be properly highlighted. Build 3176.

```c++
template<typename T> container<bar<T>> foo(T t)
{
	return container<bar<T>> { item<bar<T>>(t) };
}
```

<img width="321" alt="default" src="https://user-images.githubusercontent.com/32955896/45259994-aecbd000-b40c-11e8-9df0-f27128347219.PNG">

It seems to wrongly tell the last curly brace is surplus, which is not. The code compiles smoothly.

The crux seems to be linked with `<bar<T>>`.

To explain the code, it generates a container object list initializing with just one item, which is from a template class, with compound type `<bar<T>>`.

Please take a look. Thanks.

rwols:
Works fine for me:

<img width="509" alt="schermafbeelding 2018-09-12 om 14 25 56" src="https://user-images.githubusercontent.com/2431823/45424797-df736a00-b697-11e8-88b6-81a22ce92475.png">

Maybe the problem lies somewhere higher up in your file? It looks like that `template` token is already mis-scoped.

imfl:
@rwols Thanks for trying that on your end!

Nonetheless, if I have a `.cpp` file with only these 4 lines, the problem still exists. Build 3176. No external packages.

<img width="372" alt="default" src="https://user-images.githubusercontent.com/32955896/45537075-e637e380-b835-11e8-8e7f-6d0cbd0bcb19.PNG">

Quite strange.

rwols:
So as it turns out the problem is even earlier. It's the `container<bar<T>>` that's scoped incorrectly. It should, I think, be a type initialized with an initializer-list, but [the lookahead here](https://github.com/sublimehq/Packages/blob/master/C%2B%2B/C%2B%2B.sublime-syntax#L508) thinks it's actually a templated function call. That's because of the opening parenthesis for the `t` argument on the same line. If you put that on the next line the error goes away. Giant lookaheads are a pain in the butt for ST.

rwols:
Two minimal test cases:
```c++
template<typename T> C<T> f(T t)
{
    return C<T> { g<X<T>>(t) };
    /*     ^ - variable.function */
    /*          ^ punctuation.section.block.begin */
}

template<typename T> C<X<T>> f(T t)
{
    return C<X<T>> { g<X<T>>(t) };
    /*     ^ - variable.function */
    /*             ^ punctuation.section.block.begin */
}

```

PR #1715, opened 2018-09-14
[C++/ObjC] Fix: generic_lookahead variable

rwols:
The generic_lookahead variable didn't work quite right when uniform init syntax
was involved. It hopefully does now. A `<.*?>` was changed into `<[^{]*>` in that variable.

Fixes https://github.com/sublimehq/Packages/issues/1713.


rwols:
Superseded by https://github.com/sublimehq/Packages/pull/1727.

PR #1718, opened 2018-09-17
[C/ObjC] Fix: be more lenient about macro identifiers with enum return types

### LINKED ISSUE #1712: [C] enum scope does not close
In the following example, void is not scoped as a storage.type

```
static enum APIC_CAPABILITY apic_capabilities(void)
{
....
}
```

rwols:
Fixes #1712.

Since there were not any tests that failed after removing a line in the match rule, I am somewhat hesitant about merging this. @wbond let me know what could go wrong in this case.

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -465,12 +465,24 @@ contexts:
   data-structures-struct-definition:
     - meta_scope: meta.struct.c
     - include: data-structures-definition-common-begin
+    - include: data-structures-definition-common-macro
     - match: '{{identifier}}(?=\s*;)'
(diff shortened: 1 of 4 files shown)


wbond:
I believe the `__declspec` condition is required to handle the format:

```C++
class __declspec(dllimport) X {};  
```

(From https://msdn.microsoft.com/en-us/library/dabb5z75.aspx)

wbond:
I added some fixes and tests for handling `__declspec`, and it turned out not to require the lines you removed.

rwols:
Thanks! I totally forgot about this PR!

PR #1722, opened 2018-09-18
[JavaScript] Fix: Remove unnecessary semi-colon before 'case'

Aymkdn:
When we have the below : 
```javascript
switch (var) {
  case 'a': {
    // code
    break;
  }
  case 'b': {
    // code
   break;
  }
}
```

`case b:` is not syntax-highlighted until we add a semi-colon before it, just like the below
```javascript
switch (var) {
  case 'a': {
    // code
    break;
  }; // <--- semi-colon required to have "case 'b'" highlighted by the syntax
  case 'b': {
    // code
   break;
  }
}
```

This behavior is incorrect, especially when you use eslint because it will return `Unreachable code (no-unreachable)` for that unecessary semi-colon.

So by removing `- expect-case-colon` we can keep the syntax-highlight on `case` and have eslint happy.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -732,7 +732,6 @@ contexts:
     - match: case{{identifier_break}}
       scope: keyword.control.switch.js
       push:
-        - expect-case-colon
         - expression
 
     - match: default{{identifier_break}}


Thom1729:
This should already be fixed by #1606. That change is merged, but not yet part of a release.

Aymkdn:
Thanks ! Good to know ! So I guess I can close it.

PR #1723, opened 2018-09-18
[Graphviz] Completions

michaelblyons:
The scopes are (overly) restrictive to keep attributes from crowding out the local wordlist. Even so, the colors might need to be nixed.


wbond:
I'd say, let's try it out.

Issue #1327, opened 2017-11-25
[C++] lambda's and std::function as argument


rwols:
Some strange lookahead business is occurring within a lambda:
```c++
#include <functional>
template <class T> struct A {};
template <class T> struct B {};
struct C {};
int main() {
    auto f = [](std::function<A<B<C>>()> g) -> void { auto r = g(); };
    /*                                      ^^ missing punctuation */
    /*                                         ^^^^ missing storage.type */
    /*                                              ^ missing punctuation */
    /*                                                              ^ not invalid.illegal */
    return 0;
}
/* <- not invalid.illegal */
```

PR #1727, opened 2018-09-19
[C++/ObjC] Fix: Function type in template args

rwols:
Fixes https://github.com/sublimehq/Packages/issues/1713.
Fixes https://github.com/sublimehq/Packages/issues/1327.

For things like std::function<A<B<C>>()> f(g), the grammar got confused and
thought this was a templated function call (the function name being "function"
and the start of the opening parenthesis for the function call being the
parenthesis after the "f" identifier). However, this is a new variable called
"f" of type "std::function<A<B<C>>()>" initialized with the variable "g".

To address this, the generic_lookahead variable had to be changed to account
for possible opening and closing parentheses inside of its template argument.

The result is a somewhat insane generic_lookahead, but it makes the syntax more
correct.


rwols:
Note that I cherry-picked the added tests from https://github.com/sublimehq/Packages/pull/1715 -- they are solved automatically by the changes to the generic_lookahead variable.

PR #1739, opened 2018-10-04
[Rust] Add eprint and eprintln macros

sharkdp:
New versions of Rust support `eprint!(..)` and `eprintln!(..)` in addition to the `print!(..)` and `println!(..)` macros: https://doc.rust-lang.org/stable/std/macro.eprint.html


wbond:
Thanks!

PR #1740, opened 2018-10-10
[PHP] Add JsonSerializable built-in class

jfcherng:


FichteFoll:
Might be worth splitting that regex up in the process. Past tests have shown a flattened regexp to have better performance with sregex than nested regexps.

deathaxe:
Sure?

Just did a benchmark with flat vs. nested regexps without any noticeable performance differences.

FichteFoll:
I'm not able to find a source on that, although I strongly believe it to be mentioned in an issue or PR on this repo.

keith-hall:
Not that it's easy to find things while GitHub has a massive outage, but I believe these are relevant:

- https://github.com/sublimehq/Packages/pull/326#issuecomment-215669122
- https://github.com/sublimehq/Packages/pull/349

FichteFoll:
I knew I could count on you :wink: https://github.com/sublimehq/Packages/pull/349#issuecomment-217343544 was what I was remembering.

deathaxe:
Not sure whether this issue still applies. I heavily use such kind of regexps in my CNC syntax as well.

https://github.com/deathaxe/sublime-s840d/blob/edc30e8effba50a9839e3292e97a14b19fd418fa/gcode/s840d_gcode.sublime-syntax#L1506

Therefore I flattened them but did not realize any improvement in performance. My local syntax tests with 100k lines of code don't show any difference in processing time.

The only difference is the size of the cache files, which doubles with flattened regexps.

wbond:
I'll leave flattening the regex for another discussion/PR. I think the win there is readability/maintainability.

This seems like a functional improvement to the current state, so I am going to merge it.

jfcherng:
If the size of the cache files matter, since ST uses its own regex engine and `.sublime-syntax` is ST-only, maybe we could add some extra information to hint the engine to compact the flatten regex while compiling for cached ones? Or ST's regex engine check the group is compactable automatically without any given hint. 

`Packages/PHP/PHP Source.sublime-syntax`
```yaml
- match: |-
  (?x:
    (?#ST_REGEX_GROUP_COMPACTABLE)
    AA |
    ABC |
    ABD
  )
```

compiles to

`Cache/PHP.sublime-syntax.cache` (non-binary form)
```yaml
- match: |-
  (?x:
    A(A|B(C|D))
  )
```

keith-hall:
It's not ST/SM only - there are many other projects using `.sublime-syntax` files these days. Anyway, I believe I read that sregex already optimizes/"compacts" the literals in the regexes automatically without a need for any hints.

deathaxe:
But @jfcherng is right when he says a compacted regex pattern results in much smaller cache files. While it doesn't have any effect on performance, the bigger cache files seem to indicate the compiled pattern to need more RAM to do the same thing. In my case the flattened cache files were about 30 - 40% bigger.

Issue #1734, opened 2018-09-26
Tcl syntax and double brace issue fails to match braces


Meshparts:
Brace matching fails with Tcl syntax activated when double braces are used in code:

proc Test {args} {
set test {{1 -1 0}}
}

In this example, last brace fails to be matched with corresponding opening brace.

See also disscussion at 
https://forum.sublimetext.com/t/matching-braces-issue/39393/6

deathaxe:
Do double braces have a certain meaning or are the inner braces to be ignored only?

Meshparts:
In Tcl Braces are used to define a string which content will be send to the interpreter without any further evaluation. One can also use braces to define lists with elements separated by empty spaces. The example above defines a nested list with only one sublist. The single sublist contains the three numbers.

deathaxe:
OK, I guess I found a solution to fix your issue.

But ST still highlights the inner list as string because it expects a whitespace after the `{` like in the following example:

Example:

```TCL
set test {{ 1 -1 0 }}
```
Your descriptions sounds like this assumption is not very perfect?

Meshparts:
Thanks for the help. 
Adding white spaces is an workarround indeed. 
Do I get it right? In the next releases it will work also without white spaces?
I wonder if this bug has also other side effects. Seems like the meaning of braces for Tcl is not rely understood currently in ST.

Meshparts:
"Your descriptions sounds like this assumption is not very perfect?" 
Please reformulate. I don't undertande what you are referring to.

deathaxe:
> Please reformulate.

ST highlights all tokens beginning with a `{` and followed by a none-whitespace character as string until the next `}` is matched. This results in `{1 -1 0}` not to be highlighted as list but as string.

This is just how ST tries to distinguish between a normal string and list content. Seems like that assumption needs some refinement?

> Do I get it right? In the next releases it will work also without white spaces?

Depends on whether the related PR is accepted.

> I wonder if this bug has also other side effects.

Yes, #1681 is related. According to some statements from the core devs about TCL, all arguments are basically strings only. ST tries to highlight them best it can but TCL's syntax seems not to be very restrictive, which requires some heuristics to guess how to handle some expressions.

> Seems like the meaning of braces for Tcl is not rely understood currently in ST.

[...]

PR #1751, opened 2018-10-24
[TCL] Fix nested lists and strings

### LINKED ISSUE #1681: [TCL] Incorrect detection and highlighting of lists
Lists are not recognized correctly if there is no whitespace after curly bracket.
Here is an example to show how brackets and lists are recognized:
```tcl
foreach var ${list_of_vars} {
    set list_of_lists {{first ONE} {second TWO} {third THREE} {fourth FOUR}}
}
```
list_of_lists

I solve this by replacing \s with \s* on lines 143 and 418, but I'm not sure that it is correct solution.

deathaxe:
Fixes ~#1681,~ #1734

This commit modifies highlighting by only scoping the most inner {foo} token as string. If there are several levels of opening braces, they are handled as blocks.

This change allows highlighting of nested lists correctly.

### DIFF
--- a/TCL/Tcl.sublime-syntax
+++ b/TCL/Tcl.sublime-syntax
@@ -396,6 +396,7 @@ contexts:
       pop: true
     - match: '\\[\\{}n]'
       scope: constant.character.escape.tcl
+    - include: braces
 
   braces:
     - match: (\{)(\*)(\})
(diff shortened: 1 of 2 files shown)


dabro666:
Please check [my comment](https://github.com/sublimehq/Packages/issues/1734#issuecomment-433818514) in #1734

PR #1752, opened 2018-10-24
[TCL] Some optimizations

deathaxe:
This PR contains some smaller optimizations for the TCL syntax. They don't change functionality. The changes are described in the commit messages.

### DIFF
--- a/TCL/Tcl.sublime-syntax
+++ b/TCL/Tcl.sublime-syntax
@@ -84,15 +84,15 @@ contexts:
     - match: '{{unquoted_string}}'
 
   numbers:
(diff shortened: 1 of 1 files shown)


deathaxe:
Sounds reasonable. But anyway I have the feeling the syntax could get even more optimizations, but I hesitate to change too much as I don't really know that syntax and don't want to break something.

I was never a friend of tokens like `\b(namespace)\s+eval\s+({{var_unquoted_string}})` as user experience while typing might not be optimal.

Compared to that the "end of file" issue is negligible, I think.

PR #1757, opened 2018-10-28
[Diff] highlight diff section hints

deathaxe:
The toc-list of a changed hunk in a patch file can contain hints about the context of a change - maybe a function name.

As this hint was not yet respected in the diff syntax, the hint was highlighted as normal text. If a color scheme defines a background color not the whole line was highlighted.

Hence this commit adds support to properly highlight section hints.

### DIFF
--- a/Diff/Diff.sublime-syntax
+++ b/Diff/Diff.sublime-syntax
@@ -25,7 +25,7 @@ contexts:
         1: punctuation.definition.separator.diff
     - match: ^\d+(,\d+)*(a|d|c)\d+(,\d+)*$\n?
       scope: meta.diff.range.normal
-    - match: ^(@@)\s*(.+?)\s*(@@)\s*(.*)\s*$\n?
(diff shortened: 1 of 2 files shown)


Diff/Diff.sublime-syntax
```diff
@@ -25,12 +25,13 @@ contexts:
         1: punctuation.definition.separator.diff
     - match: ^\d+(,\d+)*(a|d|c)\d+(,\d+)*$\n?
       scope: meta.diff.range.normal
-    - match: ^(@@)\s*(.+?)\s*(@@)($\n?)?
+    - match: ^(@@)\s*(.+?)\s*(@@)\s*(.*)\s*$\n?
```
FichteFoll:
```suggestion
    - match: ^(@@)\s*(.+?)\s*(@@)\s*(.*?)\s*$\n?
```

The group match must be non-greedy, otherwise the following `\s*` will never match and everything is consumed in the group already.

deathaxe: Thanks.

PR #1761, opened 2018-11-07
[Latex] Match sizes using dots as optional argument to newline command

tobiasdiez:
Previously, commands like `\\[1.5ex]` that use a dot in the size were not matched correctly (while `\\[1ex]̀  was recognized). This is fixed.


wbond:
Sorry for the delay in getting back to you.

Could you add some tests to make sure this is working and doesn't regress?

tobiasdiez:
I've finally found the time to add tests.

wbond:
@randy3k @FichteFoll Either of you have any thoughts on this?

randy3k:
It looks good to me

wbond:
Thanks @tobiasdiez!

PR #1768, opened 2018-11-13
[JavaScript] Improved for loop conditions.

### LINKED ISSUE #1767: [JavaScript] for..of
ok:

```js
for(let element of document.querySelectorAll('a[href$=".js"]'))
  element.remove();
```

not ok (`of` not colored):

```js
let element = null;
for(element of document.querySelectorAll('a[href$=".js"]'))
  element.remove();
```

Thom1729:
Fixes #1767.

Should correctly differentiate the various types of `for` loops and highlight `in`/`of` where appropriate. `for` loops were apparently undertested, so I've added a few tests.

Semi-related issues:

1. ~~Currently, semicolons in `for` loop conditions are scoped `punctuation.terminator.statement`. This is arguably incorrect; they are separating expressions and/or variable declarations, not terminating statements. They should probably be `punctuation.separator.expression` or somesuch.~~

2. ~~`for (; function() {} /a/g;);` is broken. This is because `for-condition-contents` uses `statements-in-parens`, which uses `statements`, which is wrong because the second and third terms are conditions, not statements. As a result, the `function() {}` is parsed as a function declaration, not a function expression, so the following slash is interpreted as the start of a regexp literal and not a division operator.~~

3. The `in` and `of` in `for` loop conditions are scoped `keyword.operator.word`, which is probably wrong because they're not really operators, but strangely, all of the actual wordlike operators are merely `keyword.operator`.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -690,9 +690,7 @@ contexts:
     - match: (?:const|let|var){{identifier_break}}
       scope: storage.type.js
       set:
-        - - match: (?:of|in){{identifier_break}}
-            scope: keyword.operator.word.js
-            set: expression
+        - - include: for-condition-decider
           - match: (?=\S)
             set:
               - statements-in-parens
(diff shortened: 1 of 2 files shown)


Thom1729:
I've fixed numbers 1 and 2 above. Hopefully this should be the last of the old `for` loop quirks.

PR #1770, opened 2018-11-20
[#1769] Fix for Clojure code block

drbobbeaty:
There was an issue in that the rendering of the Clojure code block in
a Markdown file would not be rendered properly. This is documented in
Issue #1769 - with screen grabs. This addition to the syntax file
corrects that issue by adding the section for the Clojure code block in
the Markdown file.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -940,7 +940,7 @@ contexts:
       captures:
         0: meta.code-fence.definition.begin.clojure.markdown-gfm
         2: punctuation.definition.raw.code-fence.begin.markdown
-        3: constant.other.language-name.markdown
+        5: constant.other.language-name.markdown
       embed: scope:source.clojure
       embed_scope: markup.raw.code-fence.clojure.markdown-gfm
       escape: '{{code_fence_escape}}'
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -935,6 +935,21 @@ contexts:
     - match: |-
          (?x)
           {{fenced_code_block_start}}
+          ((?i:clojure|clj))
+          \s*$\n?          # any whitespace until EOL
+      captures:
+        0: meta.code-fence.definition.begin.clojure.markdown-gfm
+        2: punctuation.definition.raw.code-fence.begin.markdown
+        3: constant.other.language-name.markdown
+      embed: scope:source.clojure
+      embed_scope: markup.raw.code-fence.clojure.markdown-gfm
+      escape: '{{code_fence_escape}}'
+      escape_captures:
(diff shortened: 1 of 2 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -932,6 +932,21 @@ contexts:
         - match: $
           pop: true
   fenced-code-block:
+    - match: |-
+         (?x)
+          {{fenced_code_block_start}}
+          ((?i:clojure|clj))
+          \s*$\n?          # any whitespace until EOL
```
FichteFoll:
```suggestion
          {{fenced_code_block_trailing_infostring_characters}}
```

Markdown/Markdown.sublime-syntax
```diff
@@ -932,6 +932,21 @@ contexts:
...
+      captures:
+        0: meta.code-fence.definition.begin.clojure.markdown-gfm
+        2: punctuation.definition.raw.code-fence.begin.markdown
+        3: constant.other.language-name.markdown
```
FichteFoll:
```suggestion
        5: constant.other.language-name.markdown
```

OdatNurd:
Indeed! My fix was based on the shipped version of the syntax in the build you mentioned using, which is somewhat different from the version currently in this repo, which has unreleased changes. 😃

drbobbeaty:
You guys are right - I should have... I'm terribly sorry about that.

keith-hall:
Please add some syntax test assertions to ensure the changes you made work properly and won't be regressed in the future.

https://github.com/sublimehq/Packages#pull-requests
> Pull requests should:
> - Have multiple syntax tests

See https://www.sublimetext.com/docs/3/syntax.html#testing for details.
This can be done easily with https://packagecontrol.io/packages/PackageDev.

wbond:
I'm going to hold off on this for now. The Markdown syntax grows in the number of contexts with each syntax we add. We are seeing the sanity limit hit more frequently due to the giant include graph.

drbobbeaty:
@wbond is there something wrong with the PR that can correct this? Or is it _something else_ that is beyond this PR?

wbond:
No, there is nothing wrong with this PR. It is more an issue of us needing to figure out one of:

 - How to try and detect which syntax may be causing the sanity limit
 - Increase the sanity limit some, profiling memory usage to see if something like 50,000 is too high
 - See if there is more significant modifications that can be made to the highlighting engine to not require compiling each external include into a syntax. This is most likely not viable, since the increase of contexts is usually due to `with_prototype` being used, creating new contexts with prepended rules.

deathaxe:
Loading markdown files is already very slow, which makes me doubt an increased sanity limit to be a good solution in manner of performance.

wbond:
Since build 4075 we now have embed lazy loads, so this can be merged in without a performance penalty.

PR #1784, opened 2018-11-30
[HTML] html5 custom element and other tag names

### LINKED ISSUE #1619: [HTML] tag names don't comply with HTML5 specification
According to the html5 specification nearly any character is allowed to be part of a tag name.

see: https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state

The current HTML implementation does not comply with the specification as 

1. it only allows lowercase ASCII letters, numbers and few punctuation `[a-z_][a-z0-9:_]*-[a-z0-9:_-]+`
2. it doesn't respect uppercase (custom-) tag names being converted lowercase implicitly and therefore are valid tag names.
3. it fails to scope custom tag names like `form@custom` correctly.

Examples of valid tag names:

```html
<form@custom attribute="value" />

<FORM-CUSTOM attribute="value" />

```

P.S.: Github does not, too.  ;-)

deathaxe:
Fixes #1619 

1. According to https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state a html5 tag name may contain nearly any character, which is scoped as `entity.name.tag.other` by this commit. A couple of really weird tag names were tested and worked with Firefox.

2. According to https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts custom element tags may contain lower case characters which are not limited to ASCII or digits. Hence this commit uses the `custom_element_char` variable from the CSS.sublime-syntax to match custom-element-tags as it defines the same rule for them.

3. In order to make sure custom/other tag names don't interfere with normal html tags, the `\b` is replaced by `{{tag_name_break}}` in all tag name patterns. This avoids matching tags like `<body·other>` correctly. Without this change body would be matched as normal tag name and `other` as attribute.

4. Step 3 allows the custom-element-tag rule to be moved to the end right next to its illegal-uppercase sister and save some milliseconds on ordinary html code, which makes no use of custom/other tags. The impact is very small (ca. 3-4%) though.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -10,13 +10,15 @@ first_line_match: (?i)<(!DOCTYPE\s*)?html
 scope: text.html.basic
 
 variables:
+  ascii_space: '\t\n\f '
+
   attribute_char: (?:[^ "'>/=\x00-\x1f\x7f-\x9f])
   unquoted_attribute_value: (?:[^\s<>/''"]|/(?!>))+
   not_equals_lookahead: (?=\s*[^\s=])
 
   # https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state
-  tag_name_break: (?=[ \t\f\n/<>])
-  tag_name_char: '[^ \t\f\n/<>]'
+  tag_name_char: '[^{{ascii_space}}/<>]'
+  tag_name_break: (?=[^{{tag_name_char}}])
   tag_name: '[A-Za-z]{{tag_name_char}}*'
 
   block_tag_name: |-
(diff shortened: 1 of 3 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -15,22 +15,24 @@ variables:
   not_equals_lookahead: (?=\s*[^\s=])
 
   # https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state
-  tag_name: '[[:alpha:]_][^/>\s]*'
+  tag_name_break: (?=[\s/>])
+  tag_name_char: '[^\s/>]'
+  tag_name: '[[:alpha:]_]{{tag_name_char}}*'
```
Thom1729: I don't think that underscores are allowed. In addition, I'd spell out :alpha: as `[A-Za-z]` just to make it clear that it's ASCII-only.

HTML/HTML.sublime-syntax
```diff
@@ -15,22 +15,24 @@ variables:
   not_equals_lookahead: (?=\s*[^\s=])
 
   # https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state
-  tag_name: '[[:alpha:]_][^/>\s]*'
+  tag_name_break: (?=[\s/>])
```
Thom1729: I think that `\s` might be too broad. Only a handful of whitespace characters are prohibited: `[\t\n\f />]`.

deathaxe: `\s` does include more?

deathaxe: `unquoted_attribute_value` uses `\s` as well.

Thom1729: Outside Unicode mode, `\s` also includes CR and vertical tab. In Unicode, it includes a lot more.

HTML/HTML.sublime-syntax
deathaxe: This originates from my work at XML. You are right. It's not valid as normal tag name in HTML.

Thom1729:
Is < prohibited? Can't check the spec at.

deathaxe:
I know that from XML specs. They say, each time a `<` is detected a new tag starts. Nevertheless php syntax was broken without it.

deathaxe:
And btw. I find it a better writing experience to not break highlighting due to incomplete tags. I know some syntaxes try to be overeager with `invalid.illegal` scopes to show any wrong syntax, but it hurts while writing code.

keith-hall:
any chance that while we are here, we could change the `\xHH` format in https://github.com/sublimehq/Packages/blob/af3783f93ef59c1a5bc21f3ef0b9148edd1f00b8/HTML/HTML.sublime-syntax#L13 to `\x{00HH}` format to make it consistent with https://github.com/sublimehq/Packages/blob/af3783f93ef59c1a5bc21f3ef0b9148edd1f00b8/HTML/HTML.sublime-syntax#L49-L63 please? (and more compatible with other regex engines - like a fork of Oniguruma called Onigmo) :)

deathaxe:
@keith-hall Agree. Would be more consistent. Not sure whether to pack it into this PR as I just realized some performance issues in the way attributes are handled at all. Some more changes reduced the parsing time for attributes by about 44%!

PR #1785, opened 2018-12-01
[HTML] html5 complient attribute names

deathaxe:
This PR is an extension for PR #1784.

It provides the following changes:

1. Adds [noncharacters](https://infra.spec.whatwg.org/#noncharacter) to the list of characters an attribute must not contain.
2. Renames the variable `attribute_char` to `attribute_name`
3. Changes the hex notation to `\x{0000}` to match the `custom_element_name` style.
4. Replaces the negative lookaheads to improve parsing performance by 35..40%
5. The second commit of this PR replaces the `not_equals_lookahead` in favor of the existing `else-pop` context.

After merging PR #1784 and this one the parsing time for the following file with about 55k lines of code reduced from 1.5s to 0.9s on my box.

```html
<!DOCTYPE html>
<html>
<head>
    <title></title>
</head>
<body>
    <tag-name attribute="test" attribute="test" attribute="test" attribute="test" attribute="test" attribute="test" attribute="test" attribute="test">
    <tag-name attribute="test" attribute="test" attribute="test" attribute="test" attribute="test" attribute="test" attribute="test" attribute="test">
    <!-- 55k lines -->
</body>
</html>
```

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -11,11 +11,7 @@ scope: text.html.basic
 
 variables:
   # https://html.spec.whatwg.org/multipage/syntax.html#attributes-2
-  # 0000-0020   -- code point, C0 control (space)
-  # 007F-009F   -- controls
-  # FDD0-FDEF   -- noncharacters
-  # ?FFFE-?FFFF -- noncharacters
-  attribute_char: '[^''">/=\x{0000}-\x{0020}\x{007F}-\x{009F}\x{FDD0}-\x{FDEF}\x{FFFE}\x{FFFF}\x{1FFFE}\x{1FFFF}\x{2FFFE}\x{2FFFF}\x{3FFFE}\x{3FFFF}\x{4FFFE}\x{4FFFF}\x{5FFFE}\x{5FFFF}\x{6FFFE}\x{6FFFF}\x{7FFFE}\x{7FFFF}\x{8FFFE}\x{8FFFF}\x{9FFFE}\x{9FFFF}\x{AFFFE}\x{AFFFF}\x{BFFFE}\x{BFFFF}\x{CFFFE}\x{CFFFF}\x{DFFFE}\x{DFFFF}\x{EFFFE}\x{EFFFF}\x{FFFFE}\x{FFFFF}\x{10FFFE}\x{10FFFF}]'
+  attribute_char: '[^\t\n\f />=]'
   unquoted_attribute_value: (?:[^\s<>/''"]|/(?!>))+
 
   # https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -15,7 +15,7 @@ variables:
   # 007F-009F   -- controls
   # FDD0-FDEF   -- noncharacters
   # ?FFFE-?FFFF -- noncharacters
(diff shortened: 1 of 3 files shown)


deathaxe:
Hmm. Just see the reason for the lookaheads.

HTML/HTML.sublime-syntax
```diff
@@ -15,7 +15,7 @@ variables:
   # 007F-009F   -- controls
   # FDD0-FDEF   -- noncharacters
   # ?FFFE-?FFFF -- noncharacters
-  attribute_name: '[^''">/=\x{0000}-\x{0020}\x{007F}-\x{009F}\x{FDD0}-\x{FDEF}\x{FFFE}\x{FFFF}\x{1FFFE}\x{1FFFF}\x{2FFFE}\x{2FFFF}\x{3FFFE}\x{3FFFF}\x{4FFFE}\x{4FFFF}\x{5FFFE}\x{5FFFF}\x{6FFFE}\x{6FFFF}\x{7FFFE}\x{7FFFF}\x{8FFFE}\x{8FFFF}\x{9FFFE}\x{9FFFF}\x{AFFFE}\x{AFFFF}\x{BFFFE}\x{BFFFF}\x{CFFFE}\x{CFFFF}\x{DFFFE}\x{DFFFF}\x{EFFFE}\x{EFFFF}\x{FFFFE}\x{FFFFF}\x{10FFFE}\x{10FFFF}]+'
+  attribute_char: '[^''">/=\x{0000}-\x{0020}\x{007F}-\x{009F}\x{FDD0}-\x{FDEF}\x{FFFE}\x{FFFF}\x{1FFFE}\x{1FFFF}\x{2FFFE}\x{2FFFF}\x{3FFFE}\x{3FFFF}\x{4FFFE}\x{4FFFF}\x{5FFFE}\x{5FFFF}\x{6FFFE}\x{6FFFF}\x{7FFFE}\x{7FFFF}\x{8FFFE}\x{8FFFF}\x{9FFFE}\x{9FFFF}\x{AFFFE}\x{AFFFF}\x{BFFFE}\x{BFFFF}\x{CFFFE}\x{CFFFF}\x{DFFFE}\x{DFFFF}\x{EFFFE}\x{EFFFF}\x{FFFFE}\x{FFFFF}\x{10FFFE}\x{10FFFF}]'
```
Thom1729: We don't actually need to handle non-characters because the [parsing algorithm](https://html.spec.whatwg.org/multipage/parsing.html#attribute-name-state) doesn't. I think we can go with `[^\t\n\f />=]`.

HTML/HTML.sublime-syntax
```diff
@@ -10,9 +10,9 @@ first_line_match: (?i)<(!DOCTYPE\s*)?html
 scope: text.html.basic
 
 variables:
-  attribute_char: (?:[^ "'>/=\x00-\x1f\x7f-\x9f])
+  # https://html.spec.whatwg.org/multipage/syntax.html#attributes-2
+  attribute_char: '[^\t\n\f />=]'
   unquoted_attribute_value: (?:[^\s<>/''"]|/(?!>))+
```
deathaxe: @Thom1729 What do you think about `\s` here? I wonder, if it should be replaced by `\t\n\f ` as well.

Thom1729:
I've thought of a possible solution to the PHP issue:

```yaml
tag-generic-attribute-name:
  - meta_scope: entity.other.attribute-name.html
  - match: (?![\t\n\f />])
    pop: true
  - match: '["''<\0]'
    scope: invalid.illegal.attribute-name.html
  - match: '[^\t\n\f />"''<\0]+'
    # no scope
```

This should scope the entire attribute name (as it should be parsed) as entity.other.attribute-name.html, even if it contains invalid characters. The invalid characters will be marked `invalid.illegal`, so they will stand out. The valid characters will be parsed into single tokens, so the hash highlighting will work. And the PHP syntax will have a chance to interrupt before `<`.

The only downside is that an attribute name like `foo<bar` will still be three colors with hash highlighting, but I think that's an acceptable loss.

deathaxe:
Looks like a good strategy for tag names and unquoted values as well. Even helps with performance.

Issue #1749, opened 2018-10-23
[C++] Inconsistent scopes for method calls and definitions


mjendruk:
Currently normal and generic method/function calls and definitions with namespaces are scoped differently. For normal functions, the `entity.name` scope includes the namespace whereas for generic functions it does not.

I would like to make this more consistent but I would like to know what direction I should go. Should the namespace be included or not? In my opinion, it should not.

```c++
// definition

template<typename A, typename B>
void classname<A, B>::methodName() {
/*   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.function */
/*                  ^^ punctuation.accessor */
/*                    ^^^^^^^^^^ entity.name.function */
}

void classname::methodName() {
/*   ^^^^^^^^^^^^^^^^^^^^^^^^^ meta.function */
/*            ^^ punctuation.accessor */
/*   ^^^^^^^^^^^^^^^^^^^^^ entity.name.function */
```
```c++
// call

void func() { 
    namespace::functionName(param);
/*  ^^^^^^^^^^^^^^^^^^^^^^^ variable.function */
    namespace::genericFunctionName<float>(param);
/*             ^^^^^^^^^^^^^^^^^^^ variable.function */
}
```

mjendruk:
Okay I implemented my suggestion and it seems to work nicely. I find it much more consistent and not highlighting the namespace or class name as part of the function identifier makes sense to me.

I fixed a couple of bugs along the way. Please tell me what you think of the changes. I haven’t started adding tests and moving the changes over to the ObjC files as I don’t want to do it twice.

# Function call
**Previously**
![function_call_old](https://user-images.githubusercontent.com/2190123/49228498-c4d5a500-f3eb-11e8-8002-5498d9b4db11.PNG)

**Now**
![function_call_new](https://user-images.githubusercontent.com/2190123/49228532-d9b23880-f3eb-11e8-9336-88261df8f570.PNG)
* Note: constructor calls using curly braces are now highlighted too

# Function Definition
**Previously**
![function_definition_old](https://user-images.githubusercontent.com/2190123/49228553-e898eb00-f3eb-11e8-9039-0ed6f328267f.PNG)

**Now**
![function_definition_new](https://user-images.githubusercontent.com/2190123/49228561-ed5d9f00-f3eb-11e8-8294-9d00a2a8e267.PNG)

FichteFoll:
[...]

PR #1793, opened 2018-12-06
[C++] improve function highlighting (consistency, bugs)

mjendruk:
This PR implements the changes proposed in #1749. Please have a look into the issue for screenshots of the changes.

I haven’t copied the changes over to the ObjC++ syntax file yet, because I wanted to get your feedback first.

One problem I encountered was namespace macros such as `#define namespace greatapp {`. `greatapp {` was recognized as the beginning of a function call that never ends. Since this can happen with most opening curly braces (lines 1968ff. do not always help) and I couldn't think of a clever general solution, I handled this case separately. 

I managed to separate the highlighting of function calls, which excludes namespaces, and the local symbol list by introducing a meta scope called `meta.function.local-symbol.c++`. The exact naming is of course open for discussion.


mjendruk:
Splitting up this PR sounds actually like a really good idea. I will do this.

With regard to your comment, I can reassure you that the Goto-Symbol-in-Project symbols are not altered at all. Only local symbols for method templates are more precise, because they now also include the class name and so on just like normal methods did all along.

Issue #1771, opened 2018-11-26
[JavaScript] Move JSON out of the JavaScript package.


Thom1729:
The JSON syntax definition, tmPreferences, and tests are inside the JavaScript directory. This means that a user can't disable the JavaScript package (e.g. when using a third-party package) without also disabling the core JSON syntax (and vice versa).

JSON had its roots in JavaScript, but it is nowadays used for all sorts of things. I think that it would make sense to promote JSON to its own default package. There is no module named "JSON" in the Package Control channel, so there should be no serious risk of conflict.

Some third-party code could be hardcoding the location of `JSON.sublime-syntax`. If this is a concern, we could leave a stub at that location:

```yaml
%YAML 1.2
---
name: JSON (Compatibility redirect)
scope: ''
hidden: true
contexts:
  main:
    - match: ''
      embed: 'Packages/JSON/JSON.sublime-syntax'
      escape: \B\b # Never
```

FichteFoll:
Also worth thinking about https://github.com/sublimehq/Packages/issues/285 when this is considered.

deathaxe:
All included sublime-syntax files I've seen so far have been used without path. Something like `- include: "blabla.sublime-syntax"`. The only situations I saw paths were in syntax test files.

In case of a stub file, I think `include` would just be enough?

```
%YAML 1.2
---
name: JSON (Compatibility redirect)
scope: ''
hidden: true
contexts:
  main:
    - include: 'Packages/JSON/JSON.sublime-syntax'
```

FichteFoll:
It would still need a base scope, but other than that I suppose it should work.

FichteFoll:
Can be closed since #1805 has been merged.

wbond:
This has been completed

Issue #285, opened 2016-04-14
Separate JSON and "JSON with comments" syntaxes?


FichteFoll:
I've been thinking about this for quite a while now, but it always bothered me that Sublime Text's JSON syntax definition allows JavaScript-style comments, which are a clear violation of the standard. Users hand-editing JSON files with Sublime Text might be confused why whatever is parsing the `.json` file chokes on the comments that ST highlighted perfectly fine.

In order to make things more proper, I suggest introducing a different syntax for Sublime Text configuration files named "JSON with comments" or "JSONC" that will keep the features of the current JSON definition, while the "JSON" syntax gets comment support removed.

I tried googling for a possible standard for this that might have emerged over the time, but there doesn't seem to be any. I found:

- a [`commentjson` python module](https://commentjson.readthedocs.org/en/latest/) (also for [npm as `comment-json`](https://www.npmjs.com/package/comment-json))
- a [npm package named `json5`](http://json5.org/), which makes additional changes to its JSON standard such as allowing single quotes, unquoted object keys, more number formats and trailing commas (same as ST3) however

Opinions?

MattDMo:
Why not tighten the current JSON definition to exclude comments (as it should be), and adapt the [`PackageDev`](https://github.com/SublimeText/PackageDev/blob/master/Syntax%20Definitions) syntax definitions into Sublime's core? This way, *all* users have default highlighting of `.sublime-settings`, `.sublime-keymap`, `.sublime-mousemap`, and all of the other pseudo-JSON files.

FichteFoll:
>Why not tighten the current JSON definition to exclude comments

Yes, that is the plan.

I wouldn't object having specific syntax definitions for ST's configuration in ST's core, but not the current ones in PackageDev. I intend to redo them anyway. Will and Jon will need to decide on that however.

rchl:
I personally wouldn't be so happy about such change as I work with JSON files where comments are allowed. That is with chrome extensions whose manifest.json file accepts comments and also with chromium code itself which uses JSON for many things and uses https://github.com/open-source-parsers/jsoncpp to parse them which allows comments.

Of course if two separate syntaxes would be included in Sublime, then it would be easy to change for the more lax parser to be the default, so it wouldn't be that much problem.

jskinner:
[...]

PR #1805, opened 2018-12-08
[JavaScript] Move JSON out of the JavaScript package

deathaxe:
As suggested in issue #1771 this commit moves the JSON syntax into a dedicated package in order to be able to disable either JavaScript or JSON independently.

This commit does not yet create the suggested compatibility redirection within the JavaScript package.

The decision about it is left to the core devs right now

1. as it would cause a duplicated syntax file
2. a quick search within packagecontrol.io did not reveal any packages depending on the builtin JavaScript/JSON.sublime-syntax
3. the location of the sublime-syntax file should not be too important if it is imported via `scope:source.json` or its name `JSON.sublime-syntax` only. Conflicts would be caused only if it was imported via `Packages/JavaScript/JSON.sublime-syntax` which is easy to fix.


JavaScript/Comments.tmPreferences
```diff
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Comments</string>
 	<key>scope</key>
-	<string>source.js, source.json</string>
+	<string>source.js</string>
```
Thom1729: We might need to re-add comments to the JSON package even though JSON doesn't actually support them. See #285.

Thom1729:
I think you're right about not needing the redirect, and I'm glad you checked the existing packages. In the unlikely event that this goes out in a dev build and breaks something, it would be extremely easy to put in a quick patch before the next build.

JavaScript/Comments.tmPreferences
deathaxe: There is a dedicated `Comments.tmPreferences` in the JSON package, just to avoid cross references. Hence we don't loose anything.

wbond:
I think we are going to take a swing at this as part of the next dev cycle. There will be some changes in the C++ codebase to ensure things don't go wonky for people with open files that reference the old syntax path, hence it will be put "on hold" for now and resolved at release time.

jrappen:
relevant here: sublimehq/Packages@9c44e60

PR #1811, opened 2018-12-20
[Lua] Rewrite syntax

Thom1729:
This is a total rewrite of the Lua syntax.

The previous Lua syntax was `tmLanguage`-derived and fairly rudimentary. It had a single named context and relied heavily on large, complex regexps with many capture groups. As a result, it had trouble handling complex expressions or unexpected whitespace. In addition, it uses many negative lookbehinds to work around the simplistic parsing.

This new syntax definition is written as a `sublime-syntax` from the ground up. It should be capable of providing exact highlighting of any Lua code (with some minor caveats about line lookahead). In addition, it provides standard meta scopes for use by package developers.

The previous syntax used few enough scopes that the new syntax should be generally compatible with it.

I don't use Lua myself, so I'd appreciate feedback and examples from people who do. Some known work still remains:

- Highlighting support functions. (Should I just do the [standard library](https://www.lua.org/manual/5.3/manual.html#6)?)
- Ignoring the shebang. (Not sure how to test this, because it basically only works on the first line of a file and doing it elsewhere would break things.)
- Removing the nested lists in the syntax.
- Improving block meta scoping.


Lua/Lua.sublime-syntax
```diff
@@ -1,87 +1,464 @@
...
+        - match: =
+          scope: punctuation.separator.key-value.lua
+          push: expression
+
+        - match: (?=\S)
+          push: expression
+
+  function-literal:
+    - match: function{{identifier_break}}
+      scope: storage.type.function
```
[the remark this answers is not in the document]
Thom1729: Fixed.

Lua/Lua.sublime-syntax
```diff
@@ -1,87 +1,464 @@
...
+
+    - match: break{{identifier_break}}
+      scope: keyword.control.break.lua
+
+    - match: return{{identifier_break}}
+      scope: keyword.control.return.lua
+      push: expression-list
+
+    - match: local{{identifier_break}}
+      scope: storage.type.lua
```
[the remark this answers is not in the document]
Thom1729: Fixed.

Lua/Lua.sublime-syntax
```diff
@@ -1,87 +1,464 @@
...
+              scope: punctuation.section.brackets.end.lua
+              pop: true
+            - match: (?=\S)
+              push: expression
+
+        - match: '{{identifier}}(?=\s*=)'
+          scope: meta.key.lua
+
+        - match: =
+          scope: punctuation.separator.key-value.lua
```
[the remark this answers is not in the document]
Thom1729: To me, it's no different from a colon in a JavaScript object or a Python dict or the arrow in a Ruby hash. I'd rather highlight it accordingly based on syntactic function than highlight it as an operator just because it's an equal sign.

if we get complaints, we can always change it (or encourage the complainers to update their color schemes).

Lua/Lua.sublime-syntax
```diff
@@ -1,87 +1,464 @@
 %YAML 1.2
 ---
-# http://www.sublimetext.com/docs/3/syntax.html
 name: Lua
```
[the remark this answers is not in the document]
Thom1729: Fixed.

Lua/Symbol List.tmPreferences
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+    <key>name</key>
+    <string>Symbol List Function</string>
+    <key>scope</key>
+    <string>source.lua meta.name.function</string>
```
rwols: Nice. I think this is a good solution.

Lua/Lua.sublime-syntax
```diff
@@ -1,87 +1,480 @@
...
+
+    - match: \d+(?:\.\d*)?([Ee][-+]?\d*)?
+      scope: constant.numeric.decimal.lua
+      pop: true
+
+  string:
+    - match: \'
+      scope: punctuation.definition.string.begin.lua
+      set:
+        - meta_scope: string.quoted.single.lua
```
rwols: The prototype context should probably not be included:

```
local foo = "bar--baz"
print(foo) -- prints "bar--baz"
```

Thom1729: Fixed.

Thom1729:
A variable assigned to a function literal should now be scoped `entity.name.function`.

In `function Account:new(o) end`, only `new` is the name of the function, so Account and the colon should not be scoped `entity.name.function`. This is a deliberate change from the old syntax. However, your point about the symbol list is well taken. I've added an explicit `Symbol List.tmPreferences` that should address this. With this decoupled, we could expand the symbols if desired (e.g. from `Account:new` to `function Account:new(o)`).

I've implemented the [basic builtin functions](https://www.lua.org/manual/5.3/manual.html#6.1), plus the constants `_G` and `_VERSION`. Following the example of JavaScript, we might want to implement all of the builtin modules.

Can you mark as resolved the conversations that you think are addressed?

rwols:
> Can you mark as resolved the conversations that you think are addressed?

I tried looking for that button, but I can't seem to find it. Is it some sort of repository setting that @wbond has to enable?

At this point I'm trying out the syntax on some code bases and it seems to work well. It's basically done from my point of view.

Since the syntax can track wether we are in a "do ... end" block perfectly, and there is no preprocessor of sorts, *and* you cannot re-assign to keywords (e.g. `end = "foo"` is not allowed), it might be safe to check for stray `end` keywords in the `main` context.

> Following the example of JavaScript, we might want to implement all of the builtin modules.

Possibly, but I think this is mergeable for now.

There is one feature that you might want to consider looking deeper into: one may `yield` from a coroutine via the `coroutine` module. With the current syntax, `coroutine.yield("foo")` is scoped as a variable.function. I think that is okay and doable, but you might have other ideas about this. Either way I just wanted you to be aware of this.

Thom1729:
I've resolved the conversations. Not sure why the button wasn't showing up.

Unexpected `end` will now be scoped `invalid.illegal`. This can happen at top level, but also inside a `repeat` loop (which is ended by `until`).

I added support scopes for builtin modules. This should ensure that important builtin features like `coroutine.yield` are marked. Syntactically, that is a function, so I'd rather mark it thus rather than as a keyword.

FichteFoll:
Exciting. I've done some Lua in the past and also made a couple local fixes on the default syntax many years ago and took a couple notes on issues I noticed while working with it.

I'll test and review the new syntax in the next couple days.

Lua/tests/syntax_test_lua.lua
```diff
@@ -0,0 +1,491 @@
...
+--                                ^^^ constant.language
+--                                    ^ punctuation.section.group.end
+    end
+--  ^^^ meta.function meta.block keyword.control.end
+
+    function foo.bar:baz (...) end
+--  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.function
+--           ^^^ variable.other - entity
+--              ^ punctuation.accessor
+--               ^^^ variable.other - entity
```
FichteFoll: This should be a `meta.property`, not `variable.other`.

Works correctly for `foo.abc.bar = function() end;`, although there is no test for it.

Lua/tests/syntax_test_lua.lua
```diff
@@ -0,0 +1,491 @@
...
+
+    foo = function() end;
+--  ^^^ entity.name.function
+
+    foo.bar = function() end;
+--      ^^^ entity.name.function
+
+--STATEMENTS
+
+    end;
```
FichteFoll: Why apply the keyword scope as well? For color schemes that don't set a background color on `invalid.illegal`, this will look like a normal keyword.

Lua/tests/syntax_test_lua.lua
```diff
@@ -0,0 +1,491 @@
...
+    end
+--  ^^^ keyword.control.end
+
+    repeat
+--  ^^^^^^ keyword.control.loop
+        2 + 2;
+
+        end;
+--      ^^^ invalid.illegal.unexpected-end keyword.control.end
+    until true;
```
FichteFoll: Add a test for `true`?

Lua/tests/syntax_test_lua.lua
```diff
@@ -0,0 +1,491 @@
...
+--      ^ variable.other
+--       ^ punctuation.separator.comma
+--         ^ variable.other
+--           ^^ keyword.control.loop
+--              ^ variable.other
+--               ^ punctuation.separator.comma
+--                 ^ variable.other
+--                   ^^^^^^ meta.block
+--                   ^^ keyword.control
+--                      ^^^ meta.block keyword.control.end
```
FichteFoll: (nitpick) `meta.block` is tested twice here.

FichteFoll:
Found another issue: built-in function calls are missing the `meta.function-call` scope.

FichteFoll:
Another thing to consider:

In https://github.com/sublimehq/Packages/issues/421#issuecomment-289176723 I suggest the usage of `string.unquoted.key` for unquoted keys as they are implicitly converted to strings in the resulting table. You haven't commented on that so far and this would be the ideal opportunity to try this in a field test, so to say.

Lua/tests/syntax_test_lua.lua
Thom1729: Fixed.

Lua/tests/syntax_test_lua.lua
Thom1729: Fixed.

Lua/tests/syntax_test_lua.lua
Thom1729: Fixed.

Lua/tests/syntax_test_lua.lua
Thom1729: Fixed.

Thom1729:
> In #421 (comment) I suggest the usage of string.unquoted.key for unquoted keys as they are implicitly converted to strings in the resulting table. You haven't commented on that so far and this would be the ideal opportunity to try this in a field test, so to say. It also applies to Javascript, which you're quite involved with, too.

I agree in principle and in practice. In the particular case of JavaScript, apparently they once tried highlighting unquoted string keys with `string` and people complained. I'd certainly like to implement this if we can find a way around people's objections. (JS Custom currently provides this as an option, and that might be a viable solution for people who are really, really, really attached to the old behavior.)

Lua/Lua.sublime-syntax
```diff
@@ -496,7 +496,7 @@ contexts:
               push: expression
 
         - match: '{{identifier}}(?=\s*=)'
-          scope: meta.key.lua
+          scope: meta.key.lua string.unquoted.key
```
wbond: This should have `.lua` at the end

Thom1729: Done.

Thom1729:
Yeah, the commit history is a mess. Sorry about that. A squash sounds like the way to go.

PR #1813, opened 2018-12-25
[Latex] Add support for \parencites

### LINKED ISSUE #1762: [Latex] Commands with multiple mandatory and optional arguments
The `biblatex` package includes the `\parencites` command that has a somewhat unusual argument structure in that multiple optional and mandatory arguments follow each other, e.g.
````latex
\parencites[Proposition~1]{Firstref}[p.~2]{SecondRef}
````
Such commands are not captured successfully.

Moreover, the `parencites` command is not recognized. This can be fixed by using
````
         (?:
            [cC]ite(?:author|title|year|date|url|t|p|alt|alp|text|yearpar)?
          | (?:[pP]aren|foot|[tT]ext|[sS]mart|super|[aA]|no|full|footfull)cite(?:s)?|footcitetext
          )
````
instead of https://github.com/sublimehq/Packages/blob/289782ff2e4cb58de171579c7fc86fe00d280619/LaTeX/LaTeX.sublime-syntax#L664-L667

### PR DESCRIPTION
**Pull request by tobiasdiez**
Fixes #1762.

The `biblatex` package includes the `\parencites` command that has a somewhat unusual argument structure in that multiple optional and mandatory arguments follow each other, e.g.
````latex
\parencites[Proposition~1]{Firstref}[p.~2]{SecondRef}
````
This PR adds support for this command.

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -663,7 +663,7 @@ contexts:
           (\\)
           (?:
             [cC]ite(?:author|title|year|date|url|t|p|alt|alp|text|yearpar)?
-          | (?:[pP]aren|foot|[tT]ext|[sS]mart|super|[aA]|no|full|footfull)cite|footcitetext
+          | (?:[pP]aren|foot|[tT]ext|[sS]mart|super|[aA]|no|full|footfull)cite(?:s)?|footcitetext
           )
           \*?
         )\b
@@ -672,6 +672,16 @@ contexts:
         2: punctuation.definition.backslash.latex
       push:
         - meta_scope: meta.function.citation.latex
+        - include: general-optional-arguments
+        - match: '\{'
+          scope: punctuation.definition.group.brace.begin.latex
+          push:
+            - meta_scope: meta.function.citation.latex meta.group.brace.latex
+            - match: '[a-zA-Z0-9\.:/*!^_-]+'
+              scope: constant.other.citation.latex
+            - match: '\}'
+              scope: punctuation.definition.group.brace.end.latex
+              pop: true
         - include: general-optional-arguments
         - match: '\{'
           scope: punctuation.definition.group.brace.begin.latex
--- a/LaTeX/syntax_test_latex.tex
+++ b/LaTeX/syntax_test_latex.tex
@@ -104,12 +104,25 @@
 % ^ support.function.cite.latex keyword.other.cite.latex
 %        ^ constant.other.citation
 
-
 \cite[\command]{my:bib:key}
 % ^ meta.function.citation.latex
 % ^ support.function.cite.latex
 %           ^ support.function
 %
+
+\parencite[Propositon~1]{Ref}
+% ^ meta.function.citation.latex
+% ^ support.function.cite.latex
+%          ^ meta.group.bracket.latex
+%                         ^ constant.other.citation.latex
+
+\parencites[Proposition~1]{Firstref}[p.~2]{SecondRef}
+% ^ meta.function.citation.latex
+% ^ support.function.cite.latex
+%           ^ meta.group.bracket.latex
+%                          ^ constant.other.citation.latex
+%                                     ^ meta.group.bracket.latex
+%                                          ^ constant.other.citation.latex
 
 
 % URL COMMAND

### CODE REVIEW

**Comment by wbond** [comment]
Thanks for taking the time to send multiple fixes! Could you add tests for this PR also?

**Comment by tobiasdiez** [comment]
Tests added here as well.

**Comment by wbond** [comment]
@randy3k @FichteFoll Another LaTeX PR I'd be grateful for feedback on.

File: LaTeX/LaTeX.sublime-syntax [hunk 1]
```diff
@@ -673,6 +673,16 @@ contexts:
       push:
         - meta_scope: meta.function.citation.latex
         - include: general-optional-arguments
+        - match: '\{'
```
FichteFoll: This match is just a copy of the following `\{` match. If you just want to include `general-optional-arguments` before matching braces, you should just put that before the empty match pop pattern.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -663,7 +663,7 @@ contexts:
           (\\)
           (?:
             [cC]ite(?:author|title|year|date|url|t|p|alt|alp|text|yearpar)?
-          | (?:[pP]aren|foot|[tT]ext|[sS]mart|super|[aA]|no|full|footfull)cite|footcitetext
+          | (?:[pP]aren|foot|[tT]ext|[sS]mart|super|[aA]|no|full|footfull)cite(?:s)?|footcitetext
```
FichteFoll: This match seems to broad to me. I'm not an expert, but do `\footcites` or `\nosites` exist?

If the `s` ist just for parencizes, add it as a separate alternative. Or, if parencites is the only command that has this weird mix of optional arguments, it might even make more sense to have it in its own match and push its own set of patterns compared to all of these.

cc @randy3k

randy3k:
@FichteFoll 
I remember that I compiled the list of commands from biblatex (or something similar), and I agree that it will be an overkill.

FichteFoll:
After working with parencites last week, I can say for certain that parencites? and cites? exist, but I'm doubtful of the others.

LaTeX/LaTeX.sublime-syntax
tobiasdiez: This changes was made to properly recognize the following command `\parencites[Proposition~1]{Firstref}[p.~2]{SecondRef}`.

tobiasdiez:
There are a few other commands with this somewhat special syntax. According to [3.8.3 Qualified Citation Lists of the biblatex manual](http://mirrors.ibiblio.org/CTAN/macros/latex/exptl/biblatex/doc/biblatex.pdf) these are `cites`, `parencites`, `footcites`, `footcitetexts`, `smartcites`, `supercites` and a few capital letter versions.

How should I proceed with this PR? What changes are required to merge it?

FichteFoll:
I suggest adding a separate match for the commands with the special syntax and only for those.

Furthermore, you should remove the match in line 686 because currently it is 100% redundant and will never match.

LaTeX/LaTeX.sublime-syntax
FichteFoll: Okay, I just noticed that the difference here is that the new version uses `push` instead of `set`. But now the match at line 686 will never match, so that should be removed.

FichteFoll:
In the end, there are two options:

1. Use the `push` instead of `set` in the cite\* commands and allow all of them to have repeating required arguments. This works because of the else-pop match later. However, you should still only add \*s variants for the commands that actually have them and not *all* of them.
2. Copy the match for the complex commands where you match only those and use a `pus` instead of `set`. However, you'll end up with an almost exact copy of the source context.

Either are probably fine, but 1 sounds better to me.

LaTeX/LaTeX.sublime-syntax
tobiasdiez: I might confuse something here, but the first block (with push) matches `Firstref` while the second block (with set) matches `SecondRef`.

FichteFoll: No, this is not how syntax definitions work. You define a set of patterns that get matched at a certain point in the file. This is a context. The patterns are checked in order. If a pattern matches, its directions are executed. If not pattern matches, the position is incremented and the process is repeated. Thus, having two identical match patterns in the same context will always lead to the first pattern winning.

wbond:
Hey @tobiasdiez - is this still something you are keen to have added? If so, I think @FichteFoll has suggestions on the final tweaks needed to get this merged.

FichteFoll:
I applied my suggested changes in #2179 and also added a couple other cite comands missing I noticed while going through the biblatex docs.

tobiasdiez:
Thanks @FichteFoll, I was a bit busy with other things.

PR #1814, opened 2018-12-25
[Perl] Fix angle quoted strings break highlighting

deathaxe:
This PR adds support for angle-bracket quoted strings to the Perl syntax definition.

see: https://docs.perl6.org/language/quoting#Word_quoting:_%3C_%3E

### Example

https://github.com/erlang/otp/blob/master/make/fixup_development_runtime_dependencies

#### without this fix

grafik

#### with this fix

grafik


wbond:
Thanks!

PR #1815, opened 2018-12-28
Avoid tricky YAML syntax in the Go definition

kornelski:
Double `:` confuses my YAML parser.


FichteFoll:
I suggest fixing your parser as this is valid YAML. A whitespace must follow a colon to terminate a plain scalar and turn it into a key.

PR #1818, opened 2019-01-01
[C++/ObjC++] Improve function call syntax

mjendruk:
First part of #1793

__Previously__

__Now__

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -559,7 +559,7 @@ contexts:
             2: variable.function.c++
             3: punctuation.section.group.begin.c++
           set:
-            - meta_scope: meta.constructor-call.c++
+            - meta_scope: meta.function-call.c++
             - meta_content_scope: meta.group.c++
             - match: '\}'
               scope: meta.group.c++ punctuation.section.group.end.c++
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -164,12 +164,13 @@ contexts:
       scope: constant.numeric.c++
 
   identifiers:
-    - match: '{{identifier}}\s*(::)\s*'
+    - match: '(?:(::)\s*)?{{identifier}}\s*(::)\s*'
       captures:
-        1: punctuation.accessor.c++
+        1: punctuation.accessor.double-colon.c++
+        2: punctuation.accessor.double-colon.c++
     - match: '(?:(::)\s*)?{{identifier}}'
       captures:
-        1: punctuation.accessor.c++
+        1: punctuation.accessor.double-colon.c++
 
   function-specifiers:
     - match: \b(const|final|noexcept|override)\b
@@ -525,15 +526,11 @@ contexts:
             - include: template-common
 
   generic-type:
-    - match: '(?=(?!template){{path_lookahead}}\s*{{generic_lookahead}}\s*\()'
+    - match: '(?=(?!template){{path_lookahead}}\s*{{generic_lookahead}}\s*(\(|\{))'
       push:
         - meta_scope: meta.function-call.c++
         - match: \btemplate\b
           scope: storage.type.template.c++
-        - match: '(?:(::)\s*)?{{identifier}}\s*(::)\s*'
-          captures:
-            1: punctuation.accessor.double-colon.c++
-            2: punctuation.accessor.double-colon.c++
         - match: (?:(::)\s*)?({{identifier}})\s*(<)
           captures:
             1: punctuation.accessor.double-colon.c++
@@ -556,6 +553,19 @@ contexts:
               scope: meta.group.c++ punctuation.section.group.end.c++
               pop: true
             - include: expressions
+        - match: (?:(::)\s*)?({{identifier}})\s*(\{)
+          captures:
+            1: punctuation.accessor.double-colon.c++
+            2: variable.function.c++
+            3: punctuation.section.group.begin.c++
+          set:
+            - meta_scope: meta.constructor-call.c++
+            - meta_content_scope: meta.group.c++
+            - match: '\}'
+              scope: meta.group.c++ punctuation.section.group.end.c++
+              pop: true
+            - include: expressions
+        - include: identifiers
         - include: angle-brackets
         - match: '\('
           scope: meta.group.c++ punctuation.section.group.begin.c++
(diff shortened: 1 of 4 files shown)


wbond:
So the one things that I am unsure about with this is the whole semantically treating an initializer list as a function call, aka a direct replacement for a contractor, just with curly braces.

Part of that is treating `{}` as a "group" in some places and "block" in others, but mostly it is that a curly brace and parens really aren't equivalent.

That said, I'd be interested in getting feedback from some other people.

@rwols what do you think? Do either of @deathaxe or @Clamsax have thoughts?

deathaxe:
What I like with this PR is limiting `variable.function` to the real function identifier without the namespace parts. The overall look is more consistent with this PR.

As `meta.group` and `meta.block` are quite low-level like scopes being used for any kind of general parenthesized or curly braced expressions in most syntaxes, I'd be hesitant to mix them up.

It's a question of scoping by semantic meaning vs. common tokens.

Some syntaxes use `meta.function-call.arguments` instead of `meta.group` to express the semantic meaning of argument lists.

The provided screenshot clearly shows how impossible it is to semantically distinguish the function call `std::count` from the class constructors `std::string` and `std::vector`.

Therefore I think, the C/C++ like syntaxes should probably concentrate on highlighting things in a consistent way, rather than trying to apply too many semantic details.

Therefore the scope naming guidelines should be applied, which say `{..}` = `meta.block` and `(..)` = `meta.group`.

rwols:
I've been trying out this branch locally, with the latest changes from the master branch, for the last two days. It works nicely. I think we should merge this.

> The provided screenshot clearly shows how impossible it is to semantically distinguish the function call std::count from the class constructors std::string and std::vector.
> Therefore I think, the C/C++ like syntaxes should probably concentrate on highlighting things in a consistent way, rather than trying to apply too many semantic details.


Agreed.

> Part of that is treating {} as a "group" in some places and "block" in others, but mostly it is that a curly brace and parens really aren't equivalent.

I don't really see the problem. They are in fact equivalent in some cases:

```c++
#include <vector>
#include <initializer_list>

struct Foo {
  int x;
  int y;
  Foo(int x, int y) : x(x), y(y) {}
};

struct Bar {
    std::vector<int> integers;
    Bar(std::initializer_list<int> integers) : integers(integers) {}
};

int main() {
  // four ways to create a Foo.
  // These are all equivalent.
  // They should all receive `meta.function-call`.
  // This demonstrates the uniform initialization syntax.
  Foo foo1{1, 2};
  Foo foo2(1, 2);
  auto foo3 = Foo{1, 2};
  auto foo4 = Foo(1, 2); 
  // This uses the initializer_list constructor.
  Bar bar1{1, 2, 3, 4, 5, 6, 7};
  return 0;
}
```
([live on godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:c%2B%2B,source:'%23include+%3Cvector%3E%0A%23include+%3Cinitializer_list%3E%0A%0Astruct+Foo+%7B%0A++int+x%3B%0A++int+y%3B%0A++Foo(int+x,+int+y)+:+x(x),+y(y)+%7B%7D%0A%7D%3B%0A%0Astruct+Bar+%7B%0A++++std::vector%3Cint%3E+integers%3B%0A++++Bar(std::initializer_list%3Cint%3E+integers)+:+integers(integers)+%7B%7D%0A%7D%3B%0A%0Aint+main()+%7B%0A++//+four+ways+to+create+a+Foo.%0A++//+These+are+all+equivalent.%0A++//+They+should+all+receive+%60meta.function-call%60.%0A++//+This+demonstrates+the+uniform+initialization+syntax.%0A++Foo+foo1%7B1,+2%7D%3B%0A++Foo+foo2(1,+2)%3B%0A++auto+foo3+%3D+Foo%7B1,+2%7D%3B%0A++auto+foo4+%3D+Foo(1,+2)%3B+%0A++//+This+uses+the+initializer_list+constructor.%0A++Bar+bar1%7B1,+2,+3,+4,+5,+6,+7%7D%3B%0A++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang_trunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),lang:c%2B%2B,libs:!(),options:'-O3',source:1),l:'5',n:'0',o:'x86-64+clang+(trunk)+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4))

deathaxe:
The highlighting works fine anyway, so merging is ok.

Semantically `group` seems the better decision here.

I guess the concern is just, that `group` was only used for `(...)` and `block` for `{...}` so far in most/all syntaxes as it is proposed by the official scope naming guidelines.

wbond:
In the past we've had a hard time being able to apply semantically correct block/group when punctuation is effectively interchangeable, and thus we even allow braces/brackets.

Partially this is due to how people writing color schemes will approach things. In general it would seem users would rather symbols retain identical highlighting even upon slightly different usage rather than having symbols change highlighting based on subtly different semantics.

deathaxe:
As the scope naming guidelines don't menstion any details I wonder what the contributors think of what `semantically correct block/group` means. I guess this is the conflict we are faced to here besides the fact C/C++ using the same tokens for several meanings.

_I understand a `block` as a larger portion of code consisting of one or more `statements`.
A `group` is a smaller portion of code consisting of `expressions` or `terms` which may together with other `expressions` form a  `statement`._

Therefore I struggle with a decision, because `block` feels sematically wrong on the one hand and breaking the so far quite clear rules by scoping the curly braces as `group` on the other. 

> would rather symbols retain identical highlighting

That's a very good point, which must be able to achieved by simple color schemes. More detailed colouring based on semantics may be optionally achievable with more detailed sets of rules in a scheme.

_Maybe a statement like that would be fine to read in the docs?_

As semantically correct scoping of function calls and object creation is not possible anyway, semantically correct scoping punctuations makes no sense as well. Therefore I tend to strictly follow the scope naming guidelines.

`(...)` = `group` or `parens`
`{...}` = `block` or `braces`

Maybe `parens` vs. `braces` would be a good alternative here, but may introduce inconsistencies with existing function/method calls?

mjendruk:
> As semantically correct scoping of function calls and object creation is not possible anyway, semantically correct scoping punctuations makes no sense as well. Therefore I tend to strictly follow the scope naming guidelines.
>
> (...) = group or parens
> {...} = block or braces

I don't have a strong opinion about this but I'm slightly in favour of what @deathaxe said, that is keeping the scoping for `{...}` and `(...)` constistent.

wbond:
Alright, I'm gonna roll with this and we'll see how feedback goes.

Thanks for your work on it @mjendruk!

PR #1820, opened 2019-01-02
[Rust] scope support types in generics

### LINKED ISSUE #1817: [Rust] Type highlighting
Some of the highlighting on types is incorrect

Define a variable with an explicit type of a Vec:
```rs
let var1: Vec<String> = vec![];
```
Note that the highlighting is not present on the Vec<String>

Types are only highlighted when a space is left between the `Vec` and the chevrons:
```rs
let var1: Vec <String> = vec![];
```

    Build: 3176
    Operating system and version: Ubuntu 18.04
    Gnome3

keith-hall:
Fixes #1817


wbond:
Thanks!

PR #1823, opened 2019-01-04
[Lisp] allow ">" character in functions

whancock:
lisp allows a lot of symbols for function definitions, and we don't want to include them all, but ">" is a common pattern E.G. string->int


wbond:
Thanks for taking the time to submit this.

Could you add a test to make sure we don't regress this in the future?

wbond:
@whancock Wondering if you are keen to add a test or two for this?

wbond:
I added support for this with some tests in 9306b156ae1ce0bba0524e6263babca76219a1c5

PR #1824, opened 2019-01-05
[ObjC] properties attributes have been extended by nullability and class accessors.

lolgear:


Objective-C/syntax_test_objc.m
```diff
@@ -581,6 +581,9 @@ @interface Person : NSObject
 /* <- keyword.other punctuation.definition.keyword */
 /*^ keyword.other */
 @property (readonly) NSString *lastName;
+/* <- keyword.other punctuation.definition.keyword */
+/*^ keyword.other */
+@property (readonly, nullable, class) NSString *middleName;
 @end
```
wbond: Some assertions should be added here to make sure we are testing this change. Right now I could back out the change to the .sublime-syntax and I'm fairly sure the tests would still all pass.

lolgear:
@wbond 
Could you share tests documentation or test tool?
I would like to run them as cli if it possible.

FichteFoll:
See https://www.sublimetext.com/docs/3/syntax.html#testing for testing docs.

For CLI testing, you best refer to https://github.com/sublimehq/Packages/blob/master/.travis.yml, where the CI configuration is defined. Usually you'd just run the tests from within ST however as a build system.

lolgear:
@FichteFoll 
Wow! It has been the first time I have used Build key in ST :)

Could you help with following tests?

wbond:
Thanks for getting the tests added @lolgear. Looking over it once more, it seems the context was originally designed to only have a single keyword. Now that we match multiple, the `,` should probably be matched and scoped as `punctuation.separator.objc`.

lolgear:
@wbond 
I have added separator, but still get an error.

```
[punctuation.separator.objc] does not match scope [source.objc meta.interface-or-protocol.objc meta.scope.interface.objc meta.property-with-attributes.objc]
```

lolgear:
However, former commit passed tests :)

wbond:
To get the comma assertion to pass, you'll need to add a match rule to match the `,` and give it a scope.

wbond:
I added the comma match rule in 9929169bf4190bc04083e99e649f03c65883cf9e.

Thanks for submitting this!

PR #1825, opened 2019-01-07
[Go] support built-in types and funcs

mitranim:
The syntax now supports all built-in types and functions rather than just `make` and `new`. This feature is purely additive: built-ins receive additional `support` scopes; the user can opt out by not including these scopes into a color scheme.

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -2,10 +2,8 @@
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
 
-
 # The structure and terminology of this syntax reflects the spec:
 # https://golang.org/ref/spec
-
 
 # The following is a simplified model of Sublime Text's syntax engine, reverse-
 # engineered through experience. This model uses assertive language for brevity,
@@ -66,6 +64,12 @@ variables:
   # https://golang.org/ref/spec#Keywords
   # These are the only words that can't be used as identifiers.
   keyword: \b(?:break|case|chan|const|continue|default|defer|else|fallthrough|for|func|go|goto|if|import|interface|map|package|range|return|select|struct|switch|type|var)\b
+
+  # https://golang.org/ref/spec#Predeclared_identifiers
+  predeclared_type: \b(?:bool|byte|complex64|complex128|error|float32|float64|int|int8|int16|int32|int64|rune|string|uint|uint8|uint16|uint32|uint64|uintptr)\b
+
+  # https://golang.org/ref/spec#Predeclared_identifiers
+  predeclared_func: \b(?:append|cap|close|complex|copy|delete|imag|len|make|new|panic|print|println|real|recover)\b
 
   # Note: this matches ALL valid identifiers, including predeclared constants,
   # functions and types.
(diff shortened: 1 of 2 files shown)


Go/Go.sublime-syntax
```diff
@@ -186,25 +190,67 @@ contexts:
...
+    - match: '{{predeclared_type}}(?=(?:{{noise}}\))*{{noise}}\()'
+      scope: variable.function.go support.type.builtin.go
+    - match: '{{predeclared_func}}(?=(?:{{noise}}\))*{{noise}}\()'
+      scope: variable.function.go support.function.builtin.go
+    - match: '{{ident}}(?=(?:{{noise}}\))*{{noise}}\()'
+      scope: variable.function.go
+
+  # See notes on `match-call-or-cast`.
+  pop-call-or-cast:
+    - match: \b(?:make|new)\b(?=(?:{{noise}}\))*{{noise}}\()
```
FichteFoll: `(?=(?:{{noise}}\))*` seems like a weird pattern. (occurs a few times)

mitranim: After double-checking, this seems to be as intended. The idea is to capture an identifier followed by zero or more `)` then by a `(`, ignoring any whitespace and comments in between. I don't see a simpler way.

mitranim: To clarify, this is valid Go:

```go
var _ = (((new)))(int)
```

FichteFoll: Nevermind, I missed a backslash there and though the `*` was applied to the look-ahead, but instead this is still within the look-ahead itself.

Go/Go.sublime-syntax
```diff
@@ -186,25 +190,67 @@ contexts:
...
+  # expect it to be very rare.
+  match-predeclared-constants:
     - match: \b(?:true|false|nil)\b
       scope: constant.language.go
-    - include: match-new
-    - include: match-make
 
-  # Note: calls and casts are syntactically identical. Detecting casts and
-  # scoping them as types is beyond the capabilities of this syntax engine.
+  # Note: in Go, calls and casts are syntactically identical. Detecting casts
```
kortschak: Note that Go does not have casts. Too late for this change though.

mitranim: Go's "type conversions" are a superset of C casts. I don't understand why people insist on Go "not having casts". The term actually means different things in different languages. For example, Java's casts are a superset of Go's "type conversions". I wish Go designers just used the shorter term and chilled out.

mitranim: I guess there's no reason not to update the comments and context names. The syntax does claim to reflect the spec's lexicon.

mitranim: Also, the term is misleading to begin with, IMO. Subjectively, "type conversions" implies that only the type is converted, while the value remains unchanged. This is untrue for numeric conversions (which insert specially generated assembly and/or built-in runtime functions), string and byte conversions (which call built-in runtime functions that allocate and memcpy), and conversions involving interfaces.

Issue #1205, opened 2017-10-08
[*] Add CODEOWNERS file


jrappen:
<https://help.github.com/articles/about-codeowners/#codeowners-file-example>

Possible locations:

* `./CODEOWNERS`
* `./.github/CODEOWNERS` (I suggest here)
* `/docs/CODEOWNERS`

FichteFoll:
Related: https://github.com/sublimehq/Packages/wiki/Contributors-by-Package

jrappen:
Article from the blog introducing the feature:

* <https://github.com/blog/2392-introducing-code-owners>

jrappen:
Closing as stale.

PR #1826, opened 2019-01-07
Add support for Protobuf

gwenzek:
Syntax for Protocol Buffer definition files and Protobuff serialized as text.

Works with proto2 and proto3.
See official Protobuff documentation at https://developers.google.com/protocol-buffers/docs/proto3

There is no official documentation for the text format but it works with the official c++ implementation of the parser.

### DIFF
--- a/Protocol-Buffer/Protocol-Buffer-Text.sublime-syntax
+++ b/Protocol-Buffer/Protocol-Buffer-Text.sublime-syntax
@@ -13,26 +13,20 @@ variables:
   field_name: '\b([A-Za-z][A-Za-z0-9_]*)\b'
   integer: '(?:\+|-)?(?:0|[1-9]\d*)'
   exp: '(?i:e(?:\+|-)?{{integer}})'
-  float: |-
-    (?xi:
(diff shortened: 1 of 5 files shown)


wbond:
I imagine based on your previous contributions that this is a solid syntax. However, I don't think Protobuf is something we will want to maintain in the default packages.

Currently I believe we are intending to add just Swift and JSX.

jrappen:
@wbond add a CODEOWNERS file as described in the help pages and assign @gwenzek as being responsible for that subfolder/language? Just an idea.

jrappen:
- https://help.github.com/articles/about-codeowners/
- https://blog.github.com/2017-07-06-introducing-code-owners/

jrappen:
compare #1205

Thom1729:
There is an existing [Protobuf Syntax Highlighting package](https://github.com/VcamX/protobuf-syntax-highlighting). It has not been updated for a few years. The author may be interested in accepting improvements, or even in handing off the package.

jrappen:
@VcamX FYI

gwenzek:
The syntax has been pushed to https://github.com/VcamX/protobuf-syntax-highlighting, and should be available from Package Control.
Thanks.

PR #1828, opened 2019-01-08
[Go] Fix potential YAML syntax problem

robinst:
The YAML 1.2 parser we use in syntect has a problem handling the unquoted syntax, see https://github.com/trishume/syntect/issues/228

I also tried an [online parser that is based on pyyaml](http://yaml-online-parser.appspot.com/) where it results in this error:

```
while parsing a flow node
expected the node content, but found ':'
  in "<unicode string>", line 2, column 11:
      {match: :=   , scope: keyword.operator.a ... 
```

Ping @keith-hall :)

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -253,7 +253,7 @@ contexts:
     {match: \-   , scope: keyword.operator.go},
     {match: /=   , scope: keyword.operator.assignment.go},
     {match: /    , scope: keyword.operator.go},
-    {match: :=   , scope: keyword.operator.assignment.go},
+    {match: ':=' , scope: keyword.operator.assignment.go},
     {match: <-   , scope: keyword.operator.go},
     {match: <    , scope: keyword.operator.go},
     {match: <<=  , scope: keyword.operator.assignment.go},


wbond:
Seems to be a duplicate of #1815

robinst:
Sorry for the duplicate, I only looked through open issues.

Yeah seems to be indeed allowed by YAML 1.2, from [7.3.3. Plain Style](https://yaml.org/spec/1.2/spec.html#id2788859):

> However, the “:”, “?” and “-” indicators may be used as the first character if followed by a non-space “safe” character, as this causes no ambiguity.

There's already an issue reported against yaml-rust, I might look into trying to fix that: https://github.com/chyh1990/yaml-rust/issues/118

robinst:
FYI yaml-rust is not the only YAML 1.2 implementation that has this problem, e.g. [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) for Python:

```
>>> import ruamel.yaml
>>> from ruamel.yaml import YAML
>>> yaml = YAML()
>>> yaml.load("{x: :=}")
...
ruamel.yaml.parser.ParserError: while parsing a flow node
expected the node content, but found u':'
  in "<byte string>", line 1, column 5:
    {x: :=}
        ^ (line: 1)
```

(I haven't been able to try the [Haskell reference implementation yet](http://hackage.haskell.org/package/YamlReference).)

FichteFoll:
Good catch. I shall report this upstream.

Note that other instances of a plain scalar starting with a colon work:
```
In [6]: yaml.load("x: :=")
Out[6]: ordereddict([('x', ':=')])

In [7]: yaml.load("[:=]")
Out[7]: [':=']
```

For future referece: https://yaml.org/spec/1.2/spec.html#ns-plain-first(c) (also applies to plain scalar values in flow mappings)

Issue #888, opened 2017-04-10
[C] __attribute((section("sectionname"))) prevents text index


jameyhicks:
Using a section attribute for a declaration such as the following:
`struct foo_type foo __attribute((section("sectionname")));`
causes `foo` not to be indexed.

patakdaca24:
Just to mention: **__attribute** macro is not in the C standard, it is a GCC extension.

PR #1831, opened 2019-01-09
[C] Allow declaring a struct pointer in a struct

### LINKED ISSUE #1830: [C] struct field is seen as a declaration inside a struct
Defining a child struct pointer in a struct should not be treated as a struct declaration.

screenshot from 2019-01-09 10-30-47

```C
struct child_t {
	int foo;
};

struct parent_t {
	struct child_t *child_pointer;
//             ^ - entity.name.struct [ This check fails ]
	struct child_t child;
//             ^ - entity.name.struct
	struct nested_t {
//             ^ entity.name.struct
		int bar;
	} nested;
};
```

ismell:
Fixes #1830


rwols (dismissed):

wbond:
Could we add tests for this construct to the other C variants to make sure it get inherited?

ismell:
Done. I need to find some time to give the C syntax some love.

ismell:
Don't merge this yet... I have a couple other fixes for this.

ismell:
So actually fixing struct required a lot more work. I'm pushing up my WIP to get some initial feedback. You can ignore all the `debug.*` scopes. I will remove those for the final PR. I also haven't fixed the enum, or union keywords completely.That's the reason I have the struct-keyword context added everywhere data-structures is checked. This will eventually be cleaned up once I get around to re-writing union and enum. I want to do that work in another PR thought.

Feature Request, It would be cool if I could enable a syntax developer mode where the context that applied a scope could be determined. Then I could avoid adding all the `debug.* scopes manually.

I was also wondering why the Monokai theme doesn't highlight function arguments?

```
        {
            "name": "Function argument",
            "scope": "variable.parameter - (source.c | source.c++ | source.objc | source.objc++)",
            "foreground": "var(orange)",
            "font_style": "italic"
        },
```

I actually almost got them scoped correctly now, so it would be nice to see them highlighted by default.

ismell:
I fixed typedef used with struct declarations and function pointers. Let me know if I'm missing any other cases. I might start cleaning up `union` next.

ismell:
So I simplified my changes. This reduces the number of top level struct scopes to 1 instead of the three I had. I also fixed `union` and `enum`. I replaced the existing function/variable declaration mechanism with the one I had written for the struct body. This now correctly tags multiple variable declarations and multiple function prototype declarations.

There are still some failing cases that I need to work through. Thought I wonder how legit they are.

Expressions in a function declaration?
```
static string foo(bar() + ';');
```

Macro invocation return types. Should figure out a way to support this.
```
MACRO_CALL(int) macro_prefixed_func(){}
```

Function without the `(` on the same line. I'm not sure if I want to support this one because it would make things have to be declared as a function by default instead of a variable. I would rather not pollute the function namespace if it isn't actually a function.
```
MACRO int
funcname2
()
{
}
```

Still a WIP, but if anyone wants to test it that would be appreciated :)

ismell:
Looks like the preprocessor-practical-workarounds are causing issues. I don't think they are needed anymore. I'll play around with removing them.

ismell:
So I fixed the highlighting for macro function calls and macros as types. I think it's looking pretty good except for the couple of edge cases I posted above.

FichteFoll:
Note that there are some test failures: https://travis-ci.org/sublimehq/Packages/builds/518526136#L486

ismell:
Yeah, there is a test case I'm hesitant to support :

    static string foo(bar() + ';');

Having a `+` and string in function parameters fells wrong. I would rather they not be highlighted. Though I can work on supporting it if others agree we should.

I'll update the OP.

ismell:
OP updated and I added some images of the current state. It still doesn't confirm to #1861 yet. I'll work on that next.

+ @deathaxe who might be interested.

FichteFoll:
Thanks for updating, especially the screenshots.

The problem with failing tests is that they cannot be merged, since they would trip up CI everywhere. It may also be that you were waiting for wbond to comment on it, in which case this is of course fine, but you never mentioned this anywhere.

ismell:
So I just removed the test for the invalid syntax. All tests are passing now. The PR still needs to be cleaned up, but I think the highlighting is looking pretty good. I wish I could do a multiline match to clean up some of the stragglers, but they are pretty rare.

I still need to add the Definitions and References files so all the symbols are indexed.

ismell:
So the latest patches have variable indexing enabled.

I have the 4.14 linux kernel open:
```
index "v4.14..." collated in 8.24s from 62847 files
index "v4.14..." is using 230644344 bytes for 2556328 symbols across 26865760 locations
```

The editor does lock up every once in a while. I think its the symbol popup.

```
PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                     
108757 rrangel   20   0 8010940   1.7g 681904 R 100.0   0.9   0:57.83 sublime_text     
```

Maybe this could use some optimization, or just limit how many references are shown:

```
on_hover:
    SublimeLinter.highlight_view: 0.002s total, mean: 0.000s, max: 0.001s
    Default.symbol: 24.466s total, mean: 1.359s, max: 20.100s
```

C++/C.sublime-syntax
```diff
@@ -884,6 +884,9 @@ contexts:
         - meta_scope: debug.type-declaration.union
         - include: union-keyword
         - include: or_pop
+    - match: (?={{identifier}};)
+      pop: true
+      comment: Used for typedefing things defined in types.
```
rwols: Why do you choose to put this as an actual YAML key/value instead of a comment?

ismell: I don't actually have a good reason for that :)

ismell:
So I stopped indexing variables and member references less than 5 characters. This has cut down on the number of useless symbols.

index "v4.14..." is using 191369848 bytes for 2605987 symbols across 21084039 locations

ismell:
Fixed the build breakage. I think this is ready for review. I'm going to leave the debug scopes for now. I can remove them once reviewed. I want to keep them around for now so it's easy to fix any bugs.

C++/C.sublime-syntax
```diff
@@ -122,12 +128,45 @@ contexts:
...
       scope: storage.modifier.c
 
+  modifiers-macros:
+    - match: |-
+        (?x:
+          (?:{{macro_identifier}}|{{attr_macro_identifier}})
+          (?=
+            \(
+            \s*
+            (?>
```
deathaxe: Atomic groups are not supported by ST's sregex engine and cause the slower Oniguruma to be triggered. Is there a certain reason to use them?

ismell:
And here I thought I was making it faster... I can remove it.

On Fri, May 3, 2019, at 12:36 PM, deathaxe wrote:
> *@deathaxe* commented on this pull request.

> 
> In C++/C.sublime-syntax <https://github.com/sublimehq/Packages/pull/1831#discussion_r280884638>:

> >  
   modifiers:
     - match: \b({{modifiers}})\b
       scope: storage.modifier.c
 
+  modifiers-macros:
+    - match: |-
+        (?x:
+          (?:{{macro_identifier}}|{{attr_macro_identifier}})
+          (?=
+            \(
+            \s*
+            (?>

deathaxe:
Is issue https://github.com/sublimehq/Packages/issues/888 coverd by this PR?

ismell:
Yep, it fixes that bug as well.

C++/C.sublime-syntax
ismell: And here I thought I was making things faster. I'll remove them.

ismell:
Any other comments? I've been pretty happy with the syntax :)

C++/syntax_test_c.c
```diff
@@ -247,13 +285,329 @@ if (4) {
...
+/*                                ^^^^^^ storage.type.c */
+/*                                       ^^^^^^ support.type.c */
+/*                                              ^ storage.modifier.c */
+/*                                               ^ variable.parameter.c */
+/*                                                ^ punctuation.separator */
+/*                                                  ^^^^^^ storage.type.c */
+/*                                                         ^^^^^^^^ support.type.c */
+/*                                                                  ^ storage.modifier.c */
+
+typedef struct Node* (*get_node_t)(struct device *d, struct xhci_hcd *);
```
michaelblyons: I like the `storage.modifier` here. Is there precedent for the scope specifically, or did you pick it?

C++/syntax_test_c.c
```diff
@@ -247,13 +285,329 @@ if (4) {
...
+/*             ^^^^^^^^ entity.name.struct.c */
+/*                      ^ meta.struct.body.c punctuation.section.block.begin.c */
+    int data1;
+/*  ^^^ storage.type.c */
+/*      ^^^^^ entity.name.member.indexed.c */
+} newtype;
+/* <- meta.struct.body.c punctuation.section.block.end.c */
+/*^^^^^^^ entity.name.type.typedef.c */
+
+typedef char arrType[6];
```
michaelblyons: I am somewhat less excited about `storage.modifier` _here,_ though. I understand why you did that, but it looks odd to me, possibly because I'm conditioned to see brackets as punctuation.

C++/syntax_test_c.c
```diff
@@ -321,40 +962,813 @@ struct UI_MenuBoxData
...
+/*             ^^^^^^^^^^^^ meta.function-call.parameters.c */
+/*              ^ keyword.operator.c */
+/*               ^^^ variable.other.c */
+/*                  ^^ punctuation.accessor.c */
+/*                    ^^^^ variable.other.member.c */
+    {}
+/*  ^^ meta.function.c meta.block.c */
+
+
+void f_attributes(int *bar) __attribute__((context(&foo,1,1))) {}
```
michaelblyons: Is `keyword.operator` the right scope for `&`?

C++/syntax_test_c.c
FichteFoll: Seems unconventional indeed (for readers: `[``]` have `storage.modifier` here.

ismell: I went with `storage.modifier` because it changes the datatype of the variable.

Reading the scope naming:

`storage.modifer`
> Keywords that affect the storage of a variable, function or data structure should use the following scope. 

`meta.brackets`
> Sections of code delineated by square brackets should use the following scope. The `[` and `]` characters should additionally use the punctuation scopes.

It just felt like `storage.modifer` applied for variable declarations. For variable access I don't scope them with `storage.modifer`.

C++/syntax_test_c.c
ismell: Which instance specifically? The prototype name? I guess I picked that one. It's not really modifying the data type, but it's required to declare it. So maybe it shouldn't be `storage.modifer`?

C++/syntax_test_c.c
ismell: `keyword.operator`
> Operators are typically symbols, so the term keyword can seem somewhat contradictory. Specific variants are sometimes referenced based on the type of operator.

`&` is just a unary operator on the variable, so I think it fits.

C++/C.sublime-syntax
```diff
@@ -1118,13 +1454,13 @@ contexts:
 
   preprocessor-convention-ignore-uppercase-ident-lines:
     - match: ^(\s*{{macro_identifier}})+\s*$
-      scope: meta.assumed-macro.c
+      scope: meta.assumed-macro.c debug.preprocessor-convention-ignore-uppercase-ident-lines
```
wbond: Remove debug scope

C++/C.sublime-syntax
```diff
@@ -1118,13 +1454,13 @@ contexts:
...
       push:
         # It's possible that we are dealing with a function return type on its own line, and the
         # name of the function is on the subsequent line.
         - match: \s*({{identifier}})(?=\s*\()
           captures:
-            1: meta.function.c entity.name.function.c
+            1: meta.function.c entity.name.function.c debug.preprocessor-convention-ignore-uppercase-ident-lines.params
```
wbond: Remove debug scope

C++/C.sublime-syntax
```diff
@@ -1147,12 +1483,14 @@ contexts:
       captures:
         1: keyword.control.import.c
       push:
-        - meta_scope: meta.preprocessor.c
+        - meta_scope: meta.preprocessor.c debug.preprocessor-other
```
wbond: Remove debug scope

C++/C.sublime-syntax
```diff
@@ -1133,10 +1469,10 @@ contexts:
   preprocessor-convention-ignore-uppercase-calls-without-semicolon:
     - match: ^\s*({{macro_identifier}})\s*(\()(?=[^)]*\)\s*$)
       captures:
-        1: variable.function.assumed-macro.c
+        1: variable.function.assumed-macro.c debug.preprocessor-convention-ignore-uppercase-calls-without-semicolon
```
wbond: Remove debug scope

C++/C.sublime-syntax
```diff
@@ -1133,10 +1469,10 @@ contexts:
...
         2: punctuation.section.group.begin.c
       push:
-        - meta_scope: meta.assumed-macro.c
+        - meta_scope: meta.assumed-macro.c debug.preprocessor-convention-ignore-uppercase-calls-without-semicolon
```
wbond: Remove debug scope

C++/Symbol List.tmPreferences
```diff
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Symbol List</string>
 	<key>scope</key>
-	<string><![CDATA[(source.c | source.c++ | source.objc | source.objc++) & (entity.name.class | entity.name.struct | entity.name.union | entity.name.enum)]]></string>
+	<string><![CDATA[(source.c | source.c++ | source.objc | source.objc++) & (entity.name.class | entity.name.struct | entity.name.union | entity.name.enum | entity.name.variable.indexed | entity.name.member.indexed)]]></string>
```
wbond: We should be able to remove these change if we use the standard scopes in the definition

C++/Indexed Reference List.tmPreferences
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>scope</key>
+    <string>
+        source.c constant.other.macro.c,
+        source.c support.type.c,
```
wbond: What will `support.type` handle here?

C++/Indexed Reference List.tmPreferences
```diff
@@ -0,0 +1,18 @@
...
+        source.c variable.other.indexed.c,
+        source.c variable.other.member.indexed.c
```
wbond: These two `.indexed.` scopes can be removed if we use the standard scope in the definition

C++/C.sublime-syntax
```diff
@@ -429,125 +482,312 @@ contexts:
...
+      scope: storage.type.c
+      push:
+        - meta_scope: meta.struct.c debug.struct-keyword
+        - include: data-structures-definition-common-begin
+        - match: '{{identifier}}(?=\s*{|\s*\n)'
+          scope: entity.name.struct.c
+          comment: Struct definition
+          set: struct-body
+        - match: '{{identifier}}'
+          scope: support.type.c
```
wbond: `support.type` is usually used for built-in or framework identifiers. Is this intended to match macro calls?

C++/C.sublime-syntax
```diff
@@ -429,125 +482,312 @@ contexts:
...
+      scope: storage.type.c
+      push:
+        - meta_scope: meta.enum.c debug.enum-keyword
+        - include: data-structures-definition-common-begin
+        - match: '{{identifier}}(?=\s*{|\s*\n)'
+          scope: entity.name.enum.c
+          comment: Struct definition
+          set: enum-body
+        - match: '{{identifier}}'
+          scope: support.type.c
```
wbond: Like the other instances, is this the appropriate scope here?

C++/C.sublime-syntax
```diff
@@ -429,125 +482,312 @@ contexts:
...
+      scope: storage.type.c
+      push:
+        - meta_scope: meta.union.c debug.union-keyword
+        - include: data-structures-definition-common-begin
+        - match: '{{identifier}}(?=\s*{|\s*\n)'
+          scope: entity.name.union.c
+          comment: Struct definition
+          set: union-body
+        - match: '{{identifier}}'
+          scope: support.type.c
```
wbond: Like previously, is this the appropriate scope here?

wbond:
The other thing related to this will be porting these changes to the Obj-C, C++ and Obj-C++ syntaxes. Not saying you have to do it, but we will end up wanting that done.

ismell:
Thanks for the through review. I'l try to get to all the comments next week.

As for `support.type`, I thought that's what I scoped user content as?

> While also used for base frameworks, many syntaxes apply these to scopes unrecognized classes and types, effectively scoping all user constructs.

I can remove `storage.modifier` on variable declarations.

As for the `indexed` scopes I used. I wanted to index the struct members so `Goto Definition` was more usable. The problem was I needed to exclude very common names. e.g., anything less than 4 characters. Ideally it would be nice if Sublime Text had 2 indexes, one for types, and one for members. So Goto Definition on a type would lookup in the type index, and Goto Definition on a member access (e.g., `left_pos` in `foo->left_pos`) would lookup in the member index. I would still need to filter out really short names as to not make it overly noisy. I've been using this syntax with the indexing on the linux kernel for a few months now and it's made navigating around soo much easier. Sublime Text does struggle when hovering over a symbol that is pretty common. It usually locks the UI thread for a few seconds. So that might be something that needs to be addressed before this syntax is merged.

C++/C.sublime-syntax
```diff
@@ -429,125 +482,312 @@ contexts:
...
+          scope: punctuation.section.parens.end
+          set:
+            - include: function-definition-params-group
+            - include: or_pop
+        - match: \*
+          scope: storage.modifier.c
+        - match: '{{identifier}}'
+          scope: entity.name.variable.indexed.c debug.variable-identifier-list.function-pointer
+    - match: '{{identifier}}'
+      scope: entity.name.variable.indexed.c debug.variable-identifier-list
```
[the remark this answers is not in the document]
ismell: This is a global variable declaration. So I used entity.name since `variable.` is for scoping inside expressions/code blocks.

C++/syntax_test_c.c
ismell: Ah, you are referring to the `*`.
I picked it because of description:
> Keywords that affect the storage of a variable, function or data structure should use the following scope.

C++/C.sublime-syntax
ismell: What is the scope I should use for referencing user defined types? In the java syntax I used `support.class`

ismell:
@wbond I've addressed most of the issues. I posted some questions if you could take a look at those.
Once you are happy, I can rebase onto master to fix the merge conflict. Unless you want me to do that now. Not sure if Github freaks out and loses all the comments though. I can also merge in upstream/master, but not sure if you like that either.

deathaxe:
It would propably add all the commits to the history again. I used to merge the latest master into the PR branch, which seems to be the most secure solution.

ismell:
I filed a bug to [Implement Goto Definition Filters](https://github.com/SublimeTextIssues/Core/issues/3100) to address my comment in https://github.com/sublimehq/Packages/pull/1831#issuecomment-544152170

ismell:
Merged in upstream/master. Should be ready for review again.

C++/Indexed Reference List.tmPreferences
ismell: It handles all the references to library/user defined types.

C++/C.sublime-syntax
ismell: IMO yes, there is no real way to tell the difference between library/framework and user defined types.

C++/Indexed Reference List.tmPreferences
michaelblyons: Why not just drop the suffices altogether? You already have `source.c`; no need to specify the language scope at the end, too.

```
source.c variable.other
```

If there's some particular sub-scope that is unintentionally indexed, you can throw in an exception

```
source.c variable.other - variable.other.oops
```

Or even add [an extra file](https://github.com/braver/TypeScriptSyntax/pull/4/files#diff-5f4da57f87ace1671e461f40c7bbae7f) with the banned ones (e.g. `source.c variable.other.oops`) and `showInIndexedReferenceList` set to `0`.

jrappen:
@ismell maybe you could:

* provide some numbers regarding performance changes using `Tools > Build with ... > Syntax Tests - Performance`?
* re-merge upstream/master to re-run tests with the Build 4112 test binary

michaelblyons:
Oh no! Merge conflicts. 😞

ismell:
Yeah, I stopped working on this since Will told me he wanted to fix `C++` and `ObjC` as well instead of just `C`. FWIW, I've been using this for 2 years now doing `linux kernel`, `coreboot`, and other firmware projects and it's pretty solid. If there is actual interest in merging this PR I can make some time to do some minor clean up.

jrappen:
That would be appreciated. Will isn't working for SublimeHQ anymore. The remaining team seems to be focused on other stuff (currently) and has assigned some collaborators to take care of syntaxes (for them). So this would be useful until somebody tackles a c languages rewrite using inheritance.

ismell:
Oops, I should have tested the cpp syntax as well. I'll push a fix.

ismell:
All tests are passing.

ismell:
Can one of the new reviewers take a look?

ismell:
Any hopes of getting this merged in? It will unblock others from improving upon it.

deathaxe (commented):
Maybe not enough package maintainers use C/C++ to feel a value in reviewing. There are other C/C++ related PRs pending.

Another reason might be the long pending idea of fully rewriting C family syntaxes from scratch to reduce duplication by ST4's inheritance feature and to fix all the open issues. The content of this PR would very likely be subject of heavy changes as well then.

I can't find major flaws with a shallow view at the sources, but don't use the syntaxes enough to judge the changes.

C++/C.sublime-syntax
ismell: Looking at the tests might make it more clear.

```
struct foo { <-- This is an entity
 ...
};
```

vs

```
void func(struct bar *ptr); <-- This is a `support.type.c`
```

I used the term declaration because I'm declaring the type of a variable. Open to suggestions.

It sounds like we have new scope keywords though. I can get those updated.

ismell:
> Maybe not enough package maintainers use C/C++ to feel a value in reviewing. There are other C/C++ related PRs pending.

Right, there has been a lot of duplicated effort since this PR has sat for so long.

> Another reason might be the long pending idea of fully rewriting C family syntaxes from scratch to reduce duplication by ST4's inheritance feature and to fix all the open issues. The content of this PR would very likely be subject of heavy changes as well then.

I agree, it could use some modernization, but getting all the test cases landed would be a huge benefit to aiding in the refactor.


Side note, I have no idea how to use Github's PR interface...

deathaxe:
Merging requires approval from at least 2 authorized contributors.

PR #1835, opened 2019-01-12
[Bash] Add .zlogin, .zlogout, .zprofile, .zshenv extension

### LINKED ISSUE #1834: Consider zprofile file as Shell files by default
Originally requested over here https://github.com/sharkdp/bat/issues/465 but then forwarded to this place.

Is it possible to consider `.zprofile` file by default as Shell files so we can get proper syntax highlighting?

deathaxe:
Closes #1834

According to https://unix.stackexchange.com/questions/71253/what-should-shouldnt-go-in-zshenv-zshrc-zlogin-zprofile-zlogout all the mentioned file extensions belong to the zsh ecosystem as well es the already supported .zshrc file extension. So we could safely support them.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -23,6 +23,10 @@ file_extensions:
   - .bashrc
   - .profile
   - .textmate_init
+  - .zlogin
+  - .zlogout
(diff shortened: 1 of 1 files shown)


jrappen:
supersedes #394

PR #1846, opened 2019-01-20
[Rust] Refine operator matches (for ligatures)

FichteFoll:
Most operators were matched character-wise which breaks ligatures. Also
add more refined scope names where applicable.


FichteFoll:
Yeah I only kept the push where it was used previously (or at least for that kind of operator). I didn't pay closer attention to why this is used on some occasions only. Since Rust is an expression-based language, any expression is a valid statement anyway. Maybe something to check out later.

wbond:
Thanks!

PR #1848, opened 2019-01-20
[Scala] Arrow scoping unification

djspiewak:
This is a slightly bigger change than I normally do, at least in terms of concept. Most of my PRs are effectively bug fixes, or just refining functionality that already exists. Here I'm *changing* the scope of an existing and very common construct, and I want to be quite clear that this is intentional and *why* I think this is a good idea. Flagging specifically so you can consider this as such, @wbond.

Here's what I'm changing:

```scala
foo match {
  case bar => ???
//         ^^ storage.type.function.arrow.case.scala
}
```

The assertion above is the *new* scope. It was previously `keyword.operator.arrow.scala`. While there's nothing particularly *wrong* with the `keyword` scope here, it is inconsistent. The following is valid:

```scala
either.fold(
  x => 42,
  { case y => 24 })
```

Both the second and third lines represent lambda expressions. In the second line, the `=>` is scoped the same as it is in the JavaScript mode, which is `storage.type.function.arrow`. The third line, prior to my change, scoped the `=>` as `keyword.operator`. These are semantically almost identical constructs, and it really seemed a bit strange to me when I saw them side by side like this and realized that they were inconsistent.

To that end, I've rescoped the lambda arrows a bit and rescoped the `case` arrows a *lot*. Adding assertions to the above for documentation:

```scala
either.fold(
  x => 42,
//  ^^ storage.type.function.arrow.lambda.scala
  { case y => 24 })
//         ^^ storage.type.function.arrow.case.scala
```

People who really want the old coloring can do so by modifying their color scheme, since they still have distinct scopes, but *by default* they will be styled uniformly. I think this is better all around.

Oh, I also threw in a bug fix for good measure, since lambda lookahead was broken with comments that weren't separate by whitespace.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -197,7 +197,7 @@ contexts:
 
   lambdas:
     # lambda lookahead
-    - match: '(?=({{idorunder}}|{{idorunder}}\s*:(\s*{{id}}\s*{{withinbrackets}}?(\s*[\.#]\s*{{id}}\s*{{withinbrackets}}?)*)+|{{idorunder}}\s*:\s*{{withinparens}}|{{withinparens}})\s*(?:{{rightarrow}})[[[:alpha:]]0-9\s\)\]\}])'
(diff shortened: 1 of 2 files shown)


djspiewak:
@FichteFoll That's basically accurate. There are two different ways you can interpret the `case` block construct: either as a super-charged `switch` block, or as a lambda that can deconstruct parameters. Both are true, really. There is a syntactic difference which correlates with the two different interpretations (the `match` keyword), but I think it makes more sense to keep things consistent, and the *more* consistent interpretation is as a lambda.

wbond:
Thanks for the explanation. I do think a consistent scope makes sense here.

PR #1849, opened 2019-01-21
[SQL] handle schemas when scoping `create table` etc., for a more accurate symbol list

### LINKED ISSUE #1844: [SQL]
Hi,

it would be nice, if full quallified table names would be supported:

```
create table test( id serial );
create table some_schema.test2( id serial );

select * from test;
select * from some_schema.test2;
```

for the first select standing with the cursor on test the `goto definition` works as expeceted, but not with the second select. it does just nothing.

thank you!

keith-hall:
Fixes #1844

### DIFF
--- a/SQL/SQL.sublime-syntax
+++ b/SQL/SQL.sublime-syntax
@@ -9,15 +9,16 @@ scope: source.sql
 contexts:
   main:
     - include: comments
-    - match: '(?i:\s*\b(create(?:\s+or\s+replace)?)\s+(aggregate|conversion|database|domain|function|group|(?:unique\s+)?index|language|operator class|operator|procedure|rule|schema|sequence|table|tablespace|trigger|type|user|view)\s+)(?:(\w+)|''(\w+)''|"(\w+)"|`(\w+)`)'
+    - match: |-
+        (?xi)
+        \b(create(?:\s+or\s+replace)?)\s+
+        (aggregate|conversion|database|domain|function|group|(?:unique\s+)?index|language|operator class|operator|procedure|rule|schema|sequence|table(?:space)?|trigger|type|user|view)
+        \b\s*
       scope: meta.create.sql
       captures:
         1: keyword.other.create.sql
         2: keyword.other.sql
-        3: entity.name.function.sql
-        4: entity.name.function.sql
-        5: entity.name.function.sql
-        6: entity.name.function.sql
+      push: identifier_create
     - match: (?i:\s*\b(drop)\s+(aggregate|conversion|database|domain|function|group|index|language|operator class|operator|procedure|rule|schema|sequence|table|tablespace|trigger|type|user|view))
       scope: meta.drop.sql
       captures:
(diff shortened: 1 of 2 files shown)


SQL/SQL.sublime-syntax
```diff
@@ -222,3 +223,14 @@ contexts:
             0: punctuation.definition.string.end.sql
           pop: true
         - include: string_interpolation
+  identifier_create:
+    - meta_content_scope: entity.name.function.sql
+    - match: \w+
+    - match: \'[^']\'
+    - match: \"[^"]\"
+    - match: \`[^`]\`
```
FichteFoll: The source had ``''(\w+)''|"(\w+)"|`(\w+)` `` for these which would require a `+` after the sets.

A test case for at least one of these seems appropriate.

keith-hall: well spotted, thanks. I have updated the match patterns and included some extra test cases.

verfriemelt-dot-org:
hi, thanks for your effort, but the given example 

```
create table test( id serial );
create table some_schema.test2( id serial );

select * from test;
select * from some_schema.test2;
```

is not working with this PR?

i linked this package with your PR version into my `~/.config/sublime-text-3/Packages` as described by the repo

keith-hall:
well the syntax tests are passing in CI (and locally for me), so presumably something is wrong with your setup.
image

verfriemelt-dot-org:
ah, okay, i found the issue.

if youre standing in 'test' with your cursor and use the goto defintion shortcut, you will be taken to the thing you expect.

but then you standing somewhere with the cursor in `some_schema.test2` it takes only the part before or after the dot. so test or some_schema - but not the whole thing.

if you highlight `some_schema.test2` you get what you'll expect.

keith-hall:
You could choose to have some syntax specific preferences for SQL, and override the `word_separators` setting to remove the dot character, but it would affect double clicking to select a word etc. not just goto definition for no selection.

SQL/syntax_test_sql.sql
```diff
@@ -9,6 +9,29 @@ SELECT "My "" Crazy Column Name" FROM my_table;
...
+--^^^^ meta.create keyword.other.create
+--     ^^^^^ meta.create keyword.other
+--           ^^^^^^^^^^^^^^^^^ entity.name.function
+--                      ^ punctuation.accessor.dot
+--                            ^^^^^^^^^^^^^^^ - entity.name.function
+
+create table "testing123" (id integer);
+--^^^^ meta.create keyword.other.create
+--     ^^^^^ meta.create keyword.other
+--           ^^^^^^^^^^^^ entity.name.function
```
wbond: Should the quotes get `entity.name`? It seems to me they shouldn't since they aren't part of the identifier, but rather just allow non-word chars.

wbond:
@keith-hall What do you think about removing the quotes from the `entity.name` scope?

keith-hall:
I agree that it makes sense to remove the quotes, as it isn't part of the name/identifier. I don't think I will get chance to work on this any time soon though, so if someone wants to take over, they would be most welcome!
I think https://github.com/sublimehq/Packages/pull/2005 is also trying to achieve something similar to this PR.

wbond:
I tried to push to the origin of the PR, and was rejected, so I'll add my fix on top on master

Issue #757, opened 2016-12-23
Syntax Development Tips/Advice


wbond:
If you've spent some time writing syntaxes, take a moment here and share any revelations you've had, or tips on this to test or look for.

wbond:
## Check for Scope Doubling

The characters `(`, `)`, `{`, `}`, `[`, `]` are very easy to double scopes on via `meta_scope`. I try to add tests for `^ punctuation` and then also `^ - punctuation punctuation` to ensure they aren't there.

djspiewak:
# Stateful Chaining

## Don't Over-Use

`sublime-syntax` makes it very easy to have complex interlocking chains of stateful contexts which transition from one to the other via `set`.  This is sometimes necessary to achieve the desired scoping, but it's also very easy to get lost in the mire.  Always convince yourself that you *need* this feature before you use it.

## Push Your First State

While it is absolutely possible to have a `match` in `main` which `set`s into a chain of stateful contexts, and subsequently `set`s back into `main` at the end, it is *not* recommended.  `main` should be a stateless "baseline" context that is always the last element on the stack.  Instead, have your `match` in `main` use `push` to get into your first state, then `pop` out of the last state.  For example, imagine we wanted to match the sequence `abc` with each character scoped differently, and only when they follow each other.  For illustration purposes, we will also match numerics in `main`:

```yaml
contexts:
  main:
    - match: a
      scope: first
      push: expect-b
    - match: \d+
      scope: constant.numeric

  expect-b:
    - match: b
      scope: second
      set: expect-c

  expect-c:
    - match: c
      scope: third
      pop: true
```

Notice how `a` pushes `expect-b`.  We don't `set` the first context, only the second one.  Once we find the terminator, we `pop` out.

### Lookahead Push for Meta-Scoping

Sometimes you need to apply a meta-scope to an entire stateful chunk.  When this is the case, you almost certainly want your `push` rule to be a non-consuming lookahead, rather than a consuming scoped match.  We can modify the above:

```yaml
contexts:
  main:
    - match: (?=a)
      push: expect-a
    - match: \d+
      scope: constant.numeric

  expect-a:
    - meta_scope: meta.abc
    - match: a
      scope: first
      set: expect-b

  expect-b:
    - meta_scope: meta.abc
    - match: b
      scope: second
      set: expect-c

  expect-c:
    - meta_scope: meta.abc
    - match: c
      scope: third
      pop: true
```

## Bail Outs

[...]

PR #1850, opened 2019-01-23
[D] Syntax Improvements

BenjaminSchaaf:
This is a complete re-implementation of the D syntax highlighting to support all current features and mark entities correctly. I've validated correctness by going through some portion of the D standard library and vibed.

Also includes improvements to the build system, with error highlighting and dub support.


wbond:
It looks like a lot of work when into this, nice! I'll work on doing a review soon.

One thing that I like to do with rewrites is try to preserve the existing tests as much as possible, and have distinct commits where we can review why previous tests needed to be changed. This just helps ensure we didn't regress things that may have been addressed previously. Perhaps move the old test file into the tests dir and add a suffix `-old`? I apologize if I missed the old tests being included within the new tests.

D/D.sublime-syntax
```diff
@@ -8,339 +8,1970 @@ file_extensions:
...
-          (\w+)
-          (?:\s*,\s*(\w+))?
-          (?:\s*,\s*(\w+))?
-          (?:\s*,\s*(\w+))?
-          (?:\s*,\s*(\w+))?
-          (?:\s*,\s*(\w+))?
-          (?:\s*,\s*(\w+))?
-        )? # super class
+    - meta_include_prototype: false
+    - match: '^#!.*$'
```
deathaxe: `#` should be scoped as `punctuation.definition.comment.number-sign` to match the current shell syntax.

Otherwise I am not sure about the global strategy to support shebang directly.

The main shell scope is `source.shell` The bash syntax currently uses `soure.shell.bash`.

To follow this kind of scheme of different shell types any syntax would need to use `source.shell.<syntax>`.

Example: `source.shell.python`, `source.shell.erlang`, `source.shell.tcl`, ... .

Therefore each syntax, which wants to assign itself as a shell type, should provide a dedicated `<Syntax> Shell.sublime-syntax`, I think. (Btw. might be worth a RFC discussion?).

@all: Maybe something like that would be worth thinking about?

```YAML
%YAML 1.2
---
name: D Shell (D)
file_extensions:
  - ???
first_line_match: |-
  (?x:
    ^ \s* \#! .* \bD\b |            # shebang
    ^ \s* \#+ \s* -\*- .*? \bD\b .*? -\*-  # editorconfig
  )
scope: source.shell.d

contexts:

  main:
    - match: ^\s*(\#)
      captures:
        1:
          comment.line.number-sign.shell
          punctuation.definition.comment.shell
      push:
        - meta_content_scope: comment.line.number-sign.shell
        - match: \n
          scope: comment.line.number-sign.shell
          pop: true
    - match: ^
      set: scope:source.d

  prototype:
    - include: scope:source.d#prototype
```

_Note: We propably should not be too restrictive with whitespace in front of shebangs?_

D/D.sublime-syntax
```diff
@@ -8,339 +8,1970 @@ file_extensions:
...
-        5: storage.type.template.d
-        6: punctuation.separator.inheritance.d
-        7: entity.other.inherited-class.d
-        8: entity.other.inherited-class.d
-        9: entity.other.inherited-class.d
-        10: entity.other.inherited-class.d
-        11: entity.other.inherited-class.d
-        12: entity.other.inherited-class.d
-        13: entity.other.inherited-class.d
+        1: keyword.control.module.d
```
deathaxe: With regard of som of the most recent discussions `keyword.declaration.module` might be a more suitable choice?

D/D.sublime-syntax
```diff
@@ -8,339 +8,1970 @@ file_extensions:
...
+        2: entity.name.namespace.d
+        3: punctuation.terminator.d
+      set: statement-list-in
+    - match: '\s'
+    - match: (?=.)
+      set: statement-list-in
```
deathaxe: Instead of 78-80 just using `- match: (?=\S)` in 79 would be more suitable I think.

D/D.sublime-syntax
```diff
@@ -8,339 +8,1970 @@ file_extensions:
...
+    - meta_include_prototype: false
+    - match: '^#!.*$'
+      scope: comment.line.shebang.d
+      set: main-after-shebang-in
+    - match: '(?=.)'
+      set: main-after-shebang-in
+  main-after-shebang-in:
+    - include: attribute-specifier-in
+    - match: '\b(module)\s+({{identifier}})(;)'
+      scope: meta.module.d
```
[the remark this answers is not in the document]
BenjaminSchaaf: I don't think it makes too much sense for D, as it behaves similarly to python in that files are just modules. Everything is already in an implicit namespace regardless of whether the module is declared or not. But its probably not unhandy to have anyways, maybe some future feature could depend on this.

D/D.sublime-syntax
BenjaminSchaaf: In relation to the whitespace restriction in front of the `#!`: putting a space in front fails to compile with D.

jrappen:
Are you sure your tests for the first lines of the multi-line strings actually work as intended?

Looking at:

* L37-40
* L45-49
* L52-55
* etc.

I don't know this language, but looking at the preview here on Github makes me bring this to attention.

D/D.sublime-syntax
FichteFoll: I'm not sure why we'd want a separate syntax for these files. They are still just files in D with a shebang. Imo just highlight the shebang line as a comment, assign appropriate scopes and be done with it.

We could probably unify this across syntaxes as `comment.line.number-sign.shebang` and special-case it.

rwols: I agree with @FichteFoll, KISS!

BenjaminSchaaf:
@jrappen Are you referring to "comments" inside the strings? If so, the test runner doesn't parse the syntax, it just checks what the line starts with.

D/D.sublime-syntax
deathaxe: I am fine with this strategy as well.

But anyway: the current solution breaks `embed` statements, by calling `set: ...` to leave the `main` context.

The `escape` of the following statement would be ignored.

```YAML
    - match: ''
      embed: scope:source.d
      embed_scope: source.d.embedded.markdown
      escape: (?=^\`{3})
```

_Note: Found it while applying that strategy to Erlang._

The JavaScript syntax uses `push` which works fine with `embed`.

So something like that might work

```YAML
contexts:

  main:
    - match: ''
      push: [main-after-shebang, comment-shebang]

  comment-shebang:
    - match: ^\#!
      scope: punctuation.definition.comment.shell
      set:
        - meta_scope: comment.line.number-sign.shebang.d
        - match: \bD\b
          scope: constant.language.d
        - match: \n
          pop: true

  main-after-shebang:
```

deathaxe: Pushing away might on the other hand cause issues, if the syntax is `include`ed into another one, as the `main-after-shebang` is never popped off.

With that said: This kind of shebang handling is problematic for syntaxes which might want to be ebedded/included into other ones.

deathaxe: see: https://github.com/SublimeTextIssues/Core/issues/2395

D/D.sublime-syntax
```diff
@@ -8,339 +8,1971 @@ file_extensions:
...
+      set:
+        - match: '\)'
+          scope: punctuation.section.group.end.d
+          set: function-definition-after-first-argument-list
+        - match: '{{parameter_attribute_lookahead}}'
+          set: [function-definition-after-first-argument-list, function-argument-definition-list-content]
+        - match: '(?=\S)'
+          set: [constructor-argument-list-after-value, value]
+    - include: not-whitespace-illegal-pop
+  constructor-argument-li
```
deathaxe: Hint: indention too low.

D/D.sublime-syntax
```diff
@@ -8,339 +8,1971 @@ file_extensions:
...
+    - match: '(\+\+|--)'
+      scope: keyword.operator.arithmetic.d
+      set: value-after
+    - match: '(&&|\|\|)'
+      scope: keyword.operator.logical.d
+      set: value
+    - match: '(\+|-|/|%|\*|&|\||\^\^|\^|~|<<|>>>|>>)='
+      scope: keyword.operator.assignment.d
+      set: value
+    - match: '(\+|\-|~|/|%|\^\^)'
```
deathaxe: matching + and - in two rules will cause the second one to be ignored. Should be matched once, only.

BenjaminSchaaf: The first rule ends with a `=` though, so the 2nd one would trigger if there isn't a `=` after.

D/D.sublime-syntax
BenjaminSchaaf: I don't quite follow, sorry. What do you mean by this?

D/D.sublime-syntax
deathaxe: Sorry, my fault.

D/D.sublime-syntax
deathaxe: Just see, picked the wrong lines: Please check `invariant-block` context. SublimeLint commplained about the rules after `set` to be inproperly indented.

```
    - match: '{'
      scope: punctuation.section.block.begin.d
      set:
      - match: '}'
        scope: punctuation.section.block.end.d
        pop: true
      - include: statement-list-in
```

BenjaminSchaaf: SublimeLint looks like a framework for other linters to me. I'm not sure what linter you're running on this. If its complaining about the `- include: not-whitespace-illegal-pop` line then its certainly a false positive. The contents are handled by `(?=\S)` and `not-whitespace-illegal-pop` just highlights everything as invalid.

deathaxe: Nope: Like I said, I selected the wrong lines by accident. It's not `- include: not-whitespace-illegal-pop` the linter complains about but the construct of my previous post in line 1663ff.

Sorry for the confustion.

BenjaminSchaaf: Ah yes, sorry for not catching on sooner.

BenjaminSchaaf:
@deathaxe I've added spacing where I think it improves readability. I've kept together contexts that I think belong together and separated others.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2031 @@ file_extensions:
...
+          scope: punctuation.section.group.end.d
+          set: function-definition-after-first-argument-list
+        - match: '{{parameter_attribute_lookahead}}'
+          set: [function-definition-after-first-argument-list, function-argument-definition-list-content]
+        - match: '(?=\S)'
+          set: [constructor-argument-list-after-value, value]
+    - include: not-whitespace-illegal-pop
+  constructor-argument-list-after-value:
+    - match: ','
+      scope: punctuation.
```
deathaxe: According to #1228 `try`, `catch`, `finally` would get `keyword.control.exception` with maybe the keyword appanded as subscope.

deathaxe: Already asked the question in #1228 whether it might make sense to scope `switch` as `keyword.control.conditional` because of two reasons:

1. `keyword.control.flow` is used for many keywords like `goto`, `return`, etc. If they are scoped differently to normal conditional or loop statements, switch would do so as well, even though it might be more related with keywords like if, else, ...
2. A switch-case statement is somehow an efficient replacement of an if-elseif-elseif...else statement.

What do you think about it? 

Most syntaxes use `keyword.control.flow` so far. Hence your choice is very perfect here. I just wondered if it might make sense to change in general (other syntaxes as well).

deathaxe: Why not using `parens` here? Brackets are `[`.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2031 @@ file_extensions:
...
+    - match: '\b(static)\s+(foreach|foreach_reverse)\b'
       captures:
-        1: punctuation.definition.comment.d
-  constant_placeholder:
-    - match: '(?i:%(\([a-z_]+\))?#?0?\-?[ ]?\+?([0-9]*|\*)(\.([0-9]*|\*))?[hL]?[a-z%])'
-      scope: constant.other.placeholder.d
-  function_assertions:
-    - match: '(in|out)\s*(?=\{)'
+        1: keyword.control.flow.d
+        2: keyword.control.flow.d
```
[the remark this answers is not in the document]
BenjaminSchaaf: `static` does not have the same meaning as C/C++, it simply makes the scope static rather than dependent on the context. Sometimes it also means "compile time": `static foreach` for example.

D/D.sublime-syntax
BenjaminSchaaf: I was under the impression brackets refer to all types so I reserved parens for function calls.

BenjaminSchaaf: At least in D switch is a lot more complicated than a simple if-elseif-else statement. For instance the following is perfectly valid:

```d
switch (i) {
    case 1:
    {
        case 2:
    }
    break;
}
```

That and you can use `goto case` and `goto default`. The way I see it `switch` is just as much a conditional as `while`. You can even use switch as a loop:

```d
void main() {
    int a;
    switch (1) {
        case 1:
            writeln(a);
            if (a++ == 10)
                break;
            goto case 1;
        default:
            break;
    }
}
```

D/D.sublime-syntax
```diff
@@ -8,339 +8,2031 @@ file_extensions:
...
-        - include: $top_level_main
-    - match: |-
-        (?x)
-                    (?:  ^                                 # begin-of-line
-                      |  (?: (?<!else|new|=) )             #  or word + space before name
-                    )
-                  ((?:\b(?:private|package|protected|public|export|static|final|native|synchronized|abstract|threadsafe|transient|override)\b\s*)*) # modifier
-                    (~this) # actual name
-                     \s*(\()                           # start bracket or end-of-line
+        2: constant.numeric.binary.d
```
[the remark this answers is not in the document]
deathaxe: I can understand this kind of thinking from perspective of compatibility with existing color schemes, but it lead to weird scope names in the past and mixes up the approach to define well structured standardized scope names with creating color schemes. How such scopes are colorized depend very much on the color scheme being actually used. It might look well with your prefered color scheme, but might look totally odd with another one. Hence adapting the scope names to existing color schemes is the wrong way round.

First step is to define clearly structured scope names, which hopefully - with much effort left for future changes - reflect comparable structures of all the different syntaxes the same way and then create good color schemes upon that standard, which make look all syntaxes the same way.

The new `sublime-color-scheme` format makes tweaking existing color schemes very easy anyway.

D/D.sublime-syntax
deathaxe: In that case the general `flow` is very perfect.

D/D.sublime-syntax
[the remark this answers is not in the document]
BenjaminSchaaf: Definitely agree (hence the change), was just making light of my thought process at the time of writing

D/D.sublime-syntax
```diff
@@ -8,339 +8,2031 @@ file_extensions:
...
-        6: punctuation.separator.inheritance.d
-        7: entity.other.inherited-class.d
-        8: entity.other.inherited-class.d
-        9: entity.other.inherited-class.d
-        10: entity.other.inherited-class.d
-        11: entity.other.inherited-class.d
-        12: entity.other.inherited-class.d
-        13: entity.other.inherited-class.d
+    - meta_include_prototype: false
+    - match: '^(#)!(.*)$'
```
deathaxe: 1. Avoiding capture groups could help for performance reasons
2. Why scoping the comment with `source.shell`? 

The following looks a bit more complicated, but is faster then using capture groups.

```YAML
    - match: ^\#!
      scope: punctuation.definition.comment.shell
      set:
        - meta_scope: comment.line.number-sign.shell
        - match: \bD\b
          scope: constant.language.d
        - match: \n
          set: main-after-shebang-in
    - match: ''
      set: main-after-shebang-in
```

deathaxe: see: https://github.com/sublimehq/Packages/issues/757#issuecomment-326882205

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+      set: function-argument-definition-list-content
+    - match: '\)'
+      scope: punctuation.section.parens.end.d
+      pop: true
+    - include: not-whitespace-illegal
+  type-specialization:
+    - match: '\b(?:({{type_specialization_storage}})|({{type_specialization_keyword}}))\b'
+      captures:
+        1: storage.type.d
+        2: keyword.d
```
wbond: There is probably a second level we could add to this scope?

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+      scope: punctuation.section.group.end.d
+      set: value-group-after-parens
+    - match: '\b({{name}})\b'
+      scope: variable.parameter.d
+      set: value-group-after
+    - match: ','
+      scope: punctuation.separator.sequence.d
+      set: [value-group-after-parens, function-argument, function-argument-or-type]
+    - match: '\.\.\.'
+      scope: keyword.operator.d
```
wbond: Some syntaxes call this `keyword.operator.spread` (JS) others `keyword.operator.variadic` (C, C++, Go, Java). It probably makes sense to adopt the second.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+    - include: not-whitespace-illegal
+  value-group-after-parens:
+    - include: function-attribute-in
+    - match: (?={)
+      set: [value-after, block-statement]
+    - match: '=>'
+      scope: keyword.operator.lambda.d
```
wbond: This should be `storage.type.function.d keyword.declaration.function.arrow.d`

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
-    - match: /\*
-      scope: punctuation.definition.comment.d
+      set: [value-condition-after, value]
+    - match: '(?=function|delegate)'
+      set: [value-after, basic-type2]
+    - match: '(?=\S)'
+      pop: true
+  value-condition-after:
+    - match: ':'
+      scope: keyword.operator.logical.d
```
wbond: This seems like an odd scope for `:`, but I am not sure of how it is used

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+    - match: ','
+      scope: punctuation.separator.sequence.d
+      set: value-array-list
+    - include: not-whitespace-illegal
+
+  alias-in:
+    - match: '\b(alias)\s+({{identifier}})\s+(this)(;)'
+      captures:
+        1: keyword.control.alias.d
+        2: variable.d
```
wbond: There should probably be a second label on this scope.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+      set:
+        - match: '\b{{name}}\b'
+          scope: entity.name.type.d
+          set: alias-new
+        - include: not-whitespace-illegal
+    - include: not-whitespace-illegal
+
+  class-in:
+    - match: '\b(class)\b'
+      scope: storage.type.class.d
```
wbond: This should be `storage.type.class.d keyword.declaration.class.d`

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+    - match: ':'
+      scope: punctuation.separator.mapping.d
+      set: [optional-template-body-block, base-class-list]
+    - match: '(?={)'
+      set: optional-template-body-block
+    - include: not-whitespace-illegal
+
+  struct-in:
+    - match: '\b(struct)\b'
+      scope: storage.type.struct.d
```
wbond: `storage.type.struct.d keyword.declaration.struct.d`

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+              set: [optional-template-body-block, function-argument-definition-list]
+            - match: '(?=\S)'
+              set: optional-template-body-block
+        - match: '(?={)'
+          set: optional-template-body-block
+        - include: not-whitespace-illegal
+
+  union-in:
+    - match: '\b(union)\b'
+      scope: storage.type.union.d
```
wbond: `storage.type.union.d keyword.declaration.union.d`

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+              set: [optional-template-body-block, function-argument-definition-list]
+            - match: '(?=\S)'
+              set: optional-template-body-block
+        - match: '(?={)'
+          set: optional-template-body-block
+        - include: not-whitespace-illegal
+
+  enum-in:
+    - match: '\b(enum)\b'
+      scope: storage.type.enum.d
```
wbond: `storage.type.enum.d keyword.declaration.enum.d`

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+    - match: '\b(debug)\b'
+      scope: keyword.control.conditional.d
+      push: [optional-block-statement-or-label, debug-condition]
+
+  version-condition:
+    - match: '\('
+      scope: punctuation.section.parens.begin.d
+      set:
+        - match: '\b(unittest|assert)\b'
+          scope: keyword.d
```
wbond: We should probably add a second label to this scope.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+  static-foreach-in:
+    - match: '\b(static)\s+(foreach|foreach_reverse)\b'
       captures:
-        1: punctuation.definition.comment.d
-  constant_placeholder:
-    - match: '(?i:%(\([a-z_]+\))?#?0?\-?[ ]?\+?([0-9]*|\*)(\.([0-9]*|\*))?[hL]?[a-z%])'
-      scope: constant.other.placeholder.d
-  function_assertions:
-    - match: '(in|out)\s*(?=\{)'
+        1: keyword.control.loop.d
```
wbond: This seems like an unusual scope for `static`, but I am not familiar with its usage.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+        2: keyword.control.loop.d
+      push: [optional-block-statement, foreach-arguments]
+
+  static-assert-in:
+    - match: '\b(static)\s+(assert)\b'
+      captures:
+        1: keyword.control.conditional.d
+        2: keyword.control.conditional.d
```
wbond: This is a different scope than `assert` is getting than earlier in the syntax def.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
-  function_assertions:
-    - match: '(in|out)\s*(?=\{)'
+        1: keyword.control.loop.d
+        2: keyword.control.loop.d
+      push: [optional-block-statement, foreach-arguments]
+
+  static-assert-in:
+    - match: '\b(static)\s+(assert)\b'
+      captures:
+        1: keyword.control.conditional.d
```
wbond: Similar to above, this seems like an uncommon scope for `static`.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+      push: [expect-end-of-line, assert-arguments]
+
+  constructions-in:
+    - match: '{{construction_lookahead}}'
+      push: construction
+  construction:
+    # Post-blit constructor
+    - match: '\b(this)\s*(\()\s*(this)\s*(\))'
+      captures:
+        1: entity.name.function.post-blit.d
```
wbond: Hmm, so is this defining a function named `this`? I'm not sure how practical/useful this will be in terms of the indexer, since I would imagine Goto Definition of `this` would list a lot. But perhaps this is related to that tmPreferences file for excluding constructors that I noted earlier in the review.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+  # arguments and switch to parameters when possible
+  constructor-argument-list:
+    - match: '\('
+      scope: punctuation.section.group.begin.d
+      set:
+        - match: '\)'
+          scope: punctuation.section.group.end.d
+          set: function-definition-after-first-argument-list
+        - match: '{{parameter_attribute_lookahead}}'
+          set
```
wbond: This should probably use a similar or the same scope as `assert`, right?

wbond: One thing to look out for with constructs such as this is as the user is typing, when they've got `debug` typed in, ideally it should be highlighted the same as `debug =`. This may be the case here, but something to check for.

wbond: This is typically scoped as a separator. Imagine as a color scheme author coloring curly braces and separators differently and then for some reason this `:` looks like a curly brace.

D/D Single File.sublime-build
```diff
@@ -0,0 +1,13 @@
+{
+	"shell_cmd": "dmd $file",
```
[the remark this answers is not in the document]
BenjaminSchaaf: Fair enough, my only concern was some people use the Universal D binaries which you can "activate" in your bashrc. Specifically I did this in testing, and since the binaries aren't in the normal PATH it would only work with shell_cmd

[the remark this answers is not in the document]
wbond: In that case we just need $file quoted, I believe

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+            1: punctuation.definition.string.end.d
+            2: storage.type.string.d
           pop: true
-        - match: \b(_|:)\b
+        - match: '{{escape_sequence}}'
+          scope: constant.character.escape.d
+        - match: \\.
+          scope: invalid.illegal.unknown-escape.d
+    # Wysiwyg string
+    - match: r"
```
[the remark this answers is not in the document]
BenjaminSchaaf: Does this still make sense to use `storage.type` considering a prefix never influences the type of the string and is only lexical, whereas the postfixes do actually change the type of the string literal?

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+      pop: true
+
+  basic-type2:
+    - match: '\*'
+      scope: keyword.operator.pointer.d
+    - match: '\['
+      scope: punctuation.section.brackets.begin.d
+      push: [basic-type2-after-brackets, value-list]
+    - match: '\b(delegate|function)\b'
+      scope: keyword.other.d
```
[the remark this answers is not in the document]
BenjaminSchaaf: In this case it's a keyword for a type, ie. `void function(int)` is the type of a function pointer to a `void(into)` function

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+        - match: '\.\.\.'
+          scope: keyword.operator.d
+        - match: ','
+          scope: punctuation.separator.sequence.d
+          set: [function-type-argument, function-type-argument-type]
+        - match: '='
+          scope: keyword.operator.assignment.d
+          push: value
+        - match: ':'
+          scope: keyword.operator.assignment.d
```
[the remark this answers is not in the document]
BenjaminSchaaf: It's used for matching, in a pattern matching sense

D/D.sublime-syntax
BenjaminSchaaf: So I'm not sure about this one. The identifier can refer to anything from a member to a method and can be of literally any type.

D/D.sublime-syntax
BenjaminSchaaf: In this case the keywords have no direct meaning. They're just scoped as keyword because that's what the lexer does, but otherwise they're no different here to a normal name.

D/D.sublime-syntax
BenjaminSchaaf: `static` in this case means "compile time" (which has some other implications), but in D these kinds of statements are specifically referred to as "static if", "static foreach" and "static assert". Because of that I felt it better to have it appear more as a 2 word keyword than a "modifier" and a keyword, as this fits its use and meaning better.

D/D.sublime-syntax
BenjaminSchaaf: Constructors are all defined through `this` and destructors through `~this`. Hence the exclusion in tmPreferences. It's still useful to have indexed within a file though, since a fair number of Devices files only contain a few constructors.

D/tests/syntax_test_old.d
```diff
@@ -58,30 +58,30 @@ template testTemplate(X)
 //                  ^ storage.attribute
 
 auto takeByRef(ref int h)
-// ^ storage.type
+// ^ storage.modifier
```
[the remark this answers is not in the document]
BenjaminSchaaf: Is is kind of an implicit type but behaves much more like an annotation:
You can use it with `:`
```
auto:
static foo() {}
```

It can be in any order:
```
auto @safe foo() {}
```

It isn't even required:
```
static foo() {}
```

D/tests/syntax_test_old.d
```diff
@@ -58,30 +58,30 @@ template testTemplate(X)
...
-    // ^ storage.modifier
+    // ^ keyword.other
     {
         h += 1;
     }
 }
 
 void main(char[][] args)
-//    ^ meta.function entity.name.function
+//    ^ entity.name.function
```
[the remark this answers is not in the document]
BenjaminSchaaf: I haven't added meta scopes to pretty much anything. Maybe that should be another one of the "big things"?

D/tests/syntax_test_old.d
```diff
@@ -176,7 +176,7 @@ class CmdLin
 
     static typeof(_totalc) total()
     // ^ storage.modifier
-    //      ^ storage.type
+    //      ^ meta.block.d keyword.other.d
```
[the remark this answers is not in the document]
BenjaminSchaaf: Since typeof is used in other places, eg: `alias f = typeof(r);` or even `typeof(s).foo();` I felt it better to stay consistent with those.

D/D.sublime-syntax
BenjaminSchaaf: At least in this case `version` and `debug` are other statements, so they'll be highlighted correctly. Though it might still be better to leave the `=` parsing to those statements instead.

D/DMD Output.sublime-syntax
```diff
@@ -0,0 +1,12 @@
+%YAML 1.2
+---
+name: DMD Output
+hidden: true
+scope: source.build_output
+contexts:
+  main:
+    - match: '^(..[^:\n]*):([0-9]+):?([0-9]+)?:? '
```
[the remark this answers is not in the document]
BenjaminSchaaf: Forgot to delete this, leftover from a copy-paste from C++ :)

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
-        8: entity.other.inherited-class.d
-        9: entity.other.inherited-class.d
-        10: entity.other.inherited-class.d
-        11: entity.other.inherited-class.d
-        12: entity.other.inherited-class.d
-        13: entity.other.inherited-class.d
+    - meta_include_prototype: false
+    - match: '^#!'
+      scope: punctuation.definition.comment.number-sign
+      set:
```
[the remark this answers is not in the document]
BenjaminSchaaf: Hm, though that does make the whole checking that its at the front of the file thing impossible. Guess I'll have to live with that :)

D/tests/syntax_test_d.d
```diff
@@ -0,0 +1,2285 @@
+// SYNTAX TEST "Packages/D/D.sublime-syntax"
+
+module foo.a.b1_3;
+// <- meta.namespace.d keyword.declaration.module.d
+ // <- meta.namespace.d keyword.declaration.module.d
+//^^^^ meta.namespace.d keyword.declaration.module.d
+//     ^^^^^^^^^^ meta.namespace.d entity.name.namespace.d
```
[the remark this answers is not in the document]
BenjaminSchaaf: I guess I never really use goto definition for a module, however the full path makes Goto Symbol so much better when there are multiple modules with the same trailing name.

D/tests/syntax_test_d.d
```diff
@@ -0,0 +1,2285 @@
...
+//               ^ meta.namespace.d punctuation.terminator.d
+  module foo.a.b1_3;
+//^^^^^^ - keyword.control.module
+//       ^^^^^^^^^^ - entity.name.namespace
```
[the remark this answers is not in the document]
BenjaminSchaaf: Invalid, as `module` is only allowed on the first line of the file.

D/Indentation Rules.tmPreferences
```diff
@@ -8,15 +8,24 @@
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string>^(.*\*/)?\s*\}([^}{"']*\{)?[;\s]*(//.*|/\*.*\*/\s*)?$|^\s*(public|private|protected):\s*$</string>
+		<string>(?x)
+		^ (.*\*/)? \s* \} .* $
```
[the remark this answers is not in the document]
BenjaminSchaaf: Seems to behave the same way without, I believe that was a copy paste from C++ not sure if it should be removed there as well.

D/Symbol Index Hide Ctors.tmPreferences
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+    <key>scope</key>
+    <string><![CDATA[(source.d) & (entity.name.function.constructor | entity.name.function.post-blit | entity.name.function.destructor)]]></string>
```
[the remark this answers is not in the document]
BenjaminSchaaf: Constructors are always called `this`, post-blit constructors are called `this(this)` and destructors `~this`, hence the exclusion from the symbol index but not the symbol list.

D/tests/syntax_test_old.d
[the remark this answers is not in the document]
BenjaminSchaaf: To add to some of the weirdness, auto is sometimes invalid as a type standin:
```
foreach (auto f; foo) {} // fails to compile
foreach (f; foo) {} // only this is valid
```

D/D Single File.sublime-build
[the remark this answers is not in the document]
FichteFoll: `cmd` is only problematic on Windows because it doesn't find `dmd.exe` and the like on the PATH. You'd need a shell_cmd for that or a Windows-specific override for `cmd`.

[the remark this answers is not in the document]
FichteFoll: The "activate" thing is probably related to environment variables and most likely either to be activated by the login shell (in which case ST should pick it up regardless) or, if it's per-directory, either have ST already be launched from that directory by the shell or not something to be run by a session-global application like ST.

However, considering you're talking about bashrc, this is much more likely the be the former.

D/tests/syntax_test_d.d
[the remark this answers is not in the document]
FichteFoll: I wonder if this is something the syntax should handle or whether it's better left to a linter.

D/D.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: Iirc this question was raised for Python recently since `u` and `r` don't affect the string type, but it wasn't decided on what it should become. Maybe `storage.modifier`?

D/D.sublime-syntax
FichteFoll: If they don't do anything, why highlight them? Or is this based on a naming convention? In that case, something like `support.function` might seem appropriate?

D/D.sublime-syntax
FichteFoll: `keyword.declaration.function.arrow.d` over `keyword.declaration.function.lambda.d`? "lambda" should be the more general term.

D/D.sublime-syntax
BenjaminSchaaf: Because that's what the D grammar defines them as, they *are* keywords. They're a special case of version statements because they check for when unittests or asserts are enabled, but they're no different from any regular version check in every other sense. It's just like `.class` in ruby, where it's a keyword specialcased to serve another function. I don't think that makes it less of a keyword.

wbond: I think most users would expect that these keywords would be consistently colored, so I would use the same scope as they get in other parts of the syntax.

D/tests/syntax_test_d.d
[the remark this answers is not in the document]
BenjaminSchaaf: Since we can't use `set` in `main` I don't think this is even an option sadly, it has already been removed as a feature

D/D.sublime-syntax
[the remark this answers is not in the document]
BenjaminSchaaf: At least in python there's also the `b` prefix which does affect the type, so I can see an argument for highlighting all prefixes the same. `rb''` would look a bit odd if `r` and `b` are highlighted differently.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+        - match: '(?=\()'
+          set: [function-definition-after-condition, assert-arguments]
+        - match: '{{block_statement_loohahead}}'
+          set: [function-definition-after-condition, block-statement]
+        - include: not-whitespace-illegal
+    - match: '\bout\b'
+      scope: keyword.control.conditional.d
+      push: function-out-contract
+    - match: '\b(do|body)\b'
```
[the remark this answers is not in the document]
BenjaminSchaaf: This is actually for the body of the function. If you use contract programming you can do this:

```
void foo(int a)
in(a > 0)
do {
}
```

BenjaminSchaaf:
@wbond Thanks for the review, I think I've responded/fixed most of it. In terms of the `;` consistency I'm not sure what should be prioritized. In D `;` is used both as a terminator and a separator:

```
auto b = [1, 2, 3];
foreach (int a; b) { … }

int fn()
out(i; i > 0)
body { return 1; }
```

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
       captures:
-        1: keyword.control.statement.d
+        1: entity.name.function.destructor.d
+        2: punctuation.section.group.begin.d
+        3: punctuation.section.group.end.d
+      set: function-definition-after-arguments
+    - include: not-whitespace-illegal-pop
+  # This could either be function/template parameters or arguments. Assume
+  # arguments and switch to parameters when possible
+  constructor-argume
```
wbond: You'll probably need to shuffle things around a little, but ideally the entirety of the labels and accessors should get a meta scope of `meta.path`.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+    - match: (?={)
+      set: [value-after, block-statement]
+    - match: '=>'
+      scope: keyword.operator.lambda.d
+      set: value
+    - match: (?=\S)
+      set: value-after
+  value-identifier:
+    - match: '\b{{name}}\b(?=\s*\()'
+      scope: variable.function.d
```
wbond: This identifier, plus the parens and all parameter should be scoped with `meta.function-call`. The scopes `meta.function-call variable.function` are used for references in the Goto Definition popup and the Goto Reference command.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+      pop: true
+    - include: not-whitespace-illegal
+
+  function-or-variable-after-type:
+    - match: '\b({{operator_overloads}})\s*(?=\(|$)'
+      captures:
+        1: keyword.operator.overload.d
+      set: function-definition-begin
+    - match: '\b({{name}})\s*(?=\(|$)'
+      captures:
```
wbond: The function definition (and any declarations) should have `meta.function` applied at least for the keyword, function name, parameters and return type. If you want to, you can continue the scope until the closing brace for the function body. For parameter, the scope should switch from `meta.function` to `meta.function.parameters`. For the return type, it should switch from `meta.function` to `meta.function.return-type`. Note that the scope should be swapped out, and not stacked.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+  string:
+    - include: string-opt
+    - include: not-whitespace-illegal
+  string-opt:
+    # Regular string
+    - match: '"'
+      scope: punctuation.definition.string.begin.d
+      set:
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.double.d
```
wbond: In addition to `string.quoted`, each string should have `meta.string` that starts at the very first token that begins the string and continues until the very last. This is useful when dealing with interpolations since the `string.quoted` scope will normally be popped off the stack to allow the interpolated expression to show through without all of the colors from a string.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+      set:
+        - match: '\b{{name}}\b'
+          scope: entity.name.type.d
+          set: alias-new
+        - include: not-whitespace-illegal
+    - include: not-whitespace-illegal
+
+  class-in:
+    - match: '\b(class)\b'
+      scope: storage.type.class.d keyword.declaration.class.d
```
wbond: Ideally, everything from here to the closing brace of the class definition would get `meta.class`. However, at the least, from here until the opening brace should get `meta.class`.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
-            1: storage.modifier.d
+        - meta_include_prototype: false
+        - meta_scope: comment.line.number-sign.d
+        - match: '$\n?'
+          pop: true
+    - include: module-in
+    - include: statement-list-in
+  module-in:
+    - match: \bmodule\b
+      scope: keyword.declaration.namespace.d
```
wbond: Similar to classes, this should have `meta.namespace` at least until the opening brace, but really until the closing brace, if possible.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+    - match: ','
+      scope: punctuation.separator.sequence.d
+      set: base-class-list
+    - match: '(?={)'
+      pop: true
+    - include: not-whitespace-illegal
+
+  interface-in:
+    - match: '\b(interface)\b'
+      scope: storage.type.interface.d keyword.declaration.interface.d
```
wbond: Like a class, this should have `meta.interface` until at least the opening braces, but ideally the closing brace.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+    - include: end-of-line
+    - match: ':'
+      scope: punctuation.separator.mapping.d
+      set: [optional-template-body-block, base-class-list]
+    - match: '(?={)'
+      set: optional-template-body-block
+    - include: not-whitespace-illegal
+
+  struct-in:
+    - match: '\b(struct)\b'
```
wbond: `meta.struct` until the first, or last brace.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+              set: [optional-template-body-block, function-argument-definition-list]
+            - match: '(?=\S)'
+              set: optional-template-body-block
+        - match: '(?={)'
+          set: optional-template-body-block
+        - include: not-whitespace-illegal
+
+  union-in:
+    - match: '\b(union)\b'
+      scope: storage.type.union.d keyword.declaration.union.d
```
wbond: `meta.union` until the first or last brace.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+            - match: '(?=\()'
+              set: [optional-template-body-block, function-argument-definition-list]
+            - match: '(?=\S)'
+              set: optional-template-body-block
+        - match: '(?={)'
+          set: optional-template-body-block
+        - include: not-whitespace-illegal
+
+  enum-in:
+    - match: '\b(enum)\b'
```
wbond: `meta.enum` until the first or last brace.

D/D.sublime-syntax
```diff
@@ -8,339 +8,2057 @@ file_extensions:
...
+        1: keyword.other.assert.d
+        2: keyword.other.assert.d
+      push: [expect-end-of-line, assert-arguments]
+
+  constructions-in:
+    - match: '{{construction_lookahead}}'
+      push: construction
+  construction:
+    # Post-blit constructor
+    - match: '\b(this)\s*(\()\s*(this)\s*(\))'
```
wbond: This should get `meta.function` and specializations as appropriate.

D/D.sublime-syntax
BenjaminSchaaf: `module` can only appear once at the top of the file and has no braces. I guess I'll add it until the eof

wbond: That makes sense then 🙂

D/D.sublime-syntax
```diff
@@ -8,339 +8,2032 @@ file_extensions:
...
+          captures:
+            1: punctuation.definition.string.end.d
+            2: storage.type.string.d
+          pop: true
+        - match: '{{hex_char}}\s*{{hex_char}}'
+          scope: constant.character.escape.d
+        - match: '\S'
+          scope: invalid.illegal.unknown-escape.d
+    # Deliminated string with custom identifier
+    - match: 'q"({{name}})(.*)$'
```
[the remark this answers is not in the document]
BenjaminSchaaf: What scope should the name have, I doubt `punctuation.definition` makes any sense?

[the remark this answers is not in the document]
wbond: It does seem like an odd scope, but I'm not sure what else would work here.

I guess you could do something like `variable.other`, but that may be a little confusing, making people think it is interpolation.

Perhaps something like `constant.other.delimeter`?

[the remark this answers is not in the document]
BenjaminSchaaf: `constant.other.delimiter` sounds good.

D/D.sublime-syntax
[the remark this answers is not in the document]
BenjaminSchaaf: afaik it's always a keyword in D?

D/D.sublime-syntax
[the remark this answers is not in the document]
BenjaminSchaaf: I agree assignment doesn't make that much sense, maybe `punctuation.separator.mapping`?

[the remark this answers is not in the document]
wbond: So is `int` a default value for `A`?

D/D.sublime-syntax
BenjaminSchaaf: I'll get the others done, but `return-type` is impossible to do properly as the parser doesn't know whether the first thing on a line is a value or a type for either a variable or function until later, eg:

```
foo!(1, "bar")
// This would need meta.function.return-type, but we can't backtrace to this line to do that
bar() {}
```

wbond: In situations like this, we typically use a heuristic and/or best effort.

If we see `int main()` we can set the appropriate meta scopes. If someone writes D that is hard to highlight when highlighting a single line at a time, then they'll just get a slightly less correct parsing.

D/D.sublime-syntax
BenjaminSchaaf: It might be a bit weird if they type `static` it highlights as an attribute but finish with `static foreach` it changes to a modifier. Not sure on this.

D/D.sublime-syntax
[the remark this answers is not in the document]
BenjaminSchaaf:
```
void foo(A : int) {}
// practically equivalent to
void foo(A) if (is(A : int)) {}
```

It has the same meaning as `==` but also works with inheritance.

D/D.sublime-syntax
BenjaminSchaaf: This if for `cond ? v1 : v2`, same meaning as C++

wbond: So in C we are currently using `keyword.operator.ternary`. `logical` is mostly used for `!`, `not`, `&&` and `||`.

D/D.sublime-syntax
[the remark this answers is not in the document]
wbond: Ah, okay, thanks for explaining.

So how about `keyword.operator.comparison`?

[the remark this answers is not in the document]
BenjaminSchaaf: Sure

wbond:
Here are a few constructs from the dub codebase that seem not to be handled properly, although the issue for some may just be my understanding. 🙂 

----

https://github.com/dlang/dub/blob/master/source/dub/internal/vibecompat/core/log.d#L65

```d
appender!string()
```

The `!` should probably be something like `punctuation.section.generic`. The parens and everything between have `meta.function-call` doubled up.

----

https://github.com/dlang/dub/blob/master/source/dub/internal/vibecompat/core/log.d#L69

```d
auto threadid = () @trusted {
```

The `@` is marked as punctuation for an annotation, but there is no `meta.annotation` scope and `trusted` is not `variable.annotation`, but rather `storage.attribute`.

----

https://github.com/dlang/dub/blob/master/source/dub/commandline.d#L243

```d
dub = new Dub(NativePath(getcwd()));
```

Right now `Dub` is `storage.type`, so it won't show up in Goto References. I think we'll need to add a .tmPreferences to make sure `meta.function-call storage.type` is indexed as a reference. See `Default/Indexed Reference List.tmPreferences` for an example. However, that will probably catch type casts in function calls. We may need to change `storage.type` (in this case) to something like `variable.function` or `variable.class` (with an appropriate tmPreferences file).

---

https://github.com/dlang/dub/blob/master/source/dub/description.d#L49

```d
Exception("Package '"~name~"' not found in dependency tree.")
```

Right now the `~` is scoped as an arithmetic operator. I am guessing that is a concatenation operator?

----

https://github.com/dlang/dub/blob/master/source/dub/dub.d#L128

```d
PackageSupplier[] m_packageSuppliers;
```

`PackageSupplier` is not recognized as a type, presumably because of the `[]`.

----

https://github.com/dlang/dub/blob/master/source/dub/dub.d#L764

```d
final switch (list_type) with (ListBuildSettingsFormat) {
```

The `with` clause is marked as invalid.

----

https://github.com/dlang/dub/blob/master/source/dub/dub.d#L958

```d
(a, b) => a.version_ < b.version_
```

`b` is scoped `variable.parameter`, but `a` is `variable.other`.

----

https://github.com/dlang/dub/blob/master/source/dub/init.d#L124-L141

What exactly is this here? It seems to be D code, but inside of an unquoted string.

----

https://github.com/dlang/dub/blob/master/source/dub/init.d#L181-L196

Are the `%` placeholders a built-in string thing?

BenjaminSchaaf:
@wbond 
`@trusted` is a built-in attribute just like `pure`, `static`, `auto`, etc. It just has a `@` in front for some reason.
----
`~` is indeed concatenation
----
```
PackageSupplier[] m_packageSuppliers;
```
Can't be reliably parsed as `[]` is a valid operation:
```
PackageSupplier[].map!(foo => foo.bar);
```
----
```
(a, b) => ...
```
Is another parsing ambiguity, though one that we could probably special-case. The following is a worst-case:
```
(foo!(123, [2, 3, 4])[] a)
{}
```
```
----
`q{}` is a token string, all contents must be valid D tokens. This is useful when writing code for `mixin` or in this case when writing default code to `src/app.d` when initializing a dub project.
----
`format` does printf-style formatting. D has uniform function call syntax so:
```
void f(string a);
f("foo");
"foo".f(); // equivalent
```

PR #1857, opened 2019-01-29
[Python] added a completions file and pypy to the build system

AmjadHD:
- Added a `Python.sublime-completions` file and removed the Snippets folder
- Removed the syntax check from the build system and added pypy
- Removed pyx, pyx.ini and pxd from `file_extensions` in the Syntax, a Cython syntax is needed for these files.


AmjadHD:
I need a little help in the completions file, it's silly but I couldn't handle the caracters that need escaping in the `class` snippet `contents`.

deathaxe:
Not sure about it to be a good idea to implement completions for something which is meant to be a snippet as the fuzzy search applied to completions will most likely result in trigger words like `sc` to come into your way.

AmjadHD:
do you suggest to use the original tab triggers (from the snippets file), or the completions file is a bad idea ?

AmjadHD:
just to note , I was using those short triggers for a long time, and they have never come in my way.

Python/Python.sublime-build
```diff
@@ -8,8 +8,8 @@
 	"variants":
 	[
 		{
-			"name": "Syntax Check",
-			"shell_cmd": "python -m py_compile \"${file}\"",
```
FichteFoll: Why remove this variant?

AmjadHD: I don't see it's necessary.

AmjadHD:
I fail to see the concept here, I thought snippets files were a legacy form that `.sublime-completions` files come to replace.

Python/Python.sublime-build
keith-hall: I don't have `pypy` installed, but I have Python 3.7, so I'm not convinced the standard "syntax check" command should be replaced with a less commonly available one...

AmjadHD: > I don't have pypy installed

I don't have it either, but it is the the most common interpreter after CPython, I would argue that we shouldn't have a `Ant` build system for Java.

> standard "syntax check"
How is it standard ?

AmjadHD: > I would argue that we shouldn't have a Ant build system for Java

if pypy is not worth having in the Python build system because it's a less common command, than why theres an `Ant` build system for java other than the standard `javac` build system (not everyone have Ant), I see it's the same case here, if I'm correct.

keith-hall: > How is it standard?

it's using `python`, the same interpreter that is used to run/execute the file...

> if pypy is not worth having in the Python build system because it's a less common command

I don't think anybody said it's not worth having.

> than why theres an `Ant` build system for java other than the standard `javac` build system (not everyone have Ant), I see it's the same case here, if I'm correct.

No, because you are proposing to remove `python` (or to use your example, `javac`) and replace it with `pypy3` (`Ant` in your example), and disable the standard approach and only leave the non-standard one.

FichteFoll:
Snippets and completions function differently, mostly with regards to their priority for completions suggestions and how a precise tab trigger of a snippet will always win. What I described earlier is how they would be used.

There is some information here: http://docs.sublimetext.info/en/latest/extensibility/completions.html#sources-for-completions-and-their-priorities

Python/Python.sublime-build
FichteFoll: No issues with adding another variant that seems useful.

PR #1863, opened 2019-02-03
[ShellScript] Control Structure Improvements

### LINKED ISSUE #1868: [Bash] `if` comparisons with negated command
This one is pretty minor, but I'm going to log it for the record. The second `!` is marked as the command instead of `git`.

``` sh
if [[ ! -z "$extra_status_args" ]] && ! git diff --cached --quiet; then
    echo "Staged files exist"
fi
```

All of these look right:

``` sh
if [[ ! -z "$extra_status_args" ]] && git diff --cached --quiet; then
    echo "Staged files (don't) exist"
fi

if ! git diff --cached --quiet; then
    echo "Staged files exist"
fi

if ! git diff --cached --quiet && [[ ! -z "$extra_status_args" ]]; then
    echo "Staged files exist"
fi
```

### LINKED ISSUE #1871: Bash malformed bash-if condition
```bash
# Replace `cd` to go to file's path if DIR is a file
cd() {
	# Breaking command line:
	# cd PATH [-L|[-P [-e]] [-@]]
	local file="${!#}"

	# echo $file

	if (( "$#" )) && ! [[ -d "$file" ]]; then
		# echo "'$#'"
		# for i in "${@:1:($#-1)}" ; do echo "'$i'" ; done
		# echo "'${file%/*}'"
		builtin cd "${@:1:($#-1)}" "${file%/*}"
	else
		builtin cd "$@"
	fi
}
```
_I know the example is stupid, bear with me ..._

image

_No need to say that `[[ !` are indeed highlighted correctly, I guess ..._

deathaxe:
Fixes #1868 
Fixes #1871

This PR ...

1. proposes to rename the control keywords in order to comply with #1228 inspired by the structure of the C# syntax, which should be applied to other syntaxes in the future as well

2. fixes a design issue with the _case_ context
3. improves the keyword boundary lookahead and applies it to the case contexts

For details please refere to the commit messages.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -187,7 +187,7 @@ contexts:
       scope:  keyword.control.loop.until.shell
     - match: \bcase{{keyword_boundary_end}}
       scope: keyword.control.conditional.case.shell
-      set: [case-clause-commands, case-clause-patterns, case-clause-patterns-first-character, case-word]
+      set: [case-body, case-word]
     - match: \bcontinue{{keyword_boundary_end}}
       scope: keyword.control.flow.continue.shell
     - match: \bbreak{{keyword_boundary_end}}
@@ -199,6 +199,31 @@ contexts:
       scope: keyword.control.in.shell
       pop: true
     - include: expansion-and-string
+
+  case-body:
+    - meta_scope: meta.conditional.case.shell
+    - match: \besac\b
+      scope: keyword.control.conditional.end.shell
+      pop: true
+    - match: (?=\()
+      push:
+        - clear_scopes: 1  # remove meta.conditional.case.shell
+        - match: \(
+          scope: keyword.control.conditional.patterns.begin.shell
+          set: case-clause-patterns
+    - match: (?=\S)
+      push: case-clause-patterns
+
+  case-clause-patterns:
+    - clear_scopes: 1  # remove meta.conditional.case.shell
+    - meta_scope: meta.conditional.case.clause.patterns.shell
+    - match: \)
+      scope: keyword.control.conditional.patterns.end.shell
+      set: case-clause-commands
+    # emergency pop off if ')' is missing
+    - match: (?=\besac\b|;;&?|;&)
+      pop: true
+    - include: case-clause-patterns-body
 
   case-clause-patterns-body:
     # [Bash] 3.2.4.2: Each pattern undergoes tilde expansion, parameter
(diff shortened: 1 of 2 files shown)


ShellScript/test/syntax_test_bash.sh
```diff
@@ -1675,93 +1675,112 @@ f() {
     done
 }
 
-case "${foo}" in
-# <- keyword.control.case.begin
-#             ^^ keyword.control.case.in
+case "${foo}" in- in_ in=10 in
```
rwols: Nice job here.

ShellScript/Bash.sublime-syntax
```diff
@@ -157,50 +157,50 @@ contexts:
...
+        1: keyword.control.conditional.if.shell
         2: keyword.operator.logical.shell
     - match: \bthen\b
-      scope: keyword.control.if.then.shell
+      scope: keyword.control.conditional.then.shell
       pop: true
     - match: \b(elif){{keyword_boundary_end}}(?:\s*(!))?
       captures:
-        1: keyword.control.if.elif.shell
+        1: keyword.control.conditional.elif.shell
```
FichteFoll: If we're standardizing, this should probably be `elseif`. (https://github.com/sublimehq/Packages/issues/1228#issuecomment-337072664)

ShellScript/Bash.sublime-syntax
```diff
@@ -157,50 +157,50 @@ contexts:
...
-      scope: keyword.control.for.shell
+      scope: keyword.control.loop.for.shell
       set: [cmd-post, for-args]
     - match: \bdo{{keyword_boundary_end}}
-      scope: keyword.control.do.shell
+      scope: keyword.control.loop.do.shell
       pop: true
     - match: \bdone{{keyword_boundary_end}}
-      scope: keyword.control.done.shell
+      scope: keyword.control.loop.done.shell
```
FichteFoll: Hm, I never though about closing keywords of loop structures in languages that don't use significant whitespace or curlies.

What do you think about `keyword.control.loop.end`?

ShellScript/Bash.sublime-syntax
```diff
@@ -64,7 +64,8 @@ variables:
   is_path_component: (?=[^\s/]*/)
   is_start_of_arguments: '[`=|&;()<>\s]'
   is_variable: (?=\s*{{nbc}}(?:[({]{{nbc}}[)}])?{{nbc}}=)
-  keyword_boundary_end: (?!=)(?=[^\w_-]|$)
+  keyword_boundary_char: '[^-=\w]'
+  keyword_boundary_end: (?={{keyword_boundary_char}})
```
FichteFoll: Removing `$` from the match will cause issues at the end of the file where there is no newline character that could be matched by a negative set.

deathaxe: Good catch, but can't add the `$` as it breaks other things. A negative lookahead does the same job though.

deathaxe:
The recent commits to fix the `!` were added as they are related with control structures and would have caused conflicts with this PR if they would have been pushed as separate PR.

At this point only changes requested by review of existing commits may be added.

ShellScript/Bash.sublime-syntax
```diff
@@ -1332,12 +1333,11 @@ contexts:
...
 
-  maybe-operator-logical:
-    - match: \!
+  operator-exclamation:
+    - match: \!(?=\s)
       scope: keyword.operator.logical.shell
-      pop: true
-    - match: $|(?=\S)
-      pop: true
+    - match: \!(?=\S)
```
FichteFoll: Maybe match the entire token as something in `variable` or `support`? This is only really relevant for an interactive shell anyway (with e.g. Terminus), but we should still try our best there, imo.

ShellScript/Bash.sublime-syntax
```diff
@@ -1350,6 +1333,12 @@ contexts:
     - match: ==?|!=?|<|>|\|\||&&
       scope: keyword.operator.logical.shell
 
+  operator-exclamation:
+    - match: \!(?=\s)
```
FichteFoll:
```suggestion
    - match: \!(?!\S)
```

ShellScript/Bash.sublime-syntax
deathaxe: You mean including the history command like  `!command`?

deathaxe: ... not sure whether it works well as `!` can be followed by expression.

e.g.: `![[ $1 == '0' ]] && echo -e "$2 PASSED" || echo -e "$2 FAILED";`

deathaxe: We could enclose everything in a `meta.history.shell`?

```
  operator-exclamation:
    - match: \!(?!\S)
      scope: keyword.operator.logical.shell
    - match: \!
      scope: punctuation.definition.history.shell
      push:
        - - meta_scope: meta.history.shell
          - match: ''
            pop: true
        - cmd
```

FichteFoll: Tbh, I've never used this feature and shallow description from the net didn't exactly enlighten me. I know there are `!!` and `!<n>` short forms, but that's about it. If you don't have an idea to implement this correctly, just leave it as-is, imo.

deathaxe: `!!` is what I didn't find in http://tldp.org/LDP/abs/html/special-chars.html#NOTREF

Both of your examples are references to the history index and could therefore be handled in a dedicated rule.

michaelblyons:
Today I learned about `!cmd` history shortcuts because of the difference in syntax highlighting between `!git` and `! git` added in this PR. Thanks!

wbond:
Looks reasonable to me - thanks!

PR #1864, opened 2019-02-03
[Perl] Scope sub as storage.type.function

deathaxe:
The keyword `sub` is used to identify the definition of a subroutine. It can be compared to `def` in python or `function` in Bash.

It does not describe a control-structure nor does it a storage.type.

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -945,7 +945,7 @@ contexts:
 
   sub:
     - match: \bsub\b
-      scope: keyword.control.sub.perl
+      scope: keyword.declaration.sub.perl
       push: sub-name
 
   sub-name:
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -945,7 +945,7 @@ contexts:
 
   sub:
     - match: \bsub\b
-      scope: keyword.declaration.sub.perl
+      scope: keyword.declaration.function.perl
       push: sub-name
 
   sub-name:
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -945,7 +945,7 @@ contexts:
 
   sub:
     - match: \bsub\b
-      scope: keyword.declaration.function.perl
+      scope: storage.type.function.perl
       push: sub-name
 
   sub-name:
--- a/Perl/syntax_test_perl.pl
+++ b/Perl/syntax_test_perl.pl
@@ -1141,38 +1141,38 @@ no strict;
 ### [ SUB ]###################################################################
 
 sub
-# <- meta.function.perl keyword.declaration.function.perl
-#^^ meta.function.perl keyword.declaration.function.perl
+# <- meta.function.perl storage.type.function.perl
+#^^ meta.function.perl storage.type.function.perl
 #  ^ - invalid
 sub name
-# <- meta.function.perl keyword.declaration.function.perl
+# <- meta.function.perl storage.type.function.perl
 #^^^^^^^ meta.function.perl
-#^^ keyword.declaration.function.perl
+#^^ storage.type.function.perl
 #   ^^^^ entity.name.function.perl
 #       ^ - invalid
 sub name invalid
-# <- meta.function.perl keyword.declaration.function.perl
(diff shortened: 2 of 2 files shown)


Perl/Perl.sublime-syntax
```diff
@@ -945,7 +945,7 @@ contexts:
 
   sub:
     - match: \bsub\b
-      scope: keyword.control.sub.perl
+      scope: keyword.declaration.sub.perl
```
FichteFoll: It it's comparable to a function, shouldn't it be scoped `keyword.declaration.function` in the general sense?

wbond:
As I understand it, `sub` denotes the following a subroutine (aka function) definition, correct? If so, for consistency with other syntaxes, it should be `storage.type.function`.

The other usage of `keyword.declaration` are primarily things like `extends`, `implements` and so forth. Never is it used for the keyword that denoted the following identifier is a function.

deathaxe:
Sounds reasonable. Seems as I got confused a bit by the recent discussion about scoping this kind of general function definition keywords around the _Go_ syntax. A closer look at my "default syntax scope list" even proves `storage.type.function` to be used for most of such keywords.

That said, should Go use the same? It currently uses ...

https://github.com/sublimehq/Packages/blob/1bff674a9d17c42e43cb305b350c66c8ff514207/Go/Go.sublime-syntax#L585-L588

wbond:
Yes, Nelo really wanted to be able to scope those keywords differently than other types, so he felt strongly about making it possible to target `storage.type.keyword` in his color schemes. I'm not quite sure if that should be standardized yet. It tries to split the difference between making the scopes easily targetable and mostly backwards compatible.

In the next short while I am hoping to go through the various RFC issues and summarize my understanding of each so I may add them to the scope naming documentation.

deathaxe:
I personally don't like `storage.type.keyword`. It doesn't say anything.

wbond:
Surely, but it means that:

 1. They keyword mostly likely get the same color as `storage.type.function` in most cases (via `storage.type`), thus backwards compatibility
 2. It allows targeting `def`, `class`, `function`, etc with a single selector instead of having to enumerate `storage.type.function, storage.type.class, storage.type.enum, storage.type.impl`. This allows color schemes to treat `int` and `func` differently.

The alternative is to do something like `storage.type.function keyword.declaration` to preserve backwards compatibility and allow a path forward. This is why I said I don't think it should be standardized yet. I think more input is needed to try and get consensus.

deathaxe:
Guess you missunderstood me. I am very fine with `storage.type.function`, `storage.type.class`, etc. .

I just feel a bit confused about `storage.type.keyword.function` _(Go)_ if it seems clear to use `storage.type.function` for general purpose function definition/declaration keywords.

Other examples are

```
storage.type.keyword.chan.go
storage.type.keyword.const.go
storage.type.keyword.function.go
storage.type.keyword.interface.go
storage.type.keyword.map.go
storage.type.keyword.struct.go
storage.type.keyword.type.go
storage.type.keyword.var.go
```

`chan` and `const` are `storage.modifiers`.

`interface`, `map`, ` struct` are just `storage.type.<interface|map|struct`.

No need to inject `keyword`.

FichteFoll:
I can't find the issue where we talked about keywords for function and class definitions right now. It would be nice if someone could cross-reference this PR there (and add the <kbd>RFC</kbd> label).

wbond:
> No need to inject `keyword`.

The issue is that to apply a color to all of these *keywords*, you have to enumerate ever term used in every language. If they are all under `storage.type.keyword.*`, then you can target `storage.type.keyword`.

> `chan` and `const` are `storage.modifiers`.

Yes, I noticed yesterday when looking at Go that we may need to do a little cleanup.

> It would be nice if someone could cross-reference this PR there (and add the RFC label).

I will be working on RFC stuff next, so I'll probably create an RFC issue for this topic and we can try to come to consensus and add to the scoping naming docs.

deathaxe:
@FichteFoll : I maybe on a completely wrong track or my mind is mixing things up, but I think it was discussed in the Go PR somewhere. But it was a bit contradictionary and may not have come to a final conclusion. So please be forgiving if I got confused somehow.

Issue #1866, opened 2019-02-04
[C#] string placeholder with `\` in formatting


michaelblyons:
See [the "Using the Escape Character" section][1] of the C# Date and Time format strings documentation.

``` csharp
DateTime date = new DateTime(2009, 06, 15, 13, 45, 30, 90);
string fmt1 = "h \\h m \\m";
string fmt2 = @"h \h m \m";

Console.WriteLine("{0} ({1}) -> {2}", date, fmt1, date.ToString(fmt1));
Console.WriteLine("{0} ({1}) -> {2}", date, fmt2, date.ToString(fmt2));
// The example displays the following output:
//       6/15/2009 1:45:30 PM (h \h m \m) -> 1 h 45 m
//       6/15/2009 1:45:30 PM (h \h m \m) -> 1 h 45 m
```

This is compilable C# and should not have any invalid characters:

``` csharp
string.Format(@"GMT is {0:yyyyMMdd\THHmmss\Z}", DateTime.Now.ToUniversalTime());
```

In the trivial case, you can replace this with 

``` csharp
string.Format("GMT is {0:yyyyMMdd\\THHmmss\\Z}", DateTime.Now.ToUniversalTime());
```

but if other content necessitates the raw-string modifier, you're stuck.

[1]: https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#escape

PR #1867, opened 2019-02-05
[C#] correctly handle escapes in long string placeholders

keith-hall:
Fixes https://github.com/sublimehq/Packages/issues/1866

To get this working, I had to duplicate and tweak the `string_placeholder_format` context so that it behaves correctly for normal strings as well as "long strings" and "long format strings". It sure would be great for maintenance if contexts could take parameters and then somehow the contents of that context could use those parameters to tweak the regex patterns ;)

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1558,11 +1558,16 @@ contexts:
   string:
     - meta_include_prototype: false
     - meta_scope: string.quoted.double.cs
-    - include: escaped
+    - include: string_escaped
     - match: '"'
       scope: punctuation.definition.string.end.cs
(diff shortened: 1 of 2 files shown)


michaelblyons:
>  It sure would be great for maintenance if contexts could take parameters and then somehow the contents of that context could use those parameters to tweak the regex patterns ;)

+1: I can't seem to get #1248 working without near-duplicated rules.

wbond:
@keith-hall So perhaps something like the idea that you can define a abstract context and then provide variables to realize it in multiple places?

keith-hall:
yes, something like that could be a big help :)

PR #1869, opened 2019-02-07
[R] Apply RFCs

deathaxe:
This PR introduces the following changes:

1. Apply the `keyword` scopes discussed in #1228.
2. Fix meta scopes of function definitions and function-calls to make them comply with the majority of the other syntaxes.
3. Fixes some operators not matched correctly (scopes, ligatures)
4. Applies the `keyword.operator.logical` scope to comparison operators.


R/R.sublime-syntax
```diff
@@ -83,10 +98,11 @@ contexts:
       scope: punctuation.terminator.semicolon.r
 
   storage-types:
-    - match: \b(character|complex|double|expression|integer|list|logical|numeric|single|raw)\b(\s*\()
+    - match: \b(character|complex|double|expression|integer|list|logical|numeric|single|raw)\s*(?=\()
+      scope: meta.function-call.r
       captures:
         1: storage.type.r
```
FichteFoll: Don't we need `meta.function-call.name` here as well? At least that's how I understood your RFC.

Same for all similar occurances later in the file (`builtin-fuctions` context).

deathaxe: We can of course add it, of course. May even be better for consistency. Just didn't add it here as I did not expect sophisticated expressions to be valid for function name construction in R - but I am not a R expert though.

FichteFoll: I think it just makes sense to apply the same scopes for all syntaxes, even if they don't have a particular construct that would require it. I mean, unification of scope names is our ultimate goal.

randy3k:
@deathaxe 
Didn't know that you are also an R user.

Great job.

R/syntax_test_r.R
```diff
@@ -208,7 +232,7 @@ foo[[bar[1]]] #
 # issue #1120
 sum(x == 1)
 #   ^^^^^^ - variable.parameter.r
-#     ^^ meta.function-call.r meta.function-call.parameters.r keyword.operator.assignment.r
+#     ^^ keyword.operator.assignment.r
```
randy3k: This should be  keyword.operator.comparison.r`

randy3k: https://github.com/sublimehq/Packages/blob/16c9d826244cc265857787d1fb1bd48579837f12/R/R.sublime-syntax#L85-L88

these two scopes should be swapped.

deathaxe:
Not really a R power user. Just try to apply some recently discussed scope names around keywords etc. to some syntaxes to clean up the scoping landscape with the goal to enable ST to present most/all the syntaxes in the same way.

R/syntax_test_r.R
deathaxe: Good catch. Just removed the meta scopes from the test case without revealing this bug.

deathaxe: Added a commit to fix that and rename the operator scopes according to other syntaxes.

deathaxe:
Enough for this PR.

wbond:
Thank you @deathaxe!

PR #1876, opened 2019-02-09
[Git] Add .mailmap highlighting

michaelblyons:
It's very minimal, but names and emails are indexed


Git Formats/Mailmap.sublime-syntax
```diff
@@ -0,0 +1,18 @@
...
+file_extensions:
+  - .mailmap
+  - mailmap
+scope: text.git.mailmap
+
+contexts:
+  main:
+    - include: Git Common.sublime-syntax#comments
+    - include: Git Common.sublime-syntax#email
+    - match: '[ \t]*(\b[^<#\n]+?)[ \t]*(?=<|$)'
```
FichteFoll: Matching for `[ \t]*` at the start is *probably* redundant? Idk how greedy the other includes are.

Git Formats/Mailmap.sublime-syntax
```diff
@@ -0,0 +1,18 @@
+%YAML 1.2
+---
+# Syntax based on documentation here:
+# https://www.git-scm.com/docs/git-check-mailmap#_mapping_authors
+
+name: Mailmap
```
deathaxe: Both the file name and the name here should start with "Git" as the other files.

- Git Mailmap.sublime-syntax
- `name: Git Mailmap`

Syntaxes are addressed via file name only. Hence the name should be meaningful enough to avoid conflicts with other packages.

Git Formats/Mailmap.sublime-syntax
michaelblyons: You are right. I had that in there before adding the `\b`. I'll take it out.

I am _hoping_ that the `\b` is okay to have here. I haven't looked at valid *Nix and Windows user names. Of course there may be some goofball who decides they want to be known as `%le3t` or something. -_-;

Git Formats/Mailmap.sublime-syntax
michaelblyons: I'm not strictly opposed, but I don't think it's necessary.

- Searching DDG for `mailmap` returns lots of entries for Git-related stuff.
- Searching DDG for `mailmap -git` _still_ returns mostly Git-related stuff.
- There are no packages named Mailmap.
- There are [hardly any packages](https://packagecontrol.io/search/mail) with "mail" anywhere in the name

deathaxe: All kind of files in this package start with Git. Wouldn't break that guideline.

Git Formats/Mailmap.sublime-syntax
michaelblyons: Fixed it for goofballs, too. 😄

deathaxe:
Now that we removed the escape from the email pattern, we should probably apply that to the Git Common.sublime-syntax as well?

michaelblyons:
> Now that we removed the escape from the email pattern, we should probably apply that to the Git Common.sublime-syntax as well?

Probably so.

Also, I am tempted to remove the special `punctuation` for `.` in the domain. Right now it only highlights the first instance, but I don't really think highlighting all instances (or just the last one) is really that helpful either.

```
# current
foo@github.com
          ^ punctuation
foo@users.noreply.github.com
         ^ punctuation
                 ^      ^ - punctuation

# highlight-all; Maybe useful? idk
foo@github.com
          ^ punctuation
foo@users.noreply.github.com
         ^       ^      ^ punctuation

# highlight-last; Who cares?
foo@github.com
          ^ punctuation
foo@users.noreply.github.com
                        ^ punctuation
```

I think the most __*useful*__ place to highlight `.` is the place just to the left of the [public suffix](https://publicsuffix.org/list/), but it's a bit of a nuisance to maintain a huge list like that.

```
# current
foo@github.com
          ^ - punctuation
foo@users.noreply.github.com
                 ^ punctuation
```

For an "email address," the most important delimiter is the `@`.

michaelblyons:
Thank you, @deathaxe!

wbond:
I'm inclined to think `punctuation.separator.email` would be applied to `@` and `punctuation.separator.path` or `punctuation.separator.domain` be applied to the `.`.

michaelblyons:
@wbond Updated. Anything else?

wbond:
I'm not quite sure I understand showing names and email in the symbol list. Wouldn't users normally just use Find?

Everything else looks fine to me.

michaelblyons:
> I'm not quite sure I understand showing names and email in the symbol list. Wouldn't users normally just use Find?

Good question. I'm usually in a Mailmap when I'm trying to add someone for whom I have a stray ID, and I don't always know what canonical one they've been using.

The ST fuzzy-finder is then helpful for (e.g.) "gwenzek" which will pick up both "gwenzek" and "Guillame Wenzek" and any other near-matches.

deathaxe:
I agree with @michaelblyons. Even though I don't use the .mailmap, I find the sublime-settings symbol list very helpful, which is added by PackageDev. It's sometimes faster to fuzzy search, then anything else. Maybe a habit thing, though.

wbond:
Thanks for adding this @michaelblyons!

FichteFoll:
Also, to be clear on the wording: Emails and names aren't "indexed" (i.e. added to the project-wide symbol list) but only added to the local list for navigation. That is perfectly for fine for formats like these imo, like PackageDev does for settings files as @deathaxe mentioned.

Issue #1861, opened 2019-02-02
[RFC] Definitions and References


deathaxe:
## Motivation

The primary motivation of this RFC is to define a common guideline of how syntax definitions should scope entities in order to work well with Sublime Text's _Goto Definition_ and _Goto Reference_ features using a minimal set of rules and to provide a starting point to extend the feature syntax specifically.

Sublime Text's _Goto Definition_ and _Goto Reference_ requires scoping tokens of _definitions_ or _declarations_ and pairing them with those looking like a _reference_ of the entity.

Pairing works by adding those tokens to Sublime Text's index database.

The following code blocks show the most basic pair, which is defined by Sublime Text out of the box.

**Definitions**

```XML
<dict>
    <key>scope</key>
    <string>
        meta.function entity.name.function
    </string>
    <key>settings</key>
    <dict>
        <key>showInIndexedSymbolList</key>
        <integer>1</integer>
    </dict>
</dict>
</plist>
```

**References**

```XML
<dict>
    <key>scope</key>
    <string>
        meta.function-call variable.function,
        meta.function-call support.function
    </string>
    <key>settings</key>
    <dict>
        <key>showInIndexedReferenceList</key>
        <integer>1</integer>
    </dict>
</dict>
</plist>
```

_Note: The disussion was initiated at https://github.com/sublimehq/Packages/pull/1841#issuecomment-455788658_

## Issues

The function identifier can be one of the following atomic or complex expresions:

type               | scope
-------------------|----------------------
_identifier_       | `variable.function`
_builtin function_ | `support.function`
_macro_            | `constant.other.macro`
_builtin macro_    | `support.constant.macro`
_expression_       | `...`

**Examples with macros**

C

```C
#define FUNC_NAME function

//// FUNCTION DEFINITION //////////////////////////////////

void FUNC_NAME(int arg1, int *arg2) { }

// equal to

void function(int arg1, int *arg2) { }

//// FUNCTION CALL ////////////////////////////////////////

int main(int argc, int *argv) {
    
    return FUNC_NAME(argc, argv)
}
```

Erlang

```Erlang
-define(FUNC_NAME, function).

%%%% FUNCTION DEFINITION %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

?FUNC_NAME(arg1, arg2) -> .

% equal to

function(arg1, arg2) -> .

[...]

Issue #1842, opened 2019-01-16
[RFC] Scopes for module/namespace access


deathaxe:
#### Intro

In general a trend/principle in syntax definitions can be found which ends up with scoping the declaration/definition of constructs with `entity.name.<construct>`. When calling/using such constructs, something like `variable.<construct>` or `variable.other.<construct>` is used. 

The most popular example is `entity.name.function` vs. `variable.function`.

The goal is clear - distinguish definition and usage of the same object.

#### Question

How are `namespaces` or `modules` to handle in that manner?

A couple of syntaxes including C, C++, Python, PHP, Java, JavaScript, Erlang, ... support such concepts. Most of them use `entity.name.namespace` to scope the identifier in the definition/declaration statement as the ST3 documentation at https://www.sublimetext.com/docs/3/scope_naming.html says:

> Namespaces, packages and modules use the following scope. There are usually not multiple types of such constructs in a language, so this scope should suffice.
>
>    entity.name.namespace

But I can't find a common solution how to scope a namespace/module upon usage.

```
    module_name:function_name()
```

- *C#* distinguishes between `support.namespace.cs` and `variable.other.namespace.cs`
- *CSS* uses `entity.other.namespace-prefix.css` or `entity.name.namespace.wildcard.css`
- *Erlang* uses `entity.name.type.class.module.erlang`
- *Python* just scopes them as `meta.generic-name` upon usage or `meta.import-name` in import statements.
- ...

Can we find a common scope for that usage?

1. From my point of view anything starting with `entity.` is a no-go when we talk about usage.
2. The most pleasant approach with respect of existing scoping guidelines and implementations seems to be `variable.other.namespace`. So we'd end up in

    `entity.name.namespace` vs. `variable.other.namespace`

3. To keep up with the concept of function declaration and usage, I also could imagine to use `variable.namespace`. So we'd end up in

    `entity.name.namespace` vs. `variable.namespace`

Thoughts?

keith-hall:
I completely agree with point 1.
for namespace usage, I think its worth having the distinction between namespaces provided in the standard library and those that are user defined (i.e. like the C# syntax definition does atm), as I like to color them differently.

FichteFoll:
For method or static function calls on objects/classes, do we treat those as namespaces as well? There is no way to differentiate those from modules in most languages.

deathaxe:
I did not directly think of static class member access, but this is a very good question though.

[...]

PR #1878, opened 2019-02-10
[Erlang] Rewrite Syntax

### LINKED ISSUE #1609: [Erlang] Syntax highlighting does not work when using Macros with paramters
Sublime build 3143

When using macros with parameters (?macroname(par1, par2)), after a while syntax highlighting stops working.

The scope piles on after every statement, e.g.
image

with scope at cursor as:

source.erlang meta.function.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang meta.structure.list.erlang meta.structure.tuple.erlang meta.structure.list.erlang meta.expression.fun.erlang 

It seems the capture groups for macro-usage does not take parameters into account:

https://github.com/sublimehq/Packages/blob/1cb4c3ec368c751d6f7ecfa16fe02ceff23a1f6b/Erlang/Erlang.sublime-syntax#L441

### LINKED ISSUE #1724: [Erlang] incorrect syntax highlighting with first class functions
The picture below shows that the erlang highlighting is being used and that it is capable of highlighting functions and makes 'valid' green:

valid

However when we create another function that has a higher order function as an argument, F, and then call F in that function, the syntax highlighter no longer highlights 'valid' as green.

invalid

The code compiles and works as expected, where foo returns a new function that calls F so I believe my syntax is correct but please correct me if not as I haven't delved too deep into the languages semantics.

deathaxe:
Fixes #1609
Fixes #1724

According to #481 this PR provides an Oniguruma free Erlang syntax definition written from scratch.

## General Notes

The syntax definition is based on the [Erlang Specification](http://erlang.org/doc) and includes all sematic features of Erlang/OTP 21.2.

It follows [Sublime Text Syntax Guidelines](https://www.sublimetext.com/docs/3/syntax.html) and includes concepts from [sublimehq/Packages](https://github.com/sublimehq/Packages)

Nearly every aspect and detail of the syntax definition was designed from ground up making use of the latest features of ST's syntax engine and with the intent to apply the most recent best practice examples found in other recently updated syntax definitions.

## Changes

* Add support for Erlang ShellScript (escript)
* Add support for Erlang Type Language (`-type`, `-spec` directive)
* Add lots of detailed test cases
* Add some invalid illegal matching where appropriate
* Add builtin constants and macros highlighting
* Improve the list of builtin functions ([BIFs](http://erlang.org/doc/man/erlang.html#exports))
* Improve indention settings
* Improve shebang/editorconfig detection in `first_line_match`
* Improve support for _Goto Definition_ and _Goto Reference_
* Improve parsing performance by round about 50%
* Fix quoted atom highlighting (function names, module names, field names, ...)
* Fix bitstring data type highlighting
* Fix string escape highlighting
* Fix string placeholders highlighting
* Fix identifier break handling (`@` is no word-break)
* Fix lots of context pop offs
* ...

## Code Examples

This chapter tries to give an impression of the changes in pictures. Please note that some of the scope naming changes may look some kind of broken compared to the old syntax and others may not bee too obvious anyway.

#### Preprocessor

```Erlang
%% Erlang test module
-module('my-test-module').
-compile(export_all).

-ifndef(HAVE_DEBUG).
- define(HAVE_DEBUG, 1).
-endif.

-include_lib("common_test/include/ct.hrl").

-export([parse_result/2]).

-on_load(run_loader/0).

-type my_data() :: >=10 | millisecond | integer() | other.

-record('rec-name', {
    status   = undefined  :: integer(),
    name     = "foo"      :: string() | atom() | foo,
    code     = []         :: list(),
    callback = none       :: fun((arg1 :: binary(), atom()) -> int())
    }).
```

#### Functions

```Erlang
-spec parse_result(binary(), resulthandler(T)) -> resulthandler_result(T).

parse_result(<<ResultCode:32/little-unsigned, Rest/binary>>, ResultHandler) ->

  % anonymous functions
  Fun = fun
      % first clause
      (Arg1,Arg2) when Arg1 > 0, is_binary(Arg1); is_reference(Arg2) ->
          % nested anonyomous functions
          Fun1 = fun ?CUSTOM:name/?ARI,
          Fun3 = fun Fact(1) -> 1; Fact(X) when X > 1 -> X * Fact(X - 1) end,

          case Arg1 of
            0 -> ResultHandler(Arg2);
            1 when is_integer(Arg1); Arg2==0 -> no_magic;
            _ -> erlang:throw()
          end;
      % second clause
      (Arg1,Arg2) ->
        error
  end,
  % call anonymous function
  Fun(ResultCode, Rest),
```

#### Binaries

```Erlang
  Bin  = << << (X*2) >> || <<X>> <= << 1,2,3 >> >>.
```

#### Placeholders

```Erlang
  Fmt1 = "bin: ~B ~.16B ~4.16B ~4.16.#B ~# ~.10# ~+ ~.10+ ~X ~.16X ~-5.16X ~-5.16. X",
  Fmt2 = "str: ~p ~tp ~lp ~tlp ~ltp ~15p ~15lp ~15tp ~15tlp ~15ltp",
  Fmt3 = "flt: ~-*.*.*f",
```

#### Records

```Erlang
  Expr#record{field1="val1", Field2=3, field3, 'Field-4'={}, _=atom},
  N0n = N2#nrec2.nrec1#nrec1.nrec0#nrec0{name = "nested0a"},
```

### ST 3189

grafik

### New Syntax

grafik

## Benchmarks

The whole design and developement process included benchmarking of different solutions to find the best possible performance for the different sets of rules. The following examples provide an impression of the performance improvements achieved with the new design.

**10k binary**

```Erlang
  Bin  = << << (X*2) >> || <<X>> <= << 1,2,3 >> >>.
```

**10k maps**

```Erlang
  Expr#{name=>"adam",{age,24}:=fct(),4.0=>{july,29}}
```

**10k records**

```Erlang
  #record{field1="val1", Field2=3, field3, 'Field-4'={}, _=atom}
```

**10k fun types**

```Erlang
  -type fun() :: fun(() -> int()).
```

**10k spec**

```Erlang
  -spec is_foo(Pid) -> boolean() when Pid :: pid() | if(); (Pid) -> ok.
```

**ErlangOTP files**

[lib\wx\include\gl.hrl](https://github.com/erlang/otp/blob/master/lib/wx/include/gl.hrl)

[lib\stdlib\test\re_testoutput1_split_test.erl](https://github.com/erlang/otp/blob/master/lib/stdlib/test/re_testoutput1_split_test.erl)

[lib\xmerl\test\xmerl_sax_std_SUITE.erl](https://github.com/erlang/otp/blob/master/lib/xmerl/test/xmerl_sax_std_SUITE.erl)

[lib\crypto\test\crypto_SUITE.erl](https://github.com/erlang/otp/blob/master/lib/crypto/test/crypto_SUITE.erl)

#### Results

Type                           | before     | after       | difference
-------------------------------|------------|-------------|------------
10k binaries                   |     303ms  |      140ms  |   -53.8%
10k maps                       |     326ms  |      156ms  |   -52.1%
10k records                    |     399ms  |      130ms  |   -67.4%
10k fun type                   |      n.a.  |       79ms  |
10k spec                       |      n.a.  |      176ms  |
gl.hrl                         |      47ms  |       24ms  |   -48.9% 
re_testoutput1_split_test.erl  |     448ms  |      305ms  |   -31.9%
xmerl_sax_std_SUITE.erl        |     389ms  |      127ms  |   -67.4%
crypto_SUITE.erl               |      79ms  |       27ms  |   -65.8%


deathaxe:
@StevieSong, @tiaanvz: This PR is dedicated to all our network backbone engineers and may need some review from those experienced with Erlang.

wbond:
@mroach I think I've noticed you doing a bit with Erlang recently. Perhaps you'd be interested in trying this out?

mroach:
@wbond I've been working with Elixir, but I've had cause to read Erlang packages used by Elixir. So I'll install this and give it a go.

eproxus:
I'll start reviewing this right away. Thanks @deathaxe!

eproxus:
@deathaxe I have a couple of issues:

1. Variables change color in different situations:
   1. Variables are colored differently when they're arguments:
      Screen Shot 2019-05-07 at 11 07 24
   2. Variables are colored differently when they're called:
      Screen Shot 2019-05-07 at 11 11 09
   3. Variable are colored differently when they're assigned different values:
      Screen Shot 2019-05-07 at 11 17 44
4. Records and fields are colored differently when they're nested:
   Screen Shot 2019-05-07 at 11 09 26
5. Macros are colored differently when they're defined and when they're used:
   Screen Shot 2019-05-07 at 11 10 02

Basically, my problems boil down to not being able to use color to identify the type of entity I'm looking at.

Not sure how many of these are because of the color scheme. Realistically, could they be fixed by a color scheme? Not every scheme wants language specific adaptations. How to deal with this in Sublime Text in general?

eproxus:
The specific problem with 1.iii seems to be that the `Fun` variable in the first line doesn't even get the scope `variable.other.erlang` at all, thus the color scheme doesn't know it's a variable at all.

eproxus:
Indexing seems to be problematic as well. It can't find function definitions in the module at all.

FichteFoll:
Identifiers being scoped (and usually colored) differently when they are defined and used is the intended behavior and its details are thoroughly discussed in #1842, if you are curious.

eproxus:
@FichteFoll Are you referring to 1.iii? The variable (which it is, in the language) gets the `entity.name.function.erlang` scope which should _not_ have in my opinion. In Erlang a variable is _never_ a function name (it can be bound to one though).

deathaxe:
Thanks for taking the time to help us.

> Not every scheme wants language specific adaptations.

This is actually one of the main goals of all the recent changes made to the builtin syntaxes of ST. The community tries to make all syntaxes share as much common scopes for compareable expressions/statements as possible to enable the same user experience for all syntaxes.

Due to heavy conceptual differences of all the syntaxes this effort requires some "abstractions", which might look inaccurate in some situations though.

Example:

What _Erlang_ calls a `module` is known as `package` in Perl and maybe compared to a `namespace` in C. These kinds of comparisons are required to find common scope names resulting in common colors without syntax-specific color-scheme-extensions.

> Variables change color in different situations

To be able to drive the _Goto Definition_ vs. _Goto Reference_ feature of ST, we need to distinguish between the definition of a variable and the usage. The most common example is how function names are scoped. ST uses `entity.name.function` to name the definition while `variable.function` is used to scope the usage (function-call). Both may be applied different colors to.

This concept is tried to be applied to other situations as well.

>i. Variables are colored differently when they're arguments:

The formal argumets are scoped as `variable.parameter`, while the actual arguments are scoped `variable.other`. This is how most other default syntaxes scope them like in order to distiush definition and usage. 

_Notes:_

1. With regards to functions `variable.parameter` would probably need to be replaced by `entity.name.parameter`. 
2. If a color-scheme would address `variable` only, both would be colored the same way.

> iii. Variable are colored differently when they're assigned different values:

This is a good point, we might discuss about. I scoped the `Fun` as `entity.name.function` as the documentation calls such statements an "anonymous function declaration". I intended to include them into the _Goto Definition_ feature, but found them to bring too much noise in the end.

*I am fine to turn it into an ordinary `variable.other`.*

> 2. Records and fields are colored differently when they're nested:

I am not an Erlang expert, so please excuse the stupid question(s):

According 10.2 of http://erlang.org/doc/reference_manual/records.html `#record{field_name=value}` describes the createion (local definition?) of a record? Hence the last `#nrec0` is the new name of a record with one field named `field_name` initialized with the `value`? 

Therefore the last `#nrec0` is assigned the `entity.name.record` so far, while the rest in front is understood as existing set of records defining the path?

*If this assumption is incorrect and the last `#nrec0` can be scoped as normal `variable.other.record`, things would even get simpler.*


> 3. Macros are colored differently when they're defined and when they're used:

Basically the same here:

In `-ifndef(MACRO)` the `MACRO` is "used". Hence it is scoped `constant.other.macro`. 
In `-define(MACRO, "something")` `MACRO` is defined. Hence scoped `entity.name.constant.macro`.

... to be able to distinguish them.


**Conclusion**

I also noticed the different colors, which were equl before, but tried to follow some of the contecpts discussed in some RFCs or found in other recently updated syntaxes. It's still hard to find a "best practice". This is why I opened some RFC issues like #1861 and #1842. But the discussions just proove how hard it is so far.

Maybe we don't even matter about definition/usage distinction in times of evolving `language-server-protocols` which are superior to ST's indexing capabilities by nature?

Not sure.

wbond:
> Maybe we don't even matter about definition/usage distinction in times of evolving language-server-protocols which are superior to ST's indexing capabilities by nature?

In general, we at Sublime HQ do care about this. The user experience for language servers tends to only work for people who are looking to invest significant time into configuring their environment, and even then the performance seems sub-par. Obviously we don't have a built-in LSP host, and there are some rough edges that will need to be polished to make it possible for anyone to write LSP host without some annoying edge cases.

We obviously won't ever get to 100% of the accuracy of a parser for a specific language, but for the most part I believe we can hit the 99% use case for our users.

I foresee the information found via the indexer to continue be useful to build general-purpose features off of, and foresee the .sublime-syntax definitions to only get more powerful over time.

eproxus:
@deathaxe 
> Thanks for taking the time to help us.

Thanks for replying. Have been wanting to make a new syntax for Erlang for long time now but never had the time. 😄 

> > i. Variables are colored differently when they're arguments:
> 
> The formal argumets are scoped as `variable.parameter`, while the actual arguments are scoped `variable.other`. This is how most other default syntaxes scope them like in order to distiush definition and usage.

Ah, I see. That would be up to the color scheme to differentiate or not.

> > iii. Variable are colored differently when they're assigned different values:
> 
> This is a good point, we might discuss about. I scoped the `Fun` as `entity.name.function` as the documentation calls such statements an "anonymous function declaration". I intended to include them into the _Goto Definition_ feature, but found them to bring too much noise in the end.
> 
> _I am fine to turn it into an ordinary `variable.other`._


In Erlang, functions created using `fun` are anonymous functions. They don't have names. Thus, any concept of _goto definition_ can't be applied to them. Once the function is created and bound to a variable, that's the only handle you have. And it is a dynamic handle, not a static one. Consider the following code:

```erlang
f() ->
    Var = fun() -> ok end,
    Foo = Var,
    Foo. % What happens when you run goto definition here?

x() ->
    Var = 1,
    Var. % What happens when you run goto definition here?
```

Including anonymous functions _Goto Definition_ makes no sense, I think. 🙂

Another problem I have with the current scope is that the variable isn't even scoped as a variable. Thus any plug-in or tool that uses scopes to work with variables would break.

> > 1. Records and fields are colored differently when they're nested:
> 
> I am not an Erlang expert, so please excuse the stupid question(s):
> 
> According 10.2 of http://erlang.org/doc/reference_manual/records.html `#record{field_name=value}` describes the createion (local definition?) of a record? Hence the last `#nrec0` is the new name of a record with one field named `field_name` initialized with the `value`?
> 
> Therefore the last `#nrec0` is assigned the `entity.name.record` so far, while the rest in front is understood as existing set of records defining the path?
> 
> _If this assumption is incorrect and the last `#nrec0` can be scoped as normal `variable.other.record`, things would even get simpler._

I think it is incorrect. Records are only defined via the `-record(name, {field}).` preprocessor directive. Then they are either created or referred to.

```erlang
-record(foo, {bar}).

f() ->
    NewFoo = #foo{bar = 1}, % created
    NewFoo#foo.bar. % referred to
```

> > 1. Macros are colored differently when they're defined and when they're used:
> 
> Basically the same here:
> 
> In `-ifndef(MACRO)` the `MACRO` is "used". Hence it is scoped `constant.other.macro`.
> In `-define(MACRO, "something")` `MACRO` is defined. Hence scoped `entity.name.constant.macro`.
> 
> ... to be able to distinguish them.

That's also fine, I think. I understand the difference better now. So for definitions having the scope `entity.name.constant.macro.erlang` and usage having the scope `constant.language.macro.erlang` one could still target the scope `macro` and get both?

Is there any concern that things that are two things only get assigned one thing? 😛 What I mean is, that e.g. record field names in Erlang are both record field names and atoms. But in the current syntax proposal there's no way to see that they are atoms (they only have the scope `entity.name.field.erlang`).

I also don't see a difference in scope where a record field is defined (`-record(foo, {my_field}).` and where it is used (`#foo{my_field = 1}`). Is that an issue?

Another question I have is, did you settle on the syntax in _Goto Reference_ for Erlang specifically or is there a standard here? I mean this:

```
@
< parse_result(...)
:: my_data(...)
#'rec_name'{...}
:: parse_result(...)
```

I have a custom modification of the old Erlang plug-in which gives this as reference list:

```
?HAVE_DEBUG
-export([parse_result/…])
my_data() ::
#'rec_name'{…}
parse_result(…) ->
```

I found it mapped better to what was in the actual file and easier to read.

@wbond 
> a built-in LSP host
😍 Would love for that to exist at some point!

deathaxe:
> Is there any concern that things that are two things only get assigned one thing?

Basically yes and it was actually one reason for the already mensioned RFCs.

Macros for instance: A macro can be used as identifier for a function to replace the actual name. So the question arose, whether to scope it as `constant.other.macro` or `variable.function`. How to express that the macro is used in this way?

Example:

```Erlang
    - define(FUNMACRO, theRealFun);

    ?FUNMACRO(X);

    FunVar = ?FUNMACRO;
    FunVar(Y);
```

Erlang currently solves that issue by scoping it `meta.function-call.name constant.other.macro`.


> But in the current syntax proposal there's no way to see that they are atoms (they only have the scope entity.name.field.erlang).

I actually removed `meta.atom.erlang` scopes as I was not sure whether it is useful to have them. Basically any "smallest word/identifier" in any language is an atom.

The punctuations for quoted identifiers are actually still scoped as `punctuation.definition.atom`. Would be easy to re-add the `meta.atom` scopes.

> I also don't see a difference in scope where a record field is defined (-record(foo, {my_field}). and where it is used (#foo{my_field = 1}). Is that an issue?

both braced blocks share the same context at the moment as I interpreted the creation of a record as local definition of it. Hence the `#foo{my_field = 1}` was interpreted as definition of a record with a new `my_field`. But as this expression just assignes a different value to the field, we can also change the scopes to normal "usage" as I did with the nested records in the recent commit.

Another question I have is, did you settle on the syntax in Goto Reference for Erlang specifically or is there a standard here? 

You mean the prefixes / symbol transformation?

1. Only the `entity.name. ...` is to be indexed and therefore is displayed in the goto definition quick panel, but not complete expressions like `-export([parse_result/…])`. As a result each function in the export list is displayed separately with the prefix `<`. The character is just a personal choice.

2. The syntax tries to distinguish the different kinds of objects in the Goto Definition quick panel by prefix to efficiently filter by type and to be able to clearly see what it is. Unfortunatelly the goto definition panel doesn't provide the right handed descriptions to do so.

The current suggestion:
   ```
   :: type or function specification
   < export
   > import
   # record
   Ξ function
   Ξ ?macro function   -> -define(?fun(), ...

   ? macro    -> ooops, is actually missing
   ```

eproxus:
Hmm, I think I have a problem with the base types _not_ being assigned first hand and some secondary analyzed "usage" being assigned instead. I think (and I think I think this for all syntax modes) that first and foremost the base tokens should be assign proper names (variable, function, constant, symbol etc.). Then, and only then, can some overlying meaning be assigned to them _if and only if_ it can be properly deduced.

What in my opinion cannot be properly done on a syntax parsing level is deducing things such as what a variable is going to be used for (as we saw with the `fun` assignment). In the same way, you could never know what a macro is going to be used for so assigning meaning such as `variable.function` is just not feasible. Macros in Erlang are parameterized string substations, they can be use for _anything_ really. No syntax parsing could realistically deduce what the programmer intended.

> The current suggestion:
> 
> ```
> :: type or function specification
> < export
> > import
> # record
> Ξ function
> Ξ ?macro function   -> -define(?fun(), ...
> 
> ? macro    -> ooops, is actually missing
> ```

Is the user intended to type e.g. `Ξ` to show functions? Depending on keyboard layout, that could become problematic. As I said above, I would prefer entries that look as similar as possible to the code in the file itself. At least, I find that much more logical.

deathaxe:
> Hmm, I think I have a problem with the base types not being assigned first hand and some secondary analyzed "usage" being assigned instead. 

Not really sure about what you mean. Is it about the scope names?

The scope names comply with the original [TextMate scoping guidelines](https://manual.macromates.com/en/language_grammars#naming_conventions), which are the base for Sublime Text's [scope naming guideline](https://www.sublimetext.com/docs/3/scope_naming.html). Hence you won't find top-level scopes like `function`.

> What in my opinion cannot be properly done on a syntax parsing level is deducing things such as what a variable is going to be used for

Agree.

> you could never know what a macro is going to be used for so assigning meaning such as `variable.function` is just not feasible

Agree. This is actually the reason for scoping function-calls using macros as identifier as `meta.function-call.name constant.other.macro`

```Erlang
?FUNNAME(arguments)
^^^^^^^^ meta.function-call.name constant.other.macro
``` 

It was scoped `variable.function.macro` in the old Erlang syntax, which I find wrong.

The distinction between macro constants and macro functions in the symbol index is handled by the style of definition.

A macro constant is `-define(name, "value")` while a macro function is `-define(name(args), {args})`

> Is the user intended to type e.g. Ξ to show functions?

Was just looking for a symbol to be used for functions. 

I would prefer a solution without any additions to the name, but with a description text like we have it in completions. The description could be defined in the *.tmPreference which is used to define the index.

```
 ?macro          (Constant)
 func1           (Function)
 func2           (Function)
 ...
```

> I would prefer entries that look as similar as possible to the code in the file itself

Basically aggree.

1. IMHO encapsulating imports/exports into `-import([ ])`, `-export([])` might cause too much noise?
2. Instead of using `(...)` to suffix imports/exports, we could use the arity. `< fun_name/4`.
3. The usage of records always starts with `#`. Why not use it to identifiy records in the symbol list?
4. Macros start with ? in the code, which they do in the symbol list as well.
5. We can argue about removing the  `Ξ`  from functions though.

deathaxe:
As a side note: You may need to remove the builtin Perl.sublime-package from ST's install dir as it removes the functions from the symbol list and may cause some quirks with this new syntax.

eproxus:
> > Hmm, I think I have a problem with the base types not being assigned first hand and some secondary analyzed "usage" being assigned instead.
> 
> Not really sure about what you mean. Is it about the scope names?

I was mainly referring to the `meta.function-call.name` scope.

> A macro constant is `-define(name, "value")` while a macro function is `-define(name(args), {args})`

Macros in Erlang are just macros. There no such thing as a constant (the macro contents is actually inlined everywhere at compile time) and there are no macro "functions". A macro without arguments would be equivalent to `?MACRO()` (although the syntax is not valid). Functions have arguments and can be called in runtime, macros are string substitution performed at compile time. A macro doesn't even have to contain valid Erlang syntax.

1. Yeah, maybe you are right
2. Agree
3. Agree
4. Then it's alright
5. It's just that it isn't easy to type on a normal keyboard. I think I'd prefer just `funcname(…)`

eproxus:
I found another case which seems to be parsed as a function:
```
-callback acceptor_init() -> {ok, {
    inet:socket_protocol(),
    inet:address_family(),
    {inet:ip_address(), inet:port_number()},
    state()}
}.
```

`-callback` is a module attribute, but the body of the attribute is parsed as a function definition in the syntax.

deathaxe:
> -callback is a module attribute, but the body of the attribute is parsed as a function definition in the syntax.

As far as I understood the docs, both the `-callback` and the `-spec` statements

1. are part of Erlang's _typespec_ extension 
2. define the signature of a function.

The syntax definition handles the whole _typespec_ stuff as a special kind of function definitions in order to reuse existing standard scopes which make the function signatures look like the function definitions by default. They can be addressed in special manner by `meta.preprocessor.spec` but should otherwise be recognized as function.

Both don't have a body, but only the _name_, _parameters_ and a _return-type_. The whole `-callback` expression is like a pattern (built of types) which is used to find the function to call by event. Therefore it can be interpreted as definition of a reference to the real function. Therefore (`definition` = `entity`) the name (`acceptor_init`) is scoped as `entity.name.function`, which makes it being colored like the function defininition by default.

We could consider using `variable` instead, but this wouldn't make sense with `-spec` statements and therefore required to duplicate all the contexts just to rename the function name.

eproxus:
As long as function type specs can be distinguished from actual function definitions I'm fine with it. I don't want tools or features broken because they can't distinguish between actual function and just types.

Trying out the latest version function definitions doesn't show up in the symbol list anymore. Neither does `-callback` definitions (but I don't know if they should?).

deathaxe:
> function definitions doesn't show up in the symbol list anymore

Did you remove the originial builtin (bundled) Erlang package? If not, its _Symbol Overrides.tmPreferences_ may cause it.

eproxus:
@deathaxe I replaced the the package with a symlink as per the instructions in the forum. I also see the new list of symbol prefixes from 8e8d9d9.

deathaxe:
Assuming portable windows setup:

If you create a symlink named _Erlang_ in [SublimeText]\Data\Packages which points to the [local_repo]\Erlang, and leave [SublimeText]\Packages\Erlang.sublime-package untouched, ST merges both packages. 

As the last commit removed the _Symbol Override.tmPreference_ from this repo, the original one applies. It removes `source.erlang entity.name.function` from from the symbol list.

Hence the only way to fix it is to remove _[SublimeText]\Packages\Erlang.sublime-package_

It took me a while to find out why the hell function definitions didn't get listed during developement.

eproxus:
Using macOS. I don't have a `Erlang.sublime-package` in `Packages`.

eproxus:
In macOS it is baked into the application. Removing it there works. 😄

deathaxe:
> Using macOS

... than you need to trust me if I say: "Function definitions are actually there." :smile:

grafik


> As long as function type specs can be distinguished from actual function definitions I'm fine with it. 

The distinction is handled via meta scopes.

A definition selector may look like `meta.function entity.name.function`.

A typespec selector may look like `meta.preprocessor.spec entity.name.function`.

> I don't want tools or features broken because they can't distinguish between actual function and just types.

This is one question to answer. Didn't find too many Erlang plugins which rely on scopes, which I could use for testing.

As long as tools would address the `meta.function` to select function definitions, everything is fine.

eproxus:
Not sure if this is a regression or not, but I noticed that in type specs if I enter invalid syntax it doesn't stop parsing at the period (`.`):

```

-type {.

```

This breaks the rest of the syntax highlighting in the module. In a function this works fine.

eproxus:
ea23474 works like a charm, thank you!

eproxus:
After some more testing it seems it only works before `::` (or `->` in `-callback` attributes). This still shows weird highlighting:
```erlang
-type state() :: {term().
```

deathaxe:
These kinds of bailouts are indeed quite tricky. There may still be a couple of contexts left, which do not yet handle it well. Happy to get the required feedback to find them.

eproxus:
One thing to know about Erlang is that _all_ Erlang code or syntax is a series of expressions ending with dots `.`. Therefore the dot should probably be the end of all expressions always (except for when entering record contexts for example: `#rec.field`).

deathaxe:
@eproxus: Any issues with or remarks about the most recent commits?

eproxus:
@deathaxe I find the `variable.function.erlang` confusing. It seems function names (`func`) in the form `mod:func(Arg)` in set to `variable.function.erlang`, which is strange because they're not variables (if anything, they're constants, the same as atoms). Also, calling a bound fun marks the variable as `variable.function.erlang` but colors it differently (with my color scheme) than other variables: `Func(Arg)`.

deathaxe:
I basically can follow your concerns and agree with you. Using a `variable` scope for function identifiers is somewhat confusing. But this comes from the original scope naming guidelines of TextMate. Hence the `variable.function` is used for function identifiers in any default syntax.

This and some related topics have been discussed in several RFC issues. It's a quite sophisticated task, if we try to respect all kinds of syntaxes. Some even handle functions as variables. Even Erlang does. See the `fun()` expression.

Therefore the agreement was made to interpret `variable` as top-level scope for all kinds of identifiers.

Some contributors even suggested scoping type keywords (`storage.type`s) as `variable.type` - even much more odd, IMHO.


> calling a bound fun marks the variable as `variable.function.erlang`

The original idea was:

```
Fun = fun() -> ... end;
^^^ entity.name.function  -> local function definition

Fun()
^^^ variable.function -> usage of the local function definition
``` 

The first one was removed because of your feedback. So maybe the second one makes no sense anymore as well. `Fun` should be scoped as `variable.other` then?

michaelblyons:
Does removing `variable.function` from almost everywhere mean that the ST built-in "References" popup is basically going to be empty?

Disclaimer: I don't use Erlang at all, so I don't have an educated opinion on the effects.

eproxus:
I've thought about it a bit more and I think that adding this kind of metadata is a lost cause for syntax highlighting. It's somewhat like trying to add some weak typing during syntax highlighting. Without a full type analysis it is impossible to do, and especially for dynamic languages like Erlang it makes even less sense. My vote would be to not try to add annotations like this because they give very little value and potentially cause a lot of confusion.

To elaborate further, a variable in Erlang can be bound to a function reference or not. You can never determine just by looking at the variable if it is (therefore it's impossible to annotate _all_ instance of it). It is also valid in Erlang to assign any other term, and still try to call it as a function (which would fail at runtime, but can't be detected at compile time / syntax highlighting time). Some other terms are (or at least have been) valid to call as function, such as parameterized modules (based on tuples with atoms and other variables).

deathaxe:
Agree. All `fun()` related variables are scoped as `variable.other` now.

`variable.function` is only applied to real functions with atomic identifiers.

eproxus:
I've might have found a possible bug. The last parenthesis is incorrectly highlighted in red, whereas the code compiles without problems:
```erl
-define(CALL(Rec, Func, Args),
    erlang:apply(Rec.mod, Func, Args ++ [Rec.mod_state])
). % <-- Error highlighted here
```

deathaxe:
The culprits are `Rec.mod` and `Rec.mod_state`. Erlang docs (http://erlang.org/doc/reference_manual/records.html) dont tell anything about `Rec.mod` to be valid rexpressions, but as `Rec` is a macro parameter, I guess it is a special case here?

eproxus:
It might be a tricky edge case. Macros in Erlang can contain invalid syntax, since it's just pure string replacement.

I later use the macro as `?CALL(State#state, func, [Arg1, Arg2])` where the full record expression gets expanded.

wbond:
It seems there has been a lot of concern over making the syntax definition fit 100% (or very close to) with Erlang implementation semantics. While we obviously want to try and do a reasonable job, we don't want perfect to be the enemy of progress.

Sublime syntax definitions aren't ever going to be 100% for a very large number of languages. That said, I do think it is worth adding things we have a pretty high confidence in to the definition. The syntax definitions help the user to understand the code better, and help provide tools to navigate and manipulate the code. Since we aren't a language-specific IDE, we are not going to be able to cover everything, but it would be a shame to toss out things that work 80-90% of the time because there are a few edge cases that don't work. For instance, see the C syntax definition and the preprocessor. 🤦‍♂ There is no way to handle all kinds of things with the preprocessor, however we can use heuristics to do the right thing most of the time, and thus provide fairly sane Goto Definition and symbol lists, etc. It would be a huge shame to just punt on a bunch of that since we can't always get it right.

That's just a high-level comment, not specifically directed at any specific decision, but just from reading over the review and comments so far.

In terms of the state of this, @deathaxe do we need to come up with an alternative way to deal with macros since it is "free-form"?

deathaxe:
The macros in Erlang suffer from exactly the same issues in manner of syntax highlighting as the C/C++ ones - they can contain litterly everything - a token, incomplete expression, ... .

The most critical section is the `-define()` statement as it is impossible to reliably detect the meaning/content macro parameters. They look like normal `variable`s but can contain any (incomplete) expression. 

I guess, it was possible to add some more heuristics to try to catch some things, but this would be the  1% of the already 99% of the syntax definition, I think. (Erlang is a great syntax in manner of highlighting and many other asspects!)

I double checked the syntax definition against Erlang's standard library. Even found some bugs/syntax errors with its help ;-) but I did not see (m)any of such macros. Can't say anything about user projects.

The syntax definition should be robust enough to keep possible syntax highlighting issues limited to the `-define()` statement where they are most likely. Several bailouts were added to ensure to limit possible issues to an as small as possible junk of code.

Trying to fix all edge cases with heuristics would probably mean to add another bunch of code just for  the single `-define()` statement. 

As I already doubled the code size just to support the `-spec` and `-type` annotations, I hesitate to do the same for catching macro edge cases as the syntax is otherwise quite complete.

I'd suggest to have a look at such edge cases in case we get related feedback from users.

eproxus:
Just wanted to chime in that I've been using this syntax since I found out about the PR and it's been working really well in my daily work. In my opinion, it's ready to replace the older Erlang syntax!

eproxus:
Found one more oddity:

```
#{foo => bar}.
#{foo := bar}. % <- 'foo' is colored like a variable here
```

I would expect atoms to always be colored like atoms in all situations.

eproxus:
Not sure why it happened, but with build 3211 _Goto Symbol_ <kbd>⌘</kbd>+<kbd>r</kbd> only lists exports and not functions. I've symlinked the Erlang folder from the fork into `~/Library/Application Support/Sublime Text 3/Packages` (and tested without the symlink, where I get no symbols at all).

Does the syntax need to be updated somehow?

deathaxe:
You just need to delete the "Erlang.sublime-package" which is shipped with ST. It contains some preferences, which prevent functions from being added to the symbol list.

eproxus:
@deathaxe Found another tiny oddity. The atom `erlang` is colored differently than other atoms.

deathaxe:
As far as I understood the documentation, the `erlang` atom has a special meaning as it describes the core namespace of the language. Therefore I decided to always scope it as `support.namespace` even if it is used as normal atom (not followed by `::`).

Therefore in both of the following examples `erlang` is scoped the same way:

```erlang

  Var = erlang

  Var = erlang::whatever

```

This is a special case for the `erlang` keyword only.

I have no strong meaning about it. If you find it confusing, we can remove it.

wbond:
Thank you very much for all of your work on this @deathaxe and your feedback @eproxus! This will be part of the next dev cycle.

eproxus:
Thanks @deathaxe, stellar work!

eproxus:
@deathaxe It is the "namespace" (or rather, module) from which several built-in functions are imported (such as `spawn` or `monitor` etc.). But the atom itself doesn't hold any specific meaning in Erlang. You could for example do `[{language, erlang}]` and here `erlang` would be colored differently even though it's just a normal atom.

I found it looked weird when doing `roc:call(Node, erlang, spawn, [fun() -> ok end])` because I would have expected the module and function atoms to be the same color.

eproxus:
@deathaxe Found another tiny bug 😄 

```erl
-define(_get_stacktrace_(),
        try exit('$get_stacktrace') catch exit:'$get_stacktrace':Stacktrace -> Stacktrace end).
%                ^ This is colored like a variable
%                                   ^ catch and exit are not highlighted properly
```

wbond:
@eproxus Since this has now been merged, any bug reports should be created as a new issue. Otherwise they are liable to be lost in the shuffle.

eproxus:
@wbond Reported #2160

PR #1884, opened 2019-02-15
[Python] Fix control keywords and punctuation

deathaxe:
This commit 

1. applies `keyword.control` scopes as discussed in #1228.
2. fixes some inconsistencies with punctuation `:` scopes in control flow statements.
3. removes some unrequired capture groups.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -249,7 +249,7 @@ contexts:
     - match: \bif\b
       scope: keyword.control.conditional.if.python
       push:
-        - meta_scope: meta.statement.conditional.python
+        - meta_scope: meta.statement.conditional.if.python
         - include: line-continuation-or-pop
         - match: ':'
           scope: punctuation.section.block.conditional.python
@@ -265,7 +265,7 @@ contexts:
           pop: true
         - include: expressions
     - match: \b(else)\b(?:\s*(:))?
-      scope: meta.statement.conditional.python
+      scope: meta.statement.conditional.else.python
       captures:
         1: keyword.control.conditional.else.python
         2: punctuation.section.block.conditional.python
@@ -282,7 +282,7 @@ contexts:
     - match: \belif\b
       scope: keyword.control.conditional.elseif.python
       push:
-        - meta_scope: meta.statement.conditional.python
+        - meta_scope: meta.statement.conditional.elseif.python
         - match: ':'
           scope: punctuation.section.block.conditional.python
           pop: true
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -204,17 +204,17 @@ contexts:
     - match: \b(async +)?(for)\b
       captures:
         1: storage.modifier.async.python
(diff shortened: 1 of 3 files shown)


michaelblyons:
I don't want to derail this PR because I'm happy with it already, but I'm curious about the `try/except/else` pattern. The `else` is currently scoped as `keyword.control.conditional.else`. Is that appropriate, or should there be some "special" `.exception` scope for it?

I can argue it either way.

Python/Python.sublime-syntax
```diff
@@ -228,63 +228,63 @@ contexts:
         2: keyword.control.flow.with.python
       push: with-body
     # except ... as ...:
-    - match: \b(except)\b
-      scope: keyword.control.flow.except.python
+    - match: \bexcept\b
+      scope: keyword.control.exception.except.python
```
FichteFoll: `keyword.control.exception.catch` is the general term I proposed (and `catch` is used more commonly).

Should be adjusted for the other occurances as well.

Python/Python.sublime-syntax
```diff
@@ -228,63 +228,63 @@ contexts:
...
+              scope: meta.statement.exception.except.python punctuation.section.block.exception.except.python
               pop: true
             - include: name
         - include: target-list
     - match: \bif\b
-      scope: keyword.control.flow.conditional.python
+      scope: keyword.control.conditional.if.python
       push:
-        - meta_scope: meta.statement.if.python
+        - meta_scope: meta.statement.conditional.python
```
FichteFoll: Why have `meta.statement.exception.catch` but not use `meta.statement.conditional.if` here?

FichteFoll:
Also note that `else` is permitted after loops. This isn't so easy to apply and unrelated to the PR, however.

Python/Python.sublime-syntax
deathaxe: Because the keyword used to be `keyword.control.flow.conditional.python`. 

This raises the general question: Do we want to apply general scopes like `conditional`, `exception` to meta's at all?

FichteFoll: I'm indifferent on that tbh. I don't see myself using `meta.statement` for anything, really. I just added the scopes because it makes debugging easier and we already push into the context.

deathaxe: Me, too. I can see only one use-case which is related with the punctuations, which would basically require a 3-level-meta-scope only.

One post in #1228 requests to be able to scope the punctuation with the same color as the keyword or based on control-type.

Instead of scoping 

    punctuation.section.block.conditional.if
    punctuation.section.block.conditional.else
    ...
    punctuation.section.block.loop
    ...

the meta could propably be used to distinguish punctuation types

    meta.statement.conditional punctuation.section.block.python
    meta.statement.exception punctuation.section.block.python
    meta.statement.loop punctuation.section.block.python

The second solution would help to keep scope names shorter and more general.

FichteFoll:
FYI you should correct the last commit's commit message, although wbond usually squashes anyway.

Issue #1880, opened 2019-02-15
Haskell - highlighting of instance declaration following class declaration


tmcopeland:
### Description

I found some strange behavior in Sublime's Haskell highlighting. The [relevant code](http://learnyouahaskell.com/making-our-own-types-and-typeclasses) is from the _Learn You A Haskell_ book, specifically the paragraph called "A yes-no typeclass".
```haskell
class YesNo a where
    yesno :: a -> Bool

instance YesNo Int where
    yesno 0 = False
    yesno _ = True

instance YesNo [a] where
    yesno [] = False
    yesno _ = True

-- more instances ...
```
In the second code block, immediately following the typeclass declaration, the words `instance` and `where` do not highlight. This is irrespective of the order of the instance declarations. If I swap the instance declarations, the same thing happens, and additionally the expression `[]` does not highlight.

Oddly, this highlights as expected:
```haskell
class YesNo a where
    {yesno :: a -> Bool}
-- ...
```
as does
```haskell
class YesNo a where { yesno :: a -> Bool }
-- ...
```
but not if I break the line after the left brace.
```haskell
class YesNo a where {
    yesno :: a -> Bool }
-- ...
```
It highlights as expected in gedit. See the images below.

![haskell highlight 1](https://user-images.githubusercontent.com/505028/52829044-4b1f9000-3099-11e9-9946-27316e3ca4b3.png)
![haskell highlight 4](https://user-images.githubusercontent.com/505028/52829251-05af9280-309a-11e9-991b-db3a7c25e990.png)
![haskell highlight 2](https://user-images.githubusercontent.com/505028/52829045-4b1f9000-3099-11e9-8138-5b36ea05d9f0.png)
![haskell highlight 3](https://user-images.githubusercontent.com/505028/52829046-4b1f9000-3099-11e9-9d3d-1eae48e47d19.png)

### Environment

* Build 3176, Ubuntu 18.10, GNOME
* No packages installed that would affect Haskell highlighting

jacksoncooper:
This is similar to #1320. I have a patch written and will open a pull request when I'm done with the test cases. I don't think the brackets affect the highlighting because they're ignored by the syntax file. The issue occurs because the context responsible for highlighting type signatures is not always popped when it should be.

wbond:
This issue has been solved by a previous commit, hence it is being closed.

Issue #1321, opened 2017-11-21
[Haskell] Function definition spread over multiple lines not correctly highlighted or lexed correctly


JoeMShanahan:
Given the valid syntax:

```haskell
handlePaymentCreated
  :: EventProcessor Model m
  => PaymentId
  -> Maybe PaymentType
  -> PaymentLink
  -> PaymentAmount
  -> UserId
  -> Maybe Text
  -> Day
  -> m ()
```
which is helpful for adding argument comments where they are needed, .e.g.

````haskell
-> Maybe Text -- ^ Optional granter's note on payment
````

is not highlighted correctly.

This is what it currently looks like:

![image](https://user-images.githubusercontent.com/29176664/33065669-69b3a8cc-cea1-11e7-95cc-f6894960e8e3.png)

This is what it _should_ look like:

![image](https://user-images.githubusercontent.com/29176664/33065929-cd2e59ec-cea1-11e7-9401-612611dd4bc8.png)

The incorrect highlight also leads to the definition itself not being correctly lexed (could have that backwards..). The result is that I can not use Sublime's 'Goto Definition' feature to easily jump to this code.

I assume this is because of the separation between `handlePaymentCreated` and `::`. I believed this to be impossible to fix but @FichteFoll thought otherwise: https://github.com/SublimeTextIssues/Core/issues/1412#issuecomment-337299102

**Comment by Thom1729** [comment]
I'm not a Haskell user. If Haskell sees this:

```haskell
handlePaymentCreated
```

Is this ambiguous? Is there anything else it could be other than the beginning of a type declaration? If so, then there's no way to mark it correctly, and it's just a matter of making the best guess.

However, it absolutely *is* possible for the syntax to guess wrong, see the `::`, and get on the correct path, parsing subsequent code correctly. This is fixable.

JoeMShanahan:
A top level (no identation) expression that isn't followed up by either the function signature (`f ::`) or implementation (`f =`) is a compile error as it's a 'naked expression'. Indenastion is important, so the following is valid:

```haskell
f
  = undefined
```

The highlighter doesn't recognise the above correctly either! (See https://github.com/sublimehq/Packages/issues/1320)

However some keywords used for other syntax (`module`, `type`, `data`, `class`, `instance`) are allowed; Sublime is already aware of these, as they correctly get their own branding of syntax highlight.

Thom1729:
[...]

Issue #1320, opened 2017-11-21
[Haskell] Syntax Highlighting with Guards


JoeMShanahan:
From @kasceh (me) on November 16, 2015 12:36, duplicated by @FichteFoll 

The first piped statement with a guard when not using an `=` (which isn't necessary) isn't coloured properly, and the the statement following it isn't coloured properly either.

The following is valid Haskell:

    handleLookup :: Either String Code -> String
    handleLookup x
      | Either.isLeft x       = "Issue: " ++ fromLeft x
      | otherwise             = "Code: "  ++ fromRight x
but is highlighted like so:

![image](https://user-images.githubusercontent.com/29176664/33065332-71e4c5c2-cea0-11e7-9e25-93e2e2078bf0.png)

whereas having an = after handleLookup x on line 2, styles it like so:

![image](https://user-images.githubusercontent.com/29176664/33065342-7679c9b6-cea0-11e7-8d6a-9c603e06b3f8.png)

EDIT:

Seems to be a general issue that Sublime doesn't fully understand what can terminate a partial-statement:

Correct syntax, Incorrect highlighting:

![image](https://user-images.githubusercontent.com/29176664/33065316-6af167f2-cea0-11e7-8030-a6c68a88d4e8.png))

Incorrect syntax, correct highlighting

![image](https://user-images.githubusercontent.com/29176664/33065355-8008b38e-cea0-11e7-82db-ea09b669128c.png)

Copied from duplicate: https://github.com/SublimeTextIssues/DefaultPackages/issues/159
Copied from original issue: SublimeTextIssues/Core#1024

Wizek:
This issue affects me as well. Found it here via google: https://forum.sublimetext.com/t/haskell-syntax-highlighting-bug/9456/4

widmur:
This has been open for over three years. Is there any way to get this seen by the relevant people?

djspiewak:
I think the biggest problem we have right now is that no one has really stepped forward to seriously maintain the Haskell mode. I would honestly do it, but I have my hands full with work and the Scala mode.

My best recommendation would be to check out fixing it yourself! sublime-syntax is honestly a joy to work with, and pull requests *are* very much welcomed on this repository. The official Discord is a great place to jump in and ask questions; all of us will be more than willing to help you out.

djspiewak:
[...]

PR #1885, opened 2019-02-15
[Haskell] Fix type signature context not popping when encountering a guard or an instance declaration.

jacksoncooper:
Addresses #1320 and #1880. Shamelessly adapted from `SublimeHaskell`'s syntax file [here](https://github.com/SublimeHaskell/SublimeHaskell/blob/master/Syntaxes/Haskell-SublimeHaskell.sublime-syntax#L311).

The test on [line 75](https://github.com/sublimehq/Packages/pull/1885/files#diff-38b1f0659739bb29fa8ff5fc5899f9efL75) of `syntax_test_haskell.hs` was deleted because it relied on the `=` operator to pop the type signature context.

Sublime's default Haskell syntax is a bit dated and could use a rewrite. There are a number of problems inherent to its current design, like #1321. A merge with `SublimeHaskell` may be a good start, but I've found that it too has a couple of idiosyncrasies. I'm not sure how to best engage the community.

### DIFF
--- a/Haskell/Haskell.sublime-syntax
+++ b/Haskell/Haskell.sublime-syntax
@@ -163,8 +163,7 @@ contexts:
         3: keyword.other.double-colon.haskell
       push:
         - meta_scope: meta.function.type-declaration.haskell
-        - include: comments
-        - match: ^(?!\1\s)
+        - match: ^(?!\s*(?:--|{-|$)|\1\s)
           pop: true
         - include: type_signature
     - match: '\b[A-Z]\w*\b'
--- a/Haskell/Haskell.sublime-syntax
+++ b/Haskell/Haskell.sublime-syntax
@@ -157,19 +157,14 @@ contexts:
         4: constant.character.escape.hexadecimal.haskell
         5: constant.character.escape.control.haskell
         6: punctuation.definition.string.end.haskell
-    - match: '^\s*([a-z_][a-zA-Z0-9_'']*|\([|!%$+\-.,=</>]+\))\s*(::|∷)'
+    - match: '^(\s*)([a-z_][a-zA-Z0-9_'']*|\([|!%$+\-.,=</>]+\))\s*(::|∷)'
       captures:
-        1: entity.name.function.haskell
-        2: keyword.other.double-colon.haskell
+        2: entity.name.function.haskell
+        3: keyword.other.double-colon.haskell
       push:
         - meta_scope: meta.function.type-declaration.haskell
-        - match: |-
-            (?x)
-                ^(data|newtype|type|class|deriving)\s  # When a top level declaration starts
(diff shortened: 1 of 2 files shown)


Haskell/Haskell.sublime-syntax
```diff
@@ -157,19 +157,14 @@ contexts:
...
         - meta_scope: meta.function.type-declaration.haskell
-        - match: |-
-            (?x)
-                ^(data|newtype|type|class|deriving)\s  # When a top level declaration starts
-              | ^[^=]*(=)[\sa-zA-Z0-9_\(]                # A function declaration
-          captures:
-            1: keyword.other.haskell
-            2: keyword.operator.haskell
+        - include: comments
+        - match: ^(?!\1\s)
```
FichteFoll: You may want to ignore line comments here, too, unless the comments include above also starts from the first character of the line and would thus take precedence. (I'm in mobile.)

jacksoncooper: The `comments` context includes single-line comments and the `block_comment` context. None of their patterns are anchored to the beginning of the line. Won't the comments always take precedence because of the position of the definition?

From [Sublime's documentation](https://www.sublimetext.com/docs/3/syntax.html#contexts): "When multiple patterns match at the same position, the first defined pattern will be selected."

FichteFoll: The pattern I annodated here matches on the first character of a line (due to the `^` anchor) while the comment pattern would only match on the column the comment begins. You should be able to test this by having a comment with less leading whitespace than the original indent where this context should not be popped.

jacksoncooper: Apologies for the confusion. Is there a way to ignore comments without producing overlapping `meta_scope`'s?

FichteFoll: What kind of meta scope are you looking to remove? The meta for a function definition? I don't think that matters, or any similar scope. I mean, we are still within that conceptually and it's just a comment interweaved.

jacksoncooper: I was trying to avoid having a comment that has both `meta.function.type-declaration.haskell` and `comment.line.double-dash.haskell`. This would occur if I replaced the popping pattern with `^(?!\s*--|\1\s)`. I guess this is a nonproblem. It just thought it made more sense to think of a comment that splits a type signature as an interruption of the signature, as opposed to being contained within it.

FichteFoll: In all other syntaxes we consider comments to be extaneous and they may be appear anywhere they are legal. There is no splitting going on, it's just some dust lingering around.

wbond:
Thank you @jacksoncooper and @FichteFoll!

PR #1888, opened 2019-02-21
update hook in actually rails 5.x

pandalowry:
before_filter -> before_action, see: https://guides.rubyonrails.org/action_controller_overview.html#filters

### DIFF
--- a/Rails/Snippets/Create-controller-class.sublime-snippet
+++ b/Rails/Snippets/Create-controller-class.sublime-snippet
@@ -1,6 +1,6 @@
 <snippet>
 	<content><![CDATA[class ${1:Model}Controller < ApplicationController
-	before_filter :find_${2:model}
+	before_action :find_${2:model}
 
 	$0
 


wbond:
Is this only available in 5.x? If so, I'd be hesitant to change since I imagine there are many people using older versions of Rails.

pandalowry:
Yes, 5.x only. I do not know what to do, because I don’t know the proportions of the number of users of old versions. Perhaps you are right and change is not necessary.

FichteFoll:
How old is 5.x? If we can reasonably expect most people to run on the newer version, I think it makes sense to update the snippet. After all, this isn't like we're depreciating or removing syntax highlighting for the old variant.

pandalowry:
Rails [release 5.0 stable](https://github.com/rails/rails/tree/5-0-stable) is same 4 years ago. Now working version is Rails 5.2.3 and moreover, rails 6.0 has been stable for about 5 months. And ultimately, he will be waiting for us. But so far in the documentation of the rails even the master branch builds (6) docs are have snippet who i propose [proof](https://edgeguides.rubyonrails.org/action_controller_overview.html#filters)

pandalowry:
p.s. ling edgeguides.rubyonrails.org is "nightfly" build for official docs.

wbond:
It seems Rails 4 is no longer supported now that 6 is out, so I think it is probably reasonable to update this.

PR #1889, opened 2019-02-25
[Graphviz] Build system

michaelblyons:
Add the png-version of the Graphviz [CLI commands](https://graphviz.gitlab.io/_pages/pdf/dot.1.pdf) (plus one for `dot` to make SVGs).

Also fix some completions that throw errors as they currently exist.


wbond:
Thanks!

PR #1891, opened 2019-03-01
[PHP] Add support for inline PHPDoc

jfcherng:
See http://docs.phpdoc.org/references/phpdoc/tags/var.html .

Currently we only have multi-line doc highlighted.

```php
/**
 * @var string $name        Should contain a description
 * @var string $description Should contain a description
 */
```

However, this inline doc is valid as well.

```php
/** @var string|null Should contain a description */
//  ^ keyword.other.phpdoc
```


wbond:
Thanks!

PR #1896, opened 2019-03-09
[Python] Fix mapping element starting with 'for'

TheClams:
Add missing \b to capture inline for inside a dict.
Add relevant test.

Reported in forum: https://forum.sublimetext.com/t/python-syntax-highlight-bug-after-update/42248

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1016,7 +1016,7 @@ contexts:
       push:
         - clear_scopes: 1
         - meta_content_scope: meta.mapping.value.python
-        - match: (?=\s*(\}|,|for))
+        - match: (?=\s*(\}|,|for\b))
           pop: true
         - include: expressions
 
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1003,7 +1003,7 @@ contexts:
         - match: ','
           scope: punctuation.separator.mapping.python
           set: inside-dictionary
-    - match: (?=for)
+    - match: (?=for\b)
       push:
         - match: (?=\})
           pop: true
--- a/Python/syntax_test_python.py
+++ b/Python/syntax_test_python.py
@@ -950,6 +950,10 @@ mydict = {"key": True, key2: (1, 2, [-1, -2]), ,}
 #                                              ^ invalid.illegal.expected-colon.python
 #                                               ^ punctuation.section.mapping.end - meta.mapping.key
 
+mydict = { 'a' : xform, 'b' : form, 'c' : frm }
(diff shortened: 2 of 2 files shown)


FichteFoll:
Do you have an example for a character where the suggested `\b` would not be enough?

Thom1729:
Example, derived from the JavaScript tests:

```py
def foo(_Foobar1Ù𝓩ʷªאξ‿ᛮↂ〩):
    pass
```

TheClams:
Considering that \b is use everywhere in this syntax, it might be good to keep it the two issues in different PR

deathaxe:
@Thom1729 do we really need to kill readability and performance of each syntax by such sophisticated negative lookaheads? I think, no serious python developer will use such identifiers. The utf-support was primarily made to allow unicode characters in strings.

I wonder if it would rather be useful to be able to tell the engine what `\b` means.

FichteFoll:
I just went through a quick test with matching `foo<x>` using ST's find panel, which utilizes boost's regex, where `<x>` represents any character you used in the `_Foobar…` paramter. The only characters where `\b` matches are `ᛮↂ〩`, but at least I could confirm that the current behavior is not 100% correct. That said, I believe it easier to leave it at `\b` and only change things once someone complained.

@deathaxe I definitely agree with your stance, but I'd like to see benchmarks first before declaring something as "killing performance".

Thom1729:
If there is a performance tradeoff, then we'll have to evaluate that tradeoff.

It definitely doesn't need to be a part of this issue. Other than that, I think that it's better to be correct than not, even if no one's currently complaining, unless it turns out to be costly in terms of performance or complexity.

I played around with an implementation yesterday, but ran into a couple of minor issues. #1897 may address some of them.

Thom1729:
Addendum: it looks like sregex doesn't actually support `xid_` properties. Regexps using them run through Oniguruma, which hurts performance. Until this is fixed, it would be best to hold off.

https://github.com/SublimeTextIssues/Core/issues/2193

FichteFoll:
`\b` should be added in [line 1019](https://github.com/sublimehq/Packages/pull/1896/files#diff-62db422a04a96e067da114ba6ed7d6e9R1019) as well.

TheClams:
Correct.
Any idea for a test that would fail when the \b is missing ? I do not seem to manage to trigger an issue (just to add it to the syntax test)

FichteFoll:
No, I can't even tell why the current test case works. My only hunch is that ST detects the infinite loop and skips the offending pattern (the one with the `for` look-ahead) so that the following `expression` include matches instead and things proceed as normal.

I suggest adding the `\b` nonetheless.

PR #1897, opened 2019-03-11
[Python] Fix Oniguruma incompatibility with regex for path

robinst:
In 0550509bcd5, `path` was modified to allow spaces around `.`. Later, `simple_expression` started using `path` but within a `(?x:...)`, which means this:

    \b[[:alpha:]_][[:alnum:]_]*\b *\. *

is equivalent to:

    \b[[:alpha:]_][[:alnum:]_]*\b*\.*

While sregex doesn't complain about that, `\b*` is invalid in Oniguruma:

Screen Shot 2019-03-11 at 17 20 12

Note that even with sregex, the meaning of the regex changes.

By using a character class for the space, the pattern works the same way
whether it's within a `(?x:...)` or not.


robinst:
(Note, I tried creating a test case for this, but `simple_expression` is the only place where `{{path}}` is used like that, and `simple_expression` matches something like `foo . bar` because of other parts of the pattern anyway.)

wbond:
Thanks!

PR #1900, opened 2019-03-14
[Lua] Fix builtins + add Lua 5.1/5.2 builtins

squeek502:
Existing syntax had some erroneous builtin functions and was missing highlighting for builtin functions specific to Lua 5.1/5.2 (note that Lua 5.1 is still very much in use, e.g. LuaJIT).

Reference manuals:
- Lua 5.1: https://www.lua.org/manual/5.1/#index
- Lua 5.2: https://www.lua.org/manual/5.2/#index
- Lua 5.3: https://www.lua.org/manual/5.3/#index


wbond:
Thanks!

PR #1904, opened 2019-03-15
[Lua] Improve highlighting when typing an if statement.

### LINKED ISSUE #1903: Lua - is it intended that a missing "then" removes subsequent highlighting?
[Screenshot](https://puu.sh/D0lKP.png)

A missing "do" after a "for" does not have the same behavior for example, [screenshot](https://puu.sh/D0lLL.png)

If this is intended could someone help me edit it locally for myself? I find it super distracting while I'm writing code.

Thom1729:
Fixes #1903.

```lua
if true

42
```

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -140,16 +140,6 @@ contexts:
     - match: if{{identifier_break}}
       scope: keyword.control.conditional.lua
       push:
-        - - include: end
-          - match: elseif{{identifier_break}}
-            scope: keyword.control.conditional.lua
-            push:
-              - if-then
-              - expression
-          - match: else{{identifier_break}}
-            scope: keyword.control.conditional.lua
-            push: if-block
-
         - if-then
         - expression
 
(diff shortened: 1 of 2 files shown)


Lua/Lua.sublime-syntax
```diff
@@ -220,8 +210,15 @@ contexts:
...
-      pop: true
+    - include: end
+    - match: elseif{{identifier_break}}
+      scope: keyword.control.conditional.lua
+      set:
+        - if-then
+        - expression
+    - match: else{{identifier_break}}
+      scope: keyword.control.conditional.lua
+      set: if-block
```
FichteFoll: This does ... nothing? Besides stacking `meta.block.lua`, I guess.

wbond: It probably isn't really important, but this will end up allowing multiple `else` clauses, right?

Thom1729: Yes, and/or `elseif` after `else`. If we care, we could add another context to mark `else`/`elseif` after `else` as errors.

wbond: Eh, the more we get into trying to write a half-baked parser, the more complex and brittle it will get. I think it is probably fine as of now.

PR #1912, opened 2019-03-17
[Perl] Apply numeric scopes

deathaxe:
This commit applies the latest standardized `constant.numeric` scope names to the perl syntax.

See: https://www.sublimetext.com/docs/3/scope_naming.html#constant


wbond:
Thanks!

PR #1919, opened 2019-03-20
[Matlab] Fix typo in unwind snippet

apjanke:
There's a typo in one of the identifiers in the Octave-oriented "unwind" snippet in the Matlab package.


wbond:
Thank you!

PR #1920, opened 2019-03-20
[Matlab] Make octfun and unwind snippets available in source.octave code

apjanke:
Most of the Matlab snippets are applied to both `source.matlab` and `source.octave` code. Except for the `octfun` and `unwind` snippets, which are, curiously, Octave-specific.

This PR makes them available in `source.octave` code as well.


wbond:
Thank you for these Matlab fixes, sorry for the delay in getting them merged

Issue #1663, opened 2018-07-15
[ShellScript] Change applicable scope of snippets


Phidica:
The ShellScript snippets for `case`, `if`, `elif`, `for`, `until`, and `while` are all applied to the scope of `source.shell`. However, they are only valid for `source.shell.bash` and similar shells. They aren't valid in, for instance, the [fish shell](https://github.com/Phidica/sublime-fish), which uses the scope `source.shell.fish`. These snippets being offered when writing in fish gets in the way of selecting the actual fish snippets to correctly construct these command blocks. I think the snippets need to be scoped to `source.shell.bash` instead.

rwols:
I'm all for this, but changing the scope of the snippets breaks compatibility with users still using the (deprecated?) Shell-Unix-Generic syntax. For instance, some users had a few problems with the ApplySyntax plugin, [although I see now that that has been fixed](https://github.com/facelessuser/ApplySyntax/commit/0faad8ee6ec13b15c74375b5198137a804054cd5). I think we can at least go ahead and make the pull request to see what the maintainer thinks of it.

nhooyr:
Fish should have its own scope`source.fish` instead of being a subtype of shell. Its unfortunate but putting it under shell only helps for organizational purposes, its syntax and usage is very different from other shells.

Phidica:
Hmm. Originally I did have the scope as `source.fish`, but changed it to bring it in line with that used for other shells. The problem is really that `source.shell` existed for a time as a specific syntax rather than as a prefix for further specification of shells. But if `source.shell` is deprecated now as a specific syntax, then I would prefer to co-operate with making it a useful organisational prefix for shell syntaxes such as fish

nhooyr:
How about we put all the other shells under `source.shell.posix` for shells that conform to posix syntax and then just have fish at `source.shell.fish`?

That'd be a pretty clean way to solve this.

FichteFoll:
Good idea on grouping of similar shells. I'm not too much into the differences of shells and their history, but what about `source.shell.bourne`?

Phidica:
We should probably come to some preliminary decision, and I'll take it to a PR

Looking back through this, I'm not entirely sure I see a need for adding another layer to the scope hierarchy. We could simply stick with
`source.shell.bash`
`source.shell.fish`
`source.shell.zsh` (http://github.com/molovo/sublime-zsh is 2 years stale and using `source.shell`)
[...]

PR #1923, opened 2019-03-22
[ShellScript] Make snippets friendlier to other shell packages

Phidica:
This PR follows on from discussion in #1663. Though no clear consensus was reached, I present a potential solution to the issue for practical discussion.

Specifically, the problem is that snippets which only work in Bash (or other Bourne-like shells) are suggested for the `source.shell` scope, which includes `source.shell.fish` ([Phidica/sublime-fish](https://github.com/Phidica/sublime-fish)) despite the syntaxes being entirely incompatible. The `run_shell_script` command also provided by the ShellScript package is shell-agnostic, and makes a good case for `source.shell` to be the underlying scope for all shell scripts regardless of their compatibility with Bash.

Accepting this PR would change the applicable scope of the snippets to be the scope assigned by the Bash syntax, `source.shell.bash`, to prevent these snippets being suggested in fish scripts. It would also extend the `!env` snippet to read the intended shell from `source.shell.<name>` and insert `#!/usr/bin/env <name>` rather than always `#!/usr/bin/env bash` (though in the case of `source.shell` it would continue to assume `bash`).

`source.shell` used to be the scope for the Bash syntax, and in some older locations they may still be considered interchangeable. Is it acceptable to change these snippets such that they would no longer work in a simple `source.shell` scope?


Phidica:
Are there any comments on this?

wbond:
This seems like a reasonable change, thank you for the time you put into discussing and proposing @Phidica!

PR #1925, opened 2019-03-24
[Python] Add syntax highlighting for type comments

TeamSpen210:
This makes `#type:` comments highlighted as normal code (since type checkers expect these to be valid syntax), and treats `#type: ignore` as a keyword. I'm not sure though if it should additionally have the `comment.*` scope for the line - any schemes would probably ignore the additional highlighting and make it appear as a comment.


Thom1729:
Can there be a space in between `type` and `:`?

TeamSpen210:
It's not explicit in the PEP, but both MyPy and PyCharm treat `# type :` as a normal comment.

keith-hall:
> I'm not sure though if it should additionally have the `comment.*` scope for the line - any schemes would probably ignore the additional highlighting and make it appear as a comment.

seems unlikely, comments could have a background color for example, plus the scope specificity means the `keyword` etc will take precedence. I vote for giving it a distinct `comment.line.type-hint` scope or something so color scheme authors can choose what to do.

TeamSpen210:
In that case, `# type: SomeClass` looks a bit odd, since non-builtin names don't get a specific scope, other than `meta.generic-name.python`. So there's no specific rule to override the comment one.

keith-hall:
ah good point. But in this context, do we not know for sure that it is a class, such that it can be assigned `support.class.python` instead of a generic meta scope?

FichteFoll:
I've had this on my todo, but ideally I'd like to see https://github.com/sublimehq/Packages/issues/1842#issuecomment-464938573 resolved because it will hugely affect the type of scopes to be used. Python currently doesn't have highlighting of user types, i.e. class names, but I would like to change that for both type comments and variable/function annotations. However, I'm undecided on the particular scope to use. Do they get `variable.type`? Can I use `support.type` for classes provided by the `typing` module? How do people feel about built-in types getting a different scope than user types? This is traditionally easy to answer for languages that have primitive and complex types like Java, but not for Python.

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,96 @@ contexts:
       pop: true
 
   comments:
+    # Type-ignore must be the only hint allowed!
+    - match: '(#)\s*(type)(:)\s*(ignore)'
```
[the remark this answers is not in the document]
kaste: ignore[rule-name] is allowed here.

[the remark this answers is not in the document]
FichteFoll: Where is this defined?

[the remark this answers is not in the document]
TeamSpen210: I wasn't aware that existed, and it's not technically standard.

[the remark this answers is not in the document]
FichteFoll: Hm, yes, I agree. I think adding a `(?=\b|\[)` lookahead will make sure we're not matching extra syntax like `ignore_brace` but also not get in the way of mypy's attribute ignores.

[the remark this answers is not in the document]
kaste: I'm not super familiar with the syntax system, but after type comments other comments are allowed, t.i. `# type: List  # normal comment`. Does this standard comment also has the standard styling? Or does the whole line gets marked as a type-hint? (Basically, before the second `#` it should 'pop' the context, if that makes sense.)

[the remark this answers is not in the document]
TeamSpen210: Yes, I took that into account. It switches to a different scope (`comments_noannotation`) which allows a normal comment, but not another type comment (treating it normally). If you look at the syntax test, you can see that being verified.

Python/syntax_test_python.py
```diff
@@ -1350,6 +1350,74 @@ class Starship:
...
 
+# Type comments - type: ignore must be by itself.
+
+primes = 5  # type: ignore # type: not-a-type-comment
+#           ^ comment.line.type-hint punctuation.definition.comment.python comment.line.number-sign.python
+#             ^^^^ comment.line.type-hint keyword.other.annotation.type-comment
+#                 ^ comment.line.type-hint punctuation.separator.annotation.type-comment
+#                   ^^^^^^ comment.line.type-hint keyword.other.annotation.type-comment.ignore
+#                          ^ comment.line.type-hint comment.line.number-sign.python punctuation.definition.comment.python
+#                            ^^^^^^^^^^^^^^^^^^^^^^^^ comment.line.type-hint comment.line.number-sign.python
```
kaste: From this test, but also below, it looks like you're not popping.

Python/syntax_test_python.py
```diff
@@ -1350,6 +1350,74 @@ class Starship:
...
+#                       ^^^^ comment.line.type-hint.python meta.function.parameters.python support.class.python
+#                                  ^^ comment.line.type-hint.python meta.function.parameters.python keyword.operator.unpacking.mapping.python
+#                                    ^^^^ comment.line.type-hint.python meta.function.parameters.python support.type.python
+#                                        ^ comment.line.type-hint.python punctuation.section.parens.end.python
+#                                          ^^ comment.line.type-hint.python meta.function.return-type.python punctuation.separator.sequence.python meta.function.return-type.python
+#                                             ^^^^ meta.function.return-type.python support.class.python
+#                                                            ^ meta.function.return-type.python comment.line.number-sign.python punctuation.definition.comment.python
+#                                                              ^^^^^^^^^^^^^^^^ meta.function.return-type.python comment.line.number-sign.python
```
kaste: Here, shouldn't this be comment - meta?

kaste: Might be noobish, just asking.

Python/syntax_test_python.py
FichteFoll: Yeah, the comment scopes are stacked, and multiple times so. While this certainly isn't clean, I doubt it would break anything, so I didn't comment on it.

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+    - match: '(#)\s*(type)(:)'
+      scope: comment.line.type-hint.python
+      captures:
+        1: punctuation.definition.comment.python comment.line.number-sign.python
+        2: keyword.other.annotation.type-comment.python
+        3: punctuation.separator.annotation.type-comment.python
+      embed: type-comment
+      escape: '\n'
+      embed_scope: comment.line.type-hint.python
+
```
wbond: Style: remove this blank line

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+    - include: comments_noannotation
+
+  comments_afterignore:
```
wbond: It seems stylistically that other context named use `-` as a separator, not `_`. Let's stay consistent with that.

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+            - match: \[
+              scope: invalid.illegal.python
+              pop: true
+            - include: comments_afterignore
+    - match: \S
+      scope: invalid.illegal.python
+
+  type-comment:
+    - include: comments_noannotation
+
```
wbond: Style: remove this blank line

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+    # Don't use expressions here, only a few types are reasonable
+    # as annotations.
+    - include: constants
+    - include: builtin-exceptions
+    - include: builtin-types
+    - include: builtin-functions
+    - match: '{{identifier}}'
+      scope: support.class.python
+
```
wbond: Style: remove this blank line

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+        1: meta.item-access.python punctuation.section.brackets.begin.python
+      push:
+        - meta_content_scope: meta.item-access.arguments.python
+        - match: \]
+          scope: meta.item-access.python punctuation.section.brackets.end.python
+          pop: true
+        - match: ':'
+          scope: punctuation.separator.slice.python
+        - include: type-comment
+
```
wbond: Style: remove this blank line

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+    - match: '\('
+      scope: punctuation.section.parens.begin.python
+      push:
+        - meta_content_scope: meta.function.parameters.python
+        - match: \)
+          scope: punctuation.section.parens.end.python
+          pop: true
+        - match: \(
+          scope: invalid.illegal.stray.brace.round.python
+
```
wbond: Style: remove this blank line

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+        - match: \*{3,}
+          scope: invalid.illegal.syntax.python
+        - match: \*\*
+          scope: keyword.operator.unpacking.mapping.python
+        - match: \*
+          scope: keyword.operator.unpacking.sequence.python
+
```
wbond: Style: remove this blank line

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+        - match: '->'
+          scope: invalid.illegal.stray.return.arrow.python
+        - include: type-comment
+
```
wbond: Style: remove this blank line

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+    - match: \)
+      scope: invalid.illegal.stray.brace.round.python
+    - match: \]
+      scope: invalid.illegal.stray.brace.square.python
+    - match: \}
+      scope: invalid.illegal.stray.brace.curly.python
+    - match: \S
+      scope: invalid.illegal.python
+
+  comments_noannotation:
```
wbond: Use a `-` here instead of a `_` for stylistic consistency

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+          scope: invalid.illegal.stray.return.arrow.python
+        - include: type-comment
+
+    - match: '->'
+      scope: punctuation.separator.sequence.python meta.function.return-type.python
+      set:
+        - match: '->'
+          scope: invalid.illegal.stray.return.arrow.python
+        - include: type-comment
+        - meta_scope: meta.function.return-type.python
```
wbond: This should be moved up 3 lines so it is the first entry in the anonymous context

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+          scope: keyword.operator.unpacking.mapping.python
+        - match: \*
+          scope: keyword.operator.unpacking.sequence.python
+
+        - match: '->'
+          scope: invalid.illegal.stray.return.arrow.python
+        - include: type-comment
+
+    - match: '->'
+      scope: punctuation.separator.sequence.python meta.function.return-type.python
```
wbond: The `meta` scope should come before the `punctuation` scope

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+          scope: meta.item-access.python punctuation.section.brackets.end.python
+          pop: true
+        - match: ':'
+          scope: punctuation.separator.slice.python
+        - include: type-comment
+
+    - match: '\('
+      scope: punctuation.section.parens.begin.python
+      push:
+        - meta_content_scope: meta.function.parameters.python
```
wbond: This should probably be `meta_scope` so it applies to the `(` and `)`, right?

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+        3: punctuation.separator.annotation.type-comment.python
+      embed: type-comment
+      escape: '\n'
+      embed_scope: comment.line.type-hint.python
+
+    - include: comments_noannotation
+
+  comments_afterignore:
+    - meta_scope: comment.line.type-hint.python
+    - include: comments_noannotation
```
wbond: Why is this here? From my reading, this only matches a `#` as a start of a line comment, but we are already in a line comment.

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+      pop: true
+    - match: \[
+      # Mypy's # type: ignore[error-code1, error-code2]
+      scope: punctuation.section.brackets.begin.python
+      set:
+        - meta_scope: comment.line.type-hint.python
+        - match: \[
+          scope: invalid.illegal.python
+        - match: '[a-zA-Z_-][a-zA-Z_0-9-]+'
+          scope: comment.line.type-hint.python.mypy-errorcode
```
wbond: We don't need the `.mypy-errorcode` here. The order is reversed, but mostly it is far too specific a selector and it unlikely to really ever be used.

Python/Python.sublime-syntax
```diff
@@ -402,6 +402,120 @@ contexts:
...
+        - match: '[0-9]'
+          scope: invalid.illegal.python
+        - match: ','
+          scope: punctuation.separator.sequence.python.mypy-errorcode
```
wbond: Also remove `.mypy-errorcode` here.

Python/syntax_test_python.py
```diff
@@ -1350,6 +1350,74 @@ class Starship:
 #        ^ punctuation.separator.annotation.variable.python
 #                                   ^ keyword.operator.assignment
 
+# Type comments - type: ignore must be by itself.
+
+primes = 5  # type: ignore # type: not-a-type-comment
+#           ^ comment.line.type-hint punctuation.definition.comment.python comment.line.number-sign.python
```
wbond: The stacking of scopes here looks wrong. We shouldn't be adding a second `comment.line`, and especially not on top of the `punctuation.definition.comment`.

Python/syntax_test_python.py
```diff
@@ -1350,6 +1350,74 @@ class Starship:
...
+#                                       ^ invalid.illegal.python
+#                                             ^ comment.line.type-hint.python punctuation.section.brackets.end.python
+#                                              ^ invalid.illegal.python
+
+# With/for allow commas
+for a, b in lst: # type: str, int
+#                           ^ comment.line.type-hint.python punctuation.separator.sequence.python
+
+
+primes = 5  # type: List[Dict[*property:str, ...]], bool # comment
```
wbond: @FichteFoll What do you think about using `clear_scopes` during the type definition to remove the `comment` scope and make it more readable? I wouldn't remove it from the `type: ` part, but maybe the `List[Dict[*property:str, ...]], bool` part?

FichteFoll: For color schemes that have a background color for comments, I think it will look weird to remove the comment scope for something that clearly is a comment syntactically. We don't do this in JavaDoc and similar either.

For those that only define a foreground color (the majority), it may look weird to the punctuation characters in the comment color. Additionally, it may confuse plugins that rely on scope selectors, even though the only one that comes to mind is SublimeLinter-annotations and that's not affected.

I guess with the latter much more wide-spread, we could try removing the `comment` scope, add a meta scope for accuracy and see how that goes.

wbond: Alright, seems there are enough significant side effects to not bother.

wbond:
I think I've resolve the issues I found during review. If anyone is up for a final review, I'll plan on merging this soon.

Python/syntax_test_python.py
```diff
@@ -1490,70 +1490,80 @@ class Starship:
...
+#                                               ^ punctuation.section.brackets.end
+#                                                ^ punctuation.section.brackets.end
+#                                                 ^ punctuation.separator.sequence
+#                                                   ^^^^ support.type
+#                                                        ^^^^^^^^^^ comment.line.number-sign
+#                                                        ^ punctuation.definition.comment
 
 # Python 2.7 function annotations.
 def function(a, b, *c, **d):
     # type: (int, str, *List[str], **bool) -> Dict[str, str] # type: noncomment
```
kaste, line 1552: This is not a valid type comment, there is no `*` or `**` unpacking.

Python/syntax_test_python.py
```diff
@@ -1490,70 +1490,80 @@ class Starship:
 # Type comments - type: ignore must be by itself.
 
 primes = 5  # type: ignore # type: not-a-type-comment
```
kaste, line 1492: Just for clarification

```
primes = 5  # type: ignore # type: str   // technically ok but weird
primes = 5  # type: str  # type: ignore  // not tested here but typical
primes = 5  # type: str  # type: str     // not tested here and invalid
```

(I don't know how fine-grained these tests are here.)

Python/syntax_test_python.py
kaste:
```
# type: (int, str, *List[str], **bool) -> Dict[str, str] # type: noncomment
                                                         ^^^^^^^^^^^^^^^^^^ invalid
```

as above.

Python/syntax_test_python.py
```diff
@@ -1490,70 +1490,80 @@ class Starship:
...
+#                                              ^^^^^^^ invalid.illegal
+#                                                      ^^^^^ comment.line.number-sign - comment.line.type-hint
 
 # With/for allow commas
 for a, b in lst: # type: str, int
-#                           ^ comment.line.type-hint.python punctuation.separator.sequence.python
+#                           ^ comment.line.type-hint punctuation.separator.sequence
 
 
 primes = 5  # type: List[Dict[*property:str, ...]], bool # comment
```
kaste, line 1526:
```
primes = 5  # type: List[Dict[*property:str, ...]], bool # comment
                                                  ^^^^^^ invalid
```

This looks like a tuple but is invalid.  E.g. `# type: int, str` is invalid.

rwols:
I think this needs more work.

Screenshot from 2020-07-21 22-53-50

Screenshot from 2020-07-21 22-55-07

wbond:
@rwols Can you paste text of a couple of snippets where you see failures?

@kaste I'm tempted to get rid of trying to mark things as invalid, and instead just highlight the things known to be valid, thoughts? You may also have some thoughts @FichteFoll as you've worked on marking various things as invalid in this syntax before.

Python/syntax_test_python.py
wbond: Is this valid syntax? I haven't seen this mentioned anywhere: `Dict[property:str, ...]`

wbond:
@rwols I've added fixes for the screenshots you uploaded, if you wouldn't mind checking again?

@kaste I think I've taken out the invalid syntax you've pointed out.

Python/syntax_test_python.py
wbond: The impression I had gotten from the previous implementation was that line comments added into a type comment effectively ended the processing of mypy. Is that true?

From your comment above, it almost sounds like mypy will process nested comments that start with `type:` and the only logical combinations are `type: ignore` and a normal `type:` comment, right?

rwols:
Sorry for not pasting the text, it was late :) The last two commits e3b57a9 and 97ae638 fixed the problem.

Screenshot from 2020-07-22 18-43-05
Screenshot from 2020-07-22 18-42-57

My only remark now is that the syntax contexts from the return type and the `type: ...` annotation are different. In the screenshots above you can see that `-> List[str]` is highlighted differently than `# type: List[str]`.

Minimal reproducible case:
```python
def f() -> List[str]:
    #      ^^^^ support.class
    x = []  # type: List[str]
    #               ^^^^ support.class
```
Perhaps it can be argued that they shouldn't share the same context. But I don't know what else you'd use the `-> ...:` syntax for.

This is still broken:

```python
class SessionBufferProtocol(Protocol):

    session = None  # type: Session
    view = None  # type: sublime.View
    #                           ^ - invalid.illegal
    session_views = None  # type: WeakSet[SessionViewProtocol]
    file_name = None  # type: str

    def on_diagnostics_async(self, diagnostics: List[Dict[str, Any]], version: Optional[int]) -> None:
        ...
``` 
It's perfectly legal to type-hint something from another module.

rwols:
Minimal reproduction case:

```python
x = []  # type: List[sublime.Region]
```

rwols:
There's also this issue:

```python
x = []  # type: 'List[ParameterInformation]'
#               ^ - invalid.illegal
#                                          ^ - invalid.illegal

```
While it looks a bit weird, it's allowed to wrap the type in a string (single-quotes or double-quotes). Used for forward declarations.

wbond:
I figured I'd take a quick pass at some fixes, but it seems this requires a fair bit more work. I'm probably gonna hit pause for now and move onto other things.

Python/syntax_test_python.py
kaste: Maybe mypy stops after the first `type: ignore` comment. 🤷  After that everything is "egal".  The only thing you see in practice is `# type: str  # type: ignore` to brutally force a type.  And typical `# type: ignore[foo-rule]  # mypy bug https:\\issue-nr` because you really get hit by a mypy bug every other day.

Python/syntax_test_python.py
kaste: `Dict[foo:bar, ...]` is not valid; neither the `:` nor the `...` in that position.

kaste:
FWIW: `# type: (int, str, List[str], bool) -> Dict[str, str]  # type: comment` for the function type is illegal, no second `# type:` is allowed here.  

There is also no "tuple" thing `# type: (str, int)`. No parens allowed. Generally only `[]` as a subscript, for example `Tuple[int, str]`.  

The only fancy thing left is `str | int` for the `Union[str, int]`, but not in top position. T.i. `a = 5  # type: str | int` is not allowed; `a = {}  # type: Dict[str, str | int]` is.  Hilarious.

Python/syntax_test_python.py
```diff
@@ -1519,51 +1519,46 @@ class Starship:
 #                                                      ^^^^^ comment.line.number-sign - comment.line.type-hint
 
 # With/for allow commas
-for a, b in lst: # type: str, int
-#                           ^ comment.line.type-hint punctuation.separator.sequence
+for a, b in lst: # type: (str, int)
```
kaste, line 1522: The parens in `(str, int)`are not allowed.  It's actually `Tuple[str, int]`.  They take explicit > implicit seriously.

kaste:
Generally, I don't expect Sublime marks my syntax errors as "invalid". The invalid scope is hard to edit and read, and I sure don't expect it at all.  I always thought it is an escape hatch just before the syntax engine becomes incapable of "fixing" it, i.e. keeping the colors reasonable. (T.i. I don't expect it to read my code actually.)  

I just found it strange to have so many invalid examples in the few python test lines.

FichteFoll:
> You may also have some thoughts @FichteFoll as you've worked on marking various things as invalid in this syntax before.

I generally prefer to use invalid scopes only when the syntax parsing reaches a state it cannot recover due to invalid syntax, such as keywords being used when they are not allowed, or when the user is likely to have made an error such as for stray braces. It seems debatable, whether a type comment is a strongly parsed region, considering that on one hand comments are free-form, but on the other hand type comments are expected to be parsed by a type checker.

I have never used type comments in Python myself, so I can only comment in a general sense, but my biggest issue is matching any identifier as `support.type.python`, which strongly depends on #1842, as mentioned earlier. [Traditionally](https://macromates.com/manual/en/language_grammars#naming_conventions), `support` types were used for names provided by the language itself, including constants such as Python's `True` or `None`. While the [ST scope naming docs](https://www.sublimetext.com/docs/scope_naming.html#support) say that `support.{type,class}` may also be used for user-defined names, I personally haven't used a syntax definition doing that. 

In fact, running a grep on the entire Packages repos show `support.type` to be used for:
- Rust: selected types
- CSS: all properties (arguably a special case)
- JavaScript: provided objects from multiple environments, e.g. node modules or browser objects
- TypeScript: special types and type variables
- Go for built-in types (stacked with `storage.type`)
- Scala: any type
- C#: any non-primitive type
- ASP: selected types? (unsure)
- C++: selected types
- Java: namespaces in paths ???
- Python: selected types
- YAML: tag definitions
- Erlang: unsure

`support.class`: to be done later

wbond:
Just as a procedural note, I probably am not going to have the time to hash out RFC scope names before we get to an ST4 public release. I do want to address them at some point in the future, but they don't make the priority cut at this point.

FichteFoll:
Added results for `support.class` to my comment above.

Python/syntax_test_python.py
deathaxe: I am uncertain here. I've created a little test file with

```python
from typing import Tuple

def testing(results):
    for a, b in results:  # type: Tuple[int, str]
        print(a, b)
```

Hovering `a` or `b` with LSP and LSP-pylsp installed, doesn't show anything, while an example using current syntax with or without parentheses works fine.

```python
def testing(results):
    for a, b in results:  # type: (int, str)
        print(a, b)
```

It also makes sense as I found some comments about such type hints to date back to python 2.7 which doesn't know about `typing` module.

deathaxe: The related spec can be found at https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#declaring-multiple-variable-types-at-a-time

PR #1927, opened 2019-03-26
[JavaScript] Recognize line endings in block comments.

Thom1729:
Comments in JavaScript are usually not tokenized, but block comments containing newlines are tokenized as LineTerminators. This has implications for constructs that care about LineTerminators, like `break`, `continue`, `return`, `throw`, and `async function`.


FichteFoll:
I'm not even sure I want to know how you encountered this.

wbond:
Thank you @Thom1729!

PR #1932, opened 2019-03-31
[PHP] correctly unindent `else if:` followed by `?>`

### LINKED ISSUE #1924: [PHP] wrong if else indentation
**Version info**
 - OS: Arch Linux
 - Build: 3176

**Description**
Sublime text indentation wrongly indents else branch of php if/else construct.

**Steps to reproduce**
- Create new file and set php syntax
- Fill in
```
<?php if (true): ?>
asdfadf
<?php else if():?>
asdf
<?php endif;?>
```
- select reindent from edit menu

**Expected behavior**
First and third line should start on the same column

keith-hall:
correctly unindent `else if:` followed by `?>` - fixes #1924


wbond:
Thanks @keith-hall!

PR #1937, opened 2019-04-04
[Lua] Syntax number fixes

Resike:
Expression block fails for numbers like:
```
if number > .2 then

else

end
```

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -399,13 +399,13 @@ contexts:
       pop: true
 
   number:
-    - match: (0[Xx])\h*(?:\.\h*)?([Pp][-+]?\d*)?
+    - match: (0[Xx])\h*
       scope: constant.numeric.hexadecimal.lua
       captures:
         1: punctuation.definition.numeric.hexadecimal.lua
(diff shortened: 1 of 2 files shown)


Thom1729:
The [manual](https://www.lua.org/manual/5.3/manual.html#3.1) implies that numerals cannot begin with leading decimal points. However, an [online repl](https://repl.it/languages/lua) seems to allow it. Can you clarify this?

Resike:
I think the manual is outdated, they probably allowed it to ignore the starting zero with one of the later versions.

wbond:
We should probably add a test for the leading decimal form, right @Resike?

wbond:
Thanks @Resike!

FichteFoll:
For confirmation, in case someone is curious like me:

```
~ ∃ lua
Lua 5.3.5  Copyright (C) 1994-2018 Lua.org, PUC-Rio
> .2
0.2
> 0xAB.CD
171.80078125
```

Resike:
I see, i didn't found anything about it in the manual, and i mostly work with Lua 5.1. You guys should revert that hex deletion then.

PR #1943, opened 2019-04-20
[Ruby] Fix hash keys with reserved keywords still being marked incorrectly

### LINKED ISSUE #1934: [Ruby] Hash keys with reserved keywords are still marked incorrectly
While `begin, end, and require` are fixed from https://github.com/sublimehq/Packages/issues/53, the issue still exists for `class and module`.
subl_key_issue

Github highlighting for comparison:

github_key_issue

### Build info
Sublime Text Build 3200

SilverStrange:
Fixes 'class:' and 'module:' being marked as keywords instead of symbols.
Adds test cases for this issue to the syntax test.

Fixes #1934

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -87,9 +87,9 @@ contexts:
     - include: expressions
 
   expressions:
+    - include: constants
     - include: class
     - include: module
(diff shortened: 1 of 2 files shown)


wbond:
Thank you @SilverStrange!

PR #1944, opened 2019-04-20
[Makefile] Fix line continuation in comments

### LINKED ISSUE #1938: [makefile] Adding a trailing backslash (no spaces after it) to a comment end, comments the next line
```make
all:
#	echo I am a comment \
	echo I am a comment, too!
	echo I am no comment...
```

image

Running this outputs:
image

Exception to the rule:
```make
all:
	var="string"; \
#	echo I am a comment \
	echo I am no comment...
```

Which outputs:
image

deathaxe:
Fixes #1938


wbond:
Thanks for the fix and the reviews!

Having people chime in and approve changes definitely helps me get through these faster.

PR #1945, opened 2019-04-20
[Makefile] Fix highlighting of unbalanced quotations in functions

### LINKED ISSUE #1941: [Makefile] Broken highlighting for nested subst
image

Source code:
```make
#--------------------------- Python Script Runner ----------------------------#

define \n


endef

escape_shellstring = $(subst `,\`,$(subst ",\",$(subst $$,\$$,$(subst \,\\,$1))))
escape_printf = $(subst \,\\,$(subst %,%%,$1))
create_string = $(subst $(\n),\n,$(call escape_shellstring,$(call escape_printf,$1)))
python_script = printf "$(call create_string,$($(1)))" | python

#------------------------------- User Scripts --------------------------------#

define my_script

def good_times():
	print "good times!"

import time
print 'Hello world! $0'
print 'Hello world! $1'
print '%d + %d = %d' %($2,$3,$2+$3)
print 'Running $$BEEF \ttarget \t\n\n"%s" at time %s' %('$4', time.ctime())
good_times()

endef

#--------------------------------- Recipes -----------------------------------#

test:
	$(call python_script,my_script,1,2,$@)
```

Running it:
image

Source:
1. https://unix.stackexchange.com/questions/222911/using-embedded-python-script-in-makefile

deathaxe:
Fixes #1941

### Issue Description

A quotation mark found in the arguments of a function-call causes the parser to push into the quoted-string context. The region up to the next quotation mark is then scoped as quoted string.

But make interprets single quotation marks in functions like `subst` as ordinary characters.

The following escapes all quotation marks in $1:

    $(subst ",\", $1)

### Solution Proposal

This commit proposes to just eat escaped quotation marks (\", \') or those followed by the arguments separator.

They are scoped as whatever the parent context scopes the current region. The example posted in issue #1941 causes the quotation marks to be scoped as `string.unquoted`.

### DIFF
--- a/Makefile/Makefile.sublime-syntax
+++ b/Makefile/Makefile.sublime-syntax
@@ -411,6 +411,9 @@ contexts:
       push: textual-parenthesis-balancer
     - match: \,
       scope: punctuation.separator.makefile
+    # eat escaped or unbalanced quotation marks in front of the `,` separator
+    # to highlight them as ordinary part of the surrounding unquoted string
+    - match: '\\[''""]|[''""](?=\s*,)'
     - include: variable-substitutions
     - include: quoted-string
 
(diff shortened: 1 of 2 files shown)


wbond:
Sounds like a reasonable work-around - thanks!

PR #1948, opened 2019-04-28
[Perl] Fix regexp patterns at line start with tabs

deathaxe:
This commit fixes an issue with regexp patterns not being detected, if tabs are used for indention.

**Example:**

    $var =~
    /pattern/;

The `/pattern/` is now detected as such, even if the line starts with tabs.

~_Note: Can't add a test as PackageDev complains about tabs being used as indention._~

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -279,7 +279,7 @@ contexts:
     - include: function-call
     - include: comment-line
     # /<find>/<flags> statement at line start
-    - match: ^ *(?=/.*[^\\]/)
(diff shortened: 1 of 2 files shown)


FichteFoll:
PackageDev should only complain when the setting to use tabs is active for a view and it also shouldn't prevent you from adding one after the popup has surfaced once. If switching to use tabs for indentation, adding a tab and then switching back does not work with PackageDev, please add your use case to https://github.com/SublimeText/PackageDev/issues/177.

wbond:
Thanks!

PR #1950, opened 2019-04-29
[JavaScript] Support private property accessors.

### LINKED ISSUE #1949: Class fields
Fix from #1292 doesn't work (/cc @Thom1729)

```js
class MongoDB_Wrapper extends BaseWrapper {
  #data = getDataObject();

  fetchData(params = []) {
    let num = 0;
    for (const param of this.#data.get('value'))
      if (param === null) return -1;
      else num += param.size;
    return num;
  }
}
```

We need fast fix for this. Hope someone can resolve it soon.

Thom1729:
Fix #1949.

The original implementation in #1292 did not properly support private property accessors, but in most cases they worked almost-correctly by coincidence. #1949 identifies a case in which they do not work at all. This PR implements them correctly.

Private fields are part of the [TC39 class fields proposal](https://github.com/tc39/proposal-class-fields), which is Stage 3.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -2264,6 +2264,11 @@ contexts:
     - match: '{{identifier_part}}+{{identifier_break}}'
       scope: invalid.illegal.illegal-identifier.js
       pop: true
+    - match: (#)({{identifier}})
+      captures:
+        1: punctuation.definition.variable.js
+        2: meta.property.object.js
+      pop: true
 
   support-property:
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

rwols:
What about private methods?

```js
class SomeClass {
    #somePrivateMethod(data) {
        return "foo";
    }
}
```

image

I'm seeing this at the company I work at right now.

Thom1729:
Fixed in #2476. Thanks for reporting.

PR #1956, opened 2019-05-01
[Perl] Fix control keyword scoping

deathaxe:
This PR

1. fixes an issue with `elsif` not being scoped at all. The `elseif` was instead, which turned out not to be a valid keyword for that purpose in Perl.
2. fixes an issue with labels after `next` or `redo` keywords not being recognized by pushing into a context to scope the first identifier after the keyword as `variable.label.perl`.
3. adds detailed `keyword.control` keywords as discussed in https://github.com/sublimehq/Packages/issues/1228
4. adds several test cases to ensure keywords are scoped correctly
5. reorganizes some test cases to group `keywords`, `conditional` and `loop` related tests.


wbond:
It seems in the process of merging these various Perl PRs a conflict is now present. Do you have a minute to sort it out?

wbond:
Thank you @deathaxe!

PR #1959, opened 2019-05-01
[Perl] Fix comments

deathaxe:
This PR ...

1. removes the `begin` sub scope from the punctuation to be conform with other syntaxes.
2. includes the trailing `\n` into the `comment` scope as other syntaxes do in order to prevent completions from being triggered while editing a new comment line with the caret at the eol.

_Note: The changes this PR makes to the syntax tests causes a little conflict with https://github.com/sublimehq/Packages/pull/1952_


wbond:
It seems this PR will need a little conflict resolution also

PR #1961, opened 2019-05-02
[PHP] PHP 7.4 new syntax

jfcherng:
PHP 7.4 has been feature-freezed so this PR will no longer add new things.

## PHP 7.4 New Syntaxes

- [X] [Arrow function](https://wiki.php.net/rfc/arrow_functions_v2)
- [X] [Null coalescing assignment operator `??=`](https://wiki.php.net/rfc/null_coalesce_equal_operator)
- [X] [Typed properties](https://wiki.php.net/rfc/typed_properties_v2)
- [X] [Spread Operator in Array Expression `...`](https://wiki.php.net/rfc/spread_operator_for_array)
- [X] [Numeric Literal Separator `_`](https://wiki.php.net/rfc/numeric_literal_separator)

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -222,7 +222,7 @@ contexts:
       scope: punctuation.terminator.expression.php
     - include: heredoc
     - include: numbers
-    - match: (\+|-|\*|/|%|&|\||\^|>>|<<|\.)=
+    - match: (\+|-|\*|/|%|&|\||\^|>>|<<|\.|\?\?)=
       scope: keyword.operator.assignment.augmented.php
     - match: "=>"
       scope: keyword.operator.key.php
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -150,6 +150,7 @@ contexts:
           pop: true
     - match: \b(catch|try|throw|finally)\b
       scope: keyword.control.exception.php
+    - include: arrow-function
     - include: closure
     - include: invoke-call
     - match: |-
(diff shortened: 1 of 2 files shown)


willrowe:
The spread operator within array definitions was just accepted. I'm not sure if the syntax already supports it or not.

https://wiki.php.net/rfc/spread_operator_for_array

jfcherng:
@willrowe Thanks. Added :+1:

jfcherng:
Add support for [Numeric Literal Separator `_`](https://wiki.php.net/rfc/numeric_literal_separator) :100:

jfcherng:
There is no further new feature which will be added to PHP 7.4. This PR has **done**.

PHP/PHP Source.sublime-syntax
```diff
@@ -541,6 +544,24 @@ contexts:
...
         - match: ','
           scope: punctuation.separator.php
+    # https://wiki.php.net/rfc/typed_properties_v2
+    - match: (?=\b(?:public|private|protected|static)\b)
+      push:
+        - match: (?=\bfunction\b)
+          pop: true
+        - match: \b(?:public|private|protected|static)\b
+          scope: storage.modifier.php
+        - match: (?=(?:\?\s*)?{{identifier}})
```
FichteFoll: Why push into a separate context? You could as well assing the scopes in this match and pop, since you're not doing anything meta.

jfcherng: > Why push into a separate context?

Because I want it to do a deterministic parsing. For example,

```php
public MY_TYPE $name = '......';
public  ?MY_TYPE $name = '......';

// I want to assign a scope to "?" and especially "MY_TYPE"
```

If I don't going to a new context, I thought there is no way to know "MY_TYPE" is a constant or a class?

```php
public const MY_CONST = '...'; // public is can be omitted
```

jfcherng: Caught a bug that `(?=\bfunction\b)` should be `(?=\b(?:const|function)\b)`.

FichteFoll: I specifically meant the context that is set through this match on line 554 where you already know to assign `support.class.php` because that's what you do later inside the context with only two patterns.

jfcherng: Ah. I guess I now understand your word. Awesome catch!
So now it's like https://github.com/sublimehq/Packages/pull/1961/commits/443c71e9247705f01e5fe10862b8f1cc4828c3c2 .

FichteFoll:
Yes, that's what I meant. You didn't add a `pop: true`, however, and it didn't cause any test failures. Perhaps a test for that case is needed?

jfcherng:
> Yes, that's what I meant. You didn't add a `pop: true`, however, and it didn't cause any test failures. Perhaps a test for that case is needed?

**Without** the `pop: true`, 
image
would be successfully highlighted. It just becomes a little non-deterministic in that way, while previously we pops out immediately. Would you suggest that we should be more deterministic in this situation?

FichteFoll:
I based my comment on the previous behavior. I usually prefer deterministic behavior, unless you need to accomodate highlighting while typing, in which case the answer is different on a case-by-case basis.

I doubt you'll end up with multiple types "accidentally" while typing, so I'd go for the deterministic approach and add the pop.

PHP/PHP Source.sublime-syntax
```diff
@@ -545,21 +545,18 @@ contexts:
         - match: ','
           scope: punctuation.separator.php
     # https://wiki.php.net/rfc/typed_properties_v2
-    - match: (?=\b(?:public|private|protected|static)\b)
+    - match: (?=\b(?:var|public|private|protected|static)\b)
```
jfcherng: Related to https://github.com/sublimehq/Packages/pull/1962

wbond:
Excellent, thank you!

PR #1965, opened 2019-05-07
[Matlab] Fix transpose highlight following fields

### LINKED ISSUE #1647: [Matlab] Transpose comma as string comma
After transpose comma system is recognizing all lines as string. This can be solved using if there is no new comma on the same line or after three dot on the next line this comma is transpose.

scshoot

jwortmann:
Fix #1647.

This pull request adds a check for a transpose comma after fields of structures using dot notation `a.b'` to prevent false highlighting as a string, see [this comment](https://github.com/sublimehq/Packages/issues/1647#issuecomment-483157721). Previously, the rule of the context `transpose` did not apply to `a.b'` because it requires a variable name (`{{id}}`) preceding the transpose comma, but `a.b` was already matched by the rule of context `members`. Therefore I included the transpose comma check, which is already used after parens/brackets/braces, to matches of the form `a.b`.

Detecting the transpose comma will also work for nested structures of the form `a.b.c`, `a.b.c.d` etc. now, even though these are not entirely matched by the context `members`.

### DIFF
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -29,6 +29,7 @@ contexts:
     - match: ({{id}})(\.)({{id}})
       captures:
         2: punctuation.accessor.dot.matlab
+      push: transpose_post_parens
   all_matlab_comments:
     - match: (%%).*$\n?
       scope: comment.double.percentage.matlab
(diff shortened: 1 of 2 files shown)


wbond:
@michaelblyons Would you have a moment to look this over?

deathaxe:
After reading the matlab docs for `ctranspose()` as kindly provided by @jwortmann at https://github.com/sublimehq/Packages/issues/1647#issuecomment-516893909 this PR looks good to me.

_The overall implementation of the transpose operator could be improved and various contexts `transpose`, `variable` and `member` may be combined using the `transpose_after_parens`, but this was a separate issue._

wbond:
Thanks for this @jwortmann! This will be included when the next dev cycle starts.

PR #1966, opened 2019-05-09
[D] Fixes

### LINKED ISSUE #1909: [D] Freeze due to variable name beginning with "function"
Create a D file:
```d
void f() {
	int x;
	int functionA = 0;
}
```

Place the cursor after `x` and before `;` and type ` =`. The editor will freeze. Appears to be happening because a variable name begins with `function`.

### LINKED ISSUE #1963: [D] syntax highlighting bug
This is valid D code:
```d
extern int variable;
```
but ST highligts `extern` keyword as error:
D language highlighting bug

ST3 version 3.2.1, Build 3207

### LINKED ISSUE #1969: [D] integer hex literal syntax highlighting bug
```D
int i = 0xff;
```
will display the last `f` in italics like a floating point literal suffix.

image

### LINKED ISSUE #1972: [D] qualified name in `immutable ref A.B` in delegate highlighted as error
At this text:
```d
(immutable ref Outer.Inner inner) => inner.x
```
The `.` in `Outer.Inner` is highlighted as an error.
There is the same error if the ref is `const ref`, but not if it's a mutable `ref`. Relatedly, the color of the `ref` keyword changes based on whether `immutable` is present.
There is no such error in normal parameter lists, only for delegates.

Whole example program:
```d
import core.stdc.stdio : printf;

static struct Outer {
	static struct Inner {
		int x;
	}
}

// Sytax highlighting is fine here. Only a problem at the call site.
int callIt(immutable ref Outer.Inner inner, int delegate(immutable ref Outer.Inner) cb) {
	return cb(inner);
}

extern(C) void main() {
	immutable Outer.Inner inner = Outer.Inner(42);
	int i = callIt(inner, (immutable ref Outer.Inner inner) => inner.x);
	printf("%d\n", i);
}
```

### LINKED ISSUE #1973: [D] `pure Pair!int*` breaks syntax highlighting
Whole example program:
```
import core.stdc.stdio : printf;

struct Pair(T) {
	T first;
	T second;
}

pure Pair!int* makePair(int x) {
	return new Pair!int(x, x);
}

extern(C) void main() {
	printf("%d\n", makePair(1).first);
}
```

The `x` in `int x`, as well as the opening and closing curly braces after it, are highlighted as errors.
No such error if any of:
* `pure` is not present
* `Pair` is not a template and `!int` is removed
* It just returns `Pair!int` with no pointer.

BenjaminSchaaf:
Fixes #1963
Fixes #1909
Fixes #1969
Fixes #1972
Fixes #1973

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -532,6 +532,9 @@ contexts:
         1: keyword.other.external.d
         2: punctuation.section.parens.begin.d
       set: extern-attribute
+    - match: '\b(extern)\b'
+      scope: storage.modifier.external.d
+      pop: true
     - match: '\b(align)\b'
       scope: keyword.other.alignment.d
       set:
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -1437,7 +1437,7 @@ contexts:
     - match: '\b(alias)\b'
       scope: keyword.control.alias.d
       push:
-        - match: '{{attribute_lookahead}}'
+        - match: '{{type_qualifier_lookahead}}'
           set: [alias-old, type]
         - match: '\b({{name}})\s*(?=\(|=)'
           captures:
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -23,6 +23,7 @@ variables:
   integer_float_suffix: '[fF]i?|i'
   float_suffix: '[fFL]i?|i'
   float_exponent: '[eE][+\-]?[0-9]+'
+  hex_float_exponent: '[pP][+\-]?[0-9]+'
   character_lookahead: (?=')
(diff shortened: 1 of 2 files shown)


dcarp:
From the forum.dlang.org, here are 2 resources for testing the D language syntax highlighting:
https://github.com/colorer/Colorer-schemes/blob/master/hrc/test/other/d-sample.d
https://github.com/Emacs-D-Mode-Maintainers/Emacs-D-Mode/tree/master/tests

BenjaminSchaaf:
Thanks @dcarp, though some of those tests did help improve error handling they didn't actually find any additional edge cases.

wbond:
Looks good to me!

PR #1974, opened 2019-05-15
[Python] Handle f-string escape sequences with other escapes

FichteFoll:
Previously, `\{{` imposed problems with both regex raw strings and
normal due to the `\{` being matched first and the final `{` then being
interpreted as the opening of an f-string. We now match the entire token
and scope it appropriately.

Before: 

After:


wbond:
Gotta love escaping mixed in with formatting!

PR #1978, opened 2019-05-20
[Lua] Recognize long strings in table constructors

### LINKED ISSUE #1977: [Lua] Incorrect highlighting for long comments within table constructors
The following match causes brackets within a table constructor to be interpreted as key declarations, even if the character is actually apart of a long string:

https://github.com/sublimehq/Packages/blob/8c76dd38b2c0049247ec3f57c5b47e07ab52b178/Lua/Lua.sublime-syntax#L489-L497

Example result:

lua-syntax

FichteFoll:
Fixes #1977.


wbond:
Thank you

PR #1979, opened 2019-05-24
[OCaml] Added snippet for OCaml `if`

lwolfsonkin:


lwolfsonkin:
bump

wbond:
I think generally we are seeing snippets as less useful for the default packages. Clearly we already have some, but there are usually disagreements about what should be in them.

Because of this, I'm not very inclined to merge in new snippets, although I would accept bug fixes to existing snippets. Probably the best place for snippets moving forward is Package Control.

wbond:
I'm closing this since I think the best path forward is to encourage users to create a snippets package for a syntax if they want to see more options.

It *may* be that in the future snippets are downplayed by default, or removed from the default packages altogether.

PR #1982, opened 2019-06-04
[JavaScript] Scope builtin errors as support.class.builtin.

### LINKED ISSUE #1980: [JavaScript] Error classes should also have "builtin" scope
Currently, `Error` classes in JavaScript have the `support.class.error.js` scope:

https://github.com/sublimehq/Packages/blob/8c76dd38b2c0049247ec3f57c5b47e07ab52b178/JavaScript/JavaScript.sublime-syntax#L2055-L2056

However, since they are built-in global classes, I feel like they should also have the `support.class.builtin.js` scope, similar to these classes:

https://github.com/sublimehq/Packages/blob/8c76dd38b2c0049247ec3f57c5b47e07ab52b178/JavaScript/JavaScript.sublime-syntax#L2052-L2053

Thom1729:
Fix #1980.

The `.error` scope doesn't seem useful (it's not standard), so I removed it when adding `.builtin`.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -2053,7 +2053,7 @@ contexts:
       scope: support.class.builtin.js
       pop: true
     - match: (?:Eval|Range|Reference|Syntax|Type|URI)?Error{{identifier_break}}
-      scope: support.class.error.js
+      scope: support.class.builtin.js
       pop: true
 
     - match: (?:eval|isFinite|isNaN|parseFloat|parseInt|decodeURI|decodeURIComponent|encodeURI|encodeURIComponent){{identifier_break}}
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -2053,7 +2053,7 @@ contexts:
       scope: support.class.builtin.js
       pop: true
     - match: (?:Eval|Range|Reference|Syntax|Type|URI)?Error{{identifier_break}}
-      scope: support.class.error.js
+      scope: support.class.builtin.js
```
deathaxe: Would it make sense to merge the pattern with the former one then?

Thom1729: It would be compiled the same either way, so it shouldn't really matter.

PR #1983, opened 2019-06-06
[JavaScript] Add meta.string scopes.

Thom1729:
Also extend a few related tests. In addition to being correct, should fix Thom1729/Sublime-JS-Custom#60.


Thom1729:
Agreed that we should change the interpolation scopes to the standard, but I think that can be done separately.

wbond:
Look good

PR #1985, opened 2019-06-06
[Makefile] add just file extensions

ganziqim:
see <https://github.com/casey/just#syntax-highlighting>

### DIFF
--- a/Makefile/Makefile.sublime-syntax
+++ b/Makefile/Makefile.sublime-syntax
@@ -12,6 +12,8 @@ name: Makefile
 file_extensions:
   - make
   - GNUmakefile
+  - justfile
+  - Justfile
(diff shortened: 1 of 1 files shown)


FichteFoll:
Taking a brief look at just's feature set, I actually believe it makes more sense to provide a separate syntax for it. Especially with how it allows for e.g. Python or Ruby target scripts or how it replaces variables.

I is similar to Makefile, but I doubt it's a good idea to open it with the Makefile syntax by default for all ST instances.

wbond:
Based on @FichteFoll's review of this I don't think we should merge this in as a default extension. You can accomplish this for yourself via an override or telling ST to open all files with the extension with this syntax.

PR #1987, opened 2019-06-08
[Perl] variable reference operator

michaelblyons:
Scope `\` preceding user-defined variables. See https://perldoc.perl.org/perlref.html#Making-References

Also allow accessor chaining three-deep, and use `variable.language` rather than `variable.other.predefined`.

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -1231,13 +1231,16 @@ contexts:
       captures:
         1: punctuation.definition.variable.perl
       push: maybe-item-access
-    - match: ([\$\@\%]#?)\w+\b
+    - match: (\\)?([\$\@\%]{1,3}#?)(\w+)\b
       scope: variable.other.readwrite.global.perl
       captures:
-        1: punctuation.definition.variable.perl
+        1: keyword.operator.reference.perl
+        2: punctuation.definition.variable.perl
       push: maybe-item-access
-    - match: ([\$\@\%])(\{)
-      scope: punctuation.definition.variable.begin.perl
+    - match: (\\)?([\$\@\%]{1,3}\{)
+      captures:
+        1: keyword.operator.reference.perl
+        2: punctuation.definition.variable.begin.perl
       push:
         - meta_scope: meta.braces.perl variable.other.readwrite.global.perl
         - match: \}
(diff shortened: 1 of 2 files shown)


deathaxe:
Thank's for your attention. Missing reference operator is a good point.

I also tend to put too much into PRs, but I think it's not optimal to put the two topics 

- rescoping `variable.other.predefined`
- fixing reference operator

into one PR, because a closer look at the latest versions of https://perldoc.perl.org/perlvar.html and https://perldoc.perl.org/perlref.html reveals both topics to need some more work.

Commit 467dc56 does not handle all examples about references mensioned at perldoc.perl.org/perlref.html.

```Perl
@list = \($a, @b, %c);      # same thing!

$refrefref = \\\"howdy";
print $$$$refrefref;

$funref = \function_name
```

Commit 7bcf77a just rescopes some existing patterns to `variable.language`, while a closer look at https://perldoc.perl.org/perlvar.html reveals much more stuff like `$PID` or `$PROCESS_ID` and many more would need to be scoped as `variable.language` as well. _Looks like the docs have got heavy updates ;-) since my last visit during Perl Rewrite._

Furthermore 7bcf77a adds test cases for the former commit, which feels a bit mixed up.

---

I'll take a closer look at both topics to see how to fix them, without breaking my sophisticated local efforts to implement string interpolation support into Perl. I did not yet publish it, because it would need all the other Perl PRs to be merged first in order to avoid conflicts.

michaelblyons:
@deathaxe Sounds good. I'm also a little hesitant to use `$a` and `$b` as special variables. They _are,_ but only under a specific context. In my (very very limited) Perl experience, it's at least as likely that they're user-defined.

wbond:
@michaelblyons It appears that with merging all of the other Perl PRs this has gotten conflicted. Would you have a minute to sort it out?

deathaxe:
After some investigation I think the whole "references topic" needs a little bit more work. Testing this PR against the liked gist in my previous post reveals some of the missed edge cases.

With all the other PRs being merged now, I can check which parts to "publish" next.

michaelblyons:
> I also tend to put too much into PRs, but I think it's not optimal to put the two topics
> 
>     * rescoping `variable.other.predefined`
>     * fixing reference operator
> 
> into one PR

Agreed. I'm splitting the first bit off into #2030 


> After some investigation I think the whole "references topic" needs a little bit more work. Testing this PR against the [linked] gist in my previous post reveals some of the missed edge cases.

It seems very likely that I missed all kinds of cases. I'm happy to close this in favor of a more robust PR.

Issue #1988, opened 2019-06-09
Get rid of snippets directories in default language packages


kylebebak:
A much easier way of resolving the issue that motivated this issue in the ST Core repo: https://github.com/SublimeTextIssues/Core/issues/901

<img width="856" alt="Screen Shot 2019-06-09 at 10 34 20 AM" src="https://user-images.githubusercontent.com/1524088/59160977-29ac3f00-8aa2-11e9-855f-8c1dffb17c44.png">

The default language packages should not ship with snippets. Languages change, and the snippets get outdated. People don't write JS the way they did in 2012. Most of these JS snippets are useless and/or bizarre.

I think whoever has write access to the default language packages ought to just remove snippets from all of them.

kylebebak:
Also, I originally created this issue in the wrong repo, so I'm going to reproduce comments from that thread:

From @frou
>I agree that some dood coming up with snippets they personally thought were kinda neat, is not a high enough bar for those to be in everyone else's face forevermore.

>I mean why is there a [goofy wordplay](https://github.com/sublimehq/Packages/blob/master/Go/Snippets/go-defun.sublime-snippet) on Common Lisp terminology in the Go package? Whoever comes up with these things should just keep them to themselves in their User package.

from @keith-hall 

>We had a small discussion about this on Discord about a month ago. Some functionality should definitely still be included by default - not necessarily in the form of snippets though - like the `html` <kbd>Tab</kbd> snippet. Perhaps that one could be replicated with a keybinding that triggers on <kbd>Tab</kbd> if the syntax at the caret is plain text or html and would insert the html template (which would have to be provided as a package resource I guess, unless we can still bundle snippets but without tab triggers by default). It would ofc mean that users relying on snippets being available in the command palette or more easily discoverable through autocompletion would have their workflow interrupted. Or maybe just better to vote on which snippets to remove or keep?

frou:
This could be resolved like what happened with the [legacy Colour Schemes](https://packagecontrol.io/packages/Color%20Scheme%20-%20Legacy):

Move all of the code snippet files out of this repo and put them into a single "Snippets - Legacy" package that can be easily installed by those who still want them.

kylebebak:
Here's a PR that gets rid of Python and JavaScript snippets to get the ball rolling: #1989 

>Move all of the code snippet files out of this repo and put them into a single "Snippets - Legacy" package that can be easily installed by those who still want them.

[...]

PR #1989, opened 2019-06-09
Removes JS and Python snippets

kylebebak:
I code regularly in JS and Python. I can say with confidence that the default JavaScript snippets shipped with ST are useless and/or bizarre. They seem like they were written in 2012, and JS has changed a lot since then.

One of several `for` loop snippets in the `JavaScript` package:

```js
for (var i = 0; i < Things.length; i++) {
  Things[i]
}
```

Why `var` instead of `const` or `let`? Why this style of for loop instead of `for (const <> of <>)`?

-----

The Python snippets are also mostly useless. For example, the `for` loop:

```py
for x in xrange(1,10):
    pass
```

Why `xrange` instead of `range`? Why no space after the comma in `xrange`? Why from 1 to 10? Like, seriously, **what is this???**

-----

Even if they were less useless and/or bizarre, I think forcing the same set of snippets on all ST users is a bad design choice, especially because snippets are difficult to disable, and **impossible to disable in a way that requires no future maintenance**.

Discussion on the broader idea of removing all snippets from language packages here: #1988


michaelblyons:
I've used both those snippets in the last two weeks. The critical thing is the tabstops, not the defaults (change `xrange(1,10)`  to `range(5)` if you like. I'd support that).

I don't want to start a holy war on which TC39 loop structure someone ought to use, but the one in the snippet has worked since 1995 and still works now. 🤷‍♂

kylebebak:
Rest assured, I'm no holy warrior either. That's why I created this PR... Because even if all the snippets were "great" (which it turns out is very subjective), the main point still stands:

>I think forcing the same set of snippets on all ST users is a bad design choice, especially because snippets are [difficult to disable](https://stackoverflow.com/questions/27995926/how-do-i-disable-snippets-in-sublime-text-3), **and impossible to disable in a way that requires no future maintenance**

Snippets should be defined in user land. There's more discussion in the linked issue.

Thom1729:
It would probably be best to keep the discussion to #1988 to avoid fragmentation. As you say, the quality/relevance of the snippets is very subjective; for any given snippet, some users will likely find it useful and others will find it annoying. So while this PR is specific to Python and JavaScript, it's likely that any discussion of the PR would be applicable to the broader issue.

This is entirely up to you, but it might also be helpful to close this PR for now. That might help to avert the perception of “taking away” features that some users rely on and help to center the discussion instead on how to make snippets work for everyone.

kylebebak:
@Thom1729 
You've convinced me closing it for the time being is a good idea. Looking forward to hearing what you think about it in the #1988

PR #1990, opened 2019-06-11
[Rust] Fix next_result handling for Cargo build errors

jskinner:
Paths in error messages are relative to the top-level directory, so ensure the
working_dir is set to that.

This also sets Cargo.sublime-build to be active when there's a Cargo.toml file,
rather than when editing Rust source code, which ensures it's an available build
system in more circumstances (e.g., when editing Cargo.toml)


wbond:
Thanks! 😉

PR #1991, opened 2019-06-11
[C++] Add support for Javadoc style comments

DavidWang1024:
Allows Javadoc style comments to be recognized in scope. Word wrapping depends on it to wrap comment blocks properly.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -204,6 +204,8 @@ contexts:
           - match: \*/
             scope: punctuation.definition.comment.c
             pop: true
+          - match: ^\s*\*(?!/)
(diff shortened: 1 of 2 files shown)


C++/C++.sublime-syntax
```diff
@@ -204,6 +204,8 @@ contexts:
           - match: \*/
             scope: punctuation.definition.comment.c
             pop: true
+          - match: ^\s*\*(?!/)
```
FichteFoll: The scope should only cover the asterisk and not the leading whitespace, imo.

DavidWang1024:
This is for the default wrap command.

I have made changes to the scope to only cover the asterisk.

PR #1992, opened 2019-06-11
[D] Add support for Javadoc style comment

DavidWang1024:
Leading asterisk/plus in multi line comment will now be treated as
punctuation definition. Word wrapping depends on it to wrap comment blocks properly.


FichteFoll:
See #1991.

PR #1994, opened 2019-06-12
[JavaScript] Improve scope for Javadoc style comments

DavidWang1024:
Allow leading asterisks (excluding leading white space) in body of Javadoc style comments to be recognized as "punctuation.definition" scope. Default word wrapping will depend on it to wrap comment blocks properly.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -110,7 +110,7 @@ contexts:
         - match: \*/
           scope: punctuation.definition.comment.end.js
           pop: true
-        - match: ^\s*(\*)?(?!/)
+        - match: ^\s*(\*)(?!/)
           captures:
             1: punctuation.definition.comment.js
     - match: /\*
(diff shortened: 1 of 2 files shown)


Thom1729:
I am of two minds.

As far as JavaScript is concerned, JavaDoc-style comments are no different from any other kind of block comment. The additional asterisks have no syntactic significance. In general, we try to avoid adding nonstandard third-party behavior into syntax definitions. We've removed a lot of it from the JavaScript syntax already. If JavaDoc-style comments weren't already in the syntax, I would oppose adding them; I think that this is best left to [third-party tools](https://github.com/Thom1729/Sublime-JS-Custom). There's an argument to removing them entirely — but that's rather out of scope of this PR.

On the other hand, JavaDoc-style comments are a fairly common convention, even among different commenting systems. As it stands, the special handling of `/**` isn't that narrow of an assumption. Is this new assumption substantially narrower? I'm not sure.

If the motivation for changing so many syntax definitions is to improve word wrapping behavior, is there a plausible alternative?

wbond:
@Thom1729 This is part of some upcoming changes to word wrapping. @DavidWang1024 is an engineer at Sublime HQ, and this is the public part of some work he has been doing recently.

PR #1997, opened 2019-06-12
Improve Javadoc style comment scope for various applicable languages

DavidWang1024:
Allow leading asterisks (excluding leading white space) in body of Javadoc style comments to be recognized as "punctuation.definition" scope. Default word wrapping will depend on it to wrap comment blocks properly.

Applied to ActionScript, C, C#, CSS, Go, Graphviz, PHP, SQL, Scala


C#/C#.sublime-syntax
```diff
@@ -79,6 +79,9 @@ contexts:
         - match: '\*/'
           scope: punctuation.definition.comment.end.cs
           pop: true
+        - match: ^\s*(\*)?(?!/)
```
keith-hall: as the capture group is the only thing that receives a scope, why make it optional? it may make it harder for future maintainers to understand the purpose of consuming the leading whitespace...

PR #2001, opened 2019-06-25
[C#] include structs in symbol list

mollstam:
I finally figured out why some of my types weren't in the Goto Symbol in Project… list: they are structs!

I went ahead and looked at the Classes and Interfaces files and added one for Structs, lo and behold it just worked and I know too little about Sublime development to know what effects this has, except that now I can finally jump around to my structs, so I thought I'd make a PR for it. 😬 

Let me know if there's something you need from me. ❤️ 👏

### DIFF
--- a/C#/Symbol List Structs.tmPreferences
+++ b/C#/Symbol List Structs.tmPreferences
@@ -11,10 +11,6 @@
 		<integer>1</integer>
 		<key>showInIndexedSymbolList</key>
 		<string>1</string>
-		<key>symbolTransformation</key>
-		<string>
-		s/^/struct /;
-		</string>
 	</dict>
 </dict>
 </plist>
--- a/C#/Symbol List Structs.tmPreferences
+++ b/C#/Symbol List Structs.tmPreferences
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>name</key>
+	<string>Symbol List: Struct</string>
+	<key>scope</key>
+	<string>source.cs entity.name.struct</string>
+	<key>settings</key>
+	<dict>
+		<key>showInSymbolList</key>
+		<integer>1</integer>
+		<key>showInIndexedSymbolList</key>
+		<string>1</string>
+		<key>symbolTransformation</key>
+		<string>
+		s/^/struct /;
+		</string>
+	</dict>
+</dict>
+</plist>


C#/Symbol List Structs.tmPreferences
```diff
@@ -0,0 +1,20 @@
...
+	<string>Symbol List: Struct</string>
+	<key>scope</key>
+	<string>source.cs entity.name.struct</string>
+	<key>settings</key>
+	<dict>
+		<key>showInSymbolList</key>
+		<integer>1</integer>
+		<key>showInIndexedSymbolList</key>
+		<string>1</string>
+		<key>symbolTransformation</key>
```
wbond: We generally don't prefix the symbol type in the Goto Symbol list, so I'd like to see this and the next three lines removed.

mollstam: Consider it done

wbond:
All merged, thanks!

mollstam:
@wbond Thanks for the review, how do you feel about the symbol transformations for classes, enums, interfaces, and namespaces?
Screenshot 2019-07-30 at 13 40 20

IMO even having the namespace in the symbol list feels a bit.. polluted... 🤔

mollstam:
Well then, another day another PR I guess.

Thanks a lot. ❤️

michaelblyons:
@mollstam At some point, can you give me your thoughts on #1425?

wbond:
@mollstam Those should probably be removed also if they are there now. If you submit a PR I'd be happy to merge.

Thom1729:
@wbond 

> There are definitely some syntaxes that try to add fake syntax around the symbol to imply the type of the symbol, but I'd rather work on trying to get type info into the symbol list via the Sublime Text code base rather than adding more text that may make the matching less accurate.

Would this apply to the JavaScript syntax as well? Given the following code:

```js
function foo() {}
const xyzzy = () => 42;

({
    bar() {}
    baz: () => 42
});
```

The symbol list is:

- `function foo()`
- `xyzzy = () =>`
- `bar()`
- `baz: () =>`

There's a fair bit of special-casing to achieve that behavior. Should it be simplified?

djspiewak:
> There are definitely some syntaxes that try to add fake syntax around the symbol to imply the type of the symbol, but I'd rather work on trying to get type info into the symbol list via the Sublime Text code base rather than adding more text that may make the matching less accurate.

I'd like this as well, though it's worth noting that in my experience, the fake text actually makes matching *more* accurate. As an example, the Scala mode does this, and it helps a lot with this very common scenario:

```scala
class Foo { ... }
object Foo { ... }
```

These are called "companion objects" and they're highly ubiquitous, meaning that performing a symbol search for `Foo` is almost always going to yield two possibilities. In the global scope, there's not much that we can do about this, but within a file they will show up as `class Foo` and `object Foo`, allowing me to type `Foo obj` and rapidly narrow down to just the object.

This would be impossible without the fake syntax. I still agree that direct support within Sublime itself would be better, but there are certainly advantages to the way it is now.

mollstam:
> […] but within a file they will show up as `class Foo` and `object Foo`, allowing me to type `Foo obj` and rapidly narrow down to just the object.
> 
> This would be impossible without the fake syntax. I still agree that direct support within Sublime itself would be better, but there are certainly advantages to the way it is now.

If the type information goes into Sublime properly instead of simply being baked into the symbol name it would still be possible to take it into account when filtering. The data would just live in a more first-class place. Or have I misunderstood? In the end, functionality wise, it could be made identical as it is now, right?

PR #2002, opened 2019-06-27
[Rust] Add u128 and i128 primitive types.

willstott101:
Since https://github.com/rust-lang/rust/pull/38482 Rust has had native `u128` and `i128` integers.

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -9,7 +9,7 @@ variables:
   identifier: '(?:(?:[[:alpha:]][_[:alnum:]]*|_[_[:alnum:]]+)\b)' # include a word boundary at the end to ensure all possible characters are consumed, to prevent catastrophic backtracking
   escaped_byte: '\\(x\h{2}|n|r|t|0|"|''|\\)'
   escaped_char: '\\(x\h{2}|n|r|t|0|"|''|\\|u\{\h{1,6}\})'
-  int_suffixes: 'i8|i16|i32|i64|isize|u8|u16|u32|u64|usize'
+  int_suffixes: 'i8|i16|i32|i64|i128|isize|u8|u16|u32|u64|u128|usize'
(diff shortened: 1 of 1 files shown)


wbond:
Could we add a test assertion or two to make sure we don't regress on these in the future?

FichteFoll:
Personally, I find it superfluous to add tests for these kind of patterns with lots of options and where you only added one or two, because when following this argument you'd also have to have a test for every `support.type` identifier and that's overkill imo.

willstott101:
I could replace a few of the existing integer suffixes in the test file with 128bit suffixes.

I could add completely new tests for all suffixes.

Or perhaps via support for several occurrences of the same syntax in one comment I could make more of a matrix of supported formats.

      1i8; 1i16; 1i32; 1i64; 1i128;
    //^^^  ^^^^  ^^^^  ^^^^  ^^^^^ constant.numeric.integer.decimal
    // ^^   ^^^   ^^^   ^^^   ^^^^ storage.type - constant.numeric.integer.decimal

But without access to the test code I have no idea how sensible that is.

wbond:
> Personally, I find it superfluous to add tests for these kind of patterns with lots of options and where you only added one or two, because when following this argument you'd also have to have a test for every support.type identifier and that's overkill imo.

I think my approach is that is we missed this for a year or more and it had to be reported and fixed, it may be rare enough that someone wouldn't think to look for it.

Generally, if I see a fix with tests and it isn't complicated I can just merge. If there are no tests than more effort is spent trying to figure out why there are no tests, and if we really need them. Maybe the author hasn't spent time reading the [docs for .sublime-syntax](https://www.sublimetext.com/docs/3/syntax.html#testing) and doesn't know about the tests?

FichteFoll:
>I think my approach is that is we missed this for a year or more and it had to be reported and fixed, it may be rare enough that someone wouldn't think to look for it.

That's a good point, considering this is a nightly-only feature and also behind a feature flag, people aren't going to use it a whole lot. I didn't even know about these types.
My usual practice is to pick a couple examples from a pattern set like this, which should give appropriate coverage on them, but I don't have any set rules.

Thom1729:
Devil's advocate: why not test every single support identifier?

PR #2003, opened 2019-07-01
[JavaScript] Allow node-specific variables as valid identifiers

Hakerh400:
**Before**

**After**

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -2193,6 +2193,7 @@ contexts:
     # Module-level variables
     - match: (?:__dirname|__filename|exports){{identifier_break}}
       scope: support.constant.node.js
+      pop: true
     - match: module{{identifier_break}}
       scope: support.constant.node.js
       set:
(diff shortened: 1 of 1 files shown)


Thom1729:
The code change in this PR seems obviously correct. That rule should definitely pop, and it's surely a bug that it wasn't before.

However, I'm confused by the screenshots. After `const`, the syntax should try to match a variable binding pattern, not a variable use. I don't know why it would hit any of the support code unless there's another bug.

I'm on vacation until Monday, so I can't test this at the moment.

FichteFoll:
Support identifiers are matched in assignments.
```yaml
  variable-binding-name:
    - match: (?={{non_reserved_identifier}})
      set:
        - meta_scope: meta.binding.name.js
        - include: literal-variable
```

literal-variable => support =>

There is another pop missing in line 2287:
```yaml
    - match: (?:__defineGetter__|__defineSetter__|__lookupGetter__){{identifier_break}}
      scope: invalid.deprecated.js support.function.js
```

Thom1729:
Weird. I should probably rewrite that part. In the mean time, this PR looks correct, and adding that other missing pop would be even better.

wbond:
Thanks @Hakerh400!

PR #2004, opened 2019-07-01
[Git] End gitconfig strings before newline

michaelblyons:
I have a background color tweak for `string` that makes the problem more obvious.

### DIFF
--- a/Git Formats/Git Config.sublime-syntax
+++ b/Git Formats/Git Config.sublime-syntax
@@ -211,7 +211,7 @@ contexts:
           pop: true
 
   line-end:
-    - match: $\n?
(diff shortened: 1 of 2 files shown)


Git Formats/Git Config.sublime-syntax
```diff
@@ -211,7 +211,7 @@ contexts:
           pop: true
 
   line-end:
-    - match: $\n?
+    - match: (?=$)
```
FichteFoll: For the record, this is equivalent to `$`.

michaelblyons: A good point, thanks.

PR #2006, opened 2019-07-10
[Python] Add support for some new syntax in Python 3.8

FichteFoll:
[Python 3.8](https://docs.python.org/3.8/whatsnew/3.8.html) will include the following syntax changes:

- [x] `f'{expr=}'` expands to the text of the expression, an equal sign, then the repr of the evaluated expression (https://bugs.python.org/issue36817)
- [x] Positional-Only Parameters (https://www.python.org/dev/peps/pep-0570/)
- [ ] Assignment Expressions (https://www.python.org/dev/peps/pep-0572/)

I skipped assignment expressions for now because they are a tad bit more involved and ideally I'd like to see some consensus on #1842 regarding the scope names I should use for the identifiers that are assigned to. Because the assignments are inline, it would seem useful for those to be highlighted differently compared to normal variable usages.
Anyway, here are the small changes I've done so far.

The final release is [scheduled for 2019-10-21](https://www.python.org/dev/peps/pep-0569/).

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -624,6 +624,13 @@ contexts:
     - match: ','
       scope: punctuation.separator.parameters.python
       push: allow-unpack-operators
+    - match: /
+      scope: storage.modifier.positional-args-only.python
+      push:
+        - match: (?=[,)])
+          pop: true
+        - match: \S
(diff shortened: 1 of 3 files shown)


wbond:
Looks good, thanks!

PR #2010, opened 2019-07-13
[JavaScript] Scope builtin errors with `builtin`.

### LINKED ISSUE #1980: [JavaScript] Error classes should also have "builtin" scope

Thom1729:
Fix #1980.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -2053,7 +2053,7 @@ contexts:
       scope: support.class.builtin.js
       pop: true
     - match: (?:Eval|Range|Reference|Syntax|Type|URI)?Error{{identifier_break}}
-      scope: support.class.error.js
+      scope: support.class.builtin.js
       pop: true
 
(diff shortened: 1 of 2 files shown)


FichteFoll:
Why not go for `support.class.builtin.error.js`?

Thom1729:
I thought about that, but I'm not sure that the extra `error` scope is really useful. If you think it is, I can certainly add it back.

FichteFoll:
I mean, it wouldn't hurt. I could see use in it for coloring error/exception classes.

It's a consideration for #1842 as well. Which I'll hopefully get to structure into a concrete proposal eventually.

Thom1729:
Apparently this PR is an exact duplicate of #1982.

PR #2011, opened 2019-07-20
[JavaScript] Narrow prototype matching in utility contexts.

Thom1729:
This PR removes the prototype from a bunch of “utility” contexts, such as contexts providing meta scopes. This fixes some minor quirks where comments could get meta scopes they probably shouldn't, and it also reduces the total number of rules that the syntax has to match. In addition, this should make it much easier for [third party packages based on the core JavaScript syntax](https://github.com/Thom1729/Sublime-JS-Custom) to implement comment-based template string highlighting (e.g. ``/* sql */`select * from dual` ``).


Thom1729:
Looks like the PHP syntax is broken; it should be using `with_prototype`, but it's doing something else instead that just happened to work for the exact test case. I'll have to fix that.

Thom1729:
#2012 fixes the bug in the PHP syntax that's causing the CI failure.

PR #2012, opened 2019-07-21
[PHP] Fix JavaScript heredocs.

Thom1729:
It looks like the language-specific heredocs in PHP are a bit brittle. I've fixed the JavaScript one here because it was breaking #2011. The rest of these should probably be fixed as well, but I'm not familiar with the PHP syntax.


wbond:
This was merged in as part of #2011.

PR #2013, opened 2019-07-22
[Markdown] Support Pandoc's fenced code block format

jfcherng:
Support syntax highlight for Pandoc's fenced code block notation.

~~~~~~
``` {.bash}
CLIST=$(schroot -l | sed -n 's,^source:\(.*\-sbuild\)$,\1,p')
for ins in $CLIST; do sbuild-update -udcar $ins; done
```
~~~~~~

This is the first time that I see this kind of fenced code block notation though... but it's quite easy to support it from the current syntax definition.

References:

- https://pandoc.org/MANUAL.html#fenced-code-blocks
- https://forum.sublimetext.com/t/markdown-syntax-highlighting-misses-soem-code-blocks/45514

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -104,6 +104,7 @@ variables:
             (~){3,}    #   3 or more tildas
             (?![^~]*~) #   not followed by any more tildas on the same line
           )
+          (?:\s*\{\.)? # Pandoc's notation: https://pandoc.org/MANUAL.html#fenced-code-blocks
           \s*          # allow for whitespace between code block start and info string
         )
     fenced_code_block_trailing_infostring_characters: |-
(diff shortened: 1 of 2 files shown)


wbond:
Is there a mainstream Markdown engine that supports this syntax?

jfcherng:
> Is there a mainstream Markdown engine that supports this syntax?

Tried on some popular online markdown editor. None of them supports.

image

I read pandoc's docs again. It looks like it's just used to represent the rendered output HTML has a class attribute of `bash`. So it could be just a extended markdown by pandoc.

wbond:
If it isn't part of CommonMark and none of the big apps/sites support it, I don't think we should include it, as it would likely just lead to confusion. (Users seeing the syntax marked up like it is valid, when it really isn't.)

FichteFoll:
Oh, then I retract my approval of this.

jfcherng:
Agree since it's not common and non-standard.

PR #2015, opened 2019-07-23
[Rust] Restructure macro parsing

### LINKED ISSUE #2014: [Rust] Syntax highlighting bug with macro_rules
## Code
```rust
macro_rules! vec {
    ($($x:expr), *) => {
        {
            let mut v = Vec::new();
            $(v.push($x);)*
            v
        }
    };
}

fn main() {
    let v = vec![1, 2, 3];
    println!("{:#?}", v);
}
```

## Result
It matches the wrong `*` 
图片

## What it should be
(I removed a whitespace character so that it could be syntax highlighted correctly)
图片

FichteFoll:
- Use a deterministic approach by pushing all expected tokens at once.
- Clean up closing punctuation matches to make them easier to
  understand.
- Recover from unmatched tokens more gracevully through 'else-pop'
  context.
- Accept whitespace around multi-match separator (Rust compiles fine
  with them).
- Add more details to scope names.
- Match comma as macro matcher separator.

Fixes #2014.

Before:

After:

```rust
#[macro_export]
macro_rules! vec {
    ( $( $x:expr ) , * ) => {
        {
            let mut v = Vec::new();
            $(v.push($x);)*
            v
        }
    };
}

#[macro_export]
macro_rules! test [($($x:tt),*) => ($x)];

fn main() {
    let v = vec![1, 2, 3];
    println!("{:#?}", v);
}
```

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -23,6 +23,14 @@ contexts:
     - match: '\${{identifier}}'
       scope: variable.other.rust
 
+  else-pop:
+    - match: (?=\S)
+      pop: true
+
+  pop-immediately:
+    - match: ''
+      pop: true
+
   statements:
 
     - match: '(''{{identifier}})\s*(:)'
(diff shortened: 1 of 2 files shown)


wbond:
Thank you for the improvements in robustness

PR #2019, opened 2019-07-30
[JavaScript] Use proper scopes for interpolation in template strings.

Thom1729:
- `string.template` → `string.quoted.other`
- `punctuation.definition.template-expression` → `punctuation.section.interpolation`
- `meta.template.expression` → `meta.interpolation`

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -937,18 +937,18 @@ contexts:
       scope: punctuation.definition.string.begin.js
       set:
         - meta_include_prototype: false
-        - meta_scope: meta.string.js string.template.js
+        - meta_scope: meta.string.js string.quoted.other.js
         - match: "`"
           scope: punctuation.definition.string.end.js
           pop: true
         - match: '\$\{'
-          scope: punctuation.definition.template-expression.begin.js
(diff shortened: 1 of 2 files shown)


michaelblyons:
> Let's hope we ...

I also have a mental note to patch [BracketHighlighter](https://packagecontrol.io/packages/BracketHighlighter) when this (or something like it) is merged. /backlink #1432

leeoniya:
IMO, semantic conformance/uniformity should have priority over maintaining BC for color schemes.

FichteFoll:
I was mostly referring to the past times anything regarding standard scope compliance was done to the JavaScript syntax because it seems that this is ST's loudest user base when it comes to things looking differently. It may also be the largest for all I know. In a not so serious manner (through the xkcd comic I referenced). 😉 Probably should've added a smiley for clarification.

Past occurences of this include removing special highlighting of "dollar variables" and adding a string scope to object key literals.

leeoniya:
i have a custom js color scheme. as long as the scope specificity does not regress in a way that makes it impossible to tweak the scheme to attain the desired result, i dont have an issue making adjustments. in many cases better semantic uniformity means that it can be *simplified*.

wbond:
I think this makes sense early in a dev cycle. Let's do it!

PR #2020, opened 2019-07-30
[JavaScript] Use meta.mapping instead of meta.object-literal

Thom1729:
- `meta.object-literal` &rarr; `meta.mapping`
- `meta.object-literal.key` &rarr; `meta.mapping.key`

Also added the missing `meta.mapping.key` to computed property names.


JavaScript/tests/syntax_test_js.js
```diff
@@ -218,10 +218,10 @@ let x = import.meta;
 // method body, but we include it here to ensure that highlighting is not
 // broken as the user is typing
 let a = { otherIdentifier, foo(), baz: 1 }
-//      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.object-literal
+//      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.mapping
```
FichteFoll: This wouldn't also satisfy `meta.mapping - meta.mapping meta.mapping`, would it? Because I'm not seeing a `clear_scopes` in this diff, especially not for `object-literal-meta-key`.

Thom1729: Good point. One issue is that there's an extra meta scope around the key and value when the value is a function:

```js
+{
    funcKey: function() {
//  ^^^^^^^^^^^^^^^^^^^ meta.mapping meta.function.declaration
//  ^^^^^^^ meta.mapping meta.function.declaration meta.mapping.key entity.name.function
    },
}
```

Wrapping the whole thing like that in `meta.function` is admittedly dubious, but it's used for the symbol list. Do you have any ideas on how to address this?

Thom1729: @FichteFoll Any thoughts on this?

wbond: I think this is reasonable for now. The pattern of naming functions via object key is so pervasive that we definitely wouldn't want to lose that functionality.

wbond:
It looks like this has some assertion failures

JavaScript/tests/syntax_test_js.js
FichteFoll: Well, you could pop two scopes for the mapping key and re-push `meta.mapping.key meta.function entity.name` (and do the same for the value), but unfortunately scope mangling like this can get extremly annoying when a `set` is involved.

There is a solution to this, but having `meta.mapping` is *already* an improvement over `meta.object-literal`.

PR #2021, opened 2019-07-30
[XML] Highlight files starting with DOCTYPE as XML

deathaxe:
Apply XML syntax to all files starting with `<!DOCTYPE >`, if it is not html. `<!DOCTYPE html>` is reserved for HTML syntax.

Sometimes the <?xml ...> prolog may be omitted in XML files which may also have an arbitary extension such as `*.ts` which is not associated to XML otherwise.

### DIFF
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -13,11 +13,13 @@ file_extensions:
   - opml
   - svg
 first_line_match: |-
-    (?x)
-    ^(?:
-        <\?xml\s
-     |  \s*<([\w-]+):Envelope\s+xmlns:\1\s*=\s*"http://schemas.xmlsoap.org/soap/envelope/"\s*>
-     )
+    (?x:
+      ^(?:
+          <\?xml\s
+        | \s*<([\w-]+):Envelope\s+xmlns:\1\s*=\s*"http://schemas.xmlsoap.org/soap/envelope/"\s*>
+        | \s*<!DOCTYPE\s+(?!html\b)
+      )
+    )
 scope: text.xml
 variables:
   # The atomic part of a tag or attribute name without namespace separator `:`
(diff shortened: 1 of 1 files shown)


XML/XML.sublime-syntax
```diff
@@ -13,11 +13,13 @@ file_extensions:
...
-    (?x)
-    ^(?:
-        <\?xml\s
-     |  \s*<([\w-]+):Envelope\s+xmlns:\1\s*=\s*"http://schemas.xmlsoap.org/soap/envelope/"\s*>
-     )
+    (?x:
+      ^(?:
+          <\?xml\s
+        | \s*<([\w-]+):Envelope\s+xmlns:\1\s*=\s*"http://schemas.xmlsoap.org/soap/envelope/"\s*>
+        | \s*<!DOCTYPE\s+(?!html\b)
```
Thom1729: `\b` should probably be `(?=[ \t\n\f>])`. See https://html.spec.whatwg.org/multipage/parsing.html#doctype-name-state.

deathaxe: Good catch.

deathaxe: PS.: If I recall correctly, nested positive lookaheads don't work within other lookaheads.

XML/XML.sublime-syntax
```diff
@@ -17,7 +17,7 @@ first_line_match: |-
       ^(?:
           <\?xml\s
         | \s*<([\w-]+):Envelope\s+xmlns:\1\s*=\s*"http://schemas.xmlsoap.org/soap/envelope/"\s*>
-        | \s*<!DOCTYPE\s+(?!html\b)
+        | \s*<!DOCTYPE\s+(?!html[ \t\n\f>])
```
jfcherng: we need this to be case-insensitive?

deathaxe:
Oups. Wrong commit title.

Otherwise, thanks!

Issue #2017, opened 2019-07-27
HIghtlight problem when prefix derefference for Perl


KES777:
For the code from #1495 
```
sub x {
  my( $c ) =  shift;
  my $ctrl =  $c->stash->{ control } //=  {};
  my $id = scalar keys %{ $ctrl->{ hello } };
}
```

One problem is left for `%{ ... }` syntax:
![_3](https://user-images.githubusercontent.com/799021/61993375-bd969200-b073-11e9-8f41-872bc6fceae1.jpg)

`Modern Perl` syntax highlight OK:
![_2](https://user-images.githubusercontent.com/799021/61993391-f5053e80-b073-11e9-9a0f-a081634aafa2.jpg)

PS. this is Color scheme -> legacy -> Espresso libre

deathaxe:
The strange looking colors you see are caused by an unlucky choise of scopes. The whole `%{...}` is scoped as `variable.other`, which causes everything not assigned to any other scope to be highlighted the same. _I already replaced that by `meta.variable` in an internal dev branch to fix that issue._

Are `stash`, `control` and `hello` really functions as ModernPerl scopes them as such? The real meaning of `stash` depends on the object type, I think, while `control` and `hello` are dictionary keys?

KES777:
`stash` is function. Word after `->` always the function.  `control` and `hello` are interpreted as strings inside `{ ... }`
If I rewrite  `->{ +control }` now `control will be interpreted as function (code is interpreted as expression)

deathaxe:
@KES777: You may checkout #2048. It should fix this issue next to many more.

PR #2026, opened 2019-07-30
[Perl] Add unquoted hash keys highlighting

deathaxe:
In order to access an item of a hash (dictionary), the caller can use an unquoted string as key.

In the other way round: The `item_key` in `$the_hash{item_key}` is a simple unquoted string.

Prior to this commit, the key was not scoped at all and would so as function/method call in the future.

_Note: This fix is one result of the discussion in issue #2017._

### DIFF
--- a/Perl/syntax_test_perl.pl
+++ b/Perl/syntax_test_perl.pl
@@ -767,6 +767,9 @@ EOT
 # ^ punctuation.definition.variable.perl
 #  ^^ punctuation.accessor.double-colon.perl
 #    ^^^^^^ support.class.perl
+#          ^ punctuation.section.item-access.begin.perl
+#           ^^^^^^^^^^ string.quoted.single.perl
+#                     ^ punctuation.section.item-access.end.perl
   -f
 # ^^ keyword.operator.filetest.perl
   -foo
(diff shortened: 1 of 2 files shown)


Perl/syntax_test_perl.pl
```diff
@@ -787,6 +787,39 @@ =head1 B<--param>
...
+  $foo{10 + $bar}
+# ^^^^ variable.other.readwrite.global.perl
+# ^ punctuation.definition.variable.perl
+#     ^ punctuation.section.item-access.begin.perl
+#     ^^^^^^^^^^^ meta.item-access.perl
+#      ^^ constant.numeric.integer.decimal.perl
+#         ^ keyword.operator.arithmetic.perl
+#           ^^^^ variable.other.readwrite.global.perl
+#               ^ punctuation.section.item-access.end.perl
+  %{$foo{bar}{baz}} = 'excl';
```
FichteFoll: Wouldn't you want a test on `bar` or `baz` here as well? Otherwise LGTM.

deathaxe: Reasonable. Missed that in the quoted counter part as well.

PR #2028, opened 2019-07-31
[Batch File] Set up indexing for label "calls"

michaelblyons:


Batch File/Batch File.sublime-syntax
```diff
@@ -27,8 +27,14 @@ contexts:
     - include: escaped_characters
     - include: variables
     - include: commands
-    - match: \b(?i)(?:goto|call|exit)\b
+    - match: \b(?i)exit\b
       scope: keyword.control.statement.dosbatch
+    - match: \b(?i)(goto|call)\s*(:)(?:(eof)|(\w+))
```
FichteFoll: This will remove highlighting for plain `goto` while typing. I suggest either adding a fallback or making the latter parts of the match optional.

michaelblyons: Good catch, thanks.

wbond:
Since the symbol list and symbol index scopes are effectively equivalent, we can probably merge the two `.tmPreferences` files.

PR #2030, opened 2019-08-01
[Perl] Rescope language variables

michaelblyons:
I'm splitting this out of #1987

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -1207,7 +1207,7 @@ contexts:
         1: punctuation.definition.variable.perl
       push: maybe-item-access
     - match: (\$)0\b
-      scope: variable.other.predefined.program-name.perl
+      scope: variable.language.program-name.perl
       captures:
         1: punctuation.definition.variable.perl
       push: maybe-item-access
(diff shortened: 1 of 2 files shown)


deathaxe:
OK, just renaming the scopes shouldn't hurt too much, but like I already tried to state indirectly:

I am working on a couple of improvements for the Perl Syntax including:

- add support for interpolated strings ( which requires the following fixes)
- add support for the `\` reference operator
- add support for the `&` coderef operator
- add support for most of the language variables defined in the _English.pm_ module.
- improve support for crazy multiple levels of references like `%&$*#var{key}`
- correctly support function and method calls with argument lists with and without parentheses
- matching class, constants, functions, methods with capital and small letters as the heuristic to check for CAPITALONLY = constant etc. seems insufficient. There are classes like `URI` which are matched as constant atm.

All of those changes come with quite tricky edge cases which interfere with the other commits in the one or the other way. 

E.g.: Can't fixing function calls without modifying the way, classes and packages are matched. Same for the coderef operators. They interfere with functions and methods. Even might require changes to variables depending on whether the `&` is to be scoped as punctuation in `%&$*#var{key}` or as dedicated reference operator, ... .

That said, this PR will be superseded, but forcing me to modify all current changes (test cases I've addes so far) and the #1987 is incomplete in manner of supporting various levels of references like `\\\&coderef`.

May I therefore kindly ask to concentrate on syntaxes, which are not otherwise maintained?

PR #2033, opened 2019-08-02
Fix D multiline-comments not ending without content

BenjaminSchaaf:
`/**/` and `/++/` would cause the comment to never end

### DIFF
--- a/D/tests/syntax_test_d.d
+++ b/D/tests/syntax_test_d.d
@@ -36,11 +36,15 @@ module foo.a.b1_3;
       +
 //    ^ comment.block.nested.d punctuation.definition.comment.d
  +/
+/++/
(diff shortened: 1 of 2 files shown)


D/tests/syntax_test_d.d
```diff
@@ -36,11 +36,15 @@ module foo.a.b1_3;
       +
 //    ^ comment.block.nested.d punctuation.definition.comment.d
  +/
+/++/
+//^^ comment.block.nested.d punctuation.definition.comment.d
```
wbond: Perhaps add a test right after this with a `- comment` assertion so we are sure it is terminating?

BenjaminSchaaf: Isn't that what the rest of the file is for ;)

PR #2035, opened 2019-08-02
[Perl] Fix fully qualified sub routine identifier

deathaxe:
Each variable or function identifier can consist of a fully qualified path.

Example:

    $NS1::NS2::variable
    NS1::NS2::function

The `NS` is expressed by the file path relative to the core library or can be extended by sub routine definitions.

Example:

    sub NS1::NS2::NS3::function { }

   A real life example can be found in <..>/core_perl/B/Debug.pm

Before this commit fully qualified subroutine identifiers are scoped `invalid.illegal`. This is fixed by this commit.

Note:

  The namespace part is currently scoped as `support.class`, because
  a) this is the scope being used in other situations as well
  b) the scope naming guidelines don't yet suggest a scope for that part.

   - C# uses `variable.other.namespace`.
   - The rewritten Erlang introduces `variable.namespace` because it fits best to the `entity.name.namespace` which is to be used for namespace definitions.
   - The proposed general `variable.qualifier` would probably the most general alternative as it is hard/impossible to distinguish namespace from class access. Would require `entity.name.qualifier` as definition counterpart then?

   c) All `support.class` scopes should be renamed in one commit later.

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -1117,7 +1117,7 @@ contexts:
   sub:
     - match: \bsub\b
       scope: storage.type.function.perl
-      push: [sub-meta, sub-path]
+      push: [sub-meta, sub-parameters, sub-path]
     # special functions which are executed at compile time
     # SEE: https://perldoc.perl.org/perlmod.html#BEGIN%2c-UNITCHECK%2c-CHECK%2c-INIT-and-END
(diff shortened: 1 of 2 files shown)


deathaxe:
Close this for the moment as this commit will be added to a PR which handles all aspects of fully qualified identifiers in Perl, including function definition, function call, variable access.

PR #2036, opened 2019-08-03
[Perl] Fix ${...} meta scope name

deathaxe:
Up to this commit all `${...}` or `%{...}` tokens are scoped as `variable.other` while they contain ordinary expressions.

This commit...

1. replaces the scope by a `meta.variable` to avoid otherwise unmatched parts of the expressions being highlighted in a wrong way.
2. adds the `regexp-pop` context right after the opening punctuation in order to fix an issue with `/patter/flags` not being detected.

Note:
  This change is part to solve issue #2017

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -1238,17 +1238,17 @@ contexts:
       push: maybe-item-access
     - match: ([\$\@\%])(\{)
       scope: punctuation.definition.variable.begin.perl
-      push:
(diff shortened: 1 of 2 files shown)


deathaxe:
The expression inside the braces is used to build the variable identifier. Even though this kind of expression can be used everywhere it is a core feature to properly support string interpolations.
   
```Perl
    $name = "variable";
    $my_variable === ${"my_" . $name}
```

PR #2039, opened 2019-08-04
[Perl] Restrict whitespace surrounding namespace accessors

deathaxe:
A fully qualified identifier must not contain whitespace. Hence the accessor `::` must not be surrounded by them as well.

It was supported up to this point in order to improve the writing experience by not breaking identifiers during writing.

This will cause conflicts with parsing/scoping fully qualified identifiers in future changes an therefore needs to be restricted.

The test cases containing whitespace surrounded accessors are removed for the moment as properly handling such situations is part of a future change.

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -16,7 +16,7 @@ first_line_match: |-
 scope: source.perl
 
 variables:
-  break: (?!\w|\s*::)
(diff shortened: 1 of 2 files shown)


deathaxe:
Close this for the moment as this commit will be added to a PR which handles all aspects of fully qualified identifiers in Perl, including function definition, function call, variable access.

PR #2040, opened 2019-08-04
[Perl] Add file extensions

deathaxe:
This commit adds

- `pc`   = was found in the core_perl/Win32API/File library
- `pmc`  = mentioned in https://perldoc.perl.org/functions/require.html
           The importer loads pmc files if present in place of a
           requested pm file.

and removes:

- `PL`   = duplicate of `pl`.

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -3,11 +3,12 @@
 # http://www.sublimetext.com/docs/3/syntax.html
 name: Perl
 file_extensions:
+  - pc
   - pl
   - pm
+  - pmc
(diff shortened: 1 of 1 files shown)


Perl/Perl.sublime-syntax
```diff
@@ -3,11 +3,12 @@
 # http://www.sublimetext.com/docs/3/syntax.html
 name: Perl
 file_extensions:
+  - pc
```
wbond: I believe the order of these affects the save dialog filter options on Windows, in which case we'd want `.pl` to come first.

PR #2041, opened 2019-08-04
[C#] Removed type prefixes from symbol list

mollstam:
As discussed in #2001 this PR removes the `class`, `enum`, `interface`, and `namespace` prefixes inserted via symbol transformation. It's nice to have but really something that should be taken care of by Sublime Text in a more coherent manner.

🙅‍♂ **Before**
Screenshot 2019-08-04 at 17 49 18

🙆‍♂ **After**
Screenshot 2019-08-04 at 17 51 05

### DIFF
--- a/C#/Symbol List Classes.tmPreferences
+++ b/C#/Symbol List Classes.tmPreferences
@@ -11,10 +11,6 @@
 		<integer>1</integer>
 		<key>showInIndexedSymbolList</key>
 		<string>1</string>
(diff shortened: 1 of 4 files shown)


FichteFoll:
Because there is nothing inside ST that allows to differentiate the types of symbols in the symbol list, I'd rather not remove these annotations because they do help imo.

wbond:
@FichteFoll The next dev cycle will include some changes to help address the issue you raised.

PR #2043, opened 2019-08-04
[Scala] Scope overhaul to better align with convention

### LINKED ISSUE #1986: [Scala] Unify scopes usage with other modes
Scope guidelines have been refined quite a bit in the last three years, and as a consequence, the Scala mode has fallen somewhat out of step. This has started getting to the point where a color scheme which looks good for Scala (such as my fork of material-dark) no longer look good on *other* modes (such as JavaScript). Additionally, color schemes which generally look good in other modes (such as Mariana) don't look as good in Scala. This is obviously quite bad.

Fortunately it should be pretty easy to fix this. It's just that things are a bit out of step right now, particularly in the `variable` scopes.

djspiewak:
This changes loads of stuff. I'm sorry for the mega PR. I kept the commits granular so you can see in a bit more fine-grained fashion what changed, but I can squish a lot of it together. This mostly boils down to making the scoping of most constructs align more consistently with other modes. In particular, I relied on JavaScript and C# as the primary canon, though I also looked at Java (NB: the Java mode scopes imports inconsistently and still has some unfortunate bugs).

Some big ticket bullets:

- I changed `val` and `var` scoping. They used to use `entity.name`, and now they use `variable.other`, so as to be consistent with JavaScript. I went back and forth on this *a lot*, because `val` is not materially different from a no-argument `def` in Scala, and they are conventionally used entirely interchangeably. For that reason, `val` and `var` are both included in the project symbol index. This indexing strategy has turned out to be a very good balance in practice on both small and large codebases, but it also puts the Scala mode in a unique position of attempting to scope the construct consistently and index it inconsistently. My solution to this was to scope the construct consistently with C# and JavaScript, but I added `variable.other.constant` and `variable.other.readwrite` to the symbol set. This is quite unusual, but I decided that unconventional indexing was better than unconventional scoping.
  + As a note, I'm happy to re-explain in detail why it is that `val` and `var` are indexed (if you recall, we discussed this at length several years ago). I would also be happy to entertain removing `var` from the index, though in practice I haven't found its presence to be a problem in any sense, and it *feels* better to not have to care about the difference between `val` and `var` while navigating.
- Loads and loads of punctuation scopes. I hadn't been using a color scheme which scoped punctuation, and that caused a lot of oversights. I believe I caught everything now. There's a bit of subtlety between `punctuation.definition` and `punctuation.section`, and I tried to be as reasonable as I could be. I tried to be particularly careful about this given the special coloring that Mariana gives to `punctuation.definition`.
  + As an aside, it makes me sad that `punctuation.separator` and `punctuation.accessor` are given special treatment by Mariana, but `punctuation.ascription` is not. I understand why, but it feels a bit odd in practice, and unfortunately the `:` operator in Scala just doesn't fit into any of these other buckets.
- Imports are completely redone and now are entirely uniform with C# and entirely *not* uniform with Java. The one wart I left here is scoping of `package` declarations, which are currently using `entity.name.namespace` for the *entire* fully-qualified name, dots-included. This is highly inconsistent, but I can't fix it without changing symbol indexing for this construct, since Scala supports fully-qualified *scoped* packages (similar to namespaces in C#, but with full qualification). That full qualification needs to be part of the symbol set, but that in turn means that we either need to *not* index `entity.name.namespace` and instead index some `meta` scope which covers the entities and the punctuation, *or* we need to scope the whole thing and the `.`s end up scoped as `entity.name` rather than `punctuation.accessor`. Right now I'm doing the latter, but I'm open to the former.
- Most Scala operators are entirely free-form and we can't scope them any more than we can scope random functions, but `!=` and `==` are special magic and cannot be overridden, and thus I feel justified in giving them scoping which is more uniform with other languages of more restricted operator sets. We were already doing this for `eq` and `ne` (also special magic), though there were some bugs.
- Some random bug fixes that I found

Fixes #1986

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -810,10 +810,7 @@ contexts:
           scope: punctuation.separator.scala
         - match: '(?=[\n;])'
           pop: true
-        - match: '{{upperid}}'
-          scope: support.class.import.scala
         - match: '{{id}}'
-          scope: support.type.package.scala
         - match: \.
           scope: punctuation.accessor.dot.scala
         - match: '_'
@@ -829,10 +826,7 @@ contexts:
               scope: punctuation.separator.scala
             - match: '{{rightarrow}}'
               scope: keyword.operator.arrow.scala
-            - match: '{{upperid}}'
-              scope: support.class.import.scala
             - match: '{{id}}'
-              scope: support.type.package.scala
             - match: '_'
               scope: variable.language.underscore.scala
 
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -367,13 +367,15 @@ contexts:
         2: storage.type.class.scala
         3: entity.name.class.scala
       push: class-inheritance-extends
-    - match: '\b(package)\s+({{id}}(?:\.{{id}})*)\s*\{'
+    - match: '\b(package)\s+({{id}}(?:\.{{id}})*)\s*(\{)'
       captures:
         1: keyword.control.scala
         2: entity.name.namespace.scoped.scala
+        3: punctuation.section.block.begin.scala
       push:
         - meta_scope: meta.namespace.scala
         - match: '\}'
+          scope: punctuation.section.block.end.scala
           pop: true
         - include: main
     - match: '\b(package)\s+({{id}}(?:\.{{id}})*)'
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -821,6 +821,8 @@ contexts:
             - match: "}"
               scope: punctuation.section.group.end.scala
               pop: true
+            - match: ','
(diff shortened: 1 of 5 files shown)


wbond:
From what I looked through, everything seemed fine. In a few places `punctuation.definition.group` and `punctuation.definition.block` were used, which feels odd, but you seem to be using them for types, which is like how the `punctuation.definition.generic` is used, so it isn't without precedence.

Thanks for your work on this!

djspiewak:
> From what I looked through, everything seemed fine. In a few places punctuation.definition.group and punctuation.definition.block were used, which feels odd, but you seem to be using them for types, which is like how the punctuation.definition.generic is used, so it isn't without precedence.

Exactly my thought process. I originally used `punctuation.section.group`, but it looked really strange in Mariana and didn't necessarily seem to be correct in a type context. `punctuation.definition` looked a lot better and was uniform with `punctuation.definition.generic`, so I went with it instead.

PR #2048, opened 2019-08-14
[Perl] Feature Update 2019

### LINKED ISSUE #1029: Perl highlighting issue in regexp

### LINKED ISSUE #1325: Perl highlighting issue in regexp with newer (perl 5.20) modifiers

### LINKED ISSUE #1326: Perl variable highlighting with variables that start with an _ (underscore) is broken

### LINKED ISSUE #1495: Perl. Syntax highlight. // operator as a comment

### LINKED ISSUE #2017: HIghtlight problem when prefix derefference for Perl

deathaxe:
Fixes #1325 (already fixed in 3207)
Fixes #1326 (already fixed in 3207)
Fixes #1029 (already fixed in 3207)
Fixes #1495 (already fixed in 3207)
Fixes #2017

## Preamble

I know the PR is way too big again. Wouldn't have expected this amount of changes at all, when I started to think about implementing string interpolation in spring. But the latest perldocs, further investigations about the Perl interpreter itself, comparison with _ModernPerl_ and some discussions in the forum and issue #2017 revealed a number of short commings with the current implementation, which prevented correct interpolation support. The assumptions about how functions and variables are identified are just not accurate enough and the resulting implementation too "simple" to match them all correctly. Need a little bit more context sensitive scoping to accomblish better results.

## Main Goals

This PR therefore includes all required changes and fixes to accomblish the following main goals:

1. Correctly scope unqualified and qualified case insensitive function identifiers and variables properly (including `meta.path` and `variable.namespace` scopes as a suggestion resulting from #1842).
All the different expressions were therefore tested against the Perl interpreter to make sure the added syntax test cases are correct.
2. Add support for string interpolation including `meta.string.perl meta.interpolation.perl` and clearing the `string` scope.
3. Improve the scoping of referencing and dereferencing operators and add missing `\` and `&`. 

## Fixes

- some comparison expressions were matched as angular quoted strings 
- qualified identifiers must not contain whitespace
- sub routine definitions can contain attributes like `sub name :codeattribute ...`
- sub routine parameter lists consist of `type keywords` like `$`, `@`, ... only. No names allowed.
- the `%` modulo operator was not scoped
- escaping from some regular expressions in like `s/pattern\\/` failed due to the `\\`.
- scoping the deprecated tab indention in POD comments as `invalid.deprecated` was somehow annoying.

## Benchmarks

The overall performance impact of the changes is negligible, except parsing of code with many interpolated strings increases depending on the number of interpolated variables.

The benchmark file which was used to check performance impact of all the changes contains many interpolated strings. The overall parsing time increased by about 15% but it is still only 20% of the time which is needed with _ModernPerl_.

## Known issues

The usage of the _Regular Expressions_ package prevents proper matching of interpolated variables in regulare expressions. Fixing it means implementing new regexp syntax for Perl.

## Notes

This PR includes #2032 and #2036 for technical reasons. I leave it to @wbond to merge them first or together with this PR.


Perl/syntax_test_perl.pl
```diff
@@ -1214,470 +2103,2419 @@ =head1 B<--param>
...
+#             ^ punctuation.definition.string.end.perl
+#              ^ punctuation.terminator.statement.perl
+  $foo{bar}
+# ^^^^ variable.other.readwrite.perl
+# ^ punctuation.definition.variable.perl
+#     ^ punctuation.section.item-access.begin.perl
+#     ^^^^^ meta.item-access.perl
+#      ^^^ constant.other.key.perl
+#         ^ punctuation.section.item-access.end.perl
+  $foo{bar()}
```
michaelblyons: I think `bar()` should also have `meta.function-call.perl` so that the hover popup where `bar` is defined can find the places it is used.

deathaxe: I think, I can re-add the `meta.function-call` to the function identifier, but it is not possible for the whole expression including the arguments list, because HEREDOCs can be passed as an argument and would cause the `meta.function-call` to never end.

PR #2051, opened 2019-08-15
[Makefile] Ignore nested define directives

### LINKED ISSUE #1998: [Makefile] Syntax highlighting doesn't support nested defines
At work we have to deal with a lot of legacy Makefiles with nested defines:
```
define FOO
  BAR := 1
  define BAZ
    X := 2
  endef
Y := 3
endef
```
In the example above the block up to the first `endef` is formatted as a plain "string" (as expected) but then `Y := 3` and the last `endef` switches to normal syntax highlighting. This happens in the latest stable build (3.2.1).

deathaxe:
Fixes #1998

### DIFF
--- a/Makefile/syntax_test_makefile.mak
+++ b/Makefile/syntax_test_makefile.mak
@@ -93,6 +93,18 @@ define $(dir)_print =
 lpr $($(dir)_sources)
 endef
 
+define FOO
+  BAR := 1
+  define BAZ
+# ^^^^^^ string.unquoted.makefile - keyword
+    X := 1
+  endef
+# ^^^^^^ string.unquoted.makefile - keyword
+Y := 3
+endef   # comment
+#^^^^ keyword.control.makefile
+#       ^^^ comment.line.number-sign.makefile
+
 #########################
 # 6.5 setting variables #
 #########################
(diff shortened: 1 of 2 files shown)


Makefile/syntax_test_makefile.mak
```diff
@@ -93,6 +93,18 @@ define $(dir)_print =
 lpr $($(dir)_sources)
 endef
 
+define FOO
+  BAR := 1
+  define BAZ
+# ^^^^^^ string.unquoted.makefile - keyword
+    X := 1
+  endef
```
rwols: Oof, how annoying!

deathaxe: What is annoying? The whole content to be a string?

rwols: The fact that this is actually allowed by the Make syntax, nothing against this PR mate :)

keithrussell42:
Thank you so much! 

@rwols I'm working in Buildroot: I've seen some pretty "interesting" examples of make syntax. I think the most fun after the nested defines was 4 levels of quoting like $$$$(FOO) to get variables to expand at the right time. I haven't seen 5 levels (yet) but I'm sure it's there somewhere.

PR #2058, opened 2019-08-22
[Ruby] Fix operators ligatures and scopes

### LINKED ISSUE #1387: [Ruby] Certain ligatures are not working
Some ligatures aren't working when using Ruby syntax. The ones I've found so far are `->` and `!=` but there may be others. I assume this is an issue with the syntax definition as they work fine in Plain Text syntax.

Here is a screenshot from Ruby syntax:
image

And here's what it looks like in Plain Text mode:
image

Potentially relevant info:
* I'm using build 3156 of Sublime.
* I'm using the font "MesloLG S" from [here](https://github.com/rojiani/Ligaturizer)

deathaxe:
Fixes #1387

This commit

1) sorts rules to correctly match all operators. This fixes some operators not being replaced by ligatures.
2) tweaks scope names a little bit.
   a) `|` and `~` are bitwise operators
   b) `..` is scoped `keyword.operator.range`

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -100,6 +100,7 @@ contexts:
     - include: comments
     - include: data-section
     - include: heredocs
+    - include: punctuation
     - include: operators
     - include: identifiers-accessors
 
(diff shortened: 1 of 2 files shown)


deathaxe:
This maybe a bit tricky with Perl and Ruby due to the way regexes are matched. I wanted to make sure they can't influence each other.

PR #2059, opened 2019-08-22
[C/C++/ObjC/ObjC++] Exclude snippets from #include<...>

### LINKED ISSUE #2055: C++ autocomplete produces wrong code in vector include
If I try to enter the following in C++:
```cpp
#include <vector>
```
By using the autocomplete from sublime, I would instead get this:
```cpp
#include <std::vector<char> v;>
```
Which is not a valid statement and doesn't work in the same way as autocompleting other headers. Attached is a screen capture.
cpp_autocomplete_vector

deathaxe:
Fixes #2055

Prevent snippets from being added to the completions when the caret is within an include directive.


FichteFoll:
We'd probably want to exclude them from strings and comments as well, rigth?

deathaxe:
As we are at this poin, yes. Agree.

PR #2060, opened 2019-08-22
[Ruby] Update comments

deathaxe:
This commit

1) adds begin/end punctuation scopes to block comments
2) implements line comments the same way as done for Perl and other syntaxes to avoid `capture` for performance reasons.


Ruby/Ruby.sublime-syntax
```diff
@@ -106,16 +106,18 @@ contexts:
...
     - match: ^=begin
-      scope: punctuation.definition.comment.ruby
+      scope: punctuation.definition.comment.begin.ruby
       push:
         - meta_scope: comment.block.documentation.ruby
         - match: ^=end
-          scope: punctuation.definition.comment.ruby
+          scope: punctuation.definition.comment.end.ruby
+          pop: true
+    - match: \#+
```
FichteFoll: I'm undecided on whether we should match this syntactically or semantically. Syntactically, only the first `#` is the actual punctuation, but semantically multiple glyphs can be used. This is the first time I'm seeing this.

Otoh, the `*` asterisk in continued c-style block comments recently received a punctuation scope as well.

deathaxe: Just considder the following comment

```
#########
# the text
#########
```

I wouldn't like to only see the first `#` but all in their own color. Anything else would look odd, I think.

FichteFoll: I understand that. I'm just saying this is unprecedented.

deathaxe: These lines are basically just a copy of Perl's comments.

https://github.com/sublimehq/Packages/blob/d1494f4a5a2bf8c6ae2f3e7ad762a05273df97ef/Perl/Perl.sublime-syntax#L88-L93

wbond: For the record, I think applying the scope to all of the leading punctuation is probably what the user will generally want, so let's try it.

PR #2064, opened 2019-08-23
[Java] Fix multiple class inheritance

### LINKED ISSUE #1843: [Java] Interfaces with multiple inheritance break highlighting
Interfaces In Java can inherit from multiple parent interfaces even though it isn't allowed for classes.

In the following example the parent interface `BB` is scoped as `constant.other.java` when it should be `entity.other.inherited-class.java`

A minimal example:
```java
interface T extends A, BB {}
```

As rendered in Sublime Text:

image

deathaxe:
Fixes #1843


wbond:
Thanks!

PR #2065, opened 2019-08-23
[Java] Remove quotation marks from match patterns

deathaxe:
This PR proposes to remove some remaining quotation marks from the match patterns because most of them are already unquoted.

This PR does not change functionality, but is only for consistent code style.


FichteFoll:
You might want to squash and force-push the last commit.

PR #2069, opened 2019-08-25
[Ruby] Fix string escape sequences

deathaxe:
This commit implements string escape sequences according to the latest ruby documentation.

It fixes:

1) meta control sequences
2) unicode escapes `\uxxxx` and (`\u{...}`)
3) octal escapes, which begin with `\n`

SEE: description of escape sequences within double quoted strings at
     https://ruby-doc.org/core-2.6.3/doc/syntax/literals_rdoc.html


wbond:
@deathaxe It seems things got conflicted during all of the Ruby merges. Do you have a moment to sort this out? I took a peek, but it wasn't immediately obvious to me what needed to stay.

PR #2071, opened 2019-08-26
[CSS] Fix selector symbol index

### LINKED ISSUE #2057: [CSS] Symbol Index includes Leading Whitespace
The selector for Goto Symbol in a CSS file is `meta.selector`, which appears to be matching leading whitespace before the start of a selector.

Test case:

```
        body {
            font-family: "Open Sans", "Helvetica Neue", Helvetica, Arial;
            background-color: #f6f6f6;
            text-align: center;
        }
        h1 {
            margin-top: 100px;
            font-size: 4.0em;
            color: #333;
        }
```

deathaxe:
Fixes #2057

This commit modifies the `meta.selector` lookahead in order to exclude leading whitespace.

Whitespace between entities and after the last one keep untouched.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -788,6 +788,8 @@ contexts:
             - match: '\]'
               scope: punctuation.definition.entity.css
               pop: true
+        - match: ','
+          scope: punctuation.separator.sequence.css
 
   # Pseudo Elements
   # https://drafts.csswg.org/selectors-4/#pseudo-elements
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -733,7 +733,7 @@ contexts:
         - include: property-values
 
   selector:
-    - match: '\s*(?=[:.*#a-zA-Z\[])'
+    - match: (?=[:.*#a-zA-Z\[])
       push:
         - meta_scope: meta.selector.css
         - match: "(?=[/@{)])"
--- a/CSS/syntax_test_css.css
+++ b/CSS/syntax_test_css.css
@@ -48,11 +48,15 @@
 /*  ^   punctuation.definition.entity.css  */
 /*   ^^ entity.other.attribute-name.id.css */
 
-    html {}
+    html, h1 {}
 /*^^ - meta.selector.css  */
-/*  ^^^^^ meta.selector.css */
-/*       ^^ - meta.selector.css  */
+/*  ^^^^^^^^^ meta.selector.css */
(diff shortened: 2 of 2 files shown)


michaelblyons:
Suggestion?

```yaml
  selector:
    - match: (?=[:.*#a-zA-Z\[])
      push:
        - meta_scope: meta.selector.css
        - match: '(?=\s*[/@{)])'
          pop: true
        - match: ','
          scope: punctuation.separator.sequence.css
        ...
```
Removes whitespace from the end as well, and adds scope for `,`.

deathaxe:
The selector is ok.

I am in doubt with stripping the trailing space from the `meta.selector` for 2 reasons:

1) In many (most?) situations such meta scopes span the whole area up to the next punctuation including the surrounding whitespace. (See `meta.function-call.name`, etc. Stripping it here basically arises the basic question about how to apply meta scope boundaries.
2) Completions may fail if the `meta.` boundaries are to tight, because if the caret is behind the last keyword, the scope at the position of the first following character is used as selector to trigger completions.
3) I find it bad practice to add entities to the index based on the meta scope only, which is the core issue with CSS selectors. I understand the reasons for doing so, but I think only `meta.selector.css entity` should be indexed. This would solve the overall problem.

michaelblyons:
Argument 2 is very compelling, and argument 1 is closely related to it.

I don't think I completely agree with argument 3, but I have to think a little bit before writing an essay about it.

FichteFoll:
For 3., a symbol transformation could easily strip whitespace at the end, couldn't it?

michaelblyons:
> For 3., a symbol transformation could easily strip whitespace at the end, couldn't it?

I thinks it's already doing that. The question, as I understand it, is "When we have a CSS selector string like this:

``` css
alpha,
beta a,
gamma.foo a,
/* commented selector, */
delta:hover,
epsilon /* > div */ p {}
```

...then what do we add to the index?

1. Everything from `alpha` to `p` as one entry (maybe stripping the comments)?
2. Each `,`-delimited selector individually?
3. Each element selector individually? (`gamma.foo`, `a`)
4. Each selector-type individually? (`gamma`, `.foo`, `a`)
5. Something else?

(1) is close to what we do now. (We just don't handle comments). I _think_ DeathAxe is arguing for (3), but I'm not completely sure.

wbond:
I personally think it would be an improvement to match each selector to the local index, e.g. `gamma.foo` and `a` separately. We don't need to accomplish that here though.

deathaxe:
Think so, too. Let's see how it will play with the new completion system.

PR #2078, opened 2019-09-02
[Lua] Update mapping matching

FichteFoll:
img

Reported by @Resike on Discord.

While I was at it, I also implemented `meta.mapping.{key,value}`, as we established those as standards a while ago and as they are used in Python, JS (?) and JSON. It made sense to add to the PR because it would otherwise be conflicting.

And highlighting of metamethod and normal method assignments.


FichteFoll:
@Thom1729 please re-review :slightly_smiling_face:

FichteFoll:
And another suggestion by @Resike, this time with highlighting of metamethod assignments in tables. I expanded it to method assignments in general, including indexing.

Issue #1931, opened 2019-03-30
[C++] Pointer type as template parameter breaks syntax highlighting


brainplot:
I'm not sure if this is a bug or it's intended behavior but taking this code:
```cpp
void foo()
{
  std::copy(first, last, std::ostream_iterator<T>(std::cout, "\n"));
}
```
when `T` is not a pointer type, `std::ostream_iterator` is highlighted differently from when `T` is a pointer type.
This is what it looks like when `T` **is not** a pointer type:
![value](https://user-images.githubusercontent.com/28014843/55278590-a4244a00-530e-11e9-8d16-efcd2fa98a7e.png)
and this is what it looks like when `T` **is** a pointer type:
![pointer](https://user-images.githubusercontent.com/28014843/55278596-b1413900-530e-11e9-8f85-4578147f35fd.png)

The problem seems to be reproducible with any explicit template instantiation.

bumfo:
This can be fixed by removing '*' from C++.sublime-syntax/variables.regular. And I don't see any side effects of changing this.

bumfo:
This issue applies to references as well (which is rarer though), but fixing that is more complicated.

PR #2079, opened 2019-09-03
[C++/ObjC++] Fix #2042, generic has no form like '<<'

### LINKED ISSUE #2042: [C++] Syntax highlighting error after string ">>"
```cpp
#include <iostream>

int main(void) {
	std::cout << ">>" << std::endl;
}
```

The closing `}` is highlighted pink. No such error if `">>"` is changed to `"aa"`.

bumfo:
generic_lookahead should not include patterns like '<<', '<<<', which causes bugs like #2042.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -40,12 +40,13 @@ variables:
   non_angle_brackets: '(?=<<|<=)'
 
   regular: '[^(){}&;*^%=<>-]*'
+  regular_plus: '[^(){}&;*^%=<>-]+'
   paren_open: (?:\(
   paren_close: '\))?'
-  generic_open: (?:<
-  generic_close: '>)?'
+  generic_open: (?:{{regular_plus}}(?:<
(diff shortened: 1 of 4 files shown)


rwols:
Thanks for fixing this. Could you add a test to verify that this issue (https://github.com/sublimehq/Packages/issues/1931) is fixed too?

Due to how the C-family of syntaxes are set up you'll also need to copy your changes to the Objective-C++ syntax, and copy over the tests to the syntax tests of Objective-C++.

bumfo:
> Thanks for fixing this. Could you add a test to verify that this issue (#1931) is fixed too?
> Due to how the C-family of syntaxes are set up you'll also need to copy your changes to the Objective-C++ syntax, and copy over the tests to the syntax tests of Objective-C++.

I'd like to open a separate pull request for that one, which is much more complex. 

fixing the case for pointer is easy, but fixing the case for references is much harder, which complicates the fix, so I don't want to include these here.

wbond:
Thanks!

Issue #2080, opened 2019-09-03
[C++/ObjC++] friend methods with body highlighted incorrectly


bumfo:
According to PR #1083, there are four cases of friend declaration. The forth case of friend method contains a body, and should be highlighted & indexed as normal function/methods. 

e.g. 
```C++
class Foo {
  int hidden_data;
  friend void swap(Foo & a, Foo & b) {
    using std::swap;

    swap(a.hidden_data, b.private_data);
  }
};
```
this is part of the well-known copy-and-swap idiom, which should get prioritized support. 

Other common examples of friend declaration with body include operator overloading, etc.:

```C++
class Foo {
private:
  int hidden_data;
public:
  Foo(int data) : hidden_data(data) {}
  friend std::ostream& operator << (std::ostream& os, const Foo& foo) {
    return os << "A Foo object with value: " << foo.hidden_data;
  }
};
int main() {
  Foo foo(42);
  std::cout << foo << '\n';
};
```

Issues related with this includes: #412, #1082, but this one focuses on the case when method body is included.

PR #2081, opened 2019-09-03
[C++/ObjC++] Highlight friend methods with body

bumfo:
This PR aims to retry the attempt in #1083. 
More details can be seen in #2080. 

With this PR, friend methods **with body** is marked with `entity.name.function`, 
while other types of declarations remain untouched.


wbond:
Thanks!

PR #2083, opened 2019-09-08
[HTML] Improve class/id attribute lists

deathaxe:
This PR attempts to support a fix for https://github.com/SublimeTextIssues/Core/issues/2960

It includes 2 major changes.

1. Add support for entity highlighting in unquoted attribute values according to [whatwg.org / 12.2.5.38 Attribute value (unquoted) state](https://html.spec.whatwg.org/multipage/parsing.html#after-attribute-value-(quoted)-state)
2. Split the `meta.class-name` and `meta.toc-list.id` tokens by spaces in order to make the meta-scopes represent a single class or id only. This change makes it possible to target each class/id individually.

_Note: This PR does not add a `punctuation.separator.sequence` for the space between the classes or ids, because it is unclear, whether a space should really be scoped as punctuation._

## Example

```html
<html>
<head>
    <style type="text/css">
        #MyId {
            font-size: 33pt;
        }
        .el-ment\' {
            font-weight: bold;
        }
    </style>
</head>
<body>

    <div id=My&#x49;d class="el-m&#x65;nt'">hello</div>
</body>
</html>
```

## Screenshot

grafik

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -404,7 +404,7 @@ contexts:
           captures:
             1: punctuation.definition.tag.begin.html
             2: entity.name.tag.script.html
-          set: 
+          set:
             - meta_scope: meta.tag.script.end.html
             - include: tag-end
             - include: tag-attributes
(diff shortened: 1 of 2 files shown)


wbond:
Just to be clear, if entities are in the attribute in the HTML, then the symbol lookup isn't going to work.

Splitting the scopes on spaces seems like a reasonable change, so I'll merge.

PR #2087, opened 2019-09-11
Fix files being split on unicode newlines in diff.py

BenjaminSchaaf:
Sublime Text only considers \n a newline character, while readlines()
also splits on various unicode characters. See
https://github.com/SublimeTextIssues/Core/issues/2825 for a bug caused
by this.

### DIFF
--- a/Diff/diff.py
+++ b/Diff/diff.py
@@ -7,14 +7,21 @@ import sublime
 import sublime_plugin
 
 
-def read_file_lines(fname):
-    with open(fname, mode="rt", encoding="utf-8") as f:
-        lines = f.read().split('\n')
+def splitlines_keep_ends(text):
+    lines = text.split('\n')
 
-    # Need to insert back the newline characters between lines
+    # Need to insert back the newline characters between lines, difflib
+    # requires this.
     if len(lines) > 0:
         for i in range(len(lines) - 1):
             lines[i] += '\n'
+
+    return lines
+
+
+def read_file_lines(fname):
+    with open(fname, mode="rt", encoding="utf-8") as f:
+        lines = splitlines_keep_ends(f.read())
 
     # as `difflib` doesn't work properly when the file does not end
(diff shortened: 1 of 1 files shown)


FichteFoll:
Fixes https://github.com/SublimeTextIssues/Core/issues/2825 when merged.

Diff/diff.py
```diff
@@ -8,18 +8,24 @@
 
 
 def read_file_lines(fname):
-    with open(fname, mode="rt", encoding="utf-8", newline=None) as f:
-        lines = f.readlines()
+    with open(fname, mode="rt", encoding="utf-8") as f:
+        lines = f.read().split('\n')
+
+    # Need to insert back the newline characters between lines
+    if len(lines) > 0:
```
FichteFoll: Since the "split by \n and readd \n" sequence is repeated on line 83, it would make sense to create a function for this.

BenjaminSchaaf: I'm not sure putting `read().split('\n')` in its own function would actually be beneficial. Line 83 only has the split in common.

wbond: I think @FichteFoll has a point. Both here and line 83 consist of, roughly:

```python
lines = data.split('\n')
if len(lines) > 0:
    for i in range(len(lines) - 1):
        lines[i] += '\n'
````

Something line this would be more clear:

```python
def split_keepends(text):
    lines = text.split('\n')
    if len(lines) > 0:
        for i in range(len(lines) - 1):
            lines[i] += '\n'
    return lines

with open(fname, mode="rt", encoding="utf-8") as f:
    lines = split_keepends(f.read())

b = self.view.substr(sublime.Region(0, self.view.size()))
b = split_keepends(b)
```

BenjaminSchaaf: Ah right, sorry I clearly didn't quite understand the original suggestion.

FichteFoll: Adding comments to line ranges hasn't been a feature last month, I think :wink:

PR #2089, opened 2019-09-13
[Rust] Improve scopes for textual operators

jwortmann:
This PR applies the scope `keyword.operator.word` instead of just `keyword.operator` to `as`, `in`, `box`, so that color schemes can highlight them differently from symbolic operators.
Reference: https://www.sublimetext.com/docs/3/scope_naming.html#keyword


keith-hall (dismissed):

wbond:
This is more or less pending on the result of the conversation at #2090

PR #2090, opened 2019-09-13
[JavaScript] Improve scopes for textual operators

jwortmann:
This PR applies the scope `keyword.operator.word` instead of just `keyword.operator` to `delete`, `typeof`, `void`, `instanceof`, `in`, so that they are picked up by color schemes that want to highlight textual operators differently from symbolic operators.
Reference: https://www.sublimetext.com/docs/3/scope_naming.html#keyword

Edit: @michaelblyons is something wrong with this (because you reacted with confused emoji)? This should have no effect for color schemes that highlight all operators the same, but those that want to differentiate between symbols and word operators don't have to add an extra rule for JavaScript. And it should comply with the scope naming guidelines now.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1069,14 +1069,14 @@ contexts:
     - match: \+|\-
       scope: keyword.operator.arithmetic.js
     - match: (?:delete|typeof|void){{identifier_break}}
-      scope: keyword.operator.js
+      scope: keyword.operator.word.js
 
   binary-operators:
     - match: instanceof{{identifier_break}}
-      scope: keyword.operator.js
+      scope: keyword.operator.word.js
       push: expression-begin
     - match: in{{identifier_break}}
-      scope: keyword.operator.js
+      scope: keyword.operator.word.js
       push: expression-begin
     - match: '&&|\|\|'
       scope: keyword.operator.logical.js
--- a/JavaScript/tests/syntax_test_js.js
+++ b/JavaScript/tests/syntax_test_js.js
@@ -708,7 +708,7 @@ for (var i = 0; i < 10; i++) {
 //  ^^^ keyword.control.loop
 //      ^^^^^^^^^^^^^^ meta.group
 //       ^ punctuation.separator.expression
-//           ^^ keyword.operator
+//           ^^ keyword.operator.word
 //                  ^ punctuation.separator.expression
 
     for (a[x in list];;) {}
@@ -716,7 +716,7 @@ for (var i = 0; i < 10; i++) {
 //  ^^^ keyword.control.loop
 //      ^^^^^^^^^^^^^^^^ meta.group
 //        ^^^^^^^^^^^ meta.brackets
-//           ^^ keyword.operator
+//           ^^ keyword.operator.word
 //                   ^ punctuation.separator.expression
 //                    ^ punctuation.separator.expression
 
@@ -1479,6 +1479,7 @@ new $Dollar();
 //  ^ variable.type.dollar punctuation.dollar
 
 void {
+//^^ keyword.operator.word
     'test1': [],
     'test2': new SomeObjectHash["default"],
 //               ^^^^^^^^^^^^^^^^^^^^^^^^^ meta.function-call.constructor
@@ -1584,11 +1585,11 @@ a = /\//u + 0;
 
     x
     in y;
-//  ^^ keyword.operator
+//  ^^ keyword.operator.word
 
     x
     instanceof y;
-//  ^^^^^^^^^^ keyword.operator
+//  ^^^^^^^^^^ keyword.operator.word
 
 var π = 3.141592653
 //  ^ variable.other.readwrite


jfcherng:
I guess the confusion comes from that we hardly give a scope in a linguistics aspect.

michaelblyons:
I don't object to extending the scope chain, but my preference is for using it to better describe what a term _means,_ rather than how it _appears._

You are, however, correct that the scope naming guidelines suggest `keyword.operator.word` for `and`, `or`, and `not`, but that seems silly for languages like Python or SQL where most logical operators _are_ words.

Python even uses a different scope for `in` depending on where it appears: `keyword.control.loop...` for enumerating a structure and `keyword.operator.logical` for membership tests that return a boolean.

Side note: @Briles can you update [your handy summary](http://brianreilly.me/sublime-syntax-dashboard/#/scopes/Keyword) of `keyword` scopes?

wbond:
I think I probably wasn't specific enough in the guidelines about the rationale for `keyword.operator.word`. If I recall correctly it was originally used in a language that had primarily C-like symbol operators, but a few word-based operators that *were lower precedence* and could often trip people up.

That said, I don't have specific thoughts on its usage in this PR.

Thom1729:
I am of three minds. (Like a tree / In which there are three blackbirds):

1. The guidelines specify `word`, and this PR definitely seems to conform to those guidelines.
2. @michaelblyons's point is well taken that the `word` distinction seems of dubious value here. In many cases we'd be better off scoping operator's like Python's `and` as `keyword.operator.logical` rather than `keyword.operator.word`.
3. If we had [approximate selectors](https://github.com/SublimeTextIssues/Core/issues/2619), then we could have the best of both worlds.

As we don't have approximate selectors, we must make a choice between `word` and `logical` for Python's `and`. The current Python syntax goes with `logical`, and I don't disagree. But then what is `word` for — only operators that don't have another category? That is a fragile distinction; why shouldn't we one day choose a standard scope for `typeof` in languages that have it?

I don't want to fall into the trap of mindless consistency, but I'm having trouble imagining a reasonable rule that includes both `word` and other specific scopes like `logical` and `arithmetic`; those taxonomies seem fundamentally at odds (again, without approximate selectors).

wbond:
I think the ambiguity ties back in with what the original use of this selector was for, which was to allow a color scheme to highlight word-based operators in a syntax where word-based operators had lower precedence than symbol-based ones.

I would think in that case `keyword.operator.logical keyword.operator.word` would probably be the most appropriate scope, since it would allow a user who wants to highlight these "special" operators, whereas for most color schemes they would likely target `keyword.operator.logical`, or just `keyword.operator`.

jwortmann:
Let me explain my intention behind this PR: [my color scheme](https://github.com/jwortmann/brackets-color-scheme) does not highlight symbolic operators like `+`, `=`, etc. differently from plain text/variables, but i do want to highlight the operator keywords `and`, `in`, ... like other keywords, if they are not just symbol characters. I think most languages use `keyword.operator.word` for the latter, so i just need to specify two rules for `keyword.operator` and `keyword.operator.word` with different colors in my color scheme to make it apply to all languages. For example in the JavaScript syntax there are already some keywords with `keyword.operator.word` present (`of`, `in`, `new`) and highlighted correctly, but those mentioned in the PR were missed and therefore don't recieve highlighting:

screenshot

In this case, I was able to add a special rule for `keyword.operator.js` to highlight them, but this only worked because all of the symbolic operators in the JavaScript syntax use another (more specific) scope, so this rule did only apply to those textual operator keywords as intended. But I had to go through a lot of syntax definition files and test files to find occurences where I needed to use additional rules in the color scheme to apply highlighting for such keywords (for example `keyword.operator.new.cs` in C#, `keyword.operator.logical.lua` in Lua and `keyword.operator.logical.python` in Python), and I probably still have missed some of them. These additional rules were not necessary if the syntaxes would use `keyword.operator.word` for all textual operators. Of course they can use more specific scopes like `keyword.operator.word.new` where appropriate. 

And currently it doesn't work for example in Rust, where both symbolic and textual operators use only `keyword.operator.rust`.

The problem might be less noticeable in ST's default color schemes, because they use a special highlighting color for symbolic operators (which will then be applied to textual operators as well), but in my case (and probably there are other color schemes, which also don't use a special color for operators) it just looks as if the syntax doesn't recognize those keywords.

keith-hall:
Just to back up Will's suggestion, color schemes targeting both `keyword.operator` to cover all operators and `keyword.operator.logical` to color logical operators differently would work as expected:

```python
>>> sublime.score_selector('keyword.operator.logical keyword.operator.word', 'keyword.operator')
16
>>> sublime.score_selector('keyword.operator.logical keyword.operator.word', 'keyword.operator.logical')
24
```

though I would personally say that `logical` is more important/specific than `word`, so I'd order them as `keyword.operator.word keyword.operator.logical` in the syntax definition. But I don't have a solution for allowing color schemes which choose to color both "word" operators and (i.e.) logical operators differently to decide which one gains precedence in this scenario... Special casing all combinations of different operators wouldn't be fun... (Sorry if I'm not being clear here.) Could it be useful to include some extra guidance in the docs about how to order such "doubled-up" scopes?

Thom1729:
Is this doubling-up something that might be used elsewhere? For instance, should the Python string `rf"""foo!"""` be `string.quoted.double string.block string.raw string.interpolated`?

Briles:
> Side note: @Briles can you update [your handy summary](http://brianreilly.me/sublime-syntax-dashboard/#/scopes/Keyword) of `keyword` scopes?

@michaelblyons I've updated the dashboard with the latest data. I'll try and get the data updating automatically on a schedule so I don't have to manually update it. Glad you're finding it useful! 👍

jwortmann:
I have compiled a list where ...

1. ... color schemes need to add an extra rule (different from `keyword.operator.word`) to distinguish between word-based and symbol-based operators:
* ActionScript uses `keyword.operator.actionscript.2` for `instanceof`
* ASP uses `keyword.operator.logical.asp` for `And`, `Not`, `Or`, `Xor`, `Is`
* C# uses `keyword.operator.new.cs` for `new`
* CSS uses `keyword.operator.logic.css` and `keyword.operator.logic.media.css` for `and`, `or`, `not`, `only`
* JavaScript uses `keyword.operator.js` for `delete`, `typeof`, `void`, `instanceof`, `in` (this PR)
* Lua uses `keyword.operator.logical.lua` for `and`, `or`, `not`
* Matlab uses [`keyword.operator.matlab`](https://github.com/sublimehq/Packages/blob/d1494f4a5a2bf8c6ae2f3e7ad762a05273df97ef/Matlab/Matlab.sublime-syntax#L210) for various intrinsic functions, e.g. `all`, `and`, `isnan`, `not`, `unique`.
I would recommend to use `support.function` instead of `keyword.operator` here, because these all are indeed just function-calls, e.g. `and(A,B)` and `not(A)`. For some of them, there is also a dedicated operator symbol with the same meaning, e.g. `A & B` for `and(A,B)` and `~A` for `not(A)`, which have the scope `keyword.operator.symbols.matlab`.
* Objective-C uses `keyword.operator.logical.predicate.coca.objc` and `keyword.operator.comparison.predicate.cocoa.objc` for various word-based operators in predicate strings, e.g. `AND`, `NOT`, `CONTAINS`, `BETWEEN`
* PHP uses `keyword.operator.type.php` for `instanceof`
* Python uses `keyword.operator.logical.python` for `and`, `in`, `is`, `not`, `or`
* SQL uses `keyword.operator.logical.sql` for `is`, `not`, `and`, `or`, `like`, `having`, `between`, `in`
* SQL uses `keyword.operator.assignment.alias.sql` for `as`

2 ... it is impossible for color schemes to distinguish between word-based and symbol-based operators:
* ASP uses `keyword.operator.asp` for `Mod` and various symbol-based operators
* Matlab uses [`keyword.operator.symbols.matlab`](https://github.com/sublimehq/Packages/blob/d1494f4a5a2bf8c6ae2f3e7ad762a05273df97ef/Matlab/Matlab.sublime-syntax#L143) for `end` (when used as an array index) and various symbol-based operators
* PHP uses [`keyword.operator.logical.php`](https://github.com/sublimehq/Packages/blob/d1494f4a5a2bf8c6ae2f3e7ad762a05273df97ef/PHP/PHP%20Source.sublime-syntax#L247) for both word-based operators `and`, `or`, `xor`, `as` as well as the symbol-based operators `&&`, `||`
* Rust uses `keyword.operator.rust` for both word-based operators `as`, `in`, `box` and various symbol-based operators. Possible fix: #2089

I think there is almost no programming language (if any) where the symbols `+`, `=`, etc. have no special meaning and can be used as valid variable identifiers, so I would say it is not reprehensible for color schemes to not give them special highlighting differently from e.g. variables/plain text. But then, as mentioned before, color schemes have no means to highlight certain reserved words. Therefore it would be nice to have a simple selector rule like
```
keyword - keyword.operator, keyword.operator.word
```
to match all kind of reserved words while leaving symbol-based operators untouched. For the logical operators in ASP, CSS, Lua, Python, PHP and SQL, I think it makes sense to keep the `keyword.operator.logic` or `keyword.operator.logical` scope and use the proposed scope doubling `keyword.operator.logical keyword.operator.word`, but I really can't see any advantage for only having `keyword.operator.js` instead of `keyword.operator.word.js` for the keywords in this PR. So even if it is just for appearance reasons rather than semantics, I disagree that this PR is of "dubious value" regarding that `word` distinction.

michaelblyons:
A scope selector like your `keyword - keyword.operator, keyword.operator.word` is elegant, and even if it's not coloration I would use, it's reasonable to expect that you could want to.

I am amenable to Keith's `keyword.operator.word keyword.operator.logical`, but as he points out, there isn't any official guidance on doing so. A change like this would have to be addressed separately in every language (here or 3rd-party) that your color scheme's users cared about.

Since you have made that long list, I'd like to comment on some classes of scope:

- `typeof` in my opinion, should be `keyword.operator.type`.
- `is` (and possibly `instanceof`, depending on what it means) should be `keyword.operator.logical.type`.
    + C# uses `keyword.operator.reflection` for it.
- I don't have opinions about `delete`. The nearest equivalents I know about are Python's `del` (`keyword.other.del`) and Bash's `unset` (`support.function.unset`).
- I agree with your assertions about Matlab, but I can imagine some backlash if they are changed.
- I wrote the scope for `as` in SQL, and it is not always correct. In `CAST(foo AS INT)` expressions, it should have a different scope. I don't know what the cast-`as` scope should be. (It's in C# as `keyword.operator.reflection`.)

Thom1729:
I would like to come up with standardized sub-scopes for common operators that currently don't have them. That said, it might be best to break that out into a new discussion; it seems mostly orthogonal to the word operator issue.

deathaxe:
> Is this doubling-up something that might be used elsewhere? For instance, should the Python string rf"""foo!""" be string.quoted.double string.block string.raw string.interpolated?

1. I am not a fan of stacking (too many) scopes.
2. `string.interpolation` got obsolete as strings should use `meta.string.<lang> string.` to properly support interpolation by clearing the `string` scope within `meta.string.<lang> meta.interpolation.<lang>` blocks.

---

> I would like to come up with standardized sub-scopes for common operators

Looks quite important as operators seem to be a big mess at the moment.

wbond:
To resolve this we need to come up with a solution to two main conflicting interests:

 - Allowing users who want to highlight word-based operators that may blend in with variable names, without having to create a selector with every possible `keyword.operator.*.word` variant.
 - Using scope names for operators based on their behavior, as opposed to their presentation/implementation.

Based on this, what about revising the scope naming guidelines to recommend `keyword.operator.logical` for any operator form (symbol or word) and then in situations where an operator is a word, and that is not the norm for the language, to use `meta.operator.word keyword.operator.{type}` for the scope? This maintains the `keyword.operator` classification system, but provides a method for people wishing to highlight "different" operators?

FichteFoll:
Sounds good to me. Just for clarification, you mean the following order, correct? `keyword.operator.* meta.operator.word`

jwortmann:
In general I like that suggestion
```
meta.operator.word keyword.operator.*
```
(in that order).

But just to clarify, in order to keep the current highlighting, 4 of the 5 default color schemes (Monokai, Mariana, Breakers, Sixteen) would have to be adjusted, because they all target `keyword.operator.word`, which seems to be an established scope name in various syntaxes. And I think that a lot of custom color schemes will also target `keyword.operator.word`, so unless it is guaranteed that existing occurances of `keyword.operator.word` will not be changed in the future (which wouldn't really make sense regarding consistency), this would break backwards compatibility / change operator highlighting colors in a lot of places.

deathaxe:
I'd always vote for `meta.` scopes to prefix normal coloring scopes, so I am with @jwortmann  here.

At least in this repo `keyword.operator.word` is not used too widely. If we decide to abandon this scope here, we could tweak the other syntaxes as well with little effort. In some of the use cases the changed color wouldn't even hurt, IMHO.

wbond:
The issue of `meta` scopes was just being discussed on Discord since we need to refactor how numeric bases are scoped. In that discussion I proposed the idea for situations like this to use something along the lines of `keyword.operator.logical meta.category.word` and providing guidance in the color scheme docs that targeting a `meta.category` or `meta.component` scope is what they are designed for. This would be a break from other meta scopes where they mark up large parts of a file and aren't appropriate to target with foreground rules.

These `meta` scopes for further specialization would try to remain short, and would be targeted by adding after the base scope you want to tweak. They would also come after the primary scope name, especially in the case of `meta.component` which is the proposal to use with the `0x` of a number like `0xf00`.

frou:
Just in case anyone is wondering *why* some of us want to have the ability to disable highlighting specifically for symbol operators... I'll speak for myself and say that it's because things like = and && *already* look very distinct from other text, simply due to the shape of the glyphs. They don't need to be coloured to be visually recognisable.

jwortmann:
I think the reason why this problem wasn't that prominent yet, is because all of the default color schemes use a special highlighting color (different from plain text) for operator symbols.

Another related case that comes to mind when considering how to distinguish between symbols/words, is the "arrow operator / arrow function" `=>`, and I can imagine that some users don't want to highlight it as well. Most scope names for this seem to be `storage.type.function.{arrow|lambda|anonymous}`, I've seen `keyword.other.big-arrow.haskell` in Haskell (I'm not sure if it has the same meaning here though), and C# sometimes uses `storage.type.function.accessor.get.cs` for both `=>` and `get`, which prevents to highlight them differently. It would be great if a solution how to distinguish operator symbols, could handle this examlpe too.

When there is an "official" decision for the scope names for word-based operators, please leave a message here, so that I can update this PR.

deathaxe:
This is what I called fragmentation of scope names recently. Skimming through the packages quickly reveals its history. Some packages are quite well maintained while others have fallen behind a bit. Every syntax comes with its own and probably new challenges which results in very individual solutions and scope names. Those make it hard to create color schemes with many features which address most/all syntaxes.

Another challange are up coming user requests for new opportunities to address certain aspects of a syntax. I personally would never ever not highlighte operators. It is btw. one of the biggest issues of the Backets color schemes. It just looks incomplete to me. 

The question however is to carefully find proper scope names and some kind of scheme, which can be applied to hopefully all syntaxes.

Unfortunatelly some of the discussions seem stale and even those which came to an end have not all been applied to the major syntaxes so far. 

Hence we should carefully decide what to introduce next, because even small things can quickly become a sophisticated task. We are not yet done with the "numbers" discussion nor have the "keyword.control" scopes been applied to all syntaxes.

PR #2093, opened 2019-09-14
[SCALA] Proper parsing annotations in a class parameter list

LukaszByczynski:
This a fix for annoying pink highlights, when class using annotations for a parameter list, for example:

```scala
class TestController @Inject()(controllerComponents: ControllerComponents) extends AbstractController(controllerComponents) with I18nSupport
```
Without that fix, `extends` and `with` have this illegal syntax hightlights.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -569,6 +569,11 @@ contexts:
   class-type-parameter-list:
     - match: '\b(private|protected)\b'
       scope: storage.modifier.access.scala
+    - match: (?=@{{plainid}})
+      set:
+        - include: annotation
+        - match: '(?=\S)'
+          set: class-type-parameter-list
     - match: '(?=\()'
       set: class-parameter-list
     - match: '\['
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -182,21 +182,31 @@ contexts:
           set: annotation-parameters
     - match: '(@)({{alphaplainid}})'
       captures:
-        1: meta.annotation.scala punctuation.definition.annotation.scala
-        2: meta.annotation.identifier.scala variable.annotation.scala
+       1: meta.annotation.scala punctuation.definition.annotation.scala
+       2: meta.annotation.identifier.scala variable.annotation.scala
       push: annotation-parameters
 
   annotation-parameters:
     - meta_content_scope: meta.annotation.parameters.scala
+    - match: (?=\(\))
+      push:
+      - match: \(
+        scope: punctuation.section.arguments.annotation.begin.scala
+      - match: (?=\))
+        scope: punctuation.section.arguments.annotation.end.scala
+        pop: true
     - match: \(
       scope: punctuation.section.arguments.annotation.begin.scala
       push:
-        - match: ','
-          scope: punctuation.separator.arguments.annotation.scala
-        - match: \)
+        - match: (?=\))
           scope: punctuation.section.arguments.annotation.end.scala
           pop: true
+        - match: ','
+          scope: punctuation.separator.arguments.annotation.scala
         - include: main
+    - match: \)
+      scope: punctuation.section.arguments.annotation.end.scala
+      pop: true
     - match: \[
       scope: punctuation.section.arguments.annotation.begin.scala
       push:
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -182,8 +182,8 @@ contexts:
           set: annotation-parameters
     - match: '(@)({{alphaplainid}})'
       captures:
-       1: meta.annotation.scala punctuation.definition.annotation.scala
-       2: meta.annotation.identifier.scala variable.annotation.scala
+        1: meta.annotation.scala punctuation.definition.annotation.scala
+        2: meta.annotation.identifier.scala variable.annotation.scala
       push: annotation-parameters
 
   annotation-parameters:
--- a/Scala/syntax_test_scala.scala
+++ b/Scala/syntax_test_scala.scala
@@ -1329,12 +1329,30 @@ trait AlgebraF[F[_]] { type f[x] = Algebra[F,x] }
 //                     ^^^^ storage.type.scala
 //                               ^ keyword.operator.assignment.scala
 
+class Foo @Inject()(a: String)
+//         ^^^^^^ meta.annotation.identifier
+//                 ^- meta.annotation
+//                  ^ variable.parameter.scala
+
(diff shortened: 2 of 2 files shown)


djspiewak:
This is legal scala? TIL

LukaszByczynski:
Yeah, it's legal. Often used to cooperate with Java DI libraries like, Spring or Guice.
Playframework uses it as primary mechanism for injections:

https://www.playframework.com/documentation/2.7.x/ScalaTemplatesDependencyInjection

Scala/Scala.sublime-syntax
```diff
@@ -569,6 +569,11 @@ contexts:
   class-type-parameter-list:
     - match: '\b(private|protected)\b'
       scope: storage.modifier.access.scala
+    - match: (?=@{{plainid}})
+      set:
+        - include: annotation
+        - match: '(?=\S)'
```
djspiewak: Can you add a few tests which cover this? The following should be good starting points:

```scala
class Foo @bar (x: Int)
class Foo @bar(42) [A](x: Int)
class Foo @bar @baz (x: Int)
```

LukaszByczynski: Ok. I will add them as soon I can.

Scala/Scala.sublime-syntax
```diff
@@ -182,21 +182,31 @@ contexts:
           set: annotation-parameters
     - match: '(@)({{alphaplainid}})'
       captures:
-        1: meta.annotation.scala punctuation.definition.annotation.scala
-        2: meta.annotation.identifier.scala variable.annotation.scala
+       1: meta.annotation.scala punctuation.definition.annotation.scala
+       2: meta.annotation.identifier.scala variable.annotation.scala
```
djspiewak: The two space indentation is more conventional here. Can you revert this?

Scala/Scala.sublime-syntax
```diff
@@ -182,21 +182,31 @@ contexts:
...
       push: annotation-parameters
 
   annotation-parameters:
     - meta_content_scope: meta.annotation.parameters.scala
+    - match: (?=\(\))
+      push:
+      - match: \(
+        scope: punctuation.section.arguments.annotation.begin.scala
+      - match: (?=\))
```
djspiewak: Shouldn't this just be `- match: \)`?

Scala/Scala.sublime-syntax
```diff
@@ -182,21 +182,31 @@ contexts:
...
       captures:
-        1: meta.annotation.scala punctuation.definition.annotation.scala
-        2: meta.annotation.identifier.scala variable.annotation.scala
+       1: meta.annotation.scala punctuation.definition.annotation.scala
+       2: meta.annotation.identifier.scala variable.annotation.scala
       push: annotation-parameters
 
   annotation-parameters:
     - meta_content_scope: meta.annotation.parameters.scala
+    - match: (?=\(\))
```
djspiewak: Is there a reason we need the lookahead here? Why not just `- match: \(\s*\)`?

Scala/Scala.sublime-syntax
```diff
@@ -182,21 +182,31 @@ contexts:
...
+      push:
+      - match: \(
+        scope: punctuation.section.arguments.annotation.begin.scala
+      - match: (?=\))
+        scope: punctuation.section.arguments.annotation.end.scala
```
djspiewak: Applying a scope to a lookahead doesn't do anything. Or at least, it *shouldn't*. There might be bugs here.

Scala/syntax_test_scala.scala
```diff
@@ -1329,12 +1329,30 @@ trait AlgebraF[F[_]] { type f[x] = Algebra[F,x] }
...
+//                   ^^^^^^^ meta.annotation.identifier
+//                           ^^^^^^^ meta.annotation.parameters
+//                                           ^- meta.annotation
+//                                            ^ variable.parameter.scala
+
+class Foo @Inject(z = "a")(a: String)
+//         ^^^^^^ variable.annotation
+//                ^^^^^^^ meta.annotation.parameters
+//                         ^ variable.parameter.scala
+
```
djspiewak: Can you also try the following?

```scala
class Foo @Inject() [A](a: String)
```

Just to ensure we stay at the appropriate point in the state machine. Or are these only valid *after* type parameters? I've honestly never used this syntax.

LukaszByczynski: This code is invalid. Did you mean:

```scala
class Foo[A] @Inject()(a: String)
```

Scala/Scala.sublime-syntax
LukaszByczynski: +1

Scala/Scala.sublime-syntax
LukaszByczynski: Nope, I have to escape outside annotation-params scope. Annotaions only have one parameter list (there is no currying)

Scala/Scala.sublime-syntax
LukaszByczynski: I did it for color consitency (without that, annotations like `@Inject()` have white braces)

Scala/Scala.sublime-syntax
LukaszByczynski: It does one thing - escapes from case where is there is no annotation's params list.

Scala/syntax_test_scala.scala
djspiewak: Nice! Like I said, I never use this syntax so I don't know what is and is not valid. :-) But yes, that would be a good one to add, basically just covering the last case in the state machine.

Scala/Scala.sublime-syntax
djspiewak: Well, I mean the `scope` directive doesn't do anything, because the `match` doesn't consume anything.

Scala/Scala.sublime-syntax
djspiewak: And `- match: \(\s*\)` doesn't allow for fixing that? I guess I'm just trying to clean up the nested context.

Scala/Scala.sublime-syntax
djspiewak: Ooooooh I see what you're doing here. Gotcha.

wbond:
@djspiewak Does this seem good to merge from your perspective?

djspiewak:
@wbond It seems good to me! I'm still a little scared about the `- match: (?=\))`, `pop: true`, but it's well tested and motivated, so I think it's fine.

wbond:
Thanks for this fix @LukaszByczynski! This will be part of the default packages once the next dev cycle starts.

LukaszByczynski:
Thanks for review

PR #2095, opened 2019-09-16
[JSON] add .tfstate to file_extensions

patrickrgaffney:
*.tfstate files are [Terraform state files](https://www.terraform.io/docs/state/index.html#inspection-and-modification).

### DIFF
--- a/JavaScript/JSON.sublime-syntax
+++ b/JavaScript/JSON.sublime-syntax
@@ -16,6 +16,7 @@ file_extensions:
   - sublime-color-scheme
   - ipynb
   - Pipfile.lock
+  - tfstate
 scope: source.json
 contexts:
   prototype:


deathaxe:
You could easily add a _JSON (Terraform).sublime-syntax_ alias syntax to the Terraform package, which would even enable plugins to assign icons to.

```YAML
%YAML 1.2
---
name: JSON (Terraform)
hidden: true
file_extensions: 
  - tfstate
scope: source.json.terraform
contexts:
  main:
    - include: scope:source.json
```

see also: https://github.com/SublimeText/AFileIcon/commit/1bf2cd34c6651a55dbb5df2238c4d81a6269d982

patrickrgaffney:
Awesome, thanks. Added in [`7f7359b`](https://github.com/alexlouden/Terraform.tmLanguage/pull/39/commits/7f7359b4447350f7667b242bed303a71f61d10bc).

PR #2098, opened 2019-09-22
[Java] Fix lowercase enumeration constants

### LINKED ISSUE #1658: [Java] Enum syntax
In this (syntactically valid) Java code (assume the code is in a file named `test.java`)

```
enum test{
	a(1),
	b(2);
	test(int x){}
	public static void main(String[]a){}
}
```

the `1` and `2` has red background:

Image

deathaxe:
Fixes #1658

Requires PR #2064 to be merged first to avoid conflicts as the changes in this commit cover the same piece of code.

This commit refactors the `class` context to distinguish definitions of

  1. classes and interfaces
  2. enumerations

The body of enumerations is implemented as an extension of a class body which in addition scopes all uppercase-only or function-call like entities as enumeration constants.

Examples:

    ENUM_CONSTANT,
    ENUM_CONSTANT()
    ENUM_CONSTANT{}

    enumConstant,
    enumConstant()
    enumConstant{}

Notes:

1. The braces after an enum constant are re-scoped from `punctuation.section.braces` to `punctuation.section.block` to match the punctuation scope of the owning class/enum.

2. The `meta.block.java` was added to `meta.class.body.java` and `meta.enum.body.java` to comply with the scope naming guidelines, which ask us for a matching meta/punctuation scoping.

3. Except from (2) no other meta scopes were changed. The boundaries are kept as is, even though some of them look a bit odd.

   Examples:
   a) `meta.class.identifier` should not include the `class` keyword
   b) stacking `meta.class` and `meta.class. ...` looks strange.

4. Several `,` and `;` were not scoped at all, which is fixed for most situations by adding `punctuation-separator-comma` to the `keyword` context.


wbond:
It looks like there are some conflicts here. Would you mind checking and seeing what needs to be done?

PR #2106, opened 2019-09-27
[R] Remove .S file extension

### LINKED ISSUE #1935: [R] scope is applied to assembly code files
The R syntax file lists `.s` and `.S` files as R.
```
%YAML 1.2
---
name: R
file_extensions:
  - R
  - r
  - s
  - S
  - Rprofile
scope: source.r
<snip>
```

This has the unfortunate effect of randomly highlighting assembly code files which also use `.s` and `.S` (probably vastly more often than R files do).

Given that the R files with the `s` extensions are actually really S files, perhaps this could be removed. Reading assembly is hard enough as it is.

deathaxe:
Fixes #1935

The `.S` file extension is used by assembly code, which causes them to be highlighted as R by accident.

### DIFF
--- a/R/R.sublime-syntax
+++ b/R/R.sublime-syntax
@@ -5,8 +5,6 @@ name: R
 file_extensions:
   - R
   - r
-  - s
-  - S
   - Rprofile
 scope: source.r
 


wbond:
I looked through the file history and also online and couldn't find any reference to an `.s` file being for R. Granted, it is sort of hard to search for effectively.

kortschak:
@wbond It's a consequence of the ancestry of R (from the S language). Thanks for fixing this.

PR #2109, opened 2019-09-29
[PHP] Add scope for phpdoc's {@inheritdoc} 

jfcherng:
https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.inlineinheritdoc.pkg.html

### DIFF
--- a/PHP/syntax_test_php.php
+++ b/PHP/syntax_test_php.php
@@ -271,19 +271,12 @@ $var->meth()[10];
 /** @var Properties: class properties. */
 //  ^ keyword.other.phpdoc
 
-/**
- * @return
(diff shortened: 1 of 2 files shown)


PHP/syntax_test_php.php
```diff
@@ -271,297 +271,129 @@ function foo(?stinrg ...$args) {}
 /** @var Properties: class properties. */
 //  ^ keyword.other.phpdoc
 
-/**
- * @return
```
jfcherng: This test is already include at https://github.com/sublimehq/Packages/pull/2109/files#diff-704c2dc0506af29e37902450ea766d79R318.

PR #2110, opened 2019-10-01
[Diff] Add syntax specific setting to enforce tabs

OdatNurd:
This adds a `Diff.sublime-settings` file that reinforces that the default  setting for `translate_tabs_to_spaces` in Diff files should be `false` unless the user specifically overrides it.

The rationale here is that the `diff_files` and `diff_changes` commands provided by the package use `append` to add the computed diff to the `view`/`panel`; thus if the user enables `translate_tabs_to_spaces` in their user preferences, diff output will always be portrayed with spaces regardless of the input file.

I think the expectation when computing a diff would be that both the context lines and the changed lines be represented as in the original files; translating tabs to spaces in this case can potentially distort the diff and make it harder to see what exactly has changed.

Making sure that this setting is disabled in a diff unless the user takes a deliberate action to do so still allows for this behaviour to remain if the user wants it.

### DIFF
--- a/Diff/Diff.sublime-settings
+++ b/Diff/Diff.sublime-settings
@@ -1,0 +1,3 @@
+{
+	"translate_tabs_to_spaces": false
+}
--- a/Diff/Diff.sublime-settings
+++ b/Diff/Diff.sublime-settings
@@ -1,3 +1,3 @@
 {
-	"translate_tabs_to_spaces": false
+	"trim_trailing_white_space_on_save": false
 }
--- a/Diff/diff.py
+++ b/Diff/diff.py
@@ -52,7 +52,9 @@ class DiffFilesCommand(sublime_plugin.WindowCommand):
             v.set_name(os.path.basename(files[1]) + " -> " + os.path.basename(files[0]))
             v.set_scratch(True)
             v.assign_syntax('Packages/Diff/Diff.sublime-syntax')
+            v.settings().set('translate_tabs_to_spaces', False)
             v.run_command('append', {'characters': difftxt})
+            v.settings().erase('translate_tabs_to_spaces')
 
     def is_visible(self, files):
         return len(files) == 2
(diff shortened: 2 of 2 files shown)


FichteFoll:
In situations like these I prefer to make the specific operation to behave the same, always, and rather make the operation in dependant of other state. In this case, that would be ensuring the setting is disabled before trying to insert text that going to contain literal tabs.

This also makes it more independent, e. g. if users didn't add the settings file.

OdatNurd:
Ah yeah, that's a good point. I was originally thinking of it as an either/or type operation, but it's probably less confusing overall in this case to make sure that the diffs that Sublime generates are always the same regardless.

In regards to the settings themselves, something else to consider is that the syntax is also used if you open `diff` or `patch` files as well. 

My own exposure to `patch` files is that if I'm working with them manually I'm applying patches someone else wrote to code that someone else also wrote. In a case like that, the indent in such files may be different than the norm, which puts the decision as to how they should be indented outside of the usual settings as well.

In that case it might also help to include  `detect_indentation` here as well, though I'm sure how well it would be able to detect with only the context lines to go on, or if having a mostly un-indented file would confuse the detection.

FichteFoll:
`detect_indentation` will likely not help with diffs, as they are indented by a single character, which is one of `[+ -]` and would thus prevent the algorithm from working, at least afaik. It would certainly be nice if the source indentation was somehow detected for diffs.
Regarding the commands to build diffs, they probably don't matter as much, but using the proper characters for these would still be useful when using that diff for other purposes. And having a way to insert text into a view literally without any settings interfering would be neat.

Instead, I'd rather think about adding the likes of `"trim_trailing_white_space_on_save": false`, which users may have overridden but may turn out detrimental when editing diffs/patches and they end up not applying anymore because you stripped trailing whitespace.

OdatNurd:
Based on discussion here, I've proposed a second set of changes instead, which covers a settings change as well as a plugin change.

The commands to generate the diff output now ensure that tab translation is turned off while generating the output, but otherwise leave the settings alone. In particular while researching the idea of patches and how to install them I discovered that the unified diff headers should have tab characters separating the filenames and their modification dates so that spaces in filenames aren't ambiguous. Some patch application tools seem to rely on that.

In regards to settings, it does seem like the better way to go here would be to ensure that saving changes to files only applies changes that you made yourself, so making sure that trimming trailing white space doesn't happen seems like a better idea overall than trying to force one tab state over another, particularly since detection isn't likely to work.

wbond:
There will end up being a change in core that will end up fixing this via a new arg to `append`. I'll leave this open until then, and merge in any necessary changes then.

wbond:
This was fixed with 294517df633106cecbad739cb83c1f7fab8a1709

OdatNurd:
Apologies for this; I finally got time to check yesterday and noticed that this had already been fixed but I hadn't found the time yet to come close this one myself. :)

Issue #2111, opened 2019-10-01
[Matlab] imaginary number scoped as invalid variable name inside parens


fmanar:
The syntax highlighting for matlab scopes constants such as ```1i``` as an invalid variable name when they are inside parenthesis.

![Capture](https://user-images.githubusercontent.com/7545022/65973183-df781080-e438-11e9-97c5-d4a5152cae8f.PNG)

The ```1i``` in line 3 is correctly scoped:
- source.matlab 
- constant.numeric.matlab

The ```1i``` inside parenthesis on line 4 is incorrectly scoped:
- source.matlab 
- meta.parens.matlab 
- invalid.illegal.invalid-variable-name.matlab

I am not an expert, but it seems this is caused by the ```variable_invalid``` context being included before the ```number``` context inside the ```allofem``` context that gets included by ```parens```.  Moving ```variable_invalid``` to the end of the list (from line 112 to 114)  fixed the problem for me, but may have unintended side effects I haven't caught yet.

TheClams:
You are correct, and I do not think this would have unintended consequences considering the order between these two context is correct in the main context (line 22).
I'll make a PR for that.

fmanar:
Cool, thanks!
I'm new to all this github stuff, do I close the issue now or wait till it gets merged? Or neither?

TheClams:
In general it is better to wait for the merge so you can actually see that the issue is fixed.
Thanks for the detailed bug report, it really helps.

wbond:
Now that #2112 is merged, this will be fixed. You can either use the instructions in the readme to use a version from master, or wait for the next dev cycle of Sublime Text to start.

PR #2112, opened 2019-10-01
[Matlab] Fix imaginary number scoped as invalid (#2111)

TheClams:
Simple reorder of context fix the issue.
Add corresponding test.

Fix issue #2111


wbond:
Thanks!

PR #2114, opened 2019-10-02
[D] Fix multiple values in case statement not being parsed correctly

BenjaminSchaaf:
Example:
```d
switch (str) {
    case "foo", "bar":
        break;
}
```


evandrocoan:
May be convert the example into a syntax test?

BenjaminSchaaf:
knew I forgot something!

PR #2118, opened 2019-10-07
[R] Add support for named code sections

deathaxe:
[R] Add support for named code sections

The issue was filed at:
https://forum.sublimetext.com/t/r-named-sections-not-taken-by-sublime/46891

The function is described at:
https://support.rstudio.com/hc/en-us/articles/200484568-Code-Folding-and-Sections

Notes:

As all `#` in front of a section are scoped as `punctuation.` the same is done for normal comments to keep their look synced.

### DIFF
--- a/R/R.sublime-syntax
+++ b/R/R.sublime-syntax
@@ -16,6 +16,7 @@ variables:
 contexts:
   main:
     - include: roxygen
+    - include: codesection
     - include: comments
     - include: constants
     - include: operators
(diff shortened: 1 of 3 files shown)


R/R.sublime-syntax
```diff
@@ -29,8 +30,15 @@ contexts:
     - include: function-calls
     - include: general-variables
 
+  codesection:
+    - match: ^\s*(\#+)\s*(.+?)\s*[-=#]{4,}$\n?
```
wbond: I think this needs to be something like:

```
^\s*(\#+)\s*(.+?)\s*(?:-{4,}|={4,}|#{4,})[ \t]*$\n?
```

The current regex would allow:

```
# Code section -=#=-
```

deathaxe:
@wbond: Resolved

wbond:
Thanks!

PR #2123, opened 2019-10-10
[Go] Numeric Literal Updates (v1.13)

patrickrgaffney:
Adds the following numeric literal updates [introduced in v1.13](https://golang.org/doc/go1.13#language) — all of these changes were part of [recent number literal proposal](https://github.com/golang/proposal/blob/master/design/19308-number-literals.md):

- [Binary integer literals](https://golang.org/ref/spec#Integer_literals).
- [Octal integer literals](https://golang.org/ref/spec#Integer_literals) behind a new (additional) prefix.
- [Hexadecimal floating-point literals](https://golang.org/ref/spec#Floating-point_literals).
- [Imaginary literals](https://golang.org/ref/spec#Imaginary_literals) can now be used with any numeric literal.
- [Numeric separators](https://github.com/golang/proposal/blob/master/design/19308-number-literals.md#digit-separators) can be used in any numeric literal.

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -450,22 +450,18 @@ contexts:
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
     # Hex float literal, no fraction
-    - match: '(0[xX]){{hdigits}}([pP])([+-])?{{ddigits}}(i)?'
+    - match: '(0[xX]){{hdigits}}[pP][+-]?{{ddigits}}(i)?'
       scope: constant.numeric.float.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
-        2: punctuation.section.exponent.go
-        3: keyword.operator.go
-        4: storage.type.numeric.go
+        2: storage.type.numeric.go
     # Hex float literal, with fraction
-    - match: '(0[xX]){{ohdigits}}(\.){{ohdigits}}([pP])([+-])?{{ddigits}}(i)?'
+    - match: '(0[xX]){{ohdigits}}(\.){{ohdigits}}[pP][+-]?{{ddigits}}(i)?'
       scope: constant.numeric.float.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
         2: punctuation.separator.decimal.go
-        3: punctuation.section.exponent.go
-        4: keyword.operator.go
-        5: storage.type.numeric.go
+        3: storage.type.numeric.go
 
   match-integers:
     - include: match-octal-integer
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -432,16 +432,13 @@ contexts:
 
   match-floats:
     # Integer, no fraction, exponent
-    - match: '{{ddigits}}({{exponent}}){{ddigits}}'
+    - match: '{{ddigits}}{{exponent}}{{ddigits}}'
       scope: constant.numeric.float.go
-      captures:
-        1: punctuation.separator.exponent.go
     # Integer, fraction, optional exponent
-    - match: '{{ddigits}}(\.){{ddigits}}(?:({{exponent}}){{ddigits}})?'
+    - match: '{{ddigits}}(\.){{ddigits}}(?:{{exponent}}{{ddigits}})?'
       scope: constant.numeric.float.go
       captures:
         1: punctuation.separator.decimal.go
-        2: punctuation.separator.exponent.go
     # Dot without fraction
     - match: \d+\.(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -449,12 +449,6 @@ contexts:
         3: punctuation.section.exponent.go
         4: keyword.operator.go
         5: storage.type.numeric.imaginary.go
-    # Hex float literal, no fraction, missing exponent digits
-    - match: (0[xX])\h+([pP])([\+\-])?
-      scope: invalid.illegal.go
-    # Hex float literal, with fraction, missing exponent digits
-    - match: (0[xX])\h*(\.)\h*([pP])([\+\-])?
-      scope: invalid.illegal.go
 
   match-integers:
     - include: match-octal-integer
@@ -474,8 +468,6 @@ contexts:
       captures:
         1: punctuation.definition.numeric.octal.go
         2: storage.type.numeric.imaginary.go
-    - match: 0[oO]
-      scope: invalid.illegal.go
 
   match-hex-integer:
     - match: (0[Xx])(?:_)?\h+(?:_\h+)*(i)?
@@ -490,8 +482,6 @@ contexts:
       captures:
         1: punctuation.definition.numeric.binary.go
         2: storage.type.numeric.imaginary.go
-    - match: 0[bB]
-      scope: invalid.illegal.go
 
   match-decimal-integer:
     - match: \d+(?:_\d+)*
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -433,7 +433,7 @@ contexts:
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
     # Hex float literal, no fraction
-    - match: (0[x|X])(?:_)?\h+(?:_\h+)*([p|P])([\+|\-])?\d+(?:_\d+)*(i)?
+    - match: (0[xX])(?:_)?\h+(?:_\h+)*([pP])([\+\-])?\d+(?:_\d+)*(i)?
       scope: constant.numeric.float.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
@@ -441,7 +441,7 @@ contexts:
         3: keyword.operator.go
         4: storage.type.numeric.imaginary.go
     # Hex float literal, with fraction
-    - match: (0[x|X])(?:_)?\h*(?:_\h+)*(\.)\h*(?:_\h+)*([p|P])([\+|\-])?\d+(?:_\d+)*(i)?
+    - match: (0[xX])(?:_)?\h*(?:_\h+)*(\.)\h*(?:_\h+)*([pP])([\+\-])?\d+(?:_\d+)*(i)?
       scope: constant.numeric.float.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
@@ -450,10 +450,10 @@ contexts:
         4: keyword.operator.go
         5: storage.type.numeric.imaginary.go
     # Hex float literal, no fraction, missing exponent digits
-    - match: (0[x|X])\h+([p|P])([\+|\-])?
+    - match: (0[xX])\h+([pP])([\+\-])?
       scope: invalid.illegal.go
     # Hex float literal, with fraction, missing exponent digits
-    - match: (0[x|X])\h*(\.)\h*([p|P])([\+|\-])?
+    - match: (0[xX])\h*(\.)\h*([pP])([\+\-])?
       scope: invalid.illegal.go
 
   match-integers:
@@ -469,12 +469,12 @@ contexts:
         1: punctuation.definition.numeric.octal.go
     - match: 0[0-7]*[8-9]+
       scope: invalid.illegal.go
-    - match: (0[o|O])(?:_)?[0-7]+(?:_[0-7]+)*(i)?
+    - match: (0[oO])(?:_)?[0-7]+(?:_[0-7]+)*(i)?
       scope: constant.numeric.octal.go
       captures:
         1: punctuation.definition.numeric.octal.go
         2: storage.type.numeric.imaginary.go
-    - match: 0[o|O]
+    - match: 0[oO]
       scope: invalid.illegal.go
 
   match-hex-integer:
@@ -485,12 +485,12 @@ contexts:
         2: storage.type.numeric.imaginary.go
 
   match-binary-integer:
-    - match: (0[b|B])[_]?[1|0]+(?:_[0|1]+)*(i)?
+    - match: (0[bB])[_]?[01]+(?:_[01]+)*(i)?
       scope: constant.numeric.binary.go
       captures:
         1: punctuation.definition.numeric.binary.go
         2: storage.type.numeric.imaginary.go
-    - match: 0[b|B]
+    - match: 0[bB]
       scope: invalid.illegal.go
 
   match-decimal-integer:
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -433,10 +433,10 @@ contexts:
   match-floats:
     # Integer, no fraction, exponent
     - match: '{{ddigits}}{{exponent}}{{ddigits}}'
-      scope: constant.numeric.float.go
+      scope: constant.numeric.float.decimal.go
     # Integer, fraction, optional exponent
     - match: '{{ddigits}}(\.){{ddigits}}(?:{{exponent}}{{ddigits}})?'
-      scope: constant.numeric.float.go
+      scope: constant.numeric.float.decimal.go
       captures:
         1: punctuation.separator.decimal.go
     # Dot without fraction
@@ -447,13 +447,13 @@ contexts:
       scope: invalid.deprecated.go
     # Hex float literal, no fraction
     - match: '(0[xX]){{hdigits}}[pP][+-]?{{ddigits}}(i)?'
-      scope: constant.numeric.float.go
+      scope: constant.numeric.float.hexadecimal.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
         2: storage.type.numeric.go
     # Hex float literal, with fraction
     - match: '(0[xX]){{ohdigits}}(\.){{ohdigits}}[pP][+-]?{{ddigits}}(i)?'
-      scope: constant.numeric.float.go
+      scope: constant.numeric.float.hexadecimal.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
         2: punctuation.separator.decimal.go
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -412,6 +412,7 @@ contexts:
       scope: invalid.deprecated.go
     # Dot without integer
     - match: \.\d+(?:{{exponent}}\d+)?i
+      scope: invalid.deprecated.go
 
   match-floats:
     # Integer, no fraction, exponent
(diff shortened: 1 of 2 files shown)


Go/Go.sublime-syntax
```diff
@@ -432,28 +432,69 @@ contexts:
     # Dot without integer
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
+    # Hex float literal, no fraction
+    - match: (0[x|X])(?:_)?\h+(?:_\h+)*([p|P])([\+|\-])?\d+(?:_\d+)*(i)?
```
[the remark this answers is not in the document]
patrickrgaffney: Oof. Nice catch, thanks!

Go/Go.sublime-syntax
```diff
@@ -432,28 +432,69 @@ contexts:
     # Dot without integer
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
+    # Hex float literal, no fraction
+    - match: (0[xX])(?:_)?\h+(?:_\h+)*([pP])([\+\-])?\d+(?:_\d+)*(i)?
```
FichteFoll: `(?:_)?` is equivalent to `_?`.

Go/Go.sublime-syntax
```diff
@@ -432,28 +432,69 @@ contexts:
...
+        2: punctuation.separator.decimal.go
+        3: punctuation.section.exponent.go
+        4: keyword.operator.go
+        5: storage.type.numeric.imaginary.go
+    # Hex float literal, no fraction, missing exponent digits
+    - match: (0[xX])\h+([pP])([\+\-])?
+      scope: invalid.illegal.go
+    # Hex float literal, with fraction, missing exponent digits
+    - match: (0[xX])\h*(\.)\h*([pP])([\+\-])?
+      scope: invalid.illegal.go
```
FichteFoll: We try not to get in the way of the user with attention-seeking highlighting of invalid syntax that easily occurs while typing. I believe these would be a good example of that. If there was no highlighting for these at all, it should be enough to signal that these aren't proper numeric constants.

Go/Go.sublime-syntax
```diff
@@ -432,28 +432,69 @@ contexts:
...
   match-hex-integer:
-    - match: (0[Xx])\h+
+    - match: (0[Xx])(?:_)?\h+(?:_\h+)*(i)?
       scope: constant.numeric.hex.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
+        2: storage.type.numeric.imaginary.go
+
+  match-binary-integer:
+    - match: (0[bB])[_]?[01]+(?:_[01]+)*(i)?
```
FichteFoll: `[_]?` is also equivalent to `_?`.

Go/Go.sublime-syntax
```diff
@@ -432,28 +432,69 @@ contexts:
...
         1: punctuation.definition.numeric.octal.go
     - match: 0[0-7]*[8-9]+
       scope: invalid.illegal.go
+    - match: (0[oO])(?:_)?[0-7]+(?:_[0-7]+)*(i)?
+      scope: constant.numeric.octal.go
+      captures:
+        1: punctuation.definition.numeric.octal.go
+        2: storage.type.numeric.imaginary.go
+    - match: 0[oO]
+      scope: invalid.illegal.go
```
FichteFoll: With the above in mind, I'm unsure about `0o`. It is invalid, but it'd be hard to find this token in the wild outside of while typing. There is no match like this for `0x`, btw.

Go/Go.sublime-syntax
patrickrgaffney: Good call. Removed all the unnecessary passive groups.

Go/Go.sublime-syntax
patrickrgaffney: Sounds good. I noticed a couple of `invalid.[illegal|deprecated]` in there and probably went a little overboard adding new ones. I removed all the `illegal.*` scopes I added.

Go/Go.sublime-syntax
```diff
@@ -401,7 +401,7 @@ contexts:
 
   match-imaginary:
     # Mandatory integer, optional fraction, optional exponent, mandatory imaginary
-    - match: \d+(?:(\.)\d+)?(?:({{exponent}})\d+)?(i)
+    - match: \d+(?:_\d+)*(?:(\.)\d+(?:_\d+)*)?(?:({{exponent}})\d+(?:_\d+)*)?(i)
```
deathaxe: Would you considder a variable for `(?:\d+(?:_\d+)*)` ?

Recently found Python doing so and found it helpful with regards to readability. I therefore used and extended its use in https://github.com/sublimehq/Packages/pull/2136

```
  ddigit: '\d+(?:_\d+)*'
  hdigit: '\h+(?:_\h+)*'
```

The match would look like

```
 - match: '{{ddigit}}(?:(\.){{ddigit}})?(?:({{exponent}}){{ddigit}})?(i)'
```

Go/Go.sublime-syntax
```diff
@@ -401,7 +401,7 @@ contexts:
...
       scope: constant.numeric.imaginary.go
```
deathaxe: According to scope naming guidelines `constant.numeric.imaginary.go` would need to be renamed to `constant.numeric.complex.imaginary.go`

Go/Go.sublime-syntax
deathaxe: A further step might be to update the `exponent` variable, which would then effect many more places, of cource.

Example:

```
  exponent: '(?:([Ee])[-+]?{{ddigit}})'
```

The match above would then look like

```
 - match: '{{ddigit}}(?:(\.){{ddigit}})?{{exponent}}?(i)'
```

This is how python handles it.

_Note:_

This change excludes the sign of the exponent from the punctuation scope, which I find is wrong. If so only `e` should be scoped as separator, but I wonder whether it should be scoped as punctuation at all.

FichteFoll: Note that the variables should look more like this (plural s and non-capturing group):
```
ddigits: (?:\d+(?:_\d+)*)
hdigits: (?:\h+(?:_\h+)*)
```

Go/Go.sublime-syntax
patrickrgaffney: ✔️ Renamed.

Go/Go.sublime-syntax
patrickrgaffney: I added the `ddigits` and `hdigits` variables. I agree they improve readability.

Go/Go.sublime-syntax
```diff
@@ -432,28 +439,59 @@ contexts:
...
 
   match-integers:
     - include: match-octal-integer
     - include: match-hex-integer
+    - include: match-binary-integer
     - include: match-decimal-integer
 
   match-octal-integer:
-    - match: (0)[0-7]+(?=\D)
+    - match: (0)_?[0-7]+(?:_[0-7]+)*(?=\D)
```
deathaxe:
```
variables:
    odigits: (?:_?[0-7]+)+

# ...

    - match: (0){{odigits}}(?=\D)
```

Go/Go.sublime-syntax
```diff
@@ -432,28 +439,59 @@ contexts:
...
       scope: constant.numeric.octal.go
       captures:
         1: punctuation.definition.numeric.octal.go
     - match: 0[0-7]*[8-9]+
       scope: invalid.illegal.go
+    - match: (0[oO])_?[0-7]+(?:_[0-7]+)*(i)?
```
deathaxe:
```
    - match: (0[oO]){{odigits}}(i)?
```

Go/Go.sublime-syntax
```diff
@@ -432,28 +439,59 @@ contexts:
...
   match-hex-integer:
-    - match: (0[Xx])\h+
+    - match: (0[Xx])_?{{hdigits}}(i)?
       scope: constant.numeric.hex.go
       captures:
         1: punctuation.definition.numeric.hexadecimal.go
+        2: storage.type.numeric.imaginary.go
+
+  match-binary-integer:
+    - match: (0[bB])_?[01]+(?:_[01]+)*(i)?
```
deathaxe:
```
variables:
    bdigits: (?:_?[01]+)+

# ...

    - match: (0[bB]){{bdigits}}(i)?
```

Go/Go.sublime-syntax
```diff
@@ -86,6 +86,13 @@ variables:
 
   exponent: '[Ee][+-]?'
 
+  # Matches a digit with any number of numeric separators, while
+  # not allowing a numeric separator as the last or first character.
+  ddigits: (?:\d+(?:_\d+)*)
+
+  # Hexideciaml counterpart to ddigits.
+  hdigits: (?:\h+(?:_\h+)*)
```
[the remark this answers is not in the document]
jwortmann: I just want to mention that `(?:_?\h+)+` might be dangerous, because if I understand it right, that regex is ambiguous in how to match consecutive numbers, see https://snyk.io/blog/redos-and-catastrophic-backtracking/. I believe `(?:_?\h)+` would do the same, but prevents backtracking.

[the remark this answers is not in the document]
patrickrgaffney: [From the docs](https://golang.org/ref/spec#Integer_literals):

> ...an underscore character _ may appear after a base prefix or between successive digits...

The `_?{{hdigits}}` is catching that second case where it can appear between the base prefix and the first digit.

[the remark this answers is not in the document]
patrickrgaffney: I think adding the `_?` to the beginning of the `hdigits` variable adds to readability. Implemented in 62696fef.

Go/Go.sublime-syntax
```diff
@@ -432,28 +439,59 @@ contexts:
     # Dot without integer
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
+    # Hex float literal, no fraction
+    - match: '(0[xX])_?{{hdigits}}([pP])([\+\-])?{{ddigits}}(i)?'
```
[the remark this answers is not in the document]
patrickrgaffney: Implemented in 62696fe.

Go/Go.sublime-syntax
```diff
@@ -432,28 +439,59 @@ contexts:
...
       scope: invalid.illegal.go
+    - match: (0[oO])_?[0-7]+(?:_[0-7]+)*(i)?
+      scope: constant.numeric.octal.go
+      captures:
+        1: punctuation.definition.numeric.octal.go
+        2: storage.type.numeric.imaginary.go
 
   match-hex-integer:
-    - match: (0[Xx])\h+
+    - match: (0[Xx])_?{{hdigits}}(i)?
```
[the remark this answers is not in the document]
patrickrgaffney: Implemented in 62696fe.

Go/Go.sublime-syntax
patrickrgaffney: Implemented in 71a38577.

Go/Go.sublime-syntax
patrickrgaffney: Implemented in 71a38577.

Go/Go.sublime-syntax
patrickrgaffney: Implemented in 3df3b9f1.

Go/Go.sublime-syntax
```diff
@@ -432,28 +439,59 @@ contexts:
...
+    # Hex float literal, no fraction
+    - match: '(0[xX])_?{{hdigits}}([pP])([\+\-])?{{ddigits}}(i)?'
+      scope: constant.numeric.float.go
+      captures:
+        1: punctuation.definition.numeric.hexadecimal.go
+        2: punctuation.section.exponent.go
+        3: keyword.operator.go
+        4: storage.type.numeric.imaginary.go
+    # Hex float literal, with fraction
+    - match: '(0[xX])_?\h*(?:_\h+)*(\.)\h*(?:_\h+)*([pP])([\+\-])?{{ddigits}}(i)?'
```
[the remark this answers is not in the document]
patrickrgaffney: This one is trickier, as it uses `_?\h*` to catch the valid case where there is no hex character between the base and the decimal — the coefficient is optional, i.e. `0X.8p-0`. I couldn't come up with a good resolution without adding another variable with an optional leading hex character. I added that in e59a065b — but I'm open to other suggestions.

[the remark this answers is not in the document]
deathaxe: Yes, numbers can be tricky - especially the incomplete ones. Learned a lot while working on them for the several syntaxes. I guess Java uses the same kind of hexadecimal numbers. Maybe the related Java PR #2155 can provide some inspiration.

I've tried to put all variants into a single match pattern for performance reasons:

https://github.com/sublimehq/Packages/blob/7c699aeca4550217d45c6237dbc11d9d963ba578/Java/Java.sublime-syntax#L502-L509

_Note: As Java had already scoped some trailing/leading underscores `invalid`, I kept doing so. Not sure whether we really need it, but I found the strategy to try to keep the coloring of numbers consistent while typing a good idea.

While working on Java I learned the `{{ddigits}}` suggested here cause the whole number's color to flicker while typing. That's what I found even the old Java implementation did better._

Go/Go.sublime-syntax
```diff
@@ -432,28 +449,59 @@ contexts:
     # Dot without integer
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
+    # Hex float literal, no fraction
+    - match: '(0[xX]){{hdigits}}([pP])([+-])?{{ddigits}}(i)?'
+      scope: constant.numeric.float.go
+      captures:
+        1: punctuation.definition.numeric.hexadecimal.go
+        2: punctuation.section.exponent.go
+        3: keyword.operator.go
```
jwortmann: How about removing the special scoping for the exponent (`E+`, `p+`, ...) which is used in this syntax (`punctuation.separator.exponent` in some places and `punctuation.section.exponent` for this example), because it seems to not be used elsewhere (there is a PR #2166 to remove `punctuation.definition.exponent` in Erlang)? In my opinion it doesn't add much value and doesn't really look well in color schemes that target `punctuation.separator`. Also `keyword.operator` for the `+`/`-` doesn't make sense here in my opinion.

patrickrgaffney: Good idea. Looking at the other syntaxes it seems most don't have extra scopes for the exponent and/or `+`/`-` sections of numeric literals. Removed in f894f19, 1947eca, and 8c2b119.

patrickrgaffney:
Changed the scopes on imaginary literals to align with #2135 (and the other recent numeric literal improvement PRs). Thanks everyone.

Go/Go.sublime-syntax
```diff
@@ -416,44 +432,68 @@ contexts:
...
     - match: \d+\.(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
     # Dot without integer
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
+    # Hex float literal, no fraction
+    - match: '(0[xX]){{hdigits}}[pP][+-]?{{ddigits}}(i)?'
+      scope: constant.numeric.float.go
+      captures:
+        1: punctuation.definition.numeric.hexadecimal.go
```
deathaxe: Sorry for nagging, but would you please mind

1. to add the `[decimal|hexadecimal]` subscope to the `constant` scopes (here:  `constant.numeric.float.hexadecima.go`)
2. to use `punctuation.definition.numeric.base.go` as common scope for all the punctuation?

Go/Go.sublime-syntax
```diff
@@ -416,44 +432,68 @@ contexts:
...
   match-octal-integer:
-    - match: (0)[0-7]+(?=\D)
+    - match: (0){{odigits}}(?=\D)
       scope: constant.numeric.octal.go
       captures:
         1: punctuation.definition.numeric.octal.go
     - match: 0[0-7]*[8-9]+
       scope: invalid.illegal.go
+    - match: (0[oO]){{odigits}}(i)?
+      scope: constant.numeric.octal.go
```
deathaxe: Is this a `constant.numeric.integer.octal.go`?

Go/Go.sublime-syntax
```diff
@@ -416,44 +432,68 @@ contexts:
...
+      captures:
+        1: punctuation.definition.numeric.octal.go
+        2: storage.type.numeric.go
 
   match-hex-integer:
-    - match: (0[Xx])\h+
+    - match: (0[xX]){{hdigits}}(i)?
       scope: constant.numeric.hex.go
```
deathaxe: Is this a `constant.numeric.integer.hexadecimal.go`?

Go/Go.sublime-syntax
```diff
@@ -416,44 +432,68 @@ contexts:
...
+    - match: (0[bB]){{bdigits}}(i)?
+      scope: constant.numeric.binary.go
+      captures:
+        1: punctuation.definition.numeric.binary.go
+        2: storage.type.numeric.go
 
   match-decimal-integer:
-    - match: \d+
+    - match: '{{ddigits}}'
       scope: constant.numeric.integer.go
```
deathaxe: `constant.numeric.integer.decimal`?

Go/Go.sublime-syntax
patrickrgaffney: Resolved in 8548792 and 3b30fb7.

Go/Go.sublime-syntax
patrickrgaffney: Resolved in 7500bd7.

Go/Go.sublime-syntax
patrickrgaffney: Resolved in 167069b.

Go/Go.sublime-syntax
patrickrgaffney: Resolved in 7500bd7.

Go/Go.sublime-syntax
```diff
@@ -416,45 +432,69 @@ contexts:
...
+      scope: constant.numeric.float.decimal.go
       captures:
         1: punctuation.separator.decimal.go
-        2: punctuation.separator.exponent.go
     # Dot without fraction
     - match: \d+\.(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
     # Dot without integer
     - match: \.\d+(?:{{exponent}}\d+)?
       scope: invalid.deprecated.go
```
deathaxe: As the `invalid.deprecated` scopes are not specific, They might be merged with the ones from line 471 and 427?

```
(?:\d+\.|\.\d+)(?:{{exponent}}\d+)?i?
```

... maybe in a dedicated context at the end as it would then match both imaginary and real floats?

Go/Go.sublime-syntax
```diff
@@ -416,45 +432,69 @@ contexts:
...
+    # Hex float literal, no fraction
+    - match: '(0[xX]){{hdigits}}[pP][+-]?{{ddigits}}(i)?'
```
deathaxe: Just see the `(i)?` here. With regards to PR #2135 we should propably scope hexadecimal imaginary numbers as `constant.numeric.imaginary.hexadecimal` in Go as well?

deathaxe:
Except the two suggestions the PR looks good to me.

Go/Go.sublime-syntax
patrickrgaffney: Resolved in e921238.

Go/Go.sublime-syntax
patrickrgaffney: Resolved in ea7018d. The same was true for hex, binary, and octal integers — so I updated those as well. Thanks.

deathaxe:
Did some performance tweaks at https://github.com/patrickrgaffney/Packages/pull/1

wbond:
Thanks for your work on this @patrickrgaffney! Also, thanks to everyone who took time to review!

PR #2124, opened 2019-10-10
[Markdown] Add link/ref/image attribute support

### LINKED ISSUE #2117: Underscore in attribute values starts italics
```
>
> One day, a [link](https://example.org){:target="_blank"} was made
>
```

Screenshot 2019-10-05 at 18 01 15

This uses Karmdown-style attributes, such as used by Jekyll and GitHub Pages.

deathaxe:
Fixes #2117

This PR adds support for attributes. The implementation is borrowed from the HTML.sublime-syntax.

The different styles of attribute names (with or without `:`) are not handled in a special way.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -167,6 +167,11 @@ variables:
         (?i:address|article|aside|base|basefont|blockquote|body|caption|center|col|colgroup|dd|details|dialog|dir|div|dl|dt|fieldset|figcaption|figure|footer|form|frame|frameset|h1|h2|h3|h4|h5|h6|head|header|hr|html|iframe|legend|li|link|main|menu|menuitem|meta|nav|noframes|ol|optgroup|option|p|param|section|source|summary|table|tbody|td|tfoot|th|thead|title|tr|track|ul)
         (?:\s|$|/?>)
       )
+    ascii_space: '\t\n\f '
+    tag_attribute_name_start: (?=[^{{ascii_space}}=/>}])
+    tag_attribute_name_break: (?=[{{ascii_space}}=/>}])
+    tag_unquoted_attribute_start: (?=[^{{ascii_space}}=/>}])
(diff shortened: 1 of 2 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -608,6 +613,15 @@ contexts:
               scope: invalid.illegal.unexpected-whitespace.markdown
             - match: (?=\s*(?!\)))
               push: link-title
+  image-inline-attr:
+    - match: '([ ]*)(\{)' # spaces not allowed before the open paren, but we check for it anyway to mark it as invalid
```
FichteFoll: Since this is non-standard (i.e. not in commonmark), we should probably stay on the safe side and only highlight correct syntax.

FichteFoll:
With how similar all these `image-*` and `link-*` contexts are (at least the attr contexts only seem to differ in their meta scope), wouldn't it make sense to add a different meta_scope to the entire stack as the first context to push and then push the others on top of it?

May not be the right PR for this, but I haven't looked at the Markdown def in a long time and this really does seem redundant and overly complex. It's not like we're talking about important meta scopes like with `meta.function{,.parameters,.body}` etc. either.

deathaxe:
It definitely would make sense, but as you correctly stated. This is not the correct PR for that, because doing this change in "coding style" would need to effect many more places which are definitely out of scope here.

At least some of the existing contexts would propably be able to reworked in order to reuse more things and avoid duplications, but this would lead to bigger changes. ... Tried and abandond those changes for now.

The current implementation of this PR is to implement the requested feature/fix without breaking the existing coding style and way of doing things too much.

The existing contexts use scopes like `meta.link.inline.description` etc. which would need to be renamed. Use of `clear_scope` to maintain them didn't work well.

The markdown syntax is an exaustive `lookahead`,  `captures` and `set` hell anyway, which can't be fixed here.

wbond:
Can we add a comment to the syntax definition that links to where this format is defined since it appears to not be part of CommonMark?

deathaxe:
@wbond: References added

PR #2130, opened 2019-10-18
[JavaScript] Signs aren't part of numeric literals

Thom1729:
Since #1552, preceding signs are scoped as part of numeric literals. This is wrong. I have no idea why I thought that that was correct. I can only plead temporary insanity. This PR restores order.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1727,32 +1727,32 @@ contexts:
         - include: object-property
 
   literal-number:
-    - match: '[+-]?0[0-9]+{{identifier_break}}'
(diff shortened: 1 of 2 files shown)


FichteFoll:
So what about the `+` being part of the literal but the `-` being a unary operator? Not sure if it works like that, but `+` probably isn't a unary operator.

Thom1729:
`+` is a unary operator which converts its operand to a number. (`-` is a unary operator that converts its operand to a number, then returns the additive inverse of that number.)

FichteFoll:
Right, I forgot we were talking about JavaScript where type-coercion is a thing. Then you probably *can* make use of a `+` unary operator.

PR #2132, opened 2019-10-18
[JavaScript] Add optional chaining and nullish coalescing.

Thom1729:
[Nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing) (Stage 3)
[Optional chaining](https://github.com/tc39/proposal-optional-chaining) (Stage 3)

The regexp to match a dotted property access is now `\.(?!\?[\[(])\??`, which is regrettable.


michaelblyons:
> The regexp to match a dotted property access is now `\.(?!\?[\[(])\??`, which is regrettable.

Do you want to make that a variable? It sees a lot of reuse, and it's no longer trivial.

PR #2045, opened 2019-08-08
[PHP] Updated PHP Syntax to include PHP 7+ functions

CameronHall:
Added functions mentioned in PHP 7 Changelog
- oci_set_call_timeout
- oci_set_db_operation
- gmp_binomial
- gmp_lcm
- gmp_perfect_power
- gmp_kronecker
- openssl_pkey_derive
- is_countable
- hrtime
- ftp_append
- spl_object_id
- hash_hkdf
- is_iterable
- socket_export_stream
- error_clear_last
- http_response_code

PR #2134, opened 2019-10-18
[PHP] Applied regex expansion to most PHP regex matches

CameronHall:
I've applied the regex expansion to most of the `matches` attributes in the YAML file - as discussed in https://github.com/sublimehq/Packages/pull/2045.

I've also fixed a regex balancing error on the socket functions.


PHP/PHP Source.sublime-syntax
```diff
@@ -208,65 +233,102 @@ contexts:
...
-        - include: class-name
-        - include: variables
-    - include: instantiation
-    - match: '(?i)(goto)\s+({{identifier}})'
-      captures:
-        1: keyword.control.goto.php
-        2: support.other.php
-    - include: string-backtick
-    - match: '(\[)\s*(\])'
+        - match: (?i)\s*\(\s*(array|real|double|float|int(eger)?|bool(ean)?|string|object|binary|unset)\s*\)
```
jfcherng: the indentation is wrong here.

PHP/PHP Source.sublime-syntax
```diff
@@ -208,65 +233,102 @@ contexts:
...
+              | object
+              | real
+              | string
+            )\b
+          scope: storage.type.php
+        - match: |-
+            (?ix)\b(?:
+              parent
+              | self
+            )\b
```
jfcherng: Personally, I am not a fan about expanding short regexes like this. I think most of problems come from compact regexes for constants and function names.

FichteFoll:
While the current condensed form of the branches is nigh unreadable, I don't think expanding all of them into their own line is exactly the solution either, as it makes the syntax definition exremely large. Since they are supposed to be read by humans, I believe some special discretion is needed.

For example, groups with only a couple branches (let's say up to 8) do not need to be split per line as the necessary iformation can be grasped as a whole. The comment on `parent|self` is a good example of that.

Another thing I like to do is have multiple branches per line up to a rough column count and ideally not grouped alphabetically but categorically. Categories don't make a lot of sense for Php's many builtins, but [here's an example in the Python syntax](https://github.com/sublimehq/Packages/blob/3fff882d122f3f4f30a2ac0a01e0273bf078709d/Python/Python.sublime-syntax#L1211-L1257). The disadvantage of alphabetically (or lexicographically) sorting a list is the wrapping problem when you eventually decide a line to be too long. Section borders, e.g. a section for each first letter, can help mitiage this issue and won't cause the entire list to be rewrapped and produce an unreadable diff.

PHP/PHP Source.sublime-syntax
```diff
@@ -128,7 +128,32 @@ contexts:
           pop: true
         - match: (?=\S)
           pop: true
-    - match: \b(break|c(ase|ontinue)|d(e(clare|fault)|ie|o)|e(lse(if)?|nd(declare|for(each)?|if|switch|while)|xit)|for(each)?|if|return|switch|use|while)\b
+    - match: |-
+        (?x)\b(?:
```
deathaxe: You could easily use `\b(?x:` to simplify the term a bit. It also applies to most of the following hunks.

PHP/PHP Source.sublime-syntax
```diff
@@ -208,65 +233,102 @@ contexts:
...
-      captures:
-        1: keyword.control.goto.php
-        2: support.other.php
-    - include: string-backtick
-    - match: '(\[)\s*(\])'
+        - match: (?i)\s*\(\s*(array|real|double|float|int(eger)?|bool(ean)?|string|object|binary|unset)\s*\)
+          captures:
+            1: storage.type.php
+        - match: |-
+            (?ix)\b(?:
```
deathaxe: ... `\b(?xi:` ...

PHP/PHP Source.sublime-syntax
```diff
@@ -568,9 +865,20 @@ contexts:
       push:
         - meta_content_scope: meta.function.php
         - include: comments
-        - match: '(?i)\b(final|abstract|var|public|private|protected|static)\b'
+        - match: '(?i)(?x)\b(?:
```
deathaxe: I'd prefere `\b(?xi:` over `(?i)(?x)\b(?:`

deathaxe: ... no need for quotes here.

PHP/PHP Source.sublime-syntax
CameronHall: I agree, I'll sort this out when I come back to it.

PHP/PHP Source.sublime-syntax
CameronHall: Yeah, that's the problem I get when I copy paste the expanded regex from what is generated by the script 😞 I'll sort it out when I'm at my desktop, it's a little difficult to identify the issues on such a small screen.

CameronHall:
I'll come back and sort this one out soon. @deathaxe I'll do my best to simplify the regex, though that's the goal of this PR is legibility and expansion, pushing for too much simplification will only lead to bikeshedding. 

@FichteFoll I think eight sounds good. I'd consider the "scope" sections be a good definition of the relevant category, then we just order the function names attributed to the scope alphabetically. What do you think?

FichteFoll:
I'm not sure what you mean with "scope sections".

CameronHall:
Take BCMath for example;

```

**Comment by CameronHall** [comment]
Take BCMath for example;

```yaml
 - match: |-
        \b(?ix)(?:
          bcadd | bccomp | bcdiv | bcmod | bcmul | bcpow | bcpowmod | bcscale |
          bcsqrt | bcsub
        )\b
  scope: support.function.bcmath.php
```

PHP/PHP Source.sublime-syntax
```diff
@@ -798,19 +848,190 @@ contexts:
...
+        - match: |-
+            (\\)?\b(
+              ABDAY_1 | ABDAY_2 | ABDAY_3 | ABDAY_4 | ABDAY_5 | ABDAY_6 | ABDAY_7 | ABMON_1 |
+              ABMON_10 | ABMON_11 | ABMON_12 | ABMON_2 | ABMON_3 | ABMON_4 | ABMON_5 | ABMON_6 |
+              ABMON_7 | ABMON_8 | ABMON_9 | ALT_DIGITS | AM_STR | ASSERT_ACTIVE | ASSERT_BAIL | ASSERT_CALLBACK |
```
FichteFoll: In for cases like these, I actually find it more reable to write `ABDAY_[1-7]|ABMON_(?:[1-9]|1[0-2])`, which tells you immediately what numbers are allowed and whether all of them are covered, compared to going through each of them.

PHP/PHP Source.sublime-syntax
```diff
@@ -1342,201 +1563,767 @@ contexts:
...
+    - match: |-
+        \b(?xi:
+          mysqli_affected_rows | mysqli_autocommit | mysqli_bind_param | mysqli_bind_result | mysqli_change_user | mysqli_character_set_name | mysqli_client_encoding | mysqli_close |
```
FichteFoll: One could argue to put the `mysqli` in front of this entire match, because the prefix is the same for all these items. No strong preference on my part.
Some for other matches like this.

CameronHall: I think we should leave it in the most expanded variant here, only because if you start doing what you've mentioned here; we might as well apply the same logic to `ibase`,`bcmath, `mysql`, `mhash`,`openssl`, `pgsql` and `oct` just to name a few.

What are your thoughts?

PHP/PHP Source.sublime-syntax
CameronHall: I've updated `ABDAY_`, `ABMONTH_`, `DAY_`, `MONTH_` to use the expression you suggested.

CameronHall:
@FichteFoll I understand what you mean now by categorising things. I think the functions are categorised quite well, however the constants could do with comments to categorise them. I'm not really sure if that's worth the time investment - certainly not in the PR.

PHP/PHP Source.sublime-syntax
jfcherng: I am actually not sure about this. Once you begin doing this, you probably kind of going back with a "this maybe okay but I am not sure" state.

CameronHall: I have no opinion on the matter. I believe they both the expanded and condensed versions hold the same value, but I'm really indifferent to either way.

FichteFoll: I'm of the opinion that there is no silver line for compact vs expanded and that things like this should be decided on a case-by-case basis. This is for humans to read, after all.

PHP/PHP Source.sublime-syntax
FichteFoll: Well, yes, I just only commented on this item because repeating the same message seemed redundant. I'm fine with either, just wanted to bring this to attention.

FichteFoll:
By the way, the ABMON/DAY thing also applies to `DAY_1` and `MON_1`, which occur later in the match.

wbond:
Just to clarify, were these expanded by hand, or via a script?

jfcherng:
@wbond script https://github.com/sublimehq/Packages/pull/2045#discussion_r336170580

wbond:
I just want to confirm that. Mostly to see if more extensive testing of the expanded forms needs to be done, if it was done by hand.

CameronHall:
@FichteFoll I went along and condensed `DAY_X` and MON_X` when I did `ABMON` and `ABDAY` :)

@wbond this was a script. There were some cases where the boundaries and predicates needed to be updated by hand; mainly what @deathaxe was referring too.
```
and four expressions which I have condensed in the constants list.

FichteFoll:
Ah, apparently I missed the last commit during my review.

wbond:
Thanks for all of the work and review on this, it definitely is much more maintainable now!

PR #2135, opened 2019-10-18
[Ruby] Improve numbers

### LINKED ISSUE #2126: [Ruby] Rational and Complex number literals are not highlighted at all
In Ruby, numeric literals that are followed immediately by `r` or `i` are Rational and Complex literals, respectively. They can also be combined as `ri` (but not `ir`).

[Link to documentation](https://ruby-doc.org/core-2.6.5/doc/syntax/literals_rdoc.html#label-Rational+numbers)

Examples that should be highlighted as `constant.numeric.ruby`:

``` ruby
0.1r  # => (1/10)
0.1i  # => (0+0.1i)
0.1ri # => (0+(1/10)*i)

### PR DESCRIPTION
**Pull request by jwortmann**
This PR

1. Fixes #2126
   I used the scope ~~`constant.numeric.integer.rational.ruby` for e.g. `12r` and `constant.numeric.float.rational.ruby` for `12.3r`~~ `constant.numeric.rational.ruby` for e.g. `12r` and `12.3r`.
2. Fixes numbers with decimal or octal prefix, which are not highlighted currently
3. ~~Changes the dot in e.g. `1.` to be included in the number, instead of having the scope `punctuation.accessor` (which is wrong in my opinion)~~
4. Applies more specific scopes `integer`, `float`, `decimal`, `binary`, `octal`, `hexadecimal` instead of just `constant.numeric.ruby`

Notice: What is the purpose of the negative lookahead `(?![^[:space:][:digit:]])`, which is currently used in the regex after a decimal dot? I don't see any benefit from it, so I deleted it. In case it was to allow something like `1.`, it didn't work anyway, because there is still a word boundary `\b` required at the end.

The examples in the syntax test file were taken from https://ruby-doc.org/core-2.6.3/doc/syntax/literals_rdoc.html#label-Numbers

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -209,51 +209,46 @@ contexts:
     - match: '\b(__(FILE|LINE|ENCODING)__|self)\b(?![?!])'
       scope: variable.language.ruby
     # numbers with decimal base prefix
-    - match: '\b(0[dD]){{ddigits}}(r)?(?:(\.i)|(i)(r)?)?\b'
+    - match: '\b(0[dD]){{ddigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.decimal.ruby
       captures:
         1: punctuation.definition.numeric.decimal.ruby
         2: storage.type.numeric.rational.ruby
         3: storage.type.numeric.imaginary.ruby
-        4: storage.type.numeric.imaginary.ruby
-        5: invalid.illegal.numeric.rational.ruby
+        4: invalid.illegal.numeric.rational.ruby
     # numbers with hexadecimal base prefix
-    - match: '\b(0[xX]){{hdigits}}(r)?(?:(\.i)|(i)(r)?)?\b'
+    - match: '\b(0[xX]){{hdigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.hexadecimal.ruby
       captures:
         1: punctuation.definition.numeric.hexadecimal.ruby
         2: storage.type.numeric.rational.ruby
         3: storage.type.numeric.imaginary.ruby
-        4: storage.type.numeric.imaginary.ruby
-        5: invalid.illegal.numeric.rational.ruby
+        4: invalid.illegal.numeric.rational.ruby
     # numbers with octal base prefix
-    - match: '\b(0[oO]?){{odigits}}(r)?(?:(\.i)|(i)(r)?)?\b'
+    - match: '\b(0[oO]?){{odigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.octal.ruby
       captures:
         1: punctuation.definition.numeric.octal.ruby
         2: storage.type.numeric.rational.ruby
         3: storage.type.numeric.imaginary.ruby
-        4: storage.type.numeric.imaginary.ruby
-        5: invalid.illegal.numeric.rational.ruby
+        4: invalid.illegal.numeric.rational.ruby
     # numbers with binary base prefix
-    - match: '\b(0[bB]){{bdigits}}(r)?(?:(\.i)|(i)(r)?)?\b'
+    - match: '\b(0[bB]){{bdigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.binary.ruby
       captures:
         1: punctuation.definition.numeric.binary.ruby
         2: storage.type.numeric.rational.ruby
         3: storage.type.numeric.imaginary.ruby
-        4: storage.type.numeric.imaginary.ruby
-        5: invalid.illegal.numeric.rational.ruby
-    # complex imaginary numbers: 1i, 1.i, 1.1i, 1.1.i, 1e1i, 1e1.i, 1.1e1i, 1.1e1.i, 1ri, 1r.i, 1.1ri, 1.1r.i
-    - match: '\b{{ddigits}}(?:(\.){{ddigits}})?(?:(r)|{{exponent}}(r)?)?(?:(\.i)|(i)(r)?)\b'
+        4: invalid.illegal.numeric.rational.ruby
+    # complex imaginary numbers: 1i, 1.1i, 1e1i, 1.1e1i, 1ri, 1.1ri
+    - match: '\b{{ddigits}}(?:(\.){{ddigits}})?(?:(r)|{{exponent}}(r)?)?(i)(r)?\b'
       scope: constant.numeric.complex.imaginary.ruby
       captures:
         1: punctuation.separator.decimal.ruby
         2: storage.type.numeric.rational.ruby
         3: invalid.illegal.numeric.rational.ruby
         4: storage.type.numeric.imaginary.ruby
-        5: storage.type.numeric.imaginary.ruby
-        6: invalid.illegal.numeric.rational.ruby
+        5: invalid.illegal.numeric.rational.ruby
     # rational numbers: 1r, 1.1r
     - match: '\b{{ddigits}}(?:(\.){{ddigits}})?(r)'
       scope: constant.numeric.rational.ruby
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -203,14 +203,22 @@ contexts:
       scope: constant.language.ruby
     - match: '\b(__(FILE|LINE|ENCODING)__|self)\b(?![?!])'
       scope: variable.language.ruby
-    - match: '\b0[dD]\d(?:_?\d)*\b'
+    - match: '\b(0[dD])\d(?:_?\d)*\b'
       scope: constant.numeric.integer.decimal.ruby
-    - match: '\b0[xX]\h(?:_?\h)*\b'
+      captures:
+        1: punctuation.definition.numeric.decimal.ruby
+    - match: '\b(0[xX])\h(?:_?\h)*\b'
       scope: constant.numeric.integer.hexadecimal.ruby
-    - match: '\b0[oO]?[0-7](?:_?[0-7])*\b'
+      captures:
+        1: punctuation.definition.numeric.hexadecimal.ruby
+    - match: '\b(0[oO]?)[0-7](?:_?[0-7])*\b'
       scope: constant.numeric.integer.octal.ruby
-    - match: '\b0[bB][01](?:_?[01])*\b'
+      captures:
+        1: punctuation.definition.numeric.octal.ruby
+    - match: '\b(0[bB])[01](?:_?[01])*\b'
       scope: constant.numeric.integer.binary.ruby
+      captures:
+        1: punctuation.definition.numeric.binary.ruby
     - match: '\b\d(?:_?\d)*\.\d(?:_?\d)*r\b'
       scope: constant.numeric.float.rational.ruby
     - match: '\b\d(?:_?\d)*r\b'
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -247,7 +247,7 @@ contexts:
             # 1ri, 1.1ri, 1i, 1.1i | 1e1i, 1.1e1i
             (?: (\.) {{ddigits}} )? (?: (r) | {{exponent}}? )
             # 1.ri, 1.e1ri, 1.1e1ri | 1.i 1.e1i
-            |   (\.) (?: {{ddigits}}? {{exponent}}? (r) | ({{exponent}})? )
+            | (\.) (?: {{ddigits}}? {{exponent}}? (r) | ({{exponent}})? )
           )
         )(i)(r)?\b
       scope: constant.numeric.complex.imaginary.ruby
@@ -277,7 +277,7 @@ contexts:
         \b(?x:
           {{ddigits}} (?:
             # 1e1, 1.1e1
-            (?: (\.)     {{ddigits}}  )?  {{exponent}} \b
+            (?: (\.) {{ddigits}} )? {{exponent}}\b
             # 1., 1.1, 1.e1
             | (\.) (?: {{ddigits}}\b | ({{exponent}})\b )?
           )
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -203,8 +203,28 @@ contexts:
       scope: constant.language.ruby
     - match: '\b(__(FILE|LINE|ENCODING)__|self)\b(?![?!])'
       scope: variable.language.ruby
-    - match: '\b(0[xX]\h(_?\h)*|\d(_?\d)*(\.(?![^[:space:][:digit:]])(_?\d)*)?([eE][-+]?\d(_?\d)*)?|0[bB][01]+)\b'
-      scope: constant.numeric.ruby
+    - match: '\b0[dD]\d(_?\d)*\b'
+      scope: constant.numeric.integer.decimal.ruby
+    - match: '\b0[xX]\h(_?\h)*\b'
+      scope: constant.numeric.integer.hexadecimal.ruby
+    - match: '\b0[oO]?[0-7](_?[0-7])*\b'
+      scope: constant.numeric.integer.octal.ruby
+    - match: '\b0[bB][01](_?[01])*\b'
+      scope: constant.numeric.integer.binary.ruby
+    - match: '\b\d(_?\d)*\.\d(_?\d)*r\b'
+      scope: constant.numeric.float.rational.ruby
+    - match: '\b\d(_?\d)*r\b'
+      scope: constant.numeric.integer.rational.ruby
+    - match: '\b\d(_?\d)*(\.\d(_?\d)*)?r?i\b'
+      scope: constant.numeric.complex.imaginary.ruby
+    - match: '\b\d(_?\d)*(\.\d(_?\d)*)?ir\b'
+      scope: invalid.illegal.ruby
+    - match: '\b\d(_?\d)*(\.\d(_?\d)*)?[eE][-+]?\d(_?\d)*\b'
+      scope: constant.numeric.float.ruby
+    - match: '\b\d(_?\d)*\.(\d(_?\d)*\b)?'
+      scope: constant.numeric.float.ruby
+    - match: '\b\d(_?\d)*\b'
+      scope: constant.numeric.integer.ruby
     - match: ":'"
       scope: punctuation.definition.constant.ruby
       push:
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -208,63 +208,98 @@ contexts:
       scope: constant.language.ruby
     - match: '\b(__(FILE|LINE|ENCODING)__|self)\b(?![?!])'
       scope: variable.language.ruby
-    # numbers with decimal base prefix
-    - match: '\b(0[dD]){{ddigits}}(r)?(?:(i)(r)?)?\b'
-      scope: constant.numeric.integer.decimal.ruby
-      captures:
-        1: punctuation.definition.numeric.decimal.ruby
-        2: storage.type.numeric.rational.ruby
-        3: storage.type.numeric.imaginary.ruby
-        4: invalid.illegal.numeric.rational.ruby
-    # numbers with hexadecimal base prefix
-    - match: '\b(0[xX]){{hdigits}}(r)?(?:(i)(r)?)?\b'
-      scope: constant.numeric.integer.hexadecimal.ruby
+    # hexadecimal imaginary numbers: 0xAi, 0xAri
+    - match: '\b(0[xX]){{hdigits}}(r?i)(r)?\b'
+      scope: constant.numeric.imaginary.hexadecimal.ruby
       captures:
         1: punctuation.definition.numeric.hexadecimal.ruby
-        2: storage.type.numeric.rational.ruby
-        3: storage.type.numeric.imaginary.ruby
-        4: invalid.illegal.numeric.rational.ruby
-    # numbers with octal base prefix
-    - match: '\b(0[oO]?){{odigits}}(r)?(?:(i)(r)?)?\b'
-      scope: constant.numeric.integer.octal.ruby
+        2: storage.type.numeric.ruby
+        3: invalid.illegal.numeric.ruby
+    # octal imaginary numbers: 0o1i, 0o1ri, 01i, 01ri
+    - match: '\b(0[oO]?){{odigits}}(r?i)(r)?\b'
+      scope: constant.numeric.imaginary.octal.ruby
       captures:
         1: punctuation.definition.numeric.octal.ruby
-        2: storage.type.numeric.rational.ruby
-        3: storage.type.numeric.imaginary.ruby
-        4: invalid.illegal.numeric.rational.ruby
-    # numbers with binary base prefix
-    - match: '\b(0[bB]){{bdigits}}(r)?(?:(i)(r)?)?\b'
-      scope: constant.numeric.integer.binary.ruby
+        2: storage.type.numeric.ruby
+        3: invalid.illegal.numeric.ruby
+    # binary imaginary numbers: 0b1i, 0b1ri
+    - match: '\b(0[bB]){{bdigits}}(r?i)(r)?\b'
+      scope: constant.numeric.imaginary.binary.ruby
       captures:
         1: punctuation.definition.numeric.binary.ruby
-        2: storage.type.numeric.rational.ruby
-        3: storage.type.numeric.imaginary.ruby
-        4: invalid.illegal.numeric.rational.ruby
-    # complex imaginary numbers: 1i, 1.1i, 1e1i, 1.1e1i, 1ri, 1.1ri
-    - match: '\b{{ddigits}}(?:(\.){{ddigits}})?(?:(r)|{{exponent}}(r)?)?(i)(r)?\b'
-      scope: constant.numeric.complex.imaginary.ruby
+        2: storage.type.numeric.ruby
+        3: invalid.illegal.numeric.ruby
+    # decimal imaginary numbers: 0d1i, 0d1ri, 1i, 1ri, 1.1i, 1.1ri, 1e1i, 1.1e1i
+    - match: |-
+        \b(?x:
+          (?:
+            # 0d1i, 0d1ri, 1i, 1ri | 1.1i, 1.1ri
+            (?: (0[dD])? {{ddigits}} | {{ddigits}} (\.) {{ddigits}} ) (r?i)
+            # 1e1i, 1.1e1i
+            | {{ddigits}} (?: (\.) {{ddigits}} )? {{exponent}} (r)? (i)
+          ) (r)?
+        )\b
+      scope: constant.numeric.imaginary.decimal.ruby
       captures:
-        1: punctuation.separator.decimal.ruby
-        2: storage.type.numeric.rational.ruby
-        3: invalid.illegal.numeric.rational.ruby
-        4: storage.type.numeric.imaginary.ruby
-        5: invalid.illegal.numeric.rational.ruby
-    # rational numbers: 1r, 1.1r
-    - match: '\b{{ddigits}}(?:(\.){{ddigits}})?(r)'
-      scope: constant.numeric.rational.ruby
+        1: punctuation.definition.numeric.decimal.ruby
+        2: punctuation.separator.decimal.ruby
+        3: storage.type.numeric.ruby
+        4: punctuation.separator.decimal.ruby
+        5: invalid.illegal.numeric.ruby
+        6: storage.type.numeric.ruby
+        7: invalid.illegal.numeric.ruby
+    # hexadecimal rational numbers: 0xAr
+    - match: '\b(0[xX]){{hdigits}}(r)\b'
+      scope: constant.numeric.rational.hexadecimal.ruby
       captures:
-        1: punctuation.separator.decimal.ruby
-        2: storage.type.numeric.rational.ruby
-    # floating point numbers: 1.1, 1e1, 1.1e1
+        1: punctuation.definition.numeric.hexadecimal.ruby
+        2: storage.type.numeric.ruby
+    # octal rational numbers: 0o1r, 01r
+    - match: '\b(0[oO]?){{odigits}}(r)\b'
+      scope: constant.numeric.rational.octal.ruby
+      captures:
+        1: punctuation.definition.numeric.octal.ruby
+        2: storage.type.numeric.ruby
+    # binary rational numbers: 0b1r
+    - match: '\b(0[bB]){{bdigits}}(r)\b'
+      scope: constant.numeric.rational.binary.ruby
+      captures:
+        1: punctuation.definition.numeric.binary.ruby
+        2: storage.type.numeric.ruby
+    # decimal rational numbers: 0d1r, 1r, 1.1r
+    - match: '\b(?:(0[dD])?{{ddigits}}|{{ddigits}}(\.){{ddigits}})(r)\b'
+      scope: constant.numeric.rational.decimal.ruby
+      captures:
+        1: punctuation.definition.numeric.decimal.ruby
+        2: punctuation.separator.decimal.ruby
+        3: storage.type.numeric.ruby
+    # decimal floating point numbers: 1.1, 1e1, 1.1e1
     - match: '\b{{ddigits}}(?:(\.){{ddigits}}|(?:(\.){{ddigits}})?{{exponent}}(r)?)\b'
       scope: constant.numeric.float.decimal.ruby
       captures:
         1: punctuation.separator.decimal.ruby
         2: punctuation.separator.decimal.ruby
         3: invalid.illegal.numeric.rational.ruby
-    # integer numbers
-    - match: '\b{{ddigits}}\b'
+    # hexadecimal integer numbers: 0xA
+    - match: '\b(0[xX]){{hdigits}}\b'
+      scope: constant.numeric.integer.hexadecimal.ruby
+      captures:
+        1: punctuation.definition.numeric.hexadecimal.ruby
+    # octal integer numbers: 0o1, 01
+    - match: '\b(0[oO]?){{odigits}}\b'
+      scope: constant.numeric.integer.octal.ruby
+      captures:
+        1: punctuation.definition.numeric.octal.ruby
+    # binary integer numbers: 0b1
+    - match: '\b(0[bB]){{bdigits}}\b'
+      scope: constant.numeric.integer.binary.ruby
+      captures:
+        1: punctuation.definition.numeric.binary.ruby
+    # decimal integer numbers: 0d1, 1
+    - match: '\b(0[dD])?{{ddigits}}\b'
       scope: constant.numeric.integer.decimal.ruby
+      captures:
+        1: punctuation.definition.numeric.decimal.ruby
     # Quoted symbols
     - match: ":'"
       scope: punctuation.definition.constant.ruby
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -69,8 +69,11 @@ file_extensions:
 first_line_match: ^#!\s*/.*\bj?ruby\b
 scope: source.ruby
 variables:
-  digit: (?:\d(?:_?\d)*)
-  exponent: (?:[Ee][-+]?{{digit}})
+  bdigits: (?:[01]+(?:_[01]+)*)
+  ddigits: (?:\d+(?:_\d+)*)
+  hdigits: (?:\h+(?:_\h+)*)
+  odigits: (?:[0-7]+(?:_[0-7]+)*)
+  exponent: (?:[Ee][-+]?{{ddigits}})
   heredoc_type_css: (?:[[:upper:]_]_)?CSS
   heredoc_type_html: (?:[[:upper:]_]_)?HTML
   heredoc_type_js: (?:[[:upper:]_]_)?(?:JS|JAVASCRIPT)
@@ -206,7 +209,7 @@ contexts:
     - match: '\b(__(FILE|LINE|ENCODING)__|self)\b(?![?!])'
       scope: variable.language.ruby
     # numbers with decimal base prefix
-    - match: '\b(0[dD]){{digit}}(r)?(?:(i)(r)?)?\b'
+    - match: '\b(0[dD]){{ddigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.decimal.ruby
       captures:
         1: punctuation.definition.numeric.decimal.ruby
@@ -214,7 +217,7 @@ contexts:
         3: storage.type.numeric.imaginary.ruby
         4: invalid.illegal.numeric.rational.ruby
     # numbers with hexadecimal base prefix
-    - match: '\b(0[xX])\h(?:_?\h)*(r)?(?:(i)(r)?)?\b'
+    - match: '\b(0[xX]){{hdigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.hexadecimal.ruby
       captures:
         1: punctuation.definition.numeric.hexadecimal.ruby
@@ -222,7 +225,7 @@ contexts:
         3: storage.type.numeric.imaginary.ruby
         4: invalid.illegal.numeric.rational.ruby
     # numbers with octal base prefix
-    - match: '\b(0[oO]?)[0-7](?:_?[0-7])*(r)?(?:(i)(r)?)?\b'
+    - match: '\b(0[oO]?){{odigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.octal.ruby
       captures:
         1: punctuation.definition.numeric.octal.ruby
@@ -230,7 +233,7 @@ contexts:
         3: storage.type.numeric.imaginary.ruby
         4: invalid.illegal.numeric.rational.ruby
     # numbers with binary base prefix
-    - match: '\b(0[bB])[01](?:_?[01])*(r)?(?:(i)(r)?)?\b'
+    - match: '\b(0[bB]){{bdigits}}(r)?(?:(i)(r)?)?\b'
       scope: constant.numeric.integer.binary.ruby
       captures:
         1: punctuation.definition.numeric.binary.ruby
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by michaelblyons** [comment]
1. This is great.
2. Can regex groups which do not need special scoping become non-capturing groups? (Apparently this is slightly more performant.)

Up for discussion:

3. Would you mind adding `punctuation.definition.numeric.{binary | octal | hexadecimal}.ruby` to the indicative prefix (`0b`, etc.)?
4. Would you mind adding `storage.type.numeric[.imaginary].ruby` to the `i` and `r` suffices?
    - I do not have opinions about extending the scope an extra level for `r`.

**Comment by jwortmann** [comment]
@michaelblyons All done

**Comment by deathaxe** [comment]
My 2ct's here:

1. Looks like ruby 2.6 also likes `0xAr`, `0xAi` and `0xBri`, ... even `0.2e10i` 
  _(`0.2e10r` and `0.2e10ri` are illegal)_
  All binary, octal, hexadecimal values can be either imaginary or relational, too, even though not explicitely documented. The interpreter doesn't throw a syntax error.
2. To be very super perfect about punctuation ... the decimal dot could receive `punctuation.separator.decimal.perl`

**Comment by jwortmann** [comment]
> 1. Looks like ruby 2.6 also likes `0xAr`, `0xAi` and `0xBri`, ... even `0.2e10i`
>        _(`0.2e10r` and `0.2e10ri` are illegal)_
>        All binary, octal, hexadecimal values can be either imaginary or rational, too, even though not explicitely documented. The interpreter doesn't throw a syntax error.

Thanks, `1.i` seems to be valid too, but `1.r` and `1.ri` are not. Finding the best regex rules became a bit more complicated with all those, but I think I've covered all cases now. I adjusted the rules to inlcude `r` and `i` suffixes for hex, octal, etc. like `0xAAri` and `i` for numbers in exponential notation `0.2e10i` and for `1.i`:

* For numbers with prefix (`0x`, `0b`, etc.) I kept the `constant.numeric.integer.{decimal|hexadecimal|octal|binary}` scopes and just added `storage.type.numeric.{rational|imaginary}` for the suffix.
* `0.2e10i` and `1.i` are `constant.numeric.complex.imaginary` now, because I think it is more relevant than `constant.numeric.float` in that case. By the way, Go and R use `constant.numeric.imaginary` for this, while Pyhon uses just `constant.numeric.complex`, and the scope naming guidelines suggest `constant.numeric.complex.imaginary`, so I went with the latter.
* I simplified both `1.1r` and `1r` to `constant.numeric.rational` instead of `constant.numeric.{float|integer}.rational` because I think the float/integer distinction has not that much relevance here and would require an additional rule

>   2. To be very super perfect about punctuation ... the decimal dot could receive `punctuation.separator.decimal.perl`

Done

File: Ruby/Ruby.sublime-syntax [hunk 1]
```diff
@@ -203,8 +203,70 @@ contexts:
...
+      captures:
+        1: punctuation.separator.decimal.ruby
+        2: storage.type.numeric.rational.ruby
+        3: storage.type.numeric.imaginary.ruby
+    # 1.i
+    - match: '\b\d(?:_?\d)*(\.)(i)\b'
+      scope: constant.numeric.complex.imaginary.ruby
+      captures:
+        1: punctuation.separator.decimal.ruby
+        2: storage.type.numeric.imaginary.ruby
```
deathaxe: What do you think about merging the two rules for `# 1.1ri, 1.1i, 1ri, 1i` and `# 1.i` to:

```yaml
    # 1.1ri, 1.1i, 1.i, 1ri, 1i
    - match: '\b\d(?:_?\d)*(?:(\.)(?:\d(?:_?\d)*)?(r)?)?(i)\b'
      scope: constant.numeric.complex.imaginary.ruby
      captures:
        1: punctuation.separator.decimal.ruby
        2: storage.type.numeric.rational.ruby
        3: storage.type.numeric.imaginary.ruby
```

Should add some test cases for the `1.i` and `1.0i`, then:

```
1.i         #=> (0+1i)
#^^^ constant.numeric.complex.imaginary
#  ^ storage.type.numeric.imaginary
 1.0i        #=> (0+1i)
#^^^^ constant.numeric.complex.imaginary
#   ^ storage.type.numeric.imaginary
```

Ruby/Ruby.sublime-syntax
```diff
@@ -203,8 +203,70 @@ contexts:
...
+    # 1.1r, 1r
+    - match: '\b\d(?:_?\d)*(?:(\.)\d(?:_?\d)*)?(r)\b'
+      scope: constant.numeric.rational.ruby
+      captures:
+        1: punctuation.separator.decimal.ruby
+        2: storage.type.numeric.rational.ruby
```
deathaxe: I thought about scoping these numbers as well with exectly the considderations you described in your comment.

I am kind of uncertain about it, but I tend to scope rational numbers as ordinary `float` for 3 reasons:

1. you did not use `rational` in the oct/hex/dec/bin or imaginary numbers above
2. rational numbers are a sub group of real/float numbers
3. rational is not yet part of the scope naming guideline

Ruby/Ruby.sublime-syntax
```diff
@@ -203,8 +203,70 @@ contexts:
...
+      scope: constant.numeric.float.ruby
+      captures:
+        1: punctuation.separator.decimal.ruby
+    # 1.1, 1.
+    - match: '\b\d(?:_?\d)*(\.)(?:\d(?:_?\d)*\b)?'
+      scope: constant.numeric.float.ruby
+      captures:
+        1: punctuation.separator.decimal.ruby
+    - match: '\b\d(?:_?\d)*\b'
+      scope: constant.numeric.integer.ruby
```
deathaxe: If we decide to merge the `rational` numbers rule into float/integer, here is a suggestion to do so without introducing additional rules:

```yaml
    # 1.1r, 1r, 1.1, 1.
    - match: '\b\d(?:_?\d)*(\.)(?:(?:\d(?:_?\d)*)?(r)?\b)?'
      scope: constant.numeric.float.ruby
      captures:
        1: punctuation.separator.decimal.ruby
        2: storage.type.numeric.rational.ruby
    # 1, 1r, 1_1, 1_1r
    - match: '\b\d(?:_?\d)*(r)?\b'
      scope: constant.numeric.integer.ruby
      captures:
        1: storage.type.numeric.rational.ruby
```

Ruby/Ruby.sublime-syntax
jwortmann: The problem with that regex is that it would also match `1.ri` (which is illegal), while for this my current implementation only highlights the first part of it, but not the `ri`, so that users at least get a hint that this is actually a syntax error. If you prefer to not take that into account and favor the combined rule, I could change it though.
I will add the additional test cases.

Ruby/Ruby.sublime-syntax
jwortmann: I'm not sure if `float` is really the best choice for rational numbers, because the documentation explicitely highlights the difference between them:
> Rational numbers are exact, whereas Float numbers are inexact.
> 0.1r + 0.2r #=> (3/10)
> 0.1 + 0.2   #=> 0.30000000000000004

So rational numbers seem to be implemented as an own data type, just like complex numbers, and therefore I would prefer to give it a separate scope as well, to be consistent. To give both `12r` and `12.3r` the `constant.numeric.rational` scope, and not `constant.numeric.{integer|float}.rational` (dependant on the number part type) also corresponds with this.
> 12r         #=> (12/1)
> 12.3r       #=> (123/10)

For the numbers like `0xAAri` that fall in multiple categories, we have to choose the most relevant of them, because I think it wouldn't make sense to arbitrarily stack them into one large string (e.g. `constant.numeric.integer.hexadecimal.rational.imaginary`). This feels a bit similar to the recent discussion about `keyword.operator.logical` / `keyword.operator.word`. I think for numbers with base prefix the `hexadecimal`/`binary` part was the most important, for other notations including `i` the `complex.imaginary` property, and after that the `rational` property.

> 3. rational is not yet part of the scope naming guideline

I'm not sure which other languages have datatypes for rational numbers, but the only one from the default syntaxes I know of is Clojure, which uses `constant.numeric.ratio.clojure` for e.g. `2/3`, but I prefer `rational` more than `ratio` here. And I know that there are rational numbers `2//3` in Julia, but its syntax package doesn't even detect them. Perhaps it could be added to the guidelines?

Ruby/Ruby.sublime-syntax
deathaxe: Didn't realize that. We should of course not break highlighting due to some merges. I just came around with that with Perl's numbers in mind, which I also adapted to Python in #2136. I was able to tweak performance by about 14% by thouroghly merging regexp patterns.

Note:

I didn't use variables in Perl, but found them quite useful in manner of readability, while working on #2136.

Based on those, we could create one single match rule for all imaginary numbers including some invalid handling


```yml
variables:
  digit: (?:\d(?:_?\d)*)
  exponent: (?:[Ee][-+]?{{digit}})

  # ...

contexts:

  # ...

    # complex imaginary numbers
    - match: |-
        \b(?x:
            {{digit}} (?:
                # 1ri, 1.1ri, 1i, 1.1i | 1e1i, 1.1e1i
              (?: (\.) {{digit}} )? (?: (r) | {{exponent}}? )
              | # 1.ri, 1.1ri, 1.e1ri, 1.1e1ri | 1.i 1.e1i
              (\.) (?: {{digit}}? {{exponent}}? (r) | ({{exponent}})? )
            )
        )(i)\b
      scope: constant.numeric.complex.imaginary.ruby
      captures:
        1: punctuation.separator.decimal.ruby
        2: storage.type.numeric.rational.ruby
        3: punctuation.separator.decimal.ruby
        4: invalid.illegal.numeric.rational.ruby
        5: invalid.illegal.numeric.exponent.ruby
        6: storage.type.numeric.imaginary.ruby
```

_Note: I am about to prepare a complete proposal with some test cases for you._

Ruby/Ruby.sublime-syntax
deathaxe: > To give both 12r and 12.3r the constant.numeric.rational scope

This is the compelling argument to use the `rational` scope here. None of the numbers is an integer, indeed.

> ...  I think it wouldn't make sense to arbitrarily stack them into one large string ...

Fully agree with you. This kind of long scope names doesn't make much sense as it is unable to be queried by selectors efficiently.

The issue with `complex.{real|imaginary}` and `int,float,rational,...` is just that these are two different kinds of perspective. We don't currently handle complex numbers completely in manner of scope naming guidelines.

To do so we would need to scope both the `real` and the `imaginary` part, while both of them can consist of `float` or `rational` numbers.

We would need to implement some thing like :

```
 12.0+10.4i
#^^^^ constant.numeric.complex.real constant.numeric.float
# ^ punctuation.separator.decimal
#    ^^^^^^ constant.numeric.complex.imaginary constant.numeric.float
#    ^ punctuation.separator.complex
#       ^ punctuation.separator.decimal
```

But this is nothing anyone considdered so far.

deathaxe:
Got some inspirations for you at https://github.com/jwortmann/Packages/pull/1

Ruby/Ruby.sublime-syntax
jwortmann: And to add, in my opinion none of it should be `float` either, because from a mathematical point of view floating point numbers are a strict subset of rational numbers. The Ruby implementation seems to be a bit confusing though, because when tested for equality, the rational numbers apparently are mapped to floats (sorry if I went a bit too far with all these long numbers for this discussion :stuck_out_tongue_winking_eye:):
```ruby
r = 0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001r
f = 0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001
puts r
puts f
puts r == f
puts 10*r == 10*f
```
output (from https://repl.it, ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-linux]):
```ruby
1/1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
0.0
true
false
```
But this actually applies more to the previously mentioned languages, which have rational numbers with arbitrary denominator like `1/3`, that itself can't be mapped to floating point numbers.

I agree with your considerations about complex numbers, but I don't think it would bring much benefit. It probably would make syntax definitions for numbers unnecessarily complicated and I don't think that any color scheme targets such specific scopes anyways, or that it would result in different highlighting. And it could be problematic to parse and to justify in what cases the `real` part and the `+` should be considered as its own number/operator, or to be part of the complex number.

deathaxe: > ... but I don't think it would bring much benefit ... would make syntax definitions for numbers unnecessarily complicated

That's true. The question the otherway round than is: If we don't target `complex.real`, does it make sense to scope the imaginary part as `complex.imaginary` then? 

I just wonder about it, with a look into the other syntaxes, too.

Would it make sense to just scope imaginary numbers as normal `float|rational` with just the `i` scoped as `storate.type.imaginary` then?

jwortmann: Well, I would say in cases where the language has a builtin datatype for complex numbers, `complex.imaginary` makes sense, and I like how the third-level part of the scope `{integer|float|complex|rational}` is always an indicator of the used datatype then. However, my proposal for rational/imaginary numbers with additional base prefix like `0xAAri` to be just `constant.numeric.integer.hexadecimal` isn't really accurate from that point of view, but I don't think there exists a perfect solution for this, unless we decide for repeatings like `constant.numeric.rational.ruby constant.numeric.complex.imaginary.ruby`. But I would rather avoid that and keep it simple, even if it isn't 100% accurate.

Ruby/Ruby.sublime-syntax
```diff
@@ -272,11 +275,11 @@ contexts:
     # floating point numbers
     - match: |-
         \b(?x:
-          {{digit}} (?:
+          {{ddigits}} (?:
             # 1e1, 1.1e1
-            (?: (\.)     {{digit}}  )?  {{exponent}} \b
+            (?: (\.)     {{ddigits}}  )?  {{exponent}} \b
```
deathaxe: Ouff. Just see, I missed to remove the whitespace as in the pattern above. One space as separator should be enough.

jwortmann: Thanks, fixed.

jwortmann:
I just saw that in Ruby even numbers are objects with methods and for example the following is legal:
```ruby
5.inverse     # => 0.2
```
This is pretty different from what I'm used to in other languages, and I believe the current scoping for `1.` as `float` is wrong in this PR and the dot should not be part of the number, but instead really be `punctuation.accessor` in that case.

deathaxe:
Good catch.

The `print(1.e1)` fires 

```
`<main>': undefined method `e1' for 1:Integer (NoMethodError)
```

Strictly following this error message, your suggestion about not to mark them illegal seems reasonable. Github doesn't do so as well. So we should follow.

Therefore all cases belong to integers followed by accessor

jwortmann:
I found that `.i` is always equivalent to the `i` suffix, so that e.g. `1.1.i` and `1.1r.i` are valid imaginary numbers as well. I rewrote the regex patterns to match these notations too and to avoid marking parts of the number as illegal, unless it is inevitably a syntax error (which is: an exponent followed by `r`, and `i` followed by `r` - unless the `i` is preceded by a dot). I would say the dot in the `.i` suffix is neither a decimal separator nor accessor in that case, so I included it in the `storage.type.numeric.imaginary` scope. The test file hopefully covers all possible notations now, and should be ready to review. Please comment if there are any opinions on the used scope names or if there are any other possible improvements.

FichteFoll:
`.i` looks like a method call to me and this would confirm it:
```
irb(main):002:0> 3ri
=> (0+(3/1)*i)
irb(main):003:0> a = 3r
=> (3/1)
irb(main):004:0> a.i
=> (0+(3/1)*i)
```

jwortmann:
Yes, I thought that too. Should I exclude it then from the number? In that case for `1.i` only the `1` would be highlighted, and only as `integer` and not `imaginary`.

FichteFoll:
From a syntax perspective, yes, because the literal is a real number and only the method turns it into an imaginary number. Let's hear what @deathaxe thinks about it, seeing how much he dealt with number highlighting recently.

jwortmann:
Alright, I agree with that.

Ruby/Ruby.sublime-syntax
```diff
@@ -209,84 +209,65 @@ contexts:
...
       scope: constant.numeric.float.ruby
       captures:
         1: punctuation.separator.decimal.ruby
-        2: invalid.illegal.numeric.exponent.ruby
-    # integer decimal
+        2: punctuation.separator.decimal.ruby
+        3: invalid.illegal.numeric.rational.ruby
+    # integer numbers
     - match: '\b{{ddigits}}\b'
       scope: constant.numeric.integer.ruby
```
deathaxe: As we already define `constant.numeric.integer.binary` etc. we should propaply use `constant.numeric.integer.decimal` here to be complete. We started to do so in other syntaxes as well in case someone wants to address them.

Ruby/Ruby.sublime-syntax
```diff
@@ -209,84 +209,65 @@ contexts:
...
-            # 1.1, 1.1e1, 1.e1
-            (\.) (?: {{ddigits}} {{exponent}}? | ({{exponent}}) )
-            # 1e1
-            | {{exponent}}
-          )
-        )\b
+        2: storage.type.numeric.rational.ruby
+    # floating point numbers: 1.1, 1e1, 1.1e1
+    - match: '\b{{ddigits}}(?:(\.){{ddigits}}|(?:(\.){{ddigits}})?{{exponent}}(r)?)\b'
       scope: constant.numeric.float.ruby
```
deathaxe: In some syntaxes like R it is possible to have `constant.numeric.float.binary`, ... . Therefore we started adding `.decimal` to normal floats as well: `constant.numeric.float.decimal.ruby`. 

Not sure whether we really need it, but it would cause decimal floats to have the same scope in all languages, if we make it to apply it everywhere.

Ruby/Ruby.sublime-syntax
jwortmann: I used `decimal` only for the `0d` prefix yet, because I had not seen it often in other syntaxes and a number is always decimal unless explicitly stated different. But no problem to update this scope.

deathaxe: Just had selectors in mind. If someone wants to target decimal only, he would need to use `constant.numeric.integer - constant.numeric.integer.binary - constant.numeric.integer.octal - ...` atm, which might be a bit tedious.

Ruby/Ruby.sublime-syntax
jwortmann: Ok, does it makes sense to update the scope for `0xAAr` to be `constant.numeric.rational.hexadecimal` too then? But as discussed previously, this would require additional rules and I still can't think of an optimal and consistent scope for `0xAAri`.

deathaxe: If we consider `rational` as a nother data type next to integer and float,  `constant.numeric.rational.hexadecimal` would definitely make sense in manner of consistency.

I am a bit uncertain about imaginary numbers as this is just another perspective of how to scope numbers, but Go just ignores them. It just handles everything as integer/float/... with the `i` scoped as `storage.type`.

As we concluted scoping complex numbers to result in too complex patterns the Go approach might even be just enough.

But I don't feel like the one to decide it.

deathaxe:
> `.i` looks like a method call to me and this would confirm it:

All the extensions are method calls, even the `i` in `3i` is. 

Most keywords like `new`, `include`, etc. seem to be methods only as well.

Therefore a decision whether to scope the special methods `.i` or `.r` as `storage` or `variable.function` is not so clear here. But I prefer `storage.type` for `i` in `3i`. 

I found `.i` to be scoped as `storage` a bit strange in the first place due to the dot being included.

I don't have a strong oppinion about it, but with the dot scoped as accessor the scoping the special method `i` as storage for consistency reasons, might be something to think about, with the special scopes for some keyword like methods in mind.

It would at least make it clear for a user that `3i` and `3.i` are the same.

Ruby/Ruby.sublime-syntax
wbond: I'm not heavily versed in number theory, but from doing a little reading, it seems that for complex numbers we are talking about numbers in the form, `a+bi` where `a` and `b` are real numbers and `i` is imaginary.

From this, I think the scoping should be:

```
 12.0+10.4i
#^^^^^^^^^^ constant.numeric.complex
#^^^^ constant.numeric.complex.real
#  ^ punctuation.separator.decimal
#     ^^^^ constant.numeric.complex.real
#    ^ punctuation.separator.complex
#       ^ punctuation.separator.decimal
#.        ^ constant.numeric.complex.imaginary
```

This way the entire thing is `constant.numeric.complex`. I'm not sure there is much value in trying to get super-specific beyond this level. The primary consumer of these scopes is color schemes, and the key thing we want to highlight here is that the `+` is not an addition sign, but part of making the whole thing an imaginary number.

Consistency is good, and I've got some work to do in terms of codifying various scope names we've been standardizing on into the scope naming guidelines, but I'd rather not let perfect get in the way of getting these changes in the hands of users.

wbond:
I think we should move forward with treating `3i` having the entire thing be a `constant.numeric.imaginary` or `constant.numeric.*.imaginary`. The `i` isn't really a storage hint, but rather it changes the fundamental value. `1f` and `1` effectively define the same number. My understanding is that `1i` is fundamentally the product of `1 * I`.

In the case of `3.i` that should be scoped as a method call, since that is what it is. I'm not super concerned that Ruby may end up implementing `3i` as a method call, as it is really an atomic literal and you can't monkey patch that behavior. You could override `.i`, so I think we should make it clear it is a method call.

Ruby/Ruby.sublime-syntax
jwortmann: In `a+bi`, `a` is the real part and `b` is the imaginary part, while `i` (or sometimes `j`) is just a label for the imaginary unit. This is the important part here, so the scoping for `a` and `b` certainly has to be
```
 12.0+10.4i
#^^^^ constant.numeric.complex.real
#     ^^^^ constant.numeric.complex.imaginary
```
I'm not sure if there is much value in scoping the first part and the `+` sign at all, because a single `10.4i` will automatically parsed into `(0+10.4i)`, and the problem is to decide how to scope something like
```
12.0 +          10.4i
```
then. I would say this is the addition of a real and an imaginary number, and the `+` should be `keyword.operator`, but conventions might vary between languages.

deathaxe: The `+` here is part of the imaginary part as it describes the direction of the vector. Complex numbers describe a vector in an coordinate system where the real part is the X and the imaginary part the Y (in an X,Y coordinate system).

FichteFoll: The most important part about this is operator precedence, because the plus operator (or separator, in a semantical sense) is subject to that and syntax definitions will struggle with something like `2*2+1i`.

wbond: Okay, I was going off of https://en.wikipedia.org/wiki/Imaginary_number

> An imaginary number is a complex number that can be written as a real number multiplied by the imaginary unit *I*

However, after considering my separate comments about `3i` below, I'm thinking the whole thing `10.4i` or `12.0+10.4i` needs to be marked as imaginary. Based on the definition from Wikipedia (which may be imprecise for brevity), it seems an imaginary number is always a complex number. Based on this, I'm thinking as soon as `i` is present we need to mark it as imaginary, whether the form is "complex" or `3ri`, etc.

deathaxe:
So remove the `.complex.` part of the scope?

Ruby/Ruby.sublime-syntax
wbond: In terms of whitespace between the first part and the separator, I would only scope it all as imaginary if there is no whitespace. But I'm not a user of imaginary numbers, so my preference should be pretty low in deciding what to do.

wbond:
Honestly, based on Wikipedia, an imaginary number is a complex number and a complex number is in the form `a+bi` where `i` is an imaginary number…

wbond:
I think I'm probably out of my depth here on the subject matter, but my vote would be for the solution that is simplest, while treating the imaginary number "suffix" (`i` here, or `j` in Python I believe) as part of the number, and not a `storage.type.numeric`.

deathaxe:
Yes, imaginary numbers are complex numbers with a real part 0: One could write `i = 0+1i`. 

The most complicated part here is the combination of the `complex.real + complex.imaginary` perspective an the fact that both parts can be floating point or rational numbres.

In R those parts can even be octal, hexadecimal or binary numbers.

In theory each ordinary number is also a complex number with an imaginary part 0: `12 = 12+0i`. 

As we see: We live in a complex world. hehe.

FichteFoll:
(Moving this down here because that's where the discussion has been happening.)

Indeed, complex numbers have a real and an imaginary part and every imaginary number (literal) is a complex number (and so is every real number). The `constant.numeric.complex.imaginary` scope reflects this.

When matching the construct `1+2i` as a whole and assigning the complex scope, which I don't disagree, we end up with a not so trivial problem for syntax definitions because of my example earlier. Compare these expressions in Python and Ruby respectively:

```
In [1]: 2*2+2j
Out[1]: (4+2j)
In [2]: 2*(2+2j)
Out[2]: (4+4j)
```
```
irb(main):001:0> 2*2+2i
=> (4+2i)
irb(main):002:0> 2*(2+2i)
=> (4+4i)
```

What I'm trying to say with this is: 

1. Either every number is a complex number, including floats and integers, which are real/rational and whole numbers respectively. We can then dispose of the `complex` part and only highlight the `i` in some special way (if `storage.type` is not ideal, then I can only think of `constant.language.i`).
2. Or we specifically try to highlight the entire construct `1+2i` where possible as `*.complex`, which isn't trivial. Just imaginary numbers (`2i`) as well, of course. However, this clashes with the binary/decimal/hexadecimal scope part, which would normally appear at the same level.

jwortmann:
The operator precedence in `2*2+2i` is a very good point and it supports me in my opinion to not target the real part and the `+` symbol for complex numbers. I agree that the `complex` part wouldn't be really necessary then and could be omitted.

However, there might be other ways to define a complex number, differently from the `a+bi` notation, where eventually the `complex.real` and `complex.imaginary` distinction could make sense for semantically highlighting. For example, in Fortran a complex variable with value 1.0+2.0i is defined as follows:
```fortran
complex :: myvar = (1.0, 2.0)
```
I know this example is probably of minor relevance, and one could argue to just use `float` for both real and imaginary part in that case. But in general, the `complex` would reflect that complex numbers are implemented as its own datatype in those programming languages which have them. Ruby seems to be a special case, that it has both rational and complex numbers which can be combined (the same is true e.g. for the Julia language).

I have no strong opinion if `constant.numeric.complex.imaginary` or `constant.numeric.imaginary` or even just the usual `constant.numeric.{integer|float}` is the best decision for imaginary numbers. But I would suggest to not drop the special scoping for the `i` (whether be it `storage.type.numeric` or `constant.language`), because then it could easily be missed if not highlighted, and because it is already used by some languages.

deathaxe:
We aggree with scoping the whole complex number to be too complex compared to the little value.
As a result I'd vote to remove the `complex` part.

As all imaginary numbers are float/rational numbers per definition, the `.imaginary` scope could be used as a special form of a floating point number.

The current implementation of R reflects this circumstance very well already:

0b0101010i -> `constant.numeric.imaginary.binary`
0o017 -> `constant.numeric.imaginary.octal`
0d213i -> `constant.numeric.imaginary.decimal`
0x23ai - `constant.numeric.imaginary.hexadecimal`

Therefore scoping an number like `0d23ri` as `constant.numeric.integer` (Ruby) seems even to be wrong. Propably `constant.numeric.imaginary.decimal` is the better choice then.

This implies to add `rational` and `imaginary` to the list of child scopes of `constant.numeric`, which would then be an argument to keep `storage.type` for `i` :-)

---

I also like the special scopes for `i` / `j`. Using a different scope than `storage.type` might result in the `r` and `i` in `3.0ri` to be scoped in different colors. May be too much in the end? From the highlighting perspective I'd expect both `ri` to be colored the same way. If `r` is scoped different only. 

On the other hand `i` is just another suffix of a numeric constant to narrow down its meaning, just like the `r` or other suffixes like `f`, `L` in C or R. Furthermore `storage.type` is used for such suffixes in at least Go already.

wbond:
It seems that we've generally come to the conclusion that we won't explicitly try to scope the form `a+bi` entirely as an atomic unit. I think this makes sense, and semantically it is alright due to mathematical properties.

From here on I think we are more or less explicitly talking about how to scope these two general forms of tokens:

 - `12.0i`
 - `5ri`

----

> How about `meta.numeric.complex.imaginary constant.numeric.{integer|float}` for imaginary numbers

In this situations I'd ask what the purpose of the meta scope is for? I'm all for using accurate scopes where possible, but stacking meta scopes when there isn't a concrete use case at hand seems like a waste, and generally something I'd like to avoid codifying into the scope naming guidelines.

> One other possibility to consider for the `I` - if `storage.type` isn't appropriate - would be `keyword.other.unit`, in consistency with the CSS syntax

I don't think `i` is really controlling the storage type of the number. It isn't causing the number to be store in a 64bit memory location instead of 32bit, but fundamentally affecting the value. Because of this I think `storage.type` is wrong. I am fine with it being part of `constant.numeric` or possibly something like `keyword.other.unit`.

For example:

```
irb(main):001:0> 1 == 1r
=> true
irb(main):002:0> 1 == 1i
=> false
irb(main):003:0> 1ri
=> (0+(1/1)*i)
```

`r` is a `storage.type`, whereas `i` is not, it is effectively a constant.

> Therefore scoping an number like `0d23ri` as `constant.numeric.integer` (Ruby) seems even to be wrong. Propably `constant.numeric.imaginary.decimal` is the better choice then.

I think using `constant.numeric.imaginary.*` sound reasonable to me, from all of the details and arguments I've heard. I do think that `r` should be `storage.type` (as shown above) since it affecting how the value is stored, but not fundamentally what the value is.

jwortmann:
> In this situations I'd ask what the purpose of the meta scope is for?

The advantage of
```
meta.numeric.imaginary constant.numeric.{integer|float|rational}.*
```

(I deleted the `complex` and added the `rational` option here) would be, that it still contains the data type of the number, just as in real numbers.
```
irb(main):001:0> 1i
=> (0+1i)
irb(main):002:0> 1.0i
=> (0+1.0i)
irb(main):003:0> 1ri
=> (0+(1/1)*i)
irb(main):004:0>
```

If we scope all of `1i`, `1.0i`, `1ri` with `constant.numeric.imaginary.*`, this information is lost. From a highlighting perspective, I don't think there would be any difference, so I'm fine with `constant.numeric.imaginary.*` as well.

> I don't think i is really controlling the storage type of the number.

Well, yes but actually no. If the compiler or interpreter sees a `1i` while parsing the source code, I believe it will allocate space for two numbers in memory, the real and the imaginary part (even if the real part is 0 in this case). This is different from just a single `1`, which can be stored as a single 32bit/64bit integer. In Ruby, both real and imaginary part can furthermore be stored as an integer, float or rational number.

jwortmann:
> As all imaginary numbers are float/rational numbers per definition, the .imaginary scope could be used as a special form of a floating point number.

This statement is false. While I know that e.g. Fortran will automatically cast the real and imaginary parts of complex numbers to floating point numbers when declared with integers, as shown above in Ruby imaginary numbers can either be `integer`, `float` or `rational`. I don't know how other languages behave, but the Ruby example disproves the assumption that imaginary numbers are always stored as floating point numbers.

Even rational numbers are no floating point numbers, for example 1/3 cannot be represented as a floating point number.

As soon as we find an agreement about how to scope imaginary numbers and the `i`, I could update the rules in this PR for e.g. `0d23ri`.

deathaxe:
> As all imaginary numbers are float/rational numbers per definition, the .imaginary scope could be used as a special form of a floating point number.

This statement was made from the pure mathematical perspective and the number type hierarchie in mind which is outlined at [Python PEP 3142 > Abstract](https://www.python.org/dev/peps/pep-3141/#id3).

When talking about `float` here I just mean [Wikipedia \> real numbers](https://en.wikipedia.org/wiki/Real_number), but with trying to avoid the naming confilict with `complex.real`. According to the wiki page `float` is an implementation detail of `real numbers` and `rational numbers` to be a sub set of real. 

Therefore `3i` is not an integer. It is just a short form of a real/float number - ignoring any (propably unimportant) implementation detail (don't care about it to be stored as double, float, int64, int32, short, byte or bitmask ...) **... just in order to find a common way how to scope in in all languages**.

The point just is: 

In order to resolve the conflict of typing perspectives either use `constant.numeric.imaginary.[binary|hexadecimal|octal|decimal]` if we are interested in imaginary numbers or ignore `imaginary` at all and just use `constant.numeric.[integer|float|rational].[binary|hexadecimal|octal|decimal]`.

jwortmann:
This sounds very confusing to me. From "the pure mathematical perspective" as you call it, floating point numbers ⊂ rational numbers ⊂ real numbers. Please don't confuse floating point numbers with real numbers. When I talk about `float`, I mean [floating point numbers](https://en.wikipedia.org/wiki/Floating-point_arithmetic#Floating-point_numbers) and not [real numbers](https://en.wikipedia.org/wiki/Real_number), because arbitrary real numbers cannot be stored by computers. When I talk about `complex.real`, I mean the real part of a complex number. I don't care whether Python has an **abstract** class named "Real", which is obviously not the same as its builtin `float` data type. By the way where do you even read from this link, that imaginary numbers are per definition floating point numbers (which might even be true for Python)? If somehow you treat imaginary numbers the same as Python's "Complex" (which is wrong by itself), then the only thing you can see from `Number :> Complex :> Real :> Rational :> Integral` is, that they are a subtype of "Number".

> Therefore 3i is not an integer. It is just a short form of a real/float number

I strongly disagree with that. Could you please explain to me then, why the Ruby interpreter prints
```
irb(main):001:0> 3i
=> (0+3i)
```
for `3i`, in contrast to `(0+3.0i)` for `3.0i` and `(0+(3/1)*i)` for `3ri`? To me the `(0+3i)` clearly looks like a complex number, in which the imaginary part is stored as an integer.

> ignoring any (propably unimportant) implementation detail (don't care about it to be stored as double, float, int64, int32, short, byte or bitmask ...)

Well, I thought this is what the third-level part of the `constant.numeric.*` scope is used for: to specify how a number is stored ("data type"), i.e. floating point number (`float`), integer number (`integer`) or rational number (`rational`, `ratio` in Clojure). Complex numbers are indeed a bit difficult to classify here, because on the one hand they can be considered as its own data type, and on the other hand in e.g. Ruby the number type of the real/imaginary part appearantly isn't fixed, but can be one of `float`/`integer`/`rational`. So If you want to ignore this implementation detail for imaginary numbers, one could ask why not to be consistent and omit it for non-imaginary numbers too? Don't get me wrong, I think it could be a valid option to only ignore it for imaginary numbers because of the following reasons:
* the third-level part could be used for `imaginary` then to indicate that it is an imaginary number, while still leaving the fourth-level part for the base (`decimal`, `hexadecimal`, ...)
* Ruby seems to be an exception that it permits more than just one possible data type for imaginary numbers. As mentioned before, Julia is another example
* there is (most likely) no difference in highlighting anyways

I think we have 3 options here:
1. `constant.numeric.imaginary.{binary|hexadecimal|octal|decimal}`
2. something like `meta.imaginary constant.numeric.{integer|float|rational}.{binary|hexadecimal|octal|decimal}`
3. `constant.numeric.{integer|float|rational}.{binary|hexadecimal|octal|decimal}`

I would vote for (1) or (2). Case (1) is more simple, while (2) would be more accurate and consistent with non-imaginary numbers.

deathaxe:
My "confusing arguments" are to propose for your option (1).

wbond:
It sounds like we can move forward with your option 1 @jwortmann. The only bit left is what to scope the `i`.

jwortmann:
Alright. I think the given options for `i` were:
1. none
2. `storage.type.numeric.imaginary`
3. `constant.language.i`
4. `keyword.other.unit`

I vote for (2), because it gives the best highlighting colors and `storage.type` doesn't really feel inappropriate to me here. I think `keyword.other.unit` would be a bit more accurate semantically, but I don't like the highlighting colors for (1), (3) and (4).

jwortmann:
By the way, here is an irrefutable proof, that `1i` is stored as `integer`, and not `float` in Ruby (in case there still is someone out there, who believes in the opposite):
```
irb(main):001:0> 1i.imaginary.class
=> Integer
irb(main):002:0> 1.0i.imaginary.class
=> Float
irb(main):003:0> 1ri.imaginary.class
=> Rational
```

The `.imaginary` method returns the imaginary part of a complex number.

deathaxe:
> here is an irrefutable proof, that `1i` is stored as `integer`

This has never been the question. Of course a good implementation chooses the smallest / most efficient way to store information. ;-)

> I vote for (2a) or (2b), because it gives the best highlighting colors ...

I am with @jwortmann here. I tend to prefere (2a) for the same reasons as the next point below.

> ... I'm unsure if I prefer `imaginary` or `complex` ...

I think `imaginary` would be more accurate as the `i` actually denotes the imaginary part of a complex number. 

Of course `1i` is the short form of `0+1i` and therefore a complex number, too, but as we chose to not scope the whole `a+bi` token, we should propably not introduce the `complex` here?

FichteFoll:
> I vote for (2a) or (2b), because it gives the best highlighting colors and storage.type doesn't really feel inappropriate to me here.

Please don't try to argue scope names with colors. These are scheme-dependant and it makes no sense to talk about them in isolation without giving an example for why a certain combination is bad.

Instead, you could argue that choosing `constant.language` alongside `constant.numeric` is a poor choice because color schemes are likely to only target `constant` as a scope (instead of additionally providing a separate color for `constant.numeric` or `constant.language`, despite both being in the [recommendations for minimal coverage](https://www.sublimetext.com/docs/3/scope_naming.html)) and the two tokens would be directly next to each other.

Other than that, I am also in favor of using `storage.type` scope because it does affect the type of the literal. It does have a different usage than a type name during a variable declaration, so I could see an argument for maybe using a different `storage` sub-scope, but none of the ones currently in use (type and modifier) seem to fit and I don't have a particular suggestion myself. It should be the same scope as used in `1L` or `b'bytes'`, however.

deathaxe:
> makes no sense to talk about them in isolation without giving an example for why a certain combination is bad

The argument why scoping the `i` differently than any other suffix like `r`, `L`, etc. would be sub-optimal was actually discussed with exectly your arguments about `1L` or `b'bytes'`.

Scoping `1ri` should not result in `r` and `i` being colored differently by existing or minimal color schemes. Therefore the first 2 levels of scope naming should be the same.

Furthermore I think complex/imaginary are indeed a separate group of data types, even though on a higher level. Python/Ruby create dedicated classes for them, R knows it as dedicated language type. It's crucial to know about the data type as different vector-based maths rules need to be applied to them.

jwortmann:
> This has never been the question. Of course a good implementation chooses the smallest / most efficient way to store information. ;-)

Okay. Sorry, I didn't want to sound rude in my previous comment.

> Please don't try to argue scope names with colors

Yes, seems like I wasn't specific enough here and "highlighting colors" was poor phrasing.
What I really meant is (as discussed earlier):
For (1) (no scope) and perhaps `constant.language`, there is no distinction of the `i` suffix from the rest of the number, so it could easily be mistaken e.g. as a `1` and be unnoticed. For `1ri` it will change from the `constant.numeric` color for `1` to another color for `r` and then back to the first color. For option (4) `keyword.other.unit` (and perhaps `constant.language`) the highlighting color is probably different from `storage.type`, so that the `1ri` combination will result in 3 different colors (and perhaps font styles). If there is an additional prefix, it might even be 4 different colors in some schemes for a single number.
Monokai
(Monokai)

That just looks unpleasant to me with the changing font styles and colors, because I would expect the suffix to have a unique color, which would need both `r` and `i` to be `storage.type` here. This is of course highly subjective, but I could imagine that many users would feel the same about those highlighting styles.

> It does have a different usage than a type name during a variable declaration, so I could see an argument for maybe using a different storage sub-scope

I agree, but another new `storage` scope might also be untargeted by color schemes, because the scope coverage recommendations include only `storage.type` and `storage.modifier` but not `storage`, which then would have the same result as in option (1).

<hr>

I think the `imaginary` vs. `complex` decision depends on if we want to focus what the `i` means (imaginary unit), or what its effect is (turns a number to be stored as "Complex" type).
So from the pure meaning perspective, `imaginary` would be better, but in combination with `storage.type` the `complex` might be more appropriate, because there is no real "imaginary" data type (I think). Is this different in R? If so, I wonder how multiplication of two imaginary numbers is implemented in R, because it would have to change the data type then.

deathaxe:
When I first saw the `complex` scope my initial reflex was to think to need to scope the whole complex number `a+bi`. As we came around with not doing so in favor of simpler matching rules, I'd prefer to use `imaginary` to just avoid contributors to run into the same situation in the future.

If we scope `bi` as imaginary, it is accurate. If we scope it as `complex` someone might think about `a` to be as well.

I don't know about the details in R, but it just scopes all imaginary numbers as `constant.numeric.imaginary.[binary|hexadecimal|octal|decimal]`. Found it quite appropriate as it doesn't address the real part of the complex numbers as well.

jwortmann:
> If we scope bi as imaginary, it is accurate. If we scope it as complex someone might think about a to be as well.

This is not the question here, as we already decided for `constant.numeric.imaginary` for it. I just wanted to point out that the additional scope for the `i` would be more accurate if we use `storage.type.numeric.complex`, because there is no such thing as a type for imaginary numbers.
```ruby
 123i
#^^^^ constant.numeric.imaginary.decimal
#   ^ storage.type.complex
```
If we prefer `imaginary`, then we actually would have to use `keyword.other.unit.imaginary`, to be precise.

Without ever having used R or knowing its implementation details, I cannot imagine that there is a dedicated "imaginary" type, because e.g. how to store an array of complex numbers then? That must automatically be an array of arrays, which seems inefficient.

deathaxe:
Ok, was mislead then. I have no strong meaning about the scope of  `i`. I am basicly fine with both `storage.type.numeric.[complex|imaginary]`. We can use `complex`, too.

wbond:
I would say scope the `r`, `i` or `ri` as `storage.type.numeric` if we specialize then the `r` and the `i` will be different tokens, which would be odd.

deathaxe:
... which would then basically mean to scope all such suffixes in all syntaxes like `40L` or `10.0f` as `storage.type.numeric` only?

wbond:
Yeah, I think so. I'm not sure it is really all that useful for the suffix to be specialized. The suffixes are all part of the `constant.numeric` anyway, right? If so, then a specific suffix could be targeted via that.

deathaxe:
Further specialization might indeed not add much value to that single character. Wouldn't help selectors to interact with the whole number in a special way. It has 3 levels of scopes and is used by C# already.

Should propably mension it some where to be the scope for suffixes, primarily?

jwortmann:
> The suffixes are all part of the constant.numeric anyway, right?

I know that in Rust they are not. And D uses `storage.type.integer.d` and `storage.type.float.d` instead of `storage.type.numeric.d`. These should be adjusted for consistency.

wbond:
Thanks for all of the work on this everyone!

PR #2136, opened 2019-10-19
[Python] Improve numbers

deathaxe:
This PR proposes to ...

1. rename some scope names to comply with the recent changes to the scope naming guidelines.
2. merges the float and imaginary match rules in order to improve parsing performance by about 14%. The solution is borrowed from Perl and adapted to Python.

EDIT:

Another banchmark even shows up to 30% improvement. The test case consists of some more complicated numbers.

```
12_34_5.1_23_45e+10 12_34_5.1_23_45e+10 12_34_5.1_23_45e+10 12_34_5.1_23_45e+10
```

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -33,8 +33,8 @@ variables:
   identifier_continue: '[[:alnum:]_]'
   identifier: '\b[[:alpha:]_]{{identifier_continue}}*\b'
   identifier_constant: '\b(?:[\p{Lu}_][\p{Lu}_\d]*)?[\p{Lu}]{2,}[\p{Lu}_\d]*\b'  # require 2 consecutive upper-case letters
-  digit: (?:\d+(?:_\d+)*)
-  exponent: (?:[Ee][-+]?{{digit}})
+  digits: (?:\d+(?:_\d+)*)
+  exponent: (?:[eE][-+]?{{digits}})
   path: '({{identifier}}[ ]*\.[ ]*)*{{identifier}}'
   sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|WITH)\b
   illegal_names: (?:and|as|assert|break|class|continue|def|del|elif|else|except|finally|for|from|global|if|import|in|is|lambda|not|or|pass|raise|return|try|while|with|yield)
@@ -457,10 +457,12 @@ contexts:
     # complex
     - match: |-
         (?x:
-                      (\.)  {{digit}}  {{exponent}}? |
-          \b{{digit}} (\.)? {{digit}}? {{exponent}}?
-        )([Jj])
-      scope: constant.numeric.complex.imaginary.python
+          # 1.j, 1.1j, 1.1e1j, 1.1e-1j, 1.e1j, 1.e-1 | 1e1j, 1e-1j
+          \b{{digits}} (\.)? {{digits}}? {{exponent}}?
+          # .1j, .1e1j, .1e-1j
+          | (\.) {{digits}} {{exponent}}?
+        )([jJ])
+      scope: constant.numeric.complex.imaginary.decimal.python
       captures:
         1: punctuation.separator.decimal.python
         2: punctuation.separator.decimal.python
@@ -468,8 +470,10 @@ contexts:
     # floating point
     - match: |-
         (?x:
-                          (\.) {{digit}}  {{exponent}}? |
-          \b{{digit}} (?: (\.) {{digit}}? {{exponent}}? | {{exponent}} )
+          # 1., 1.1, 1.1e1, 1.1e-1, 1.e1, 1.e-1 | 1e1, 1e-1
+          \b{{digits}} (?: (\.) {{digits}}? {{exponent}}? | {{exponent}} )
+          # .1, .1e1, .1e-1
+          | (\.) {{digits}} {{exponent}}?
         )
       scope: constant.numeric.float.decimal.python
       captures:
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -425,7 +425,7 @@ contexts:
       scope: constant.numeric.integer.hexadecimal.python
       captures:
         1: punctuation.definition.numeric.hexadecimal.python
-        2: storage.type.numeric.long.python
+        2: storage.type.numeric.python
     - match: \b(?i)(0x)(_?\h)+
       scope: constant.numeric.integer.hexadecimal.python
       captures:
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -454,39 +455,29 @@ contexts:
...
-    - match: (?i){{digitpart}}?(\.){{digitpart}}(?:e[\-\+]?{{digitpart}})? # mandatory fraction
-      scope: constant.numeric.float.python
-      captures:
-        1: punctuation.separator.decimal.python
-    - match: \b(?i)(?:{{digitpart}}(?:(\.){{digitpart}}?)?|(\.){{digitpart}})(?:e[\-\+]?{{digitpart}}) # mandatory exponent
-      scope: constant.numeric.float.python
+    - match: |-
+        (?x:
+                          (\.) {{digit}}  {{exponent}}? |
+          \b{{digit}} (?: (\.) {{digit}}? {{exponent}}? | {{exponent}} )
```
FichteFoll: The `|` being on the same column as above is kinda irritating because it's actually within a nested group. I'd put `\b{{digit}} {{exponent}}` on a third line.

Python/Python.sublime-syntax
```diff
@@ -33,7 +33,8 @@ variables:
   identifier_continue: '[[:alnum:]_]'
   identifier: '\b[[:alpha:]_]{{identifier_continue}}*\b'
   identifier_constant: '\b(?:[\p{Lu}_][\p{Lu}_\d]*)?[\p{Lu}]{2,}[\p{Lu}_\d]*\b'  # require 2 consecutive upper-case letters
-  digitpart: (?:\d(?:_?\d)*)
+  digits: (?:\d+(?:_\d+)*)
```
FichteFoll: I was just writing a review and suggesting this to be `digits` when it was updated. 👀

FichteFoll:
I like how much more readable the combined match patterns are (barring the one I commented on). Good job on those.

Python/Python.sublime-syntax
deathaxe: Hahaha.

Python/Python.sublime-syntax
deathaxe: The idea is to prevent the parser from parsing a token twice. Therefore a dedicated `\b{{digit}} {{exponent}}` next to `\b{{digit}} (\.) {{digit}}? {{exponent}}?` would reduce the performance (slightly, but meassurable).


How about just ...

```yml
    - match: |-
        (?x:
          \b{{digits}} (?: (\.) {{digits}}? {{exponent}}? | {{exponent}} )
          | (\.) {{digits}} {{exponent}}?
        )
      scope: constant.numeric.float.decimal.python
      captures:
        1: punctuation.separator.decimal.python
        2: punctuation.separator.decimal.python
```

Python/Python.sublime-syntax
```diff
@@ -454,39 +455,33 @@ contexts:
       captures:
         1: punctuation.definition.numeric.binary.python
     # complex
```
eric-wieser: nit: Should this line say `imaginary` now?

deathaxe: The comment? An imaginary is still a part of a complex number. Maybe someone is keen enough to add the real part in the future, too?

FichteFoll: Complex numbers in Python are constructed by adding any numeric with an imaginary number annotated with `j`, e.g. `1+2j`. There is a convention that suggests not to use spaces around this operator to indicate the creation of a complex number with a non-zero real part, which we may target, but that would incure duplicating basically every match pattern for numeric literals and I don't know what your intention with the scopes is.

deathaxe: ... that's why "keen" ... :-)  

Was discussed with @jwortmann in the Ruby PR and we agree it to make things too complex with only little or even no value.

FichteFoll: https://github.com/sublimehq/Packages/pull/2135#discussion_r337142160 for reference.

wbond: My idea of how to scope complex numbers is at https://github.com/sublimehq/Packages/pull/2135#discussion_r338225323.

eric-wieser: > the key thing we want to highlight here is that the + is not an addition sign, but part of making the whole thing an imaginary number

This isn't super meaninful in python - constant folding means that `1+1j` is no more an atom for constructing complex numbers than `1 + 1 + 1j + 1j` is.

My original remark was simply that the comment should be updated to match the code change, maybe to "imaginary literals for complex numbers" or something.

deathaxe: The ongoing discussion is just about some aspects of R or Ruby with regards of imaginary numbers being combined with other type modifiers like `r` for rational numbers or even hexadecimal expressions.

wbond:
I'm thinking `storage.type.numeric.python` instead of `storage.type.numeric.imaginary.python` because of https://github.com/sublimehq/Packages/pull/2135#issuecomment-546696712 and https://github.com/sublimehq/Packages/pull/2135#issuecomment-547543859.

Python/Python.sublime-syntax
```diff
@@ -421,20 +422,20 @@ contexts:
...
       scope: constant.numeric.integer.hexadecimal.python
       captures:
         1: punctuation.definition.numeric.hexadecimal.python
     # octal
     - match: \b(?i)(0o?)(?=o|[0-7])[0-7]*(L) # py2
-      scope: constant.numeric.integer.long.octal.python
+      scope: constant.numeric.integer.octal.python
       captures:
         1: punctuation.definition.integer.octal.python
-        2: storage.type.numeric.long.python
```
eric-wieser: Why the removal of `long`?

wbond: Because we are standardizing on `storage.type.numeric` for literal storage type suffixes across syntaxes. Certainly languages allow multiple suffixes, and by specializing the scope per letter suffix, we actually result in having multiple tokens for the suffix, which is odd, and messy. Plus, there isn't a whole lot of value in adding `.long` since you can infer that from `constant.numeric.integer storage.type.numeric.python`.

eric-wieser: > we actually result in having multiple tokens for the suffix,

can you elaborate on this?

eric-wieser: Ah, sorry - I think this comment was intended for line 434 - I can see the argument that having `long` in both inner and outer scopes is not useful.

wbond: >> we actually result in having multiple tokens for the suffix,
> can you elaborate on this?

Every token in a text buffer in Sublime Text has a scope. Any time the scope changes, a there must, necessarily then, be a token boundary.

For example, imagine `123ri` in Ruby. The entire thing will have the scope `source.ruby constant.numeric.*.ruby`. However, the `ri` will also have `storage.type.numeric` since it is a suffix. Thus `123` is one token, and `ri` is another. If we specialize the suffix scope so that `r` gets `storage.type.numeric.rational.ruby` and `i` gets `storage.type.numeric.imaginary.ruby`, then we now have three tokens, `123`, `r` and `i`.

By **not** specializing the `r` and the `i`, we have two tokens, `123` and `ri`.

PR #2137, opened 2019-10-20
[R] Improve numbers

deathaxe:
This commit ...

1. adds punctuation scopes to numbers.
2. renames the imaginary numbers to `constant.numeric.complex.imaginary`
3. merges some rules to compensate performance impacts caused by the added capture groups required for (1).
4. makes use of character classes where appropriate (e.g.: [eE] )
5. removes some unnecessary escapes (e.g. (\+\-) => [-+])
6. adds a couple of test cases for the different types of numbers

### DIFF
--- a/R/R.sublime-syntax
+++ b/R/R.sublime-syntax
@@ -45,13 +45,13 @@ contexts:
       scope: constant.language.r
     # complex imaginary
     - match: \b(0[xX])\h+(?:(i)|(I))\b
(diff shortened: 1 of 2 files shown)


wbond:
My hunch is that once the discussions in #2135 are resolved this will probably need some tweaks, right?

deathaxe:
I guess this is the status quo. May still need some input there about what's the best strategy.

PR #2138, opened 2019-10-21
[RestructuredText] Improve handling of directives, verbatim, and comments

eric-wieser:
This:
* Allows comments to start at columns other than the first
* Allows blank lines to appear in comments, directives, and verbatim blocks

---

My eventual goal here is to allow:
* Latex environments to be embedded in sphinx RsT `.. math::` blocks
* RsT blocks to be embedded in python docstrings

### DIFF
--- a/RestructuredText/reStructuredText.sublime-syntax
+++ b/RestructuredText/reStructuredText.sublime-syntax
@@ -24,10 +24,6 @@ contexts:
         4: punctuation.separator.key-value.restructuredtext
       push:
         - match: '^(?!\1[ \t])'
-          captures:
-            2: meta.directive.restructuredtext
-            3: punctuation.definition.directive.restructuredtext
-            4: punctuation.separator.key-value.restructuredtext
           pop: true
         - include: scope:text.html.basic
     - match: '(\.\.)\s[A-z][A-z0-9-_]+(::)\s*$'
@@ -44,9 +40,6 @@ contexts:
       push:
         - meta_scope: meta.raw.block.restructuredtext
         - match: '^(?=\1[^\s]+)'
-          captures:
-            2: markup.raw.restructuredtext
-            3: punctuation.definition.raw.restructuredtext
           pop: true
         - match: .+
           scope: markup.raw.restructuredtext
--- a/RestructuredText/reStructuredText.sublime-syntax
+++ b/RestructuredText/reStructuredText.sublime-syntax
@@ -16,7 +16,7 @@ contexts:
           pop: true
         - include: inline
   inline:
-    - match: '^(\s*)((\.\.)\sraw(::)) html'
+    - match: '^(\s*)((\.\.)\s+raw(::)) html'
       comment: directives.html
       captures:
         2: meta.directive.restructuredtext
@@ -27,7 +27,7 @@ contexts:
           pop: true
         - meta_content_scope: text.html.basic
         - include: scope:text.html.basic
-    - match: '^(\s*)((\.\.)\s[A-z][A-z0-9\-_]+(::))\s*'
+    - match: '^(\s*)((\.\.)\s+[A-z][A-z0-9\-_]+(::))\s*'
       comment: directives
       captures:
         2: meta.other.directive.restructuredtext
--- a/RestructuredText/reStructuredText.sublime-syntax
+++ b/RestructuredText/reStructuredText.sublime-syntax
@@ -9,37 +9,42 @@ file_extensions:
 scope: text.restructuredtext
 contexts:
   main:
-    - match: '^([ \t]*)(?=\S)'
+    - match: '^(?=(\s*)\S)'
       push:
         - meta_content_scope: meta.paragraph.restructuredtext
         - match: ^(?!\1(?=\S))
           pop: true
         - include: inline
   inline:
-    - match: '^([ \t]*)((\.\.)\sraw(::)) html'
+    - match: '^(\s*)((\.\.)\sraw(::)) html'
       comment: directives.html
       captures:
         2: meta.directive.restructuredtext
         3: punctuation.definition.directive.restructuredtext
         4: punctuation.separator.key-value.restructuredtext
       push:
-        - match: '^(?!\1[ \t])'
+        - match: '^(?!\1\s+)(?=\s*\S+)'
           pop: true
+        - meta_content_scope: text.html.basic
         - include: scope:text.html.basic
-    - match: '(\.\.)\s[A-z][A-z0-9-_]+(::)\s*$'
+    - match: '^(\s*)((\.\.)\s[A-z][A-z0-9\-_]+(::))\s*'
       comment: directives
-      scope: meta.other.directive.restructuredtext
       captures:
-        1: punctuation.definition.directive.restructuredtext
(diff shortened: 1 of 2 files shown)


RestructuredText/reStructuredText.sublime-syntax
```diff
@@ -9,44 +9,42 @@ file_extensions:
 scope: text.restructuredtext
 contexts:
   main:
-    - match: '^([ \t]*)(?=\S)'
+    - match: '^(?=(\s*)\S)'
       push:
         - meta_content_scope: meta.paragraph.restructuredtext
-        - match: ^(?!\1(?=\S))
+        - match: '^(?!\1\s+)(?=\s*\S+)'
```
FichteFoll: A paragraph should end when there is a blank line inbetween, don't you think?

RestructuredText/reStructuredText.sublime-syntax
```diff
@@ -9,44 +9,42 @@ file_extensions:
...
+    - match: '^(\s*)((\.\.)\s[A-z][A-z0-9\-_]+(::))\s*'
       comment: directives
-      scope: meta.other.directive.restructuredtext
       captures:
-        1: punctuation.definition.directive.restructuredtext
-        2: punctuation.separator.key-value.restructuredtext
-    - match: '^([ \t]*).*?((::))'
+        2: meta.other.directive.restructuredtext
+        3: punctuation.definition.directive.restructuredtext
+        4: punctuation.separator.key-value.restructuredtext
```
FichteFoll: These aren't really key-value pairs. The colons rather serve the purpose of starting the directive's body.

RestructuredText/syntax_test_restructuredtext.rst
```diff
@@ -60,3 +66,69 @@ This is *italic*.
 
 this is **bold**.
 ..      ^^^^^^^^ markup.bold.restructuredtext
+
+
+Directive tests
+--------------
+
+.. not possible to test a multi-line directive, since comments cannot appear
+.. mid-directive
```
FichteFoll: Syntax tests prioritize the "comment test marker token" over valid syntax, meaning you can use them anywhere you like and still test the previous line. You do this in line 87 as well.

RestructuredText/syntax_test_restructuredtext.rst
```diff
@@ -13,6 +13,12 @@
 ..
   a multi-line comment ends at the first character in the
   first column
+This is not a comment
+.. <- meta.paragraph.restructuredtext
+
+.. multi line comments can contain
+
+ blank lines are still part of the comment if they're indented by more
```
FichteFoll: There should be a test here.

RestructuredText/syntax_test_restructuredtext.rst
eric-wieser: The presence of a comment marker invalidates the directive contents. This RST produces a note

```
.. note::

    My note
```

This produces only a comment

```
.. note::
.. <-- test comment here

    My note
```

RestructuredText/reStructuredText.sublime-syntax
eric-wieser: This is just continuing with what was already there - I'd prefer to leave scope names out of scope for this review, since I didn't change any of them.

RestructuredText/reStructuredText.sublime-syntax
eric-wieser: Is there any way to test this?

RestructuredText/syntax_test_restructuredtext.rst
eric-wieser: Added

RestructuredText/syntax_test_restructuredtext.rst
eric-wieser: To write such a test, I would need a marker like
```
.. note ::
    .. <-[2] scope.name
    .. <-[2] means "apply to column 2
```

eric-wieser: Ah, found a workaround in the rST to get what I wanted

RestructuredText/reStructuredText.sublime-syntax
FichteFoll: Hm, I see now that the old syntax didn't do this either. Probably not actually worth bothering with.

RestructuredText/reStructuredText.sublime-syntax
FichteFoll: I only looked at the diff and noticed that you added this scope, but I see that it is used elsewhere already. I agree that this isn't the place to change them.

RestructuredText/reStructuredText.sublime-syntax
eric-wieser: I think you're right though, this change was incorrect, I'll revert this line

RestructuredText/syntax_test_restructuredtext.rst
FichteFoll: The usual "workaround" is to repeat the mutli-line construct and then add tests at a later point.

eric-wieser: That's exactly what I ended up doing in the first rev of this, but I think my approach now of doing both is better anyway. Removed this comment, so will resolve this review chain.

wbond:
> My eventual goal here is to allow:
>
> Latex environments to be embedded in sphinx RsT .. math:: blocks
> RsT blocks to be embedded in python docstrings

Unfortunately this likely won't be merged in until we resolve the issue with embedding syntaxes and memory growth/load time. Currently when you embed or include a syntax in another, at syntax compile time, all of the contexts are copied into the definition. This is done for performance, but also to allow different syntaxes to be used by different background loading threads at the same time, rather than sharing a single syntax across all threads, and have each wait for a lock on the shared syntax definition.

Because of this, syntaxes that include lots of others get big, and even the cached version gets large on disk and is slow to load. Unfortunately LaTeX embeds a whole bunch of other syntaxes to allow them to be used in code blocks. By embedding LaTeX in RST and then embedding RST in Python would cause Python to become huge and slow to load.

Granted, we need to find a way to solve this in Sublime Text, but I just wanted to give you a heads up that until that happens, such changes probably won't be merged into the default packages.

wbond:
Thanks for your work on this!

eric-wieser:
> By embedding LaTeX in RST and then embedding RST in Python would cause Python to become huge and slow to load.

Would you be in favor of doing just the former? Strictly it would be rST/Sphinx at that point since `..math::` is sphinx-specific - but in practice, I suspect that represents a large fraction of rst use.

Also, is there any easy way for me to create a syntax derived from python but with a change to the docstring section short of copying and modifying

FichteFoll:
>Also, is there any easy way for me to create a syntax derived from python but with a change to the docstring section short of copying and modifying it?

Your best bet for injecting matches into an existing syntax's context currently is https://packagecontrol.io/packages/YAMLMacros.

eric-wieser:
@FichteFoll: Ultimately I'd still need to rebuild the macro'd version each time the Python.sublime-syntax changes though, right?

FichteFoll:
Yes, that is correct.

PR #2139, opened 2019-10-22
[Go] Templates String Updates

patrickrgaffney:
Two minor updates to template strings:

- New `slice` function [added in v1.13](https://golang.org/doc/go1.13#text/template).
- Re-assignment operator `=` [added in v1.11](https://golang.org/doc/go1.11#text/template).

> *Sidebar*: I added a test for the new `slice` function. Previously the only function being tested was `printf`. If we think there's value in testing all those template functions, I can add additional tests.


wbond:
Awesome, thanks!

PR #2141, opened 2019-10-22
[Git Formats] Match hash in commit message

FichteFoll:
I feel that hashes in commit messages deserve more attention than they do now (which is none at all), because they are frequently used to cross-reference commits.

I intended to use a `variable.other.hash` scope, because that's basically what hashes are used like (as references), but we already have `constant.numeric.hex.hash.git.commit` in the syntax definition, so I didn't change that. I totally could, if people are in favor of it, however.

### DIFF
--- a/Git Formats/Git Commit.sublime-syntax
+++ b/Git Formats/Git Commit.sublime-syntax
@@ -75,6 +75,8 @@ contexts:
           pop: true
         - include: Git Common.sublime-syntax#references
         - include: signed-off
+        - match: '{{hash}}'
+          scope: constant.numeric.hex.hash.git.commit
 
   signed-off:
     - match: ^\s*(Signed-off-by)\s*(:)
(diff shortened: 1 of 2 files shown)


deathaxe:
Might not be a strong argument as it is easy to change, but unfortunatelly numerous color schemes scope variables just like plain text, which would therefore cause scoping hashes as such not to make a visual difference for many users. 

Furthermore we are talking about text. A commit message feels more like a markdown.

Furthermore repository references use `constant` as well - borrowed from markdown. ... Ok - emails don't.

With regards to scope naming guidelines the `constant.numeric.hex` is a bit out dated ;-) as well, but  `constant.numeric.integer.hexadecimal.hash.git.commit` feels quite too long.

Semantically the hash could also use `constant.other[.reference].hash.git.commit`.

Thoughts?

wbond:
I like the idea of `constant.other.hash.git.commit`

FichteFoll:
I like it, too. Adjusted.

PR #2145, opened 2019-10-25
[RestructuredText] Add support for sphinx `math` directives and roles

eric-wieser:
These are treated as being embedded latex math of the right type (block/inline).

Notably, this makes them compatible with the inline preview rendering of the LaTeXTools plugin.

These are specifix to Sphinx, but represent:
* A frequent use case of rST
* One of the few things that `math` is likely to be defined as anyway

Notes on scope names:
- `support.function.role.builtin.restructuredtext` is a compromise between getting colored under the default highlighter while indicating the rst term for the `` :name:`text` `` construct
- `punctuation.definition.interpreted.restructuredtext` is used elsewhere in this file for backticks
- Including `text.tex.latex` as a root scope for embedded latex is enough to encourage the LaTeXTools plugin to get involved. Strictly, this would work without any syntax embedded at all, but having LaTeX highlighting feels natural here

---

To actually work for LaTeX previews, this needs https://github.com/SublimeText/LaTeXTools/pull/1438

### DIFF
--- a/RestructuredText/reStructuredText.sublime-syntax
+++ b/RestructuredText/reStructuredText.sublime-syntax
@@ -27,6 +27,17 @@ contexts:
           pop: true
         - meta_content_scope: text.html.basic
         - include: scope:text.html.basic
+    - match: '^(\s*)((\.\.)\s+math(::))\s*'
+      comment: directives.math
+      captures:
+        2: meta.directive.restructuredtext
+        3: punctuation.definition.directive.restructuredtext
+        4: punctuation.separator.key-value.restructuredtext
+      push:
+        - match: '^(?!\1\s+)(?=\s*\S+)'
+          pop: true
+        - meta_content_scope: text.tex.latex meta.environment.math.block.restructuredtext
+        - include: scope:text.tex.latex#math-content
     - match: '^(\s*)((\.\.)\s+[A-z][A-z0-9\-_]+(::))\s*'
       comment: directives
       captures:
@@ -77,6 +88,17 @@ contexts:
       scope: markup.underline.substitution.restructuredtext
       captures:
         1: punctuation.definition.substitution.restructuredtext
+    - match: (:math:)(`)
+      comment: links `...`_ or `...`__
+      captures:
+        1: support.function.role.builtin.restructuredtext
+        2: punctuation.definition.interpreted.restructuredtext
+      push:
+        - match: '`'
+          scope: punctuation.definition.interpreted.restructuredtext
+          pop: true
+        - meta_content_scope: text.tex.latex meta.environment.math.inline.restructuredtext
+        - include: scope:text.tex.latex#math-content
     - match: \b(\w+)(_)\b
       comment: links `...`_ or `...`__
       scope: meta.link.reference
--- a/RestructuredText/reStructuredText.sublime-syntax
+++ b/RestructuredText/reStructuredText.sublime-syntax
@@ -89,7 +89,7 @@ contexts:
       captures:
         1: punctuation.definition.substitution.restructuredtext
     - match: (:math:)(`)
-      comment: links `...`_ or `...`__
+      comment: embedded math-mode latex, https://docutils.sourceforge.io/docs/ref/rst/directives.html#math
       captures:
         1: support.function.role.builtin.restructuredtext
         2: punctuation.definition.interpreted.restructuredtext
--- a/RestructuredText/syntax_test_restructuredtext.rst
+++ b/RestructuredText/syntax_test_restructuredtext.rst
@@ -135,3 +135,18 @@ Not verbatim
 
     </b>
 ..  ^^^^ text.html.basic
+
+.. math::
+
+    \sum_{i=0}^N x_i^2
+..  ^^^^ text.tex.latex support.function.math.tex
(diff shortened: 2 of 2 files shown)


RestructuredText/syntax_test_restructuredtext.rst
```diff
@@ -134,4 +134,19 @@ Not verbatim
...
+..  ^^^^^^^^^^^^^^^^^^^ text.tex.latex meta.environment.math.block.restructuredtext
+
+
+:test:
+
+   :math:`\alpha + \beta`
+.. ^^^^^^ support.function.role.builtin.restructuredtext
+..       ^ punctuation.definition.interpreted.restructuredtext
+..        ^^^^^^^^^^^^^^ text.tex.latex meta.environment.math.inline.restructuredtext
+..                      ^ punctuation.definition.interpreted.restructuredtext
```
eric-wieser: With the linked patch to LaTexTools, and a config change there to also render inline math, this displays as:
image

RestructuredText/reStructuredText.sublime-syntax
```diff
@@ -27,6 +27,17 @@ contexts:
...
+      comment: directives.math
+      captures:
+        2: meta.directive.restructuredtext
+        3: punctuation.definition.directive.restructuredtext
+        4: punctuation.separator.key-value.restructuredtext
+      push:
+        - match: '^(?!\1\s+)(?=\s*\S+)'
+          pop: true
+        - meta_content_scope: text.tex.latex meta.environment.math.block.restructuredtext
+        - include: scope:text.tex.latex#math-content
```
FichteFoll: Have you tried using `embed` for this?

RestructuredText/reStructuredText.sublime-syntax
```diff
@@ -77,6 +88,17 @@ contexts:
       scope: markup.underline.substitution.restructuredtext
       captures:
         1: punctuation.definition.substitution.restructuredtext
+    - match: (:math:)(`)
+      comment: links `...`_ or `...`__
```
FichteFoll: This comment is clearly wrong. Just remove it.

RestructuredText/reStructuredText.sublime-syntax
```diff
@@ -77,6 +88,17 @@ contexts:
       scope: markup.underline.substitution.restructuredtext
       captures:
         1: punctuation.definition.substitution.restructuredtext
+    - match: (:math:)(`)
```
FichteFoll: Only the `math` part should become `support.function`, not the punctuation.

RestructuredText/reStructuredText.sublime-syntax
eric-wieser: Copy paste, whoops

RestructuredText/reStructuredText.sublime-syntax
eric-wieser: What's the difference between this approach and that one?

wbond: `embed` ensures that no matter what context the "embedded" syntax/context is in, it always pops all the way out as soon as the `escape` pattern is found.

For instance, if `math-content` pushed into a sub-context for a `(`, it may get stuck in there if there isn't a `)`. If using `embed` and `escape`, the first thing that happens is that we look for the `escape` pattern. If that is found, any content until then is parsed using the specified context. So it wouldn't matter if there is no `)`, it will exit the embedded context once finding the `escape` pattern.

Beyond that, the special nature of `embed` allows us to apply certain optimizations behind the scenes for performance. There will likely also be more optimizations made in the future.

wbond:
I should note, that this should not run into the performance problems I was afraid of with LaTeX since you are using only a specific, rather limited context from LaTeX.

wbond:
@eric-wieser If you address the comments from @FichteFoll then this should be in shape to merge.

RestructuredText/reStructuredText.sublime-syntax
eric-wieser: I'm not sure I agree, based on how github highlights the word `:test:` in the rst file. If nothing else, I find it more helpful as a reader to see `:math:` in a single color. Ok to be overruled if you think being precise is more important than looking pretty

FichteFoll: I tried to think of similar applications and apparently we've been doing things in either of two ways:

1. Scope everything, including punctuation, as the keyword. An example are at-rules in CSS, i.e. `@media`, or dollar signs for languages that *require* them for variables, like Php with `$var`.
2. Only scope the name, not the punctuoation. Examples for this are annotations in Java or decorators in Python, but those are different because the punctuation there doesn't describe an atomic but rather a construct, which is why it also gets a meta scope.

This case fits into 1., so I we probably should include the colons at least until we decided on a uniform standard for this type of token.
Either way, punctuation *should* receive a `punctuation` scope, either on top of the `support.function` scope or without it entirely.

jrappen:
@eric-wieser Could you update your PR for it to be merged?

RestructuredText/reStructuredText.sublime-syntax
eric-wieser:
```suggestion
      comment: embedded math-mode latex, https://docutils.sourceforge.io/docs/ref/rst/directives.html#math
```

PR #2150, opened 2019-10-25
[Markdown] Add test case for ligatures vs. block-quote

deathaxe:


wbond:
I'm not quite sure how this is testing that ligatures work?

deathaxe:
This test case is just to ensure that `>` in `>=` is scoped as punctuation for an quotation depending on the indention level. If a line starts with 3 or more spaces, it should no longer be scoped as punctuation, which seems not to be tested until now at all.

Basically there is no real possibility to test ligatures without adding rules to scope them in a certain way, which I don't know would be worth/helpful?

wbond:
Ok, just wanted to make sure I understood the intent!

PR #2153, opened 2019-10-26
[PHP] Fix missing (?x:) for regexes

jfcherng:
After trying the merged https://github.com/sublimehq/Packages/pull/2134 in my daily work. I found constants no longer get highlighted. And according my test, PHP's consts are case-*sensitive* hence should not have `(?i:)` except for some special constants like `true`, `false`, `null`.

```
[Clover@Clover-NB nthu-ee-ust-enroll-censor](master)$ php74 -r "echo JSON_THROW_ON_ERROR;"
4194304
[Clover@Clover-NB nthu-ee-ust-enroll-censor](master)$ php74 -r "echo json_THROW_ON_ERROR;"
Warning: Use of undefined constant json_THROW_ON_ERROR - assumed 'json_THROW_ON_ERROR' (this will throw an Error in a future version of PHP) in Command line code on line 1
json_THROW_ON_ERROR
```

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -324,7 +324,7 @@ contexts:
 
   class-builtin:
     - match: |-
-        (?i)(\\)?\b(
+        (\\)?\b(?xi:
           AMQPConnection | AMQPExchange | AMQPQueue | APCIterator | AppendIterator | ArrayAccess | ArrayIterator | ArrayObject |
           BadFunctionCallException | BadMethodCallException | CachingIterator | Collator | Countable | DOMAttr | DOMCharacterData | DOMComment |
           DOMDocument | DOMDocumentFragment | DOMElement | DOMEntityReference | DOMImplementation | DOMNamedNodeMap | DOMNode | DOMNodelist |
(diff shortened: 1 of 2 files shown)


PHP/PHP Source.sublime-syntax
```diff
@@ -324,7 +324,7 @@ contexts:
 
   class-builtin:
     - match: |-
-        (\\)?\b(?xi:
+        (?xi)(\\)?\b(
```
jfcherng: feel like a WTF to me

wbond:
Perhaps we should have a test for at least one const from each list so something like this doesn't regress?

PHP/syntax_test_php.php
```diff
@@ -1098,6 +1098,28 @@ function generate2()
...
 
+// class name should be case-insensitive
+$object = new ArRaYoBjEcT();
+//            ^^^^^^^^^^^ support.class.builtin
+
+// constant name should be case-sensitive
+$const = E_aLL;
+//       ^^^^^ - support.constant.core
+
+// function name should be case-sensitive
```
jfcherng: oh god, here's a typo. it should be case-*insensitive*.

Issue #1440, opened 2018-02-23
[Meta] A new workflow for scope naming guidelines


FichteFoll:
## Current Situation

The [scope naming guidelines on the official docs](http://www.sublimetext.com/docs/3/scope_naming.html) can't catch up with how frequently and detailed scoping is being discussed at this repo currently. Because the website source isn't public, the only people to work on them are Will and Jon, and they arguably have more important things to do.

We currently have a couple labeled issues on this repo, most of which are still open and without having reached a conclusion. The direction of the discussions is sometimes clear, but not always as it's fairly loose (which is fine by me), but the lack of a properly formulated conclusion that can be applied to actual syntax definitions is frustrating.

Thus, I propose a more formal workflow for scope naming guidelines that will hopefully produce more, faster and better results.

## Goals

1. Have a referable collection of scope naming conventions and guidelines for specific syntax constructs that we have discussed previously. The guidelines should be clear and cover all usual and the most common edge cases for the topic they address. 
2. That collection should have a certain authority over scope naming so that it can be adapted for the default packages as well as third-party syntax definitions.
3. Reduce delay between discussions of guidelines to the actual formulation of these.

For example, there would be a document about string scoping that covers current practices of the general `string` scopes but also how string interpolation is handled and what that actually is.

## Proposal

I propose to adapt a workflow similar to [PEPs](https://www.python.org/dev/peps/pep-0001/#start-with-an-idea-for-python), where a document is prepared by a member of the community and other community members can discuss and refine it until it is eventually added to the guideline collection or rejected.

To concretize:

- There would be a repository on the SublimeText Github organization.
- That repo will be used to store files with markup text explaining the individual guidelines, as well as outline the workflow.
- New proposals are suggested through pull requests and then discussed and adjusted until it is "accepted" (see later).
- Guidelines can be edited later (again though a PR and review process)
- Before a guideline is formulated, discussion may take place in issues on the repo.

Details:

- The potentially multiple authors of a guideline become the primary negotiators for any further changes to that guideline.
- Guidelines will need a number associated with them so they can be renamed without losing referability, sort of. I think a zero-padded three-digit number will suffice. Filenames would then follow the schema `<number>-<title>.<ext>`.
- The guidelines need to be versioned. A header section with metadata, including the document's version, last change date and authors would be useful. Semver seems overkill, but makes thinking about our own version schema unnecessary. The first accepted version should then always be 1.0.0.

## Questions

1. How should our repo and the proposals be named?

[...]

PR #2156, opened 2019-10-27
[Ruby] Improve keyword scopes

deathaxe:
This PR applies the #1228 and tries to find proper solutions for some special keywords/situations where needed.

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -118,13 +118,13 @@ contexts:
   class:
     # Defining a class method
     - match: \bclass\b(?=\s*<)
-      scope: storage.type.class.ruby
+      scope: storage.type.class.ruby keyword.declaration.class.ruby
     - match: \bclass\b
-      scope: meta.class.ruby storage.type.class.ruby
+      scope: storage.type.class.ruby keyword.declaration.class.ruby
       push: class-declaration
 
   class-declaration:
-    - meta_content_scope: meta.class.ruby
+    - meta_scope: meta.class.ruby
     - match: '(?={{path_lookahead}})'
       set: class-name
     # Escape if no valid match
@@ -158,14 +158,14 @@ contexts:
 
   module:
     - match: \bmodule\b
-      scope: meta.module.ruby storage.type.module.ruby
+      scope: storage.type.namespace.ruby keyword.declaration.namespace.ruby
       push: module-declaration
 
   module-declaration:
-    - meta_content_scope: meta.module.ruby
+    - meta_scope: meta.namespace.ruby
     - match: '(?=(::)?({{identifier}}::)*{{identifier}})'
       set:
-        - meta_content_scope: meta.module.ruby entity.name.module.ruby
+        - meta_content_scope: meta.namespace.ruby entity.name.namespace.ruby
         - include: name-parts
         - match: ''
           pop: true
@@ -824,7 +824,7 @@ contexts:
 
   method:
     - match: \bdef\b
-      scope: meta.function.ruby support.type.function.ruby
+      scope: meta.function.ruby storage.type.function.ruby keyword.declaration.function.ruby
       push:
         - meta_content_scope: meta.function.ruby
         - match: '(self)(\.)({{identifier}}{{method_punctuation}}|===?|>[>=]?|<=>|<[<=]?|[%&`/\|]|\*\*?|=?~|[-+]@?|\[\]=?)'
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -289,9 +289,8 @@ contexts:
         2: invalid.illegal.character.ruby
 
   blocks:
-    - match: \b(do)\b\s*
-      captures:
-        1: keyword.control.start-block.ruby
+    - match: \bdo\b
+      scope: keyword.control.block.do.ruby
       push: maybe-block-parameters
     - match: \{
       scope: punctuation.section.scope.ruby
@@ -330,20 +329,77 @@ contexts:
             - include: expressions
 
   keywords:
-    - match: '\b(BEGIN|END)\b(?![?!])'
-      scope: keyword.control.ruby
-    - match: '\b(begin|end|ensure|rescue)\b(?![?!])'
-      scope: keyword.control.ruby
-    - match: '\b(case|else|ensure|for|in|then)\b(?![?!])'
-      scope: keyword.control.ruby
-    - match: '\b(elsif|if|unless|when|while|until)\b(?![?!])'
-      scope: keyword.control.ruby
+    # blocks
+    - match: \bbegin\b(?![?!])
+      scope: keyword.control.block.begin.ruby
       push: after-keyword
-    - match: \b(and|not|or)\b
+    - match: \bend\b(?![?!])
+      scope: keyword.control.block.end.ruby
+    # conditional
+    - match: \bcase\b(?![?!])
+      scope: keyword.control.conditional.case.ruby
+      push: after-keyword
+    - match: \belse\b(?![?!])
+      scope: keyword.control.conditional.if.ruby
+      push: after-keyword
+    - match: \belsif\b(?![?!])
+      scope: keyword.control.conditional.elseif.ruby
+      push: after-keyword
+    - match: \bif\b(?![?!])
+      scope: keyword.control.conditional.if.ruby
+      push: after-keyword
+    - match: \bthen\b(?![?!])
+      scope: keyword.control.conditional.then.ruby
+    - match: \bunless\b(?![?!])
+      scope: keyword.control.conditional.unless.ruby
+      push: after-keyword
+    - match: \bwhen\b(?![?!])
+      scope: keyword.control.conditional.when.ruby
+      push: after-keyword
+    # exception
+    - match: \bensure\b(?![?!])
+      scope: keyword.control.exception.ensure.ruby
+      push: after-keyword
+    - match: \brescue\b(?![?!])
+      scope: keyword.control.exception.rescue.ruby
+      push: after-keyword
+    # loop
+    - match: \bfor\b(?![?!])
+      scope: keyword.control.loop.for.ruby
+      push: after-keyword
+    - match: \buntil\b(?![?!])
+      scope: keyword.control.loop.until.ruby
+      push: after-keyword
+    - match: \bwhile\b(?![?!])
+      scope: keyword.control.loop.while.ruby
+      push: after-keyword
+    # flow
+    - match: \bbreak\b(?![?!])
+      scope: keyword.control.flow.break.ruby
+    - match: \bnext\b(?![?!])
+      scope: keyword.control.flow.next.ruby
+    - match: \bredo\b(?![?!])
+      scope: keyword.control.flow.redo.ruby
+    - match: \bretry\b(?![?!])
+      scope: keyword.control.flow.retry.ruby
+    - match: \breturn\b(?![?!])
+      scope: keyword.control.flow.return.ruby
+    - match: \bsuper(?![?!])
+      scope: keyword.control.flow.super.ruby
+    - match: \byield\b(?![?!])
+      scope: keyword.control.flow.yield.ruby
+    # operators
+    - match: \b(?:and|not|or)\b
       scope: keyword.operator.logical.ruby
       push: after-keyword
-    - match: '\b(alias|alias_method|break|next|redo|retry|return|super|undef|yield)\b(?![?!])|\bdefined\?|\bblock_given\?'
-      scope: keyword.control.pseudo-method.ruby
+    - match: \bin\b(?![?!])
+      scope: keyword.operator.logical.ruby
+      push: after-keyword
+    # other
+    - match: \b(?:BEGIN|END)\b(?![?!])
+      scope: entity.name.function.prepocessor.ruby
+    - match: \b(?:(?:alias|alias_method|undef)\b(?![?!])|(defined|block_given)\?)
+      scope: support.function.builtin.ruby
 
   operators:
     - match: '>>'
@@ -477,11 +533,30 @@ contexts:
       scope: variable.other.constant.ruby
 
   well-known-methods:
-    - match: '\b(initialize|new|loop|include|extend|prepend|raise|fail|attr_reader|attr_writer|attr_accessor|attr|catch|throw|module_function|public|protected|private)\b(?![?!])'
+    # exceptions
+    - match: \bcatch\b(?![?!])
+      scope: keyword.control.exception.catch.ruby
+      push: function-call-arguments
+    # flow
+    - match: \braise\b(?![?!])
+      scope: keyword.control.flow.raise.ruby
+      push: function-call-arguments
+    - match: \bfail\b(?![?!])
+      scope: keyword.control.flow.fail.ruby
+      push: function-call-arguments
+    - match: \bthrow\b(?![?!])
+      scope: keyword.control.flow.throw.ruby
+      push: function-call-arguments
+    # loop
+    - match: \bloop\b(?![?!])
+      scope: keyword.control.loop.loop.ruby
+      push: function-call-arguments
+    # other
+    - match: \b(initialize|new|include|extend|prepend|attr_reader|attr_writer|attr_accessor|attr|module_function|public|protected|private)\b(?![?!])
       scope: keyword.other.special-method.ruby
       push: function-call-arguments
     - match: \b(require|require_relative|gem)\b
-      scope: keyword.other.special-method.ruby
+      scope: keyword.control.import.ruby
       push:
         - meta_scope: meta.require.ruby
         - match: $|(?=[#}])
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -408,6 +408,7 @@ contexts:
       push: after-keyword
     - match: \bsuper\b(?![?!])
       scope: support.function.builtin.ruby
+      push: after-keyword
 
   operators:
     - match: '>>'
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -117,10 +117,10 @@ contexts:
 
   class:
     # Defining a class method
-    - match: \b(class)\b(?=\s*<)
-      scope: keyword.control.class.ruby
-    - match: '\b(class)\b'
-      scope: meta.class.ruby keyword.control.class.ruby
+    - match: \bclass\b(?=\s*<)
+      scope: storage.type.class.ruby
+    - match: \bclass\b
+      scope: meta.class.ruby storage.type.class.ruby
       push: class-declaration
 
   class-declaration:
@@ -332,7 +332,7 @@ contexts:
   keywords:
     - match: '\b(BEGIN|END)\b(?![?!])'
       scope: keyword.control.ruby
-    - match: '\b(class|module)\b(?![?!])'
+    - match: '\b(module)\b(?![?!])'
       scope: keyword.control.ruby
     - match: '\b(begin|end|ensure|rescue)\b(?![?!])'
       scope: keyword.control.ruby
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -384,7 +384,7 @@ contexts:
       scope: keyword.control.flow.retry.ruby
     - match: \breturn\b(?![?!])
       scope: keyword.control.flow.return.ruby
-    - match: \bsuper(?![?!])
+    - match: \bsuper\b(?![?!])
       scope: keyword.control.flow.super.ruby
     - match: \byield\b(?![?!])
       scope: keyword.control.flow.yield.ruby
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -740,8 +740,8 @@ contexts:
     - include: try-regex
 
   method:
-    - match: \b(def)\b
-      scope: meta.function.ruby keyword.control.def.ruby
+    - match: \bdef\b
+      scope: meta.function.ruby support.type.function.ruby
       push:
         - meta_content_scope: meta.function.ruby
         - match: '(self)(\.)({{identifier}}{{method_punctuation}}|===?|>[>=]?|<=>|<[<=]?|[%&`/\|]|\*\*?|=?~|[-+]@?|\[\]=?)'
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -546,13 +546,7 @@ contexts:
       scope: keyword.control.exception.catch.ruby
       push: function-call-arguments
     # flow
-    - match: \braise\b(?![?!])
-      scope: keyword.control.flow.raise.ruby
-      push: function-call-arguments
-    - match: \bfail\b(?![?!])
-      scope: keyword.control.flow.fail.ruby
-      push: function-call-arguments
-    - match: \bthrow\b(?![?!])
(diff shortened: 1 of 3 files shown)


Ruby/Ruby.sublime-syntax
```diff
@@ -117,10 +117,10 @@ contexts:
...
   class:
     # Defining a class method
-    - match: \b(class)\b(?=\s*<)
-      scope: keyword.control.class.ruby
-    - match: '\b(class)\b'
-      scope: meta.class.ruby keyword.control.class.ruby
+    - match: \bclass\b(?=\s*<)
+      scope: storage.type.class.ruby
+    - match: \bclass\b
+      scope: meta.class.ruby storage.type.class.ruby
```
FichteFoll: Via https://github.com/sublimehq/Packages/issues/582#issuecomment-453231444,
we decided on `keyword.declaration` for these types of keyword. `storage.type.class keyword.declaration` should be used in the transition period.

Ruby/Ruby.sublime-syntax
```diff
@@ -117,10 +117,10 @@ contexts:
 
   class:
     # Defining a class method
-    - match: \b(class)\b(?=\s*<)
-      scope: keyword.control.class.ruby
-    - match: '\b(class)\b'
-      scope: meta.class.ruby keyword.control.class.ruby
+    - match: \bclass\b(?=\s*<)
+      scope: storage.type.class.ruby
```
FichteFoll: If I'm correct and this depicts a class method (idk Ruby), then shouldn't this be `storage.modifier`?

Ruby/Ruby.sublime-syntax
```diff
@@ -157,8 +157,8 @@ contexts:
       pop: true
 
   module:
-    - match: \b(module)\b
-      scope: meta.module.ruby keyword.control.module.ruby
+    - match: \bmodule\b
+      scope: meta.module.ruby storage.type.module.ruby
```
FichteFoll: `keyword.declaration.module` imo.

Ruby/Ruby.sublime-syntax
```diff
@@ -289,9 +289,8 @@ contexts:
         2: invalid.illegal.character.ruby
 
   blocks:
-    - match: \b(do)\b\s*
-      captures:
-        1: keyword.control.start-block.ruby
+    - match: \bdo\b
+      scope: keyword.control.block.do.ruby
```
FichteFoll: Hm, what about using `keyword.control.context` from #1228 here?

Ruby/Ruby.sublime-syntax
```diff
@@ -330,22 +329,77 @@ contexts:
...
-    - match: '\b(case|else|ensure|for|in|then)\b(?![?!])'
-      scope: keyword.control.ruby
-    - match: '\b(elsif|if|unless|when|while|until)\b(?![?!])'
-      scope: keyword.control.ruby
+    # blocks
+    - match: \bbegin\b(?![?!])
+      scope: keyword.control.block.begin.ruby
+      push: after-keyword
+    - match: \bend\b(?![?!])
+      scope: keyword.control.block.end.ruby
```
FichteFoll: Same as with `do`.

Ruby/Ruby.sublime-syntax
```diff
@@ -330,22 +329,77 @@ contexts:
...
+    - match: \bnext\b(?![?!])
+      scope: keyword.control.flow.next.ruby
+    - match: \bredo\b(?![?!])
+      scope: keyword.control.flow.redo.ruby
+    - match: \bretry\b(?![?!])
+      scope: keyword.control.flow.retry.ruby
+    - match: \breturn\b(?![?!])
+      scope: keyword.control.flow.return.ruby
+    - match: \bsuper(?![?!])
+      scope: keyword.control.flow.super.ruby
```
FichteFoll: We haven't talked about `super` in #1228, but I guess this is an okay scope for how it appears to be used in Ruby (calls the overridden function with given parameters).

What do you think about C++ or Java?

Ruby/Ruby.sublime-syntax
```diff
@@ -330,22 +329,77 @@ contexts:
...
+      scope: keyword.operator.logical.ruby
+      push: after-keyword
+    - match: \bin\b(?![?!])
       scope: keyword.operator.logical.ruby
       push: after-keyword
-    - match: '\b(alias|alias_method|break|next|redo|retry|return|super|undef|yield)\b(?![?!])|\bdefined\?|\bblock_given\?'
-      scope: keyword.control.pseudo-method.ruby
+    # other
+    - match: \b(?:BEGIN|END)\b(?![?!])
+      scope: entity.name.function.prepocessor.ruby
```
FichteFoll: What's this?

Ruby/Ruby.sublime-syntax
```diff
@@ -479,11 +533,30 @@ contexts:
...
 
   well-known-methods:
-    - match: '\b(initialize|new|loop|include|extend|prepend|raise|fail|attr_reader|attr_writer|attr_accessor|attr|catch|throw|module_function|public|protected|private)\b(?![?!])'
+    # exceptions
+    - match: \bcatch\b(?![?!])
+      scope: keyword.control.exception.catch.ruby
+      push: function-call-arguments
+    # flow
+    - match: \braise\b(?![?!])
+      scope: keyword.control.flow.raise.ruby
```
FichteFoll: `raise` and `throw` should be unified to a single name because they behave in the same or at a least similar way. In #1228 I proposed `keyword.control.throw` because of the number of C++-like languages.
I see that Ruby has a `throw` keywor as well. How do they relate?

(Unless you're saying we should always use the keyword's name at the fourth subscope position, which I'd like to discuss first.)

Ruby/Ruby.sublime-syntax
```diff
@@ -479,11 +533,30 @@ contexts:
...
+      push: function-call-arguments
+    - match: \bfail\b(?![?!])
+      scope: keyword.control.flow.fail.ruby
```
FichteFoll: Does this behave like `die` or `panic` in other languages?

Ruby/Ruby.sublime-syntax
```diff
@@ -742,8 +815,8 @@ contexts:
     - include: try-regex
 
   method:
-    - match: \b(def)\b
-      scope: meta.function.ruby keyword.control.def.ruby
+    - match: \bdef\b
+      scope: meta.function.ruby support.type.function.ruby
```
FichteFoll: `keyword.declaration.function`

Ruby/Ruby.sublime-syntax
deathaxe: Except D all syntaxes use `storage.type.class`. As stacking those keywords doesn't have any value, I prefer to stay with the majority of syntaxes which use `storage` only.

Ruby/Ruby.sublime-syntax
deathaxe: I was thinking about `storage.type.namespace` as a module can be compared to a namespace, I think, but in general most syntaxes use `storage.type.namespace` for this kind of syntax, and so I perefer here.

Ruby/Ruby.sublime-syntax
deathaxe: possible.

Ruby/Ruby.sublime-syntax
deathaxe: possible

Ruby/Ruby.sublime-syntax
deathaxe: I am not sure about `super` actually. It sounds like a special keyword to tell ruby to call the ancestor method. In other languages it's sometimes more like a special variable compared to `self` or `this`. Haven't investigated all the different languages so far.

Ruby/Ruby.sublime-syntax
deathaxe: This is a feature borrowed from Perl, I think. It is a feature to hack the preprocessor/import of an module. In perl it can be used to manipulate the lookup path of imports or the whole environment. 

The scope is borrowed from:

https://github.com/sublimehq/Packages/blob/f60f80afdb66e317c3f2a0852861171ab677621f/Perl/Perl.sublime-syntax#L461-L465

Ruby/Ruby.sublime-syntax
deathaxe: Both keywords are explicitly added to the scope names in the following languages for instance.

`keyword.control.throw` is not used anywhere. At least `throw` should be part of the `keyword.control.flow` scope.

Java

https://github.com/sublimehq/Packages/blob/f60f80afdb66e317c3f2a0852861171ab677621f/Java/Java.sublime-syntax#L629

Python

https://github.com/sublimehq/Packages/blob/f60f80afdb66e317c3f2a0852861171ab677621f/Python/Python.sublime-syntax#L86

Ruby/Ruby.sublime-syntax
deathaxe: I don't know. I just followed the rule to add such kind of flow control keywords to the group of `keyword.control.flow`

Ruby/Ruby.sublime-syntax
deathaxe: @wbond suggested `storage.type.function` for `sub` in Perl and `def` is scoped exactly like that in Python. Many other syntaxes use this scope for general purpose function declaration keywords and it matches the scope names for `namespace`, `module`, `class`, ... . Therefore it is done here as well.

Ruby/Ruby.sublime-syntax
deathaxe: I found examples the following example, which says something about to use that notation `class << self` just to avoid repetition of the class name in the method declaration.

Not sure whehter it should be scoped differently in such a situation.

```
class TestClass
  # bad
  def TestClass.some_method
    # body omitted
  end

  # good
  def self.some_other_method
    # body omitted
  end

  # Also possible and convenient when you
  # have to define many class methods.
  class << self
    def first_method
      # body omitted
    end

    def second_method_etc
      # body omitted
    end
  end
end
```

Ruby/Ruby.sublime-syntax
deathaxe: The question is just whether we want to create some kind of relationship between `punctuation.section.block` and keywords which basically do the same - opening/closing blocks of code. If not, `context` is very fine as well here.

Ruby/Ruby.sublime-syntax
deathaxe: https://ruby-doc.org/docs/ruby-doc-bundle/Manual/man-1.4/syntax.html#BEGIN_proc

deathaxe: ... sorry for being too lazy to create a dedicated context for them as in Perl at this point.

Ruby/Ruby.sublime-syntax
FichteFoll: So your argument is you don't want to use the new scope because the old one is still in use by other syntaxes?

Ruby/Ruby.sublime-syntax
FichteFoll: Okay, that looks *weird*, to say the least. Looks more like a `keyword.control.context` than a class declaration, though.

Ruby/Ruby.sublime-syntax
FichteFoll: Same argument as for classes.

Ruby/Ruby.sublime-syntax
FichteFoll: Function calls *are* control flow, but I guess this most closely relates to `yield` (and maybe `async`). The only other scope I can think of right now would be something in `keyword.other`, which isn't ideal. No strong opinion.

Ruby/Ruby.sublime-syntax
FichteFoll: Thanks for the reference. The scope name is fine, then.

Ruby/Ruby.sublime-syntax
FichteFoll: Ah, yes, that was a typo. I meant to write `keyword.control.flow.throw`.

I'm just worried about us discussing scope names and eventually leading to a conclusion somewhat but then not acting on it when given the chance.

Ruby/Ruby.sublime-syntax
FichteFoll: See my other comments.

Ruby/Ruby.sublime-syntax
deathaxe: Yep. But I am unsure about using different scopes for the same keyword.

Ruby/Ruby.sublime-syntax
deathaxe: Possible due to their special meaning. But basically, they are ordinary functions.

Ruby/Ruby.sublime-syntax
deathaxe: We came arround with adding dedicated scopes for `die`, `panic`, `exit`, ... even though they basically have the same meaning in the one way or the other. But with `raise` and `throw` we argue to use the same 4th level scopes which might propably never be addressed by color schemes at this depth.

That's just weird. It just feels more natural to keep the first 3 levels balanced and use the keyword as 4th level for those who really want to use different scopes for each keyword.

I think its worth to keep something like `elif` or `elsif` or `elseif` with the same scope.

Ruby/Ruby.sublime-syntax
FichteFoll: Using different scopes for the same keyword is no problem for the `in` operator in Python, for example. We prioritize semantics over syntax anyway and the alternative scope would still be a `keyword`.

Ruby/Ruby.sublime-syntax
deathaxe: As noone haven't made any efforts to change it, yes.

Ruby/Ruby.sublime-syntax
FichteFoll: I don't remember us deciding on unsing different 4th level scopes for `die` or `panic`. My proposal (https://github.com/sublimehq/Packages/issues/1228#issuecomment-461077685) only includes `panic` as well. Maybe you misunderstood something?

Ruby/Ruby.sublime-syntax
FichteFoll: :man_shrugging: I'll note down to do a large PR changing various scope names to what we decided on in the past. In fact, I already took notes for preparing proper standards tracking for #1440. Hopefully I can manage this within the current dev cycle.

Ruby/Ruby.sublime-syntax
wbond: I would probably argue the other way. Most users are going to find it weird to have the same keyword scoped differently.

Ruby/Ruby.sublime-syntax
deathaxe: > ... I don't remember us deciding on using different 4th level scopes ...

Oh. So I missed something. Keywords were a while ago.

> The actual keyword could then be added in the 5th level.

1. The distinction between "meaning" and "keyword" is sometimes hared to see.
2. Not sure if scoping "each keyword" with its dedicated scope is really useful. A reasonable number of groups to target different types of keywords is propably enough.

Ruby/Ruby.sublime-syntax
deathaxe: That's my concern, too.

Ruby/Ruby.sublime-syntax
wbond: @deathaxe Multiple contributors have gotten hung up on using `storage.type.…` for declaration keywords in the past. Part of the hang up is that many color schemes don't specify past `storage.type`, so `int` looks the same as `def`, and some languages will have both a keyword and a return type (if I recall correctly).

In an attempt to make a forward-looking change, but preserve not breaking all existing color schemes, the standard approach for now (which will be reflect in the naming docs when I do the next update) is to double up with `storage.type keyword.declaration`.

Ruby/Ruby.sublime-syntax
FichteFoll: I suggest we discuss that in #1228 separately.

Ruby/Ruby.sublime-syntax
FichteFoll: Are you sure about that? `in` is imo a good example,
because it behaves *entirely* different.

```py
for a in b:
    if 1 in a:
        pass
```

Unless you didn't mean to argue for the general case but about `class` in particular here.

deathaxe: _Off Topic: I saw several languages scoping `in` as `keyword.operator`. While `keyword.operator.logical` makes sense I am not sure about the use case in the `for` statement._

In `class << self` the `class` keyword has the same meaning as `with` or `using`. It would then need to be scoped as `keyword.control.[block|context]`.

_Sorry, but I still don't feel comfortable with `keyword.control.[exception|resource]` as discussed in #1228 because the former alternatives seem to be a more general approach, which can be applied to more situations/languages without sematical conflicts (see python or the keywords below)._

djspiewak: I generally am pretty happy to scope the same symbol differently in different contexts. A great example of this is `=>` in Scala:

```scala
val f: Int => String = i => i.toString
//         ^^ keyword.operator.arrow
//                       ^^ storage.type.function.arrow.lambda
```

I think it's not at all unprecedented.

Ruby is complicated and weird because it has things that are functions but behave like keywords, things that are keywords that behave like functions, and things that are neither that behave differently in different syntactic contexts. In general, I would say that the solution to the *specific* problems raised here (`class` and `in`) should be to scope them differently in the different contexts, biasing toward the differing semantics rather than toward the identical characters.

FichteFoll: `keyword.control.[exception|resource]` has been replaced by `keyword.control.context.[exception|resource]` (or, ideally, `keyword.context.[exception|resource]`). I just haven't made a definitive statement on that yet because I was collecting input on a new second-level scope.

Ruby/Ruby.sublime-syntax
djspiewak: >  sounds like a special keyword to tell ruby to call the ancestor method

That is exactly what it is. Or rather, it's a special *function* which points to the ancestor method. I would actually *almost* consider scoping it as `support.function`. I don't think `keyword.control.flow` is appropriate. `keyword.other.super` is not terrible. @FichteFoll is right that it's similar to Python's `yield`.

Ruby/Ruby.sublime-syntax
FichteFoll: @djspiewak, you should  add `keyword.declaration` to the second arrow, via https://github.com/sublimehq/Packages/pull/2156#discussion_r339352039.

Ruby/Ruby.sublime-syntax
deathaxe: With keywords being used very often, I find `support.function` a good choice here. It also makes clear a function is called.

deathaxe: ... but the ruby docs name [super](https://ruby-doc.org/docs/keywords/1.9/files/keywords_rb.src/M000034.html) as keyword.

deathaxe: A bit confusing is to compare `super` with `yield` which is a `keyword.control.flow` but then argue to not scope `super` as such.

Ruby/Ruby.sublime-syntax
themilkman: Sorry for jumping in (just stumbled in here), but if the question of @FichteFoll was related to the `fail` keyword:
It is a synonym for `raise`, which is like "throw Exception" in other languages.

Ruby/Ruby.sublime-syntax
FichteFoll: Keywords are special because they have a high priority in syntax parsing and can't be used as identifiers. Still, `super` and `this`, in some languages, just refer to different *things* depending on the context. 

- In Java and C#, `super` kind of behaves like `this` and you use it as `super.printMethod();` or just `super()` for the constructor. 
- In Ruby, it behaves like a magic function that resolves to the appropriate function. 
- In Python, it is an actual function that does these two things combined.

Afaik, we generally scope `this` as `variable.language` based on TM's guidelines, despite it being a keyword. I don't know if this is the best choice, but it's what we have currently, at least. If this is fine, then we could use `support.function.{language,super}` **if** it resembles a function call, like it does here. For Java and C# it would then be scoped `variable.language`. 

We could assign these scopes on top of `keyword.other` to represent the fact that they *are* keywords.

---

I don't think the comparison with `yield` was a good one. The only behavior they share is that they exit the current context at the position of the keyword and give a re-entry point at the same position. Ruby's `super` behaves more like a function call. Either way, it's not `keyword.control.flow`

Ruby/Ruby.sublime-syntax
deathaxe: The `die` or `panic` functions in Perl are nothing else. They just cause the process to die just like `exit` in some other syntaxes. The point just is ... Which keywords to treat as synonymes and which not. 

The same question arises with the `redo`, `next`, `retry` keywords above. All are some kind of `continue` synonym but with little different meanding?

themilkman: Ah okay. Well I feel like `raise` should be the original and `fail` the synonym as the latter is much less common.
Whereas `next` and `retry` for example are not synonyms for me (except from the point of syntax color... which probably is the point here).

Ruby/Ruby.sublime-syntax
deathaxe: I think we can follow Python and use `support.function.builtin` here. As keywords are already used very often and with regards to `this` and `self`, I don't think we need to stack it with `keyword.other`.

wbond:
So the general consensus is that this is good to merge?

deathaxe:
The two questions which we might need some feedback for are:

1. What do you think about the `keyword.control.block` as a general scope for the keywords `do`, `begin` and `end`? We also discussed `keyword.control.context` or @FichteFoll even suggested something within a new 2nd level scope `keyword.context`. There are some comments about it at #1228

2. The question about the scope for `class <<` is still open. I currently decided to use `storage.type.class keyword.declaration.class`, but from the sematic aspect a scope from 1.) would be appropriate as well.

Ruby/Ruby.sublime-syntax
FichteFoll: >As keywords are already used very often and with regards to this and self, I don't think we need to stack it with keyword.other.

Could you explain this in more detail? I don't see how keywords frequently being used in a language would be related to not using a `keyword` scope for a keyword.

deathaxe: see commit message a11cf6d

FichteFoll: Well, let's tackle this at another time.

FichteFoll: Well, let's tackle this at another time.

FichteFoll:
We might also want to consider (also in the context of #1228) whether `die` and `panic` do the same and should be unified to a subscope of `keyword.control.flow` instead of using `keyword.control.flow.<keyword-name>`. But that's not as significant and can be talked about later.

wbond:
@deathaxe I like `keyword.control.block` for 1. For 2 `keyword.control.context` seems most appropriate since I am familiar with the mechanics of Ruby. That said, I would also be okay with `storage.type.class keyword.declaration.class`.

@FichteFoll If I had to pick, I'd do something like `keyword.control.flow.termination`.

wbond:
On second thought, I think `storage.type.class keyword.declaration.class` is much more correct. While `class` is being passed a block, it is really re-opening the definition and using the block to append the definition.

deathaxe:
So everything would be ready to merge then, if `keyword.control.flow.termination` is addressed to `die`, `panic`, etc.

I've choosen `keyword.control.flow.throw` for `throw`, `raise` and `fail` here.

wbond:
Thanks @deathaxe and everyone who spent time reviewing!

PR #2158, opened 2019-10-28
[Git Formats] Rename the remaining hash constants

deathaxe:
This commit renames the remaining scopes to `constant.other.hash` as it was introduced by commit 71aeb9ad2812bbd4bc8cf6e139e963d7a37b7dd8.


FichteFoll:
I totally forgot there were multiple syntaxes when I made this change. 😅

PR #2159, opened 2019-10-29
[JavaScript] Improve numbers

deathaxe:
This PR

1. uses `storage.type.numeric` scope for numeric literal suffixes.
2. adds more detailed `constant.numeric.[integer|float]` scopes.

Note:

1. The 2nd change required to change the pattern for floating point numbers and its order in the context in order to correctly distinguish between integers and floats.
2. The 1st rule being used to match illegal octal numbers is removed as its function is handled by the 2nd illegal octal rule.
3. The changes don't have an impact on parsing performance.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -12,8 +12,8 @@ variables:
   oct_digit: '[0-7_]'
   dec_digit: '[0-9_]'
   hex_digit: '[\h_]'
-  integer: (?:0|[1-9]{{dec_digit}}*)
-  exponent: (?:[Ee](?:[-+]|(?![-+])){{dec_digit}}*)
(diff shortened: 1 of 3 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -8,6 +8,10 @@ file_extensions:
 first_line_match: ^#!\s*/.*\b(node|js)\b
 scope: source.js
 variables:
+  ddigits: (?:[0-9_]+)
+  ddigits0: (?:0|[1-9][0-9_]*)
```
Thom1729: The variable names might be a bit unclear.

deathaxe: They are borrowed from the other syntaxes which have been recently tweaked.

PR #2166, opened 2019-10-31
[Erlang] Improve numbers

deathaxe:
This commit

1. removes the punctuation scope from the `e` exponent character
2. scopes the whole base part of integers as `punctuation.definition.numeric.base`.

_Note:_

When I saw the `punctuation.definition.numeric.base` scope I wondered whether it was a good general scope name for all kinds of numeric prefixes like `0x`, 0b`, etc.

Most syntaxes use specialized scopes like `punctuation.definition.numeric.hexadecimal`.

Thoughts?


wbond:
I think the punctuation base scope seems like a good fit to me.

PR #2171, opened 2019-11-01
[YAML] Improve numbers

deathaxe:
This commit adds some capturing groups to the existing numeric literal match patterns in order to distinguish the types (int/float/...) and to add scopes for the separators.

### DIFF
--- a/YAML/tests/syntax_test_block.yaml
+++ b/YAML/tests/syntax_test_block.yaml
@@ -18,11 +18,11 @@
 
 - >1
 # ^ keyword.control.flow.block-scalar
-#  ^ constant.numeric.indentation-indicator
+#  ^ constant.numeric.integer.decimal.indentation-indicator
 
 - |1-
 # ^ keyword.control.flow.block-scalar
-#  ^ constant.numeric.indentation-indicator
+#  ^ constant.numeric.integer.decimal.indentation-indicator
 #   ^ storage.modifier.chomping-indicator
 
 # Headers and content ################
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -134,7 +134,7 @@ variables:
     )
   _type_timestamp: |- # http://yaml.org/type/timestamp.html
     (?x:
-        \d{4} (-) \d{2} (-) \d{2}        # (y-m-d)
+        \d{4} (-) \d{2} (-) \d{2}       # (y-m-d)
       | \d{4}                           # (year)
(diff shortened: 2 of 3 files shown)


Thom1729:
That is a lot of capture groups. Would it make more sense to split those mega-matches up into multiple rules?

YAML/tests/syntax_test_block.yaml
```diff
@@ -18,11 +18,11 @@
 
 - >1
 # ^ keyword.control.flow.block-scalar
-#  ^ constant.numeric.indentation-indicator
+#  ^ constant.numeric.integer.decimal.indentation-indicator
```
wbond: Can you expand upon why this punctuation is part of the integer literal?

It is probably just that even after reading the YAML spec and this, I have no idea what the indentation-indicator does/is.

deathaxe: Actually, I didn't see it before as well. We might ask @FichteFoll for it.

It just intents to follow _Bash_ which scopes numeric file descriptors like `constant.numeric.integer.decimal.file-descriptor.shell`.

... just to ensure it gets highlighted as integer if someone will ever address `constant.numeric.integer.decimal` in a color scheme.

Otherwise I would suggest something like `constant.other` or `constant.language` as we did with the hash in the git syntaxes.

wbond: Oh, sorry, I guess I misread what was getting the `indentation-indicator`. I now see it only applies to the `1` and the `indentation-indicator` is just a specialization. I thought it was applying to the `>`.

Thom1729: `>` introduces a folded block scalar. The `1` controls the indentation of the content; it isn't part of the content. (This particular block scalar doesn't actually have any content; it ends at the beginning of the next line when it sees the `#`.)

deathaxe:
@Thom1729  I didn't want to change the overall structure of the syntax as the variables being used reflect the specs pretty well.

The number of capture groups is not performance critical. Once you match some it doesn't matter whether there are 5 or 20. The only critical thing was being able to count from 1 ... 30 ;-)

All splitted rules would need to be implemented twice with a lot of more lines of code. Not sure whether it is worth to do so.

FichteFoll:
We could probably prevent duplicating all these captures with a look-ahead, but let's do so in a separate PR. For now, this is fine.

deathaxe:
I always try to avoid lookaheads where possible, just for performance reasons.

PR #2175, opened 2019-11-01
[Python] punctuation.definition.numeric.base

FichteFoll:
Via https://github.com/sublimehq/Packages/pull/2166.


jwortmann:
Change `punctuation.definition.integer.octal.python` to `punctuation.definition.numeric.base.python` in [line 443](https://github.com/sublimehq/Packages/pull/2175/files#diff-62db422a04a96e067da114ba6ed7d6e9R440-R443) as well?

PR #2183, opened 2019-11-06
Ruby rescue block, idiomatic Snippet

mark-d-holmberg:
* Don't trap Exception by default, but rather StandardError

If you rescue from Exception, you're unintentionally trapping Signals from the system like SIGQUIT


FichteFoll:
Also, the description should remain as a single line (and not be changed) because that's what is shown in the autocomplete popup on the right side to give a hint on the contents.

mark-d-holmberg:
Thanks. I'll make the changes you request and provide you with an updated PR.

mark-d-holmberg:
That should do it. I'd suggest squash and merge if you end up approving my changes. Thanks!

PR #2185, opened 2019-11-07
[C/C++/ObjC/ObjC++] Improve numbers

deathaxe:
This commit...

1. applies scope naming guidelines to all numeric literals
2. implements dedicated C style and C++ style contexts with the syntax specific behavior.

   a) C/ObjC supports only limited amount of predefined suffixes C++/ObjC++ supports all kind of suffixes which are either predefined (see C) or user defined via `operator ""` keyword.
   b) C++14 introduces `'` as digit to improve readability of large numbers, while C does not support it.
   c) both C and C++ support hexadecimal floats

   Dedicated means, C++ does no longer import C number contexts as the required patterns are simply differ too much.

3. adds illegal matching of octal or hexadecimal digits as well as unsupported suffixes (in C). The implementation is inspired by Bash.sublime-syntax and complies with the compiler error messages.

4. fixes an issue with decimal points being matched as accessor in some situations.

5. Depending on the mix of numbers the parsing performance is improved by 4 ... 40% especially in C++.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -20,6 +20,21 @@ file_extensions:
 first_line_match: '-\*- C\+\+ -\*-'
 scope: source.c++
 variables:
+  # number digits
+  dec_digits: '(?:\d(?:[\d'']*\d)?)'
+
+  # number exponents
+  dec_exponent: '(?:[eE][-+]??{{dec_digits}})'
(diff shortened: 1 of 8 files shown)


deathaxe:
Oversaw a couple of `constant.numeric.preprocessor` in Objective-C/C++ which needed to be fixed.

PR #2188, opened 2019-11-08
[JSON] Improve numbers

deathaxe:
This commit applies the numeric literal scope naming guidelines.


jrappen:
The regexes might have been easier to read for beginners before this change. Whatever works 🤷🏼‍♂️

wbond:
@jrappen That is probably true, however it may have been a little over-the-top before.

This seems to be an improvement, so I'm going to merge.

FichteFoll:
Comments should be used to convey extra information that can not trivially be inferred from the code (or expression). The comments in the old regex did not do that.

PR #2189, opened 2019-11-09
[CSS] Add property value: fit-content

jfcherng:
https://caniuse.com/#search=fit-content


deathaxe:
Also need the completions: https://github.com/jfcherng/Sublime-Official-Packages/pull/2

PR #2191, opened 2019-11-09
[CSS] Add completions for CSS colors

deathaxe:
This commit adds the predefined color names to the css_completions.py and makes sure to match the names case-insensitive in the syntax as there are suggestions out there which propose camelCase names.

see: 
https://www.quackit.com/css/css_color_codes.cfm
https://www.w3schools.com/cssref/css_colors.asp


jrappen:
maybe add tests?

CSS/css_completions.py
```diff
@@ -36,7 +36,44 @@
     'break_inside': [
         'auto', 'avoid', 'avoid-page', 'avoid-column', 'avoid-region'
     ],
-    'color': ['currentColor', 'rgb($1)', 'rgba($1)', 'hsl($1)', 'hsla($1)', 'transparent'],
+    'color': [
+        'rgb($1)', 'rgba($1)', 'hsl($1)', 'hsla($1)', 'hwb($1)',
+        'blend($1)', 'blenda($1)', 'color($1)', 'shade($1)', 'tint($1)',
```
wbond: I believe these are part of the color module v4, which the last I checked was in fairly heavy flux. I know we implemented things like `blend()`, `blenda()`, etc in minihtml, but they've since decided to rip a whole bunch of that out since they wouldn't do what people expected since they defaulted to the RGB color space instead of Lab.

deathaxe:
@jrappen: This commit doesn't add new features to the syntax except just making existing color value patterns case-insensitive. 

@wbond This commit just copies existing color functions already defined in the syntax. According to your comment, it means to remove them from the syntax as well?

deathaxe:
We propably shouldn't delete functions from the syntax definition which are used in the minihtml even though they are not officially supported. The CSS syntax is included in PackageDev's Sublime Color Scheme syntax definition.

FichteFoll:
Imo the standard definition should respect the official standards of a language and only provide additions where they are reasonably needed. I'd be fine with adding support for the scraped CSS4 color functions within PD itself.

deathaxe:
As this PR is primarily about completions - the color functions were removed - any changes to the syntax definition should be part of a separate PR.

wbond:
I refactored this on top of the current state of `css_completions.py`

PR #2198, opened 2019-11-11
[C++/ObjC++] Add C++20 coroutine keywords

rwols:
Coroutines are a new feature of C++20. The new keywords are

- co_await
- co_return
- co_yield

Any function with such a keyword in its body is considered a [coroutine](https://en.cppreference.com/w/cpp/language/coroutines).


wbond:
Thanks!

PR #2202, opened 2019-11-13
[PHP] Fix "fn" function call

jfcherng:
Congrats for ST 4 :beer: 

```php
$var = fn($x)
//     ^^ meta.function.arrow-function.php - meta.function-call
   => $x * 2;
// ^^ punctuation.definition.arrow-function

$var = fn($x)
//     ^^ meta.function-call - meta.function.arrow-function.php
;
```

- Before PHP 7.4, `fn` can be used as a function name.
- Since PHP 7.4, `fn` is a reserved word for arrow function.

[Previously](https://github.com/sublimehq/Packages/pull/1961/commits/c6210f7a1cfa62376f0aa9aed499e983b337e430), I assumed that `fn` is always a reserved word because of ST's parsing engine's limitation.

---

PHP 7.4.0 will just be released by the end of this month.

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -635,7 +635,18 @@ contexts:
 
   # https://wiki.php.net/rfc/arrow_functions_v2
   arrow-function:
-    - match: '(?i)\b(fn)\b\s*(&)?\s*(?=\()'
+    # "fn" can be a function name before PHP 7.4
+    # Since PHP 7.4, "fn" is a reserved word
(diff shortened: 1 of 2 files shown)


wbond:
Right now `syntax_tests` links to libGL, so we'll need to get that sorted (either fixing the dependency to be optional or getting travis to install libGL) before any PRs will pass.

wbond:
If you rebase on master that should allow the tests to pass

jfcherng:
OHHHHHHHH! YEEEEEEEEE!

PR #2205, opened 2019-11-15
Try fixing Travis CI

FichteFoll:


FichteFoll:
The current tests [fail](https://travis-ci.org/sublimehq/Packages/builds/611507946#L199) because of
```
./syntax_tests: error while loading shared libraries: libGLU.so.1: cannot open shared object file: No such file or directory
```

A [search on Ubuntu's package repository](https://packages.ubuntu.com/search?keywords=libglu&searchon=all&suite=all&section=all) lead me to believe that the new build requires the `libglu1-mesa` package, but it seems like that's not available for tavis.

Maybe switching to Github Actions might be worth exploring? I have no experience with that, however, and afaik only collaborators or owners can change that.

wbond:
There is already a branch with the Travis packages, but the executable was segfaulting. I had just quickly compiled it on Fedora 31, so there is probably a runtime diff. I’ll compile on Ubuntu 18.04 once I get my VM back up and running.

wbond:
I have been using GitHub Actions on some of my Python repositories, and it still has a little ways to go, IMO. There is no way to re-run individual jobs in a build matrix when one fails, trying to get some build info is a pain, the seem to quickly get rid of old language runtimes and OSes. Not that any of those affect this repo, but hopefully we'll see continued improvements for Actions.

Issue #2147, opened 2019-10-25
[Rust] Ligatures mis-grouped


Shadlock0133:
- Sublime Version: 3.2.2
- OS Version: Windows 8.1

``` rust
bitmask >>= 1;
```

I'm using a Fira Code font.

In Rust file:
![image](https://user-images.githubusercontent.com/3766527/67568083-7b8cf300-f72b-11e9-9f53-c135820446f9.png)

For comparison, same line in C file:
![image](https://user-images.githubusercontent.com/3766527/67568124-9a8b8500-f72b-11e9-8179-ee7dd9dcac94.png)

urschrei:
Does this problem also occur when using Rust-Enhanced? (https://github.com/rust-lang/rust-enhanced). If it does, you should open an issue there, as that's where the Rust syntax development happens.

wbond:
@urschrei Unfortunately, no, Rust Enhanced generally does not upstream on contribute bug reports or fixes here, even after I have asked if they would work with us. They don’t like waiting for new builds to get fixes out, which is understandable, but also prefer not to follow the scope naming guidelines, so the syntaxes have diverged.

Shadlock0133:
I tested it and the syntax file from Rust-Enhanced fixes the issue.

jrappen:
@wbond Fixed by @TheClams in #2208.

deathaxe:
Luckily #2305 doesn't break the fix by #2208.

PR #2208, opened 2019-11-16
[Rust] Fix scope for assignment operator

TheClams:
Simply swap comparison and assignement operator to be sure assignment get tested first.
Add relevant test.

Fixes issue #2147

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -247,9 +247,11 @@ contexts:
   operators:
     - match: \.{2,3}
       scope: keyword.operator.range.rust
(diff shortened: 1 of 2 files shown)


jwortmann:
I think these rules have to be splitted to ensure the right testing order. With the current change, `==` will be `keyword.operator.assignment.rust` instead of `keyword.operator.comparison.rust`, I believe.

TheClams:
Correct, thanks for the review

jrappen:
@TheClams Please use `Fixes #2208` or `Closes #2208` in commit messages. `Fixes issue #2208` doesn't auto-close that on merge of this PR.

PR #2211, opened 2019-11-22
[Java] Unicode and octal escapes in strings

lastsecondsave:
Extended rules for escapes to handle all types present in Java:

1. Unicode escape:
```
System.out.println("\u0078 \u0079 \u007A");
```

2. Octal escapes:
```
System.out.println("\170 \171 \172");
```

3. Simple escapes now limited to a set of allowed ones instead of matching `\\.`

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -1128,8 +1128,19 @@ contexts:
     - match: \n
       scope: invalid.illegal.newline.java
       pop: true
-    - match: \\.
+    - include: strings-escape
+
+  strings-escape:
+    - match: \\u+\h{4}
+      scope: constant.character.escape.unicode.java
+    - match: \\[0-3][0-7]{2}
+      scope: constant.character.escape.octal.java
+    - match: \\[0-7]{1,2}
+      scope: constant.character.escape.octal.java
+    - match: \\[btnfr"'\\]
       scope: constant.character.escape.java
+    - match: \\.
+      scope: invalid.illegal.escape.java
 
   module:
     - match: (?=\b(?:open\s+)?module\b)
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -1133,9 +1133,7 @@ contexts:
   strings-escape:
     - match: \\u+\h{4}
       scope: constant.character.escape.unicode.java
-    - match: \\[0-3][0-7]{2}
-      scope: constant.character.escape.octal.java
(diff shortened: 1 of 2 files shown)


Java/Java.sublime-syntax
```diff
@@ -1128,8 +1128,19 @@ contexts:
     - match: \n
       scope: invalid.illegal.newline.java
       pop: true
-    - match: \\.
+    - include: strings-escape
+
+  strings-escape:
+    - match: \\u+\h{4}
```
deathaxe: Java allows at most 2 leadig `u`. So this should look like `\\uu?\h{4}`

Java/Java.sublime-syntax
```diff
@@ -1128,8 +1128,19 @@ contexts:
...
+      scope: constant.character.escape.unicode.java
+    - match: \\[0-3][0-7]{2}
+      scope: constant.character.escape.octal.java
+    - match: \\[0-7]{1,2}
+      scope: constant.character.escape.octal.java
```
deathaxe: The octal escape rules might be merged to `\\[0-3]?[0-7][0-7]?`

Java/Java.sublime-syntax
```diff
@@ -1128,8 +1128,19 @@ contexts:
     - match: \n
       scope: invalid.illegal.newline.java
       pop: true
-    - match: \\.
+    - include: strings-escape
+
+  strings-escape:
```
deathaxe: Many contexts normally use plural (see: `keywords`, `constants-and-special-vars`, ...). So the context should propably named `string-escapes`.

Java/Java.sublime-syntax
lastsecondsave: Did not notice these 2 can be merged. Thanks for the tip!

Java/Java.sublime-syntax
lastsecondsave: I've matched it with `strings-common` but agree it's better to use plurals like everywhere else.

Java/Java.sublime-syntax
lastsecondsave: [The language specification](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.3) does not limit the amount of 'u' characters. The compiles also seems not to care about it.

deathaxe: You're right, didn't read the BNF carefully enough.

wbond:
Thanks for these fixes!

Issue #1942, opened 2019-04-16
[Regular Expression] Unified scope names for (embedded) Regular Expressions


jwortmann:
Various languages have inbuilt support for Regular Expressions and some default syntax packages provide rules to apply suitable scopes within RegExp strings. However, these scope names seem to differ in various places, for example the standalone RegExp syntax and Clojure use `keyword.operator.alternation.regexp` for the `|` symbol, while JavaScript, Python and PHP use `keyword.operator.or.regexp`. Another example are character classes such as `\d` or `\w`, which get the scope `keyword.control.character-class.regexp` in the standalone RegExp syntax, `constant.other.character-class.escape.backslash.regexp` in JavaScript and `constant.character.character-class.regexp` in Python and PHP. Other languages such as Tcl and Ruby recognize RegExp strings, but do not apply specific scopes other than `string.regexp`, which prevents syntax highlighting of Regular Expressions in these languages.

I want to refine my color scheme for consistent RegExp highlighting, but the currently used scope names make it difficult to find common highlighting rules for all languages. My knowledge of syntax definitions is somewhat limited, but as far as I know there is the possibility to embed a syntax within another language syntax (e.g. CSS in HTML). Maybe this could be applied to parse RegExp strings and allow consistent syntax highlighting of Regular Expressions in more languages.

Regular Expression syntax:
```regex
    (?<=(T|t)he\s)(cat)$
(?#  ^^^ constant.other.assertion )
(?#       ^ keyword.operator.alternation )
(?#            ^^ keyword.control.character-class )
(?#                    ^ keyword.control.anchors )
```

JavaScript syntax:
```js
var regex = /(?<=(T|t)he\s)(cat)$/;
//            ^^^ punctuation.definition.group.assertion
//                 ^ keyword.operator.or
//                      ^^ constant.other.character-class.escape.backslash
//                              ^ keyword.control.anchor
```

Python syntax:
```python
regex = r'(?<=(T|t)he\s)(cat)$'
#          ^^^ constant.other.assertion
#               ^ keyword.operator.or
#                    ^^ constant.character.character-class
#                            ^ keyword.control.anchor
```

Clojure syntax:
```clojure
#"(?<=(T|t)he\s)(cat)$"
;  ^^^ constant.other.assertion
;       ^ keyword.operator.alternation
;            ^^ keyword.control.character-class
;                    ^ keyword.control.anchors
```

deathaxe:
[...]

PR #2214, opened 2019-11-23
[Regex Replace] Add syntax

deathaxe:
This PR proposes to add the _Regex Replace.sublime-syntax_ to highlight the content of the replace input widget if the regexp mode is enabled.

The knowledge about the _Regex Replace Widget.sublime-settings_ was published by @keith-hall at Discord.

The syntax definition is sourced from PackageDev with some fixes and scope name modifications being applied. Scoping of conditional expressions and groups was removed as ST doesn't seem to support that.

Both file name and the `name` attribute use `Regex` as the widget setting is named like that. The scope name on the other hand uses `regexp` as this is what the general `Regular Expressions` syntax
defines.

A new syntax package is created to enable users to selectively disable it on demand.

If ST supports conditionals in the future, the following contexts could be added as well:

```YAML
  groups:
    - match: \(
      scope: punctuation.section.group.begin.regexp-replace
      push:
        - meta_scope: meta.group.regexp-replace
        - match: \)
          scope: punctuation.section.group.end.regexp-replace
          pop: true
        - include: main

  conditionals:
    - match: (\?)([1-9]|\{\w+\})
      captures:
        1: keyword.operator.conditional.regexp-replace
        2: variable.language.backref.regexp-replace
      push:
        - match: ':'
          scope: keyword.operator.alternation.regexp-replace
          pop: true
        - match: (?=\))
          pop: true
        - include: main
```


FichteFoll:
While technically a violation of PD's license, there isn't exactly much of value with that definition, so in case this ever becomes relevant, I hereby grant all the necessary rights for this "code" to be included and distributed by putting it under the public domain.

I was confused about conditionals not being supported at first, because I do know that ST uses boost, but with some testing they do indeed not seem to work. See https://github.com/SublimeTextIssues/Core/issues/1524.

deathaxe:
> While technically a violation of PD's license, ...

Seriously?

FichteFoll:
I mean, that's what licenses are for and I'm sure a business such as Sublime HQ is more interested in being legally secure than someone scraping together random snippets from Github.

michaelblyons:
Is there a reason this isn't nested under "Regular Expressions" instead of a new dir?

deathaxe:
> someone scraping together random snippets from Github.

With all the changes to the code I could have easily claimed it as a new invention or at least being extracted from Perl.sublime-syntax as well, which is part of this repo and lincensed to be published!

And honesty: Without a note from @keith-hall at discord about this syntax being part of PackageDev, I would even have. 

Finally it was just a try to make a nice feature available for a majority of users as I guess nearly noone knows about it at all. Only those handful of guys who live within SublimeTextIssus/Core. As discussed at Discord, the corresponding setting could have been added to PackageDev 3 years ago, already.

> Is there a reason this isn't nested under "Regular Expressions" instead of a new dir?

This is indeed a constructive question to discuss about as they are indeed related.

The only reason for this PR to suggest a dedicated package is to enable users to disable highlighting of the replace input by disabling the package.

With #1942 pending and the discussion going on in here in mind, I even suggest to make a new PR with a completely rewritten syntax based on Perl, with some general scope and syntax names .

What we have here is basically a _Perl Format String.sublime-syntax_ which is a tiny subset of Perl.sublime-syntax. With Perl.sublime-syntax using the _Regular Expressions.sublime-syntax_ already, it would also make sense to share this one.

michaelblyons:
> The only reason for this PR to suggest a dedicated package is to enable users to disable highlighting of the replace input by disabling the package.

It's true that a separate directory would allow you to disable it by just adding to `ignored_packages`, but I think other User settings could be used instead to do the disabling for anyone who wanted to, e.g. `Regex Replace Widget.sublime-settings`:

``` json
{
    "syntax": "Plain Text.tmLanguage"
}
```

FichteFoll:
>>someone scraping together random snippets from Github.
>With all the changes to the code I could have easily claimed it as a new invention or at least being extracted from Perl.sublime-syntax as well, which is part of this repo and lincensed to be published!

I didn't mean to offend you in anything I said. I did, however, struggle with finding the correct words to express what I mean, so I'll blame my poor English on this misunderstanding. Sorry for that.

What I meant to do was clear up potential license issues towards Sublime Hq, because you specifically mentioned using/basing this on the syntax from PackageDev, and to somewhat relativize it to private projects that technically have to fear the same concequences but are very unlikely to be targeted over a trivial issue such as this. While this could be interpreted as me harassing you for violating the license terms, I did not mean any harm.

---

>The only reason for this PR to suggest a dedicated package is to enable users to disable highlighting of the replace input by disabling the package.

The included regular expression syntax definition appears to be used primarily by the find panel and is based on boost's PCRE dialect. While other syntaxes can include it, they may find slight differences with their implementations, like with Python, and are likely better advised to implement their own flavor. Thus, I don't believe splitting these two very closely related definitions in origin and usage into two packages makes sense.

deathaxe:
> Sorry for that.

Me, too. Clearly overreacted.

deathaxe:
I'll merge it into the _Regular Expressions_ package.

The conflict with the file names `Regex` vs. `Regexp` drives me crazy though. Any suggestions  to resolve that conflict?

Interestingly the _Default.sublime-syntax_ of ST4055 contains 

- Regex Format Widget.sublime-settings
- Regex Replace Widget.sublime-settings
- Regex Widget.sublime-settings

with the latter one pointing to "Packages/Regular Expressions/RegExp.tmLanguage" .

I would at least expect it to use "RegExp.sublime-syntax".

More over, I wonder, if the Replace syntax should go the same way. Just adding the syntax here, while asking to add a reference to it within the _Regex Replace Widget.sublime-settings_ in the _Default.sublime-package_?

jwortmann:
I think it would make sense to let the scope names end with just `.regexp` or perhaps better `.regexp.replace` instead of `.regexp-replace`, because apparantly all other lanuage syntaxes use only the `.regexp` ending for scopes in regular expressions, and that way it could be reused by color schemes.

deathaxe:
I considered that, too, but actually the replacement is no `regexp` and thus should probably not be scoped as such.  Using `.regexp.replace` would indicate it to be a special form of a regular expresionn, but it's actually a `perl format string`.

jwortmann:
After looking at the files again, I saw that it indeed seems to be different from "normal" regular expressions. From the file names and discussion here I assumed it was a subset of Regex first.

michaelblyons:
This is a little silly, so feel free to disagree:

Is there a way to add another trivial syntax definition for the globbing in the "find in files" entry of the panel? I think all it needs is `**`, `*`, `?`, and `,` (unless someone is about to tell me about a feature I did not know---which is kind of the point of having the highlighting).

deathaxe:
You mean the "Where" field in the "find in files" panel?

This PR is about the "Replace" field only at the moment.

wbond:
Thanks for the discussion and your work on this @deathaxe!

PR #2215, opened 2019-11-24
[Java] Allow annotations in generics

lastsecondsave:
For situations like this:
```java
List<@NotBlank(message="Address must not be blank") String> addresses;
```


wbond:
Thank you!

PR #2217, opened 2019-11-24
[Go] Identify entity.name.namespace.go

piotrkubisa:
Add rule to interpret the `package {{ident}}` as a `entity.name.namespace.go`. Squash-able.

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -727,6 +727,14 @@ contexts:
   match-keyword-package:
     - match: \bpackage\b
       scope: keyword.other.package.go
+      push:
+        - match: \b_\b
+          scope: variable.language.blank.go
+
+        - match: '{{ident}}'
+          scope: entity.name.package.go
+
+        - include: pop-before-nonblank
 
   match-keyword-range:
     - match: \brange\b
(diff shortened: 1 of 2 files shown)


Go/Go.sublime-syntax
```diff
@@ -727,6 +727,14 @@ contexts:
   match-keyword-package:
     - match: \bpackage\b
       scope: keyword.other.package.go
+      push:
+        - match: \b_\b
+          scope: variable.language.blank.go
+
+        - match: '{{ident}}'
+          scope: entity.name.package.go
```
FichteFoll: You can pop on these matches.

deathaxe: Would we prefer the more general `entity.name.namespace.go`, wouldn't we?

piotrkubisa: Changed `entity.name.package.go` to `entity.name.namespace.go` and also added `pop: true` for the `match: '{{ident}}'`.

FichteFoll: I actually meant the `_` match as well, though I'm not sure what that is for. Why declare a package when you use a dump name? Is it for single-file executables?

piotrkubisa: Yeah, my purpose was to have easy-identifiable scope where package name starts.

piotrkubisa:
I suppose I don't need it anymore. I am closing this issue.

FichteFoll:
What do you mean you don't need it anymore? I'm sure other people writing Go with ST would appreciate it.

PR #2219, opened 2019-11-28
 [Python] Fix f-string ending punctuation scope name 

FichteFoll:
Reported at https://forum.sublimetext.com/t/python-f-block-string-scoping-bug/48027.


wbond:
Thanks!

PR #2220, opened 2019-12-01
[Git] Add new v2.25.0 pretty-formats

michaelblyons:
See
- git/git@c22f63c
- git/git@d37cfe3

### DIFF
--- a/Git Formats/Git Common.sublime-syntax
+++ b/Git Formats/Git Common.sublime-syntax
@@ -8,7 +8,7 @@ scope: text.git.common
 
 variables:
   comment_char: '[#;]'
-  pretty_formats_builtins: oneline|short|medium|fuller|full|email|raw
+  pretty_formats_builtins: oneline|short|medium|fuller|full|email|raw|reference
   pretty_formats_empty_value_modifiers: '[-+ ]?'
 
 contexts:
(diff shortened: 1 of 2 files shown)


jrappen:
@michaelblyons maybe update the test for gitconfig?

michaelblyons:
Okay. I found a different bug while making the tests. Fixed that, too.

Git Formats/Git Common.sublime-syntax
```diff
@@ -283,6 +283,7 @@ contexts:
     - include: pretty-formats-wrap-formatting
     - include: pretty-formats-string-formatting
     - include: pretty-formats-trailers
+    - match: \S
```
deathaxe: What is that `-match: \S` for? We stay in that context until the `(?!\S)` on line 238 is matched. Removing it doesn't make a difference in your test cases.

michaelblyons: Good catch! It was added intentionally, along with the Line 394 test. But the test must only have been broken while I was tinkering with it. I'll take that out.

Git Formats/syntax_test_git_config
```diff
@@ -377,14 +381,26 @@ stray-bracket]
...
+#                             ^^ constant.other.placeholder.pretty-formats.git
+    foo = log --pretty=tformat:%h
+#                      ^^^^^^^ keyword.declaration.pretty-formats.git
+#                             ^ keyword.operator.assignment.option.pretty-formats.git
+#                              ^^ constant.other.placeholder.pretty-formats.git
     foo = log --pretty=%a
 #                      ^^ - constant
-    foo = log --pretty=%aD
+    foo = log --pretty=%aDx%cD
 #                      ^^^ constant.other.placeholder.pretty-formats.git
```
jrappen: @michaelblyons I assume you should change this as well, when changing the previous line.

wbond:
Thanks!

PR #2223, opened 2019-12-02
[Matlab] Improve numbers

jwortmann:
* Add scope for base prefixes for [hexadecimal and binary number literals](https://www.mathworks.com/help/matlab/matlab_prog/specify-hexadecimal-and-binary-numbers.html), wich were introduced in R2019b
* Add scope for type suffixes for hexadecimal and binary number literals
* Add scope for decimal separator in floating point numbers
* Notice: all numbers without base prefix are stored as floating point by default in Matlab

### DIFF
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -434,9 +434,25 @@ contexts:
       comment: Not equal is written ~= not !=.
       scope: invalid.illegal.invalid-inequality.matlab
   number:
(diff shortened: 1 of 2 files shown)


jrappen:
Does Matlab not differentiate between integers and floats?

jwortmann:
I think every programming language differentiates between integers and floats, but if you write `123` in Matlab, it will be stored as a floating point number. If you want it to be stored as integer, you have to explicitely cast it to integer, e.g. `int32(123)`, see https://www.mathworks.com/help/matlab/matlab_prog/integers.html

wbond:
Great, thanks for these improvements!

PR #2225, opened 2019-12-03
[Haskell] Rewrite Haskell syntax

### LINKED ISSUE #214: [Haskell] "&" does not get highlighted

FichteFoll:
After working with Haskell a bit, I couldn't help but notice how poor the syntax highlighting for it was. It got the job done, but there were and still are several things blatantly missing or using the wrong scopes.

This PR is in a mergeable state. I'm not done or satisfied with it, but I submitted it anyway to collect some feedback if people happen to stumble upon it. I am by no means a Haskell pro. However, this is already a significant improvement over the current syntax.

Notably, the following things still need attention:

- [ ] class definitions
- [ ] type definitions
- [ ] lambdas & `\case`
- [ ] curlies & semicolon
- [ ] records
- [ ] separate subscopes for known operators, such as assignment
- [ ] `|` and `->` are actual keywords, but context-dependant. Others?
- [ ] [unicode operators](https://www.schoolofhaskell.com/school/to-infinity-and-beyond/pick-of-the-week/guide-to-ghc-extensions/basic-syntax-extensions#unicodesyntax)
- [ ] accessors
- [ ] (builtin) types
- [ ] import scope names
- [ ] more I don't remember or know about yet

Fixes #214.


Haskell/syntax_test_haskell.hs
```diff
@@ -93,21 +102,28 @@
 --
    a a = (+) a 2
 --     ^ keyword.operator.haskell
```
djspiewak: Shouldn't this be `keyword.operator.assignment.haskell`? Also perhaps `keyword.operator.declaration.haskell`.

Haskell/Haskell.sublime-syntax
```diff
@@ -5,68 +5,140 @@ name: Haskell
 file_extensions:
   - hs
 scope: source.haskell
+
+variables:
+  # In case this regex seems overly general,
+  # note that Haskell permits the definition of new operators
+  # which can be nearly any string of punctuation characters,
+  # such as $%^&*.
+  operator_infix: '[*|!%$?~+:\-.=</>\\]+'
```
djspiewak: With extensions, Haskell also permits arbitrary unicode operators, so this regex is actually overly-specific.

Haskell/Haskell.sublime-syntax
```diff
@@ -5,68 +5,140 @@ name: Haskell
...
+  operator_parens: '(?:{{operator_infix}}|,+)'
+  module_name: (?:[A-Z][A-Za-z._']*)
```
djspiewak: Unicode should probably be considered here as well.

Haskell/Haskell.sublime-syntax
```diff
@@ -198,56 +305,55 @@ contexts:
         - match: '-\}'
           scope: punctuation.definition.comment.end.haskell
           pop: true
```
djspiewak: This seems insufficient. For example, it would not catch the following:

```haskell
Prelude> {- {- test -} 42 -} 42
42
```

Haskell/syntax_test_haskell.hs
```diff
@@ -93,21 +102,28 @@
...
---       ^^^ entity.name.function.infix.haskell
+--       ^^^ variable.function.infix.haskell
+--        ^ keyword.operator.haskell
 --             ^ constant.numeric.integer.decimal.haskell
    a a = (-) a 2
 --     ^ keyword.operator.haskell
---       ^^^ entity.name.function.infix.haskell
+--       ^^^ variable.function.infix.haskell
```
djspiewak: Should the `variable.function` scope *really* cover the whole thing? I generally think of the parentheses more as operators in this context (similar to the backticks in the ascii case) and the contents as the actual function.

Haskell/syntax_test_haskell.hs
```diff
@@ -93,21 +102,28 @@
...
 --             ^ constant.numeric.integer.decimal.haskell
    a a = (/) a 2
 --     ^ keyword.operator.haskell
---       ^^^ entity.name.function.infix.haskell
+--       ^^^ variable.function.infix.haskell
 --             ^ constant.numeric.integer.decimal.haskell
 
+
+   a `member` x
+--   ^^^^^^^^ keyword.operator.function.infix.haskell
```
djspiewak: This *definitely* shouldn't be `keyword` anything. This syntax is analogous to the prefix parentheses syntax for infix operators. IMO, the contents of the backticks should be `variable.function.prefix` and the backticks themselves should be just `punctuation` scope.

Haskell/syntax_test_haskell.hs
```diff
@@ -120,6 +136,8 @@
 
    instance TooMany Int where
 -- ^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.declaration.instance.haskell
+-- ^^^^^^^^ keyword.declaration..haskell
```
djspiewak: `..`?

Should probably be `keyword.declaration.instance` or `keyword.definition.instance`. That second component should mirror whatever the C mode does with function declaration vs definition. I'm assuming that `class` is `keyword.declaration.class`?

Haskell/syntax_test_haskell.hs
```diff
@@ -145,6 +163,62 @@
...
 
+--KEYWORDS
+
+import qualified Data.Vector.Mutable as MutableVector
+-- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.import.haskell
+-- ^^^ keyword.control.import.haskell
+--     ^^^^^^^^^ keyword.control.import.haskell
+--               ^^^^^^^^^^^^^^^^^^^ support.other.module.haskell
+--                                   ^^ keyword.control.import.haskell
+--                                      ^^^^^^^^^^^^^ support.other.module.haskell
```
djspiewak: This scoping is inconsistent with Scala, which has a similar import mechanism and which is itself consistent with JavaScript, Python, and C#. For reference:

```scala
   import Data.Vector.{Mutable => MutableVector}
// ^^^^^^ meta.import.scala keyword.other.import.scala
//        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.import.scala
//                    ^ meta.import.scala meta.import.selector.scala punctuation.section.group.begin.scala
//                     ^^^^^^^^^^^^^^^^^^^^^^^^^ meta.import.scala meta.import.selector.scala
//                             ^^ meta.import.scala meta.import.selector.scala keyword.operator.arrow.scala
```

Haskell/syntax_test_haskell.hs
```diff
@@ -145,6 +163,62 @@
...
+import Data.List.Split (splitOn)
+-- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.import.haskell
+-- ^^^ keyword.control.import.haskell
+--     ^^^^^^^^^^^^^^^ support.other.module.haskell
+--                     ^^^^^^^^^ meta.declaration.exports.haskell
+--                     ^ punctuation.section.group.begin.haskell
+--                      ^^^^^^^ variable.function.haskell
```
djspiewak: I do like scoping this as `variable.function`, but it *is* inconsistent with other import scopes.

Haskell/syntax_test_haskell.hs
```diff
@@ -145,6 +163,62 @@
...
+--            ^^ keyword.control.context.haskell
+    where
+--  ^^^^^ keyword.control.context.haskell
+        y = 1
+--        ^ keyword.operator.haskell
+
+test a = case a of
+--       ^^^^ keyword.control.conditional.select.haskell
+--              ^^ keyword.control.conditional.select.haskell
+    Nothing -> 0
```
djspiewak: Are the arrows scoped? They should be.

Haskell/syntax_test_haskell.hs
```diff
@@ -145,6 +163,62 @@
...
+
+--MISC
+
+    (group)
+--  ^^^^^^^ meta.group.haskell
+--  ^ punctuation.section.group.begin.haskell
+--        ^ punctuation.section.group.end.haskell
+
+    [1,2]
+--  ^^^^^ meta.sequence.haskell
```
djspiewak: I would say "list" rather than "sequence".

Haskell/syntax_test_haskell.hs
```diff
@@ -186,3 +260,25 @@
     0XdeafBEEF42
 --  ^^^^^^^^^^^^ constant.numeric.integer.hexadecimal
 --  ^^ punctuation.definition.numeric.base
```
djspiewak: Probably worth also adding a test for the `0x` prefix, just to ensure the case insensitivity is working.

FichteFoll:
@djspiewak please note that none of the things you commented on have been changed in the PR, but I appreciate the comments. I'll translate them into todo items for the OP.

Haskell/syntax_test_haskell.hs
FichteFoll: Sequence is the undocumented standard for … sequential types, along with `meta.mapping` for mappings.

Haskell/syntax_test_haskell.hs
FichteFoll: They are matched by the operator pattern.

Haskell/syntax_test_haskell.hs
FichteFoll: Operators (`keyword.operator`) are behind the keyword top scope. We could probably make this `variable.function`, but I would need to think about this more considering how basically everything is a function and the backtick notation only turns a function into an infix function.

Haskell/syntax_test_haskell.hs
FichteFoll: Same problem as https://github.com/sublimehq/Packages/pull/2145#discussion_r349674489.

Haskell/Haskell.sublime-syntax
FichteFoll: How would I differentiate a unicode name from a unicode operator?

Haskell/Haskell.sublime-syntax
FichteFoll: Not exactly arbitrary. I intended to go through the list I linked in OP and add those to this variable.

Haskell/syntax_test_haskell.hs
djspiewak: I think it's probably worth scoping them specifically, since operators should not be scoped as `keyword`, but `->` *should*.

Haskell/Haskell.sublime-syntax
djspiewak: > How would I differentiate a unicode name and a unicode operator?

Hmm, I'll see if I can find a more comprehensive description of Haskell's lexical syntax in this area. It's entirely possible that you can't. In the Scala mode, for example, I gave up entirely on this and just matched identifiers as… identifiers, since there's no meaningful distinction. Haskell does distinguish both at definition site and at call site, but only in a context-sensitive fashion:

- At definition site, parens/backticks may distinguish names from operators, as well as infix usage of the type-declared name (this is the context-sensitive bit)
- At call site, parens/backticks may also be found, but also the use of an identifier in infix position rather than prefix, which should theoretically be parseable

There's got to be *some* distinguishing lexical feature though, which should give us something to hang our hats on. I'll do some digging

Haskell/syntax_test_haskell.hs
djspiewak: Ah, gotcha.

Haskell/syntax_test_haskell.hs
djspiewak: I would generally consider `keyword` in any language to be from a statically fixed set. Anything which *isn't* from a statically fixed set probably shouldn't be a `keyword`. My two cents anyway.

FichteFoll: Seeing how most operators in functional languages (like Haskell) aren't proper keywords, are you saying an "operator" like `+` shouldn't be scoped `keyword.operator`? If so, what other scope should be used? `variable.function`?

How do other functional languages like Scheme, Lisp or Scala do it? Since you wrote the latter, I imagine you having spent more time with this problem than I did.

Haskell/syntax_test_haskell.hs
FichteFoll: Good catch, I was probably debating which 3rd level scope to use there and got sidetracked, which explains the `..`. Relating Haskell operations onto other languages is annoying, because the `class` keyword seems more related to a trait and `instance` declares the implementation of such a trait for a type. This is confusing, especially seeing how `class` is very different from what it does in object-oriented languages.

We don't use `keyword.definition` for anything, afaik.

Haskell/syntax_test_haskell.hs
FichteFoll: I agree, `->`should be a proper keyword, much like guards are. It is used in many different situations and you can't assign to it.

Haskell/syntax_test_haskell.hs
djspiewak: In Scala, I just gave up and I don't use `variable.function` at all. Basically, there's no meaningful way to distinguish functions and operators in Scala, and in all but a very very small set of cases, they behave identically. Haskell does give a bit more information in some cases – specifically the backticks for infix function usage, and the parentheses for prefix oprator usage – the question is just whether or not we want to *use* that information and how we would use it.

Scoping either of them as `keyword` seems very much wrong. In my view, the best thing to do would be to scope the prefix paren syntax as `variable.function.infix`, which we could also use for operators in all positions if the lexical syntax is sufficiently disambiguated. Infix backtick syntax would be something like `variable.function.prefix`. Scoping all non-operator identifiers as `variable.function.prefix` would technically be *correct* in a sense, but it's also probably worthless and we shouldn't do that.

Haskell/syntax_test_haskell.hs
djspiewak: > Relating Haskell declarations to other languages is annoying, because the class keyword declares something like a trait and instance declares the implementation of such a trait for a type. This is confusing, seeing how class is very different from what it does in object-oriented languages.

I kinda wanted to use the C/C++ precendent, since class vs instance *really* is very much like the declaration/definition split in that language. Unfortunately, neither mode actually detects declarations (probably because it wasn't really possible before branching), so we have nothing to go on.

Haskell/syntax_test_haskell.hs
FichteFoll: Imo we *should* highlight infix function calls because they affect the syntax, but we can leave parenthesized infix operators to use them as prefix functions alone or do whatever we do for any other function. They should still be scoped as *something*, but not `variable.function`.

Now, to me, the question remains what infix functions should be scoped as. Semantically, they function like binary operators and this matches well with other languages. If the operator scope is a subscope of `keyword`, then so be it. Scoping them as `variable.function` is also an option, but I worry about color schemes not targeting that or assigning a style that is too subtle (my current scheme just makes them italics). I'm unsure how much bias I have regarding this.
Maybe I'll just try both and see how things work out.

djspiewak: Well, at least to me, the backticks and paren syntaxes are really the same thing: each takes one fixity and swaps it to the other, and both affect syntax. Regarding scoping, I'm just quite worried that `keyword` is too much.

FichteFoll:
Seeing how I probably won't be able to return back to this any time soon, I'm removing the draft status. As mentioned in the OP, the PR *is* mergeable and already an improvement, but there is much more to be improved on.

Haskell/Haskell.sublime-syntax
```diff
@@ -5,68 +5,142 @@ name: Haskell
...
+    - include: list
+    - include: ident
+    - match: ','
+      scope: punctuation.separator.comma.haskell
+
+  group:
+    - match: \(
+      scope: punctuation.section.group.begin.haskell
+      push:
+        - meta_scope: meta.group.haskell
```
moodmosaic: Do we really need/care about this one? On the screenshot below (in the middle), it matches `round` but it shouldn't.

image

Haskell/Haskell.sublime-syntax
```diff
@@ -5,68 +5,142 @@ name: Haskell
...
+        - match: \)
+          scope: punctuation.section.group.end.haskell
+          pop: true
+        - include: expression
+
+  list:
+    - match: \[
+      scope: punctuation.section.sequence.begin.haskell
+      push:
+        - meta_scope: meta.sequence.haskell
```
moodmosaic: Same as above, we shouldn't need that scope, it matches functions that shouldn't be matched with that scope. Those are supposed to be built-in prelude function names (see on the left, where I'm highlighting them manually)

image

FichteFoll: There is no occurance of `[` in your screenshot here.

Haskell/Haskell.sublime-syntax
FichteFoll: Yes, we care about groups in general and mark the entire thing with a meta scope to signify its spanning tree, sort of. We do this for all syntaxes, so not doing it for Haskell would be more than odd.

This sounds like an XY problem to me. What is the behavior you expect (for a given code sample) and what happens instead?

moodmosaic: I'd expect that `round` wouldn't match with `meta.group.haskell` :thinking:

Haskell/Haskell.sublime-syntax
moodmosaic: Yes, that screenshot goes up on that other comment about groups. But it's more or less the same issue. [If you'd like to match names coming from the prelude package](https://github.com/FichteForks/Packages/pull/4/files), for example, you'll see that matching groups and sequences kind of gets in the way...

Haskell/Haskell.sublime-syntax
FichteFoll: The prelude not being matched inside the group is a separate issue and while the removal of these matches work, it's not an actual fix for the issue. I absolutely won't remove these matches and instead work on a proper fix for matching items of the prelude in groups, which is what I interpreted your report as.

moodmosaic: >and instead work on a proper fix for matching items of the prelude in groups, which is what I interpreted your report as.

:100:

moodmosaic:
>Notably, the following things still need attention:

After merging those latest pull requests, apart from
* unicode operators, which I haven't thoroughly checked
* and the groups discussion in https://github.com/sublimehq/Packages/pull/2225#discussion_r380352626

the rest look *good enough* after doing a comparison with @lierdakil's [atom-haskell/language-haskell grammar](https://github.com/atom-haskell/language-haskell/blob/e036e449909816e616b880157e2703e70fc9b5df/grammars/haskell.cson) (which I consider one of the most precise grammars for Haskell).

lierdakil:
Fwiw, I think both atom-haskell/language-haskell and Sublime's Haskell grammar are more or less based on the same TextMate grammar originally. Which in theory means that it should be possible to backport atom-haskell changes here. Or alternatively, convert language-haskell to tmlanguage. Although I think sublime grammar is strictly more powerful than tmlanguage, from which Haskell grammar could benefit a lot -- my implementation has a few limitations that stem from the lack of direct control over context, which I hear sublime's grammar engine allows.

FichteFoll:
I'm pretty sure Sublime's grammar was copied from Textmate and basically never touched since then.

Trying to backport atom's grammer doesn't look like a good option because of how convoluted its regular expressions are and I'd like to take advantage of ST's power with the context stack and branching where feasible, which is better done from scratch. Thanks for the heads-up, however.

On the PR at hand, I incorporated the changes suggested by @moodmosaic. I still believe record syntax to be pretty broken, as it's basically unhandled right now, but the syntax doesn't break on it and has *some* highlighting at least.

moodmosaic:
@FichteFoll 

>I still believe record syntax to be pretty broken, as it's basically unhandled right now, but the syntax doesn't break on it and has *some* highlighting at least.

There's only one thing missing to fix the record syntax, **just don't match the first field and be done**:

image

As you can see above, we're matching the first field as `meta.function.type-declaration.haskell` but if we can avoid that, we're all set.

AllanZyne:
This [syntax file](https://github.com/AllanZyne/Packages/blob/master/Haskell/Haskell.sublime-syntax) is what I wrote when I was interested in Haskell. It might be helpful as reference for **unicode operator**, which I studied from [The Haskell 2010 language report](https://haskell.org/definition/haskell2010.pdf) by some effort.

```
  id_char: '[\p{Ll}\p{Lu}\p{Lt}\p{Nd}_'']' 
  con_id: '[\p{Lu}\p{Lt}]{{id_char}}*'
  var_id: '[\p{Ll}_]{{id_char}}*'
  reserved_id : 'case|class|data|default|deriving|do|else|foreign|if|import|in|infix|infixl|infixr|instance|let|module|newtype|of|then|type|where|_'
  mod_id: (?:(?:{{con_id}}.)*{{con_id}})
  symbol: '[\p{S}\p{P}&&[^(),;\[\]`{}_"'']]'

  qcon_id: '{{con_id}}(?:\.{{con_id}})*'
  qvar_id: (?:{{qcon_id}}\.)?{{var_id}}
  qvar_sym: (?:{{mod_id}}\.)?{{symbol}}+

  qvar: \b(?:{{qvar_id}}|\({{qvar_sym}}\))
  qvar_op: \b({{qvar_sym}}|`{{qvar_id}}`)

  operator_char: (?:[\p{S}\p{P}&&[^(),;\[\]`{}_"']])
  operator: '{{operator_char}}+'
  operator_fun: (?:\((?!--+\)){{operator}}\))
```

moodmosaic:
@FichteFoll 

>There's only one thing missing to fix the record syntax, just don't match the first field and be done:

What I mean is that we go from this:
image

To this:
image

moodmosaic:
Ping

FichteFoll:
It's still on my radar, but so are many other things. I don't have any updates to share atm.

moodmosaic:
@FichteFoll, @wbond, this should be merged, even without the improvement described in https://github.com/sublimehq/Packages/pull/2225#issuecomment-593154731. If there's anything I can do to help further, let me know.

wbond:
@FichteFoll So you are still of the opinion that this is a pretty significant step forward?

FichteFoll:
Definitely, it's mergeable and a significant improvement. The unchecked items in the OP are merely a checklist of things to also check out that I haven't gotten around to yet and probably won't in the near future either. They are no blockers.

wbond:
Thanks for the work on writing this and the reviews!

PR #2226, opened 2019-12-05
[Scala] Removed fake jump markers and corrected class entity scoping

djspiewak:
I couldn't think of a better name for these things. They're obsolete now, and removing them uncovered a bug (which I also fixed).

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -338,7 +338,7 @@ contexts:
         1: storage.type.function.scala
         2: entity.name.function.scala
       push: function-type-parameter-list
-    - match: '\b(case\s+)?(class|trait|object)(\s+({{id}}))'
+    - match: '\b(case\s+)?(class|trait|object)\s+({{id}})'
       scope: meta.class.identifier.scala
       captures:
         1: storage.type.class.scala
--- a/Scala/Symbols-def.tmPreferences
+++ b/Scala/Symbols-def.tmPreferences
@@ -9,10 +9,6 @@
   <dict>
     <key>showInSymbolList</key>
     <integer>1</integer>
-    <key>symbolTransformation</key>
-    <string>
-      s/(.+)/  def $1\(...\)/
-    </string>
   </dict>
 </dict>
 </plist>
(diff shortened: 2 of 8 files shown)


Scala/Scala.sublime-syntax
```diff
@@ -338,7 +338,7 @@ contexts:
         1: storage.type.function.scala
         2: entity.name.function.scala
       push: function-type-parameter-list
-    - match: '\b(case\s+)?(class|trait|object)(\s+({{id}}))'
+    - match: '\b(case\s+)?(class|trait|object)\s+({{id}})'
```
deathaxe: `\b(?:(case)\s+)?` - to ensure the word boundaries of `case` are respected as well.

wbond: @djspiewak I think the capture groups need to be tweaked as @deathaxe mentioned so that the space after case doesn't get the same scope name.

Scala/Symbols-def.tmPreferences
```diff
@@ -9,10 +9,6 @@
   <dict>
     <key>showInSymbolList</key>
     <integer>1</integer>
-    <key>symbolTransformation</key>
-    <string>
-      s/(.+)/  def $1\(...\)/
-    </string>
```
wbond, line 15: This file should also be able to be removed since there is no symbol transform any longer and `entity.name.function` should be part of the symbol list by default.

Scala/Symbols-type.tmPreferences
```diff
@@ -9,10 +9,6 @@
   <dict>
     <key>showInSymbolList</key>
     <integer>1</integer>
-    <key>symbolTransformation</key>
-    <string>
-      s/(.+)/  type $1/
-    </string>
```
wbond, line 15: This file can be removed since `entity.name.type` is included in the symbol list by default

wbond:
I'm working on all of the tmPreferences merge requests now, so I'll probably merge this in with fixes after that

wbond:
Actually, it looks like @deathaxe has incorporated all of the changes from here along with the naming changes.

wbond:
This was superseded by #2293.

PR #2227, opened 2019-12-06
[PHP] Fix catch multiple exceptions in a single catch

jfcherng:
The previous implementation is too strict so that it fails parsing exceptions scattered in multiple lines.

```php
try {
    // ...
} catch (
    // vvv failed
    \Custom\Exceptoin1 |
    \Custom\Exceptoin2 $e
    // ^^^ failed
) {
    // ...
}
```

If the `st3` branch is still maintained, I can make a PR to it.

---

Fixes https://github.com/sublimehq/Packages/issues/2358


wbond:
Thanks!

PR #2228, opened 2019-12-07
[Java] Rewrite Java Server Page Syntax

deathaxe:
This PR addresses an issue reportet at:
https://forum.sublimetext.com/t/jsp-source-theme-change/48165/2

This PR proposes a rewritten JSP syntax definition in order to ...

1. make use of the `embed` technology to inject the `<%%>` tags into the HTML.sublime-syntax, which fixes an issue with `<%%>` not being highlighted within tag attributes
2. apply the HTML like tag scope names.


deathaxe:
It turned out JSP to need more work related with Java itself to maintain contexts etc. Thus it will be handled at a later point.

PR #2229, opened 2019-12-09
[JavaScript] Add meta.number

deathaxe:
This PR replaces the `punctuation.definition.numeric` and `storage.type.numeric` by `meta.number.[base|type]` scope.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -13,7 +13,7 @@ variables:
   dec_digit: '[0-9_]'
   hex_digit: '[\h_]'
   dec_integer: (?:0|[1-9]{{dec_digit}}*)
-  dec_exponent: (?:[Ee](?:[-+]|(?![-+])){{dec_digit}}*)
+  dec_exponent: ([Ee](?:[-+]|(?![-+])){{dec_digit}}*)
 
   identifier_escape: (?:\\u(?:\h{4}|\{\h+\}))
   identifier_start: (?:[_$\p{L}\p{Nl}]|{{identifier_escape}})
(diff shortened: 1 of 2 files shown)


jwortmann:
Will this be a new convention for the scope names in numbers?
Also I would advice against using `meta.number.mantissa` in integer numbers, because afaik a mantissa only exists for floating point numbers. https://en.wikipedia.org/wiki/Significand#Use_of_%22mantissa%22

deathaxe:
The sublimehq core members decided against `punctuation.definition.numeric` and asked for a proper alternative, because:

1. Prefixes like `0x`, `0b` are no `punctuation.definition` (only things like `$` in `$var` might be)
2. Using `punctuation` may cause bad user experience with some color schemes which address these scopes. I can confirm that with one of my custom syntaxes ending up in nearly everything of a number being highlighted as `storage` or `punctuation`, which looks really bad.

The idea therefore is to add new scopes to enable color schemes to colorize the different parts of a number if they want without changing the behavior for existing color schemes.

Different solutions with extra meta scopes or even a new top-level scope were discussed on Discord. Something like `meta.component.[prefix|suffix]` was one of them.

After some days of thinking about appropriate scopes, I ended up in suggesting the following scheme:

```
meta
    number
        sign
        base
        mantissa
        exponent
        type
```

This PR is to discuss the general direction of how to handle numbers.

---

I know that `mantissa` might not be the best choice for integers, but I think all parts of a number should consist of 3 scope levels so color schemes or plugins can select each of them easily. The main part of a number should receive the same meta scope no matter of which type it is. The type is distinguished by the `constant.numeric. ...` already.


---

A second question which is not so clearly visible in the JavaScript syntax, is how strictly numbers should be matched. One single invalid character within a number might cause the whole number to be scoped as `illegal` or not to be scoped at all, which ends up in poor user experience while writing new numbers or modifying existing ones. The color of the number may change with each key stroke.

Only few syntaxes do `illegal` highlighting correctly though. Means it doesn't match the interpreter/compiler errors in most cases. Doing so would require a lot more effort in many cases.

**Example:**

Considder the incomplete number `10_20_`. All interpreters/compilers throw an error on the trailing underscore, but the core syntaxes act differently. 

- Perl and JavaScript are quite permissive scoping everything as `constant.numeric.
- Ruby doesn't scope this token at all
- Java scopes the trailing underscore `invalid`
- Some others just don't highlight the trailing underscore but the `10_20` token.

I'd vote to go the permissive way in such situations in order to improve the user experience with incomplete numbers. While proper invalid highlighting might help finding typing errors, it is somehow annyoing to see tokens change their color very often - especially invalids.



``

jwortmann:
A few more things to think about:
1. Even though `punctuation.separator` doesn't feel to be a wrong term for the decimal point, that scope could as well be replaced with e.g. `meta.number.decimal-separator` then, to improve highlighting with color schemes that target `punctuation`. In my opinion that would improve user experience especially for languages like Ruby and JavaScript, which allow accessor dots right after numbers, because these no longer could be confused with decimal points.
2. Is it really necessary to distinguish between `mantissa` and `exponent`? I can't think of a situation where I want those to be highlighted differently or where plugins could make use of that, so maybe these should be combined to a more appropriate scope for both integer and floating point numbers, e.g. something like `meta.number.digits`.
3. I guess the `i`/`j` for imaginary parts will be `meta.number.type` too then? Or does it make sense to use `meta.number.unit` for it and perhaps for other units like `px`, `rem`, ... in CSS as well (instead of `keyword.other.unit`)? I think I would prefer the latter, i.e. `meta.number.unit`, however in CSS that would change the highlighting style of numbers for almost all color schemes.

<hr>

For incomplete numbers, I also think it is best to not be too restrictive, and the whole number shouldn't change its color while typing, especially not become `invalid`. In my opinion the following would give the most information with no bad user experience for the trailing underscore example:
> Some others just don't highlight the trailing underscore but the `10_20` token.

I unfortunately didn't pay much attention to that when we updated the Ruby syntax recently, so there is definitely room for improvement.

deathaxe:
1. The most color-schemes I've seen so far colorize `punctuation.separator` and `punctuation.accessor` differently and a decimal point is basically a punctuation. The already long living dedicated `punctuation.separator.decimal` could always be used to avoid confusion with accessors. But yeah, these might be the questions to discuss.

2. If we distinguish betwenn the `base`, `type` and `unit` and even considder to add `meta.number.separator`, why not `mantissa` vs. `exponent`? I used those to slightly dimm the exponent part, same as I do for the base part. A scope like `digits` would also match the leading `0` in the base notations but not the `e` or the sign of the exponent, so I am not sure whether it would be a good choice.

3. Yes, anything like `i/j`, `f`, `D`, ... would be scoped `meta.number.type` then. Didn't have the `px`, ... units in mind so far. Would indeed be a breaking change. But honestly, I don't want to end up in stacking 3 or 4 scopes for just two letters, like `meta.number.unit constant.numeric.integer keyword.other.unit`.

_I don't want to make the numbers more complicated than they are. Except the `punctuation.definition.numeric.base` thing, I am basically very fine with what we have._

---

> I unfortunately didn't pay much attention to that when we updated the Ruby syntax recently

Me not, too. I just got aware of those aspects while working on all the different syntaxes as all of those caused different implementations for nearly the same situations.

jwortmann:
1. I just checked the default color schemes:
    * Monokai doesn't target `punctuation.separator`, so it won't be affected
    * Mariana targets `punctuation.separator` and `punctuation.accessor`, but uses the same color for both
    * Sixteen targets `punctuation.separator` and `punctuation.accessor`, but uses the same color for both
    * Celeste targets only `punctuation`, so both will have the same color as well
    * Breakers targets `punctuation.separator` and `punctuation.accessor` and uses different colors ("red2" & "red3")

    It might be a personal preference, but I mean with the suggested changes of using only `meta` scopes for the prefix and suffix, these parts will lose their highlighting in all color schemes (unless those new scopes will be added into the scheme, which could be done with the default color schemes, but don't expect that to happen for custom schemes soon).  The only part which would be left in a different highlighting color then, is the decimal point. I don't think that is optimal from a visual perspective, even though `punctuation.separator.decimal` *is* an appropriate name.

2. With the `meta.number.digits` example I just meant the "main part" of a number, be it `10_20` or `12.34e+56`, including underscores, `e` and the sign after `e`, but not the `0` in the prefix. Perhaps someone can think of a better name for it, ... maybe `meta.number.value`? If we consider my previous point, then another question would be if the decimal point should have both of the `meta` scopes stacked, or just that one with `separator`. If we prefer to have `mantissa` and `exponent` for floating point numbers, then let me ask why did you chose `meta.number.mantissa` for integer numbers, and not `meta.number.exponent`? :smirk:
I know that `exponent` would feel pretty wrong, but `mantissa` is equally wrong, as that doesn't exists for integers, and this choice would just be arbitrary then.

3. I didn't think of stacking the 3 scopes `meta.number.unit constant.numeric.integer keyword.other.unit` (only for CSS), but actually I don't dislike it, because CSS numbers wouldn't change their highlighting and I still think `unit` is more appropriate than `type` for `i/j`.

All these are just suggestions / my point of view, but I think this is what this discussion is intended for.

deathaxe:
Maybe the `meta.number.type` and `meta.number.unit` scopes could be replaced by a more general `meta.number.suffix`.

PR #2231, opened 2019-12-12
[Erlang] Distinguish operators that are words from those that are symbols

frou:
Colour scheme authors can want to target these differently. For example, personally I don't colour operators like `= + *` but do colour operators like `and`.

Use of the scope `keyword.operator.word` is officially endorsed: https://www.sublimetext.com/docs/3/scope_naming.html#keyword

### DIFF
--- a/Erlang/Erlang.sublime-syntax
+++ b/Erlang/Erlang.sublime-syntax
@@ -1928,13 +1928,13 @@ contexts:
   operator:
     # http://erlang.org/doc/reference_manual/expressions.html#arithmetic-expressions
     - match: (div|rem){{ident_break}}
-      scope: keyword.operator.arithmetic.erlang
+      scope: keyword.operator.word.arithmetic.erlang
     - match: (band|bor|bxor|bnot|bsl|bsr){{ident_break}}
-      scope: keyword.operator.bitwise.erlang
+      scope: keyword.operator.word.bitwise.erlang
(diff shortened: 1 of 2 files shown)


michaelblyons:
Very related discussion at #2090

deathaxe:
We should resolve the `[arithmetic|binary|logical|...]` vs. `word` discussion first. I am not sure whether those scope types should be mixed up like that, because of their totally different meaning.

frou:
Overarching standardisation discussion aside, where's the mixed up meaning? The scopes seem like a matter of fact

deathaxe:
At this point it is just another step towards fragmentation of scope names as such fundamental changes have always been applied to single syntaxes for someones own sake. The decision which of the both - representation form (word vs. symbol) or logical meaning (arithmetic|binary|logical|...) - have a heigher priority and thus should go to scope level 3 or 4. With regards to other use cases (keyword.control, constant.numeric) both should but can't stay at the same level. As both are not logically related to each other they shouldn't be combined as proposed in this PR. 

The same way as you want to just distinguish between word like and symbol like operators others might don't want to care about it but the logical meaning of an operator.

Therefore a generally acceptable solution should be found first at #2090 and than I want to see who is willing to apply it to all syntaxes to reduce/abandon the current fragmentation of scope naming.

frou:
Not worth keeping this PR open, since it's just a small satellite of #2090

PR #2234, opened 2019-12-16
[R] Rename scope meta...name->meta...identifier

deathaxe:
This commit renames all `meta.[function|macro|...].name` scopes to
`meta.[function|macro|...].identifier` because

1. the scope naming guideline uses `meta.annotation.identifier`
2. the identifier may consist of a qualifier and the name
3. this scope name is already used in syntaxes like Java.


wbond:
Thanks!

Issue #2236, opened 2019-12-22
[HTML] Doctype should be case insensitive


twhb:
The HTML5 doctype is case insensitive. `<!doctype html>` should be treated the same as `<!DOCTYPE html>`, but currently is not.

Reference: https://html.spec.whatwg.org/multipage/syntax.html#the-doctype

Sublime Version: 3.2.2, Build 3211

PR #2237, opened 2019-12-23
[HTML] DOCTYPE should be case-insensitive

jfcherng:
Fixes https://github.com/sublimehq/Packages/issues/2236.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -257,7 +257,7 @@ contexts:
           scope: invalid.illegal.bad-comments-or-CDATA.html
 
   doctype:
-    - match: (<!)(DOCTYPE)
+    - match: (<!)((?i:DOCTYPE))
       captures:
         1: punctuation.definition.tag.begin.html
         2: keyword.declaration.doctype.html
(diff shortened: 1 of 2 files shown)


deathaxe:
According to whatwg.org the `(?:PUBLIC|SYSTEM)\b` should be case-insensitive, too.

jfcherng:
@deathaxe nice catch. done.

HTML/HTML.sublime-syntax
```diff
@@ -280,7 +280,7 @@ contexts:
     - include: else-pop
 
   doctype-content-type:
-    - match: (?:PUBLIC|SYSTEM)\b
+    - match: (?i)(?:PUBLIC|SYSTEM)\b
```
deathaxe: Why not the shorter `(?i:PUBLIC|SYSTEM)\b` ??

jfcherng: for coding style

deathaxe: Most tags use the `(?i: ...)` style.

deathaxe: ... ok. A bit a mess.

jfcherng: changed to `(?i: ...)` style

jfcherng:
Moved to https://github.com/sublimehq/Packages/pull/2239

PR #2239, opened 2019-12-23
[HTML] Consolidate tag/attribute name matching

deathaxe:
This PR ...

1. consolidates the way the case-insensitive flag `(?i)` is applied to match rules. All remaining rules are modified to apply it as local flag for strings only.

   Notes:
   a) The `{{javascript_mime_type}}` is case-insensitive already.
   b) It is removed from lookaheads with non-word characters only.
   c) All tag attributes are matched case-insensitive now.

2. removes the leading `\b` in some remaining tag and attribute names as the way the syntax works already ensures any match to start at an appropriate position (not in the middle of a word).

3. replaces the trailing `\b` in some remaining tag and attribute names by `{{tag_name_break}}` or `{{attribute_name_break}}` to make sure to properly match the name boundaries according to the whatwg specs under all circumstances.

4. consolidates the `</script>` and `</style>` closing tag match rules to behave like any other recently updated tags. The `>` can be at the next line now.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -22,8 +22,9 @@ variables:
   unquoted_attribute_break: (?=[{{ascii_space}}]|/?>)
 
   # https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state
-  tag_name_char: '[^{{ascii_space}}/<>]'
-  tag_name_break: (?=[^{{tag_name_char}}])
+  tag_name_break_char: '[{{ascii_space}}/<>]'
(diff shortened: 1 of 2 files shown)


jfcherng:
Would you also have [this](https://github.com/sublimehq/Packages/pull/2237) done here. I feel there will be conflict and mine PR is less useful. I would like to close mine if its done somewhere in other PR.

deathaxe:
Guess, you ask because of the `{{tag_name_break}}` changes?

jfcherng:
I just feel it's more suitable to be in this PR :D Thanks

wbond:
Thanks for these improvements!

PR #2240, opened 2019-12-23
[HTML] Consolidate quotes

deathaxe:
This PR proposes to consolidate the usage of escapes vs. quotes in all remaining matches. It prefers unquoted values, if the required escapes are shorter and still well readable.

It does not provide any functional changes.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -269,7 +269,7 @@ contexts:
 
   doctype-meta:
     - meta_scope: meta.tag.sgml.doctype.html
-    - match: \>
+    - match: '>'
       scope: punctuation.definition.tag.end.html
       pop: true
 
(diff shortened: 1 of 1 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -269,7 +269,7 @@ contexts:
 
   doctype-meta:
     - meta_scope: meta.tag.sgml.doctype.html
-    - match: '>'
```
FichteFoll: Personally, I prefer the quoted version instead of a backslash escape here because `>` as syntax is interpreted by YAML, not by the regular expression engine, where the escape is simply ignored (but not in boost, for example).

Since we need to escape > for YAML, I say we should also use YAML's means for doing so through a quoted scalar. The same applies to other syntax punctuation like colons and quotation marks.

FichteFoll:
What about `\"` and `\'`? The same applies for those.

deathaxe:
Sorry, but I hat that mix of `"'"` vs. `'"'` and got used to use the escaped ones in nearly every syntax/change I made - same for `\(`, `\[` and `\{'. This was originally the reason to also do so for `>` as it just feels inconsitent to use different styles for comparable types of section punctuations.

It's not a question about escaping regexp vs. "required by yaml" but a cleaner overall look and feel of the code itself.

PR #2242, opened 2020-01-02
[Lua] Fix goto function references

edubart:
This fixes missing references in "Goto Reference..." (Shift+F12) for Lua source files by using the appropriate scope `meta.function-call.lua` in functions calls.

Missing `keyword.declaration` scope names for **function** keyword were also added according to the documented syntax naming conventions to have a better Lua syntax highlighting in some newer color schemes.

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -85,7 +85,7 @@ contexts:
       scope: punctuation.terminator.statement.lua
 
     - match: function{{identifier_break}}
-      scope: storage.type.function.lua
+      scope: storage.type.function.lua keyword.declaration.function.lua
       push:
         - function-meta
         - block-contents
@@ -344,7 +344,7 @@ contexts:
       scope: meta.property.lua entity.name.function.lua
       pop: true
     - match: '{{identifier}}{{function_call_ahead}}'
-      scope: meta.property.lua variable.function.lua
+      scope: meta.property.lua meta.function-call.lua variable.function.lua
       pop: true
     - match: '{{metaproperty}}'
       scope: meta.property.lua support.other.metaproperty.lua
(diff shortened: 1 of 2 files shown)


jrappen:
Please update the tests, as well. Compare <https://www.sublimetext.com/docs/3/syntax.html#testing>

edubart:
Tests for the new scope names were added in Lua syntax tests and they seams to pass successfully.

Lua/Lua.sublime-syntax
```diff
@@ -102,7 +102,7 @@ contexts:
       push: expression-list
 
     - match: local{{identifier_break}}
-      scope: storage.modifier.lua
+      scope: storage.modifier.lua keyword.declaration.local.lua
```
jwortmann: Should `local` really get the `keyword.declaration` scope? According to the guidelines, this applies only to function/class/type definition keywords (`storage.type`), but not to modifiers (`storage.modifier`).

edubart: I think you are correct, the motivation of doing this was to fix syntax highlighting inconsistencies with the **local** keyword on some color themes, but I've now realized that actually the color scheme was doing wrong. The change was removed.

Lua/Lua.sublime-syntax
```diff
@@ -423,8 +423,8 @@ contexts:
           |next|pairs|pcall|print|rawequal|rawget|rawlen|rawset|select
           |setmetatable|tonumber|tostring|type|xpcall
           |require|getfenv|module|setfenv|unpack
-        ){{identifier_break}}
-      scope: support.function.builtin.lua
+        ){{function_call_ahead}}
```
FichteFoll: I actually think this match might need to be doubled, because if there is a function call ahead, we apply `meta.function-call`, but if there isn't, the function is just as an object where it should *still* be scoped as `support.function`. Non-deterministic highlighting *might* also help.

edubart: I actually think is odd to highlight any variable with these names, because as a Lua programmer is common for me to declare a local variable named `next`, `type`, `module` in certain circumstances and they are certainly not the builtin functions on these cases so are wrongly highlighted. Thus was intentional to change this.

Plus the highlighter already can't (and shouldn't) highlight non builtin function references without a call ahead, so doing the same for builtin functions keeps the highlighting more uniform.

wbond:
Looks good to me, thanks @edubart!

PR #2244, opened 2020-01-04
[JavaScript] Improve implementation of arrow function branching.

Thom1729:
- Reuse existing contexts where possible.
- Detect arrow functions with a single bare parameter.
- Add tests.


wbond:
Thanks for cleaning this up!

PR #2246, opened 2020-01-14
[Python] Non-deterministic format-spec matching

### LINKED ISSUE #2232: [Python] Incorrect format context in string
- Sublime Version: 4061
- OS Version: Ubuntu 19.10

Given the following Python dictionary:
```python
{
    'diagnostics':
    [
        {
            'message': "Argument of type '{ props: { translations: { type: (ObjectConstructor | FunctionConstructor)[]; required: boolean; }; }; data: () => { showBanner: boolean; fallback: { title: string; text: string; agreeButtonText: string; closeTitle: string; }; }; computed: { ...; }; mounted: typeof mounted; methods: { ...; }; }' is not assignable to parameter of type 'new (...args: any[]) => any'.\n  Type '{ props: { translations: { type: (ObjectConstructor | FunctionConstructor)[]; required: boolean; }; }; data: () => { showBanner: boolean; fallback: { title: string; text: string; agreeButtonText: string; closeTitle: string; }; }; computed: { ...; }; mounted: typeof mounted; methods: { ...; }; }' provides no match for the signature 'new (...args: any[]): any'.",
            'code': 2345,
            'severity': 1,
            'range':
            {
                'end':
                {
                    'character': 0,
                    'line': 0
                },
                'start':
                {
                    'character': 0,
                    'line': 0
                }
            },
            'source': 'Vetur'
        }
    ],
    'uri': 'file:///Users/cookie.vue'
}
```

The value for the `message` key starts a `constant.other.placeholder.python` scope whereas it should not.

I believe a minimal example for this would be

```python
x="{:{ {}}"
```

My Python interpreter says this is not a formatting string:

```
>>> x="{:{ {}}".format("1", "2", "3")
  File "<stdin>", line 1, in <module>
ValueError: unmatched '{' in format spec
```

### LINKED ISSUE #2241: [Python] Breaking on strings with { and !a inside single quotes
The source code is from: 
* https://github.com/python/cpython/blob/91874bb07161bb481b6f5ea18ffafe69cb8cac30/Lib/test/test_logging.py#L3864-L3878

```Python
        self.assertRaises(ValueError, logging.Formatter, '{{foo!r:4.2}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{{foo!r:4.2}}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{foo/bar}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{foo:{{w}}.{{p}}}}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{foo!X:{{w}}.{{p}}}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{foo!a:random}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{foo!a:ran{dom}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{foo!a:ran{d}om}', style='{')
        self.assertRaises(ValueError, logging.Formatter, '{foo.!a:d}', style='{')

        # Dollar style
        # Testing failure for mismatch bare $
        self.assert_error_message(
            ValueError,
            "invalid format: bare \'$\' not allowed",
            logging.Formatter, '$bar $$$', style='$'
        )
        self.assert_error_message(
            ValueError,**
```

image

After that point, the scope becomes all `source.python meta.function-call.arguments.python meta.string.python string.quoted.single.python constant.other.placeholder.python meta.format-spec.python constant.other.format-spec.python`

### Environment

  * Windows `10 build 15063 x64`
  * Resolution `1920x1080`
  * dpi_scale `1.0`
  * Sublime Text:
  * Build `3207`
  * 64 bit

FichteFoll:
Correctly match some tricky format specifications using branches. Some
quite unusual ones are not correctly matched but tests have been added
to document the exact behavior.

My first experiement with non-determinism and quite a tricky case at that because we're not aiming for 100% accuracy but it still allows for more correct highlighting.

Closes #2232.
Closes #2241.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1395,6 +1395,8 @@ contexts:
 
   formatting-syntax:
     # https://docs.python.org/3.6/library/string.html#formatstrings
+    # Technically allows almost every character for the key,
+    # but those are rarely used if ever.
     - match: |- # simple form
         (?x)
         (\{)
@@ -1410,14 +1412,27 @@ contexts:
         2: storage.modifier.conversion.python
         3: constant.other.format-spec.python
         4: punctuation.definition.placeholder.end.python
-    - match: \{(?=[^\}"']+\{[^"']*\}) # complex (nested) form
+    - match: (?=\{[^{}"']+\{[^"']*\})  # complex (nested) form
+      branch_point: formatting-syntax-branch
+      branch:
+        - formatting-syntax-complex
+        - formatting-syntax-fallback
+
+  formatting-syntax-fallback:
+    - match: \{
+      scope: meta.debug.formatting-syntax-fallback.python
+      pop: true
+
+  formatting-syntax-complex:
+    - match: \{
       scope: punctuation.definition.placeholder.begin.python
       push:
         - meta_scope: constant.other.placeholder.python
         - match: \}
           scope: punctuation.definition.placeholder.end.python
-          pop: true
-        - match: '[\w.\[\]]+'
+          pop: 2
+        # TODO could match numeric indices or everything else as a key
+        # and also [] indexing
         - match: '![ars]'
           scope: storage.modifier.conversion.python
         - match: ':'
@@ -1425,7 +1440,10 @@ contexts:
             - meta_scope: meta.format-spec.python constant.other.format-spec.python
             - match: (?=\})
               pop: true
-            - include: formatting-syntax
+            - match: (?=\{)
+              push: formatting-syntax-complex  # need to push extra because of pop: 2
+        - match: '[{"''\n]'
+          fail: formatting-syntax-branch
 
   f-string-content:
     # https://www.python.org/dev/peps/pep-0498/
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1400,18 +1400,19 @@ contexts:
     - match: |- # simple form
         (?x)
         (\{)
-          (?: [\w.\[\]]+)?             # field_name
-          (   ! [ars])?                # conversion
-          (   : (?:{{format_spec}}|    # format_spec OR
-                   [^}%]*%.[^}]*)      # any format-like string
+          (?: [\w.\[\]]+)?           # field_name
+          (   ! [ars])?              # conversion
+          (?: (:) ({{format_spec}}|  # format_spec OR
+                   [^}%]*%.[^}]*)    # any format-like string
           )?
         (\})
       scope: constant.other.placeholder.python
       captures:
         1: punctuation.definition.placeholder.begin.python
         2: storage.modifier.conversion.python
-        3: constant.other.format-spec.python
-        4: punctuation.definition.placeholder.end.python
+        3: punctuation.separator.format-spec.python
+        4: meta.format-spec.python constant.other.format-spec.python
+        5: punctuation.definition.placeholder.end.python
     - match: (?=\{[^{}"']+\{[^"']*\})  # complex (nested) form
       branch_point: formatting-syntax-branch
       branch:
@@ -1426,22 +1427,23 @@ contexts:
   formatting-syntax-complex:
     - match: \{
       scope: punctuation.definition.placeholder.begin.python
-      push:
+      set:
         - meta_scope: constant.other.placeholder.python
         - match: \}
           scope: punctuation.definition.placeholder.end.python
-          pop: 2
+          pop: true
         # TODO could match numeric indices or everything else as a key
         # and also [] indexing
         - match: '![ars]'
           scope: storage.modifier.conversion.python
         - match: ':'
+          scope: punctuation.separator.format-spec.python
           push:
-            - meta_scope: meta.format-spec.python constant.other.format-spec.python
+            - meta_content_scope: meta.format-spec.python constant.other.format-spec.python
             - match: (?=\})
               pop: true
             - match: (?=\{)
-              push: formatting-syntax-complex  # need to push extra because of pop: 2
+              push: formatting-syntax-complex
         - match: '[{"''\n]'
           fail: formatting-syntax-branch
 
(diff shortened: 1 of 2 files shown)


Python/syntax_test_python_strings.py
```diff
@@ -549,6 +549,45 @@
...
+# ^ - constant.other.placeholder
+
+# Syntactically correct, but hardly come up in real code
+"{:{ ()}}".format(0, **{" ()": "d"}) == '0'
+# ^ constant.other.placeholder
+'{foo/bar}'.format(**{"foo/bar": 1}) == '1'
+# ^ - constant.other.placeholder
+
+# Legal but non-standard format spec
+'{foo:{{w}}.{{p}}}}'  # illegal double {{ in format spec
```
wbond, line 584: This line confuses me. Above it says "Legal but non-standard", whereas this line says "illegal".

Python/syntax_test_python_strings.py
```diff
@@ -549,6 +549,45 @@
...
+# ^ - constant.other.placeholder
+'{foo!a:random}'
+# ^ - constant.other.placeholder
```
wbond, line 587: Is this another case of a legal format that the current implementation doesn't handle?

Python/Python.sublime-syntax
```diff
@@ -1410,22 +1412,38 @@ contexts:
...
         - match: '![ars]'
           scope: storage.modifier.conversion.python
         - match: ':'
           push:
             - meta_scope: meta.format-spec.python constant.other.format-spec.python
             - match: (?=\})
               pop: true
-            - include: formatting-syntax
+            - match: (?=\{)
+              push: formatting-syntax-complex  # need to push extra because of pop: 2
```
wbond, line 1444: Could this be worked around using `set` on line 1429?

Python/syntax_test_python_strings.py
FichteFoll: The syntax is legal and translates to `format(foo, "{w}.{p}")`. The comment refers to how a format spec of `{w}.{p}` is unusual and must be interpreted by a custom `__format__` implementation, because this is unheard of in the entire stdlib. I'll reword it to be clearer.

FichteFoll: Actually, the spec is illegal because `}` (or rather `}}` in the escaped form) doesn't work. This doesn't seem to be documented anywhere, but the following snippet can be used to test behavior:
```py
class Foo:
    def __format__(self, spec):
        return spec.format(w=1, p=123)
fmt = '{foo:{{w}}.{{p}}}}'
print(fmt.format(foo=Foo()))
```
```
  File "/tmp/format-spec.py", line 5, in <module>
    print(fmt.format(foo=Foo()))
ValueError: Single '}' encountered in format string
```

FichteFoll: Here's an example implementation showing how it works:
```py
class Foo:
    def __format__(self, spec):
        return spec.format(w=1, p=123)
fmt = '{foo:{{w}}.{{p}}}'
print(fmt.format(foo=Foo()))
```
```
1.123
```

Python/syntax_test_python_strings.py
FichteFoll: Yes, but deliberately.

Python/syntax_test_python_strings.py
```diff
@@ -549,6 +549,45 @@
...
+'{foo.!a:d}'
+# ^ constant.other.placeholder
+"{:{ {}}"  # issue 2232
+# ^ - constant.other.placeholder
+
+# Syntactically correct, but hardly come up in real code
+"{:{ ()}}".format(0, **{" ()": "d"}) == '0'
+# ^ constant.other.placeholder
+'{foo/bar}'.format(**{"foo/bar": 1}) == '1'
+# ^ - constant.other.placeholder
```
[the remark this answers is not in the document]
FichteFoll: The opening comment explains why we don't need or want to match all legal syntax.
```
# We don't actually want to match the syntax 100% of the time,
# since we never know for sure if the string is used as a format string,
# so some of these matches are incorrect because of implementation details.
```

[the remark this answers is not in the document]
wbond: Right, it just isn't crystal clear to me which is which, based on some of the ambiguous comments. I just figured it would be good to know on a case by case basis which is a non-match due to the implementation.

Python/syntax_test_python_strings.py
```diff
@@ -549,6 +549,45 @@
...
+'{foo!a:ran{dom}'  # unclosed
+# ^ - constant.other.placeholder
+'{foo!a:ran{dom}'  # unclosed
+# ^ - constant.other.placeholder
+
+# Invalid field names
+'{foo{d}}'
+# ^ - constant.other.placeholder
+'{foo.!a:d}'
+# ^ constant.other.placeholder
```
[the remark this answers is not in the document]
FichteFoll: That is correct-ish. Trying to format this raises `ValueError: Empty attribute in format string`, but in reality it's the same as typing `foo.` outside of a format string which we still highlight because of a) convenience for the highlighter and b) highlighting while typing.

Python/Python.sublime-syntax
FichteFoll: Wasn't possible before because I couldn't `set` out of the `\{` match due to a recently fixed issue in the syntax, but I'll try now.

Python/syntax_test_python_strings.py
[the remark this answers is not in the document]
FichteFoll: I have a commit updating the comments locally. Will push once I got to the `set` part.

FichteFoll:
I refined the commits a but and removed the pop:2. The extra push is still there, but this time because it *should* be there and should generally behave better in one particular edge case.

wbond:
Thanks for the extra info!

PR #2249, opened 2020-01-16
[Markdown] Add GitHub list checkboxes

### LINKED ISSUE #2248: [Markdown] Add GFM checkbox notation
- Sublime Version: 4062
- OS Version: MacOS Catalina

Immediately following a list item marker `-`/`+`/`*`, match `[ ]`, `[X]`, and `[x]` as some kind of keyword or other entity. This is particularly so they don't get confused as link text identifiers when a real link follows (see below).

``` markdown
Integers
========
- [ ] [PEP 0237][]: Essentially, `long` renamed to `int`. That is, there is only one...
- [x] [PEP 0238][]: An expression like `1/2` returns a `float`. Use `1//2` to get the...
- [x] Octal literals are no longer of the form `0720`; use `0o720` instead.

[PEP 0237]: https://python.org/dev/peps/pep-0237
[PEP 0238]: https://python.org/dev/peps/pep-0238
```

michaelblyons:
Fixes #2248

Unsure on the scope to use, but I like the `.github` tag at the end. In a sense, `constant.language` is "right," considering that these represent Booleans, but I didn't think that would go over well.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -203,10 +203,12 @@ contexts:
         - include: table
         - match: ''
           pop: true
-    - match: '^([ ]{0,3})([*+-])(?=\s)'
+    - match: ^([ ]{0,3})([*+-])( (\[[ xX]\]))?(?=\s)
       captures:
         1: markup.list.unnumbered.markdown
         2: markup.list.unnumbered.bullet.markdown punctuation.definition.list_item.markdown
+        3: markup.list.unnumbered.markdown
+        4: constant.language.checkbox.markdown-gfm
       push:
         - meta_content_scope: markup.list.unnumbered.markdown
         - match: ^(?=\S)
@@ -987,10 +989,12 @@ contexts:
       push:
         - match: ^\s*$
           pop: true
-        - match: '([ ]*)([*+-])(?=\s)'
+        - match: ([ ]*)([*+-])( (\[[ xX]\]))?(?=\s)
           captures:
             1: markup.list.unnumbered.markdown
             2: markup.list.unnumbered.bullet.markdown punctuation.definition.list_item.markdown
+            3: markup.list.unnumbered.markdown
+            4: constant.language.checkbox.markdown-gfm
           push:
             - clear_scopes: 1
             - meta_content_scope: markup.list.unnumbered.markdown meta.paragraph.list.markdown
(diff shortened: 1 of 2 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -203,10 +203,12 @@ contexts:
         - include: table
         - match: ''
           pop: true
-    - match: '^([ ]{0,3})([*+-])(?=\s)'
+    - match: ^([ ]{0,3})([*+-])( (\[[ xX]\]))?(?=\s)
       captures:
         1: markup.list.unnumbered.markdown
         2: markup.list.unnumbered.bullet.markdown punctuation.definition.list_item.markdown
+        3: markup.list.unnumbered.markdown
+        4: keyword.other.markdown.github
```
keith-hall: [in other places in this syntax def](https://github.com/sublimehq/Packages/blob/8f997996e381ae473aaae0e6b0d6a137705a12b5/Markdown/Markdown.sublime-syntax#L768), we are using `.markdown-gfm` as the suffix to scope features specific to GitHub Formatted Markdown

Markdown/Markdown.sublime-syntax
```diff
@@ -203,10 +203,12 @@ contexts:
         - include: table
         - match: ''
           pop: true
-    - match: '^([ ]{0,3})([*+-])(?=\s)'
+    - match: ^([ ]{0,3})([*+-])( (\[[ xX]\]))?(?=\s)
       captures:
         1: markup.list.unnumbered.markdown
         2: markup.list.unnumbered.bullet.markdown punctuation.definition.list_item.markdown
+        3: markup.list.unnumbered.markdown
+        4: keyword.other.checkbox.markdown-gfm
```
FichteFoll, line 211: Random thought while browsing this: What about a `storage` scope?

michaelblyons: I'd be fine with that.

michaelblyons: Preference for `storage.type` or `storage.modifier`?

FichteFoll: Modifier, imo. Actually, now that you mention it it might as well be a language constant. I think I prefer that.

michaelblyons: > Modifier, imo. Actually, now that you mention it it might as well be a language constant. I think I prefer that.

Done.

michaelblyons:
> Could you add a comment at the top of the fail that mentions this extension as being supported by the syntax? LGTM otherwise.

"Top of the... file?" It already mentions GFM, and to look at the suffix.

FichteFoll:
True that. I though we had an actual feature list for extensions over commonmark in there, but mentioning GFM is certainly enough to cover checkboxes. :+1: Maybe such a list could be added by @keith-hall later, since he did most of it.

wbond:
Looks good, thanks for the implementation and reviews!

PR #2250, opened 2020-01-16
[R] updated support functions and added raw string support of R 4.0

randy3k:
Updated the support functions in R syntax to R 4.0. At the same time, it fixes an escaping bug of `.`, for example `all\.equal.default` -> `all\.equal\.default`.

Added the raw string support of R 4.0.


FichteFoll:
OP should be updated to mention the raw strings addition also included in this PR.

Would you mind grouping the really long lines for matching the stdlib into reasonable line chunks (and not just wrap them at x characters), so they don't reflow the entire block when exceeding the x character line length? It's basically impossible to see which names have been added currently.

randy3k:
@FichteFoll 
Actually I also do not know what names are added. The function names are automatically generated with

https://github.com/sublimehq/Packages/blob/master/R/R.sublime-syntax#L303-L307

randy3k:
Actually, let me update the list for R 4.0.

FichteFoll:
Oh, okay, in that case this would be hard.

In that case, the lists *could* be written out with one entry per line into a separate hidden syntax that the main syntax includes only for the buildin function matches. Doesn't need to happen within this PR though and is generally just a suggestion.

wbond:
Thanks for the updates and improvements!

PR #2252, opened 2020-01-17
[C/C++] Recognize C++20 spaceship operator

swlody:
Since C++20, the `<=>` operator needs to be supported in all the same contexts as existing comparison operators. Update `C++.sublime_syntax` to recognize it in existing patterns: `operator_method_name`, `non_angle_brackets`, and `early-expressions-before-generic-type`.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -38,7 +38,7 @@ variables:
   identifier: \b[[:alpha:]_][[:alnum:]_]*\b # upper and lowercase
   macro_identifier: \b[[:upper:]_][[:upper:][:digit:]_]{2,}\b # only uppercase, at least 3 chars
   path_lookahead: '(?:::\s*)?(?:{{identifier}}\s*::\s*)*(?:template\s+)?{{identifier}}'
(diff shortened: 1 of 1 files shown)


C++/C++.sublime-syntax
```diff
@@ -52,7 +52,7 @@ variables:
   visibility_modifiers: 'private|protected|public'
   other_keywords: 'typedef|nullptr|{{visibility_modifiers}}|static_assert|sizeof|using|typeid|alignof|alignas|namespace|template'
   modifiers: '{{storage_classes}}|{{type_qualifier}}|{{compiler_directive}}'
-  non_angle_brackets: '(?=<<|<=)'
```
swlody: I'm not certain this change is necessary, but it looks like this pattern already includes `<=` which shares the same starting tokens. I can't think of a scenario that would need to match `=<<` though...?

C++/C++.sublime-syntax
```diff
@@ -52,7 +52,7 @@ variables:
...
+  non_angle_brackets: '(?=<<|<=>?)'
```
swlody: I'm not certain this change is necessary, but it looks like this pattern already includes <= which shares the same starting tokens. I can't think of a scenario that would need to match =<< though...?

PR #2253, opened 2020-01-18
[C/C++] Recognize C++20 spaceship operator

swlody:
* Update syntax definitions for C++ and Objective-C++
* Add new tests

### DIFF
--- a/C++/syntax_test_cpp.cpp
+++ b/C++/syntax_test_cpp.cpp
@@ -530,6 +530,10 @@ template <class T>
 bool A<T>::operator    ==    (const A& other) { return false; }
 /*   ^^^^^^^^^^^^^^^^^^^^ meta.function meta.toc-list.full-identifier */
 /*         ^^^^^^^^^^^^^^ entity.name.function */
+template <class T>
+bool A<T>::operator   <=>    (const A& other) { return false; }
+/*   ^^^^^^^^^^^^^^^^^^^^ meta.function meta.toc-list.full-identifier */
+/*         ^^^^^^^^^^^^^^ entity.name.function */
 
 typedef std :: vector<std::vector<int> > Table;
 /*          ^^ punctuation.accessor */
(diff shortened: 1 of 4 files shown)


C++/syntax_test_cpp.cpp
```diff
@@ -789,8 +797,9 @@ int main() {
 
 template <typename T>
 struct A<T, enable_if_t<std::is_arithmetic<T>::value && !is_std_char_type<T>::value>> {
-    using x = conditional_t<sizeof(T) <= sizeof(long), long, long long>;
-    /*                                ^^ keyword.operator */
+    using x = conditional_t<sizeof(T) <= sizeof(long) <=> true, long, long long>;
```
rwols: I'm not very familiar with the spaceship operator yet. I only know that one can define it for a class to alleviate most of the boilerplate for defining value-like types. What is its purpose in the expression `sizeof(T) <= sizeof(long) <=> true` ?

swlody: There's not really any reason to write code like that in real life. One would typically just define the operator and let the compiler generate the other comparison operators, like you said. I was really just trying to come up with a way to test the `<=>` operator inside of a template argument, which is technically valid C++, if a little pathological.

Testing it a little bit more, this wouldn't actually compile in real C++20 because the `<=>` operator returns a value of type [`std::strong_ordering`](https://en.cppreference.com/w/cpp/utility/compare/strong_ordering) and not a bool required by `static_assert`.

Something like the following would actually compile: (https://godbolt.org/z/asYMp4)

```
#include <compare>
static_assert(((sizeof(int) <= sizeof(long)) <=> true) == std::strong_ordering::equivalent, "");
```

I can add a separate test for this, or leave it as is, since it's syntactically valid but semantically nonsense.

swlody: With some struggling, I was able to come up with a legitimate but still pretty convoluted test for this case: https://godbolt.org/z/y8PZfo

rwols: Thanks. Was about to say that, at least for me, tests that are semantically valid are preferred :)

wbond:
Thanks @swlody, @rwols and @FichteFoll!

PR #2254, opened 2020-01-20
[Scala] Fix infinite loop

BenjaminSchaaf:
In `single-type-expression-tail`, `@/` would match the lookahead on line `1774` but not the annotation, thus causing an infinite loop.


wbond:
Thanks!

PR #2255, opened 2020-01-20
[JavaScript] Fix key bindings in template strings.

Thom1729:
When the interpolated string scopes were changed, the key bindings broke. This PR updates the keymap to use the new scopes.


wbond:
Thanks!

PR #2259, opened 2020-01-29
this is so that if the page has any Unicode character, it won't show …

KennethKinLum:
…up as weird characters on some browsers


wbond:
Thanks!

PR #2261, opened 2020-02-01
[Python] Simplify Symbol List and Index

### LINKED ISSUE #2257: [Python] Classes and func'ns listed in symbol ref as "keywords"
I'm not totally sure whether this belongs here or at [sublimehq/sublime_text](https://github.com/sublimehq/sublime_text).

- Sublime Version: 4062
- OS Version: MacOS Catalina

## Expected behavior
<kbd>Cmd</kbd>+<kbd>R</kbd> shows classes with a `c` and functions with an `f`.

## Actual behavior
Both have a `k` (for keyword?) because the indexer is matching a `meta` scope that includes the `class` and `def`.

## Steps to reproduce

1. Open a Python file with ST4
2. <kbd>Cmd</kbd>+<kbd>R</kbd>

deathaxe:
Fixes #2257

This commit merges the `Symbol Index.tmPreferences` into the `Symbol List.tmPreferences`.

The symbol transformations are removed. The functions/classes are listed without modifications/indention right next after each other, but with the correct type icon.


Python/Symbol List.tmPreferences
```diff
@@ -4,16 +4,16 @@
 	<key>name</key>
 	<string>Symbol List</string>
 	<key>scope</key>
-	<string>source.python meta.function - meta.function.inline, source.python meta.class</string>
+	<string>
+		source.python entity.name.class,
+		source.python entity.name.function
```
FichteFoll: I wonder, why not use `entity.name.class.python, entity.name.function.python`? That's more precise, although we don't include other syntaxes in Python.

michaelblyons: @FichteFoll Not an issue _here_ per se, but I prefer the `source.x foo` rules over `foo.x`.

This makes a difference when someone has a `foo.Y.x` that they expect to be matched, but isn't. Say for instance there's an `entity.name.function.CONSTRUCTOR.python` and suddenly that one doesn't show up in indexing.

FichteFoll: Well, yes, but we control the syntax definition, so that would still be doable.

I didn't know about conventions for scopes *after `entity.name.function`, so if `.constructor` is a standard in other languages, we could definitely apply that to `__init__` and `__new__` as well.

deathaxe: Indeed I have several questions regarding symbol lists and how to propably find a common way to handle them in all packages the same way, which might better be discussed in a dedicated issue.

One of those is the way how selectors are formed. Many packages seem to follow the rule to use `source.xyz entity.name. ...` while `entity.name. .... .xyz` would have the same result.

In a manner of finding a common way of creating selectors, which work for all kinds of syntaxes the same way, I tend to find @michaelblyons' arguments a good point. It doesn't matter, whether there are conventions for such sub scopes or not - is not prohabited. The current form of selectors just makes sure all such entities are indexed without the need to tweak the selector each time such an entity is added. They just keep simple with the least scopes required to index a certain kind of entity.

wbond: Generally I think the scope syntax suffixes tend to be more of a backward compatibility thing with TextMate. They _do_ allow distinguishing between syntaxes when they are nested, but as noted, they tend to break if someone uses a more specialized scope, e.g. `entity.name.function.constructor`. I think a lot of the built-in functionality of Sublime Text tends to never interact with them, and they tend to break syntax tests somewhat often.

In terms of usage, I think it is sort of a "6 of one, half dozen of another" situation.

PR #2262, opened 2020-02-02
[Ruby] Fix SHELL heredoc tag boundaries

deathaxe:
Add word boundary checks to the heredoc type names to ensure correct scoping of short and long tag names like SH vs. SHELL.

Without those only SH is scoped in SHELL.

A tag name like SHOUT was scoped as SH heredoc as well.


wbond:
Thanks!

PR #2264, opened 2020-02-03
[Graphviz] Change relationship scope

michaelblyons:
Since relationships are the primary thing of interest in Graphviz files, they shouldn't be marked as `punctuation`.

I've had this locally for ages and forgot to push it.


wbond:
Looks good, thanks!

PR #2266, opened 2020-02-06
[Markdown] Optional preview to the side

jrappen:
Requires:

- `ST4065+` for new HTML sheets
- `Package Control` for dependencies

Usage:

- right-click for context menu in Markdown file
- context menu item grayed out (`is_visible` is `False`) unless `mdpopups` (dependency) can be imported and cursor is within `Markdown` view

---

ScreenFlow

### DIFF
--- a/Markdown/Context.sublime-menu
+++ b/Markdown/Context.sublime-menu
@@ -1,0 +1,8 @@
+[
+    { "caption": "-" },
+    {
+        "caption": "HTML Preview (read_only)",
+        "command": "preview_markdown_via_html_sheet"
+    },
+    { "caption": "-" }
+]
--- a/Markdown/dependencies.json
+++ b/Markdown/dependencies.json
@@ -1,0 +1,13 @@
+{
+    "*": {
+        ">=3124": [
+            "markupsafe",
+            "mdpopups",
+            "pygments",
+            "pymdownx",
+            "python-jinja2",
+            "python-markdown",
+            "pyyaml"
+        ]
+    }
+}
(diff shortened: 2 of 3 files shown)


FichteFoll:
> PackageControl for dependencies 

Pretty sure this is already a guarantee for not being added to the default packages.

jrappen:
I changed `is_visible` to `False` unless:

* `mdpopups` can be imported
* and context menu is in Markdown.

jrappen:
@math2001 FYI as your package does something similar, although for earlier versions of Sublime Text.

math2001:
Thanks for the ping @jrappen :+1: 

[trying to be as objective as possible here] I'm not sure if this would be a good thing to include in the default package considering the limited support ST has for html (`table` for example). I think it'd be a bit odd for the end user right now. Installing a third party plugin, it's more acceptable for it to not be perfect. Though having this feature as a default might be nice, from a very markdown-preview-centered perspective, because it would push ST to support more standard markdown elements. I'm not certain it's that important though. But that's just my almost irrelevant opinion. I'd be more than happy for MarkdownLivePreview to become redundant with the default package.

BTW, if you're interested in contributing to MarkdownLivePreview, feel free, there are a few low hanging fruits.

jrappen:
The support seems to currently be limited to output elements listed on:

https://www.sublimetext.com/docs/3/minihtml.html

deathaxe:
1. The external dependencies are a no go for builtin functionality.
2. It doesn't work at this point as all builtin plugins run in python3.8 while PC and all the dependencies only run/support python3.3 
3. Why not just extending MarkdownLivePreview to use html sheets when running on ST4?

jrappen:
1. I am certain that including all that code here would make maintainability worse.
2. That is an issue that Will will be taking care of soon, I assume.
3. Output (and/or Experience) wouldn't change as the limiting factor is Sublime Text with its `minihtml`.

---

I personally use this browser extension:

https://github.com/painyeph/GitLabMarkdownViewer

PR #2268, opened 2020-02-08
[Applescript]Fixed the multi-line comment characters that are used/reported.

Gary-Ash:
The old file was causing the wrong multi-line comment characters to be reported to a package extension i have. This change fixes it


wbond:
Thanks for this fix!

PR #2269, opened 2020-02-08
[Git Formats] Assign git ignore syntax to sparse-checkout

deathaxe:
see: https://www.git-scm.com/docs/git-sparse-checkout


wbond:
@michaelblyons Your comment it saying that this works in multiple situations, right? I just wanted to make sure you weren't suggesting something be changed.

michaelblyons:
Correct. It works for both the normal and the slightly weird versions.

wbond:
Thank you both!

PR #2276, opened 2020-02-16
[Matlab] Fix tmPreferences files

deathaxe:
This commit ...

1. renames the tmPreferences files according to their functions
2. removes duplicated rules
   - Indention rules were found in Indent.tmPreferences and
     Micsellaneous.tmPreferences
3. removes unsupported rules (highlightPairs, smartTypingPairs)


deathaxe:
It appears the meta scopes used in the file are not defined anywhere. Didn't pay attetion to it.  Looks like we can even remove that file for now. The meta scopes for `meta.function` look poor and would  need some rework. But all of that would end up in functions being indexed by the builtin symbol index list.

PR #2277, opened 2020-02-16
[OCaml] Fix tmPreferences files

deathaxe:
This PR ...

1. removes duplicated symbol list files with different names.
2. renames all files according to their function.


wbond:
I think our plan is to move forward with removing symbol manipulation in the default syntaxes.

That said, it appears that some of the scopes here may not coincide with the built-in "kind" detection.

I'll circle back to this later to see what needs to be done in terms of getting scope names in place so we can kill the symbol transformations.

wbond:
From looking over this again, I don't think there is anything actionable for the kind mapping. There is one place where `storage.type.user-type` is used, but considering how few users are OCaml, and that in the future we should have something like https://github.com/sublimehq/sublime_text/issues/3315, I am doing to merge as-is and then strip the symbol transforms.

PR #2278, opened 2020-02-16
[JSON] Fix tmPreferences

deathaxe:
This PR...

1. removes duplicated indention rules, which are already defined in "Indention Rules.tmPreferences".
2. removes the `name` keys from the tmPreferences files to also remove JavaScript relation ships.


FichteFoll:
Imo, grouping these PRs would also make sense because the act of "unifying file names across the repo" is a more or less atomic action that doesn't need to be split into separate packages, although it might lead to conflicts should the files be modified by a different PR.

deathaxe:
Honestly, I started with one single PR. But I found it too complex as I took the chance to slightly tweak formatings or content of some files as well.

PR #2287, opened 2020-02-16
[JavaScript] Fix tmPreferences

deathaxe:
This commit ...

1. renames the tmPreferences according to their function
2. removes the `name` key from tmPreferences files
3. removes symbol transformations to display symbols as collected from the source code
3. merges symbol index and symbol list files and their selectors in order to fix the symbol kind icons. Most functions got the `k` instead of the `f`. (same as for Python)

The goal is a common naming scheme to be applied to all packages, so files of a certain function have the same name in each package.


wbond:
Nice catch fixing the kind info!

PR #2293, opened 2020-02-16
[Scala] Fix tmPreferences

deathaxe:
This commit ...

1. renames the tmPreferences according to their function
2. removes the `name` key
3. removes symbol transformations to display symbols as collected from the source code
4. merges symbol index and symbol list files and their selectors in order to fix the symbol kind icons. Most functions got the `k` instead of the `f`. (same as for Python)

Note: 

This commit fixes a word boundary issue for class identifiers in the syntax to ensure the correct display of symbols.

The goal is a common naming scheme to be applied to all packages, so
files of a certain function have the same name in each package.


djspiewak:
#2226

PR #2303, opened 2020-02-24
[PHP] Add Sort Flag Constants

willrowe:
Sort flags can be passed to functions like [sort](https://www.php.net/manual/en/function.sort.php). The flags [`SORT_FLAG_CASE` and `SORT_NATURAL`](https://www.php.net/manual/en/function.sort.php#refsect1-function.sort-parameters) were [added in PHP v5.4](https://www.php.net/manual/en/function.sort.php#refsect1-function.sort-changelog)


wbond:
Thanks!

PR #2304, opened 2020-02-26
[Git] Add 'bright' versions of colors

michaelblyons:
https://github.com/git/git/commit/87f17d790d8e350e8df1101476882f7751581ee0

The release notes [say](https://github.com/git/git/blob/master/Documentation/RelNotes/2.26.0.txt#L38-L39) "the basic 7 colors" get "bright" modifiers, but there are 8. I think they forgot to count "0" in 0-7.


wbond:
Thanks!

PR #2308, opened 2020-03-12
[C/C++] Fix hexadecimal floating point numbers

deathaxe:
Adds missing patterns to match suffixes of hex floats.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -186,7 +186,7 @@ contexts:
       push:
         - meta_include_prototype: false
         - meta_scope: constant.numeric.float.hexadecimal.c++
-        - match: '{{hex_exponent}}([fFlL])?'
+        - match: '{{hex_exponent}}([fFlL]\b)?'
           captures:
             1: storage.type.numeric.c
           pop: true
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -186,7 +186,9 @@ contexts:
       push:
         - meta_include_prototype: false
         - meta_scope: constant.numeric.float.hexadecimal.c++
-        - match: '{{hex_exponent}}'
+        - match: '{{hex_exponent}}([fFlL])?'
+          captures:
+            1: storage.type.numeric.c
           pop: true
         - match: \.
           scope: punctuation.separator.decimal.c++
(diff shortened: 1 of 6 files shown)


C++/C++.sublime-syntax
```diff
@@ -186,7 +186,9 @@ contexts:
       push:
         - meta_include_prototype: false
         - meta_scope: constant.numeric.float.hexadecimal.c++
-        - match: '{{hex_exponent}}'
+        - match: '{{hex_exponent}}([fFlL])?'
```
FichteFoll: There should still be a `\b` at the end, right?

deathaxe:
Not neccessarily. This is one inconsistency, I've found in different syntaxes. Some match numbers only if terminated by a word boundaries, others don't. In general a number should be matched as such, even if directly followed by a word, if possible.

It helps to keep the editing experience more fluently, even if for some reason the number is temporarily followed by a word. IIRC the C/C++ compilers even complain about invalid trailing characters in such situations, which means the number is interpreted correctly.

I know about some required tweaks at other situations, but this is what I'd have a look at, if I know how to handle the `meta.number` stuff.

FichteFoll:
Not a fan of that particular behavior, since it doesn't follow how tokenization works in most cases and not highlighting the number works better for recognition without being too distracting while typing.

deathaxe:
In general tokenization of a syntax definition used for highlighting should at least try to follow the strategy a compiler/interpreter of a language does, which can be evaluated by passing several kinds of incomplete/invalid numbers to compilers/interpreters and use the error messages to decide how to design the syntax definition - always with the goal to avoid changes in highlighting while typing.

wbond:
This probably makes sense to update for the new numeric scopes, right?

wbond:
I think requiring there be a break makes the most sense for now. I'm open to discussing, but I don't want to delay getting this PR in any longer over this issue, especially since we need to work on the numeric scopes.

deathaxe:
Syntax highlighting without them would follow closer the way C compilers parse those literals. They just throw errors about invalid trailing characters. Python doesn't add `\b` after those, too, which was authored by the one who complained about here ;-).

wbond:
I'm definitely open to a change highlighting anything but those in the suffix as invalid.

PR #2310, opened 2020-03-13
[ShellScript] Fix merged compact flags highlighting

### LINKED ISSUE #2306: [Bash] Flags emphasis inconsistent for "merged" flag  `-o`
- Sublime Version: Sublime Text Build 3211
- OS Version: Ubuntu 16.04.6 LTS

```bash
# Minimum code snippet to demonstrate the issue
set -Eeuo pipefail
```

Flag emphasis is inconsistent for some reason when the `-o` flag is "glued" to others.
image

Bug or feature?

(_I don't remember the term for writing the flags like that; feel free to educate/edit me_)

### LINKED ISSUE #2311: [ShellScript] The `allow-end-of-options-token` setting in auto-generated builtin syntax has no effect
- Sublime Version: 4068
- OS Version: Windows 10

```bash

shift 1 -- -nooption
```

**Expected behavior**

The `--` is not supported by the `shift` command and should therefore not be highlighted.

**Actual behavior**

By including the `scope:source.shell.bash#cmd-args-boilerplate` the `allow-end-of-options-token` machinary seems not to have any effect atm, bacause `cmd-args-boilerplate` contains a rule to match the `--` token.

```yaml
  cmd-args-boilerplate:
    - match: (?={{is_end_of_interpolation}})
      pop: true
    - include: cmd-args-common
    - match: (?:\s+|^)--(?=\s|$)   ; <--- always applies to all commands and all builtins!
      scope: keyword.operator.end-of-options.shell
      set:
        - meta_content_scope: meta.function-call.arguments.shell
        - include: end-of-options-common
```

**Solutions**

1. We could either remove `allow-end-of-options-token` and the included contexts, so the `--` keeps highlighted as a general end-of-options tokens even if a command doesn't need/supports it.

2. We could tweak the contexts in a way which makes `--` being highlighted only if supported by a command, which seems to be the intended behavior.

deathaxe:
Fixes #2306
Fixes #2311

For details, please refer to the commit messages

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -41,7 +41,6 @@ first_line_match: |
 #-------------------------------------------------------------------------------
 variables:
   call_token: \./
-  cmd_boundary: (?=\s|;|$|>|<)
   extension: \.sh
   identifier: '[[:alpha:]_][[:alnum:]_]*'
   identifier_non_posix: '[^{{metachar}}\d][^{{metachar}}=]*'
@@ -410,7 +409,7 @@ contexts:
       pop: true
     - match: (?=\s*#)
       pop: true
-    - include: cmd-args-boilerplate
+    - include: cmd-args-boilerplate-with-end-of-options
     - match: (?=\S)
       push: [vardef-value, vardef-assign, vardef-name]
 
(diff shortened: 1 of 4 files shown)


michaelblyons:
What do you think about adding a `\b` to the beginning of `opt_break`, so that `-` doesn't match until you type a letter?

deathaxe:
I basically like highlighting incomplete tokens in a way that feels like ST knows what I am going to type. So if there I no compelling reason to prevent the `-` from being highlighted already (something like breaking some types of highlighting), I'd keep it as is.

FichteFoll:
But `-` by itself is a valid and frequently used command line parameter that is not an option. This only works for incomplete and invalid tokens.

PR #2312, opened 2020-03-18
[Lisp] Rescopes `parens` to be consistent with the guidelines and Clojure.

s-clerc:
## Changes

1. `parens` is changed from `punctuation.definition`... to `punctuation.section`.
2. `()` is included as a language constant.
3. Tests for both.
## Rationale
1. This is to more strictly [follow the scoping guidelines](https://www.sublimetext.com/docs/3/scope_naming.html#meta) which state that blocks of code delineated in brackets should be scoped in `meta.parens` with `punctuation.section.parens` for the brackets.
  The use of the `.definition` scope led to (perceived) incorrect colouring in Monokai Pro, and this change brings `Lisp.sublime-syntax` in line with `Clojure.sublime-syntax` which currently uses `punctuation.section.parens`.

2. Adding `()` as a `language.constant` because `()` ≡ `nil` for the sake of consistency.
3. By necessity.

### DIFF
--- a/Lisp/syntax_test_lisp.lisp
+++ b/Lisp/syntax_test_lisp.lisp
@@ -87,6 +87,9 @@
 (nil)
 ;^^^ constant.language
 
+()
+;^^^ constant.language
+
 ;#######################
 ; ARITHMETIC OPERATORS #
 ;#######################
--- a/Lisp/syntax_test_lisp.lisp
+++ b/Lisp/syntax_test_lisp.lisp
@@ -37,11 +37,11 @@
 ;                    ^^^ invalid.illegal.stray-bracket-end
 
  (( #| nested |# ))
-;^ meta.group punctuation.definition.group.begin
-; ^ meta.group meta.group punctuation.definition.group.begin
+;^ meta.parens punctuation.section.parens.begin
+; ^ meta.parens meta.parens punctuation.section.parens.begin
 ;         ^ comment
-;                ^ meta.group meta.group punctuation.definition.group.end
-;                 ^ meta.group punctuation.definition.group.end
+;                ^ meta.parens meta.parens punctuation.section.parens.end
+;                 ^ meta.parens punctuation.section.parens.end
 
 ;##########
 ; STRINGS #
--- a/Lisp/syntax_test_lisp.lisp
+++ b/Lisp/syntax_test_lisp.lisp
@@ -87,8 +87,8 @@
 (nil)
 ;^^^ constant.language
 
-()
-;^^^ constant.language
+(())
+;^^ constant.language
 
 ;#######################
 ; ARITHMETIC OPERATORS #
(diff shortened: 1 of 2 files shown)


Lisp/syntax_test_lisp.lisp
```diff
@@ -87,6 +87,9 @@
 (nil)
 ;^^^ constant.language
 
+()
+;^^^ constant.language
```
jwortmann: I expect this test to fail, because the ^ are not aligned correctly. They should be directly underneath the parens:
```
 ()
;^^ constant.language
```

s-clerc: Apologies, it's fixed.

s-clerc:
I think I switched to `parens` because the guidelines aren't too clear, so my thinking was that `group` was more used in cases like the grouping of the arguments after a function `f(x, y z)` or overriding operator precedence `(8 – 2) × 3`. While the former use case is quite common in lisp, the latter isn't. As s-expressions are very general (grouping, but also denoting blocks, assignment, lists, and most other data types) and I felt it was inappropriate to treat them the same way as simple grouping operators.

I'm not a Sublime Syntax expert by any means. If you think that it would be better to use `meta.group` instead, we could change it to that. My main concern really is that if schemes want to colour s-expressions particularly, then it might be difficult as Clojure could use `meta.group.clojure` but Common Lisp uses `meta.group.lisp` and Racket `meta.group.rkt` whereas there is unification under `meta.parens`, although this is under the assumption that Color Schemes cannot use regular expressions.

deathaxe:
Agree with scope naming guidelines being a bit unclear at this point. As I didn't follow/attend the discussion about when which of them is appropriate, I'd propably wait for someone who did.

But with regards to your description, your choice feels more appropriate. I'd use groups for operator precedence like use cases only, as well.

Not sure about your concerns about color schemes as they can use parts of scopes as selectors. If a selector in a color scheme addresses `meta.group`, all scopes beginning with that token would match. Hence all of your examples would (`meta.group.closure`, `meta.group.lisp`, `meta.group.rkt`, etc.).

s-clerc:
> Not sure about your concerns about color schemes as they can use parts of scopes as selectors. If a selector in a color scheme addresses meta.group, all scopes beginning with that token would match. Hence all of your examples would (`meta.group.closure`, `meta.group.lisp`, `meta.group.rkt`, etc.).

I was talking about if a colour scheme wanted to colour things like `print()` in Python differently from `(print)` in Lisp. Although, we could probably solve it if we standardised on `meta.group.lisp.`dialect` `, and then changed the Clojure syntax file to follow this standard.

wbond:
I'd love to converge the work on Lisp by a few different people to a single course of action, if only to try and make the integration process easier. I see you've already been providing feedback on #2387 @s-clerc. Is that in a state where it supersedes this here?

As a comment about `meta.parens`, that is left generally vague as in some situations languages have very clear semantics about things like parens and braces, and in others the lexing is rather ambiguous. Here with lisp it would seem to me that there is probably a fair bit of ambiguity, so I'd opt for `meta.parens` instead of `meta.group`.

deathaxe:
Closing as superseded by #3896.

PR #2313, opened 2020-03-20
[Git Formats] Improve reference matching in messages

FichteFoll:
Added support for FichteForks/Packages@759d6ee style matching (`FichteForks/Packages@759d6ee`), (`FichteForks@759d6ee`) and user/org/repository names with hyphens.


wbond:
I haven't looked at the extent of the conflict, but it appears previous PRs are making it so I can't just merge this.

Git Formats/Git Common.sublime-syntax
```diff
@@ -8,6 +8,8 @@ scope: text.git.common
 
 variables:
   comment_char: '[#;]'
+  hash: \b\h{7,}\b
+  repository_ref: '(?x: \b(?: [\w-]+ (/) )? [\w-]+)'  # user (optional) / repo
```
michaelblyons, line 12: Leave expanded? Or tighten into `(?:\b(?:[\w-]+(/))?[\w-]+)`? Also, do you want more `\b`s, or is it okay to begin/end a repo or user with `-`?

FichteFoll: Seems like repos can have `-` anywhere. 


I don't see a need to compact the regex. I would want to wrap it in a non-capturing group anyway, so I might as well use a couple spaces in there.

PR #2319, opened 2020-03-31
[Lisp] Expand symbol character support

samuel-jimenez:
Expand characters available to symbols.
Allow Goto Definition on constants and variables as well.

### DIFF
--- a/Lisp/Lisp.sublime-syntax
+++ b/Lisp/Lisp.sublime-syntax
@@ -50,7 +50,7 @@ contexts:
       scope: keyword.control.lisp
 
   functions:
-    - match: \b(?i:(defun|defmethod|defmacro))\b\s+([\w\-!?<>]*)
+    - match: \b(?i:(defun|defmethod|defmacro|defvar|defconstant))\b\s+([\w\-!?<>%+*]*)
       scope: meta.function.lisp
       captures:
         1: storage.type.function-type.lisp
--- a/Lisp/Lisp.sublime-syntax
+++ b/Lisp/Lisp.sublime-syntax
@@ -18,6 +18,8 @@ file_extensions:
   - lsp
   - fasl # Scheme dialect of Lisp
 scope: source.lisp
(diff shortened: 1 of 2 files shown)


s-clerc:
Why should globals be in the `entity.`[…]`.functions` scope?

samuel-jimenez:
Good catch; updated.

s-clerc:
Why should globals and constants use the `entity` and `storage` scope at all actually? I just looked at [the docs](https://www.sublimetext.com/docs/3/scope_naming.html#variable) and it seems the appropriate scope would be `variable` but there doesn’t seem to be an appropriate subscope, I’d probably use `variable.other.global` like already done in the file.

For constants `variable.other.const` seems more appropriate.

samuel-jimenez:
In order for Goto Definition to fire.
Currently, this feature will only work with specific scopes. Used in this file are the following:

- entity.name.function
- entity.name.type

There may be others that exist, but documentation is sparse.

Attempting to hack in showInIndexedSymbolList as referenced in [this post](https://forum.sublimetext.com/t/extend-st3-project-tagging/10468/8) allows limited functionality, but only when a symbol is wrapped in parentheses, which is obviously counterproductive for a variable.

s-clerc:
I didn’t realise that; as a temporary solution I guess it’s fine although I worry it’ll mess up syntax highlighting with certain themes such as *Monokai Pro*. Hopefully we can see a better solution for ST4.

wbond:
Is this something that can be converged under #2387? We've got a few different lisp PRs open right now, and it would be great to have the work sort of centralized under the one with the most significant changes. I am sure @michaelblyons would be happy for testing and feedback also.

michaelblyons:
I believe #2387 supersedes this by allowing nearly any character to be used in a symbol. There's a blacklist instead of a whitelist.

deathaxe:
Closing as superseded by #3896.

PR #2320, opened 2020-04-01
[Makefile] Fix embedded ShellScript prototypes

### LINKED ISSUE #2318: [Makefile] Comment inside single quotes
``` lang
.build/vernum: ../meta/version
	sed -i.bak 's/.*automatically updated.*/version = "$(VER)" # automatically updated/' setup.py
```

image

### Environment

  * Windows `10 build 15063 x64`
  * Resolution `1920x1080`
  * dpi_scale `1.0`
  * Sublime Text:
  * Build `3207`
  * 64 bit

*I am using the latest changes in this master branch for the `Makefile` syntax.

deathaxe:
Fixes #2318

### Problem Description

  The Makefile.sublime-syntax includes the Bash.sublime-syntax.

  The `scope:source.shell#prototype` is injected into the embedded Bash.sublime-syntax using `with_prototype` maybe as workaround for https://github.com/sublimehq/sublime_text/issues/1062

  The `with_prototype` command ignores `meta_include_prototype: false`, which is heavily used by Bash.sublime-syntax to prevent the `prototype` context from being included into a bunch of contexts such as strings.

### Fix Description

  As the prototype context is not included into source.shell's main context only, it is not necessary to inject it via `with_protoype`. Hence it is moved to the `shell-body` together with `source.shell`, which is suggested by the core issue as proper workaround.

### Notes:

  The `prototype` context of the Bash.sublime-syntax also includes proper rules to pop off all contexts at eol with respect of line continuation. Hence those don't need to be part of the `with_proptype`  statement as well. They are moved to just pop off the Bash's main context as soon as the eol is reached.

  The `source.shell.embedded` scope is applied to all embedded shell syntax regions.

### DIFF
--- a/Makefile/Makefile.sublime-syntax
+++ b/Makefile/Makefile.sublime-syntax
@@ -602,8 +602,8 @@ contexts:
       set: value-to-be-defined
 
   embedded-shell:
-    - include: scope:source.shell#prototype
     - include: scope:source.shell
+      apply_prototype: true
 
   value-to-be-defined:
     - meta_content_scope: meta.string.makefile string.unquoted.makefile
--- a/Makefile/Makefile.sublime-syntax
+++ b/Makefile/Makefile.sublime-syntax
@@ -325,7 +325,7 @@ contexts:
       set:
         # I don't like putting the string "source.shell" here, as the delegation
         # seeps into this syntax now. But I don't see a cleaner way right now.
-        - meta_scope: source.shell comment.line.number-sign.shell
+        - meta_scope: source.shell.embedded comment.line.number-sign.shell
         - include: comments-pop-on-line-end
     # Otherwise, delegate to the shell syntax (with various extra prototypes).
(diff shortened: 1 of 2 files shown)


Makefile/Makefile.sublime-syntax
```diff
@@ -592,14 +593,18 @@ contexts:
...
+      embed_scope: meta.string.makefile meta.interpolation.makefile source.shell.embedded
       escape: (?!<\\)\2
       escape_captures:
         0: meta.string.makefile meta.interpolation.makefile punctuation.section.interpolation.end.makefile
     - match: ''
       set: value-to-be-defined
 
+  embedded-shell:
+    - include: scope:source.shell#prototype
+    - include: scope:source.shell
```
wbond, line 606: This should be able to use the new `apply_prototype: true` option instead of explicitly including the prototype context on the line above.

PR #2326, opened 2020-04-07
[XML] Fix comment trigger scopes

### LINKED ISSUE #2325: [XML] unable to toggle comments in some circumstances
- Sublime Version: 4070
- OS Version: Ubuntu 19.10

## Expected behavior

It should be possible to comment out a line of XML, or a selected XML element, using the `toggle_comment` command.

## Actual behavior

Nothing happens

## Steps to reproduce

1. Create a new file
1. Set the syntax to XML
1. Type `<hello />`
1. Press <kbd>Home</kbd>
1. Press <kbd>Ctrl+/</kbd>
1. Press <kbd>Ctrl+Shift+/</kbd>
1. Select `<hello />` / press <kbd>Ctrl+A</kbd>
1. Press <kbd>Ctrl+/</kbd>
1. Press <kbd>Ctrl+Shift+/</kbd>
1. See that no commenting happened

Note that <kbd>Ctrl+/</kbd> does work if you have a newline after the end of `<hello />`, and your caret is at the end of the line, or if you have indentation/non-tag text before the start of the tag, and your caret is in that text or at column 0. But if your caret is right before the `<`, or the tag (and only the tag) is selected, then nothing happens.

# Extra info

Caused by https://github.com/sublimehq/Packages/pull/1621/files#diff-a64be0dee389e05e2f9f892168c6354d CC @deathaxe (I think the justification was that [comments are not valid inside tags](https://github.com/sublimehq/Packages/pull/1621#issuecomment-398177798), which is true, but this currently also causes comments to not be toggleable *around* tags.)
I think that ideally, the behavior should be consistent with the HTML syntax (- which currently allows to end up with invalid code, so that perhaps should be tweaked too if we find a better fix?)

deathaxe:
Fixes #2325

This commit includes the punctuation.definition.tag scope into the
comments trigger selector to enable commenting out a tag if

1. the caret is positioned directly before the tag
2. the caret is located at the end of a tag, which also is the end of file as this situation prevents the `meta.tag` scope from being popped off the stack correctly.

As a result the toggle_command does

1. still not work when toggled from within a tag (which is intended behavior) in order to
2. prevent a tag from being messed up by comments.

### DIFF
--- a/XML/Comments.tmPreferences
+++ b/XML/Comments.tmPreferences
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Comments</string>
 	<key>scope</key>
-	<string>text.xml - meta.tag</string>
+	<string>text.xml - (meta.tag - punctuation.definition.tag)</string>
 	<key>settings</key>
 	<dict>
 		<key>shellVariables</key>
--- a/XML/Comments.tmPreferences
+++ b/XML/Comments.tmPreferences
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Comments</string>
 	<key>scope</key>
-	<string>text.xml - (meta.tag - punctuation.definition.tag)</string>
+	<string>text.xml</string>
 	<key>settings</key>
 	<dict>
 		<key>shellVariables</key>


lastsecondsave:
I maybe late for a discussion, but what is the point of limiting user to toggle a comment? If the comment makes a file invalid it's user's mistake and he should adjust the selection. Really hate the idea to move a cursor to be able to put a comment.

AFAIK no other syntax does it.

deathaxe:
This commit actually does not enforce limiting comments, but fixes an issue which caused comments not to be triggered with the curser being placed in front of the tag.

lastsecondsave:
I understand but the scope this fix applies in still not just `text.xml`. Which means at some cursor positions in xml file the `toggle_comment` will do nothing. This will be an unexpected behavior for a lot of people.

deathaxe:
The primary idea of limiting comments was to avoid weired behavior like ...

Animation

... while it indeed looks like the XML syntax definition solves that issue by design and causes comment toggling to work under all circumstances.

Animation

Removing `- meta.tag` would enable something like ...

Animation

I guess this is what you call expected behavior?

As comments can be toggled within tags without the weird behavior of the old syntax (1st screencast), I guess it would be safe to remove all restrictions.

lastsecondsave:
You are toggling block comments in the first two screencasts. But if you toggle line ones with `ctrl+/` Sublime will comment out the whole line leaving the xml totally valid. This what I expect. This what was happening since version 2 and it isn't anymore.

On the last one you now use line comment and yes, this is also what I expect from the _text_ editor. Although even IDEs allow you to insert comments wherever you like.

deathaxe:
> You are toggling block comments in the first two screencasts.

That's the point. All types of comments should work properly under all circumstances. As `ctrl+shift+/` is meant to toggle comments, comments should toggle what is obviously not the case in the first screencast.

> On the last one you now use line comment 

Yes, but it works for block comments as well.

Animation

> even IDEs allow you to insert comments wherever you like

Even IDEs may not be perfect under all circumstances. It's not a valuable argument for or against something at least.

lastsecondsave:
The point with IDE was that even tools which use AST and precisely know at what cursor position the toggle of comment is invalid don't prevent user from doing so. And the benefit of Sublime is that it provides you a lot of help but doesn't try to be too clever like some IDEs do. At least for me.

Thanks for returning the old behavior. People who have to deal with huge xml configs will appreciate that.

PR #2327, opened 2020-04-08
[ShellScript] Recognise Herestrings with no space following the operator 

frou:
Notably, the popular code auto-formatter `shfmt` chooses not to use a space.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -558,7 +558,7 @@ contexts:
         2: keyword.operator.assignment.redirection.shell
 
   redirection-here-string:
-    - match: (\d*)(<<<)\s
+    - match: (\d*)(<<<)
       captures:
         1: constant.numeric.integer.decimal.file-descriptor.shell
         2: keyword.operator.herestring.shell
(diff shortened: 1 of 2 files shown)


ShellScript/Bash.sublime-syntax
```diff
@@ -558,7 +558,7 @@ contexts:
         2: keyword.operator.assignment.redirection.shell
 
   redirection-here-string:
-    - match: (\d*)(<<<)\s
+    - match: (\d*)(<<<)
```
FichteFoll, line 561: I believe we can still do something useful with e.g. a (negative) look-ahead here since shell operators (e.g. `<` or `&`) immediately following the `<<<` token are not allowed.

wbond: The idea being that the match should not occur if there is a `<<<<` or `<<<&`?

FichteFoll: Exatly.

deathaxe: The PR is ready to discuss. It superseeds all open PRs. So the question is just. Merge the open PRs, which are in a good shape first, or just close them in favor of the whole big patchset?

wbond: @deathaxe It seems this is the only open ShellScript PR, so I'd say open a PR with your changes if they include the fix here and the extra check that @FichteFoll pointed out.

deathaxe: Yes. I'll collect the issues to close and create a PR, hopefully soon.

wbond:
This should be fixed, with the extra handling discussed by the changes in #2423

PR #2330, opened 2020-04-10
[*](ci): Remove '.git' folder, move verbosely

jrappen:


.github/workflows/ci.yml
```diff
@@ -17,9 +17,11 @@ jobs:
         run: mv st_syntax_tests/syntax_tests ./
       - name: Cleanup syntax_tests dir
         run: rm -R st_syntax_tests*
+      - name: Cleanup git dir
+        run: rm -R .git*
```
FichteFoll, line 21: Instead of removing the git folder, I would rather explicitly unselect it in the `find` call later.

jrappen:
closing as stale

PR #2335, opened 2020-04-17
[PHP] Add highlight for multiline regex string

jfcherng:
Our PHP syntax has already detected single/double-quoted regex strings which are written in a single line. Thanks to `branch_point` in ST 4, this PR makes it also detect those which are written in multiple lines.

```php
$regex = '/
    a{0,6}
//   ^^^^^ keyword.operator.quantifier.regexp
/ux';

$regex = '/foo?/ux';
//            ^ keyword.operator.quantifier.regexp

$not_regex = 'foo?';
//               ^ string - source.regexp

$not_regex = '/foo?';
//                ^ string - source.regexp

// there is no "T" regex modifier
$not_regex = '/foo?/uTx';
//                ^ string - source.regexp
```

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -1465,43 +1465,63 @@ contexts:
       scope: meta.tag.inline.phpdoc.php
       captures:
         1: keyword.other.phpdoc.php
-  regex-double-quoted:
-    - match: (?="/(\\.|[^"/])+/[eimsuxADJSUX]*")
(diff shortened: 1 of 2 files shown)


jfcherng:
I would like to make `\b[eimsuxADJSUX]*\b` as a variable `regex_modifier` but it results in https://github.com/sublimehq/sublime_text/issues/1488.

jfcherng:
Look like I have to wait for the syntax test runner update and re-push :smile:

PR #2256, opened 2020-01-21
[Groovy] Adding first_line_match to help identification

arjhall:
This will help when editing files with no extension, such as those groovy config files commonly used for Jenkins.

PR #2336, opened 2020-04-17
[Groovy] Adding Groovy detection using shebang line

oshmyrko:
This will allow Sublime Text to detect syntax for files without extension but with specified shebang line, such as Jenkinsfiles with non-standard names (e.g. `Jenkinsfile-build`, `Jenkinsfile-deploy`, etc).

Tested locally (`~/.config/sublime-text-3/Packages/User/Groovy.sublime-syntax`) with different types of shebang lines:
* `#!/usr/bin/env groovy`
* `#!/usr/bin/groovy`
* `#! groovy`
* `#!groovy`
* etc

### DIFF
--- a/Groovy/Groovy.sublime-syntax
+++ b/Groovy/Groovy.sublime-syntax
@@ -7,6 +7,9 @@ file_extensions:
   - gvy
   - gradle
   - Jenkinsfile
+
+first_line_match: ^#!.*\bgroovy\b
+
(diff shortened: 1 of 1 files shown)


oshmyrko:
This is similar to #2256, but allows different shebang lines.

wbond:
Thanks for submitting this enhancement!

PR #2339, opened 2020-04-19
[Markdown] Add YAML frontmatter highlighting

### LINKED ISSUE #1783: Highlight Frontmatter in Markdown as YAML
Markdown documents may occasionally have a so-called "frontmatter" that is frequently using YAML syntax.

Using an embed, highlighting of this should be trivial.

Before settling onto YAML, however, have you seen other kinds of frontmatter in Markdown documents and if yes, which syntax do they use and how are they separated from the body?

FichteFoll:
Because this may only match at the very beginning of the file, it cannot be tested. I tried.

Closes #1783.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -180,7 +180,21 @@ variables:
     #     \s*$      # any amount of whitespace until EOL
     #   )
 contexts:
+  file-start:
+    - match: ^---$\n
+      scope: punctuation.section.block.begin.frontmatter.yaml.markdown
+      embed: scope:source.yaml
+      embed_scope: source.yaml.embedded
+      escape: ^---(?=\s)
+      escape_captures:
+        0: punctuation.section.block.end.frontmatter.yaml.markdown
+
   main:
+    - include: file-start
+    - match: ''
+      push: default
+
+  default:
     - match: |-
         (?x)^
         (?=  {{block_quote}}
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -181,20 +181,22 @@ variables:
     #   )
 contexts:
   file-start:
-    - match: ^---$\n
-      scope: punctuation.section.block.begin.frontmatter.yaml.markdown
+    - match: (---)\s*
+      captures:
+        0: meta.frontmatter.markdown
+        1: punctuation.section.block.begin.frontmatter.yaml.markdown
       embed: scope:source.yaml
-      embed_scope: source.yaml.embedded
-      escape: ^---(?=\s)
+      embed_scope: meta.frontmatter.markdown source.yaml.embedded
+      escape: ^(---)\s*$  # pandoc requires the remainder of the line to be blank
       escape_captures:
-        0: punctuation.section.block.end.frontmatter.yaml.markdown
+        1: meta.frontmatter.markdown punctuation.section.block.end.frontmatter.yaml.markdown
 
   main:
     - include: file-start
     - match: ''
-      push: default
+      set: markdown
 
-  default:
+  markdown:
     - match: |-
         (?x)^
         (?=  {{block_quote}}
--- a/Markdown/MultiMarkdown.sublime-syntax
+++ b/Markdown/MultiMarkdown.sublime-syntax
@@ -1,6 +1,7 @@
 %YAML 1.2
 ---
 # http://www.sublimetext.com/docs/3/syntax.html
+version: 2
 name: MultiMarkdown
 first_line_match: (?i)^format:\s*complete\s*$
 scope: text.html.markdown.multimarkdown
@@ -25,4 +26,12 @@ contexts:
     - match: ''
       push:
         - meta_scope: meta.content.multimarkdown
-        - include: scope:text.html.markdown
+        - match: ''
+          embed: scope:text.html.markdown
+          escape: $(?!$)  # should never match
+    # Replace with the following once
+    # https://github.com/sublimehq/sublime_text/issues/2395#issuecomment-662515815 is fixed.
+    # - match: ''
+    #   embed: scope:text.html.markdown
+    #   embed_scope: meta.content.multimarkdown
+    #   escape: $(?!$)  # should never match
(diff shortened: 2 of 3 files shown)


michaelblyons:
Can you make a test to ensure that it _doesn't_ match further down in the file?

FichteFoll:
There are a couple tests regarding the `thematic-break` token (which is what these would normally be matched as), but none specifically with exactly three hyphens. I guess I could add a test for that since I already verified this manually.

Markdown/Markdown.sublime-syntax
```diff
@@ -180,7 +180,21 @@ variables:
     #     \s*$      # any amount of whitespace until EOL
     #   )
 contexts:
+  file-start:
+    - match: ^---$\n
```
deathaxe, line 184: Using `^` here might cause some issues if Markdown is embedded in another syntax for some reason. Using it here would also mean to use it in the push statement for `default`.

If `\n` is matched without the need to detect the end-of-file, the `$` before can be omitted.

Result: 
```
- match: ---\n
```

Markdown/Markdown.sublime-syntax
```diff
@@ -180,7 +180,21 @@ variables:
...
+      scope: punctuation.section.block.begin.frontmatter.yaml.markdown
```
deathaxe, line 185: Do we really want to go the way to add such subscopes?

There is no corresponding meta scope for frontmatters.

Result:

```
    scope: meta.frontmatter.markdown punctuation.section.frontmatter.begin.markdown
```

Markdown/Markdown.sublime-syntax
```diff
@@ -180,7 +180,21 @@ variables:
...
+      embed: scope:source.yaml
+      embed_scope: source.yaml.embedded
```
deathaxe, line 187: Would need to add `meta.frontmatter.markdown` here, too.

Markdown/Markdown.sublime-syntax
```diff
@@ -180,7 +180,21 @@ variables:
...
+      escape: ^---(?=\s)
+      escape_captures:
+        0: punctuation.section.block.end.frontmatter.yaml.markdown
```
deathaxe, line 190: Same here.

```
      escape_captures:
        0: meta.frontmatter.markdown punctuation.section.frontmatter.end.markdown
```

Markdown/Markdown.sublime-syntax
```diff
@@ -180,7 +180,21 @@ variables:
...
+
   main:
+    - include: file-start
+    - match: ''
+      push: default
```
deathaxe, line 195: We can set into the `default` context as the embed-escape issue was resolved.

I'd call the main context `markdown` though.

Markdown/Markdown.sublime-syntax
```diff
@@ -180,7 +180,21 @@ variables:
...
+    - match: ^---$\n
+      scope: punctuation.section.block.begin.frontmatter.yaml.markdown
+      embed: scope:source.yaml
+      embed_scope: source.yaml.embedded
+      escape: ^---(?=\s)
+      escape_captures:
+        0: punctuation.section.block.end.frontmatter.yaml.markdown
+
   main:
+    - include: file-start
```
deathaxe, line 193: Not sure if it is really worth adding a named context for just one match rule.

FichteFoll: The name of the context helps convey the semantic meaning of the match, imo, so I left that in.

wbond:
One the test failure is fixed, this should be good to merge

Markdown/MultiMarkdown.sublime-syntax
```diff
@@ -25,4 +26,12 @@ contexts:
     - match: ''
       push:
         - meta_scope: meta.content.multimarkdown
-        - include: scope:text.html.markdown
+        - match: ''
+          embed: scope:text.html.markdown
+          escape: $(?!$)  # should never match
```
wbond, line 31: Why an embed here that never exits? This will add overhead where we keep track of a state-based Oniguruma regex that is run on every line, before the regular matching is done.

Markdown/MultiMarkdown.sublime-syntax
```diff
@@ -25,4 +26,12 @@ contexts:
...
+    # Replace with the following once
+    # https://github.com/sublimehq/sublime_text/issues/2395#issuecomment-662515815 is fixed.
+    # - match: ''
+    #   embed: scope:text.html.markdown
+    #   embed_scope: meta.content.multimarkdown
+    #   escape: $(?!$)  # should never match
```
wbond, line 37: I _think_ this will result in the same scoping, without adding an Oniguruma overhead. It pushes an extra context down so when markdown does a `set` from `main` it will retain the `meta_scope`.

```suggestion
        - match: ''
          push:
            - include: scope:text.html.markdown
```

Markdown/MultiMarkdown.sublime-syntax
FichteFoll: It seemed like the proper directive semantically, except that it doesn't have the ability to never exit. I agree that having to run a never-matching pattern for every line is suboptimal, however, especially if it's not using sregex.
I initially went that route because the only other alternative I could think of was to add another nesting level so the markdown context doesn't set out of the context with `meta_scope`, but using a multi-push is much more elegant.

Markdown/MultiMarkdown.sublime-syntax
FichteFoll: That is correct, but I simply replaced it with a multi-push and an anonymous context. 🙃

wbond:
LGTM, thanks!

Markdown/MultiMarkdown.sublime-syntax
```diff
@@ -24,5 +25,5 @@ contexts:
           scope: string.unquoted.multimarkdown
     - match: ''
       push:
-        - meta_scope: meta.content.multimarkdown
-        - include: scope:text.html.markdown
+        - - meta_scope: meta.content.multimarkdown
```
deathaxe, line 28: Is that style a new feature in version 2?

FichteFoll: No, that was possible before.

michaelblyons: Yeah, but Will told Thom not to do it for complicated stuff.

deathaxe: I know about chaining anonymous contexts via `- - ...`. I did not expect anonymous contexts with only a `meta_scope` derective to be valid.

PR #2341, opened 2020-04-20
[Markdown] Add YAML syntax highlight to code blocks

TheSecEng:
Add yaml syntax highlight to code blocks

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -1317,6 +1317,21 @@ contexts:
     - match: |-
          (?x)
           {{fenced_code_block_start}}
+          ((?i:yaml))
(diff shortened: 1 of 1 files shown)


FichteFoll:
Previously, addition of new embedded languages for markdown was [put on hold](https://github.com/sublimehq/Packages/pull/1770#issuecomment-440709348) in order not to grow it even more, but maybe with YAML front matter from #2339, adding that at another location wouldn't hurt as much.

deathaxe:
IMHO, yaml is such a common and basic syntax that it should be supported. There are other less common syntaxes to argue about.

wbond:
Now that we have lazy embeds, this incurs effectively no overheard.

PR #2347, opened 2020-04-21
[Scala] Added newline bailout for annotation type tails

djspiewak:
Annotations *at* end-of-line within a single type expression would result in the subsequent line(s) being incorrectly scoped due to the way the lookahead was being processed. Catching this case explicitly allows us to (correctly) pop out of the context.


Scala/syntax_test_scala.scala
```diff
@@ -2202,3 +2202,9 @@ foo eq bar
 
 foo ne bar
 //  ^^ keyword.operator.comparison.scala
+
+{
+   type Apply = F @uncheckedVariance
+   type A
+// ^^^^ storage.type.scala
+//      ^ entity.name.type.scala
```
wbond, line 2210: Should there be a `}` on the last line?

djspiewak: Very likely yes! It's been a while since I looked at this PR. I'll try to update it.

Issue #2349, opened 2020-04-23
[JavaScript] Optional chaining does not work properly


ghost:
Examples:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```js
object?.key;
object?.[key];
object?.(key);
```

Actual:

![](https://user-images.githubusercontent.com/64195804/80082748-13da2800-8555-11ea-9fc5-fa0311094cde.png)

Expected:

![](https://user-images.githubusercontent.com/64195804/80082763-176daf00-8555-11ea-85c2-1110e94d69a2.png)

or

![](https://user-images.githubusercontent.com/64195804/80082771-19d00900-8555-11ea-9365-4015293edaec.png)

Introduced in https://github.com/sublimehq/Packages/pull/2132

/cc author @Thom1729

jfcherng:
In #2132, it uses `.?` as the accessor. But the correct one should be `?.`.

fastcoder15:
Given that many JavaScript engines already support this feature, can we expect a reasonably fast fix? @Thom1729

PR #2350, opened 2020-04-28
[JavaScript] Fix optional chaining typos

jfcherng:
Fixes https://github.com/sublimehq/Packages/issues/2349#issuecomment-618329776

---

@Thom1729 feel free to take over this since you definitely have better understanding than me in JS syntax definition rules to do this PR

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -42,9 +42,10 @@ variables:
 
   non_reserved_identifier: (?:(?!{{reserved_word}}){{identifier}})
 
+  dot_accessor: (?:\??\.)
   either_func_lookahead: (?:{{func_lookahead}}|{{arrow_func_lookahead}})
   binding_pattern_lookahead: (?:{{identifier}}|\[|\{)
-  left_expression_end_lookahead: (?!\s*[.\[\(])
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -1362,7 +1356,7 @@ contexts:
   function-assignment:
     - match: |-
         (?x)(?=
-          (?:{{identifier}} \s* \. \s*)*
+          (?:{{identifier}} \s* {{dot_accessor}} \s*)*
```
fastcoder15, line 1359: This is incorrect. Expression involving optional chaining cannot be a Lvalue.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -1192,7 +1186,7 @@ contexts:
       push: expression-begin
 
   ternary-operator:
-    - match: '\?'
+    - match: '\?(?=[^.]|\.[0-9])'
```
jfcherng, line 1189:
```js
let a = true ?.5 : .1; // this is not optional chaining but number 0.5
```

JavaScript/JavaScript.sublime-syntax
jfcherng: nice catch!

PR #2354, opened 2020-05-04
[Regex] Sublime internal file patterns

michaelblyons:
Minimal syntax for ST [file patterns](https://www.sublimetext.com/docs/file_patterns.html). Like #2214 does with Regex Replace, this is for internal use if/when the "Where" (which files) portion of "Find in Files" can be scoped.

Also provides an entry point for things like PackageDev's [Project.sublime-syntax](https://github.com/SublimeText/PackageDev/blob/master/Package/Sublime%20Text%20Project/Sublime%20Text%20Project.sublime-syntax) that might not want the `,` included. ~(Although the bits anchored with `^` probably won't work either as JSON values.)~ _[Fixed?]_

I tried to distinguish the `/*` and `*/` behavior from the normal `*` behavior so that people might notice it works differently.

### DIFF
--- a/Regular Expressions/File Pattern.sublime-syntax
+++ b/Regular Expressions/File Pattern.sublime-syntax
@@ -1,0 +1,38 @@
+%YAML 1.2
+---
+# https://www.sublimetext.com/docs/file_patterns.html
+name: Sublime File Pattern
+file_extensions:
+  - sublime-file-patterns
+hidden: true
+scope: source.file-pattern
+
+contexts:
+  main:
+    - include: multiple
+
+  multiple:
+    - match: (,)\s*(?:(-)|(//))?
+      captures:
+        1: punctuation.separator.file-pattern
+        2: keyword.operator.logical.file-pattern
+        3: constant.language.file-pattern
+    - include: single
+
+  single:
+    - match: \*{2,}
+      scope: invalid.illegal.file-pattern
+    - match: \[|\]
+      scope: invalid.illegal.file-pattern
+    - match: ^\s*(?:(-)|(//))
+      captures:
+        1: keyword.operator.logical.file-pattern
+        2: constant.language.file-pattern
+    - match: (\*)/|/(\*)(?!\*)
+      captures:
+        1: variable.language.wildcard.file-pattern
+        2: variable.language.wildcard.file-pattern
+    - match: \*
+      scope: keyword.operator.wildcard.file-pattern
+    - match: \?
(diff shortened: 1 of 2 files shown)


Regular Expressions/File Pattern.sublime-syntax
```diff
@@ -0,0 +1,38 @@
...
+        2: keyword.operator.logical.file-pattern
+        3: constant.language.file-pattern
+    - include: single
+
+  single:
+    - match: \*{2,}
+      scope: invalid.illegal.file-pattern
+    - match: \[|\]
+      scope: invalid.illegal.file-pattern
+    - match: ^\s*(?:(-)|(//))
```
FichteFoll, line 27: This won't match project-relative patterns at any position that is not the first, or exclusive patterns for that matter because of the `^` anchor. You'll need to restructure the to match certain characters only at the first position, for example through a multi-push on `(?=[^\s,])`. (I'm not sure how white-space-sensitive patterns are.)

Of course, you should also add tests for these cases.

michaelblyons: How about now?

keith-hall:
is it worth considering having a separate package for this instead of putting them in the Regular Expressions package?

deathaxe:
With the discussion about find/replace related regexp syntaxes at https://github.com/sublimehq/Packages/pull/2214, I'd say it is in the right package.

Regular Expressions/File Pattern.sublime-syntax
```diff
@@ -0,0 +1,39 @@
+%YAML 1.2
+---
+# https://www.sublimetext.com/docs/file_patterns.html
+name: Sublime File Pattern
+file_extensions:
+  - sublime-file-patterns
```
wbond, line 6: I feel like this will add future confusion, since there isn't such a file format.

michaelblyons: Makes sense. I'll drop it.

michaelblyons:
What do you think about adding "builtins" for `<open folders>`, `<open files>`, and `<current file>`?

wbond:
Having them and `<project filters>` scopes some way special would be good, but I haven't spent time thinking about what scope yet.

michaelblyons:
> if/when the "Where" (which files) portion of "Find in Files" can be scoped.

I'm not sure which build added this, but I have noticed it. Thanks HQ!

PR #2355, opened 2020-05-05
[Clojure] Support ClojureScript as its own language scope.

### LINKED ISSUE #2351: [Clojure] Supporting ClojureScript
ST supports ClojureScript by treating it as if it was Clojure (https://github.com/sublimehq/Packages/blob/master/Clojure/Clojure.sublime-syntax#L9). For the purposes of highlighting the source, this is probably the right choice.

Unfortunately it complicates tooling that wants to interact with Clojure vs ClojureScript. Both [Zeal](https://github.com/vaanwd/Zeal) and [SublimeLinter-contrib-clj-kondo](https://github.com/ToxicFrog/SublimeLinter-contrib-clj-kondo) _[^1]_ use the scope to determine what the contents of the file are and how to resolve them; meaning that they don't quite work for ClojureScript out of the box. [SublimeREPL](https://github.com/wuub/SublimeREPL) works around the issue by including a [partial definition of ClojureScript](https://github.com/wuub/SublimeREPL/blob/master/config/ClojureScript/clojurescript.tmLanguage).

Another complication is that ".cljc" files are either Clojure or ClojureScript depending on how/where they are executed. And the correct way to handle ".edn" files as they aren't quite Clojure or ClojureScript but a subset.

As an experiment, I copied https://github.com/sublimehq/Packages/blob/master/Clojure/ into /Packages/ClojureScript and renamed all references to "Clojure" or "clojure" to "ClojureScript" and "clojurescript" respectively. This appeared to work as I desired but is likely the wrong choice for a number of reasons (including pointless duplication of code).

I expect there is at least 1 cleaner way to do the above (probably embedding the Clojure syntax in ClojureScript) and maybe another (I've seen ".tmPreferences" that seem to do a similar embedding).

Anyway, I am trying to get a feel for the right way to proceed here. I'd like to see ST natively support ClojureScript, but it was handled differently than that so don't want to break things either. I can submit a PR to add ClojureScript. I can make a standalone ClojureScript package. I can work with the package authors to deal with seeing a Clojure scope when seeing ClojureScript code. There are probably other options I haven't considered.

[^1]: https://github.com/ToxicFrog/SublimeLinter-contrib-clj-kondo/issues/1

oconnor0:
Should resolve #2351


Clojure/Comment.tmPreferences
```diff
@@ -4,7 +4,7 @@
 	<key>name</key>
 	<string>Comment</string>
 	<key>scope</key>
-	<string>source.clojure</string>
```
michaelblyons, line 7: This works fine, but note that _usually_ commas are used unless there is a multilevel command that needs an ORed context.

oconnor0: I will get that fixed.

michaelblyons: Probably also in keymap? I don't often map keys, so I'm less familiar with those.

oconnor0: I updated both of them to use `,` instead of `|`.

Clojure/tests/syntax_test_clojurescript_old.cljs
```diff
@@ -0,0 +1,103 @@
+; SYNTAX TEST "Packages/Clojure/ClojureScript.sublime-syntax"
```
FichteFoll, line 1: Since the syntax just imports the main clojure syntax, I don't see a need to duplicate the entire test suite. It would also need to be kept in sync.

Clojure/ClojureScript.sublime-syntax
```diff
@@ -0,0 +1,14 @@
+%YAML 1.2
+---
+# https://www.sublimetext.com/docs/3/syntax.html
+
+name: ClojureScript
+scope: source.clojurescript
```
FichteFoll, line 6: Using `source.clojure.script` would be more desirable, as this is not only subfamily of clojure but actually the same syntax (at least currently). Linters explicitly targetting the version on Java could exclude `source.clojure.script` from their selector, as could Zeal, while features that apply to all languages of the clojure family (such as the comment preferences below) only need to target `source.clojure`.

Clojure/tests/syntax_test_clojurescript_old.cljs
oconnor0: Sound good.

Clojure/ClojureScript.sublime-syntax
oconnor0: Thoughts on `source.clojure.clojurescript` or `source.clojure.cljs` vs `source.clojure.script`?

FichteFoll: I could see `source.clojure.clojurescript` as a more descriptive version, though I find it hard to believe that `source.clojure.script` would not be understood by anyone reading it. Still, I never used either. Definitely not `source.clojure.cljs`, though.

oconnor0: `source.clojure.clojurescript` made sense to me. And it avoids having to modify plugins.

PR #2357, opened 2020-05-06
[HTML] WIP Entity completion

michaelblyons:
This is very WIP. It's just a naive parse of the entity list at w3.org. I'm hoping to add the symbols themselves in the "Details" if they're printable.


michaelblyons:
Now that I think about it, I should add a listener to `&` in html_completions.py instead.

FichteFoll:
And use `html.entities`, I hope. 🙂

PR #2359, opened 2020-05-08
[HTML] Entity completion

michaelblyons:
You have to add `&` in your HTML/XML `auto_complete_triggers` to see it working. Thanks to @FichteFoll for the `html.entities` tip.

Apparently, I cannot reopen the other PR after force-pushing. 🤷

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -1,7 +1,6 @@
 import sublime, sublime_plugin
-import re
-
 import html.entities
+import re
 
 
 def match(rex, str):
@@ -19,15 +18,6 @@ def make_completion(tag, completion=None):
         completion=completion,
         completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
         kind=(sublime.KIND_ID_MARKUP, 't', 'Tag')
-    )
-
-def make_entity_completion(entity, rendered_char=None):
-    return sublime.CompletionItem(
-        trigger=entity,
-        completion=entity if entity.endswith(';') else f'{entity};',
-        completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
-        kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
-        details=f'Renders as <code>{rendered_char}</code>' if rendered_char else None
     )
 
 def get_tag_to_attributes():
@@ -280,18 +270,30 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
                 completion='#${1:00};',
                 completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
                 kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
-                details='Base-10 Unicode',
+                details='Base-10 Unicode character',
             ),
             sublime.CompletionItem(
                 trigger='#x0000;',
                 completion='#x${1:0000};',
                 completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
                 kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
-                details='Base-16 Unicode',
+                details='Base-16 Unicode character',
             ),
         ]
-        for ent in html.entities.html5:
-            completions.append(make_entity_completion(ent, html.entities.html5[ent]))
+
+        completions += [
+            sublime.CompletionItem(
+                trigger=entity,
+                annotation=printed,
+                completion=entity,
+                completion_format=sublime.COMPLETION_FORMAT_TEXT,
+                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
+                details=f'Renders as <code>{printed}</code>',
+            )
+            for entity, printed in html.entities.html5.items()
+            if entity.endswith(';')
+        ]
+
         return completions
 
     def default_completion_list(self):
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -1,6 +1,8 @@
 import sublime, sublime_plugin
 import re
 
+import html.entities
+
 
 def match(rex, str):
     m = rex.match(str)
@@ -17,6 +19,15 @@ def make_completion(tag, completion=None):
         completion=completion,
         completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
         kind=(sublime.KIND_ID_MARKUP, 't', 'Tag')
+    )
+
+def make_entity_completion(entity, rendered_char=None):
+    return sublime.CompletionItem(
+        trigger=entity,
+        completion=entity if entity.endswith(';') else f'{entity};',
+        completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
+        kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
+        details=f'Renders as <code>{rendered_char}</code>' if rendered_char else None
     )
 
 def get_tag_to_attributes():
@@ -231,6 +242,11 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
         # print('ch:', ch)
 
         completion_list = []
+
+        if ch == '&':
+            completion_list = self.entity_completion_list()
+            return (completion_list, sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS)
+
         if is_inside_tag and ch != '<':
             if ch in [' ', '\t', '\n']:
                 # maybe trying to type an attribute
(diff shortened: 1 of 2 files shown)


HTML/html_completions.py
```diff
@@ -257,6 +273,27 @@ def get_completions(self, view, prefix, locations, is_inside_tag):
...
+            ),
+            sublime.CompletionItem(
+                trigger='#x0000;',
+                completion='#x${1:0000};',
+                completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
+                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
+                details='Base-16 Unicode',
+            ),
+        ]
+        for ent in html.entities.html5:
```
FichteFoll, line 293: Use `.items()` to iterate over key-value pairs.

deathaxe:
Looks like there are several duplicates in the entities list (e.g.: `'AElig', 'AElig;', 'AMP', 'AMP;', 'Aacute', 'Aacute;',`). Need some kind of dedubbing? Add missing semicolon and dedup via set()

HTML/html_completions.py
```diff
@@ -19,6 +21,15 @@ def make_completion(tag, completion=None):
         kind=(sublime.KIND_ID_MARKUP, 't', 'Tag')
     )
 
+def make_entity_completion(entity, rendered_char=None):
+    return sublime.CompletionItem(
+        trigger=entity,
+        completion=entity if entity.endswith(';') else f'{entity};',
+        completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
```
deathaxe, line 28: It should be safe to use `COMPLETION_FORMAT_TEXT` here.

rwols: Is the implication of `COMPLETION_FORMAT_TEXT` that things like `${1:foo}` are inserted into the buffer verbatim?

deathaxe:
I'd propably just ...

```python
    def entity_completion_list(self):
        completions = [
            sublime.CompletionItem(
                trigger='#00;',
                completion='#${1:00};',
                completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
                details='Base-10 Unicode',
            ),
            sublime.CompletionItem(
                trigger='#x0000;',
                completion='#x${1:0000};',
                completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
                details='Base-16 Unicode',
            ),
        ]

        dedupped = set()
        for entity, printed in html.entities.html5.items():
            if not entity.endswith(';'):
                entity += ';'
            if entity not in dedupped:
                dedupped.add(entity)
                completions.append(
                    sublime.CompletionItem(
                        trigger=entity,
                        completion=entity,
                        completion_format=sublime.COMPLETION_FORMAT_TEXT,
                        kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
                        details=f'Renders as <code>{printed}</code>'
                    )
                )

        return completions
```

FichteFoll:
The doc says:

> Note that the trailing semicolon is included in the name (e.g. 'gt;'), however some of the names are accepted by the standard even without the semicolon: in this case the name is present with and without the ';'.

So it should be okay to just drop the variants without a semicolon.

As a suggestion, I would add *just* the represented character as the annotation.

michaelblyons:
Thank you, all!

HTML/html_completions.py
```diff
@@ -257,6 +263,39 @@ def get_completions(self, view, prefix, locations, is_inside_tag):
...
+                completion_format=sublime.COMPLETION_FORMAT_TEXT,
+                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
+                details=f'Renders as <code>{printed}</code>',
+            )
+            for entity, printed in html.entities.html5.items()
+            if entity.endswith(';')
+        ]
+
+        return completions
+
```
deathaxe, line 298: Doesn't make a big difference, but how about

```python
    def entity_completion_list(self):
        return [
            sublime.CompletionItem(
                trigger='#00;',
                completion='#${1:00};',
                completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
                details='Base-10 Unicode character',
            ),
            sublime.CompletionItem(
                trigger='#x0000;',
                completion='#x${1:0000};',
                completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
                details='Base-16 Unicode character',
            ),
        ] + [
            sublime.CompletionItem(
                trigger=entity,
                annotation=printed,
                completion=entity,
                completion_format=sublime.COMPLETION_FORMAT_TEXT,
                kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
                details=f'Renders as <code>{printed}</code>',
            )
            for entity, printed in html.entities.html5.items()
            if entity.endswith(';')
        ]
```

deathaxe: By defining a class-level or global attribute for `(sublime.KIND_ID_MARKUP, 'e', 'Entity')` the time needed to compute completions could be slightly reduced.

```python
class HtmlTagCompletions(sublime_plugin.EventListener):
    """
    Provide tag completions for HTML
    It matches just after typing the first letter of a tag name
    """

    KIND_ENTITY = (sublime.KIND_ID_MARKUP, 'e', 'Entity')
    ...

Comment (FichteFoll) [reply HTML/html_completions.py:298]: Then I want to suggest `return `["#00;", "#x0000;", *(comprehension)]` for  nice optimizations.

**Comment by wbond** [comment]
The two snippet completions feel a little strange in that they are presented the same way as the literal text completions.

However, it may just be me getting used to having "proper" kind info on snippets.

**Comment by deathaxe** [comment]
You would expect `s`?

**Comment by wbond** [comment]
Yeah, I like it that it stays as an `e` with the name `Entity`, but the green to imply snippet feels like an improvement.

File: HTML/html_completions.py [hunk 4]
```diff
@@ -190,6 +191,9 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
     Provide tag completions for HTML
     It matches just after typing the first letter of a tag name
     """
+
+    KIND_ENTITY = (sublime.KIND_ID_MARKUP, 'e', 'Entity')
+
```
deathaxe, line 196: Maybe add a `KIND_SNIPPET = (sublime.KIND_ID_SNIPPET, 'e', 'Entity')` here.

May also be named `KIND_ENTITY_SNIPPET`?

HTML/html_completions.py
```diff
@@ -257,6 +266,36 @@ def get_completions(self, view, prefix, locations, is_inside_tag):
 
         return (completion_list, flags)
 
+    def entity_completion_list(self):
+        return [
+            sublime.CompletionItem(
+                trigger='#00;',
+                completion='#${1:00};',
+                completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
+                kind=self.KIND_ENTITY,
```
deathaxe, line 275: And replace `self.KIND_ENTITY` by `self.KIND_SNIPPET` here.

HTML/html_completions.py
```diff
@@ -257,6 +266,36 @@ def get_completions(self, view, prefix, locations, is_inside_tag):
...
+                details='Base-10 Unicode character',
+            ),
+            sublime.CompletionItem(
+                trigger='#x0000;',
+                completion='#x${1:0000};',
+                completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
+                kind=self.KIND_ENTITY,
```
deathaxe, line 282: ... and here.

HTML/html_completions.py
michaelblyons: I think I did these slightly differently than you did. Any objections?

HTML/html_completions.py
```diff
@@ -288,7 +288,7 @@ def entity_completion_list(self):
                     annotation=printed,
                     completion=entity,
                     completion_format=sublime.COMPLETION_FORMAT_TEXT,
-                    kind=self.KIND_ENTITY,
+                    kind=(sublime.KIND_ID_MARKUP, 'e', 'Entity'),
```
deathaxe, line 291: This now causes the tuple to be created for each entity in the list instead of using a cached one. The class level attribute's reason was to avoid that.

michaelblyons: Right. I forgot. Thanks.

wbond:
Awesome, thanks!

michaelblyons:
Sure thing. Do you want an issue on sublimehq/sublime_text to update the default `auto_complete_triggers` setting?

wbond:
Yes, that would be great, thank you!

PR #2362, opened 2020-05-16
[Markdown] Separate bibliography reference links

### LINKED ISSUE #2360: [Markdown] Need GoTo separation between bibliography links
- Sublime Version: 4074
- OS Version: macOS 10.15.4

## Expected behavior

Consecutive bibliography links are noted separately in the local Go-To entry

## Actual behavior

The Go-To scope used for bibliography links bleeds into the next entry, making all of them Go-To'd as a single block.

## Steps to reproduce

1. Open a markdown file like [this one](https://github.com/SublimeText/sublime-text-docset/blob/master/README.md)
2. <kbd>Cmd</kbd>+<kbd>R</kbd> and type `http` or the name of any bib link
3. The whole block is highlighted, and the Palette entry is unreadable

deathaxe:
Fixes #2360

Reference definitions need to be separated by line to correctly display them in the Goto Symbol Overlay. The newline `\n` is no longer included into the `meta.link.reference` scope in order to separate the lines from each other.

**Before**

grafik

**After**

grafik

**Notes**

1. As `\s*` includes the `\n` the `link-title` context is modified accordingly.
2. The former `after-link-title` and `link-ref-def` contexts are merged into a `link-ref-def-expect-end`

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -275,7 +275,7 @@ contexts:
         6: markup.underline.link.markdown
         7: punctuation.definition.link.end.markdown
         8: markup.underline.link.markdown
-      push: [link-ref-def, after-link-title, link-title]
+      push: [link-ref-def-expect-end, link-title]
     - match: '^(?=\S)(?![=-]{3,}\s*$)'
       branch_point: heading2-branch
       branch:
@@ -340,10 +340,6 @@ contexts:
         1: markup.heading.2.setext.markdown punctuation.definition.heading.setext.markdown
       pop: true
 
-  link-ref-def:
-    - meta_scope: meta.link.reference.def.markdown
-    - match: ''
-      pop: true
   ampersand:
     - match: (?!{{html_entity}})&
       comment: |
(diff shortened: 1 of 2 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -1426,45 +1422,40 @@ contexts:
...
-    - match: '(.*)$\n?'
-      captures:
-        1: invalid.illegal.expected-eol.markdown
+  non-terminated-link-title:
+    - match: ^\s*$\n?
+      scope: invalid.illegal.non-terminated.link-title.markdown
+      pop: true
+  link-ref-def-expect-end:
+    - meta_scope: meta.link.reference.def.markdown
+    - match: .*$
```
FichteFoll, line 1457: This will match trailing whitespace. I know it matches the previous behavior, but I don't think whitespace actually matters here. It certainly doesn't on https://spec.commonmark.org/dingus/.

Would be a good oppurtunity to change this here.

deathaxe: Sure.

wbond:
This is probably a strong argument for something like https://github.com/sublimehq/sublime_text/issues/3315

deathaxe:
Also like the reference completions discussed at https://forum.sublimetext.com/t/list-markdown-reference-links-updating-old-plugins/52761. Might be handled by a 3rd party plugin though.

PR #2366, opened 2020-05-17
[Python] Fix function return type annotation

deathaxe:
This commit fixes the return type annotation after line continuation.

It also removes the `comments` from the `function-after-parameters` as it seems not to be used at all.

The `:` is moved into a dedicated `function-terminator` context to avoid the `(?=:)` lookahead used to pop the return type annotation context.


FichteFoll:
Comments don't work in situations where line continuations must be explicit anyway.

PR #2370, opened 2020-05-24
[Hansl] Add initial language support

atecon:
Hi,

I would like to add support for the open-source statistics and econometrics software named "Gretl" (http://gretl.sourceforge.net/). Its scripting language is called "Hansl". For a primer to "Hansl" see here: https://sourceforge.net/projects/gretl/files/manual/hansl-primer-a4.pdf/download
I've added:
1) sublime-syntax file for Hansl
2) sublime-completion file for Hansl
3) a test file for the Hansl syntax
4) various snippets for Gretl
5) a build-system file for Gretl

Best,
Artur


FichteFoll:
This language doesn't seem to be popular enough to be added to the default packages and have it maintained here. See also the comment in the README about adding new syntaxes. You instead want to add this as its own package to package control.

I'm not an authority on this repository, but from my experience I'm quite confident that this won't make it in.

atecon:
> This language doesn't seem to be popular enough to be added to the default packages and have it maintained here. See also the comment in the README about adding new syntaxes. You instead want to add this as its own package to package control.

Thanks. Putting it to its own package is actually a good suggestion. Thanks.

> I'm not an authority on this repository, but from my experience I'm quite confident that this won't make it in.

I am fine with this. I just read the relevant sentence in the README that currently no further languages are being added.

PR #2374, opened 2020-06-01
[HTML] Add scopes for utf8 percent encoded characters

deathaxe:
specs: https://url.spec.whatwg.org/#userinfo-percent-encode-set

Addresses a "lively" discussion at https://forum.sublimetext.com/t/advanced-other-color-or-hide-html-20-in-st-user-interface/51447

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -227,6 +227,10 @@ contexts:
       captures:
         1: punctuation.definition.entity.html
         2: punctuation.terminator.entity.html
+    - match: (%)\h{2}
+      scope: constant.character.escape.html
+      captures:
+        1: punctuation.definition.escape.html
 
   cdata:
     - match: (<!\[)(CDATA)(\[)
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -228,7 +228,7 @@ contexts:
         1: punctuation.definition.entity.html
         2: punctuation.terminator.entity.html
     - match: (%)\h{2}
-      scope: constant.character.escape.html
+      scope: constant.character.escape.url.html
       captures:
         1: punctuation.definition.escape.html
 
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -227,10 +227,6 @@ contexts:
       captures:
         1: punctuation.definition.entity.html
         2: punctuation.terminator.entity.html
-    - match: (%)\h{2}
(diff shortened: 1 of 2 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -227,6 +227,10 @@ contexts:
       captures:
         1: punctuation.definition.entity.html
         2: punctuation.terminator.entity.html
+    - match: (%)\h{2}
+      scope: constant.character.escape.html
```
keith-hall, line 231: maybe better to be more specific here - `constant.character.escape.url.html`, in case users want to target it specially in their color schemes etc.?

FichteFoll:
Adding this match to the `entities` context means it will be matched globally, even inside normal text, right? I would expect it to only be active in attributes, ideally only href and src attributes at that, since these sequences are only relevant for URIs, not HTML in general.

keith-hall:
Ah I missed that in my review because GitHub didn't show enough context ;)
I assumed it was a copy of [the fix proposed on the forum](https://forum.sublimetext.com/t/advanced-other-color-or-hide-html-20-in-st-user-interface/51447/13?u=kingkeith).

deathaxe:
I have no strong meaning about it, but was not sure about a most common approach, which works for all kinds of templating syntaxes which use html.

wbond:
I think the issue @FichteFoll raised is worth trying to find a solution for. I would consider it a bug if percent encoded triplets were highlighted outside of something that is guaranteed to be a URL.

deathaxe:
Ok. Can we merge the interpolation branch first? The changes here would conlfict with it otherwise.

HTML/syntax_test_html.html
```diff
@@ -234,6 +234,12 @@
         ##                        ^ - meta.attribute-with-value
         ##                         ^^ punctuation.definition.tag.end.html
 
+        <img srs="https://foo%2Fbar.com" />
```
wbond, line 237: This should be `src`, right?
 
```suggestion
        <img src="https://foo%2Fbar.com" />
```

deathaxe: Nope, it should ensure not to match `%2f` in something else that `href` or `src` tag. Can choose a different one if that counts.

deathaxe: The positive test is one above.

deathaxe: I've renamed the tag and added some comments to clarify the meaning of the test case.

wbond: Ah, yes, thanks for clarifying 😄

PR #2378, opened 2020-06-06
[C#] assorted fixes

keith-hall:
Some fixes for bugs in the C# syntax definition which I noticed during day-to-day work; there are no open issues relating to these:

- fix `implicit operator` methods
- fix `for` loops without `var` etc.
- add support for `readonly get` accessors
- fix `stackalloc` and ternary operators

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -860,6 +860,10 @@ contexts:
         1: keyword.control.using.cs
         2: meta.group.cs punctuation.section.group.begin.cs
       set: [using_block, line_of_code]
+    - match: \b(using)\s*
+      captures:
+        1: keyword.control.using.cs
+      push: var_declaration
     - match: \b(fixed)\s*(\()
       captures:
         1: keyword.control.other.fixed.cs
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -768,8 +768,10 @@ contexts:
                 - include: line_of_code_in
             - match: (?=\S)
               pop: true
-        - match: \b(get)\b
-          scope: storage.type.function.accessor.get.cs
+        - match: (?:\b(readonly)\b\s+)?\b(get)\b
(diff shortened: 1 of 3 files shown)


michaelblyons:
Rebase?

C#/C#.sublime-syntax
```diff
@@ -525,6 +525,9 @@ contexts:
       scope: storage.modifier.cs
     - match: '{{visibility}}'
       scope: storage.modifier.access.cs
+    - match: '\b(operator)\b'
```
wbond, line 528: We can drop the capture, right?

C#/C#.sublime-syntax
```diff
@@ -858,6 +860,10 @@ contexts:
         1: keyword.control.using.cs
         2: meta.group.cs punctuation.section.group.begin.cs
       set: [using_block, line_of_code]
+    - match: \b(using)\s*
```
wbond, line 863: I don't know the context, but I presume it isn't possible for another identifier starting with `using` to be present here?

C#/C#.sublime-syntax
```diff
@@ -1478,10 +1490,10 @@ contexts:
       push: type_argument
     - match: '{{name}}'
       scope: support.type.cs
-    - match: '[{:]'
-      scope: invalid.illegal
-      pop: true
-    - match: (?=\}|\)|>|\]|,|;|>|=>)
+    # - match: '[{:]'
+    #   scope: invalid.illegal
+    #   pop: true
```
wbond, line 1495: We can probably just remove these commented-out lines, right?

C#/C#.sublime-syntax
keith-hall: I believe that is correct, yes. If one wants to define or reference a variable called `using`, it has to be escaped with `@` - `var @using = 123;`

wbond:
LGTM, thanks!

PR #2379, opened 2020-06-06
[Python] Use Windows launcher for build system

### LINKED ISSUE #381: [Python] Use Python Launcher on Windows for build system
Currently, the build system invokes `python` via shell_cmd. However, on Windows this should _definitely_ use the `py` launcher that any Python version >=3.3 installs globally and into the PATH by default, which is capable of a couple neat things but most importantly interprets shebang lines and is on the PATH by default, which afaik a normal installation is not.

Now, here comes a problem: Any Python version lower than that does not get the launcher installed and thus the build system will fail for those users that only have older versions installed. Since everyone still using Py2 is either living under a rock, using a (possibly old) (and LTS) Linux version that still only comes with Py2 (unimportant since this is Windows-only) or otherwise ... whatever, I suppose the number of people affected by this is rather small. Just for compatibility's sake, we could define a variant that always calls `python` even on Windows, so that users only have to modify their PATH and select the legacy build variant instead of modifying the build system directly. The disadvantage of that is that the variant would also be displayed for non-Windows platforms.

Thoughts?

FichteFoll:
Since Python 3.3, a `py` launcher is installed on Windows, which is by
far the best choice to execute python scripts on Python. It can
interpret shebang lines, read the default Python version configuration
and even handle virtualenvs (although the latter is unlikely to occur
within ST's environment). It will also be available in the PATH with the
default installer configuration, unlike the `python` executable.

Also change the main command to use `cmd` instead of `shell_cmd`.

https://docs.python.org/3/using/windows.html#launcher

Closes #381.

### DIFF
--- a/Python/Python.sublime-build
+++ b/Python/Python.sublime-build
@@ -1,9 +1,13 @@
 {
-	"shell_cmd": "python -u \"$file\"",
+	"cmd": ["python", "-u", "$file"],
 	"file_regex": "^[ ]*File \"(...*?)\", line ([0-9]*)",
 	"selector": "source.python",
 
 	"env": {"PYTHONIOENCODING": "utf-8"},
+
+	"windows": {
+		"cmd": ["py", "-u", "$file"],
+	},
 
 	"variants":
 	[


FichteFoll:
I forgot to mention that I'm opening this now that Python 2 has been EOLd and the officially supported Python versions are 3.5-3.8 now.

wbond:
I'm sure this will affect some users, but it seems that we've waited for a fairly reasonable amount of time.

PR #2382, opened 2020-06-09
[Go] Additional Snippets

TheSecEng:
Implementation of snippets for the following

- switch expression { case condition...
- select { case condition: }
- map[type]type
- make(type, 0)
- if else
- for _, var := range var
- method
- hello webapp


Go/Snippets/go-ifelse.sublime-snippet
```diff
@@ -0,0 +1,10 @@
+<snippet>
+    <content><![CDATA[if $0 {
+
```
Ultra-Instinct-05, line 3: Maybe have placeholders within `if-else` blocks so users could directly place content in that ?

Go/Snippets/go-make.sublime-snippet
```diff
@@ -0,0 +1,6 @@
+<snippet>
+    <content><![CDATA[make(${1:type}, 0)]]></content>
```
FichteFoll, line 2: This seems more like a completion than a snippet to me. Same for `map`.

TheSecEng: is it okay to include a sublime-completions file in the default packages?

Go/Go.sublime-completions
```diff
@@ -1,321 +1,328 @@
...
+            "trigger": "make",
+            "annotation": "make(type, 0)",
+            "contents": "make(${1:type}, 0)",
+            "kind": "snippet"
         },
         {
-            "trigger": "string",
-            "contents": "string",
-            "kind": "type",
+            "trigger": "make",
```
FichteFoll, line 202: If I were to use this, I think I would prefer the bare `make` to be ranked higher than the other "smarter" completion, but I'm not sure if the order in the completion files affects this. LGTM otherwise.

TheSecEng: Ah, you know I’m going to check that. I would agree with you on this.

michaelblyons: I got tired of [doubling up](https://github.com/robballou/sublimetext-sshconfig/blob/master/Support/SSH%20Config%20-%20Match%20Conditions.sublime-completions#L14-L29) completions and snippets, so I started building them both from [the same source](https://github.com/robballou/sublimetext-sshconfig/blob/master/src/options.yaml#L11-L13). (Inspired by the Bash [builtins generator](https://github.com/sublimehq/Packages/blob/master/ShellScript/commands-builtin-shell-bash.yml).)

I don't use Go, so I don't know if there are enough completions to be significant here. You _may_ want to consider adding `details` while you're in here, though.

PR #2383, opened 2020-06-09
[Markdown] don't treat lone underscores in link text as formatting

keith-hall:
this fixes a bug in the Markdown syntax definition whereby inline italic/bold formatting was taking precedence over the square bracket marking the end of the link text. The closing square bracket will now take precedence, in line with the CommonMark specification.


wbond:
Looks like we've got a conflict with some recently-merged Markdown changes.

wbond:
I ended up cherry-picking this into master via 1c5205f57aed01d48eeb52b643daea5889a16559

Issue #1968, opened 2019-05-10
[Lisp] Highlighting seems very gone


JellyWX:
The highlighting on Scheme dialect of Lisp seems very dead.

![Selection_004](https://user-images.githubusercontent.com/5578434/57516622-9b0e9b80-7304-11e9-87e9-d68cc0fbaea0.png)

1: Minus symbols as variable names are highlighted inconsistently
2: Some internal functions like map are highlighted, whereas others like pair are not (maybe this depends what Lisp dialect is in use)
3: define isn't highlighted

Another note I just found: `null?` highlights the `null` section purple but the qmark remains grey/white
+: `rest` is highlighted as a function like `map`

michaelblyons:
``` lisp
(define (square x)
    (* x x))

(define (square-tree tree)
    (map (lambda (sub-tree)
        (if (pair? sub-tree)
            (square-tree sub-tree)
            (square sub-tree)))
    tree)
```

Lisp (mostly) still has a "highlight the keywords" structure to its syntax definition file, rather than the stack of contexts that most of the recent language rewrites have used. On the plus side, Lisp's elegance _probably_ makes it relatively easy to rewrite.

rotty:
I've noticed this issue as well, and especially the dash issue is extremely annoying I'm not a Sublime user, but the Rust ecosystem syntax highlighting support is based on Sublime's syntax definitions, via the `syntect` crate, which in turn is used by the static site generator I happen to be using.

@michaelblyons  Could you give some guidance on what "modern" syntax definitions to base a Scheme/CL syntax definition on? I'd be interested at giving it a go, probably focusing on Scheme first.

michaelblyons:
I don't know what syntax definition might be closest to Lisp/Scheme, mostly because I don't use any languages that look like them. :wink: In my opinion, Javascript is the syntax definition that is most meticulously maintained. Perl had a fairly recent rewrite. There's an overhaul of Haskell in PR #2225.

The Sublime Text Discord is happy to help with questions, and the pointers in #757 may help you out. (I believe someone specifically mentions `syntect --debug` as being helpful to them.)

jrappen:
[![SublimeHQ Discord](https://img.shields.io/discord/280102180189634562?label=SublimeHQ%20Discord&logo=discord)](https://discord.gg/D43Pecu)

Thom1729:
I'd also recommend taking a look at the Lua syntax definition, which has a similar structure to the JavaScript definition but is much simpler.

[...]

PR #2386, opened 2020-06-16
[Lisp]

gighmmpiob:
Hyphen is a valid character in identifiers:

```lisp
(defun log-1(log-1) (if log-1 "1" "0"))
(format t (log-1 (< 1 2)))
```


michaelblyons:
I see you have added negative lookbehinds. Can you verify that the change is still compatible with the "fast" regex engine? <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>B</kbd> and choose the regex compatibility option.

jfcherng:
Failing test cases should be added.

gighmmpiob:
Besides hyphens, Lisp allows a lot more other characters inside identifiers. This is a valid code:

```lisp
(write (funcall (lambda (+ -) (* + -)) 3 4)) ; prints 12
```

Here `+` and `-` are bound variables in the lambda function. Only `*` is the actual operator. Any suggestion on how to resolve this? If you want to open a PR yourself let me know and I'll close this.

jfcherng:
> If you want to open a PR yourself let me know and I'll close this.

No. I won't. I don't write Lisp. I can't even know I have solved it or not.

michaelblyons:
I also have never written Lisp, but I spent a little time just now mucking about in the syntax definition cleaning up a little cruft. Some things are not ideal, and some are downright broken: `do*`? wtf?

I have a [still-very-naive modification](https://github.com/michaelblyons/SublimeSyntax-Defaults/tree/lisp/errata) of the existing syntax definition that helps... a little. There are two big problems with a full rewrite: (1) Like @jfcherng, most (all?) of the people who spend time improving the default syntax definitions do not know Lisp, and (2) Edits to the Lisp syntax affect lots of file extensions, so we don't want to accidentally clobber Scheme because we did something for CLisp.

If you're willing to work on a PR, please take what I have (or start from an empty file!). Just be aware that people will review your code. These people may or may not know anything about Lisp, but we'd love to have someone fix it up. See also: #2319 and #1968.

jfcherng:
Reminds me https://iamwilhelm.github.io/bnf-examples/lisp + https://github.com/BenjaminSchaaf/sbnf.

PR #2393, opened 2020-06-25
[JavaScript] Improve import.meta detection with branching.

Thom1729:
Also splits up the import/export contexts in the process, which is probably a good thing.


JavaScript/JavaScript.sublime-syntax
```diff
@@ -164,21 +164,27 @@ contexts:
...
+  import-statement:
+    - match: import{{identifier_break}}
       scope: keyword.control.import-export.js
       set:
         - meta_scope: meta.import.js
         - match: (?=[.(]) # Recovery for import expressions
-          set:
-            - expression-statement-end
-            - import-expression-end
+          fail: import-statement
```
deathaxe, line 180: I am surprised to see `fail` work, when being used within a `set` section. I can remember a discussion about `set` to set away from a branch. The last time I tried that in Java, `fail` was just ignored.

Thom1729: Same, honestly. I thought I'd tried it before and it didn't work. But it seems to work here. Maybe one of the fixes along the way did the trick. 🤷 Either way, I'm glad; without `set`, we'd need a lot more dummy contexts and multiple `pop`s.

deathaxe: It appears to be related with the consuming match rule which sets the underlying context. If I do the same with non-matching rules like `- match: ''` or `- match: (?=...)` it still fails by ignoring the `fail` then.

I run into the situation of needing tons of duplicated contexts just because of different pop-levels. Thats a fundamental tradeoff of the new branching mechanism, which makes syntaxes unneccessarily complicated.

I'd expect `branch` to act as a `push` with builtin reset-command per branch-tree. Thus a nested `set` should always work like in this example by just replacing the first context on stack. Thus not leaving a branch statement.

Would make things much easier.

wbond:
Thanks!

PR #2397, opened 2020-06-28
[HTML] Improve string interpolation and embedding CSS/JavaScript

deathaxe:
This PR adds string interpolation support for embedded JavaScript and CSS tag attributes according to the scope naming guidelines by adding the `meta.string` and `meta.interpolation` scopes accordingly.

It also fixes `source.js` and `source.css` main scopes to be applied by putting the embedded scopes into dedicated contexts.

Furthermore it fixes an issue with `meta.attribute-with-value.event` being applied multiple times.


deathaxe:
As of ST 4075 some of the changes made in this PR may not be needed anymore.

To make use of it the syntax needs to be converted to `version: 2`.

As it includes some changes which are not related directly with this PR, I'd like to put them into another PR.

This new PR with a HTML version 2 would then 

a) include required changes to be compatible with sublime-syntax version 2
b) sort/organize the contexts the same as XML.sublime-settings does

This change is already prepared offline with all other PRs merged in.

deathaxe:
OK, updated to sublime-syntax version 2 here as the remaining changes to fix the interpolation staff are too trivial after all.

wbond:
Could you rebase this on master to see if we can get the tests to pass?

deathaxe:
Ok, done. Merged the master into the feature branch though as it was easier.

wbond:
How was your general experience of migrating to .sublime-syntax version 2?

deathaxe:
Only few things to fix so far, most likely those related to the set/push meta boundary fixes. Fixing obvious backward compatibility breaks seems easy. I am a bit uncertain about subtile changes though as many test cases don't cover accurate meta scope boundaries or checks for duplicated or overlapping scopes. Hence, I'd prefer to have a closer look at each syntax to ensure not to break untested stuff.

wbond:
Having a rundown of the exact changes in the docs I am sure will help.

deathaxe:
With regards to the HTML syntax I am pretty sure the move to version 2 shouldn't break too much. Meta scopes seem to be ok and the clear_type statement is not used.

wbond:
Thanks for your work on this!

PR #2400, opened 2020-07-01
[Haskell] comment have false possitive closed #2399

### LINKED ISSUE #2399: [Haskell] comment have false possitive closed #2399
- Sublime Version: none
- OS Version: Gentoo Linux

```hs
myManageHook = composeAll
  [ isDialog                   --> doFullFloat
  , return True                --> doShift "1"
  ]
```

`-->` is not comment out.
However sublime def comment to start with `--`.

>  An ordinary comment begins with a sequence of two or more consecutive dashes (e.g. --) and extends to the following newline. The sequence of dashes must not form part of a legal lexeme. For example, “-->” or “|--” do not begin a comment, because both of these are legal lexemes; however “--foo” does start a comment. 
> [2 Lexical Structure](https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-170002.3)

I think solve to fix <https://github.com/sublimehq/Packages/blob/f36b8f807d5f30d2b8ef639232a9fc5960f550fa/Haskell/Haskell.sublime-syntax#L202>.

```yaml
    - match: '--'
```

to 

```yaml
    - match: '--(?![[:punct:]])'
```

However I have not test env.
So I open issue instead pull request.

ncaq:
`-->` is not comment.
However sublime def comment to start with `--`.

>  An ordinary comment begins with a sequence of two or more consecutive dashes (e.g. --) and extends to the following newline. The sequence of dashes must not form part of a legal lexeme. For example, “-->” or “|--” do not begin a comment, because both of these are legal lexemes; however “--foo” does start a comment.
> [2 Lexical Structure](https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-170002.3)


deathaxe:
You should propably add the following to the PR message to auto-close the issue once this PR is merged.

Fixes #2399

ncaq:
@deathaxe I added closed word on commit message.
However I added.

deathaxe:
AFAIK issues are closed only by tags in PR messages. The link in commit messages has no effect - at least it was the case in the past.

ncaq:
@deathaxe I must be remembering things wrong.

PR #2409, opened 2020-07-15
[Erlang] Remove obsolete query keyword

### LINKED ISSUE #2401: [Erlang] 'query' keyword is deprecated
The atom `query` is not a keyword since at least R16B (?) and should be removed.

https://github.com/sublimehq/Packages/blob/f36b8f807d5f30d2b8ef639232a9fc5960f550fa/Erlang/Erlang.sublime-syntax#L764

deathaxe:
Fixes #2401

### DIFF
--- a/Erlang/Erlang.sublime-syntax
+++ b/Erlang/Erlang.sublime-syntax
@@ -760,13 +760,6 @@ contexts:
         - meta_scope: meta.block.erlang
         - include: expr-control-flow-end
         - include: expr-control-body
-    # http://erlang.org/doc/reference_manual/...?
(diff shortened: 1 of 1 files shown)


wbond:
For a language that has been around since 1998, removing a deprecated keyword from 2014 seems pretty aggressive.

deathaxe:
I have no strong meaning about it, but you may be right. Just saw that I didn't have a link to the specs in the syntax definition.

I am fine to close it unmerged with the issue rejected as well.

wbond:
I think I'll close as a won't fix for now. We can revisit in the future.

PR #2328, opened 2020-04-09
[PHP] add indentation tests

keith-hall:

PR #2411, opened 2020-07-17
[Ruby] Add lambda operator to the list of builtin functions

joseramonc:
This is mostly to fix the ligature issue, since `->` is currently tokenizing as arithmetic and comparision

Before:
image
image

After (ligature works):
image

### DIFF
--- a/Ruby/syntax_test_ruby.rb
+++ b/Ruby/syntax_test_ruby.rb
@@ -1129,7 +1129,7 @@ exit! 2
 #^^^^ support.function.builtin
 
  ->
-#^^ support.function.builtin
+#^^ meta.function.ruby storage.type.function.ruby keyword.declaration.function.lambda.ruby
 
 
 ##################
--- a/Ruby/syntax_test_ruby.rb
+++ b/Ruby/syntax_test_ruby.rb
@@ -1128,8 +1128,8 @@ abort "Ending"
 exit! 2
 #^^^^ support.function.builtin
 
- ->
-#^^ meta.function.ruby storage.type.function.ruby keyword.declaration.function.lambda.ruby
+get :name, -> { "John" }
+#          ^^ meta.function.ruby storage.type.function.ruby keyword.declaration.function.lambda.ruby
 
 
 ##################
--- a/Ruby/syntax_test_ruby.rb
+++ b/Ruby/syntax_test_ruby.rb
@@ -1128,6 +1128,8 @@ abort "Ending"
 exit! 2
 #^^^^ support.function.builtin
 
+ ->
+#^^ support.function.builtin
 
 
 ##################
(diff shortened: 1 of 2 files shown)


jwortmann:
I assume `->` here in Ruby has the same meaning as in Java or `=>` in JavaScript and C#. I think the scope should not be `support.function` here, because that's meant for built-in function names like e.g. `println` instead. Here are some alternatives which are currently used in this repository:

* C#: [storage.type.function.lambda.cs](https://github.com/sublimehq/Packages/blob/33804834e8b926cc2a2834bb98a2dd746c481f69/C%23/C%23.sublime-syntax#L989)
* D: [storage.type.function.d keyword.declaration.function.lambda.d](https://github.com/sublimehq/Packages/blob/33804834e8b926cc2a2834bb98a2dd746c481f69/D/D.sublime-syntax#L1324)
* Java: [storage.type.function.anonymous.java](https://github.com/sublimehq/Packages/blob/33804834e8b926cc2a2834bb98a2dd746c481f69/Java/Java.sublime-syntax#L1025)
* JavaScript: [storage.type.function.arrow.js](https://github.com/sublimehq/Packages/blob/33804834e8b926cc2a2834bb98a2dd746c481f69/JavaScript/JavaScript.sublime-syntax#L1502)
* Scala: [storage.type.function.arrow.lambda.scala](https://github.com/sublimehq/Packages/blob/33804834e8b926cc2a2834bb98a2dd746c481f69/Scala/Scala.sublime-syntax#L228)

I think the scope name used in D is the most appropriate here and up to date with the scope naming guidelines.

joseramonc:
Thanks for your feedback, I just changed the scope to match D, however, I'm not sure why tests failed (seems unrelated to this) or is there anything else I need to do regarding adding this lambda operator?

michaelblyons:
I think the tests are not your fault. See https://github.com/sublimehq/Packages/pull/2328

wbond:
Yeah, I'm working on getting the CI test runner fixed

wbond:
If you rebase on master, the CI should be working now.

joseramonc:
Awesome, just rebased and it's passing now. Thanks!

Ruby/syntax_test_ruby.rb
```diff
@@ -1128,6 +1128,8 @@ class MyClass
 exit! 2
 #^^^^ support.function.builtin
 
+ ->
+#^^ meta.function.ruby storage.type.function.ruby keyword.declaration.function.lambda.ruby
```
wbond, line 1132: This is a good first step, but perhaps provide a slightly more detailed example also? I'd recommend perhaps some parameters and a simple function body.

joseramonc: Got it, I've pushed a better example for the `->`

wbond:
Thanks!

PR #2420, opened 2020-07-21
[HTML] Add sections and logically sort contexts

deathaxe:
This PR proposes to ...

1. add section headers `###[ name ]###` to group related contexts.
2. sort sections and contexts logically based on the model of the XML.sublime-syntax.

It does not provide any functional changes.

Note: I tracked all the changes but at some point it was not easy to update all the small commits, thus had to squash them.


wbond:
Works for me, thanks!

PR #2423, opened 2020-07-21
[ShellScript] A set of several patches

### LINKED ISSUE #1191: [Bash] Proposal: Give "sudo" a special parsing path
image

It would be good if a `sudo` command could be highlight as the 3rd one in the above screenshot. What do you guys think?

Or, at least, the `-rf` is not really an option of `sudo`.

### LINKED ISSUE #1248: [Shell] Clear string scope in "interpolated subcommands"

### LINKED ISSUE #1781: Bash: '#' (comment) overrides command-substitution range
I found a fancy way to add comments in a multiline command:
```bash
bc | ` # get quarks ` \
ac |  ` # remove flair` \
ab ` # important; this and that ` "${USELESS_TEXT}" | ` # match text` \
cat

```

Bash is running it _(not ... it, but the un-obsfucated one :stuck_out_tongue_closed_eyes:)_, Github is (as of commit time) highlighting them correctly, however, Sublime is not:
image

_Note that his "fancy syntax" **does not apply** to `$()`!_

Build 3176

### LINKED ISSUE #2391: [Bash] Comment after redirection make next line wrong scope
```bash
foo > ${bar} # comment
if [ $? -eq 0 ]; then
    # comment
fi
```

image
See the `if` and the condition.

## References

- https://www.reddit.com/r/SublimeText/comments/heauml/if_not_recognized_as_a_keyword_if_and_in_the/

### LINKED ISSUE #2408: [Bash] Wrong highlighting for inline comment in backticks
```bash

sed -i'' \
    `# comment 1` \
    -e 's/search 1/replace 1/g' \
    `: # comment 2` \
    -e 's/search 2/replace 2/g' \
    "input.txt"
```

Both `` `# comment 1` \ `` and `` `: # comment 2` \ `` are wrong.

**Expected** (Photoshop-ed):
image

**Actual**:
image

- Sublime Version: 4077
- OS Version: Win7 x64

deathaxe:
This PR proposes to update the Bash.sublime-syntax in order to fix most of the known issues which can be handled with reasonable effort.

Fixes #1191 (scope `sudo` as builtin)
Fixes #1248
Addresses all bash related issues of #1358
Fixes #1781 
Fixes #2327 
Fixes #2391 (the main trigger whose fixes ended up in fixing many others found on the way)
Fixes #2408

The main goals of this patch set are:

1. Fix meta.function-call boundaries
2. Fix issues with comments being matched within function call or expression scopes (related with 1.).
3. Improve complience with scope naming guidelines 
4. Improve meta scope consistency and shorten scope names where possible.
5. Remove the `-bt` contexts which add a lot of duplicated contexts
6. Fix as many as possible known issues (hopefully most of #1358, too but maybe not all of them yet)
7. Add sections and sort contexts logically by a top-down scheme with more complex statements at the top of the file and smaller more common ones below.
8. Update the syntax definition to version 2.
9. Prepare to be extended for proper embedding in Makefile

**Benchmarks**

Compared to ST 4079's revision of Bash, the new syntax is about 35-55% faster depending on the structure of source code. The mean performance improvement is about 40%.

Tested against the syntax_test_bash.sh an enlarged [bash2048.sh](https://github.com/mydzor/bash2048/blob/master/bash2048.sh) and several scripts shipped with Git for Windows (e.g. merge_tool.sh, ...).

Hope this fixes more than it breaks.

This PR is rather huge and should propably undergo some deeper tests. Maybe someone could also update #1358 to see what's still missing.


ShellScript/Bash.sublime-syntax
```diff
@@ -704,501 +926,298 @@ contexts:
...
+        - match: \s*(\))
+          captures:
+            1: punctuation.section.interpolation.end.shell
+          pop: 1
+        - include: main
+    - match: \`
+      scope:
+        meta.interpolation.command.shell
+        punctuation.section.interpolation.begin.shell
+      embed: main
```
rwols, line 1113: Nice!

deathaxe: Yes, this change made many things much easier and the code a bit cleaner.

deathaxe:
Looks like I trapped into https://github.com/sublimehq/sublime_text/issues/3495 with the recent shebang changes.

jfcherng:
Some bugs:

- The `f` flag:
  image

- The leading `7`:
  image


---

I think it would be good to give `true`/`false` a dedicated scope rather than saying they are simply strings since they can be used in the `if` test like
image

---

deathaxe:
Looks I was mislead with first chars in commands as Bash expects variables to not begin with digits.

I see `sudo` needs special handling somehow. `-rf` is interpreted as argument of `sudo` which does not support `-f`.

Is `true` and `false` a special constant in Bash? Already thought about special scoping, but was not sure.

I already thought about a separate list of `commands-builtin-linux.shell` or something like that to separate builtin bash commans from builtin/special linux commands. May be useful if we start creating dialects for ZSH etc. anyway.

jfcherng:
> Is true and false a special constant in Bash? Already thought about special scoping, but was not sure.

I just realized that my code works because in

```bash
if $var; then
    # ...
fi
```

`$var` will be executed as a command. In my case, it's the command `true` or the command `false`, so it works.

I don't find `boolean` type exist in Bash. But highlighting them could be useful though.

deathaxe:
Bash seems to provide some defaults for `true` and `false` as someighing like `if true ; then printf "foo"; fi` prints text (at least with bash for git for windows), while `if false ; then ... fi` prints nothing.

jfcherng:
I think that's because there is a command called `true` and it returns `0` (success).
image

Similarly, there is a `false` command.

jfcherng:
Yet another flag issue:

image

`-9r` means max compression and recursive include dirs.

jfcherng:
Just want to say you are so awesome :+1:

ShellScript/Bash.sublime-syntax
```diff
@@ -704,714 +1159,586 @@ contexts:
...
+    - match: \!
+      scope: keyword.operator.logical.shell
+      pop: 1
+    - match: \#
+      scope: keyword.operator.arithmetic.shell
+      pop: 1
+    - include: immediately-pop
+
+  expansions-parameter-chars:
+    - include: expressions-parameter-variable
```
jfcherng, line 1430: There are trailing spaces in this line.

deathaxe: Thanks

michaelblyons:
Minor regression: Both these expressions used to have no `constant.numeric`s. Now the top one does.

``` sh
cd foo/bar-2345
cd foo/bar2345
````

deathaxe:
Numbers have not been matched in general arguments before. The only way is to disable that again.

wbond:
In terms of the larger testing, I think the path forward with significant changes like these is to have people use it and report issues.

wbond:
This seems like a significant chunk of work. Thanks for all of the effort you put in!

FichteFoll:
I find it quite a bit sad to have all the organizing of splitting this changeset into 240 commits with reasonable commit messages to be lost by squashing everything into https://github.com/sublimehq/Packages/commit/880696d8284000cd35118098d02b9c0334047e73.

wbond:
I preserve all of the messages when I squash, but unfortunately the packages repo ends up littering the ST git repo with an immense number of useless commits if I don't squash. Some of the contributors do a great job in organizing and making logical commits, but that tends to be the exception. We don't need 50 commits in the history of a syntax with all of the mistakes and attempts at fixing it that sometimes happens.

deathaxe:
I am with Will here. Normally we should have small PRs fixing certain kinds of issues. In case of such huge changes as within this PR, commits may help understand and following certain changes during review. In the end this PR could also have been titled "Rewrite Syntax". Many older commits are superseeded by more recent changes with no certain issue being assigned. Having all of those in the main history would just create too much noise.

I just kept all the commits for reviewers to have a chance to follow. I am very fine with the current way of squashing and merging PRs.

PR #2424, opened 2020-07-21
[Markdown] Make code spans end at new list item

michaelblyons:
Thoughts? GitHub's syntax highlighter clearly disagrees with this change.

I wrote it while I was editing a big Markdown file and adding code spans in list items that switched the highlighting for large swathes of the rest of the file. I found that to be disconcerting... but if someone forgot to close a code span in a README and pushed it, that could be worse, depending on their MD implementation. 🤷‍♂️

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -1399,6 +1399,8 @@ contexts:
           scope: punctuation.definition.raw.end.markdown
           pop: true
         - match: '`+'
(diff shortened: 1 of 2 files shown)


keith-hall:
Technically the backtick shouldn't mark the start of a code span in this case (I think only Common Mark is really relevant) - something we can fix with non deterministic parsing later. But for now, it seems better / more correct to not erroneously continue that inline code span highlighting past the inline list item.

PR #2433, opened 2020-07-24
[HTML] Improvements to attribute completions

wbond:
This include two changes that make typing HTML a slightly less frustrating experience:

 - Allow over-typing the closing > of an HTML tag. This seems fairly straightforward, and shouldn't cause any trouble, at least from situations I could think of. If a user does with to have multiple >>, then can press the > key twice.
 - Augment attribute completions with extra fields. Now when a user completes an attribute, they will always have a field they can tab to "finish" the completion. If the completion closes the tag, they will have two fields, one after the attribute to allow for more attributes to be added, and a second one after the closing > of the tag.

### DIFF
--- a/HTML/Default.sublime-keymap
+++ b/HTML/Default.sublime-keymap
@@ -1,0 +1,9 @@
+[
+	{ "keys": [">"], "command": "move", "args": {"by": "characters", "forward": true}, "context":
+		[
+			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+			{ "key": "following_text", "operator": "regex_contains", "operand": "^>", "match_all": true },
+			{ "key": "selector", "operator": "equal", "operand": "punctuation.definition.tag.end.html", "match_all": true },
+		]
+	},
+]
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -582,6 +582,9 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
             # add a space if not there
             suffix = ' '
 
+        # ensure the user can always tab to the end of the completion
+        suffix += '$0'
+
         # got the tag, now find all attributes that match
         return sublime.CompletionList(
             [


HTML/Default.sublime-keymap
```diff
@@ -0,0 +1,10 @@
+[
+	{ "keys": [">"], "command": "move", "args": {"by": "characters", "forward": true}, "context":
+		[
+			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+			{ "key": "following_text", "operator": "regex_contains", "operand": "^>", "match_all": true },
+			{ "key": "selector", "operator": "equal", "operand": "punctuation.definition.tag.end.html", "match_all": true },
+			{ "key": "selector", "operator": "equal", "operand": "text.html.basic" }
```
deathaxe, line 7: I'd propably not limit that to `.basic` to also enable that in extended syntaxes such as `text.html.jinja` or `text.html.embedded.whatever`? 

IMHO the previous selector `punctuation.definition.tag.end.html` is accurate enough to make it match in appropriate situations only.

deathaxe:
```yaml
%YAML 1.2
---
# https://www.sublimetext.com/docs/syntax.html
name: HTML (jinja2)
scope: text.html.jinja2

file_extensions:
  - html

contexts:
  main:
    - match: ''
      set:
        - include: scope:text.html.basic
      with_prototype:
        - match: '{%'
          scope: punctuation.section.jinja2.begin.html
          embed: scope:source.python
          embed_scope: source.jinja2.embedded
          escape: '%}'
          escape_captures:
            0: punctuation.section.jinja2.end.html
        - match: ({{)(.+?)(}})
          scope: source.jinja2.embedded
          captures:
            1: punctuation.definition.placeholder.begin.html
            2: variable.parameter.placeholder.html
            3: punctuation.definition.placeholder.end.html
```

HTML/Default.sublime-keymap
```diff
@@ -0,0 +1,10 @@
+[
+	{ "keys": [">"], "command": "move", "args": {"by": "characters", "forward": true}, "context":
```
FichteFoll, line 2: Should this binding respect `auto_match_enabled` like the bindings for `)'"]}`?

wbond: Well, what prompted this was the attribute completions that add `”>`. Those aren’t triggered by auto_match_enabled, although I do realize the overwriting of “ is controlled by that setting.

wbond:
I was using this for a while yesterday and I found the two fields with completions ending in `”>` to require too many presses to [tab]. It ended up requiring three - one to complete a class name, one to exist the class attribute and a third to go past the end of the tag.

HTML/Default.sublime-keymap
deathaxe: I wonder if we'd need a separate setting (e.g.: `smart_type_enabled`) to enable such kind of features, which are related with `auto_match_enabled` only. Some of the new bindings in CSS which auto-add spaces to maintain code style would be a use case, too.

HTML/Default.sublime-keymap
wbond: I've incorporated this improvement

HTML/Default.sublime-keymap
wbond: I'm going to kick the can down the road for now. We can spend some time later thinking about potential settings that affect overwriting, but aren't strictly related to auto-matching.

PR #2435, opened 2020-07-25
[HTML] Prepare for inheritance

deathaxe:
This PR...

1. proposes to use `pop: 1` instead of `pop: true` in sublime-syntax version 2 only.
2. renames the `attribute-...-equals` context to `attribute-...-assignment` which looks more appropriate.
3. adds a lot of named contexts in order to introduce entry points for inherited syntax definitions to add custom highlighting.
4. tweaks the recently added `src/href` attribute contexts 
   a) to ensure maximum re-usablility in inerhited syntax definitions by avoiding  `pop: 2`
   b) by adding `meta.path.url` to the url values.
5. tweaks the usage of `meta_scope` and `meta_content_scope` in the script/style attriubtes to avoid assigning meta scopes via normal `scope` directives.
6. moves `meta.class-name`, `meta.toc-list.id` and `meta.path` so that an inherited syntax can easily clear string scopes while maintaining symbol list related meta scopes within interpolations.
7. adapts ASP test cases to the new meta scope order
8. improves ASP<->HTML interaction by using `embed` and `extends`, which results in much smaller syntax cache files.

### DIFF
--- a/ASP/ASP.sublime-syntax
+++ b/ASP/ASP.sublime-syntax
@@ -672,23 +672,28 @@ contexts:
         - match: '<%=?'
           scope: punctuation.section.embedded.begin.inside-block.asp
           pop: true
-        - match: '\s+$' # eat whitespace so that the lookahead on the next match pattern can match the next line if appropriate
+        # Eat whitespace so that the lookahead on the next match pattern
+        # can match the next line if appropriate.
+        - match: '\s*\n'
           push:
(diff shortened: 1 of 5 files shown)


wbond:
These looks like some excellent forward-looking changes, and should hopefully set the stage for higher quality templating syntaxes.

Thank you for all of your effort on this PR, and all of your other ones too!

Issue #2430, opened 2020-07-22
[RFC] Sublime Syntax Main Key Sort Order


deathaxe:
The number of main keys in the sublime-syntax file format increases. While it seems common practise to have `variables` context before `contexts` the header keys are sorted alphabetically only.

I'd propose to group the name, scope and version keys as they describe the the syntax itself.

The file_extensions  and first_line_match should follow and be separated by an empty line,

Here's an example of how it could look like:

```yaml
%YAML 1.2
---
# https://www.sublimetext.com/docs/syntax.html
# https://www.mysyntax.com/specs/index.html
name: MySyntax
scope: source.mysyntax
version: 2

# with a look forward towards diamont extends ;)
extends:
  - souse.basesyntax
  - souse.basesyntax2

file_extensions:
  - main_ext
  - ext2

hidden_file_extensions:
  - any
  - thing

first_line_match: |-
  (?x:
    ^\#! .* \b(mysyntax)\b |                                   # shebang
    ^/[/*] \s* -\*- .*? \bmysyntax\b \s* -\*-(?: \s* [/*]/)?  # editorconfig
  )

variables:
  var1: ...

contexts:
  main:
    ...
```

keith-hall:
As there has been no oppositions to this, I'd personally say that this is now the standard and perhaps should be written in our RFC repository https://github.com/SublimeText/ScopeNamingGuidelines :)

ngc92:
> As there has been no oppositions to this, I'd personally say that this is now the standard and perhaps should be written in our RFC repository https://github.com/SublimeText/ScopeNamingGuidelines :)

In addition, it would be good to update the examples given in https://www.sublimetext.com/docs/syntax.html

PR #2436, opened 2020-07-26
[XML] Update to sublime-syntax version 2

deathaxe:
This commit...

1. updates XML/DTD syntax to sublime-syntax version 2.
2. sorts the main keys according to proposal #2430
3. sorts hidden file extensions
4. removes some quotation marks from values.

### DIFF
--- a/XML/DTD.sublime-syntax
+++ b/XML/DTD.sublime-syntax
@@ -1,11 +1,13 @@
 %YAML 1.2
 ---
 name: DTD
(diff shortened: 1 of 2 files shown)


wbond:
It doesn't appear the hidden file extensions were actually sorted, but that isn't really important to me

deathaxe:
It's actuatlly a mix of sorting and grouping. ST / Visual Studio Project files / others.

PR #2441, opened 2020-08-02
[Markdown] Add Erlang support to fenced code blocks

deathaxe:


TheSecEng:
Deathaxe at it again. Improving all of the syntax files!

PR #2446, opened 2020-08-11
[Markdown] Assign specific scopes to level 3~5 headings

jfcherng:
This PR proposes to use

- `markup.heading.3.markdown` for level 3 heading (changed from `markup.heading.markdown`)
- `markup.heading.4.markdown` for level 4 heading (changed from `markup.heading.markdown`)
- `markup.heading.5.markdown` for level 5 heading (changed from `markup.heading.markdown`)
- `markup.heading.6.markdown` for level 6 heading (changed from `markup.heading.markdown`)

---

My original idea is to give headings different background colors via color scheme rules according to their levels.

Snipaste_2020-08-11_19-26-09

But soon I found that there are only `markup.heading.1.markdown`, `markup.heading.2.markdown` and other levels are all `markup.heading.markdown`. It's weird that we highlight `punctuation.definition.heading.begin.markdown` up to 6 levels but only gives level 1 and level 2 specific `meta_scope`s.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -585,7 +585,7 @@ contexts:
         - meta_content_scope: entity.name.section.markdown
         - include: atx-heading-terminator
         - include: inline-bold-italic
-    - match: '(##)(?!#)\s*(?=\S)'
+    - match: '(#{2})(?!#)\s*(?=\S)'
       captures:
         1: punctuation.definition.heading.begin.markdown
       push:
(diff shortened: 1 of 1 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -585,15 +585,39 @@ contexts:
...
+        - include: inline-bold-italic
+    - match: '(#{5})(?!#)\s*(?=\S)'
+      captures:
+        1: punctuation.definition.heading.begin.markdown
+      push:
+        - meta_scope: markup.heading.5.markdown
+        - meta_content_scope: entity.name.section.markdown
+        - include: atx-heading-terminator
+        - include: inline-bold-italic
+    - match: '(#{6})(?!#)\s*(?=\S)'
```
deathaxe, line 620: Would we also want level 6 to have a specific scope?

```
- meta_scope: markup.heading.6.markdown
```

jfcherng: I am happy to use it for sure. :+1:

wbond:
Nice enhancement, thanks!

PR #2448, opened 2020-08-12
[PHP] Support PHP 8 attributes syntax change

jfcherng:
Although PHP 8 has been feature-frozen, devs decide to hold a poll for changing the syntax `@@` for `attributes`.

https://wiki.php.net/rfc/shorter_attribute_syntax_change

It looks like to be either (more likely) `#[]` or `@[]`, but it's easy to switch between them in this PR anyway.

---

**UPDATE**: The voting result is `#[]`.


wbond:
Looks like the voting goes until Sept 2. Currently `#[]` is in first place.

jfcherng:
Yeah, `#[]` is kind of dominating. I don't think the result will change given that there are ~60 PHP core devs will vote and now 56 people has voted.

wbond:
Is this good to merge now?

jfcherng:
> Is this good to merge now?

I think yes since the voting has been ended.

wbond:
Thanks!

PR #2450, opened 2020-08-18
[Makefile] Embed extended ShellScript

### LINKED ISSUE #2388: [Makefile] Shell flag highlighting
- Sublime Version: 3.2.2 b3211
- OS Version: Windows 10

There's a bug in the built-in makefile syntax. Specifically, color highlighting. 
The first flag in a command is properly colored blue, but any following flags aren't. 

See the test string:
    
``` Makefile
html:
	$(PELICAN) $(INPUTDIR) -o $(OUTPUTDIR) -s $(CONFFILE) $(PELICANOPTS)
```
In Sublime Text, -o is correctly highlighted, but -s is not.

This may be related to #1699

deathaxe:
Fixes #2388

This commit creates an intermediate `Makefile Shell.sublime-syntax` as extended `Bash.sublime-syntax` with `variable-substitutions` from `Makefile.sublime-syntax` being added to proper contexts so that they are applied in a sane way, which respects the syntax structure of Bash.

The Makefile Shell.sublime-syntax is included into Makefile.sublime-syntax normally without `with_prototype`.

### DIFF
--- a/Makefile/Makefile Shell.sublime-syntax
+++ b/Makefile/Makefile Shell.sublime-syntax
@@ -1,0 +1,24 @@
+%YAML 1.2
+---
+scope: source.shell.embedded.makefile
+version: 2
+hidden: true
+
+extends: Packages/ShellScript/Bash.sublime-syntax
+
+contexts:
+  main:
+    # skip shebang
+    - include: statements
+
+  expansions-parameter:
+    - meta_prepend: true
+    - include: Packages/Makefile/Makefile.sublime-syntax#variable-substitutions
+
+  string-ansi-c-body:
+    - meta_append: true
+    - include: Packages/Makefile/Makefile.sublime-syntax#variable-substitutions
+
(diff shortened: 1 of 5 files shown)


michaelblyons:
Don't you want to keep the "skip shebang" override in the Makefile-specific one, and let Shell-Unix-Generic have shebangs?

deathaxe:
No, because `scope:source.shell` is included in other syntaxes such as Ruby, which then failed due to pushing into contexts without returning to main. That's the reason for the second commit.

deathaxe:
Not quite sure how to solve that issue best atm. Wasn't too aware of the edge cases when introducing shebang scope.

It was easy to fix things for builtin packages, but it might break things for 3rd-party ones. That's why `source.shell` behavior shouldn't change.

michaelblyons:
No problem. You've convinced me. 😃

PR #2452, opened 2020-08-18
[PHP] Add scope for ternary operator

jfcherng:
Yet another one why we didn't notice this... so weird

### DIFF
--- a/PHP/tests/syntax_test_php.php
+++ b/PHP/tests/syntax_test_php.php
@@ -237,6 +237,21 @@ function foo(?stinrg ...$args) {}
 //                   ^^^ keyword.operator.spread
 //                      ^^^^^ variable.parameter
 
+$a = $b ? $c::MY_CONST : $d * 5;
+//      ^ keyword.operator.ternary
+//        ^^ variable.other
+//          ^^ punctuation.accessor.double-colon
+//            ^^^^^^^^ constant.other.class
+//                     ^ keyword.operator.ternary
+//                          ^ keyword.operator.arithmetic
+
+$a = $b ? : $c::MY_CONST;
+//      ^ keyword.operator.ternary
+//        ^ keyword.operator.ternary
+//          ^^ variable.other
+//            ^^ punctuation.accessor.double-colon
+//              ^^^^^^^^ constant.other.class
+
 $arr4 = ['a', ...$arr1, 'b', ...$arr2, 'c'];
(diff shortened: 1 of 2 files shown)


PHP/tests/syntax_test_php.php
```diff
@@ -237,6 +237,21 @@ function foo(?stinrg ...$args) {}
...
+//                     ^ keyword.operator.ternary
+//                          ^ keyword.operator.arithmetic
+
+$a = $b ? : $c::MY_CONST;
+//      ^ keyword.operator.ternary
+//        ^ keyword.operator.ternary
+//          ^^ variable.other
+//            ^^ punctuation.accessor.double-colon
+//              ^^^^^^^^ constant.other.class
+
```
FichteFoll, line 254: I'm missing a test case for `:` without a `?` here. What should that be scoped as? I'd think it be a keyword still, but since the pattern for it is only pushed after a `?`, I'm not sure about it.

jfcherng: I think that would be either a [label](https://github.com/sublimehq/Packages/blob/7b279e87cf54ad0ad8c0565e3dfdff5408b69f06/PHP/tests/syntax_test_php.php#L1429-L1430) or a syntax parsing error.

jrappen:
@jfcherng I know this is already merged into `master`, but since `PHP Source` uses sublime syntax v2 maybe it should've been `pop: 1` instead of `pop: true`?

jfcherng:
> @jfcherng I know this is already merged into `master`, but since `PHP Source` uses sublime syntax v2 maybe it should've been `pop: 1` instead of `pop: true`?

Yes. I had added it in another PR. https://github.com/sublimehq/Packages/pull/2448/files#diff-1e446a02188b28f39505f6fd68756a27R283

jrappen:
@jfcherng current `master`, though:

https://github.com/sublimehq/Packages/blob/d086cf77a9a8af06fb2d96c218922ab2883a02ce/PHP/PHP%20Source.sublime-syntax#L283

jfcherng:
> @jfcherng current `master`, though:
> 
> https://github.com/sublimehq/Packages/blob/d086cf77a9a8af06fb2d96c218922ab2883a02ce/PHP/PHP%20Source.sublime-syntax#L283

Not sure what do you want me to do.

jrappen:
Yes, disregard then.

PR #2453, opened 2020-08-18
[Ruby] Improve heredoc tag scoping

deathaxe:
This PR proposes to adapt HEREDOC scopes to those of _Perl_ and _Bash_.

It uses `meta.tag.heredoc entity.name.tag` to scope opening and closing HEREDOC tags.


wbond:
Looks logical to me, thanks!

PR #2454, opened 2020-08-18
[CSS] Fix <0x01> in completion labels

deathaxe:
Looks like python 3.8? requires a raw string to perform proper regexp replacement. The intention here is to replace `${1:param}` by `param` so the completion list displays named function arguments.

grafik

### DIFF
--- a/CSS/css_completions.py
+++ b/CSS/css_completions.py
@@ -639,7 +639,7 @@ class CSSCompletions(sublime_plugin.EventListener):
 
                 for value in values:
                     if isinstance(value, str):
-                        desc = self.re_trigger.sub("\1", value)
+                        desc = self.re_trigger.sub(r"\1", value)
                         snippet = value
                     else:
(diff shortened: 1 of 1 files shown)


FichteFoll (approved):
Not sure why this behavior should have changed in 3.8. It does't seem to from my tests, but this is the correct solution regardless.

```

~ $ python3.8
Python 3.8.5 (default, Jul 27 2020, 08:42:51)
[GCC 10.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "\1"
'\x01'
>>>

~ Ξ python3.3
Python 3.3.7 (default, Oct  5 2017, 02:25:52)
[GCC 7.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "\1"
'\x01'
>>>
```

wbond:
Most likely it is just that we draw these now?

PR #2456, opened 2020-08-19
[HTML] Default Completions Selector

deathaxe:
This commit suggests to...

1. move main completion selector to a syntax specific `default_completions_selector` setting to make it easier for end users to customize it.

2. fix HTML completions not being provided within string interpolation by replacing `- meta.string` by `- string`. This issue was revealed when trying to auto complete html tags within Perl heredocs.

   ```Perl
   $var <<HTML

   <

   HTML
   ```

3. simplify and increase readability of selectors by avoiding nested subtraction.

_Note: The same selector might be used within `auto_complete_triggers`_

```json
"auto_complete_triggers":
[
	{
		"characters": "<&",
		"selector": "text.html - source - string - text.html.markdown, text.html.embedded - text.html.embedded source, markup.raw.code-fence text.html - markup.raw.code-fence text.html source, text.xml - source - string, source text.xml"
	},
	{
		"rhs_empty": true,
		"selector": "punctuation.accessor",
	}
],
```

### DIFF
--- a/HTML/HTML.sublime-settings
+++ b/HTML/HTML.sublime-settings
@@ -1,0 +1,7 @@
+{
+    // Set to true to disable default completions.
+    "disable_default_completions": false,
+
+    // Controls what scopes default completions will be provided in.
+    "default_completions_selector": "text.html - string - source - text.html.markdown, text.html.markdown text.html, source text.html"
+}
--- a/HTML/HTML.sublime-settings
+++ b/HTML/HTML.sublime-settings
@@ -3,5 +3,13 @@
     "disable_default_completions": false,
 
     // Controls what scopes default completions will be provided in.
-    "default_completions_selector": "text.html - string - source - text.html.markdown, text.html.markdown text.html, source text.html"
+    // Can be a list of strings which are joined before matching.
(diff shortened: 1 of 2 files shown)


michaelblyons:
Does this fix #438, or does that need changes in core like your suggestion above?

deathaxe:
No, it doesn't fix it. The selectors of `auto_complete_triggers` and the one used by the plugin might need to be equal for consistent behavior though.

I realized HTML completions to not trigger within HEREDOCs in Perl due to `- meta.string` and have a feeling about the possible need to be able to adjust the selector. Hence suggesting a setting which may be tweaked quickly without knowledge about python coding.

That's just an idea. Maybe we can find better/common solution.

deathaxe:
Things become even more interesting with embedded CSS/JS in embedded HTML :D

Looks like selectors need some further work to fix some edge cases.

```perl
$var <<HTML

<abbr><abbr><bdo class="test" id=""></bdo></abbr></abbr>
<abbr></abbr>
<style type="text/css">
    tr {
        color: |;
    }
</style>
HTML
```

PR #2457, opened 2020-08-20
[CSS] Default Completions Selector

deathaxe:
This commit suggests to...

1. move main completion selector to a syntax specific `default_completions_selector` setting to make it easier for end users to customize.

2. simplifies default completions selector.

   The `meta.attribute-with-value.style` part is no longer needed as those are correctly scoped `source.css.embedded`.


wbond:
Thanks!

PR #2458, opened 2020-08-20
[CSS] Omit semicolon in completions before closing quotes

deathaxe:
With this PR the last CSS attribute completion before a closing brace or quotation mark will no longer be terminated with semi-colon.

It's ment to improve workflow in inline `style` attributes.

Example:

```
   <p style="color: |">

   <p style='color: |'>

   p { color: |}
```


rchl:
It's probably fine when used within quotation marks to do it this way but inside selectors I wouldn't want to skip semicolons and it seems quite opinionated to do it.

I think that style is less common and also has same downside as skipping trailing commas on objects - adding something after requires to also change previous line.

keith-hall:
What's the motivation for this change? Just personal preference? I'm with @rchl on this one - if you really want it, maybe gate it behind (yet another) preference? But if you're trying to save bytes sent out over the wire, isn't that what minification and gzip compression are for?

PR #2463, opened 2020-08-25
[JavaScript] Meta Numbers

deathaxe:
This PR is a proposal created from the latest discussion at #2460

It is to illustrate what it means to add meta.number scopes. It means to add one or up to 1 (2) additional capture groups per match pattern to assign meta.numbers.[base|value|suffix]. The complexity is the same for applying meta.numbers on top of below constant.numeric.

Note: It also contains the older more complex solutions for comparison.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1835,44 +1835,49 @@ contexts:
           # .1, .1e1, .1e-1
           | (\.) {{dec_digit}}+ (?:{{dec_exponent}})?
         ){{identifier_break}}
+      scope: meta.number.float.decimal.js constant.numeric.value.js
       captures:
-        0: meta.number.float.decimal.js constant.numeric.value.js
         1: punctuation.separator.decimal.js
         2: punctuation.separator.decimal.js
       pop: true
 
     # integers
     - match: (0)({{dec_digit}}+){{identifier_break}}
+      scope: meta.number.integer.octal.js
       captures:
-        1: meta.number.integer.octal.js constant.numeric.base.js invalid.deprecated.numeric.octal.js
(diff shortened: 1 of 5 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -1836,43 +1836,43 @@ contexts:
...
+        1: meta.number.integer.octal.js constant.numeric.base.js invalid.deprecated.numeric.octal.js
+        2: meta.number.integer.octal.js constant.numeric.value.js invalid.deprecated.numeric.octal.js
       pop: true
 
     - match: (0[Xx])({{hex_digit}}*)(n)?{{identifier_break}}
       captures:
-        1: meta.number.base.js constant.numeric.integer.hexadecimal.js
-        2: meta.number.value.js constant.numeric.integer.hexadecimal.js
-        3: meta.number.suffix.js constant.numeric.integer.hexadecimal.js
+        1: meta.number.integer.hexadecimal.js constant.numeric.base.js
```
michaelblyons, line 1853: Can you `scope: meta.number.integer.hexadecimal.js` and leave just the `constant`s broken out in the capture groups?

wbond:
Looks like there are some test failures also

deathaxe:
I somehow need a way to run all test cases locally on my windows machine. It's not the first time I miss to update some test files.

JavaScript/JavaScript.sublime-syntax
deathaxe: Done.

wbond:
You don't use the build variant to run all tests?

deathaxe:
It does just freeze ST but nothing else. I therefore even removed it from command palette, because I got sick of killing ST when hitting it by accident.

It is meant to run all syntax test files it finds? My impression was it just tries to run all syntaxes against the single file or something like that.

PR #2465, opened 2020-08-26
[D] Fix mixin not being considered a type

### LINKED ISSUE #2375: [D] string mixin in "is" expression
- OS Version: Fedora 32 x86_64

There is a problem with highlighting string mixin in `is` expression:

```d
is(mixin("a == struct"))
```
image

```d
is(mixin("identifier." ~ symbol) == struct)
```
image

BenjaminSchaaf:
Fixes #2375

### DIFF
--- a/D/tests/syntax_test_d.d
+++ b/D/tests/syntax_test_d.d
@@ -1603,6 +1603,31 @@ extern(1)
 //          ^ punctuation.definition.string.end.d
 //           ^ punctuation.section.parens.end.d
 //            ^ punctuation.terminator.d
+  is(mixin("int") == int);
+//^^ keyword.other.d
+//  ^ punctuation.section.parens.begin.d
+//   ^^^^^ keyword.control.d
+//        ^ punctuation.section.parens.begin.d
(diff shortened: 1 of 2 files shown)


D/tests/syntax_test_d.d
```diff
@@ -1603,6 +1603,31 @@ extern(1)
 //          ^ punctuation.definition.string.end.d
 //           ^ punctuation.section.parens.end.d
 //            ^ punctuation.terminator.d
+  is(mixin("int") == int);
+//^^ keyword.other.d
+//  ^ punctuation.section.parens.begin.d
+//   ^^^^^ keyword.control.d
+//        ^ punctuation.section.parens.begin.d
+//         ^ meta.string.d string.quoted.double.d punctuation.definition.string.begin.d
+//          ^^^^ meta.string.d string.quoted.double.d
```
wbond, line 1612: Mostly a style thing, but over time I've found that using assertions such as the following tend to be easier to modify and understand:

```suggestion
//         ^^^^^ meta.string.d string.quoted.double.d
//         ^ punctuation.definition.string.begin.d
```

PR #2466, opened 2020-08-26
[ShellScript] Remove interpolation symbol list

deathaxe:
Interpolations/Expansions don't declare any symbols, but denote variable references or embedded scripts which output values are used to form the value.

The symbol selector is also too generous.


michaelblyons (approved):

rwols (approved):

FichteFoll (approved):

wbond:
Thanks!

PR #2472, opened 2020-09-01
[Markdown] Add scope for EOL in setext headings

jfcherng:
Continued from https://github.com/sublimehq/Packages/pull/2446 but for setext-style headings. This PR makes the user be able to style the EOL (\n) char (such as adding a background color) just like it can already be done in atx-style headings.

image


keith-hall (approved):

FichteFoll (approved):

wbond:
Looks like a nice change, thanks!

PR #2476, opened 2020-09-10
[JavaScript] Implement private methods.

Thom1729:
Fixes the bug @rwols reported in #1950.


JackGoff98:
The patch does not work for me. Can the method name be green please?
And also the private method call should be blue.

wbond:
I tested this and it appears to be working. Thanks!

PR #2486, opened 2020-09-17
[Regular Expressions] Match literal characters greedily

### LINKED ISSUE #2483: [Regular Expressions] Tokenization
The regex syntax seems to split ever char as a token, which results in funky highlighting with Celeste (or any other syntax using hashed highlighting).

- Sublime Version: 4086
- OS Version: macOS 10.15.6

```regex
storage
```

Example in the Find input:

Screen Shot 2020-09-16 at 8 36 58 AM

jwortmann:
Fix #2483.

### DIFF
--- a/Regular Expressions/RegExp.sublime-syntax
+++ b/Regular Expressions/RegExp.sublime-syntax
@@ -345,5 +345,5 @@ contexts:
       scope: keyword.other.any.regexp # https://github.com/sublimehq/Packages/issues/314
     - match: \)
       scope: invalid.illegal.unmatched-brace.regexp
-    - match: .
+    - match: \w+|.
       scope: meta.literal.regexp


michaelblyons:
I don't think you need to escape most things in a `[ ]` expression, but this is also what i had been thinking for #2483.

jwortmann:
Yes. The syntax tests in Clojure fail, because it uses `"` as character to escape from the embedded syntax. However, according to https://www.regular-expressions.info/characters.html a `"` is no metacharacter in Regex. It could easily be fixed by excluding `"` from the `meta.literal.regexp` match as well, but I wonder whether that implementation in Clojure is correct then.

I will look into that and the failed tests in Perl tomorrow.

michaelblyons:
I was just talking about this on Discord. Try `[^[:punct:]]+|.`?

jwortmann:
Thanks for the suggestion. However, `[^\p{P}\r\n]+|.` would not work, because `\p{P}` apparently doesn't match backtick quotes: `` ` ``. But it could be adjusted to ``[^\p{P}\r\n`]+|.`` then.

> You can't necessarily predict the characters of every single language that might embed RegExp.

This is the important part, and we therefore should be careful with this change, because we don't know whether there is some syntax package out there which embeds the Regex syntax and uses another character which is also not matched by `\p{P}` as delimiter.

Another drawback of `\p{P}` is that it *does* match underscores, but I guess we can assume that no language uses underscores as a delimiter for Regex strings, so we probably don't want to split on underscores.
Hence, maybe just using `\w+|.` instead would be better?

wbond:
Hashed syntax highlighting only deals with word chars, so we only need to match runs of letters, numbers and `_`. I believe the current state of the PR should be sufficient.

wbond:
Thanks for the fix and review!

jwortmann:
> Hashed syntax highlighting only deals with word chars, so we only need to match runs of letters, numbers and _.

I see a different behavior:

regex

Nevertheless, I agree that word characters should be sufficient here.

wbond:
I suppose I should clarify. Technically hashed syntax highlighting hashes based on whatever tokens you throw at it. I think most users will expect "identifiers" to be unique tokens. Based on the language, you can write a syntax where various punctuation are part of a token. Since regex isn't language-specific, I believe that treating a run of "word chars" as a token/identifier is the logical thing.

Issue #2356, opened 2020-05-06
[CSS] mark invalid calc args as invalid


keith-hall:
- Sublime Version: 4073
- OS Version: Xubuntu 19.10

Take the following CSS example:

```css
.example {
    height: calc(100vw-5rem);
}
```
Here, there is no whitespace between the `vw` and the `-` or the `-` and the `5`, so browsers treat it as invalid. It would be great if ST would scope the `-` as `invalid.illegal.missing-whitespace.css` to help spot the problem - currently the highlighting implies that it should work, which is misleading.

![image](https://user-images.githubusercontent.com/11882719/81147315-b5646f00-8f82-11ea-8529-29e005275284.png)

Related: https://stackoverflow.com/a/49239963/4473405

PR #2488, opened 2020-09-17
[YAML] Update numeric scopes to meta.number constant.numeric

deathaxe:
This PR applies `meta.number.<type> constant.numeric.[base|value|suffix]` to all numeric literals.

Each number type is matched by a dedicated pattern to improve readability. It is even slightly (but negligible) faster than a monolitic pattern with dozens of capture groups.

Note: We may want to discuss how to scope special number tokens like `NaN` and `inf`.

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -110,29 +110,33 @@ variables:
 
   # patterns for plain scalars of implicit different types
   # (for the Core Schema: http://www.yaml.org/spec/1.2/spec.html#schema/core/)
-  _type_null: (?:null|Null|NULL|~) # http://yaml.org/type/null.html
-  _type_bool: |- # http://yaml.org/type/bool.html
+  
+  # http://yaml.org/type/null.html
+  _type_null: (?:null|Null|NULL|~)
+  
+  # http://yaml.org/type/bool.html
+  _type_bool: |-
     (?x:
        y|Y|yes|Yes|YES|n|N|no|No|NO
       |true|True|TRUE|false|False|FALSE
       |on|On|ON|off|Off|OFF
     )
-  _type_int: |- # http://yaml.org/type/int.html
-    (?x:
-        ([-+]? (0b) [0-1_]+) # (base 2)
-      | ([-+]? (0)  [0-7_]+) # (base 8)
-      | ([-+]? (?: 0|[1-9][0-9_]*)) # (base 10)
-      | ([-+]? (0x) [0-9a-fA-F_]+) # (base 16)
-      | ([-+]? [1-9] [0-9_]* (?: :[0-5]?[0-9])+) # (base 60)
-    )
-  _type_float: |- # http://yaml.org/type/float.html
-    (?x:
-        ([-+]? (?: [0-9] [0-9_]*)? (\.) [0-9.]* (?: [eE] [-+] [0-9]+)?) # (base 10)
-      | ([-+]? [0-9] [0-9_]* (?: :[0-5]?[0-9])+ (\.) [0-9_]*) # (base 60)
-      | ([-+]? (\.) (?: inf|Inf|INF)) # (infinity)
-      | (      (\.) (?: nan|NaN|NAN)) # (not a number)
-    )
-  _type_timestamp: |- # http://yaml.org/type/timestamp.html
+
+  # http://yaml.org/type/int.html
+  _type_int_binary: ([-+]?)(0b)([0-1_]+) # (base 2)
+  _type_int_octal: ([-+]?)(0)([0-7_]+) # (base 8)
+  _type_int_decimal: ([-+]?)(0|[1-9][0-9_]*) # (base 10)
+  _type_int_hexadecimal: ([-+]?)(0x)([0-9a-fA-F_]+) # (base 16)
+  _type_int_other: ([-+]?)([1-9][0-9_]*(?::[0-5]?[0-9])+) # (base 60)
+
+  # http://yaml.org/type/float.html
+  _type_float_decimal: ([-+]?)((?:[0-9][0-9_]*)?(\.)[0-9.]*(?:[eE][-+][0-9]+)?)
+  _type_float_other: ([-+]?)([0-9][0-9_]*(?::[0-5]?[0-9])+(\.)[0-9_]*) # (base 60)
+  _type_float_infitity: ([-+]?)((\.)(?:inf|Inf|INF)) # (infinity)
+  _type_float_nan: (\.)(?:nan|NaN|NAN) # (not a number)
+
+  # http://yaml.org/type/timestamp.html
+  _type_timestamp: |-
     (?x:
         \d{4} (-) \d{2} (-) \d{2}       # (y-m-d)
       | \d{4}                           # (year)
@@ -148,19 +152,11 @@ variables:
         )?                              # (time zone)
     )
 
-  _type_value: '=' # http://yaml.org/type/value.html
-  _type_merge: '<<' # http://yaml.org/type/merge.html
+  # http://yaml.org/type/value.html
+  _type_value: '='
 
-  _type_all: |-
-    (?x:
-        ({{_type_null}})
-      | ({{_type_bool}})
-      | ({{_type_int}})
-      | ({{_type_float}})
-      | ({{_type_timestamp}})
-      | ({{_type_value}})
-      | ({{_type_merge}})
-     )
+  # http://yaml.org/type/merge.html
+  _type_merge: '<<'
 
 ##############################################################################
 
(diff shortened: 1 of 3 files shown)


wbond:
`NaN` and `inf`: perhaps `meta.number.conceptual constant.numeric.value`?

wbond:
@jwortmann Perhaps you may suggest a better term for `NaN` and `inf` than "conceptual"?

jwortmann:
The usage of `NaN` and `inf` seems really special in YAML compared to other languages, where usually `constant.language` is used for these. Since the YAML documentation explicitly lists them under floating point numbers and requires a leading dot (`.NaN`/`.inf`), I don't have objections against the proposed scopes here in this PR with `meta.number.float.infinite` and `meta.number.float.not-a-number`. Perhaps shorten the latter one to `meta.number.float.nan`? If a combined scope name for both is desired, I would suggest `meta.number.float.other`.

`meta.number.float.other` is currently used for [sexagesimal](https://en.wikipedia.org/wiki/Sexagesimal#YAML) (base 60) numbers like `190:20:30.15`, but why not use the actual name for that base there, i.e. `meta.number.float.sexagesimal`? The same applies for base 60 integer numbers: `meta.number.integer.sexagesimal`.

deathaxe:
Also thought about `constant.language` as it would match the boolean/null scopes, but always found it useful to see those highlighted as numbers in the past.

`meta.number.float.nan` was my first shot as well, but I found it not enough self-explaining.

`meta.number.float.sexagesimal` is quite special. Introducing it might mean to introduce scopes for any base between 2..60. Erlang explicitly supports base 2..36 for instance. Bash seems to support any base. Therefore I find `meta.number.float.other` best here.

I'd like to keep an option to address those values special with the discussion about word-operators in mind.

If other languages know about those tokens, they might be scoped as numbers, too then. Maybe we shouldn't use the `meta.scope` to address them, but the a sub scope of value, then? Just in case other syntaxes use it for integers as well?

Maybe `constant.numeric.value.[infinite|nan]` ?

jwortmann:
I didn't know about the arbitrary bases in Erlang and Bash, but I see the PRs now. Maybe it's useful to make an exception for sexagesimal numbers in YAML, because they use their own unique notation and base 60 is arguably more important than other arbitrary bases. But I guess to inlcude it into `other` is okay too, because support for those numbers was dropped in YAML 1.2 anyway.

> If other languages know about those tokens, they might be scoped as numbers, too then.

Do I understand it correctly that you mean `inf` and `NaN` here? At least in Matlab and R, which have them built-in into the language, they are scoped under `constant.language` and I really would not like to change that. JavaScript also uses [`constant.language.nan.js`](https://github.com/sublimehq/Packages/blob/01d4e6007ee96ab10a3d1933023000502c680dd6/JavaScript/JavaScript.sublime-syntax#L2012) for `NaN` and [`constant.language.infinity.js`](https://github.com/sublimehq/Packages/blob/01d4e6007ee96ab10a3d1933023000502c680dd6/JavaScript/JavaScript.sublime-syntax#L2015) for `Infinity` and I think those scope names are the best. Now where I think about it, I would actually prefer these scopes from JavaScript for YAML too, to be consistent (and they feel more correct). The only reasons I see for using a `constant.numeric` scope in YAML would be that it is the currently used scope in this syntax, and the weird choice of the YAML documentation to list them under floating point numbers.
Your preference to highlight them with the color for numbers might be a unlucky coincidence in the Brackets schemes (which I believe you use) to use the same color for `constant.language` and `string` (especially for YAML).

Another suggestion you could possibly add to base 60 numbers in YAML would be to give the `:` characters in `190:20:30` an additional scope like it is done for the decimal point. In case of doing so, either `punctuation.separator.sexagesimal`, or maybe I would use `punctuation.separator.decimal` here as well, because that would include them if color schemes define such a rule, even though the latter scope wouldn't be 100% correct here.

wbond:
If all of the other syntaxes use `constant.language`, then I think YAML should also.

deathaxe:
I am fine with `constant.language.[nan|infity]` if we agree to use it in all syntaxes. I can follow your arguments with a look at `null`, `true`, ... which use the same scope family.

> weird choice of the YAML documentation to list them under floating point numbers

I partly disagree to call it `weird` from a sematic and mathematical point of view - at least for `inf` as infinity belongs to real numbers space. It may be arguable from the highlighting perspective though.

> ... which I believe you use

Yes, a slightly bugfixed/tweaked version ;-)

> ... add to base 60 numbers in YAML ... ` `: characters ... an additional scope

Would be possible, but...

1. Splits the number token in (too?) many small tokens
2. Would imply to also (want/propose to) scope other number separators (other than ordinary decimal dot), too.

We already agreed not to add special scopes for `e` exponents etc, so we shouldn't for separators, too.

wbond:
The logic of using `constant.language` being, if someone wants them to look like `constant.numeric`, they can do so via their color scheme. But color scheme authors shouldn't have to deal with "special snowflake" syntaxes.

deathaxe:
Just might feel a bit odd to see `constant.numeric.sign` and `punctuation.separator.decimal` in tokens like `-.inf` if `inf` uses 
`constant.language`.

wbond:
> Just might feel a bit odd to see constant.numeric.sign and punctuation.separator.decimal in tokens like -.inf if inf uses
constant.language.

That is true. In practice, I presume it is mostly a non-issue.

jwortmann:
> I partly disagree to call it weird from a sematic and mathematical point of view - at least for inf as infinity belongs to real numbers space.

Then let's be correct here :wink:. Infinity is *not* a real number. Real numbers are all numbers between -∞ and +∞.
https://math.stackexchange.com/questions/750777/is-infinity-a-real-number

> Yes, a slightly bugfixed/tweaked version ;-)

Very off-topic here and could be done somewhere else, but mind to share your customizations with me? For some things like operators I'd like to stick to the highlighting style in the original Brackets editor, but it's certainly possible that there still are some bugs around that I might want to fix then.

> We already agreed not to add special scopes for e exponents etc, so we shouldn't for separators, too.

Well, `:` is punctuation like the decimal dot is and hence would fit into the `punctuation.separator` scope, while `e` is not. But it was just a suggestion anyway.

> Just might feel a bit odd to see constant.numeric.sign and punctuation.separator.decimal in tokens like -.inf if inf uses
constant.language.

You could also use `keyword.operator`, since a number sign can also be interpreted as a unary operator. That's the typical usage in other syntaxes as well, at least for the "real" programming languages. E.g. see https://github.com/sublimehq/Packages/pull/2130

jwortmann:
Also the dots in `inf` and `NaN` are not decimal seperators, they should be part of the `constant.language` scope instead. Maybe you could argue for `punctuation.definition.constant`, but I wouldn't use that either. It's the same here like e.g. the boolean constants in Fortran: `.true.` and `.false.`.

deathaxe:
> Well, `:` is punctuation like the decimal dot

I don't think so. From the implementation it feels more like the separator between thousands, like `1.000.000.000`.

deathaxe:
Geht mir doch nich offn sagg.

jwortmann:
> 
> 
> > Well, `:` is punctuation like the decimal dot
> 
> I don't think so. From the implementation it feels more like the separator between thousands, like `1.000.000.000`.

I don't understand this. You want to argue that a colon is *not* punctuation? It just consist of two stacked points?! On the contrary, `e` is a letter, and therefore not punctuation.
Also the colon is used to separate each digit in the positional notation of base 60. So `separator` seems correct to me too.


> > they should be part of the constant.language scope
>
> They actually are. ;-)

I was referring to your previous post and hadn't looked at the updated code. But the dots still have the `punctuation.separator.decimal` scope currently stacked on top, which is just wrong and should be removed.


> `constant.numeric.sign`

I think we shouldn't really introduce this scope at all, and use `keyword.operator` instead in general, because it's strange to sometimes have the minus or plus sign colored as an operator, and in other occasions have it colored like a number. That is just inconsistent.

deathaxe:
1. I just argue to use `punctuation.separator` only to separate decimal places but nothing else.
2. I more and more understand and follow why YAML specs put `inf` into `float` scope. For me it feels odd and wrong to have `+` and `.` be part of a plain `constant.language`. I don't care about outdated languages such as fortran which were born in stone age. Most of those old languages consist of weird constructs of weak structure. `inf` is just a token to express a certain kind of value be it part of a real number or just expressing it a concept.
3. For arguments for `constant.numeric.sign` please refer to https://github.com/sublimehq/Packages/pull/2503.

deathaxe:
Another issue with removing `punctuation.separator` from infinity (e.g. `-.inf` ) is ligature tokenization. 

The `-.` is rendered as ligature, which is sub optimal, IMHO:

grafik

It looks more consistent with separate scope for dots.

grafik

jwortmann:
I don't get why the `+` or `-` should be part of the constant now. That is never the case in other syntaxes as well. Just use `keyword.operator` for the sign, then it won't be rendered as a ligature.
The only argument in #2503 for `constant.numeric.sign` I can find is your claim
> In some syntaxes it is even crucial to be able to distinguish operators from number signs

but only a vague hint that it applies to Clojure and OCaml and no concrete examples. I've never used Clojure nor OCaml, so perhaps you're right and the sign scope cannot be replaced by `keyword.operator` there.
I say it's inconsistent to sometimes highlight it as a number and sometimes as an operator, and that it's strange to have different rules for it in JavaScript Object Notation and in JavaScript.

And right above you wrote yourself
> to use punctuation.separator only to separate decimal places but nothing else

that would mean:
1. to remove it here, because it is clearly not the case here
2. add it to `:`, because that's exactly what `:` is used for, with the analogy to decimal places in base 60 positional notation

Perhaps you meant "to separate digits with positional value <1 and ≥1 only"?

wbond:
For consistency between syntaxes, which I think should generally be of higher importance than syntax-specific concerns, I think the next steps should be:

 - `.` gets `punctuation.separator` - consistency with pretty much all other syntaxes
 - `-` and `+` become `keyword.operator` - consistency between JavaScript and JSON seems pretty logical
 - `inf` and `NaN` get `constant.language` - consistency with most other languages

I think `-.inf` should be `meta.number` in its entirety, with the scopes defined above. I don't think it is important for the `.` to share any scope but `meta.number` with `inf`. I think it is fine for `meta.number` to contain `keyword.operator`.

FichteFoll:
The notion of `keyword.operator.sign` for signs is because most languages interpret them as unary operators, whereas in YAML there are no artithmtics in number parsing. From a syntactic standpoint the scope naming makes sense, but from the semantic PoV I say consistency with the other syntaxes is more important, especially if the result is the same. In the same sense, I would scope NaN and inf the same as for other languages.

Regarding `:`, I believe it should be compared to `_` in numeric literals of various languages, which can be inserted at arbitrary points in a number but is usually used as a thousands separator. Iirc, we don't currently scope this and I don't think we need to, so I don't see a particular need to scop `:` here either. As always, consistency is king.

jwortmann:
But `_` in numeric literals is only a visual thing used to make a number more readable, while `:` gives a meaning to a number, so I don't see the connection here. Could you clarify more?
I for example have a special rule in my color scheme for the `punctuation.separator.decimal` point, to applie the same color as the rest of the number (like in Monokai, Breakers etc.). But I could imagine that I want to highlight the `:` as separators, because they are very important for the value of a number, and since multiple `:` are possible in a singe number, they might be are harder to recognize visually than the decimal dot.
For example when I have the time 1 hour, 30 minutes, 5 seconds and write it as a number `1:30:5` it could be nice to seperate the different parts visually.

<hr>

For when `inf` or `nan` stand alone, they are strings, but only with the leading dot in `.inf` `.nan` they become a constant, so what is the reason to not include the dot in `constant.language` then?

deathaxe:
You argue against highlighting decimal separators by defining color scheme rules but want a mechanism to highlight lower prior separators. While most color schemes highlight operators with a dedicated color you decide to tint them as plain text. You have to oppose to any argument and seem not even to try to think about or understand others. Sorry, that's too cumbersome to continue discussing and never seems to and with a conclusion.

> only with the leading dot in `.inf` `.nan` they become a constant, so what is the reason to not include the dot in `constant.language` then?

To repeat myself again: The dot is actually part of `constant.language`.

---

> The notion of `keyword.operator.sign` for signs is because most languages interpret them as unary operators, whereas in YAML there are no artithmtics in number parsing.

That means you suggest to introduce `keyword.operator.sign` as alternative to `constant.numeric.sign`, right? 

Not only JSON and YAML don't have a concept of arithmetic operators other it may be important or at least useful in some syntaxes (like my CNC and even CSS) to be able to distinguish arithmetic operators from negative number signs.

The main intention of the `constant.numeric.[sign|base|value|suffix]` concept is to enable color-schems to highlight the whole token with a single color using `constant.numeric` but also being able to distinguish colors by addressing the more specific sub-scopes instead of needing complicated rules to like Breakets does to achieve the whole token to have the same color.

We can keep everything as it was with suffixes being scoped as `storage.type` or `keyword.other` etc. if we break that rule as it wouldn't help to increase consistency.

As I stated in

---

jwortmann:
> You argue against highlighting decimal separators by defining color scheme rules but want a mechanism to highlight lower prior separators

No, I suggested that it would be consistent to apply `punctuation.separator` to seperator punctuations, to be consistent with the practice for the decimal seperator. What makes you think that `:` is a lower priority separator than `.`?

<hr>

> While most color schemes highlight operators with a dedicated color you decide to tint them as plain text.

The color scheme is named "Brackets Color Scheme", and it is intended to follow the highlighting style from Adobe Brackets, which doesn't highlight operator symbols. I don't understand what this matter has to do with the discussion here about scopes in numbers and constants, though.

<hr>

> You have to oppose to any argument and seem not even to try to think about or understand others. Sorry, that's too cumbersome to continue discussing and never seems to and with a conclusion

No, I just give my opinion where I see inconsistent usages of scopes that makes no sense, so that it can be improved. **I have always stayed factual and gave very precise arguments about why I think what I think.** *You* were the one who wrote things like
> I don't care about outdated languages such as fortran which were born in stone age. Most of those old languages consist of weird constructs of weak structure.

I believe rules and scope guidelines should apply to all languages for consistency, and not just those used here in the default packages. If the dot is decimal punctuation in `.inf` and `.nan`, that would conclude that it is decimal punctuation in `.true.` as well, which just doesn't make sense to me. If you don't want to answer how you would scope the dots there, then I can only conclude that you ignore my arguments, and I agree that this discussion leads to nothing.

<hr>

> To repeat myself again: The dot is actually part of constant.language.

That question by me was directed to @wbond, who suggested that it should *not* be part of `constant.language`:

> * `.` gets `punctuation.separator` - consistency with pretty much all other syntaxes
> * `-` and `+` become `keyword.operator` - consistency between JavaScript and JSON seems pretty logical
> * `inf` and `NaN` get `constant.language` - consistency with most other languages

<hr>

> That means you suggest to introduce keyword.operator.sign as alternative to constant.numeric.sign, right?

I guess this question is directed to @FichteFoll , because I was not the one who suggested it.

<hr>

> Not only JSON and YAML don't have a concept of arithmetic operators other it may be important or at least useful in some syntaxes

Yes, and that is why I wrote that I agree with it in #2503. However, I wrote that I value consistency between languages higher, and therefore prefer to use the same scopes for everywhere, especially in JSON and JavaScript.

<hr>

> I give up, do what you want.

In general I don't understand why you became aggressive now, and closed this PR. You don't have to agree with me, but I try to give precise arguments for my opinions everywhere! And when I don't understand something, I ask them to clarify. I don't want to argue with you, all I want is to find consistent rules for scope names and allow useful highlighting in color schemes!

deathaxe:
Here's another example why distinguishing number signs from operators may be required/useful as they are just no operators in some situations, which must be visualized.

https://github.com/sublimehq/Packages/blob/a1c3d2a959ab631d027f2cc2b788e8641af6ac32/CSS/syntax_test_css.css#L381-L399

jwortmann:
> Here's another example why distinguishing number signs from operators may be required/useful as they are just no operators in some situations, which must be visualized.

1. This has no relevance for YAML, since there are no arithmethic operations between multiple numbers where a possible usage of `keyword.operator.arithmetic` for both number signs and binary operations could be confusing
2. Those two syntax tests in CSS are actually *wrong* and should be corrected:
>     top: calc(1px+1px); 
>     /*           ^ -keyword.operator.css */ 
>     top: calc(1px-1px); 
>     /*           ^ -keyword.operator.css */ 

The first one is to calculate the addition of two numbers, which *is* an arithmetic operation. Therefore the correct scope for `+` would be `keyword.operator.arithmetic` here!
The second line is to calculate the difference between two numbers, which *is* an arithmetic operation as well.
These two syntax test are currently wrong and should be corrected.

<hr>

I give you a few examples from other syntaxes where the `keyword.operator` scope is used for the sign of a number.

Matlab:
```matlab
foo = -inf
%     ^ keyword.operator.symbols.matlab
```
(The Matlab syntax chooses `keyword.operator.symbols` for all operator symbols, and `keyword.operator.arithmetic` doesn't occur at all in this syntax. But the current state of scope names used in the Matlab syntax is garbage in general.)

R:
```r
foo <- -Inf
#      ^ keyword.operator.arithmetic.r
```

JavaScript:
```js
var foo = -Infinity
//        ^ keyword.operator.arithmetic.js
```

Python:
```python
foo = -123
#     ^ keyword.operator.arithmetic.python

import math
foo = -math.inf
#     ^ keyword.operator.arithmetic.python
```


Java:
```java
int foo = -123;
//        ^ keyword.operator.arithmetic.java
```

C:
```c
int foo = -123;
//        ^ keyword.operator.arithmetic.c
```

D:
```d
int foo = -123;
//        ^ keyword.operator.arithmetic.d
```

Lua:
```lua
foo = -123
--    ^ keyword.operator.arithmetic.lua
```

Ruby:
```ruby
foo = -123
#     ^ keyword.operator.arithmetic.ruby
```

These are only a few examples, but basically it's the same everywhere.

jwortmann:
So to summarize, I would recommend the following scopes:

```yaml
  -.inf,  -.nan
# ^ keyword.operator.arithmetic.yaml
#  ^^^^ constant.language.inf.yaml - meta.number
#         ^ keyword.operator.arithmetic.yaml
#          ^^^^ constant.language.nan.yaml - meta.number

 190:20:30.15
#^^^^^^^^^^^^ meta.number.float.sexagesimal.yaml constant.number.value.yaml
#   ^ punctuation.separator.sexagesimal.yaml
#      ^ punctuation.separator.sexagesimal.yaml
#         ^ punctuation.separator.decimal.yaml
```
(or alternatively, convert all `punctuation.separator` scopes within numbers to uncolored meta scopes, but this was already rejected)

There still seems to be disagreement for my other two points too:

In contrary to this statement from @wbond
> `.` gets `punctuation.separator` - consistency with pretty much all other syntaxes

I say it would *not* be consistent; it would be very inconsistent instead, because `punctuation.separator.decimal` is only used for the decimal point within a number in other syntaxes, which is clearly not the case here, and it is never used for dots in built-in constants. So I don't understand why `punctuation.separator` should be consistent here, and if you think so, please give me an explanation for it.

Also nobody could give me a single argument yet for the claim why `:` separators in sexagesimal numbers would "feel" like the underscores in `1_000_000` notation. And I have disproven this (imo) by the argument that the underscores are only a visual thing to make a number more readable, but without any relevance for the compiler, whereas both the `:` in sexagesimal numbers and the decimal point are very significant for the actual value of a number, and they can not be omitted nor their position be changed without modifying the value or producing a syntax error. These are two completely different concepts. If we decide for scope names based on how things allegedly "feel" for someone, but ignore the facts, we lower ourself to statements like this: https://www.youtube.com/watch?v=17ne6-ypj_4

keith-hall:
@jwortmann related to CSS calc invalid scopes: https://github.com/sublimehq/Packages/issues/2356

I'm not sure if this has any relevance here but an expression like `-123` or `-variable` is functionally/mathematically equivalent to `0 - 123` or `0 - variable`. In the latter case, the minus is clearly an arithmetic operator, in the former it is an inversion sign, or it could be construed like an implicit zero minus... Probably not relevant for YAML, but maybe more interesting for expressions like `1 - -variable` etc. Maybe also not that common or worth worrying about?

wbond:
@deathaxe I just wanted to say thank you for your work on this. How the PR went down is _not_ how I like things to happen, as it was drawn out and seemed rather contentious.

I hope you are alright with me using your base to get this completed. If not, please let me know and I'll revert your changes and make changes on my own.

wbond:
@jwortmann In terms of the `punctuation` on `.inf` and `.nan`, I think that I was interpreting those as sort of "casting" the value to be float. From looking at the YAML grammar, I see that there is no version of those without the `.`, so it is really part of the identifier, and not something added on. Based on that, I think it should be just the `constant.language` and not get `punctuation`.

@deathaxe In order to preserve the different between `keyword.operator.arithmetic` as a sign on a number as just a plain operator, I've put the sign into `meta.number` in YAML. Hopefully this provides enough context for situations where a syntax and/or user wants to know which is which. A color scheme could choose to style `meta.number keyword.operator.arithmetic` the same color as `constant.numeric`. I am hoping this will be a reasonable solution for the syntaxes you are working on.

PR #2492, opened 2020-09-17
[TypeScript] Support function parameter decorators

### LINKED ISSUE #2480: [Typescript] Argument decorators not supported
- Sublime Version: 4086

```ts
class Controller {
    public async get(@Args() { id, credits }: GetArgs): Promise<Artist> {
        const url = `{id}`
        const b = callme();
    }
```

Screenshot 2020-09-15 at 15 18 02

The `@Args` decorator seems to be what is breaking the syntax.

@Thom1729

Thom1729:
Fix #2480.


wbond:
Looks like there is a test failure

Issue #2461, opened 2020-08-24
[JSX] Comment not working in HTML content


jfcherng:
**Disclaimer**: I don't write JSX.
**Disclaimer**: I don't write JSX.
**Disclaimer**: I don't write JSX.

But there must be something wrong here.

---

- Sublime Version: 4083
- OS Version: Win7 x64

```jsx
export default function Foo(props) {
  return (
    <div>
      <Foo>
      </Foo>
    </div>
  );
}
```

If you want to comment out the `<Foo>...</Foo>` part with `ctrl+/` (or `ctrl+shift+/`), it gives
![image](https://user-images.githubusercontent.com/6594915/91063100-1ca44d00-e660-11ea-9319-95b82971e8ca.png)

VSCode gives this
![image](https://user-images.githubusercontent.com/6594915/91063137-2a59d280-e660-11ea-97da-747408f86375.png)

## References

- https://forum.sublimetext.com/t/reactjs-syntax-highlighting-in-sublimetext/53412/5
- https://github.com/babel/babel-sublime/issues/390

michaelblyons:
Commenting should be fairly simple. You just make another tmPreferences for `source.jsx meta.jsx`. The interesting thing might be _uncommenting:_ Can you make it remove the whole `{/*...*/}` if it's contiguous, but fall back to `{foo}` if it sees `{foo /* here's a comment */}`?

Thom1729:
I've added a PR to copy over JS Custom's JSX commenting. It's not completely satisfactory. See https://github.com/Thom1729/Sublime-JS-Custom/issues/21 for another discussion.

Most likely, getting it perfect would require changing how the commenting command works, which is unlikely to happen.

jfcherng:
> Most likely, getting it perfect would require changing how the commenting command works, which is unlikely to happen.

Maybe need a dedicate built-in `.py` plugin to hijack the `toggle_comment` and perform (de-)commenting according to the cursor scope.

FichteFoll:
There are some issues related to how `toggle_comment` currently works, particularly https://github.com/sublimehq/sublime_text/issues/446.

PR #2494, opened 2020-09-17
[JSX] Commenting

Thom1729:
For #2461. See discussion there.


wbond:
Seems like a reasonable option for now

PR #2503, opened 2020-09-18
[JSON] Update meta.number

deathaxe:


JSON/JSON.sublime-syntax
```diff
@@ -68,12 +68,17 @@ contexts:
...
     # handles integer and decimal numbers
-    - match: -?(?:0|[1-9]\d*)(?:(?:(\.)\d+)(?:[eE][-+]?\d+)?|(?:[eE][-+]?\d+))
-      scope: constant.numeric.float.decimal.json
+    - match: (-?)((?:0|[1-9]\d*)(?:(?:(\.)\d+)(?:[eE][-+]?\d+)?|(?:[eE][-+]?\d+)))
+      scope: meta.number.float.decimal.json
       captures:
-        1: punctuation.separator.decimal.json
-    - match: -?(?:0|[1-9]\d*)
-      scope: constant.numeric.integer.decimal.json
+        1: constant.numeric.sign.json
```
wbond, line 74: In most other syntaxes this would be `keyword.operator`, right?

deathaxe: Depends on use case. If it is not possible to distinguish per syntax definition, syntaxes fall back to just use `keyword.operator`, yes. If possible with reasonable effort, number signs may be part of the number.  In some syntaxes it is even crucial to be able to distinguish operators from number signs to get a hint for correct syntax.

JSON doesn't know about operators and used to scope the sign as part of the number literal. Thus I found it most appropriate here. It should be easy to assign `constant.numeric.sign` the same color as `keyword.operator` if wanted. But there's no easy way to assign different colors if they share the same scope. Using `keyword.operator` would also imply to not include it into the `meta.number` scope as it might be weird to have `meta.number keyword.operator`, right?

I don't have a strong opinion about it. Would also be fine to use keyword.operator everywhere, for consistency and simplicity reasons. With a short looks like clojure and ocaml asign special scopes to signs atm. (Doesn't matter, but I want/need to distinguish it in my CNC language as well).

wbond: Alright, sounds reasonable. I'll merge as is.

jwortmann: Just reading this here. Regarding the discussion for YAML, it would be nice to find a general guideline for this first. I agree with some parts here from @deathaxe, such that the sign could be interpreted as part of the number in languages like JSON or CSS which don't have operators.

>  In some syntaxes it is even crucial to be able to distinguish operators from number signs to get a hint for correct syntax.

Do you have a concrete example for this, so we can see that/where `constant.numeric.sign` is really neccessary?

I'm not sure what's the best solution for JSON, but JSON is just "JavaScript Object Notation", so it is strange to have different rules for this here and in JavaScript (https://github.com/sublimehq/Packages/pull/2130).

deathaxe: Please read the comment to find it. Just because JS doesn't know about negative numbers it doesn't mean they don't exist.

PR #2508, opened 2020-09-20
[ShellScript] Fix numeric job identifier

deathaxe:
This commit removes `constant.numeric` from `%1` job identifiers as those already use `variable.language.job`.

Applying `meta.number constant.numeric` here would just be more odd.


wbond:
I agree, thanks!

PR #2509, opened 2020-09-20
[HTML] Split HTML syntax into plain vs. basic

deathaxe:
This PR suggests to move most basic HTML parts to HTML-Plain.sublime-syntax and inherit HTML.sublime-syntax from it.

This does not change any pattern or behavior of published HTML syntax, but provides a very basic HTML implementation to base work on, which doesn't need any of the embedded CSS/JS syntax definitions or all the different tag/attribute types available.

HTML-Plain is meant as base syntax for HTML supporting documentation comments or such kinds of template syntaxes.

The motivation for this step is JavaDoc, which can be inherit from HTML, but doesn't need CSS/JavaScript support. Injecting patterns to ignore leading asterisk punctuation in comments into HTML and all embedded syntaxes even creates unnecessary overhead and huge caches. It may even cause the context sanity limit to be hit soon.

**Example:**

```Java
  /**
   * <script
   *   type="text/javascript"
   * > the java script body
   * </script
   **/
```

grafik

Trying to support CSS/JS highlighting seems of no value as all JavaDocs seen out in the wild so far just use simple plain html tags.

Furthermore the current implementation of `id` and `class` attributes makes correctly clearing meta scopes hard or even impossible. We need another approach here.

Parsing times of JavaDoc like comments is increased significantly due to missing overhead, if it can be based on HTML-Plain.

The current HTML implementation is not really `basic` anymore.

### DIFF
--- a/HTML/HTML-Plain.sublime-syntax
+++ b/HTML/HTML-Plain.sublime-syntax
@@ -1,0 +1,359 @@
+%YAML 1.2
+---
+# This is a very low-level HTML implementation, which is meant to be used as
+# base definition for any kind of context/syntax, which just needs plain HTML
+# tag and attribute highlighting support without the overhead of:
+# a) embedded CSS/JS syntaxes
+# b) all the different kinds of tags and attributes.
+#
+# Any kind of high-level highlighting support such as additional tag types,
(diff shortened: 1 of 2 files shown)


wbond:
This sounds like a good change to me, thanks for taking the time to suggest it.

PR #2510, opened 2020-09-20
[SQL] Add storage types bool, enum, longtext and tinytext

anticultist:
This PR adds some missing storage types (bool, enum, longtext, tinytext) to improve the compatibility with MySQL/MariaDB.


wbond:
Thanks for the fixes!

PR #2511, opened 2020-09-20
[RUBY] Add Ruby Safe Navigation Operator(&.)

MatthewKennedy:
Add Ruby Safe Navigation Operator(&.)

Screenshot 2020-09-20 at 23 11 48

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -540,7 +540,7 @@ contexts:
     - match: =
       scope: keyword.operator.assignment.ruby
       push: after-operator
-    - match: '!+|&&|\|\|'
+    - match: '!+|&&|&\.|\|\|'
       scope: keyword.operator.logical.ruby
       push: after-operator
     - match: '[~&|^]'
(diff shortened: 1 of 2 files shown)


Ruby/Ruby.sublime-syntax
```diff
@@ -540,7 +540,7 @@ contexts:
     - match: =
       scope: keyword.operator.assignment.ruby
       push: after-operator
-    - match: '!+|&&|\|\|'
+    - match: '!+|&&|&\.|\|\|'
```
wbond, line 543: This probably should be part of this match since this isn't a logical operator. It should probably be a separate match with a scope such as `punctuation.accessor.safe.ruby`. The logic here being that `punctuation.accessor.dot.ruby` is used for `.`.

wbond (changes requested):
Thanks for taking the time to suggest this fix. I've made a suggestion on how to tweak the scope name to be more in line with https://www.sublimetext.com/docs/scope_naming.html.

Ruby/Ruby.sublime-syntax
michaelblyons: >  probably should be part

"shouldn't" ?

wbond: Yes, you are correct @michaelblyons. I meant to type “shouldn’t”.

PR #2521, opened 2020-09-21
[C#] Update missing meta.number constant.numeric

### LINKED ISSUE #2516: [C#] Highlighting numerics in formatting placeholders
- Sublime Version: 4086
- OS Version: Win 10

```cs
string multiple = String.Format("0x{0:X} {0,-15} {0,5:N1}", Int64.MaxValue);
```

How to highlight numeric literals in placeholders?

Python doesn't add special scopes, atm.

C# still scopes them `constant.numeric.integer`.

deathaxe:
Closes #2516

This commit replaces scopes of remaining numeric literals with old scope schemes for consistency reasons with the new guideline.

We may want to preserve numeric scopes in placeholder strings as their meaning may be compared to the formatting/substring operands in _Batch File_, which scopes them as numeric literals as well.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -129,7 +129,7 @@ contexts:
     - match: '\b(line)\s+(\d*)\s+((").*("))?'
       captures:
         1: keyword.other.preprocessor.cs
-        2: constant.numeric.integer.decimal.cs
+        2: meta.number.integer.decimal.cs constant.numeric.value.cs
         3: string.quoted.double.cs
         4: punctuation.definition.string.begin.cs
         5: punctuation.definition.string.end.cs
@@ -144,7 +144,7 @@ contexts:
           push:
             - meta_scope: string.quoted.double.hash.cs
             - match: '[-\h]+'
-              scope: constant.numeric.integer.hexadecimal.cs
+              scope: meta.number.integer.hexadecimal.cs constant.numeric.value.cs
             - match: '}"'
               scope: punctuation.definition.string.end.cs
               pop: true
@@ -1582,7 +1582,7 @@ contexts:
     - match: '(\{)(\d+)'
       captures:
         1: punctuation.definition.placeholder.begin.cs
-        2: constant.numeric.integer.decimal.cs
+        2: meta.number.integer.decimal.cs constant.numeric.value.cs
       push: string_placeholder
     - match: $\n?
       scope: invalid.illegal.unclosed-string.cs
@@ -1639,7 +1639,7 @@ contexts:
       scope: constant.other.placeholder.cs
       captures:
         1: punctuation.definition.placeholder.begin.cs
-        2: constant.numeric.integer.decimal.cs invalid.illegal.unclosed-string-placeholder.cs
+        2: meta.number.integer.decimal.cs constant.numeric.value.cs invalid.illegal.unclosed-string-placeholder.cs
 
   inside_string_placeholder:
     - match: '(\})(\}(?!\}))?'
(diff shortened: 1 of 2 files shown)


wbond:
> We may want to preserve numeric scopes in placeholder strings as their meaning may be compared to the formatting/substring operands in Batch File, which scopes them as numeric literals as well.

It isn't exactly clear to me what you are referring to here, mostly since I'm not extremely familiar with Batch File. What scope names are you suggesting we preserve in C#/Batch File?

deathaxe:
The main question is propably how to handle numeric literals within string format placeholders/expressions. I am a bit uncertain about the correct strategy, but finally found following the scheme used by Batch File might be a possibly good choice here especially as C# already used to apply numeric scopes.

While Python doesn't handle numeric literals in placeholders like `"{0} {1} {1:20}"` C# currently stacks `constant.numeric.integer.formatting` onto `constant.other.placeholder`. 

Batch File has constructs such as `%var:1:20%`. These are no format placeholders in the first place but may be compared to slice operations or `substr()` functions, but they look very simelar to format strings. They even may be nested, so could also be some kind of interpolation in the one way or the other.

I therefore see two possible options:

1. Keep scoping numeric literals within format placeholders and make them follow the scope naming guildelines.
2. Remove numeric scopes from such constructs for simplicity reasons (the python way) and just keep important separators/operators scoped.

Just took the first option to keep numeric scopes.

Maybe we need someone's oppinion who used C# actively here.

wbond:
I generally like the sound of adding `meta.number constant.numeric` to numeric arguments in formatting syntax or expansions like in Bash, especially if it is known that a value should be a number, and isn't just a text string that happens to contain digits.

C#/C#.sublime-syntax
```diff
@@ -1665,10 +1665,11 @@ contexts:
       scope: invalid.illegal.unexpected-character-in-placeholder.cs
 
   string_placeholder_format:
-    - match: '\s*(?:(,)\s*(-?\d+)\s*)?'
+    - match: '\s*(?:(,)\s*((-?)\d+)\s*)?'
       captures:
         1: punctuation.separator.arguments.cs
-        2: constant.numeric.integer.decimal.formatting.cs
```
keith-hall, line 1671: out of curiosity, was the removal of the `.formatting` specialization deliberate? As far as I can tell, it is now no longer possible for a color scheme to target the formatting number/padding amount separately from the argument number.
i.e. in the example `string multiple = String.Format("0x{0:X} |{0,-15}| -{0,5:N2}-", 12);`, the `0` after `{` has the same scope as the `15` and the `5`. Should we start another RFC for how to scope these?

For reference, the above example outputs:
`0xC |12             | - 12.0-`

So we can see that the `5`/`-15` reflects left/right padding. So I propose `meta.formatting.padding-amount` for this, which can be combined with the `meta.number` and/or `constant.numeric` for the color scheme to target.

deathaxe: Removed the formatting as I strictly followed the modified `meta.number constant.numeric` guidelines which basically remove all kinds of sematic relation from `constant.numeric` due to limiting the 3rd-level scope to `base|value|suffix` in the first place.

We may move the `formatting` to the `meta.number. ... ` scope if requried, as the meta scope now holds sematic information. Not sure if we want to create `constant.numeric.value.formatting` as I already saw things like `constant.numeric.value.exponent` to denote exponent part of floats.

I'd appreciate a common sense how to handle such formatting literals as we can find them in several languages. I agree with @wbond about having numbers scoped as such in formatting placeholders may be helpful and you are right, they should propably be distinguishable from "normal" numbers in some way.

Tokens like `{0:X}` are assigned  `constant.other.placeholder`. The `0` within therefore can be identified by `constant.other.placeholder constant.numeric`. I could also imagine to use `meta.placeholder constant.numeric` for such kinds of tokens, which whould match the scheme of `meta.interpolation` and avoid stacking different `constant.` scopes.

Not sure whether it was suitable for C-style formats though, which just use `%02d` or `%02.04f`. They just receive `constant.other.placeholder` atm and I guess it is the reason for python to scope `{0}` and friends the same way.

keith-hall: Thanks, I like your reasoning :) I will create an RFC issue about it when I can make time, unless someone beats me to it :)

PR #2524, opened 2020-09-21
[YAML] Update numeric indentation indicator

deathaxe:
This commit removes `integer.decimal` from indentation indicator literals as those sub scopes are no longer used.

As yaml version is scoped `constant.numeric.yaml-version`, `constant.numeric.decimal-indicator` should be sufficient here.


wbond:
Yes, I think those sub-scopes of `constant.numeric` seem sufficient.

PR #2525, opened 2020-09-21
[Git Formats] Update meta.number constant.numeric

deathaxe:
This commit applies updated meta.number scope naming guidelines.

Note: Do we really need all the `pretty-format` sub scopes everywhere?

### DIFF
--- a/Git Formats/Git Common.sublime-syntax
+++ b/Git Formats/Git Common.sublime-syntax
@@ -303,7 +303,7 @@ contexts:
           scope: meta.function-call.arguments.pretty-formats.git punctuation.section.parens.end.pretty-formats.git
           pop: true
         - match: \d+
-          scope: constant.numeric.integer.pretty-formats.git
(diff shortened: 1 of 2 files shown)


wbond:
> Note: Do we really need all the pretty-format sub scopes everywhere?

That's a good question to ask.

@michaelblyons Do you think that have the overarching meta scopes for `pretty-formats` would be sufficient and that we could discard the `pretty-formats` suffix from most scope names?

michaelblyons:
No objection. 👍

deathaxe:
I may add some commits with simplified scopes, here or do it in another PR.

wbond:
I'm going to merge this right now since I'm going to be pushing the current state of the repo to our internal mirror shortly in prep for the next build

deathaxe:
Looks like there are more possible changes/improvements with regards to string interpolation, which would be far off-topic anyway.

PR #2531, opened 2020-09-22
[YAML] Fix constant.numeric of decimal numbers

deathaxe:
This commit fixes decimal number values being scoped as `constant.numeric.base` by accident.


wbond:
Looks like there are a couple of base vs value test failures.

PR #2532, opened 2020-09-22
[CSS] Replace remaining `keyword.other.unit`

deathaxe:
This commit replaces the last remaining `keyword.other.unit` as it was decided to use `constant.numeric.suffix` for all kinds of units and storage type denoting suffixes in numeric literals.

This commit handles the only situation units appear as standalone token without a numeric literal in front of them.


wbond:
I think I'm going to hold off on this for now. I feel like it would be weird to have a unit by itself get a `constant.numeric` scope.

PR #2536, opened 2020-09-23
[CSS] Fix counter identifier

### LINKED ISSUE #2431: [CSS] Counter Identifiers with Numbers Show as Multiple Tokens
- Sublime Version: 4079
- OS Version: macOS 10.15.6

```css
ul > li:first-child
{
    counter-reset: counter1;
}
```

The entirety of `counter1` should be a single token. With CSS it shows as two, one being `counter` and the second being `1`, which is currently scoped as a number.

There are a handful of places these identifiers are valid: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters.

deathaxe:
Fixes #2431

This PR...

1. splits the anonymous property-names context in `rule-list-body` into several named contexts, such as `counter-property`, `font-property`,  ... .
2. Moves `counter-(...)` patterns from `other-properties` to a dedicated context in order to apply special scoping to the first property value as it denotes a counter name.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -616,126 +616,160 @@ contexts:
 
   rule-list-body:
     - include: comment-block
-    - match: "(?=[-a-z])"
+    - match: (?=[-a-z])
       push:
         - meta_scope: meta.property-name.css
-        - match: "$|(?![-a-z])"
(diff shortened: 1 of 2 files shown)


wbond:
Thanks for all of the CSS fixes!

PR #2539, opened 2020-09-26
[SQL] Add INDEX and KEY variations to improve compatibility with MySQL/MariaDB

anticultist:
This PR adds missing variations for INDEX (fulltext, spatial) and KEY (fulltext, unique) to improve the compatibility with MySQL/MariaDB.


wbond:
Thanks!

PR #2540, opened 2020-10-04
[SQL] Add various syntax tests to prevent regressions

anticultist:
Due to few syntax tests for the SQL language it is more likely to add regressions on modifications to the syntax file. This PR tries to address this issue. It adds syntax tests to preserve the status quo but also has SQL code which is not covered by the syntax definition file yet. To avoid failing tests that code parts have no scope annotations. Its like a collection of missing improvements for the syntax definition file. Some of them I will try to improve in the future.


wbond:
Looks good. Thanks for the additions!

PR #2541, opened 2020-10-08
[ShellScript] Fix ebuild/eclass file extensions

sharkdp:
These two patterns for Gentoo packaging files should be treated as file extensions, not as full filenames.


wbond:
Thanks!

PR #2548, opened 2020-10-12
[C#] improvements to the syntax highlighting to cater for C#9

keith-hall:
improvements to the syntax highlighting to cater for C#9 - including pattern matching :)

https://dotnetcoretutorials.com/2020/08/10/relational-pattern-matching-in-c-9/
https://devblogs.microsoft.com/dotnet/welcome-to-c-9-0/#improved-pattern-matching

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -972,8 +972,6 @@ contexts:
         1: support.type.cs
         2: variable.other.cs
       set: variables_declaration
-    # - match: \.
-    #   scope: punctuation.accessor.dot.cs
     - match: '(?=\S)'
       set: line_of_code_in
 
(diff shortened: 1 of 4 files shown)


C#/C#.sublime-syntax
```diff
@@ -983,8 +972,8 @@ contexts:
         1: support.type.cs
         2: variable.other.cs
       set: variables_declaration
-    - match: \.
-      scope: punctuation.accessor.dot.cs
+    # - match: \.
```
wbond, line 975: We should probably just kill these lines, right?

wbond (changes requested):
Looks good, just one stray commented-out rule.

wbond:
Thanks!

PR #2551, opened 2020-10-14
Enable TSX toggle comment

alecmev:
Not sure about the notation in `JSX Comments.tmPreferences`, but it works.

### DIFF
--- a/JavaScript/JSX Comments.tmPreferences
+++ b/JavaScript/JSX Comments.tmPreferences
@@ -5,7 +5,7 @@
     <key>name</key>
     <string>JSX Comments</string>
     <key>scope</key>
-    <string>source.jsx meta.jsx - source.jsx.embedded.jsx, source.ts meta.jsx - source.ts.embedded.jsx</string>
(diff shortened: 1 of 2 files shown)


Thom1729 (approved):

JavaScript/JSX Comments.tmPreferences
```diff
@@ -5,7 +5,7 @@
     <key>name</key>
     <string>JSX Comments</string>
     <key>scope</key>
-    <string>source.jsx meta.jsx - source.jsx.embedded.jsx, source.ts meta.jsx - source.ts.embedded.jsx</string>
+    <string>source.jsx meta.jsx - source.jsx.embedded.jsx, source.ts meta.jsx - source.ts.embedded.jsx, source.tsx meta.jsx - source.tsx.embedded.jsx</string>
```
wbond, line 8: With the restructuring, we not longer have `jsx` in `ts`, only `tsx`, so I think the suggestion below is more correct, right?

```suggestion
    <string>source.jsx meta.jsx - source.jsx.embedded.jsx, source.tsz meta.jsx - source.tsx.embedded.jsx</string>
```

alecmev: @Thom1729?

wbond:
Thanks @alecmev!

PR #2553, opened 2020-10-15
[GitHub Actions] Enhance ci.yml script (#2552)

### LINKED ISSUE #2552: [GitHub Actions]  Content-Disposition header?
In the GitHub workflow ci.yml, there's an invocation of wget with an experimental (according to wget docs) `--content-disposition` option. It should instruct wget to "honor the Content-Disposition header when choosing local file names".

https://github.com/sublimehq/Packages/blob/7608ae7d6c33850d680a0e2a07d49f409b6c817f/.github/workflows/ci.yml#L19

Either I'm missing something, or the server's response doesn't even contain such header:

```http
HTTP/1.1 200 OK
Server: nginx
Date: Wed, 14 Oct 2020 13:58:52 GMT
Content-Type: application/octet-stream
Content-Length: 1118840
Last-Modified: Mon, 14 Sep 2020 06:18:34 GMT
Connection: keep-alive
ETag: "5f5f0b3a-111278"
Accept-Ranges: bytes

...rest is tar.gz binary content...
```

On the other hand, why using this option at all? Automation scripts are no place for random filenames dictated by the server. It would be much more robust approach to use deterministic file names, e.g. with `-O/--output-document` option:

```sh
wget -O st_syntax_tests.tar.xz https://url...
```

ratijas:
Fixes #2552


ratijas:
As soon as this is accepted, I will upload a mirror patch to [SublimeText/syntax-test-action](https://github.com/SublimeText/syntax-test-action/blob/cd9fa51ddd87ad43b77c8f918bb7788b3f6adc52/syntax-tests.sh#L33).

FichteFoll (approved):

FichteFoll:
FIY, you are free to already start work on the syntax-test-action, since I maintain that. Alas (via #2552), I don't think the usage of globs there is inherently wrong or bad, but I'm open to reasonable alternatives.

ratijas:
Welp, good thing I didn't rely on this for the hacktoberfest challenge too much :laughing: But hey, there's still two day left for reviewing.

Nevermind. I'm happy it finally made it upstream.

PR #2555, opened 2020-10-16
[HTML] Add dedicated CSS/JS content contexts

deathaxe:
This commit intends to make it easier for an inherit HTML syntax to replace the plain `source.js` and `source.css` by inherit definitions.

A syntax such as ASP or JSP needs to inherit CSS and JS in order to extend them by ASP/JSP tag support. Those extended CSS/JS variants need to be embedded in HTML-ASP or HTML-JSP then.

The strategy may be compared to what Makefile does with Bash via Makefile Shell.sublime-syntax.

Unfortunately the `embed:` command doesn't accept variables which could be used to replace the scope name more easily/consistently. So the dedicated context is the only way to minimize required code duplication.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -199,10 +199,13 @@ contexts:
       scope: punctuation.definition.tag.end.html
       set:
         - include: script-close-tag
-        - match: (?=\S)
-          embed: scope:source.js
-          embed_scope: source.js.embedded.html
-          escape: '{{script_close_lookahead}}'
+        - include: script-javascript-content
+
+  script-javascript-content:
+    - match: (?=\S)
+      embed: scope:source.js
+      embed_scope: source.js.embedded.html
+      escape: '{{script_close_lookahead}}'
 
   script-html:
     - meta_scope: meta.tag.script.begin.html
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

Thom1729:
Seems fine. I'd only add that the embedding can probably be simplified now that `embed` can coexist with `push`.

deathaxe:
Not sure if I can follow.

Embed is called after pushing into the context for 2 reasons.

1. Match possible `<!--` before switching to CSS/JS scopes (works only for JS atm due to `(?=\S)` though).
2. The closing tag may be unclosed `</script` or the `>` may be on the next line. That use-case is not possible via `escape_captures`, hence using the lookahead for `</script` only.

If we decide to drop commented CSS/JS and be satisfied with escaping via complete tags `</script>` only, we can simplify those patterns, yes.

But that won't solve my propblem with exchanging the basic CSS and basic JS syntaxes by extended ones in JSP syntax without also copying all the tag-name matching stuff around that.

Ideally `embed` should support variables such as:

```
- match: whatever
  embed: 'scope:{{my_scope_name}}`
```

Thom1729:
I forgot that the comment stuff complicated things. If it weren't for that, we could do this:

```yaml
  script-javascript:
    - meta_scope: meta.tag.script.begin.html
    - include: script-common
    - match: '>'
      set:
        - script-close-tag
        - script-javascript-content

  script-javascript-content:
    - match: ''
      embed: scope:source.js
      embed_scope: source.js.embedded.html
      escape: '{{script_close_lookahead}}'
      pop: true
```

But that doesn't support comments, and modifying it to support comments would not be simpler than this PR.

deathaxe:
Does that really work? (... asking without having tried that so far). According to the documentation `pop` is performed` before `embed` when combined.

> When combined, the specified number of contexts will be popped off of the stack before the other action is performed.

Thom1729:
It works — except for the comment thing, which I'd forgotten about. YMMV on which way is simpler, but since my idea doesn't work with comments, it's a moot point.

Thom1729 (approved):

PR #2556, opened 2020-10-18
[CSS] Major Upgrade including syntax-version 2

### LINKED ISSUE #1017: [CSS] uppercase HTML tags and CSS properties are currently unscoped
I've recently seen some HTTP responses from some Microsoft products, which unfortunately seem to be stuck in the 90s and have HTML tags and CSS properties in upper case.

According to the CSS spec, most tokens should be treated as case-insensitive, but the CSS syntax definition in this repo currently only scopes everything if it is lowercase.

https://www.w3.org/TR/css-syntax-3/#style-rule
> Tokens match if their value is an ASCII case-insensitive match for the value defined in the grammar.

More related reading: https://stackoverflow.com/a/42959444/4473405

(Ugly) examples:

```css
A {
    FONT-WEIGHT: bold; FONT-SIZE: 10pt; COLOR: #005a80; FONT-FAMILY: tahoma
}

A:hover {
    FONT-WEIGHT: bold; FONT-SIZE: 10pt; COLOR: #0d3372; FONT-FAMILY: tahoma
}
```

here, the `A` should be scoped as `entity.name.tag.css`, and `FONT-WEIGHT` for example as `support.type.property-name.css`

### LINKED ISSUE #1100: [CSS] Parentheses in parentheses still breaks highlighting; Property in animation looks strange.

deathaxe:
While the original intention of this PR was to upgrade the syntax definition to sublime-syntax version 2 and apply some related changes which might make sense:

1. sorting header contexts (see: #2430)
2. replacing `pop: true` by `pop: 1`
3. add some sections for logical grouping contexts

it more and more escalated to a major update with various other fixes and improvements.

1. Fixes #1017 by consequently matching all tokens case-insensitive (except. custom-element-tags). Some of them already were, others not.
2. Fixes #1100 by adding contexts to ensure bracket balancing, if unexpected tokens are matched. This change was made to comply with CSS specifications which ask to ensure balancing under all circumstances.
3. Improves bailout strategy from blocks and groups to ensure not to break highlighting of the whole document just because of a missing `)` or `}`.
4. Rule boundaries are respected according to the specs. At-rules span everything beginning with `@` up to `;` or `{...}` for instance.
5. At-rules have been updated to handle nested declaration-lists. (see `@page` or `@supports`).
6. Efforts to support goto definition and goto reference were made with regards to `@counter` related entity names. _(Needs some scope changes in the future though)_.
7. Some logical operators have been added in `@media`.
8. Larger lists of predefined tokens have been moved to variables.

Further details may be found in related commit messages.

### Performance

According to the most recent benchmarks parsing time is the same as with the old syntax. While several changes and optimizations reduced parsing times by 45% all of those improvements are eaten by case-insensitive patterns.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -883,8 +883,10 @@ contexts:
 
   media-query-group-content:
     - meta_scope: meta.group.css
+    - match: (?=[,;{}])
+      pop: 1
     - include: group-end
-    - include: at-rule-end
+    - include: comments
     - include: media-query-conditions
     - include: var-functions
     - include: calc-functions
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -825,6 +825,7 @@ contexts:
   at-supports-group-content:
     - meta_scope: meta.group.css
     - include: group-end
+    - include: at-rule-end
     - include: at-supports-groups
     - include: at-supports-operators
     - include: rule-list-body
@@ -883,6 +884,7 @@ contexts:
   media-query-group-content:
     - meta_scope: meta.group.css
     - include: group-end
+    - include: at-rule-end
     - include: media-query-conditions
     - include: var-functions
     - include: calc-functions
@@ -1078,8 +1080,7 @@ contexts:
           scope: punctuation.section.group.begin.css
           set:
             - meta_scope: meta.function-call.arguments.css meta.group.css
-            - include: group-end
-            - include: var-functions
+            - include: function-arguments-common
             - include: unquoted-strings
         - include: immediately-pop
 
@@ -1091,7 +1092,7 @@ contexts:
           scope: punctuation.section.group.begin.css
           set:
             - meta_scope: meta.function-call.arguments.css meta.group.css
-            - include: group-end
+            - include: function-arguments-common
             - include: selectors
         - include: immediately-pop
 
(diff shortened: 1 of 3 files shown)


CSS/CSS.sublime-syntax
```diff
@@ -92,1903 +432,1490 @@ contexts:
...
-    - match: '{{ident}}'
-      scope: variable.other.counter.css
-      pop: true
-    - match: (?=\S)
-      pop: true
+  # Custom Properties
+  # https://drafts.csswg.org/css-variables/#typedef-custom-property
+  custom-properties:
+    - match: (--)({{nmchar}}+)
+      scope: support.type.custom-property.css
```
deathaxe, line 982: I am not very happy about `support.type` at all, but also the fact those are stacked on each other here.

Thoughts?

deathaxe:
I am done with, what I'd like to do at this point.

From here on, only requested changes from reviewers might be added.

wbond:
I didn't read every line of the changes, but did peruse it a bit, and checked out the plugin changes.

Looks like a significant amount of work and a whole bunch of useful changes. Thanks for all of your work on this!

Issue #618, opened 2016-09-12
Javadoc syntax highlighting doesn't work in JS files


3Nex:
Javadoc syntax highlighting works in PHP files for example, highlighting keywords such as @param, @return etc. 
![javadoc_highlighting](https://cloud.githubusercontent.com/assets/851674/18429215/a13c10c0-78d2-11e6-97e1-cdd994dd761d.png)

But none of it works in Javascript files. 

(I'm not sure if there's a valid reason behind it maybe?)

wbond:
This is likely because no one has implemented it.

Further, I think that JS has not standardized on JavaDoc syntax like PHP has.

I'd be interested in getting info from the JS community about documentation standards before taking the time to implement this.

3Nex:
Originally I just wanted to know if it's in any way configurable so that I could get that to work at least just for myself. But I was advised on the Sublimetext forum to report it as an issue. 
https://forum.sublimetext.com/t/can-i-get-javadoc-syntax-highlighting-in-js-files/22762

Is it possible to just add the `.js` file extension somewhere in some config file to get it to work?

wbond:
No, syntax highlighting isn't a case of configuration, but rather writing regex rules as part of the syntax itself. There isn't a way to turn on or off parts of highlighting via settings at the moment.

I think the way forward for this would be to look up info about the various JS block comment syntax modes to try and see which would be the best for us to support. Or perhaps we can find that multiple can be supported at the same time since the syntax does not conflict.

FichteFoll:
I think the most commonly used standard (if there even are multiple) is jsdoc: http://usejsdoc.org/

I'm not a webdev, but I did use that the last time I was involved with it.

Thom1729:
@3Nex If you're still interested in this functionality, open an issue at https://github.com/Thom1729/Sublime-JS-Custom/issues and I'll see what I can do.

jingyuexing:
> 
> 
> Javadoc syntax highlighting works in PHP files for example, highlighting keywords such as `@param`, `@return` etc.
> 
> But none of it works in Javascript files.
> 
> (I'm not sure if there's a valid reason behind it maybe?)
> 
> ![javadoc_highlighting](https://cloud.githubusercontent.com/assets/851674/18429215/a13c10c0-78d2-11e6-97e1-cdd994dd761d.png)

user Vscode syntax defined file  transformation to sublime syntax
## like this :
[...]

PR #2557, opened 2020-10-20
[JavaScript] Initial support for JSDoc documentation syntax

rchl:
Initial changes adding support for block tags [1]. There is no validation of tag name as JSDoc accepts any unknown tag name by default.

Block tags are only allowed at the beginning of the comment line (after comment-starting pattern "/**" on the first line, or after "*" on the following lines) [2].

[1] https://jsdoc.app/about-block-inline-tags.html
[2] https://jsdoc.app/about-configuring-jsdoc.html#default-configuration-options

Relates to #618


rchl:
Things like scopes (and by extension colors) to use are very much up for discussion.

I've tried a bunch of alternative scopes like (with color as seen in Mariana):
 - keyword.other - pink
 - entity.name - orange
 - entity.name.tag - red
 - support.type - blue (this should be used for actual JSDoc type in the future)
 - support.type.property-name - white

I'm not entirely happy with any of them.

It might be a good idea for documentation tags to use fainter colors so that those don't stand out more than the code itself. But there is not really any existing mapping with such colors so an option would be to add new ones. That would mean that all color schemes would have to be updated to support those tags...

It also could be that using existing colors/scopes is OK and it's just a matter of getting used to.

deathaxe:
Choosen scopes are ok. Not sure if jsdoc, but you might want to save some context switches or `set-set-set` chains.

```
  comments:
    - match: /\*\*(?!/)
      scope: punctuation.definition.comment.begin.js
      push:
        - jsdoc-comment
        - jsdoc-tag
    // ...

  jsdoc-comment:
    - meta_scope: comment.block.documentation.js
    - match: \*/
      scope: punctuation.definition.comment.end.js
      pop: 1
    # JSDoc "block" tags (i.e. @param) are only accepted at the beginning of the documentation
    # line so directly after the '/**' or after the '*' marker on the next lines.
    - match: ^\s*(\*)(?!/)
      captures:
        1: punctuation.definition.comment.js
      push: jsdoc-tag

  jsdoc-tag:
    - match: '{{jsdoc_tag}}'
      scope: meta.tag.jsdoc entity.name.jsdoc
      pop: 1
    - match: (?=\S)|$
      pop: 1
```

rchl:
Thanks for your comment. This looks like an awesome suggestion - much simpler.

FichteFoll:
Not exactly feeling the usage of `entity.name.jsdoc` here, notably because `entity.name.xxx` is the usual pattern and because tags are keywords moreso than definitions. As such, a keyword seems more appropriate to me, so something like `keyword.other` seems fine.
The only downside I see is that keyword scopes tend to receive high prority colors that you wouldn't necessarily want inside a comment. Color schemes *could* use a selector like `comment keyword`, but realistically I don't think any scheme does currently.

Do we have any other instances of highlighting keywords in documentation comments?

rwols:
> Do we have any other instances of highlighting keywords in documentation comments?

Yes. Paste this into a new view

```csharp
/// <summary>This method changes the point's location by
///    the given x- and y-offsets.
/// <example>For example:
/// <code>
///    Point p = new Point(3,5);
///    p.Translate(-1,3);
/// </code>
/// results in <c>p</c>'s having the value (2,8).
/// </example>
/// </summary>

public void Translate(int xor, int yor) {
    X += xor;
    Y += yor;
} 
```
and set the syntax to C#.

deathaxe:
- C# uses xml notation.
- JavaDoc is based on HTML and additionally supports `@param` like tags via `keyword.other.documentation`.
- Perl supports `=POD` tags using `entity.name.tag.pod`

In order to achive some common scoping I'd find `entity.name.tag` not too bad for `@param` and friends.

With regards to JavaDoc `keyword` makes those `@...` tags stand out as normal HTML tags in the content are highlighted as `entity.name.tag` already.

FichteFoll:
I can get behind `entity.name.tag`, especially since that is already used in HTML, but not `entity.name.jsdoc`.

JavaDoc using a keyword scope for `@param` kind of defeats the purpose of common scoping for such constructs, though. I get why it was done, but I don't think it's a good decision.

deathaxe:
JavaDoc is part of the current work to get the syntax up to date. I am open to change the scope of those keywords especially due to the argument of `keyword` being highlighted very prominently in most color schemes.

We should propably find a common sub-scope to distinguish it from normal html tags though. Maybe something like `entity.name.tag.doc` or `entity.name.tag.documentation` ... not sure.

rchl:
As for the colors, I think `entity.name.tag.documentation` wins because `entity.name.jsdoc` has the same color as arguments in a function while those are not arguments.

`entity.name.tag.documentation`
Screenshot 2020-10-25 at 18 25 04

`entity.name.jsdoc`
Screenshot 2020-10-25 at 18 25 36

The argument names in documentation comments will also be highlighted and scoped in the future so it would be good to match the color in both places.

And note that those are just initial changes and it will eventually look even more "festive" as both block tag, the type, and parameter name will have scopes and probably colors assigned. Might look something like:
Screenshot 2020-10-25 at 18 32 33

rchl:
BTW. There is also PHPDoc which uses `keyword.other.phpdoc.php` for tag names:
Screenshot 2020-10-25 at 18 37 07

deathaxe:
Yells for a RFC to note/discuss the desired scopes to apply to all syntaxes.

rchl:
Created https://github.com/sublimehq/Packages/issues/2569 to discuss the scopes but hopefully, we can decide on how to scope the `@` block tags at least so that there is a chance to get this minimal support in at least.

wbond:
I commented at https://github.com/sublimehq/Packages/issues/2569#issuecomment-726101166, voicing my support for `entity.name.attribute-name.documentation`.

deathaxe:
`entity.name.` or `entity.other`?

wbond:
> entity.name. or entity.other?

Good catch! It should be `entity.other`.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -136,6 +133,26 @@ contexts:
...
+    # JSDoc "block" tags (i.e. @param) are only accepted at the beginning of the documentation
+    # line so directly after the '/**' or after the '*' marker on the next lines.
+    - match: ^\s*(\*)(?!/)
+      captures:
+        1: punctuation.definition.comment.js
+      push: jsdoc-block-tag
+
+  jsdoc-block-tag:
+    - match: '{{jsdoc_block_tag}}'
+      scope: meta.tag.jsdoc entity.other.attribute-name.documentation.jsdoc
```
wbond, line 151: I would probably drop the `meta` scope unless you are going to scope the entire line with it. The suffix on the scope should just be `js`, IMO.

```suggestion
      scope: entity.other.attribute-name.documentation.js
```

wbond:
Awesome, thanks for your work on this @rchl!

PR #2562, opened 2020-10-24
[Matlab] Fix ellipsis comments

### LINKED ISSUE #2561: [MATLAB] Comment at end of line and continuation
Comments at end of line and continuation are currently not highlighted as comment. See https://www.mathworks.com/help/matlab/matlab_prog/comments.html for a official reference of my minimal example.

- OS Version: Ubuntu 20.04.1 LTS

Minimal example:
``` matlab
var_name = 1 + 2 ... comment at end of line and continuation
    3;
```

Screenshot in Sublime:
image
We see here that after the `...` the text is highlighted as code. But it should be highlighted as a comment.

deathaxe:
Fixes #2561

### DIFF
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -48,6 +48,10 @@ contexts:
       scope: comment.line.percentage.matlab
       captures:
         1: punctuation.definition.comment.matlab
+    - match: (\.{3}).*$\n?
+      scope: comment.line.ellipsis.matlab
+      captures:
+        1: punctuation.definition.comment.matlab
   all_matlab_keywords:
     - include: matlab_keyword_control
     - include: matlab_keyword_operator
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -48,10 +48,6 @@ contexts:
       scope: comment.line.percentage.matlab
       captures:
         1: punctuation.definition.comment.matlab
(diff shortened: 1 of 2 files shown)


Matlab/Matlab.sublime-syntax
```diff
@@ -48,6 +48,10 @@ contexts:
       scope: comment.line.percentage.matlab
       captures:
         1: punctuation.definition.comment.matlab
+    - match: (\.{3}).*$\n?
+      scope: comment.line.ellipsis.matlab
+      captures:
+        1: punctuation.definition.comment.matlab
```
jwortmann, line 54: I would suggest an implementation with something like the following:
```yaml
- match: (\.{3})(.*$\n?)
  captures:
    1: punctuation.separator.continuation.line.matlab
    2: comment.line.ellipsis.matlab
```

This would allow the `...` to not be rendered as comment, because its main purpose is line continuation. The [official .tmLanguage grammar](https://github.com/mathworks/MATLAB-Language-grammar/blob/40d9a0cd3b628f80cdcf948bbe1747a527ed5dd5/Matlab.tmbundle/Syntaxes/MATLAB.tmLanguage#L1375-L1396) maintained by Mathworks uses `keyword.operator.symbols` for it, but I think `punctuation.separator.continuation.line` is more consistent with the usage in other default syntaxes here.

deathaxe: Reusing the existing continuation pattern.

PR #2563, opened 2020-10-24
[Matlab] Merge percentage comment patterns

deathaxe:
This commit merges double-percentage and single-percentage comments.

The result is all types of comments beginning with an arbitrary number of percentage sign are accepted with highlighting all percentages as `punctuation.definition`.

### DIFF
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -40,7 +40,11 @@ contexts:
           captures:
             1: punctuation.definition.comment.matlab
           pop: true
-    - match: (%+).*$\n?
+    - match: (%%(?!%)).*\n?
+      scope: comment.line.double-percentage.matlab
+      captures:
+        1: punctuation.definition.comment.matlab
+    - match: (%+).*\n?
       scope: comment.line.percentage.matlab
       captures:
         1: punctuation.definition.comment.matlab
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -31,20 +31,16 @@ contexts:
         2: punctuation.accessor.dot.matlab
       push: transpose_post_parens
   all_matlab_comments:
-    - match: (%%).*$\n?
-      scope: comment.double.percentage.matlab
-      captures:
-        1: punctuation.definition.comment.matlab
-    - match: '^\s*(%\{)\s*\n'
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):

jwortmann:
I don't think this change would be useful, because it would remove the ability for color schemes to distinguish between single and double `%`-comments.

For example the Matlab IDE displays `%%`-comments in bold style, because they are used as section headers and to separate code blocks.

deathaxe:
1. Those are not distinguished at https://www.mathworks.com/help/matlab/matlab_prog/comments.html
2. `comment.double` does not comply with the scope naming guideline. At least level 1-2 better 1-3 should stay as close as possible.
3. Recent syntax modifications introduced highlighting of arbritary numbers of comment definition chars as `punctuation.definition.comment` for a better highlighting experience in case punctuation is addressed by color-schemes.

   after

   grafik

   vs. before

   grafik

jwortmann:
> 2. comment.double does not comply with the scope naming guideline. At least level 1-2 better 1-3 should stay as close as possible.

Yes, sounds reasonable. How about `comment.line.double-percentage` for 2 or more `%` then, so that the ability to highlight section titles differently from normal comments is preserved?
https://www.mathworks.com/help/matlab/matlab_prog/run-sections-of-programs.html

deathaxe:
With regards to the scope name and the explicit definition of two percentage signs to denote sections, I'd prefer to explicitly scope comments beginning with exactly two percentages in a special way, only.

Scoping `%%% hello world` as `double-percentage` seems odd.

jwortmann:
Okay that seems fine, but the current state of this PR doesn't reflect this. What I try to explain is, that it would be unfortunate to loose the ability to distinguish between section title comments (`%%`) and "normal" comments (`%`) here in the ST syntax, while this is possible in other editors, and in particular different styles are applied for them in the Matlab IDE.

So `comment.line.double-percentage.matlab` for lines with exactly 2 percentages, and in case of 1 or more than 2 percentages `comment.line.percentage.matlab` then?

deathaxe:
It does reflect it. See https://github.com/sublimehq/Packages/pull/2563/commits/95b79a3f33534003f9c449506ec70db9342870a8

It's just out of order because the solution was ready before you expressed your critics. :D

jwortmann:
Alright, I can see the change now. Looks good to me now.

PR #2564, opened 2020-10-25
[YAML] Replace entity.name.tag with meta.mapping.key

FichteFoll:
I've been wanting to do this for a while, as `entity.name.tag` was really just a hack for a lack of an appropriate scope name when I made this syntax. Now that JSON, JavaScript and Python (and probably other syntaxes) successfully use `meta.mapping.key`, YAML should not be exempted.

---

*(via https://github.com/sublimehq/Packages/pull/2564#pullrequestreview-894390337)*

Without addressing https://github.com/sublimehq/sublime_text/issues/3590, the UX of default color schemes with regards to highlighting keys differently from values will be *worse* but in line with JSON and the other languages using `meta.mapping.key` as the primary differntiator between keys and values.

The only thing needed for Mariana is something like:

```json
{
	"name": "Mapping Key Names",
	"scope": "meta.mapping.key string",
	"foreground": "var(red2)"
}
```

Monokai should be fixed as well to maintain the highlighting users are used to.

### DIFF
--- a/PHP/tests/syntax_test_php.php
+++ b/PHP/tests/syntax_test_php.php
@@ -1935,12 +1935,12 @@ echo <<< yml
 //       ^^^ entity.name.tag.heredoc
 one: two
 //^^^^^^ meta.embedded.yaml source.yaml
-//^ string.unquoted.plain.out entity.name.tag
+//^ meta.mapping.key string
 // ^       punctuation.separator.key-value.mapping
-//   ^^^ string.unquoted.plain.out
+//   ^^^ string
 three: "$four"
 //^^^^^^^^^^^^ meta.embedded.yaml source.yaml
-//^^^ string.unquoted.plain.out entity.name.tag
+//^^^ meta.mapping.key string
 //   ^       punctuation.separator.key-value.mapping
 //     ^^^^^^^ string.quoted.double
 //      ^^^^^ variable.other.php
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -37,6 +37,7 @@
 ---
 name: YAML
 scope: source.yaml
+version: 2
 
 file_extensions:
   - yaml
@@ -107,6 +108,34 @@ variables:
           \s* $
         | \s+ \#
         | \s* : (\s|$)
+      )
+    )
+
+  _flow_key_in_lookahead: |-
+    (?x:
+      (?=
+        {{ns_plain_first_plain_in}}
+        ( [^\s:{{c_flow_indicator}}]
+        | : [^\s{{c_flow_indicator}}]
+        | \s+ (?![#\s])
+        )*
+        \s*
+        :
+        (\s|$)
+      )
+    )
+
+  _flow_key_out_lookahead: |-
+    (?x:
+      (?=
+        {{ns_plain_first_plain_out}}
+        ( [^\s:]
+        | : \S
+        | \s+ (?![#\s])
+        )*
+        \s*
+        :
+        (\s|$)
       )
     )
 
@@ -512,15 +541,13 @@ contexts:
     - match: \[
       scope: punctuation.definition.sequence.begin.yaml
       push:
-        # Pushes an extra 'meta' on the scope stack that flow-pair can then clear,
-        # so we don't have to duplicate the context
-        - meta_scope: meta.sequence.flow.yaml meta.workaround.yaml
+        - meta_scope: meta.sequence.flow.yaml
         - match: \]
           scope: punctuation.definition.sequence.end.yaml
           pop: true
         - match: ','
           scope: punctuation.separator.sequence.yaml
-        - include: flow-pair
+        - include: flow-pair-no-clear
         - include: flow-node
 
   flow-mapping:
@@ -534,50 +561,83 @@ contexts:
         - match: ','
           scope: punctuation.separator.mapping.yaml
         - include: flow-pair
+        - include: flow-node  # for sets
 
   flow-pair:
     - match: \?
-      scope: meta.mapping.key.explicit.yaml punctuation.definition.key.begin.yaml
+      scope: meta.mapping.key.yaml punctuation.definition.key.begin.yaml
       push:
-        - - match: ''
-            set: flow-pair-value
-        - - clear_scopes: 1
-          - meta_content_scope: meta.mapping.key.explicit.yaml
-          - match: (?=[},\]])  # Empty mapping keys & values are allowed
-            pop: true
-          - match: :(?=\s|$|{{c_flow_indicator}})
-            scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
-            pop: true
-          - include: flow-node
+        - flow-pair-value-set
+        - flow-pair-key
     # Attempt to match plain-in scalars followed by a colon
-    - match: |
-        (?x)
-        (?=
-          {{ns_plain_first_plain_in}}
-          ( [^\s:{{c_flow_indicator}}]
-          | : [^\s{{c_flow_indicator}}]
-          | \s+ (?![#\s])
-          )*
-          \s*
-          :
-          (\s|$)
-        )
+    - match: '{{_flow_key_in_lookahead}}'
       push:
-        - - clear_scopes: 1
-          - meta_scope: meta.mapping.key.yaml
-          - include: flow-scalar-plain-in-implicit-type-12
-          - match: ''
-            pop: true
-        - - match: (?=\S)
-            pop: true
+        - flow-pair-value-set
+        - flow-pair-key-12
     - match: :(?=\s|$|{{c_flow_indicator}}) # Empty mapping keys allowed
       scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
       push: flow-pair-value
-    # TODO match as meta.mapping.key
+
+  # Variant without clearing the parent scope for pairs in a sequence
+  flow-pair-no-clear:
+    - match: \?
+      scope: meta.mapping.key.yaml punctuation.definition.key.begin.yaml
+      push:
+        - flow-pair-value-set-no-clear
+        - flow-pair-key-no-clear
+    - match: '{{_flow_key_in_lookahead}}'
+      push:
+        - flow-pair-value-set-no-clear
+        - flow-pair-key-no-clear
+    - match: :(?=\s|$|{{c_flow_indicator}})
+      scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
+      push: flow-pair-value-no-clear
+
+  flow-pair-key:
+    - clear_scopes: 1
+    - meta_content_scope: meta.mapping.key.yaml
+    - match: (?=[},\]])  # Empty mapping keys & values are allowed
+      pop: true
+    - match: :(?=\s|$|{{c_flow_indicator}})
+      scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
+      pop: true
     - include: flow-node
+
+  flow-pair-key-12:
+    - clear_scopes: 1
+    - meta_content_scope: meta.mapping.key.yaml
+    - match: (?=[},\]])  # Empty mapping keys & values are allowed
+      pop: true
+    - match: :(?=\s|$|{{c_flow_indicator}})
+      scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
+      pop: true
+    - include: flow-scalar-plain-in-implicit-type-12
+
+  flow-pair-key-no-clear:
+    - meta_content_scope: meta.mapping.key.yaml
+    - match: (?=[},\]])
+      pop: true
+    - match: :(?=\s|$|{{c_flow_indicator}})
+      scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
+      pop: true
+    - include: flow-node
+
+  flow-pair-value-set:
+    - match: ''
+      set: flow-pair-value
 
   flow-pair-value:
     - clear_scopes: 1
+    - meta_content_scope: meta.mapping.value.yaml
+    - include: flow-node
+    - match: (?=[},\]])
+      pop: true
+
+  flow-pair-value-set-no-clear:
+    - match: ''
+      set: flow-pair-value-no-clear
+
+  flow-pair-value-no-clear:
     - meta_content_scope: meta.mapping.value.yaml
     - include: flow-node
     - match: (?=[},\]])
@@ -637,32 +697,18 @@ contexts:
     # Attempt to match plain-out scalars and highlight as "meta.mapping.key",
     # if followed by a colon
     # TODO rewrite with branching
-    - match: |
-        (?x)
-        (?=
-          {{ns_plain_first_plain_out}}
-          (
-              [^\s:]
-            | : \S
-            | \s+ (?![#\s])
-          )*
-          \s*
-          :
-          (\s|$)
-        )
+    - match: '{{_flow_key_out_lookahead}}'
       push:
-        - - meta_scope: meta.mapping.key.yaml
-          - match: ''
-            pop: true
-        - - include: flow-scalar-plain-out-implicit-type-12
-          - match: '{{_flow_scalar_end_plain_out}}'
-            pop: true
-          - match: (?={{ns_plain_first_plain_out}})
-            set:
-              - meta_scope: string.unquoted.plain.out.yaml
-              - meta_include_prototype: false
-              - match: '{{_flow_scalar_end_plain_out}}'
-                pop: true
+        - meta_scope: meta.mapping.key.yaml
+        - include: flow-scalar-plain-out-implicit-type-12
+        - match: '{{_flow_scalar_end_plain_out}}'
+          pop: true
+        - match: (?={{ns_plain_first_plain_out}})
+          set:
+            - meta_scope: meta.mapping.key.yaml string.unquoted.plain.out.yaml
+            - meta_include_prototype: false
+            - match: '{{_flow_scalar_end_plain_out}}'
+              pop: true
     - match: :(?=\s|$)
       scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
 
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -433,6 +433,28 @@ contexts:
         7: punctuation.separator.time.yaml
         8: punctuation.separator.time.yaml
 
+  flow-scalar-plain-in-11:
+    # http://yaml.org/spec/1.2/spec.html#style/flow/plain
+    # ns-plain(n,c) (c=flow-in, c=flow-key)
+    - include: flow-scalar-plain-in-implicit-type-11
+    - include: flow-scalar-plain-in-common
+
+  flow-scalar-plain-in-12:
+    # http://yaml.org/spec/1.2/spec.html#style/flow/plain
+    # ns-plain(n,c) (c=flow-in, c=flow-key)
+    - include: flow-scalar-plain-in-implicit-type-12
+    - include: flow-scalar-plain-in-common
+
+  flow-scalar-plain-in-common:
+    - match: (?={{ns_plain_first_plain_in}})
+      push: flow-scalar-plain-in-content
+
+  flow-scalar-plain-in-content:
+    - meta_scope: string.unquoted.plain.in.yaml
+    - meta_include_prototype: false
+    - match: '{{_flow_scalar_end_plain_in}}'
+      pop: true
+
   flow-scalar-plain-out-implicit-type-11:
     - match: '{{_type_bool_11}}{{_flow_scalar_end_plain_out}}'
       scope: constant.language.boolean.yaml
@@ -524,28 +546,6 @@ contexts:
     - meta_scope: string.unquoted.plain.out.yaml
     - meta_include_prototype: false
     - match: '{{_flow_scalar_end_plain_out}}'
-      pop: true
-
-  flow-scalar-plain-in-11:
-    # http://yaml.org/spec/1.2/spec.html#style/flow/plain
-    # ns-plain(n,c) (c=flow-in, c=flow-key)
-    - include: flow-scalar-plain-in-implicit-type-11
-    - include: flow-scalar-plain-in-common
-
-  flow-scalar-plain-in-12:
-    # http://yaml.org/spec/1.2/spec.html#style/flow/plain
-    # ns-plain(n,c) (c=flow-in, c=flow-key)
-    - include: flow-scalar-plain-in-implicit-type-12
-    - include: flow-scalar-plain-in-common
-
-  flow-scalar-plain-in-common:
-    - match: (?={{ns_plain_first_plain_in}})
-      push: flow-scalar-plain-in-content
-
-  flow-scalar-plain-in-content:
-    - meta_scope: string.unquoted.plain.in.yaml
-    - meta_include_prototype: false
-    - match: '{{_flow_scalar_end_plain_in}}'
       pop: true
 
   flow-sequence:
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -634,8 +634,9 @@ contexts:
           scope: meta.mapping.yaml invalid.illegal.expected-newline.yaml
           pop: true
         - include: block-node
-    # Attempt to match plain-out scalars and highlight as "entity.name.tag",
+    # Attempt to match plain-out scalars and highlight as "meta.mapping.key",
     # if followed by a colon
+    # TODO rewrite with branching
     - match: |
         (?x)
         (?=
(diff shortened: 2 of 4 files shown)


YAML/YAML.sublime-syntax
```diff
@@ -513,51 +515,48 @@ contexts:
...
-        - match: (?=[},\]]) # Empty mapping keys & values are allowed
-          pop: true
-        - include: flow-pair
-        - include: flow-node
-        - match: :(?=\s|$|{{c_flow_indicator}})
-          scope: punctuation.separator.key-value.mapping.yaml
-          set: flow-pair-value
-    # Attempt to match plain-in scalars and highlight as "entity.name.tag",
-    # if followed by a colon
+        - - match: ''
```
deathaxe, line 520: This multi-push might not be needed if we'd move on to sublime-syntax version 2.

Note: Setting `version: 2` doesn't break any of the existing tests.

Otherwiese: 

1. With inheritance in the house it might be a better choice/style for basic syntaxes to avoid anonymous contexts.
2. AFAIK we are asked to avoid anonymous multi-push multi-set for readability reasons.

YAML/YAML.sublime-syntax
```diff
@@ -513,51 +515,48 @@ contexts:
...
+            set: flow-pair-value
+        - - clear_scopes: 1
+          - meta_content_scope: meta.mapping.key.explicit.yaml
+          - match: (?=[},\]])  # Empty mapping keys & values are allowed
+            pop: true
+          - match: :(?=\s|$|{{c_flow_indicator}})
```
deathaxe, line 526: With sublime-syntax version 2 this `:` wouldn't receive the meta_content_scope then.

YAML/YAML.sublime-syntax
```diff
@@ -513,51 +515,48 @@ contexts:
...
-            - meta_include_prototype: false
-            - match: '{{_flow_scalar_end_plain_in}}'
-              pop: true
-    - include: flow-node
+        - - clear_scopes: 1
+          - meta_scope: meta.mapping.key.yaml
+          - include: flow-scalar
+          - match: ''
+            pop: true
+        - - match: (?=\S)
```
deathaxe, line 549: What is this `match: (?=\S)` pattern for? As the lookahead which pushes this context begins with `{{ns_plain_first_plain_in}}` - a non-whitespace pattern - this anonymous context popps immediatelly.

I don't see any tests breaking, when turning that into a simple single-push statement.

YAML/YAML.sublime-syntax
```diff
@@ -513,51 +515,48 @@ contexts:
...
-          scope: punctuation.separator.key-value.mapping.yaml
-          set: flow-pair-value
-    # Attempt to match plain-in scalars and highlight as "entity.name.tag",
-    # if followed by a colon
+        - - match: ''
+            set: flow-pair-value
+        - - clear_scopes: 1
+          - meta_content_scope: meta.mapping.key.explicit.yaml
+          - match: (?=[},\]])  # Empty mapping keys & values are allowed
+            pop: true
```
deathaxe, line 525: I'd find it safe to replace this and the next `pop: true` by `set: flow-pair-vlaue` once version 2 was set. As a result the multi-push could be simplified to a single-push.

YAML/YAML.sublime-syntax
```diff
@@ -490,7 +490,9 @@ contexts:
     - match: \[
       scope: punctuation.definition.sequence.begin.yaml
       push:
-        - meta_scope: meta.flow-sequence.yaml
+        # Pushes an extra 'meta' on the scope stack that flow-pair can then clear,
+        # so we don't have to duplicate the context
+        - meta_scope: meta.sequence.flow.yaml meta.workaround.yaml
```
deathaxe, line 495: `meta.workaround` is assigned to `[` and `]` and all other non-pair tokens, which might not be ideal.

deathaxe (commented):
meta.mapping and meta.sequence ... nice.

YAML/YAML.sublime-syntax
FichteFoll: That's the downside of this workaround. I suppose `meta_content_scope` would excempt the brackets at least, but won't do anything for non-pair tokens, which is why I chose such a scope name.

YAML/YAML.sublime-syntax
FichteFoll: Good point, I forgot this was already after a lookahead.

FichteFoll:
Thanks for the suggestions. I think I addressed all your comments, please re-review.

YAML/YAML.sublime-syntax
```diff
@@ -490,74 +519,94 @@ contexts:
...
+      push: flow-pair-value-no-clear
+
+  flow-pair-key:
+    - clear_scopes: 1
+    - meta_content_scope: meta.mapping.key.yaml
+    - match: (?=[},\]])  # Empty mapping keys & values are allowed
+      pop: true
+    - match: :(?=\s|$|{{c_flow_indicator}})
+      scope: meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
+      pop: true
```
deathaxe, line 581: If we don't intend to re-use `flow-pair-key` anywhere else, the `pop: true` could be replaced by `set: flow-pair-value` and the multi-push could be turned into a simple `push: flow-pair-key` in `flow-pair`. The intermediate `flow-pair-value-set` context was obsolete then.

On the other hand, the current solution clearly spearates key and value related contexts with maximum level of reusability, which might be to prefer in such a basic syntax. The proposed change wouldn't result any performance improvements, so I am very fine with the current solution.

deathaxe (approved):
LGTM

jwortmann:
> If this goes through, Mariana.sublime-color-scheme needs updating. Maybe others?

:point_right: https://github.com/sublimehq/sublime_text/issues/3590

FichteFoll:
> The `:` is scoped using `meta.mapping` while key and value receive `meta.property` only.

The key is also scoped as a key for me, but the property isn't. I'll add that to the look-ahead for correctness, but values currently aren't scoped in block collections at all, because matching is too inaccurate for those. We can add that later, imo.

FichteFoll:
Turns out this isn't as simple as I was hoping to and it does make the syntax unnecessarily complex, imo. What should happen is rewriting the key and value pair detection using branching, but I won't get to that in the near future, so this quick(er) fix should do for now.

wbond:
Beyond the default color schemes, this is going to need a forward fill for color schemes in general. Moving from `entity.name.tag` to `string.unquoted` will leave the syntax pretty bare in most color schemes.

deathaxe:
I wonder if we can find some kind of scope for such mapping keys in general. We need it in python mappings, json and even CSS might be a candidate to apply the `meta.mapping` to properties.

Basically, even though it complies with JSON/YAML specs, I find it a bit weird to scope keys `string`. A string feels more like a value or argument. In case of YAML, which is also used in PackageDev it makes navigation via goto definition harder/impossible as `string` is always excluded from such functions.

wbond:
"Symbols" in Ruby use `constant.other`. What about something like `constant.other.key` for unquoted key values?

deathaxe:
It definitly might be a good scope in case a key is used to access an item, like `dict[key]`, but I am not quite sure how well it plays in CSS for instance. Nearly all key like tokens (html-tag names, attribute names, property names, ...) use `entity.other` or `entity.name` there, at the moment.

If we think of `meta.mapping` to be a form of structured data like XML and JSON, YAML as an alternative representation of such, something like `entity.name.tag` makes sense. At least I understand why it was choosen in the past.

With the intention of #1861 in mind, too, I wonder if such a key should stay within the `entity.` scope?

Not sure if it needs to apply to unquoted keys only. In Erlang for instance all identifiers may be quoted, without them denoting strings. The same applies to SCL/ST (structured text) used in automation systems. 

One point of this PR is to avoid too much scopes being stacked. Hence I guess `entity` was removed in favor of `meta.mapping.key`, but finally scoping keys as string is the same as scoping them boolean which has been fixed as well. TCL also chooses to scope strings semantically as nearly everything is one, etc.

wbond:
So perhaps something like `entity.other.key` and skipping the `string.unquoted`?

I could potentially see `using `entity.other.key` for string keys in JSON, and performing a forward fill of a color scheme rule to populate it from `string`. That would preserve existing colors, but also give color schemes the ability to provide special colors for all mapping keys, wether they be unquoted, quoted, etc. The only thing we probably wouldn't want to scope would be something like a number as a key, right?

deathaxe:
I have a good feeling about `entity.other.key`. I would remove `string` as well, then, yes. The `meta.mapping.key` is then used to denote expression like keys as the property thing discussed here.

The point about `meta.mapping string` in json is, it already "broke" or removed highlighting the same way as we discuss it here for older color schemes due to removing the scope used for keys before. I haven't investigated but maybe `entity.other` would even improve or restore the original behavior for them even without a forward fill?

jwortmann:
@wbond:
> So perhaps something like entity.other.key and skipping the string.unquoted?

I'm not sure about `entity.other.key`. Not that I'm per se against it, and I can see that it's kind of unfortunate to have `string` for both keys and values, which is why I asked in the linked issue above to distinguish them via the `meta.mapping.key` scope in the default color schemes. But the only `entity.other.*` scopes in use seem to be `entity.other.attribute-name` and `entity.other.inherited-class`, and `entity.other` is not beyond the recommended minimal scope coverage for color schemes. So how exactly would such a forward fill from `string` exactly work? Let's assume `entity.other.key` is not defined, then take the color for `string` instead => that would mean that color schemes which have adopted and assigned a different color for `meta.mapping.key string` in the meanwhile (it is in use in the lasted ST3 release at least for JSON and quoted keys in Python) would "break". Or alternatively, if the color from `meta.mapping.key string` is taken instead => this would not be effective as most color schemes probably don't have a rule for it. It might work if for both `entity.other.key` and `meta.mapping.key string` is tested and the correct color is taken accordingly, but that makes things quite complicated.

And besides JSON, which is a strict subset of YAML, such a change would also require changing it again in JavaScript, Python, Lua and possibly more if we aim for consistent scopes between languages.

Furthermore, as you already mentioned, it's unclear to me whether `entity.other.key` should only apply to string keys or also to other kind of keys. For example, Python or Julia allow almost arbitrary things as keys, e.g. numbers or custom types. Keys that are numbers are currently scoped as `constant.numeric` in Python, should this be changed to `entity.other.key` as well then? If not, what makes string keys special from other types of keys, that they "deserve" `entity.other.key`?

@deathaxe:
> I am not quite sure how well it plays in CSS for instance. Nearly all key like tokens (html-tag names, attribute names, property names, ...) use entity.other or entity.name there, at the moment

I think these are only CSS selectors and I would say they are somewhat different from key-value mappings. The properties actually use `support.type.property-name`, and this should be fine, because only predefined properties with known names are allowed (afaik?). So in my opinion it's not necessary to change the CSS scopes, especially since they are quite unique and there are probably a lot of color schemes which specifically target them.

> The point about meta.mapping string in json is, it already "broke" or removed highlighting the same way as we discuss it here for older color schemes due to removing the scope used for keys before

JSON used `string` for keys even before, I believe the only difference is a rule for `"meta.structure.dictionary.json string.quoted.double.json"` in Monokai, which doesn't apply anymore. Therefore I can only reference https://github.com/sublimehq/sublime_text/issues/3590 again as a pretty easy fix. As explained above, I guess `entity.other` (without a forward fill) would not apply any color at all in most color schemes, so it would fall back to the default foreground color. Not sure whether that's an improvement or not.

deathaxe:
The main issue I have with `string` is it being treated special by many functions. It's a major reason for goto definition not to work properly in PackageDev's sublime-syntax syntax. Each key (context name) is a string and thus excluded by ST's symbol.py

Of course, we can define a special syntax for that which doesn't use strings for context names. But that's basically YAML, so it makes sense to use the same base scopes. There are other use cases where users want goto defintion functionality for enumeration keys in C/C++. How is that so much different to structured data?

Otherwise the `meta.mapping [string|int|constant| ...]` approach is pretty much ok.

FichteFoll:
> but finally scoping keys as string is the same as scoping them boolean which has been fixed as well.

No, boolean keys `true` and `false` are still scoped as booleans. What was removed were the matches for boolean aliases in keys, i.e. `y: 12`.

---

In general, I believe we have the following ways forward:

1. Scope every key as `mapping.key` and then scope the key's token as it would be when written as a value, i.e. use `string` for a string and use `constant.numeric` for numbers. Python example: `{"string": "hi", 2: 3}`. Syntaxes with implictly unquoted string keys, such as JavaScript, would then use `string.unquoted` so they wouldn't be confused with an identifier and seeing how they are equivalent to strings. Ruby's symbols would still get `constant.other.symbol`.

2. Scope every key as `mapping.key` and add `entity.other.key` for scalar keys, such as strings (quoted or unquoted) or integers. 

3. Do both.

The following evaluation is from the perspective of "what if we started from zero". All solutions would require a forward-fill for compatibility with old color schemes anyway.

**Advantages**:

1. Low-level approach that allows fine-grained control for color schemes or plugins.

2. It becomes clear what token represents a key. Color schemes can simply target the key scope and be done with it.

3. Both of the above.

**Disadvantages**:

1. `meta.mapping.key string` also matches strings in complex keys, e.g. `{["list", "key"]: "value"}` (Python). Context management of clearing the non-key `meta.mapping` scope is sometimes not so straight-forward. `string` is special-cased for some ST features like goto definition.

2. Can't re-use expression-style contexts for matching keys, as `entity.other.key` needs to be applied on top. What is `entity.other` supposed to mean in the long run?

3. Both of the above. Also more scope stacking.

I should mention that if we were to go for 2, I would question the usefulness of `meta.mapping.{key,value}`, as the primary goal of that has been to distinguish between key and value expressions. We may be able to remove this scoping, then.

jwortmann:
> ... in complex keys, e.g. `{["list", "key"]: "value"}` (Python)

This expression is illegal in Python, only immutable types can be keys, but lists cannot.

Otherwise I agree, if we use only `entity.other.key` instead of the actual type for keys, then `meta.mapping.{key|value}` has no more use and could be removed.

deathaxe:
Maybe ST can just ignore its `string` exclusion in `meta.mapping.key` to enable goto definition and everything is fine.

michaelblyons:
> Maybe ST can just ignore its `string` exclusion in `meta.mapping.key` to enable goto definition and everything is fine.

I like this for an additional reason: YAML keys can be explicitly quoted strings, too:

``` yaml
foo:
  - bar
  - baz
'bim': true
```

Unfortunately, they're not scoped as keys at present.

> This expression is illegal in Python, only immutable types can be keys, but lists cannot.

Because I'm silly and pedantic sometimes: sure, but that list as a tuple _can_ be a key. `{("list", "key"): "value"}` If you want to be even sillier, you can use namedtuples and pretend your key is a dict.

jwortmann:
> Because I'm silly and pedantic sometimes: sure, but that list as a tuple can be a key. {("list", "key"): "value"} If you want to be even sillier, you can use namedtuples and pretend your key is a dict.

True, it indeed works with tuples. But I guess examples like this aren't used very often in practice, so I wonder how relevant it is for this discussion. And actually, I would see this more as an *advantage* for `meta.mapping.key` than a disadvantage, because the meta scope can span the whole tuple (since the key does consist of the whole tuple here), while the parentheses and comma can get `punctuation` and the strings get `string.quoted.double` as expected.
In case of using `entity.other.key` instead, would you apply this scope to all of the characters, including the punctuation and spaces? That might make it a bit more visible that the whole tuple-structure is the key, but at same time it would probably look quite unusual if all these characters have the same highlighting color then.

But I'm not sure which way is simpler or more robust from the implementation side and how exactly it is implemented right now in Python.

FichteFoll:
> In case of using `entity.other.key` instead, would you apply this scope to all of the characters, including the punctuation and spaces?

In my proposal, I would only apply it to scalar elements, i.e. strings and numbers and without punctuation, since this would be more of a semantic scope name.

jwortmann:
> In my proposal, I would only apply it to scalar elements, i.e. strings and numbers and without punctuation, since this would be more of a semantic scope name.

Ok, but is it really useful to drop the punctuation, i.e. the string quotes from `entity.other.key` even for scalar elements? While it wouldn't make a semantic difference in JavaScript (where unquoted keys implicitly get converted to a string), in Python the keys `"42"` and `42` aren't the same. For item access like `mydict["42"]` you also can't just omit the quotes in Python, if you want to select the quoted key. So I would expect the `entity.other.key` scope to match the same behaviour. And in case of Goto functionality or something like a symbol list for mapping keys, which might use these scopes, I wouldn't expect two unquoted identical entries `42` if there exist both a quoted and an unqouted key, which are actually different. Maybe this decision should be syntax dependent, based on whether the language does an implicit string conversion or not?
Assuming no `entity.other.key` for the punctuation, and if there is also no more `string` scope for the key anymore, should the quotes still get `punctuation.definition.string` then? Most color schemes, such as Monokai, probably don't target `punctuation.definition` alone, so the quotes wouldn't get any color at all.

Anyways, I took some time and made a few experiments in Monokai and Mariana for the resulting highlighting colors, assuming a rule for `meta.mapping.key string` or `entity.other.key`, respectively, will be added to the color schemes. That should probably be equivalent to the suggested "forward fill", even though I still don't know how exactly or with which scopes this forward fill should work (or is it perhaps meant to apply only to YAML?). Note that Mariana already has special rules for `entity.name.tag.yaml` and for `source.yaml string.unquoted`. So for Mariana I used the color from `entity.name.tag.yaml` for these extra rules, while for Monokai I used orange as an example, because the color for `entity.name.tag` is the same as for `keyword` in that scheme, which would look a bit confusing in Python code.

**Current highlighting** (as of latest ST3):

current

**Alternative 1** (this PR): `meta.mapping.key` and actual type, with a corresponding color scheme rule for `meta.mapping.key string`:

meta_mapping_key_string

**Alternative 2**: `entity.other.key`, without `string`, but only for scalar keys and excluding the punctuation as suggested. In Mariana the excluded quotes make no difference (assuming they still get `punctuation.definition.string` or a similar scope), because Mariana has a rule for `punctuation.definition` with the same color as for `entity.name.tag.yaml`. But for Monokai it looks kind of "incomplete" to me without any highlighting for the quotes:

entity_other_key

I would still prefer the `meta.mapping.key` solution, which allows to color e.g. number keys differently from string keys.

michaelblyons:
I currently support `meta.mapping.key`, although if I were to make a color scheme rule, I'd exclude the tuple coloring (`meta.mapping.key - meta.mapping.key meta.group` or something).

With DeathAxe's idea to drop the `string` exclusion for go-to refs, I think that covers the things I'm interested in.

keith-hall (dismissed):

deathaxe:
Yeah, color schemes need to be tweaked anyway as JSON looks pretty plain as well.

Maybe someone should have a look into those merge conflicts. I am fine with this PR as well.

YAML/YAML.sublime-syntax
```diff
@@ -599,46 +648,37 @@ contexts:
...
-          (
-              [^\s:]
-            | : \S
-            | \s+ (?![#\s])
-          )*
-          \s*
-          :
-          (\s|$)
-        )
+    # TODO rewrite with branching
```
jrappen, line 667: Were you going to leave this TODO here, or fix it in this PR?

YAML/tests/syntax_test_flow.yaml
```diff
@@ -112,47 +113,53 @@ continuation"
...
 #  ^^^^^ storage
-#        ^^^ string entity.name.tag
-#            ^ punctuation.separator.key-value.mapping -string
-#              ^ string.unquoted.plain.in
+#        ^^^ string.unquoted.plain.in.yaml
+#            ^ meta.mapping.yaml punctuation.separator.key-value.mapping.yaml
+#              ^^^^ meta.mapping.value.yaml string.unquoted.plain.in.yaml
 }
 
+# TODO catch `!!set` and set meta.set scope
```
jrappen, line 159: Were you going to leave this TODO here, or fix it in another PR?

jrappen:
@FichteFoll Please fix the conflicts.

jrappen:
⚠️ Since you switched to `version: 2`, all `pop: true` should be changed to `pop: 1`.

YAML/tests/syntax_test_flow.yaml
FichteFoll: Definitely not in this PR.

YAML/YAML.sublime-syntax
FichteFoll: That will have to wait for a different PR. While it is certainly something that *could* be done in a similar scope, it's not something that I will find the time for in the forseeable future.

FichteFoll:
I could do that, but it will add a lot of noise to the PR that is otherwise self-sufficient (even with the version change). There is a merge conflict that I will resolve, however.

keith-hall (approved):

deathaxe:
Just was skimming through the different types of key names, trying to create a color scheme patch for mariana to tint them equally, when I found quoted key names not receiving `meta.mapping.key`.

```yaml
"string": "string"
'string': 'string'
```

grafik

Do we want to support `meta.mapping` scoping here as well?

PHP/tests/syntax_test_php.php
```diff
@@ -1935,12 +1935,12 @@ function generate2()
...
-//^ string.unquoted.plain.out entity.name.tag
+//^ meta.mapping.key string
 // ^       punctuation.separator.key-value.mapping
-//   ^^^ string.unquoted.plain.out
+//   ^^^ string
 three: "$four"
 //^^^^^^^^^^^^ meta.embedded.yaml source.yaml
-//^^^ string.unquoted.plain.out entity.name.tag
+//^^^ meta.mapping.key string
 //   ^       punctuation.separator.key-value.mapping
```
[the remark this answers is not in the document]
jrappen: Shouldn't it rather be `punctuation.separator.mapping.key-value`? Also compare the related `punctuation.separator.mapping.pair`.

FichteFoll:
Definitely, although I suspect we'll need branching for this.

PHP/tests/syntax_test_php.php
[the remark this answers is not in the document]
jwortmann: Note that it *was* `punctuation.separator.mapping.key-value` before, and it got changed to `punctuation.separator.key-value.mapping` in https://github.com/sublimehq/Packages/commit/60e7c391deb32d1d39b5add199f827c7d3bb8050 for a reason. Could it be that PackageDev was just not updated with the newer scope? It would be great to have a consistent scope for key-value separators between syntaxes, and the only ones which still seem to be left are JSON and Python. Putting the more specific `mapping` at the third level would require color schemes to target at least two scopes for key-value separators, unless you want to introduce this `mapping` at the third level also for syntaxes like CSS and HTML, which don't use the `meta.mapping` scope.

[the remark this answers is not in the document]
jrappen: The use of one order or the other is split across syntaxes, might as well flip a coin at this point.

deathaxe:
Sure. A look ahead may even be sufficient as it is probably quite unusual to have quoted multi-line keys.

michaelblyons:
Very relevant, in terms of Go To references: https://github.com/sublimehq/sublime_text/issues/2546

deathaxe:
Situation won't change with this PR assuming id keeps `string` scopes being applied to keys. But as JSON also scopes keys `string` a solution including `meta.mapping.key` in _Default/symbol.py_ might be needed anyway.

PHP/tests/syntax_test_php.php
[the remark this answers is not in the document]
deathaxe: We should keep it as is with regards to #3228.

deathaxe:
An attempt to fix quoted `meta.mapping.key` can be found at https://github.com/FichteForks/Packages/pull/10

FichteFoll:
Applied a similar solution to block keys as well while fixing some invalid explicit block key syntax.

https://github.com/yaml/yaml-editor with the c-libfyaml parser is very useful for verifying YAML syntax, by the way.

YAML/YAML.sublime-syntax
FichteFoll: Handled via https://github.com/FichteForks/Packages/pull/10.

jrappen (dismissed):

keith-hall (dismissed):

YAML/YAML.sublime-syntax
```diff
@@ -620,47 +704,36 @@ contexts:
...
-        - match: ^ *(:)
-          captures:
-            1: punctuation.separator.key-value.mapping.yaml
-          pop: true
-        - match: ':'
-          scope: invalid.illegal.expected-newline.yaml
-          pop: true
-        - include: block-node
-    # Attempt to match plain-out scalars and highlight as "entity.name.tag",
+        - block-key-explicit
```
deathaxe, line 710: Maybe use a direct `push: block-key-explicit` as a single context is pushed only?

YAML/YAML.sublime-syntax
```diff
@@ -620,47 +704,36 @@ contexts:
...
-        - include: flow-scalar-plain-out-implicit-type-12
-        - match: '{{_flow_scalar_end_plain_out}}'
-          pop: true
-        - match: (?={{ns_plain_first_plain_out}})
-          set:
-            - meta_scope: string.unquoted.plain.out.yaml entity.name.tag.yaml
-            - meta_include_prototype: false
-            - match: '{{_flow_scalar_end_plain_out}}'
-              pop: true
+        - block-key-implicit
```
deathaxe, line 715: Same here.

deathaxe (commented):
LGTM! Maybe two tiny code style improvements...

deathaxe:
> the use or non-use of *.yaml-endings for scopes in tests is really random

That's a general issue in most syntaxes. Tests started quite simple when introduced and matured to capture more edge cases and details. I guess PRs with improved test cases are wellcome in general.

jrappen:
@FichteFoll Maybe you could update the opening post for the commit message of the squash merge? It would probably make sense to mark this as addressing 3228.

PHP/tests/syntax_test_php.php
```diff
@@ -1935,12 +1935,12 @@ function generate2()
 //       ^^^ entity.name.tag.heredoc
 one: two
 //^^^^^^ meta.embedded.yaml source.yaml
-//^ string.unquoted.plain.out entity.name.tag
+//^ meta.mapping.key string
```
keith-hall, line 1938: Was there a reason why the sub-scope(s) of `string` are no longer asserted here? 
```suggestion
//^ meta.mapping.key string.unquoted
```

keith-hall (approved):

PHP/tests/syntax_test_php.php
FichteFoll: I made the test more general so that it wouldn't break as easily when YAML is updated again. I don't think we'd remove the `unquoted` part, but the purpose of the test is primarily to check whether a freuqent use case of embedding works, not all the details of the YAML syntax. `meta.mapping.key string` seems to me to be the most general assertion that would apply here, especially seeing that it's likely to be used by color schemes as well.

keith-hall: Makes sense, thanks for the explanation.

jrappen:
🤷🏼‍♂️ Not sure why the commit message of the merged commit lost all the details that were provided above.

FichteFoll:
Same here :unamused:

PR #2566, opened 2020-10-25
[YAML] Don't match controversial plain bools in keys

### LINKED ISSUE #1602: YAML highlighting issue
Currently, this YAML file highlights incorrectly:

    x: 10
    y: 20
    z: 30

Namely, the 'y' appears in red italics because the syntax highlighter treats as short for the boolean value 'Yes'. But I think that such highlighting should only occur if 'y' appears as a value.

FichteFoll:
YAML 1.1 made the prominent mistake of accepting many aliases for boolean values that many parsers adopted. For 1.2, these aliases have been removed, but for accuracy with the older version and especially
considering how older parsers are still seeing use, we cannot simply remove matching them.

However, the short names "y" and "on" are fairly common in keys and having those highlighted as a boolean is annoying, so we only match for the 1.2 scalars in keys.

I believe this solution to be a satisfactory compromise for #1602, where this issue has been discussed at length.

Closes #1602.

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -115,12 +115,14 @@ variables:
   _type_null: (?:null|Null|NULL|~)
   
   # http://yaml.org/type/bool.html
-  _type_bool: |-
+  _type_bool_11: |-
     (?x:
        y|Y|yes|Yes|YES|n|N|no|No|NO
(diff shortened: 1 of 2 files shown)


deathaxe (approved):

wbond:
Seems like a reasonable resolution, thanks @FichteFoll!

PR #2568, opened 2020-10-28
[Git Formats] Add support for subject length

niksy:
[As documented in blog post](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html), this adds support for two new ways to style subject, longer than 50/less than 72 characters is considered warning, everything over 72 is considered error.

### DIFF
--- a/Git Formats/Git Commit.sublime-syntax
+++ b/Git Formats/Git Commit.sublime-syntax
@@ -57,6 +57,7 @@ contexts:
     - match: $\n?
       set: commit-separator
     - include: Git Common.sublime-syntax#references
+    - include: commit-subject-length
 
   commit-separator:
     # empty line between subject and message
@@ -81,6 +82,12 @@ contexts:
       captures:
         1: keyword.other.signed-off-by.git.commit
         2: punctuation.separator.mapping.pair.git.commit
+
+  commit-subject-length:
+    - match: '^((.{1,50}\s*)|(.{51,72}\s*)|(.+\s*))$'
+      captures:
+        3: invalid.warning.line-too-long.git.commit
+        4: invalid.error.line-too-long.git.commit
 
 ##[ COMMENTS ]#########################################################
 


jwortmann:
A few comments about this:

* Even though the 50/72 character limit is a good convention, longer commit messages are not really *invalid*. So I don't think it is the best approach to consider them as an error via syntax highlighting. Personally I use the following `Git Commit.sublime-settings` file, to add some more subtle indications about the character limit:
```json
{
    "rulers": [72, 50]
}
```

* The way this is implemented in your PR right now completely masks the `- include: Git Common.sublime-syntax#references` statement (unless they occur at the start of the line), which results in the loss of highlighting for issue numbers, usernames, etc. in commit message headings. This is why the syntax tests fail. Also it doesn't work as intended if the start of the line is one of the things listed above.

* *If* this approach with syntax highlighting for long commit messages should really be the desired way (personally I wouldn't like it much):
  * The scopes `invalid.warning` and `invalid.error` which used here are unlikely to be adressed by color schemes. Alternatives are `invalid.illegal`, `invalid.deprecated` or maybe `markup.error`, `markup.warning` (but the latter ones would probably also not be adressed by schemes often).
  * Maybe you would want to apply this kind of highlighting only to the part of the line which exceeds the character limit.

niksy:
Yeah, I see your point. I already use `rulers` setting but this is only addition to that.

This is only suggestion, maybe someone else can also add their opinion on this?

deathaxe:
Fully agree with @jwortmann here.

Limiting line length is of linting category but no syntax error thing. It was discussed in the past with the result not to add invalid highlighting for too long lines as it is of user's preference to limit it or not.

Some color schemes use quite obvious colors for `invalid` scopes which easily becomes a pain when reading existing commit messages with too long lines.

Hence the `ruler` approach and/or `word_wrap` are the best default, IMHO.

niksy:
OK, closing PR then! Thank you!

PR #2570, opened 2020-11-03
[C#] Add string scopes to characters

michaelblyons:
https://discord.com/channels/280102180189634562/280102180189634562/773291314984910849


keith-hall (approved):
Does it make sense to also scope it with `meta.string` or not (because it isn't a string... Its just that normally `string` scopes are also accompanied by `meta.string`.)?

wbond:
I think the `meta.string` would be appropriate.

wbond:
Thanks @michaelblyons!

PR #2572, opened 2020-11-05
[RFC][PHP] Split out extensions into dedicated files

jfcherng:
This PR basically split out those 70+ extensions into their own dedicated files. 

---

**Pros**:

- Make things more structured and manageable.
- Make things more customizable?

**Cons**:

- I noticed that, on my machine, the main compiled cache file size increase from `9480 KB` to `10101 KB`.

---

I am not sure whether this way is preferred.


deathaxe:
I agree with an approach to split a syntax definition into a `plain` part which contains the very basic patterns to highlight it (excluding embedded syntaxes and huge lists of builtins). 

That's the approach of 

1. _HTML.sublime-syntax_ vs. _HTML-Plain.sublime-syntax_ 
   where `-Plain` contains the least required patterns to highight HTML, while the normal HTML.sublime-syntax adds patterns for all the different kinds of tags and embedds JS/CSS.

2. _Bash.sublime-syntax_ making use of a `commands-builtin-shell-bash.sublime-syntax` which contains all builtins

But we need to keep in mind, even though a `hidden: true` syntax isn't displayed in command palette or menues, it still "pollutes" the overall scope namespace. Any kind of API or syntax using functionality sees them and needs to handle/ignore them, even though they only have special purpose to complete their parent core syntax.

Hence we should be careful with adding too many syntaxes especially those which don't even work standalone. 

With this PR merged `sublime.list_syntaxes()` outputs nearly as much `php.ext` syntaxes as all the other default syntaxes together.

_That doesn't feel right, even though I fully understand the motivation and appriciate the efforts to increase maintainability._

### My suggestion

How about renaming `PHP Source` to `PHP-Plain.sublime-syntax` with only the basic patterns and some "entry-point" contexts to plug-in the builtins then.

The `PHP Source` could then be an inherited syntax from `PHP-Plain` adding each extension as its own named context, which is then included by the extended `entry-point` contexts. The file would mainly consist of those extension patterns organized in contexts one after another. Many lines, but still maintainable via "Goto Symbol" features using PackageDev.

Each extension could be separated by `###[ Extension Name ]#####...#` comment of 80chars of length. (See HTML for instance).

We'd end up in 2 syntax files each of them being usable standalone, just with different feature levels. Some purists could even decide to use the plain syntax without language extensions then.

jfcherng:
@deathaxe Thanks for all these inputs. I will check it out when I am available. :+1:

wbond:
The increased cache size is likely due to new context names from the extensions. With our new lazy loading, we have to explicitly track the names on contexts, even after the syntax is compiled, so that we can lazily reference contexts and load them as needed for embeds.

I think the idea that @deathaxe proposed sounds good where we have a plain syntax and one with all of the extensions added in. If the inheritance is done right, I _think_ it would likely also bring the compiled size back down.

jfcherng:
Close as `extend` is a better way to do this.

Issue #2468, opened 2020-08-26
[TypeScript] Fails to highlight arrow function with types


rchl:
- Sublime Version: 4084

There are some failures with highlighting this code.

```ts
const merge = (target: JSONSchema, sourceRoot: JSONSchema, sourceURI: string, refSegment: string | undefined): void => {
	const path = refSegment ? decodeURIComponent(refSegment) : undefined;
	const section = findSection(sourceRoot, path);
	if (section) {
		for (const key in section) {
			if (section.hasOwnProperty(key) && !target.hasOwnProperty(key)) {
				(<any>target)[key] = section[key];
			}
		}
	} else {
		resolveErrors.push(localize('json.schema.invalidref', '$ref \'{0}\' in \'{1}\' can not be resolved.', path, sourceURI));
	}
};
```

![Screenshot 2020-08-26 at 23 07 50](https://user-images.githubusercontent.com/153197/91357084-f82baa80-e7f0-11ea-9aa6-a511c8194c46.png)

This is extracted from this file and line: https://github.com/microsoft/vscode-json-languageservice/blob/76f4b43d475f67fd58c6bd2eaea1b0cd01826f84/src/services/jsonSchemaService.ts#L423

michaelblyons:
Works if you remove `: void`. Missing type hint after function assignment?

I'm not too surprised no one had encountered this before. I imagine most people would write

``` ts
function merge (...) : void {
   ...
}
```

rchl:
Pink @Thom1729 :)

Thom1729:
FYI, I'm looking at this, and I hope to have a proper solution soon. This is probably the trickiest thing to get right with TypeScript because of situations like this:

```ts
x ? (y) : z;
x ? (y) : T => r : z;
```

rchl:
For good measure here is a slightly different case but probably the same issue: https://github.com/Anidetrix/rollup-plugin-styles/blob/7532971ed8e0a62206c970f336efaf1bcf5c3315/src/loaders/postcss/url/index.ts#L59-L192

The two ending braces of that block are rendered in red.

deathaxe:
Here's another failing example:

```tsx
function bloadware() {

    return (
        <ul type="cards">
        { 
            list.map((item: Faq)) => (

                <li
                    class={item.class}
                >
                    {item.text}
                </li>
            ) 
        };
        </ul>
    );
};
```

Thom1729:
That example has a typo:

```
[...]

Issue #2267, opened 2020-02-06
[JavaScript] Rationalize function name detection.


Thom1729:
In some cases, the current JavaScript syntax will highlight a variable name, mapping key, or property name as a function name (`entity.name.function`):

```js
const f = function() {};

const obj = {
    f: function() {}
};

obj.f = function() {};
```

Originally, this seems to have been a convenience feature for display purposes. The implementation uses lookaheads and is not reliable, leading to an uneven user experience.

However, there is a real JavaScript feature by which anonymous functions will be named using the surrounding context in some cases. We could detect most of these cases using branching and reimplement the special name highlighting in terms of the language feature. This would seem to be the most correct way of handling what is otherwise a vaguely-motivated special case.

michaelblyons:
Closed with https://github.com/sublimehq/Packages/pull/2417?

Thom1729:
That one improves arrow function detection, but it still uses awkward lookaheads to highlight the “names” of anonymous functions. I've been looking into ways to rewrite that to avoid the messy lookaheads, but haven't hit upon the right implementation yet.

PR #2574, opened 2020-11-06
[TypeScript] Implement arrow function return types.

Thom1729:
For #2468.

In regular JavaScript, when we see an identifier or open paren, we branch and first try to parse them as an identifier or open paren. Then, if the next token is `=>`, we fail the branch and parse it as an arrow function.

In TypeScript, we have another case. If the possible identifier or parenthesized expression is followed by a colon, then we have an arrow function if and only if the colon is followed by a syntactically valid type followed by `=>`. In this implementation we branch again and try to parse a type followed by an arrow. If we succeed, then we fail the original `arrow-function` branch. Otherwise, we fail the inner branch and pop out of the `arrow-function` branch. This is admittedly convoluted, but in principle it should do the right thing.

This PR should mostly work, but there are a few known issues:

- ~~Only arrow functions with parenthesized arguments lists should be able to have type arguments, and now arrow function with a single bare argument, but this PR does not yet make that distinction because both cases are handled the same in the base JavaScript syntax. They should probably be split, though this will involve some code duplication due to https://github.com/sublimehq/sublime_text/issues/3494.~~ Fixed
- In some cases, invalid type expressions will be accepted, e.g. `T||U`. The type expression contexts have some special handling to make this work for function type arguments, but because of https://github.com/sublimehq/sublime_text/issues/3494 this can't be re-used.
- There are a lot of magic multi-pop expressions, which bothers me personally. This may be fixable.
- If an arrow function has a return type, then the type will actually be parsed twice, which is not ideal. I'm not sure there's a good way around this.
- In statements like `const closeDialog = (): any => setDialogOpen(false);`, `closeDialog` will not be highlighted as a function name. This can probably be fixed by tinkering with the lookahead, though it won't be perfect. ~~Fixed, although it won't be exact in corner cases.~~ Never mind; the fix broke other things. I will probably punt this to #2267.

Comments welcome. I don't have a lot of TypeScript code laying around, so I'd appreciate anyone else giving it a whirl.

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -50,6 +50,7 @@ contexts:
       push:
         - ts-detect-arrow
         - ts-type-expression-end
+        - ts-type-expression-end-no-line-terminator
         - ts-type-expression-begin
 
   ts-detect-arrow:
@@ -673,6 +674,7 @@ contexts:
     - include: ts-type-basic
 
     - match: (?=\()
+      pop: true
       branch_point: ts-function-type
       branch:
         - ts-type-parenthesized
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -48,11 +48,11 @@ contexts:
   ts-detect-arrow-function-return-type:
     - match: ':'
       push:
-        - ts-detect-arrow-foobar
+        - ts-detect-arrow
         - ts-type-expression-end
         - ts-type-expression-begin
 
-  ts-detect-arrow-foobar:
+  ts-detect-arrow:
     - match: (?==>)
       fail: arrow-function
     - match: (?=\S)
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -25,6 +25,51 @@ variables:
     ))
 
 contexts:
+  branch-possible-arrow-function:
+    - meta_include_prototype: false
+    - match: ''
+      push:
+        - ts-detect-arrow-function
+        - branch-not-arrow-function
+
+  ts-detect-arrow-function:
+    - match: (?=:)
+      branch_point: ts-arrow-function-return-type
+      branch:
+        - ts-detect-arrow-function-return-type
+        - ts-immediately-pop-3
+    - include: detect-arrow
+
+  ts-immediately-pop-3:
+    - meta_include_prototype: false
+    - match: ''
+      pop: 3
+
+  ts-detect-arrow-function-return-type:
+    - match: ':'
+      push:
+        - ts-detect-arrow-foobar
+        - ts-type-expression-end
+        - ts-type-expression-begin
+
+  ts-detect-arrow-foobar:
+    - match: (?==>)
+      fail: arrow-function
+    - match: (?=\S)
+      fail: ts-arrow-function-return-type
+
+  arrow-function-declaration:
+    - meta_include_prototype: false
+    - match: ''
+      set:
+        - function-meta
+        - arrow-function-expect-body
+        - function-declaration-meta
+        - arrow-function-expect-arrow
+        - ts-type-annotation
+        - arrow-function-expect-parameters
+        - function-declaration-expect-async
+
   ts-import-type:
     - match: type{{identifier_break}}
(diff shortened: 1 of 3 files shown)


JavaScript/tests/syntax_test_typescript.ts
```diff
@@ -359,6 +359,26 @@ function f(this : any) {}
...
+
+    x ? (y) : T => r : z;
+//  ^ variable.other.readwrite
+//    ^ keyword.operator.ternary
+//      ^^^^^^^^^^^^^ meta.function
+//      ^ meta.function.declaration
+//       ^ meta.binding.name variable.parameter.function
+//          ^ punctuation.separator.type
+//            ^ meta.type support.class
+//              ^^ storage.type.function.arrow
```
Ultra-Instinct-05, line 377: Maybe `storage.type.function.arrow keyword.declaration.function.arrow` as per scope naming guidelines 
https://www.sublimetext.com/docs/scope_naming.html#keyword ?

Thom1729: I don't have an opinion either way, but it's probably out of scope for this PR.

Ultra-Instinct-05: Alright. You can go ahead & close this review.

JavaScript/TypeScript.sublime-syntax
```diff
@@ -25,6 +25,51 @@ variables:
...
+      pop: 3
+
+  ts-detect-arrow-function-return-type:
+    - match: ':'
+      push:
+        - ts-detect-arrow-foobar
+        - ts-type-expression-end
+        - ts-type-expression-begin
+
+  ts-detect-arrow-foobar:
```
rwols, line 55: Why is this called ts-detect-arrow-**foobar**?

rchl:
Doesn't seem to help for this case: https://github.com/getsentry/sentry-javascript/blob/e9bad4792b80bc3de21bb6fcd806f31b2123602a/packages/node/src/integrations/console.ts#L32-L71

Screenshot 2020-11-06 at 10 25 51

rchl:
Also, this one has issues: https://github.com/vuejs/vetur/blob/master/server/src/vueServerMain.ts

Screenshot 2020-11-06 at 12 50 27

JavaScript/TypeScript.sublime-syntax
Thom1729: Fixed. It was a leftover temp name.

Thom1729:
Thanks for the examples. I've fixed both of them.

Thom1729:
FYI, I decided to punt on scoping `closeDialog` in `const closeDialog = (): any => setDialogOpen(false);` as a function name. Right now, we're using a fairly simplistic lookahead to scope it, and changing the lookahead broke other things. All of the anonymous function name detection code should probably be rewritten at some point (see #2267); until then, I think that failing to highlight `closeDialog` here is a tolerable bug.

rwols:
But is it really a bug? It’s a variable which happens to be a function. Should variables which are assigned a constant number be scoped as a number? IMO it’s fine to not have this “feature”.

Thom1729:
It matters for one weird technical reason and one practical reason.

The weird technical reason is that in an expression like `f = () => 42`, the anonymous function's `name` attribute will be set to `f`. This is useful because it shows up with a name in stack traces and so on. So there's a legitimate syntactic thing going on.

The practical reason is that people would like these names to show up in the symbol list as functions. This is actually pretty important because a lot of people use arrow functions preferentially over old-style function declarations.

It's an annoying special case and the syntax would be simpler without it. But it's also a genuinely useful feature and I think it's worth including despite the complexity cost.

rchl:
Looks like https://github.com/sublimehq/Packages/pull/2574#issuecomment-723039966 has regressed with latest changes.

wbond:
I tried to resolve this, but it appears #2577 significantly changed the same area of code, and it wasn't trivially clear to me how to resolve.

Thom1729:
Yep; I'm going to have to do some more work on it. #2577 should fix the peripheral issues that came up while working on this, but I'll have to re-apply the core strategy on top of those changes. Probably won't be until tomorrow, unfortunately.

Thom1729:
Closing in favor of #2590.

PR #2576, opened 2020-11-06
[TypeScript] Fix bug at end of function types.

Thom1729:
Reported in https://github.com/sublimehq/Packages/pull/2574#issuecomment-722975089, but can be merged independently (especially since that PR may need more work).

In a type expression, an opening paren may indicate either a type in parentheses or the beginning of a function type. The TypeScript syntax uses branching to distinguish them. However, because `branch` is like `push`, after parsing the type the syntax was still in the `expression-begin` state instead of `expression-end`. Adding a `pop` to the `branch` rule will fix this so that the stack is correct after parsing the function type.

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -628,6 +628,7 @@ contexts:
     - include: ts-type-basic
 
     - match: (?=\()
+      pop: true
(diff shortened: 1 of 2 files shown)


rchl:
Been running with it for a few days and haven't noticed any issues.

PR #2577, opened 2020-11-07
[JavaScript] Improve/simplify function parsing.

Thom1729:
This PR consists of a number of minor fixes and improvements to function parsing. It fixes a few minor bugs and annoyances that I ran into while working on #2574. Clearing away these issues should ease the completion of that PR and — in the future — #2267.

- Reimplement async arrow functions. Previously, async arrow functions were recognized by a lookahead in `expression-begin` (via `either-function-declaration`). This was generally fragile, and it may have been an oversight to begin with (there were no tests). This PR includes dedicated handling for async arrow functions that should work in all cases and that does not use lookaheads. It also adds tests for each of the obvious cases, including weird ones like `async => {}`.
- Simplify the lookaheads used to detect functions. There is one lookahead identifying regular functions and another identifying arrow functions. This PR removes leading spaces from both lookaheads, which requires slightly modifying some existing special cases for function identifiers. Because the regular function lookahead is exact (there is now a comment to that effect), we still use it for detecting function literals in expressions, but the arrow function lookahead (which is not exact) is now only used for highlighting identifiers as function names. The `either-function-declaration` context is gone, and there are a couple of minor bug fixes associated with that.
- Remove the `function-or-class-declaration` context. In the case of classes, it was doing the same work as the `class` context anyway, and for regular functions I added a new `regular-function` context. These are now both included explicitly where `function-or-class-declaration` was previously used.
- Add `pop: true` to the arrow function branching rule. This is consistent with the other rules in the `expression-begin` context and allows us to reduce the depth of some of the `pop`s in the branches. I would like to reduce the `pop`s even further, but I'm having trouble when replacing some `push`es with `set`s. Maybe I'll take another whack at this another time.

This work was intended to be just refactoring, but in the process it fixed some minor bugs, so it's probably worth merging on its own.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - name: Download syntax_tests
-        run: wget --content-disposition https://download.sublimetext.com/st_syntax_tests_build_4086_x64.tar.xz
+        run: wget --content-disposition https://download.sublimetext.com/st_syntax_tests_build_4090_x64.tar.xz
       - name: Extract syntax_tests
         run: tar xf "$(ls st_syntax_tests_build_*_x64.tar.xz)"
       - name: Move syntax_tests
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1634,10 +1634,9 @@ contexts:
       scope: punctuation.separator.key-value.js
       push: expression-no-comma
 
-    # This can't occur in an object literal,
-    # but it can occur in parenthesized arrow function parameters,
-    # so we handle it gracefully here.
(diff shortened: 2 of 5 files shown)


Thom1729:
At the expense of cliché, it works on my machine. 🤷  I suspect it's some kind of obscure branching issue. I'll see if I can get the CI to succeed.

rwols:
https://github.com/sublimehq/Packages/blob/master/.github/workflows/ci.yml#L19

The CI downloads 4086. Maybe something was fixed in 4090.

Thom1729:
Thanks for pointing that out. I downloaded 4086, and the same error appears. There's something about the nested branches that 4086 doesn't like. 4090 has no trouble; I know there were some fixes to branching in the meantime.

I've opened #2578 to resolve this.

PR #2584, opened 2020-11-08
[Go] Help ST with proper bracket matching

deathaxe:
ST needs all brackets scoped `string` or `comment`, which are to be ignored by bracket matching or goto matching bracket command.

Before this commit if the code contains `'}'` it was included into bracket matching causing false positives.

This commit therefore scopes "runes" as `string.quoted.single` with only the containing character assigned `constant.character.[literal|escape]`.

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -499,12 +499,15 @@ contexts:
 
   # https://golang.org/ref/spec#Rune_literals
   match-runes:
-    - match: \'({{char_escape}})'
-      scope: constant.character.go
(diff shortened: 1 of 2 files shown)


michaelblyons:
If `constant.character.literal` is okay, I'll put it in #2570, too. So far, I only see it in Scala.

deathaxe:
I found it just suitable to be able to distiunguish both easily as it was one topic in the discussion about it and because I like to be able to distinguish tokens per scopes with the same level as it avoids selectors like `constant.character - constant.character.escape` to just exclusively match literal chars.

PR #2586, opened 2020-11-09
[ASP] Rename syntax file from HTML-ASP to HTML (ASP)

deathaxe:
This commit renames the sublime-syntax file to follow the more commonly used naming scheme, which is already used by Erlang, Ruby and TCL to describe an extended version of HTML.


wbond:
I think this will require some code in core so we don't break open buffers using the old syntax filename.

PR #2588, opened 2020-11-10
[HTML] Exclude prototype from strings by default

deathaxe:
This commit moves the `meta_include_prototype` settings from ASP over to HTML package because it turns out strings to need special handling for proper `meta.string meta.interpolation source.embedded.xyz.html` in general.

The reason is the `string` scope which needs to be cleared. As a result patterns added to prototype may not be injected into string and attribute value contexts as it would make it impossible to override those with special patterns.

To handle that ASP used to override a lot of attribute value contexts of HTML just to add `meta_include_prototype: false`. Any inherited templating syntax would need to do so, which seems cumbersome.

So it feels more convenient to exclude prototype by default if it is considered useless or might prevent proper handling of "special cases".


wbond:
I'd like to change all of the comments in `HTML-Plain.sublime-syntax` from:

```
# Exclude prototypes from strings by default to kindly ask an inherited
# syntax definition to explicitly handle string interpolation.
```

to

```
# See the top of this file for why prototype is excluded
```

and in `HTML.sublime-syntax` from:

```
# Exclude prototypes from quoted strings by default to kindly ask an inherited
# syntax definition to explicitly handle string interpolation.
# See: HTML (Plain).sublime-syntax to learn more.
```

to:

```
# See HTML (Plain).sublime-syntax for why prototype is excluded
```

wbond:
Awesome, thanks!

PR #2589, opened 2020-11-11
[TSX] Support comments.

Thom1729:


wbond:
Seems to duplicate #2551

PR #2591, opened 2020-11-11
[JavaScript] Remove meta.function.declaration scope.

Thom1729:
Historically, the JavaScript syntax scoped the parts of a function before the body as `meta.function.declaration`. This is not a standard scope. In the past, it was used to populate the symbol list so that a function might appear in the list as `function foo (bar)`. However, with the new symbol list functionality, only the function name itself is shown in the symbol list, so the `meta.function.declaration` scope serves no purpose.

~~I've marked this WIP because I think it might conflict with #2590. If so, it would be a trivial fix, but it's better to apply that fix here rather than in the more complicated PR.~~

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -559,7 +559,7 @@ contexts:
     - include: else-pop
 
   function-initializer:
-    - meta_scope: meta.function.declaration.js
+    - meta_scope: meta.function.js
     - match: (=){{nothing}}
       captures:
         1: keyword.operator.assignment.js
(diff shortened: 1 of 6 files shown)


Thom1729:
I've also improved the accuracy of the `meta.function` scope:

```js
const f = () => {};
```

Previously, all of `f = () => {}` would get `meta.function` and `f = () =>` would get `meta.function.declaration`. The only reason for this was the aforementioned symbol list behavior that isn't used anymore. The latest commit limits `meta.function` to the function itself, `() => {}`. The name `f` still gets `entity.name.function`, which is all that's needed now for the symbol list.

The new behavior is more correct and it allows for removing some special cases. This should provide a clean basis for #2267.

deathaxe:
wouldn't `()` need to be scoped `meta.function.parameters`?

Thom1729:
It should be, but it wasn't before. Removing `meta.function.declaration` should clear the way to adding that. I think I'd rather do it in a separate PR — because this one touches so many lines, I'd like to keep it simple and focused.

PR #2601, opened 2020-11-11
[C#] Remove storage.type scope

deathaxe:
This commit removes storage.type in patterns which already make use of
keyword.declaration no avoid obsolete stacking of those scopes.

Note: `storage.type.variable` is interpreted as variant data type, so it
      is not renamed to keyword.declaration.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -729,7 +729,7 @@ contexts:
       scope: storage.modifier.other.cs
       push: type_constraint
     - match: (=>)
-      scope: storage.type.function.cs
+      scope: keyword.declaration.function.anonymous.cs
       set:
         - meta_content_scope: meta.method.cs
         - include: line_of_code_in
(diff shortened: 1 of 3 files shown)


jwortmann:
I found two other occurences for the arrow operator / lambda function arrow in this syntax:
https://github.com/sublimehq/Packages/blob/1218ebbb58c01641361402bfeb0a2633479dd5b3/C%23/C%23.sublime-syntax#L982
https://github.com/sublimehq/Packages/blob/1218ebbb58c01641361402bfeb0a2633479dd5b3/C%23/C%23.sublime-syntax#L1002

And there is also `storage.type.function.accessor.get.cs` used for both `=>` and `get`.
https://github.com/sublimehq/Packages/blob/1218ebbb58c01641361402bfeb0a2633479dd5b3/C%23/C%23.sublime-syntax#L582
https://github.com/sublimehq/Packages/blob/1218ebbb58c01641361402bfeb0a2633479dd5b3/C%23/C%23.sublime-syntax#L774
Seems like the arrow has a different meaning here, but I'm not sure as I'm no C# user. Maybe @michaelblyons can help.
Could the arrow please receive a different scope here at least, e.g. `storage.type.function.accessor.get.arrow.cs`, so that color schemes can distinguish between symbols and keywords?

Issue #2585, opened 2020-11-09
[PHP] Greedily scope the $ as punctuation.definition.variable


rwols:
- Sublime Version: 4090

```php
<?php
$
# <- missing punctuation.definition.variable
```

In order for plugins to better filter when to show completions via an entry in the `auto_complete_triggers` of the view.settings(), it would be nice to have this.

PR #2602, opened 2020-11-12
[PHP] Greedily scope the $ as punctuation.definition.variable

jfcherng:
Resolves https://github.com/sublimehq/Packages/issues/2585

---

Add a workaround for dangling `$` in interpolated regex because it means EOL.

```php
$var = "foo";
$regex = "/$var $/";
                ^ not punctuation.definition.variable, but keyword.control.anchor.regexp
```

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -1583,6 +1583,10 @@ contexts:
           with_prototype:
             - match: (?=(/)({{regex_modifier}})(")|")
               pop: 1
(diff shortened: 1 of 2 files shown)


wbond:
Can we add a test for the regex usage of `$`?

jfcherng:
> Can we add a test for the regex usage of `$`?

It has been already there. https://github.com/sublimehq/Packages/blob/26df3267f588b248b1bd9eecf6fbd511c7859093/PHP/tests/syntax_test_php.php#L1661

wbond:
Ok, thanks!

PR #2604, opened 2020-11-12
[JavaScript] Use keyword.declaration.function instead of storage.type.function

Thom1729:
@deathaxe 

Currently, `async` is still `storage.type` and the asterisk in a generator function is `keyword.generator.asterisk`. Either or both of these could be re-scoped if anyone has a suggestion.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -947,7 +947,7 @@ contexts:
 
   async-arrow-function:
     - match: async{{identifier_break}}
-      scope: storage.type.js
+      scope: keyword.declaration.async.js
       set:
         - function-meta
         - arrow-function-expect-body
@@ -1374,7 +1374,7 @@ contexts:
         - match: (?=\S)
           fail: prefixed-method
     - match: (?:async){{identifier_break}}
-      scope: storage.type.js
+      scope: keyword.declaration.async.js
       set:
         - meta_scope: meta.function.js
         - match: (?=\*|{{class_element_name}})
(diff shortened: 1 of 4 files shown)


deathaxe:
**async**

I propose `keyword.declaration.async.python` for `async` in the Python PR. It is clearly a declaration of a coroutine, thus `keyword.declaration` seems suitable for it.

**asterisk**

I am not sure about the asterisk usage as generator in JavaScript, but struggled with trying to find a common scope for such tokens in CSS and Java recently. Those are basically wildcards in general. Makefile scopes them `variable.language.wildcard`. A couple of syntaxes use `keyword.other.wildcard`. Being used in a string literal we also might scope it `constant.other.placeholder`.

We might also think of anonymous/blank variables like `_` in python or scala, which are scoped `variable.language.blank` or `variable.language.anonymous`. It feels like the same category as wildcards somehow.

I wonder if we can cum up with general scopes for such things as

- `*` == `variable.language.wildcard` with `.asterisk` added optionally in case a distinction is needed/wanted. I have `?` in mind.
- `_` == `variable.language.anonymous`

A special usage/meaning may be appended then.

The ladder one leads us to anonymous function definitions aka. lambdas.

**lambdas**

I'd suggest to scope `=>` in JS, `->` in Java, `lambda` in python, `fun()` in Erlang, ... as `keyword.declaration.function.anonymous` or `keyword.declaration.lambda`. The ladder one feels a bit uncommon with regards to wording in Erlang, but wouldn't be wrong there. The first would match variable scopes with regards to `.anonymous` and expresses clearly the function character.

Thom1729:
I've changed `async` to `keyword.declaration.async`. Following that example, I've changed the asterisk to `keyword.declaration.generator`. It's not really acting as a wildcard here, so I think it makes more sense to stay with `keyword.declaration`.

PR #2606, opened 2020-11-12
[ActionScript] Rename storage.type scope

deathaxe:
This commit replaces
a) storage.type.[class|function] by keyword.declaration.[class|function]
b) storage.modifier.extends by keyword.declaration.extends


wbond:
I'm not sure how I feel about changing `extends`/`implements` to a `keyword.declaration`. I don't think those are the same thing as `class`, `def`, `var`.

deathaxe:
Renamed that as I used it in Java for some reason I don't recall anymore, but honestly got in doubt with it, too. `storage.modifier` may indeed be the better choice.

PR #2607, opened 2020-11-12
[C/C++/Objective-C/Objective-C++] Keyword declaration

deathaxe:
This commit renames some `storage.type` scopes into `keyword.declaration` to highlight keywords like `class`, `union`,
`template`, `typename`, `struct`, ... as declaration keywords rather than primitive/ordinary data types.

Note:
It is an as good as possible with least effort approach, which means some `class`, ... keywords to be scoped `keyword.declaration.[c|c++|...]` only due to existing patterns. Don't want to rework the syntax at this point by splitting those.


wbond:
This seems like a step in the right direction, thanks!

PR #2610, opened 2020-11-13
[C#] Use comment.line scope for single-line doc comments

jwortmann:
This PR changes the scope for single-line documentation comments (denoted by `///`) from `comment.block.documentation` to `comment.line.documentation`, because:

* They *are* line comments and not block comments
* The Microsoft C# documentation at https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments explicitely distinguishes between *single-line-doc-comment* (starting with `///`) and *delimited-doc-comment* (starting with `/**`)
* There are several examples of such doc comments in the linked C# docs which indeed use only a single line, so they don't necessarily appear in blocks
* This is consistent with R, which also uses `comment.line` for line doc comments
* This complies with the scope naming guidelines, which recommends `comment.block.documentation` only for "*Multi-line* comments used as documentation"

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -56,7 +56,7 @@ contexts:
   comments:
     - match: '^\s*(///)'
       captures:
-        1: comment.block.documentation.cs punctuation.definition.comment.documentation.cs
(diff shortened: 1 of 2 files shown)


wbond:
Looks like a good change to me, thanks!

PR #2611, opened 2020-11-13
[ShellScript] Add anonymous functions

### LINKED ISSUE #2608: [Bash] Zsh anonymous functions highlighting
- OS Version: Arch Linux

```zsh
if [[ "$OSTYPE" == linux* ]]; then
  echo "outside is OK"
fi
() {
  if [[ "$OSTYPE" == linux* ]]; then
    echo "inside () {} the syntax is broken"
  fi
}
```

Screenshot_20201112_214200

AFAIK, the anonymous function syntax is only for zsh, but unfortunately there's no separate package for it. They are kind of functions which solely exist to provide scoping for local variables; other than that they are executed immediately at the place of definition.

deathaxe:
Fixes #2608

This commit handles `() { ... }` as incomplete function definition in Bash, while it is valid anonymous function definition in ZSH.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -99,6 +99,7 @@ contexts:
     - include: line-continuations
     - include: operators
     - include: redirections
+    - include: def-anonymous
     - include: cmd-alias
     - include: cmd-arithmetic
     - include: cmd-compound
@@ -191,6 +192,22 @@ contexts:
 
 ###[ FUNCTION DEFINITIONS ]####################################################
 
+  def-anonymous:
+    - match: (\()\s*(\))
+      scope: meta.function.parameters.shell
+      captures:
+        1: punctuation.section.parameters.begin.shell
+        2: punctuation.section.parameters.end.shell
+      push:
+        - def-functions-redirection
+        - def-anonymous-body
+
+  def-anonymous-body:
+    - include: def-function-body-braces
+    - include: line-continuations
+    - include: comments
+    - include: else-pop
+
   def-functions:
     # [Bash] 3.3 Shell Functions
     - match: (?={{cmd_literal}}\s*\(\s*\))
@@ -230,9 +247,18 @@ contexts:
         1: punctuation.section.parameters.begin.shell
         2: punctuation.section.parameters.end.shell
       pop: 1
+    - include: line-continuations
+    - include: comments
     - include: else-pop
 
   def-functions-body:
+    - include: def-function-body-braces
+    - include: def-function-body-parens
+    - include: line-continuations
+    - include: comments
+    - include: else-pop
+
+  def-function-body-braces:
     - match: \{  # Bash expects `{{cmd_break}}` but we don't care.
       scope: punctuation.section.compound.begin.shell
       set:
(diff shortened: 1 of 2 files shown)


ShellScript/Bash.sublime-syntax
```diff
@@ -191,6 +192,22 @@ contexts:
 
 ###[ FUNCTION DEFINITIONS ]####################################################
 
+  def-anonymous:
+    - match: (\()\s*(\))
```
ratijas, line 196: I'm pretty sure, whitespaces aren't allowed between parentheses:

```zsh
% () { echo hi; }
hi
% (   ) { echo hi; }
zsh: parse error near `{'
```

```suggestion
    - match: (\()(\))
```

ShellScript/test/syntax_test_bash.sh
```diff
@@ -865,6 +865,46 @@ coproc foobar {
...
 ####################################################################
 
+   ()
+#^^ - meta.function
+#  ^ meta.function.parameters.shell
+#   ^ meta.function.parameters.shell
+#    ^ meta.function.shell
+
+   ()
+   {}
```
ratijas, line 875: Funny enough, this is perfectly valid zsh syntax:

```zsh
% () () () { echo hi; }
hi
```

ratijas (changes requested):
Hi! Thanks for the fix. I have one little suggestion below.

ShellScript/Bash.sublime-syntax
deathaxe: Yes, but we might not want to break highlighting due to such details.

ShellScript/test/syntax_test_bash.sh
deathaxe: I'd expect `() () (echo hi;)` to work as the first `()` would be the function name, the second could be the arguments dummy and the 3rd one the function body.

But honestly, we can find dozens of such inconsistencies in all script/shell syntaxes created in the stoneage of PCs's (80s and 90s) when chaos and adhoc implementations drove development. All of those lack correctly structured syntax which causes them to be filled with dozens of ambigulties.

If we want to highlight them all, syntax will probably grow to several Megabytes.

ShellScript/Bash.sublime-syntax
michaelblyons: You could make the `\s*` itself a capture group that is `illegal`

deathaxe: We could, yes. But this context is a copy of the original implementation of arguments lists only. 

ST's auto match feature makes `( )` a very common pattern if we start editing groups or other things. To be honest, I don't like all the illegal highlighting too much. It may help to find really illegal patterns, but should not disturb a user too much while createing expressions. Incomplete expressions are illegal very often with regards to specs.

ShellScript/test/syntax_test_bash.sh
ratijas: I mean, isn't that awesome? :D

ShellScript/Bash.sublime-syntax
ratijas: If you think about it, most other editors and IDEs highlight illegal stuff with red underline or something like that — not as distracting as ST which just flashes red background all over the text.

deathaxe: Underlining illegals is what I guess would happen with an LSP server running in the background.

wbond:
Thanks for the fix and discussion on this!

PR #2613, opened 2020-11-13
[D] Fix conditional statements

### LINKED ISSUE #2558: [D] Ternary statement incorrectly marks as syntax error
### Description

The changes to syntax definitions in ST4 have greatly improved parsing D code. However, ternary statements now mark as though they're a syntax error.

### Steps to reproduce

Type this code into a D source file:

```d
void func() {
  int a = b ? c : d;
}
```

Result:

Image

### Environment

* Build: 4088
* Operating system and version: Elementary OS 5.1
* Linux Desktop Environment and/or Window Manager: Pantheon/Gala

### LINKED ISSUE #2567: [D] Switch-case throws off highlighting
- Sublime Version: 4090
- OS Version: Elementary OS 5.1 Freya

```d
void func() {
  switch (perform_air_step(m, 0)) {
    case AIR_STEP_NONE:
      if (m.vel[1] < 0.0f && m.faceAngle[0] > -0x2AAA) {
        m.faceAngle[0] -= 0x200;
        if (m.faceAngle[0] < -0x2AAA) {
          m.faceAngle[0] = -0x2AAA;
        }
      }
      break;
    
    default: break;
  } 
}
```

image

Thank you!

deathaxe:
Fixes #2558
Fixes #2567

Problem Description

D matches mapping keys globally in a general `value` context, which causes identifiers followed by colon such as `c :` to be matched as mapping key instead of variable followed by ternary operator or case label separator.

Thus ternary statements and switch case statements fail in such situations.

Solution

This commit creates intermediate contexts `value-condition` and `value-no-mapping-key` to avoid matching keys within ternary statements.

Case statements are fixed by `value-list-no-mapping-key`.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -1494,6 +1494,11 @@ contexts:
       pop: true
     - match: '(?=\S)'
       set: [value-list-after, value]
+  value-list-no-mapping-key:
+    - match: '(?=\)|}|]|;|:)'
+      pop: true
+    - match: '(?=\S)'
+      set: [value-list-after, value-no-mapping-key]
   value-list-after:
     - match: '(?=\)|}|]|;|:)'
       pop: true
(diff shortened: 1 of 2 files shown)


wbond:
Thanks!

PR #2614, opened 2020-11-14
[JavaScript] Enable auto-indentation rules for Typescript

rchl:


rchl:
I suppose also `sources.tsx` and `source.jsx` should be added?

wbond:
Thank you!

PR #2615, opened 2020-11-14
[JavaScript] Improve handling of identifiers.

Thom1729:
If we encounter an unexpected reserved word while parsing an expression, bail out of the expression. This should improve highlighting while typing in some cases.

In strict mode, additional words are reserved, but handling this would require duplicating the expression contexts, so it's not worth it.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -449,7 +449,7 @@ contexts:
           scope: punctuation.section.block.end.js
           pop: true
         - include: variable-binding-spread
-        - match: (?={{identifier_name}}|\[|'|")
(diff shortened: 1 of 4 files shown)


Thom1729:
I've also renamed the `identifier` variable to `identifier_name`. The JavaScript spec uses `IdentifierName` to include both identifiers and reserved words, and `Identifier` to exclude reserved words. The `non_reserved_identifier` variable could be renamed `identifier`, but I think leaving it as-is will encourage anyone using `identifier_name` or `non_reserved_identifier` to consider which one is correct.

I also added a couple of tests verifying that reserved words are allowed in certain situations, like object properties and JSX tags.

wbond:
Thank you @Thom1729!

PR #2616, opened 2020-11-14
[C#] Use keyword.declaration.function

deathaxe:
This commit scopes all remaining function declaration keywords `keyword.declaration.function`.

This commit scopes the `=>` as `keyword.declaration.function.anonymous`.

It was applied to C#, D recently and will be used in Java an Python as a common alternative to the existing variants of defining anonymous functions.

We currently have:

- keyword.declaration.function.anonymous
- keyword.declaration.function.arrow
- keyword.declaration.function.inline
- keyword.declaration.function.lambda


wbond:
Looks good!

PR #2617, opened 2020-11-14
[JavaScript] Scope arrow as anonymous function

deathaxe:
This commit scopes the `=>` as `keyword.declaration.function.anonymous`.

It was applied to C#, D recently and will be used in Java an Python as a common alternative to the existing variants of defining anonymous functions.

We currently have:

- keyword.declaration.function.anonymous
- keyword.declaration.function.arrow
- keyword.declaration.function.inline
- keyword.declaration.function.lambda

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -987,7 +987,7 @@ contexts:
 
   arrow-function-expect-arrow-or-fail-async:
     - match: '=>'
-      scope: keyword.declaration.function.arrow.js
+      scope: keyword.declaration.function.anonymous.js
       pop: true
     - match: (?=\S)
       fail: async-arrow-function
@@ -1530,7 +1530,7 @@ contexts:
 
   arrow-function-expect-arrow:
     - match: '=>'
-      scope: keyword.declaration.function.arrow.js
+      scope: keyword.declaration.function.anonymous.js
       pop: true
     - include: else-pop
 
--- a/JavaScript/tests/syntax_test_js.js
+++ b/JavaScript/tests/syntax_test_js.js
@@ -1226,7 +1226,7 @@ class{}/**/
 //  ^^ meta.function.parameters
 //  ^ punctuation.section.group.begin
 //   ^ punctuation.section.group.end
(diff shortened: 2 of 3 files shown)


Thom1729:
This seems reasonable as a convention, but it's worth noting that JavaScript supports anonymous “regular” function literals as well, e.g. `(function() {})`. Marking those as `anonymous` would require branching and probably wouldn't be worth it, but it's worth mentioning.

Thom1729 (approved):

deathaxe:
I wonder, whether we might need to distinguish scopes for real keywords and arrow operators here. What comes to my mind here is the `keyword.operator.word` discussion.

Thom1729:
The arrow isn't really an operator; it's part of a literal. I think it makes sense to keep the `keyword.declaration` scope.

I was never a fan of the `keyword.operator.word` scope anyway.

deathaxe:
Not a fan of it, too. Didn't want to suggest to move something to `keyword.operator.word` - just point to keywords which have operator like vs. word like counterparts.

Color schemes might want to highlight an arrow (`=>`) differently than a literal keyword (`function`) without targeting each syntax individually. As JS supports both, it might even be more important to avoid ambiguities.

While Java uses `->` to express lambdas, python uses `lambda`, ... . Erlang knows about `fun()` as anonymous function declaration.

I am not quite sure how to sovle that, but maybe ...

- keyword.declaration.function.anonymous  -> literal keywords like `fun`, `function`, `lambda`
- keyword.declaration.function.arrow -> all arrow like lambda definitions `->`, `=>`

... or alternatively maybe ...

- keyword.declaration.function.anonymous  -> literal keywords like `fun`, `function`, `lambda`
- punctuation.definition.function.anonymous  -> all arrow like lambda definitions `->`, `=>`

wbond:
> keyword.declaration.function.anonymous -> literal keywords like fun, function, lambda
> keyword.declaration.function.arrow -> all arrow like lambda definitions ->, =>

I like this convention.

deathaxe:
So we may ...

1. note it for scope naming guidelines
2. and adjust the other syntaxes

and close this PR.

PR #2618, opened 2020-11-15
[XML] Add xsl file extension

deathaxe:


rwols:
XSL is a separate language ID in the LSP spec https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem

Wouldn't this change cause ST to not be able to distinguish between XML and XSL?

keith-hall:
> Wouldn't this change cause ST to not be able to distinguish between XML and XSL?

Isn't that already the case unless someone has a third party package like https://packagecontrol.io/packages/XSL installed providing an XSLT syntax definition?

rwols:
@keith-hall you're right this syntax already has .xslt as file extension https://github.com/sublimehq/Packages/blob/1218ebbb58c01641361402bfeb0a2633479dd5b3/XML/XML.sublime-syntax#L12

Since we also have a DTD.sublime-syntax it might we worthwhile to have an XSL.sublime-syntax for the file extensions .xsl and .xslt with base scope `text.xml.xsl` that just includes `scope:text.xml`. Well, since @deathaxe also maintains [lemminx](https://packagecontrol.io/packages/LSP-lemminx), which is probably the only language server interested in xml and xsl, I trust he knows what he's doing :)

deathaxe:
You mean this one?

grafik

I actually have that XSL.sublime-syntax in my private setup which is an alias for XML to provide a dedicated scope to handle XSL related snippets.

I haven't proposed it as @wbond actively removed that syntax from this package, earlier, even thouh it would definitly be the cleaner solution.

I was working on A File Icon today, when I realized xsl files not showing xml icons in a vanilla ST, so this was just a quick shot.

wbond:
> I haven't proposed it as @wbond actively removed that syntax from this package, earlier, even thouh it would definitly be the cleaner solution.

I don't recall the rationale behind that, perhaps I wasn't aware of a practical reason at the time.

If you think having an XSL syntax would be beneficial, I'd be up for that. You've invested a lot of time in the syntaxes, but especially HTML and XML, so I'm willing to trust your judgement.

deathaxe:
A dedicated XSL.sublime-syntax would be as huge as DTD.sublime-syntax. It would be XML with a replaced top-level scope such as `text.xml.xsl` in order to support XSL related snippets and completions as well as LSP sending the correct language id `xsl` to a language service out of the box.

I'd use `text.xml.xsl` rather than `text.xsl` because

a) it is basically a subtype of XML
b) DTD.sublime-syntax uses `text.xml.dtd` already.

XSL related snippets could be provided via simple 3rd-party package as LemMinX matures and provides better intellingence features even for XSL files with each build, so most users might want no snippets, probably.

deathaxe:
FWIW, the PR assigns xsl, xsd and xslt extensions to a dedicated XSL.sublime-syntax

wbond:
Looks good, thanks!

PR #2619, opened 2020-11-15
[ASP] Add vbs file extension

deathaxe:
ASP is visual basic. Thus we can also use that syntax for vbs files.

### DIFF
--- a/ASP/ASP.sublime-syntax
+++ b/ASP/ASP.sublime-syntax
@@ -4,6 +4,7 @@
 name: ASP
 file_extensions:
   - asa # asp is handled by HTML-ASP.sublime-syntax
+  - vbs # Visual Basic Script
 scope: source.asp
 variables:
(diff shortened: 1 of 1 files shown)


keith-hall:
> ASP is visual basic. Thus we can also use that syntax for vbs files.

Technically, classic ASP pages can choose to use VBScript or JScript script blocks inside them, so it'd be more accurate to rename the ASP syntax definition to VBScript (while keeping the HTML (ASP) one as is.)

keith-hall (approved):

deathaxe:
You are absolutely right. The reason why this is not done here is because we Maybe can create a VBScript syntax without HTML interpolations as base for ASP at some point.

ASP needs some work with regards to supporting ASP tags within embedded CSS/JS anyway.

wbond:
Sounds like this is a reasonable option for now, so I'll merge.

PR #2620, opened 2020-11-15
[Python] Don't cancel completions for function definitions

### LINKED ISSUE #2404: [Python] Consider relaxing completion cancelation rules
- Sublime Version: 4077
- OS Version: macOS 10.15.5

In python files completions are inhibited quite aggressively, I feel. Not sure what is the reasoning for this behavior so I would suggest to reconsider current behavior and potentially relax it.

## Expected behavior
It should be possible to trigger completion popup both in:

 - case A (currently possible):

```py
def |
```

 - case B (currently impossible):

```py
def a|
```

## Actual behavior

It's only possible to trigger completions popup in case A.

## Steps to reproduce

1. Insert code (where `|` is the caret position):
```py
def aaa:
	pass

def a|
```
2. Try to invoke completions (using ctrl+space or whatever is the relevant shortcut for your platform)

The behavior is controlled by:
https://github.com/sublimehq/Packages/blob/40ec1f2f9b56fb55d739e17ecd003cc9e8c9b096/Python/Completion%20Rules.tmPreferences#L8-L9
and it seems like this logic was there more or less from the beginning.

FichteFoll:
Other object-oriented languages in the default packages only disable
completions for class/type names, because in object-oriented languages
you can make use of function overloading, including the commonly
defined dunder functions. Most notably, this allows auto-completion
plugins like LSP to function properly.

Aside from that, the regex was merely restructured.

Closes #2404.

### DIFF
--- a/Python/Completion Rules.tmPreferences
+++ b/Python/Completion Rules.tmPreferences
@@ -6,7 +6,12 @@
 	<key>settings</key>
 	<dict>
 		<key>cancelCompletion</key>
-		<string>^(.*\b(and|or)$)|(\s*(pass|return|and|or|(class|def)\s*[a-zA-Z_0-9]+)$)</string>
+		<string>(?x)
+		# Prevent completions when typing keywords
+		\b(and|or|pass|return)$
+		| # Prevent completions for class names
+		^\s* class \s+ [a-zA-Z_0-9]+ $
+		</string>
 	</dict>
 </dict>
 </plist>


Python/Completion Rules.tmPreferences
```diff
@@ -6,7 +6,12 @@
 	<key>settings</key>
 	<dict>
 		<key>cancelCompletion</key>
-		<string>^(.*\b(and|or)$)|(\s*(pass|return|and|or|(class|def)\s*[a-zA-Z_0-9]+)$)</string>
+		<string>(?x)
+		# Prevent completions when typing keywords
+		\b(and|or|pass|return)$
+		| # Prevent completions for class names
+		^\s* class \s+ [a-zA-Z_0-9]+ $
+	</string>
```
rchl, line 14: Indentation doesn't seem right :)

PR #2624, opened 2020-11-15
[JavaScript] Move control statement tests to separate file.

Thom1729:


deathaxe:
I don't find splitting tests into dozens of separate files very useful as there is no sane way to run them together.

Creating good logical sections within a test files seems more useful.

Thom1729:
If you run the Syntax Tests build system on a syntax definition, it will find and run all associated syntax tests.

deathaxe:
I haven't been aware of that, indeed. Maybe because running syntax tests for a sublime-syntax doesn't have any effect here except a traceback being printed to console.

```
  File "C:\Apps\Sublime Text\Lib\python38\sublime_plugin.py", line 1473, in run_
    return self.run(**args)
  File "C:\Apps\Sublime Text\Packages\Default.sublime-package\run_syntax_tests.py", line 41, in run
IndexError: list index out of range
```

But even if that works, I am not sure if it would make sense to run syntax tests agains all benchmarking files on present in various syntaxes I've been working at.

Thom1729:
Is it necessary for benchmarking files to be syntax tests? You can run performance tests on any file.

It looks like the error you're seeing would occur if there's a resource whose name begins with `syntax_test` that is completely empty. This is arguably a bug in `run_syntax_tests`.

FWIW, the JavaScript syntax already uses eight separate test files (not counting JSX, TypeScript, or TSX), so it's sort of a moot point in that case.

Thom1729:
I've opened https://github.com/sublimehq/sublime_text/issues/3760 for the `run_syntax_tests` error.

deathaxe:
> Is it necessary for benchmarking files to be syntax tests?

Syntax tests such as the Performance test is enabled for `syntax_test_` files only, isn't it?

> This is arguably a bug in run_syntax_tests.

Yes, it is, but it looks more likely to happen due to `load_resource` failing with large files or those not using utf-8 encoding.

> FWIW, the JavaScript syntax already uses eight separate test files (not counting JSX, TypeScript, or TSX),

Yes, one for JSX, one for TSX, one for TS and so should JS as well.

Thom1729:
> Syntax tests such as the Performance test is enabled for syntax_test_ files only, isn't it?

It works on any file. If the file name begins with `syntax_test`, then it will automatically select the Syntax Tests build system; otherwise, you would have to select it manually. You don't need a `// SYNTAX TEST` header, though; the performance test command doesn't care. So if you have separate performance tests, you can simply omit that header and the `run_syntax_tests` command won't run them.

> Yes, one for JSX, one for TSX, one for TS and so should JS as well.

To clarify, the pure-JavaScript tests are already spread across eight files containing a total of 9,677 assertions across 3,188 lines. A number of other syntaxes also split tests across several files. If we do decide that we want to consolidate these into a single file, then this would be a substantial change.

Thom1729:
I think that if we're talking about consolidating syntax tests in general, then we should probably move that discussion to an RFC where it will get more input.

PR #2629, opened 2020-11-18
[JavaScript] Handle type argument in JSDoc documentation

rchl:
Enables parsing of types that can follow some block tags.

The type is scoped with a single scope instead of trying to parse the
type in a "correct" way (how typescript syntax does for types). This is
for one due to the fact that I don't know how to do it "properly" but
even more importantly because typing it properly would make the type
more noisy visually. The only exception is that the dot is scoped as
an accessor so that it triggers completions (especially useful when
using LSP).

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -193,7 +193,7 @@ contexts:
       pop: 1
     - match: '^\s*\*'
     - match: \.
-      scope: meta.type.js punctuation.accessor.js
+      scope: meta.type.js punctuation.accessor.dot.js
     - match: '\{'
       scope: meta.type.js
       push: jsdoc-type-expression-body-nested
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -107,7 +107,7 @@ variables:
   # '@' followed by a pattern like \S but excluding literal '*' and '@'.
   jsdoc_block_tag: \@[^\n\t\f\v *@]+
   # Tags that are optionally followed by a type expression.
-  jsdoc_block_tag_with_type: \@(?:arg(?:ument|)|param|type(?:def|)|returns?|prop(?:erty|)|enum|implements|member|module|namespace|(?:pr(?:ivate|otected))|template|this|throws|exception|yields?|var)\b
+  jsdoc_block_tag_with_type: \@(?:arg(?:ument)?|param|typedef|type|returns?|prop(?:erty)?|enum|implements|member|module|namespace|private|protected|template|this|throws|exception|yields?|var)\b
 
 contexts:
   main:
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -193,16 +193,16 @@ contexts:
       pop: 1
     - match: '^\s*\*'
     - match: \.
-      scope: support.type.js punctuation.accessor.js
+      scope: meta.type.js punctuation.accessor.js
     - match: '\{'
-      scope: support.type.js
+      scope: meta.type.js
       push: jsdoc-type-expression-body-nested
     - match: '(?:(?!\*\/|^\s*\*|\}|\{|\.).)*'
-      scope: support.type.js
+      scope: meta.type.js
 
   jsdoc-type-expression-body-nested:
     - match: '\}'
-      scope: support.type.js
+      scope: meta.type.js
       pop: 1
     - include: jsdoc-type-expression-body
 
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -106,6 +106,8 @@ variables:
 
   # '@' followed by a pattern like \S but excluding literal '*' and '@'.
   jsdoc_block_tag: \@[^\n\t\f\v *@]+
+  # Tags that are optionally followed by a type expression.
+  jsdoc_block_tag_with_type: \@(?:arg(?:ument|)|param|type(?:def|)|returns?|prop(?:erty|)|enum|implements|member|module|namespace|(?:pr(?:ivate|otected))|template|this|throws|exception|yields?|var)\b
 
 contexts:
   main:
@@ -164,11 +166,45 @@ contexts:
       push: jsdoc-block-tag
 
   jsdoc-block-tag:
+    - match: '{{jsdoc_block_tag_with_type}}'
+      scope: entity.other.attribute-name.documentation.js
+      set:
+        - match: '\s+(?=\{)'
+          set: jsdoc-type-expression
+        - match: ''
+          pop: 1
     - match: '{{jsdoc_block_tag}}'
       scope: entity.other.attribute-name.documentation.js
       pop: 1
     - match: (?=\S)|$
       pop: 1
+
+  jsdoc-type-expression:
+    - match: '\{'
+      scope: punctuation.definition.type.begin.js, meta.tag.jsdoc
+      set:
+        - match: '\}'
+          scope: punctuation.definition.type.end.js, meta.tag.jsdoc
+          pop: 1
+        - include: jsdoc-type-expression-body
+
+  jsdoc-type-expression-body:
+    - match: (?=\*/)
+      pop: 1
+    - match: '^\s*\*'
+    - match: \.
+      scope: support.type.js punctuation.accessor.js
+    - match: '\{'
+      scope: support.type.js
+      push: jsdoc-type-expression-body-nested
+    - match: '(?:(?!\*\/|^\s*\*|\}|\{|\.).)*'
+      scope: support.type.js
+
+  jsdoc-type-expression-body-nested:
+    - match: '\}'
+      scope: support.type.js
+      pop: 1
+    - include: jsdoc-type-expression-body
 
   line-comments:
     - match: /{4,}
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -197,7 +197,17 @@ contexts:
     - match: '\{'
       scope: meta.type.js
       push: jsdoc-type-expression-body-nested
-    - match: '(?:(?!\*\/|^\s*\*|\}|\{|\.).)*'
+    - match: |-
+        (?x:
+          (?! # break on patterns handled explicitly above
+            \*\/|
+            ^\s*\*|
+            \}|
+            \{|
+            \.
+          ).
+        # consume as many consecutive characters as possible to not break the ligatures
+        )*
       scope: meta.type.js
 
   jsdoc-type-expression-body-nested:
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -191,7 +191,9 @@ contexts:
   jsdoc-type-expression-body:
     - match: (?=\*/)
       pop: 1
-    - match: '^\s*\*'
+    - match: '^\s*(\*)'
+      captures:
+        1: punctuation.definition.comment.js
     - match: \.
       scope: meta.type.js punctuation.accessor.dot.js
     - match: '\{'
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -184,7 +184,7 @@ contexts:
       scope: punctuation.definition.type.begin.js, meta.tag.jsdoc
       set:
         - match: '\}'
-          scope: punctuation.definition.type.end.js, meta.tag.jsdoc
+          scope: meta.tag.jsdoc punctuation.definition.type.end.js
           pop: 1
         - include: jsdoc-type-expression-body
 
--- a/JavaScript/tests/syntax_test_js_jsdoc.js
+++ b/JavaScript/tests/syntax_test_js_jsdoc.js
@@ -56,13 +56,17 @@
   /**
    * @param {{
 //           ^ meta.type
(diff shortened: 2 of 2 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -103,6 +103,8 @@ variables:
 
   # '@' followed by a pattern like \S but excluding literal '*' and '@'.
   jsdoc_block_tag: \@[^\n\t\f\v *@]+
+  # Tags that are optionally followed by a type expression.
+  jsdoc_block_tag_with_type: \@(?:arg(?:ument|)|param|type(?:def|)|returns?|prop(?:erty|)|enum|implements|member|module|namespace|(?:p|rivate|protected)|throws|exception|yields?|var)\b
```
jwortmann, line 107: No JSDoc expert here, but is there a mistaken extra `|` in `p|rivate`? And what's the purpose of this uncaptured group around `private|protected`?
Also e.g. `arg(?:ument)?` would be the same as `arg(?:ument|)`, but more readable imo. Or maybe even just write them out as `arg|argument`.

rchl: > but is there a mistaken extra `|` in `p|rivate`?

This looks wrong. Good catch.

> Also e.g. `arg(?:ument)?` would be the same as `arg(?:ument|)`, but more readable imo.

Yes, that should be the same. I can go with either. No strong feelings.

> Or maybe even just write them out as `arg|argument`.

I could but that's a question I should try to optimize here or not bother with it.

michaelblyons: > > Or maybe even just write them out as `arg|argument`.
> I could but that's a question of whether I should try to optimize here or not bother with it.

IIRC the engine optimises them internally, whichever of those you write. So pick the one that looks nice for humans.

rchl: I know there was a discussion about that somewhere but I can't find it now. Are you able to provide a reference for that?

jwortmann: I think I've also seen something about this related to the PHP syntax before somewhere here, but I don't know where exactly.

Another thing I noticed from looking at this implementation (I might be wrong because I haven't tested it), is that you push/set into an anynomous context with the first opening `{`, but only pop on `}`. What happens if someone writes
```
/** @type {string  no closing brace here */
```
I guess that might confuse the JSDoc parser, but it is still a valid comment and a valid JS file, right? If it is, you might better add a syntax test for such cases with missing single or double braces. Right now I assume the rest of the file be highlighted as a comment then, until it hits a closing brace somewhere.

rchl: You are right. This should be handled better. Should handle the comment-end marker at least so that highlighting doesn't continue to the rest of the file.

(Just a note that those type blocks can be multiline so we can't interrupt parsing at the end of the line.)

wbond:
@Thom1729 Wondering if you have any thoughts on this?

JavaScript/JavaScript.sublime-syntax
rchl: Note: That issue was fixed with later commits.

rchl:
@Thom1729 ping :)

Thom1729:
Whoops; I started typing a response and never posted it.

Because JSDoc is nonstandard, there is some risk to including it in the core syntax. Right now, the implementation is very simple and the chance of conflict is small.

JSDoc type expressions apparently use [the syntax from Google's Closure compiler](https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System). This syntax looks to be fairly complicated — comparable to TypeScript's type syntax, though not compatible with it. In fact, it looks like TypeScript will read JSDoc type annotations, and it will accept them in either Closure or TypeScript form, which sounds like it would be fun to parse. I think that full support for Closure type expressions would be a lot more complexity than we'd want to add to the core syntax. (There's an open ticket for JS Custom, which I admittedly haven't done anything with.)

Of course, this PR isn't nearly so complicated — it's a minor improvement that doesn't require that much code. It's hard for me to evaluate because I usually try to start from an underlying guiding principle and apply it to the particular case. Here we don't have clear guidelines. If JSDoc were a standard language feature, I'd say we should go the whole way and implement the Closure syntax. But since it's not, I'd rather err on the side of caution.

It might help if I better understood the specific use cases here. Is the main benefit improved LSP completions? Are there any simple examples of annotations which look bad with the current highlighting?

rchl:
> Because JSDoc is nonstandard, there is some risk to including it in the core syntax. Right now, the implementation is very simple and the chance of conflict is small.

Note that the implementation has already started before with #2557. That initial PR implemented parsing of _@tags_. This extends the support further into _types_. And potentially there is gonna be a follow-up with implementation of `@param`'eter names (`foo` in `@param {string} foo This is a parameter`).

> JSDoc type expressions apparently use [the syntax from Google's Closure compiler](https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System). ... I think that full support for Closure type expressions would be a lot more complexity than we'd want to add to the core syntax. (There's an open ticket for JS Custom, which I admittedly haven't done anything with.)

Yes, not only it would be very complicated to implement full parsing of types but also we **really** wouldn't want to do that because it would make the JSDoc comments really noisy without providing any additional benefit.

> It might help if I better understood the specific use cases here. Is the main benefit improved LSP completions?

The main reason for implementing parsing of JSDoc comments was IMO to make them stand-out and improve readability. It's easier to parse them visually when there is some mild highlighting of special keywords and structure is easy to see at a glance rather than just having a gray block of documentation text.

That said, we don't want them to stand out too much as then they would look too much like the actual code, blurring the separation of concerns.

So really the highlighting of **types** implemented in this PR is to make them stand out by applying a single scope (color). Nothing more, nothing less. The special handling of interpunction is a little extra to trigger completions and improve the experience with LSP (without LSP possibly too). It was never the intention (and IMO shouldn't be) to parse them in the same way as we do for TS types (or worse, implementing parsing of Closure types whose popularity is pale comparing to TS).

> Are there any simple examples of annotations which look bad with the current highlighting?

Well, without this PR JSDoc types look like a plain comment so I wouldn't call it "bad" but it's lacking the semantic highlighting. :)

rchl:
Can we decide on something here?

I'm running with those changes since I've made them and I can't complain about anything. Nothing annoys me about it and documentation blocks are more readable IMO.

Since the highlighting of tags was already implemented and merged, I think it would be weird to stop now. At least if the argument is that this is not a standard language feature.

Also, various other language syntaxes (PHP for example) also have support for highlighting the documentation comments and those are not always an integral part of those languages either.

Thom1729:
This should be fine to merge. Even if parsing JSDoc seems messy to me in principle, this PR itself doesn't make it any messier.

Hopefully one of these days I'll get around to complete JSDoc support in JS Custom, and that might help to illuminate the issue, at least from my own perspective.

rchl:
Pushed a small refactor to avoid scoping leading `*` when jsdoc types span multiple lines.

Before | After
-------|------
Screenshot 2021-03-24 at 21 19 41 |  Screenshot 2021-03-24 at 21 19 50

This unfortunately required a bit of ugly solution with negative lookahead that excludes all cases that can be matched and consumed within the type expression.
Couldn't use `meta_scope` anymore as then I can't remove it when leading `*` is matched.
Couldn't just do a simple `.` match as then ligatures wouldn't work within type expressions.

Thom1729:
Would it work to handle the leading asterisk with a `with_prototype` rule?

Also, mostly unrelated to this PR, but would it make sense to move the JSDoc implementation out into its own file, like JS regular expressions?

michaelblyons:
> Also, mostly unrelated to this PR, but would it make sense to move the JSDoc implementation out into its own file, like JS regular expressions?

I believe this is also the way JavaDoc is set up, both currently and in DeathAxe's huge rewrite (#2654).

rchl:
I can extract the code into a separate file once/if this gets merged so that the changes done here are easy to diff.

rchl:
Force-pushed since I inadvertently did a rebase locally instead of merging.

Still not clear if there is even a will to get it in but will keep it open until some concrete decision is made.

I will just say that this is something that is provided by VSCode so not having it can be seen as a disadvantage to some.

rchl:
Since there doesn't seem to be a consensus whether this should be merged (not sure why), maybe you would accept this in JSCustom as an option @Thom1729 ?

rchl:
Not that I have hope that this will get merged but since I'm using it myself, I'll keep pushing my changes here until someone closes it :)

I've changed the scope for type annotations from `support.type` to `meta.type` to match the other Python PR.

With `meta.type` there is no coloring (in Mariana at least) so maybe that's less controversial. It still always the user to apply own styling with a custom color scheme rule. And there is still added benefit of scoping dot as punctuation to allow completions to trigger.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -106,6 +106,8 @@ variables:
 
   # '@' followed by a pattern like \S but excluding literal '*' and '@'.
   jsdoc_block_tag: \@[^\n\t\f\v *@]+
+  # Tags that are optionally followed by a type expression.
+  jsdoc_block_tag_with_type: \@(?:arg(?:ument|)|param|type(?:def|)|returns?|prop(?:erty|)|enum|implements|member|module|namespace|(?:pr(?:ivate|otected))|template|this|throws|exception|yields?|var)\b
```
keith-hall, line 110: Generally literals are very easy for regex engines to optimize themselves, so we tend to make them more readable and searchable in this repository :)
Also a small style thing, empty alternations can be confusing (i.e. maybe something is missing) but with optional quantifiers the intentions are clear
```suggestion
  jsdoc_block_tag_with_type: \@(?:arg(?:ument)?|param|typedef|type|returns?|prop(?:erty)?|enum|implements|member|module|namespace|private|protected|template|this|throws|exception|yields?|var)\b
```

JavaScript/JavaScript.sublime-syntax
```diff
@@ -164,12 +166,46 @@ contexts:
...
+      set:
+        - match: '\}'
+          scope: punctuation.definition.type.end.js, meta.tag.jsdoc
+          pop: 1
+        - include: jsdoc-type-expression-body
+
+  jsdoc-type-expression-body:
+    - match: (?=\*/)
+      pop: 1
+    - match: '^\s*\*'
```
keith-hall, line 194: A while ago, we started scoping the leading asterisk in `/**` style comments because some ST functionality was updated to handle it.
I'm not 100% sure this is the right scope but I'm on mobile and with limited time ATM, and figured a hint and vague review could be better than none...
```suggestion
    - match: '^\s*(\*)'
      captures:
        1: punctuation.definition.comment.js
```

JavaScript/JavaScript.sublime-syntax
```diff
@@ -164,12 +166,46 @@ contexts:
...
+    - match: \.
+      scope: meta.type.js punctuation.accessor.js
```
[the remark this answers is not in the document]
rchl: Looks like the Javascript syntax in particular doesn't include `dot` in those scopes. But I guess it makes sense given that majority does.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -164,12 +166,46 @@ contexts:
...
+    - match: '\{'
+      scope: meta.type.js
```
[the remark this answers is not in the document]
rchl: `punctuation.section.block` or `punctuation.definition.block`?

And also, this would apply a distinct color. At least in Mariana the first would become white and the second blue. So it would be a bit in conflict with my latest change.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -164,12 +166,46 @@ contexts:
...
       pop: 1
     - match: (?=\S)|$
       pop: 1
 
+  jsdoc-type-expression:
+    - match: '\{'
+      scope: punctuation.definition.type.begin.js, meta.tag.jsdoc
+      set:
+        - match: '\}'
+          scope: punctuation.definition.type.end.js, meta.tag.jsdoc
```
keith-hall, line 187: meta scopes come before punctuation, and no need to have a comma after `.js`
```suggestion
          scope: meta.tag.jsdoc punctuation.definition.type.end.js
```

JavaScript/JavaScript.sublime-syntax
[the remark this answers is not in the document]
keith-hall: but aren't you already scoping `{` and `}` in type comments? https://github.com/sublimehq/Packages/pull/2629/files#diff-04a9b98e8498a3df8d16447d91d8c3cb9131fa4359f113b2917651aa0ed77c23R183-R184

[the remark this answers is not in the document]
keith-hall: I would guess `section` as I don't really see `definition` listed on http://www.sublimetext.com/docs/scope_naming.html#punctuation regarding curly braces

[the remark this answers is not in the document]
rchl: > but aren't you already scoping `{` and `}` in type comments? [#2629 (files)](https://github.com/sublimehq/Packages/pull/2629/files#diff-04a9b98e8498a3df8d16447d91d8c3cb9131fa4359f113b2917651aa0ed77c23R183-R184)

Yes but with the same scope as the rest of the type contents.

The outer braces are scoped distinctly but those can be see as part of the jsdoc tag syntax rather than the type itself.

Screenshot 2023-05-14 at 11 07 08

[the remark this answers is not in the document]
keith-hall: ah I see, thanks for the explanation. Then feel free to ignore this suggestion :)

Issue #2630, opened 2020-11-21
Decimal point scope in some of the recent PRs for number scopes is wrong


jwortmann:
Looks like there was a little mistake in a few of the recent PRs to floating point number scopes. The `punctuation.decimal.separator` scope for the decimal point should be applied *on top* of `constant.numeric.value` like this:
```python
  12.34
# ^^^^^ constant.numeric.value
#   ^ punctuation.separator.decimal
```
But currently the scopes are alternating between `constant.numeric.value` / `punctuation.separator.decimal` / `constant.numeric.value` in a few syntaxes. This way a single number would be considered as two separate tokens, e.g. if a plugin searches for `constant.numeric`, and also the highlighting is different than before in many color schemes.

The corresponding PRs are: #2498, #2499, #2500, #2501, #2512, #2513. #2514

Should be an easy fix and some regexes could even be simplified by it.

PR #2641, opened 2020-11-26
[Go] Fix decimal point scope in floats

jwortmann:
As described in #2630.


jwortmann:
I've also removed the special scoping for the exponent part in float numbers here.

wbond:
Thanks!

PR #2642, opened 2020-11-27
[D] Fix decimal point scope in floats

deathaxe:
As described in #2630.

Note: I've also removed `exponent` scope to reduce capture groups and because they are not implemented in the majority of syntaxes.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -194,7 +194,7 @@ contexts:
     - include: integer-opt
     - include: not-whitespace-illegal-pop
   integer-opt:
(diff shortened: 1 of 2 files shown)


jwortmann:
The Go syntax and the other mentioned syntaxes also have the "exponent" scope. I don't think it's really necessary or useful and could remove it in #2641 as well, but I'm not sure if there are users who maybe want to use this scope in their color schemes. Any opinions about that?

deathaxe:
IIRC, the conclusion was not to add it for simplicity reasons. I think it's safe to remove it for exactly that reason as those changes are young enough for them not to be widely used.

wbond:
Thanks!

PR #2646, opened 2020-12-06
[C] Improve scoping of function parameters identifier

TheClams:
Thanks to branching it is now possible to properly recognize a function
 parameter identifier even when there is a line-return or comment before
 the comma or closing parenthesis.

Applying the same fix to C++ requires a lot more change like changing the 
shared context expressions-minus-generic-type which likely have more implication

### DIFF
--- a/ASP/HTML (ASP).sublime-syntax
+++ b/ASP/HTML (ASP).sublime-syntax
@@ -13,8 +13,14 @@ file_extensions:
 
 variables:
 
+  jscript_lang: |-
+    (?ix: js | jscript )
+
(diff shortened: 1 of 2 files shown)


deathaxe:
Please note such changes would need to be applied to all C variants. I see requirement for Objective-C at least. How about C++?

TheClams:
Yes C should be fix as well, but the fix seems a lot more complex: it
implies changing other included scope, and I thought it might worth waiting
for the rework through inheritance before going through this rabbit hole

PR #2647, opened 2020-12-06
[Go] use the `keyword.declaration` convention

mitranim:
Changed every occurrence of `storage.type.keyword` to `keyword.declaration` to accommodate this newer convention.

Additionally, corrected the scope order "non-meta meta" to "meta non-meta" for `go:` annotations. (Meta scopes usually go first.)

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -126,7 +126,7 @@ contexts:
         3: meta.punctuation.accessor.colon.go
         4: meta.variable.function.go
       push:
-        - meta_scope: comment.line.go meta.annotation.go
+        - meta_scope: meta.annotation.go comment.line.go
         - match: \S+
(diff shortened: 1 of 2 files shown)


Go/Go.sublime-syntax
```diff
@@ -705,7 +721,7 @@ contexts:
 
   match-keyword-interface:
     - match: (?=\binterface\b)
-      scope: storage.type.keyword.interface.go
+      scope: keyword.declaration.interface.go
```
deathaxe, line 724: Assigning a scope to a lookahead has no effect.

mitranim: Nice catch, thanks. Made a correction.

deathaxe (approved):

wbond:
Thanks @mitranim!

PR #2648, opened 2020-12-06
[Ruby] Fix Constant vs. Class Syntax Definition

### LINKED ISSUE #506: [Ruby] Incorrect scope of nested constants used outside of class definition contexts

### LINKED ISSUE #2410: [Ruby] Wrong scope name for nested classes
- Sublime Version: 3.2.2 build 3211
- OS Version: MacOS Catalina 10.15.6

``` ruby
SomeClass::Nested
```

image

I'd expect the above to have `support.class.ruby`, the same as `SomeClass`

danillos:
When a Class is defined in Ruby it is assigned to a Constant, so every Class is a Constant.

This PR make them behavior as the same and improves the Match regex.

Close #2410
Close #506


deathaxe:
I am not sure if it is desirable to have both `support.class` and `variable.other.constant` stacked to each class object. Classes and constants are fundamentally different things. That's why applying both of those scopes to the same token seems odd. From highlighting perspective it doesn't make much sense, too, as most color schemes provide rules for `variable.other` which cause such tokens to not be highlighted as classes anymore.

deathaxe:
The test cases don't reflect the intended issue which is to be fixed. The added test case succeeds without your syntax changes as well.

danillos:
> I am not sure if it is desirable to have both `support.class` and `variable.other.constant` stacked to each class object. Classes and constants are fundamentally different things. That's why applying both of those scopes to the same token seems odd. From highlighting perspective it doesn't make much sense, too, as most color schemes provide rules for `variable.other` which cause such tokens to not be highlighted as classes anymore.

That is the question, in Ruby they are the same because Class are defined in Constants.

danillos:
> The test cases don't reflect the intended issue which is to be fixed. The added test case succeeds without your syntax changes as well.

Thanks, I'll double-check it.

timfjord:
I also faced this problem, and indeed it is correctly highlighted when there is a method call after the constant(like in the test case)
But when there is no method call(e.g. `include MyModule::MySubmodule`), the last part isn't correctly highlighted

danillos:
@deathaxe I added a new test and updated it based on master.

> Classes and constants are fundamentally different things. That's why applying both of those scopes to the same token seems odd.

About it, [Ruby classes are Constants](https://www.rubyguides.com/2017/07/ruby-constants/) so makes sense to them be defined as the same, but I don't know how bad is it on SublimeText Syntax Definition.

CleanShot 2021-06-27 at 19 38 34

deathaxe:
So you would also scope right hand side of `f = MyConstScopedAsClass` as class? I don't think everybody will be happy about it.

Still stacking `variable.other` onto `support.class` doesn't change highlighting for default nor most 3rd-party themes.

I'd at least try to distinguish those a bit more.

deathaxe:
See suggestion at https://github.com/danillos/Packages/pull/1

danillos:
> So you would also scope right hand side of `f = MyConstScopedAsClass` as class? I don't think everybody will be happy about it.
> 
> Still stacking `variable.other` onto `support.class` doesn't change highlighting for default nor most 3rd-party themes.
> 
> I'd at least try to distinguish those a bit more.

Since Class and CONSTANT are the same in Ruby, and we still want to have some difference between then on SublimeText, and I agree with it, my suggestion is to use [Ruby Style Guide](https://rubystyle.guide/#screaming-snake-case), where CONSTANTs definitions are all in `UPPER_CASE` and Classes in `CamelCase`.

```
MyClass     = Class
MY_CONSTANT = Constant
```

What do you think @deathaxe ?


CleanShot 2021-06-29 at 15 45 05

timfjord:
👍🏻 on distinguish constants and classes so they can be highlighted differently(like GitHub does) 
but I think in this case `f = MyModule::MYCLASS.new`, `MYCLASS` should be a `variable.other.constant.ruby` and not a `support.class.ruby` (again, similar to GitHub)

danillos:
Thanks, @deathaxe, I just merged your PR changes.

deathaxe (approved):

jfcherng (approved):

PR #2650, opened 2020-12-12
[Matlab] Better syntax highlighting

jwortmann:
The current Matlab syntax isn't in the best condition in my opinion could need some improvements:

- some language features are just not implemented and not highlighted
- various scopes don't follow the scope naming guidelines; for example there are a few "uncommon" creations like `keyword.mathematics`, `keyword.operator.symbols` or `storage.control`
- there are almost 700 built-in function names scoped as `keyword`, which results in poor highlighting and makes it difficult to distinguish actual keywords from built-in functions
- syntax tests are very sparse
- the overall style and implementation of the syntax is a bit unstructured, not very readable and contains some duplication

I thought to give it a comprehensive rework and make a PR so that other users can benefit from better highlighting. Most parts of the syntax file were reworked, but all changes are split into seperate, self-contained commits and they have explanative commit messages.

Things that were changed include:

- Add support for more language features:
  - support for more than one superclass in [class definitions](https://www.mathworks.com/help/matlab/matlab_oop/subclass-syntax.html)
  - support for the [! operator](https://www.mathworks.com/help/matlab/matlab_external/run-external-commands-scripts-and-programs.html)
  - support for the [? operator](https://www.mathworks.com/help/matlab/ref/metaclass.html)
  - support for the [.? operator](https://www.mathworks.com/help/matlab/matlab_prog/function-argument-validation-1.html#mw_de9b2056-9dc2-421c-9c7e-d2c4400a95be) in `arguments` blocks
  - support for the [@ operator](https://www.mathworks.com/help/matlab/ref/function_handle.html)
  - support for [function argument validation](https://www.mathworks.com/help/matlab/ref/arguments.html) blocks inside of function definitions

- Add meta scopes to blocks (function/class declarations, flow control blocks)

- Add the correct scopes to punctuation characters which had a wrong scope or no scope before:
  - `punctuation.terminator` for semicolon when used as a statement terminator
  - `punctuation.separator` for comma & semicolon in arrays and comma between function arguments
  - proper scopes for curly braces
  - ~~`variable.language.blank`~~ `variable.language.anonymous` for the `~` character (argument placeholder), except for when it is used as negation operator
    ~~*Note:* The `blank` in this scope name comes from what's currently used in the Go syntax; another alternative for it could be `variable.language.wildcard`~~
  - `punctuation.accessor.dot` for all dots when accessing fields in structure arrays and also for trailing dots to trigger autocompletions

- Change some other scopes to comply with the scope naming guidelines:
  - change the scope for built-in functions from various `keyword` scopes to `support.function.builtin` (but no further subscopes here like they had under `keyword`, because lots of functions are overloaded/used in various "parts" of the language, so it's impossible to distinguish them and choose the "correct" subscope for it)
    *Note:* ~~I still kept the `keyword` scope for a few functions which are usually only used in [command syntax](https://www.mathworks.com/help/matlab/matlab_prog/command-vs-function-syntax.html) form and take predefined arguments. However, I used a uniform scope `keyword.other.command` for them now. Examples are `import` or `tic`/`toc`. Maybe it's useful to keep the `keyword` scope for those, so that they stand out against the various other built-in functions which are mostly used in function syntax form. But I would be open to change all their scopes to `support.function` if that is preferred, because that is what they actually are.~~ Keep in mind though, that almost everthing in Matlab is implemented as a function, including "constants" that are mostly used without input parameters (e.g. `true`/`false`, `pi`, `Inf`). The Matlab documentation seems to even list [control flow keywords](https://www.mathworks.com/help/matlab/referencelist.html?type=function&category=control-flow) like `if`, `for`, `while`, `return` under functions.
    Likewise I kept the `storage.type` scope for some type-conversion functions like `double`, `int32`, `string`, but only when they  appear without parentheses now, e.g. in function argument validation blocks.
  - use `variable.language` scope for built-in variables (`ans`, `nargin`, `nargout`, `varargin`, `varargout`) instead of the `variable.other.function` or `constant.language` scopes used before

- Update the list of built-in functions to the most recent Matlab version (R2020b) from the Matlab docs
    *Note:* I deleted the additional functions from Matlab toolboxes (those require to purchase separate licenses to use), because they were very incomplete/outdated

- A few small fixes, e.g.
  - floating point numbers with trailing dots
  - better detection of the transpose operator
  - nested block comments
  - escaped and placeholder characters in strings

~~Something I didn't change yet is the scope `meta.variabel.other.valid.matlab` which is used for ordinary identifiers (variables and function calls), but I think this should either be changed to `variable.other.matlab` or shortened to `meta.variable.matlab`. This would of course break backward compatibility, but the former uncommon meta scope probably isn't targeted by color schemes anyway. Any opinions about this?~~
*Edit*: Scopes for identifiers (variables and function calls) were changed from `meta.variabel.other.valid.matlab` to `variable.other.matlab`.

Another thing I was wondering about is why the syntax file lists (only) `matlab` as the file extension for Matlab files. The actual file extension is `m` and I have never seen a file with a `matlab` extension. Was this one somehow arbitrarily chosen because Objective-C also has `m` as a file extension? If there is no special reason against it, the Matlab extension should be changed to `m` as well.

It would be nice if the Matlab syntax can get enhanced with this PR. Feel free to test it and to provide feedback if you find something that could be improved.


Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
+    - match: \bif\b
+      scope: keyword.control.conditional.if.matlab
+      push:
+        - meta_scope: meta.block.if.matlab
+        - match: \belseif\b
+          scope: keyword.control.conditional.elseif.matlab
+        - match: \belse\b
+          scope: keyword.control.conditional.else.matlab
+        - match: \bend\b
+          scope: keyword.control.end.if.matlab
```
deathaxe, line 502: With regards to the former comment about `end` this is an example of what we used `keyword.control.conditional.end` for in some other languages.

Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
+        - include: expressions
+
+  while:
+    - match: \bwhile\b
+      scope: keyword.control.loop.while.matlab
       push:
-        - meta_scope: string.quoted.double.matlab
-        - match: '"'
-          scope: punctuation.definition.string.end.matlab
+        - meta_scope: meta.block.while.matlab
```
deathaxe, line 548: These kinds of use cases of sub-scoping `meta.block` are quite interesting. While this style is found in older syntaxes (e.g.AppleScript) most recently modified syntaxes seem to just use `meta.block.<syntax>` without sub-scoping. 

They rather use `meta.while meta.block`, maybe because the loop condition `while (condition)` doesn't really belong to the block, but may rather be interpreted as some kind of argument for the while keyword. The block is only one of several parts of such a statement. While it is easy to apply for syntaxes which use `{` `}` to denote blocks, I am a bit uncertain about the correct style with regards to syntaxes like matlab or ruby, ... which use keywords only.

I am also not very certain about `meta.while meta.block` is the best solution. I use it in the rewritten Java, but it quite quickly fills the scope stack because of each level of statement adding 2 meta scopes.

Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
+    - match: \b(who|whos)(?:\s+(global))?\b
+      captures:
+        1: keyword.other.command.matlab
+        2: storage.modifier.matlab
+
+  # Parentheses are usually omitted when functions are used without parameters.
+  # This command syntax is also possible for functions with string parameters.
+  # Only the function names are matched here to allow both command syntax and function syntax form.
+  # https://www.mathworks.com/help/matlab/matlab_prog/command-vs-function-syntax.html
+  builtin-functions:
```
deathaxe, line 801: Maybe combining all the different lists of builtin functions and classes in the `variables` section may make sense?

To illustrate what I have in mind:

https://github.com/sublimehq/Packages/blob/4b66388fd5b206357769a9650c163f1a6958cea6/CSS/CSS.sublime-syntax#L55-L72

and than just use:

https://github.com/sublimehq/Packages/blob/4b66388fd5b206357769a9650c163f1a6958cea6/CSS/CSS.sublime-syntax#L2536-L2549

Another example may be the different HTML tag types in CSS.

deathaxe (commented):
I like this rewrite. It is clean and very well structured. The one way or the other, it is a great step forward for Matlab, IMHO.

I've made some suggestions about things which came to my mind reading the syntax, which I just found worth discussing about. Some of them are not directly related with this PR but are of a more general sense,  which may touch other syntaxes as well.

Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
+      scope: storage.type.matlab
+
+  last-index:
+    - match: \bend\b
+      scope: variable.language.matlab
+      push: transpose
+
+  argument-placeholder:
+    - match: \~(?![A-Za-z])
+      scope: variable.language.blank.matlab
```
deathaxe, line 669: It appears several languages such as Python, Java or Scala have such kinds of special variables to denote unrequired or arbritary variables. I'd find it suitable to have a common scope for all of them. While `blank` fits quite well for that use case, I wonder whether `anonymous` may be an alternative. The reason for it is scoping lambda functions `anonymous` as well, so reusing this subscope name for unnamed variables seems logical.

Thoughts?

jwortmann:
Thanks for the suggestions, they are really helpful! I have addressed most of them for now.

Matlab/Matlab.sublime-syntax
```diff
@@ -1,36 +1,61 @@
...
+##############################################################################
+
 contexts:
+
   main:
-    - include: function
-    - include: constants_override
-    - include: brackets
-    - include: curlybrackets
+    - include: function-declaration
```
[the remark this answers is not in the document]
jwortmann: Seems like a good idea. Done in 639bc95.

Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
+      captures:
+        1: punctuation.separator.decimal.matlab
+        2: punctuation.separator.decimal.matlab
+      push: transpose
+
+  string:
+    - match: \'
+      scope: punctuation.definition.string.begin.matlab
+      push:
+        - meta_scope: string.quoted.single.matlab
```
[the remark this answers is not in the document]
jwortmann: Added `meta.string` in 74f2158.

Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
+    - match: \%\%|\\\\|\\[abfnrtv]
+      scope: constant.character.escape.matlab
+
+  format-spec:
+    - match: \%(?:\d\$)?[-+\s0#]?(?:\d+|\*)?(?:\.\d+|\.\*)?(?:[cdeEfgGis]|[bt]?[ouxX])
+      scope: constant.other.placeholder.matlab
+
+  variable:
+    - match: \b{{identifier}}\b
+      scope: meta.variable.other.valid.matlab
```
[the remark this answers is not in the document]
jwortmann: I preferred `variable.other.matlab` as well. I adjusted it in b57bc48.

Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
+                  set: class-body
+                - match: '{{identifier}}'
+                  scope: entity.other.inherited-class.matlab
+                - match: \&
+                  scope: punctuation.separator.sequence.matlab
+
+  class-body:
+    - meta_scope: meta.class.matlab
+    - match: \bend\b
+      scope: keyword.control.end.class.matlab
```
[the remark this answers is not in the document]
jwortmann: I think I first scoped them with `keyword.control.end.<statement>.matlab`, because the single rule for `end` which was used before in this syntax assigned `keyword.control`, and with this it would still be easy for color schemes to target all `end` keywords with a single scope. But after thinking a bit about your argument, I agree that it probably is an advantage to have it scoped similar to the "opening" keyword.

I went with your second option now, i.e. `keyword.declaration.class.end`, because it is consistent with `keyword.control.{loop|conditional}.end` to have the "end" part of the scope at the 4th level, and because if a color scheme wants to use a special color for all `end` keywords, it needs to target more than only a single scope anyway now.

Matlab/Matlab.sublime-syntax
jwortmann: There was a question/discussion for this in https://github.com/SublimeText/ScopeNamingGuidelines/issues/5, and I mentioned that `variable.language.blank` from the Go syntax might be a good solution. A search on this repo revealed that you used `variable.language.anonymous` in Erlang, so I adjusted this scope here in 12a1469. Would definitely be useful to have a common convention for it and unified scopes in the default syntaxes.

Maybe worth to mention that in the [Matlab docs](https://www.mathworks.com/help/matlab/matlab_prog/ignore-function-outputs.html) it is called "tilde operator", but I don't find that wording a good choice, because `~` can also be used as the negation operator.

Matlab/Matlab.sublime-syntax
jwortmann: I assume the restriction for `meta.block` to `{ ... }` was due to it beeing one of the scope recommendations in https://www.sublimetext.com/docs/scope_naming.html#meta.

Some languages use an additional `begin` keyword, e.g.
```julia
while condition begin
   ...
end
```
so it would be analogous here to have `meta.block` from (and including) `begin` to `end`. But I don't really think there is a big advantage to have separate `meta.while meta.block` scopes (at least in Matlab which doesn't use a `begin` keyword). And I when I say "block" here, I would probably mean the whole statement, including `while` keyword and condition. But it's probably a matter of interpretation.

Anyway, I would suggest to keep a single `meta.block.while` here.

Matlab/Matlab.sublime-syntax
```diff
@@ -48,477 +76,792 @@ contexts:
...
             - meta_scope: meta.function.parameters.matlab
-            - match: \b\w+\b
-              scope: variable.parameter.input.function.matlab
-            - match: ','
-              scope: punctuation.separator.matlab
-            - match: '\)'
+            - include: line-continuation
+            - match: '{{eol}}'
+              set: function-body-arguments
+            - match: \(
```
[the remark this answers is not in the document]
jwortmann: I'm not exactly sure why I used that lookahead here, maybe because it was necessary for correct boundaries of the meta scopes. I'll check it soon and look whether I can adjust it.

Matlab/Matlab.sublime-syntax
jwortmann: I'm not sure whether it makes sense here to have all function names defined as a variable, because they are used only once in this syntax.

The classification represents the function lists from the Matlab docs (left side menu at https://www.mathworks.com/help/matlab/referencelist.html?type=function&listtype=cat&category=index&blocktype=all&capability=&s_tid=CRUX_lftnav).
I hope that I've removed all duplicates and "non-functions". To merge all lists would probably make it easier to find duplicates.

Matlab/Matlab.sublime-syntax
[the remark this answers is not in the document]
jwortmann: I've cleaned up the rules for function declarations a bit and eliminated the lookahead. There should be no nested parens for function parameters, but correct balancing will be ensured now, like for the other parens/brackets expressions.

jwortmann:
I just made two additional adjustments:

There are several built-in functions which are usually used in "command syntax form" and previously I kept them scoped as `keyword`, as explained in the opening comment of this PR. But now I think this isn't the best scope name for them, because they are not really keywords in Matlab and it isn't guaranteed that these identifiers are not used in normal function syntax. So I decided to change the scope to `support.function` now, like all other built-in functions are scoped. One example which would probably be a keyword in most other languages is `import`. Parameters for functions in command syntax form are always strings, but there are a handful of functions that have some predefined parameters with a special meaning, e.g. `clear all`, and I've kept `support.constant` for these special parameters.

In general, command syntax form can be used for all functions in case all of their parameters are strings (or if they don't have any parameters), see https://www.mathworks.com/help/matlab/matlab_prog/command-vs-function-syntax.html. I'm still not entirely sure what the best approach for scoping functions and parameters in command syntax form is, and I'd be happy to hear opinions about it.

I also took a look at how command syntax functions are handled in the Matlab package for VS Code: they use `variable.other.command` for all function names (imo not optimal) and `string` for all parameters (which is correct). However, this means that built-in functions are not recognized if they appear in command syntax form, and there is no highlighting for special parameters.

vscode

And the implementation for it seems very complicated; I think I have never before seen such a [regex pattern monster](https://github.com/mathworks/MATLAB-Language-grammar/blob/40d9a0cd3b628f80cdcf948bbe1747a527ed5dd5/Matlab.tmbundle/Syntaxes/MATLAB.tmLanguage#L931) like it is used there...

***

The second change is a fix for a rule which is present in the current Matlab syntax and which I haven't touched in this rewrite, with the intention to highlight unescaped quotes within strings as `invalid.illegal`:
```matlab
 'a'a'
%  ^ string.quoted.single.matlab invalid.illegal.unescaped-quote.matlab
```
(The highlighting on GitHub is similar because it uses the same TextMate syntax definition for linguist, which seems to has been used as a basis for the ST syntax as well.)

But the way this was implemented here can lead to false highlighting of valid code, for example `'a'%a` is a valid string followed by a comment. Also this highlighting is different from how the Matlab IDE interprets it; Matlab marks only the trailing part after the 2nd quote in the `'a'a'` example as error. So even though it could be fixed, I think it's safe to remove this error highlighting here completely, especially because you can easily see already from the following highlighting if an unescaped quote was not intended to end a string.

deathaxe:
> ... functions are handled in the Matlab package for VS Code

I'd be careful with using VS Code as reference for scope names. My experience or impression is that many syntaxes are just converted from very old TextMate syntaxes. Even those which have been updated sometimes still do.

> In general, command syntax form can be used for all functions in case all of their parameters are strings ...

This answers the question, how to scope it correctly, right? I think it's the correct step. Bash also scopes all builtin commands `support.function` and others `variable.function` as this is what most of them are - functions/procedures.

I'd probably use only one scope for known predefined values though - either `string` or `support.constant`. The reasons are

1. `string` scope disables auto-completions, which might be a nice feature for predefined values either by static file, plugin or LSP.
2. IMHO we should try to avoid stacking two non-meta scopes wherever possible. They are highlighted `support.constant` anyway. If a color scheme doesn't address the latter scope it wouldn't highlight such constants in other standalone use cases, too.
3. Several syntaxes use `meta.string meta.interpolation` for string interpolation. We therefore should stack `meta.string string. ...` in all syntaxes in case color schemes or plugins want o make use of it somehow. So we already have 2 scopes stacked per string, at least.

> The second change is a fix for a rule which is present in the current Matlab syntax 

As illegal code is a common use case during authoring I think we should use `illegal` scope with care and only  if we are sure it helps illustrating syntax errors. When in doupt we should ommit it. 

The more I work with certain syntaxes of PackageDev the more I hate illegal highlighting because many incomplete or temporarily wrong code blocks are highlighted in red. It doesn't help at all.

To illustrate what you already stated ...

grafik

The end of the string is clearly visible and that's very much ok.

Matlab/Matlab.sublime-syntax
deathaxe: The variables approach in general has following motivations:

1. avoid those endless lines and replace them by hard wrapped code blocks of fixed width
2. variables are not indented and therefore can use most available space to achieve (1.)
3. variables may help inerhited syntaxes a bit to replace or extend certain kinds of variables. It would require to group commands from different libraries in dedicated variables though. This is what the "units" example should illustrate. The html tags variables in CSS may be a better one though.

All of those are experiences I've made. They are of rather low priority for Matlab as it is quite unlikely it being extended or inherited.

jwortmann: Okay, I've merged most of the function lists and moved them to the variables section.

Your third point might actually be quite useful even for Matlab; there is the Octave software, which aims to be a free and open source replacement for Matlab, with mostly compatible syntax. Not all Matlab functions are implemented in Octave though, and there are a few other extension. For example in the "Indentation Rules.tmPreferences" file in this Matlab package, there are patterns for `unwind_protect` and for `end\w*` (instead of just `end`), which are Octave-specific and have no meaning in Matlab. Perhaps they should better be removed here. Some snippets seem to be targeting Octave as well. There is also a [OcataveSupport](https://packagecontrol.io/packages/OctaveSupport) package on Package Control with its own syntax, but it seems to be just a copy of this Matlab syntax from an earlier state with adjusted header.

 And to the indentation patterns file for Matlab should be added `parfor`, `spmd` and `arguments`, but maybe that's another PR.

deathaxe: I guess the solution depends on two factors:

1. How easy is it to select correct syntax for an end user? As Octave makes use of the same file extensions it might require active setup or interaction to choose correct dialect.
2. Whether or not a dialect such as Octave is interpreted as new syntax and/or is accepted to be added to default packages.

Basically a syntax should keep focused and precise. On the other hand end users might expect "Octave" support if those syntaxes are mostly compatible. Python for instance also supports both 2 and 3 which have slightly different lists of builtins.

If there are no unsolvable conflicts and differences are builtin functions only, it might be the simples solution to just add those additional Octave functions to Matlab.

An alternative might be to add an Octave syntax as inherited one, if sublimehq is interested. I have an impression of trying to keep syntaxes which "inherite each other" together might be the most robust solution.

jwortmann: I have only used Matlab and not Octave, but according to the [Octave Wiki](https://wiki.octave.org/wiki/index.php?title=Differences_between_Octave_and_Matlab#What_features_are_unique_to_Octave.3F) there seem to be quite some differences which would need changes or additions to this syntax. These are not only built-in functions, but also various new operators, a few more keywords and block-constructs, and another character for comments. And I think none of the Octave-specific syntax is currently recognized by this Matlab syntax (except for the .tmPreferences file and the snippets), so it's probably better to leave it to the Octave package on Package Control.

jwortmann:
> I'd probably use only one scope for known predefined values though - either string or support.constant.

The `string`/`support.constant` scopes were adjusted to avoid stacking of more than one non-meta scope, but still give a hint that arguments are unquoted strings (also see the explanation below).

I very much agree regarding `invalid.illegal` scopes, and the only rule left from the current syntax with this is for invalid variable names which start with a number or underscore. I could remove this as well if desired, but I think this is safe to never produce false positives for actually valid syntax, so I kept it for now.

I would consider the following points of the current state of this syntax rewrite as not perfect, but I believe there is probably no good way to improve them:

1. There is a single lookbehind used in this syntax ([here](https://github.com/sublimehq/Packages/pull/2650/files#diff-721e75fbe18ecb70c64982e61a40c8046331d61e98421a389ca03270ceab9cf7R1133)) to ensure the greedily scoping of trailing dots as `punctuation.accessor`, and didn't find a way to eliminate it.
2. Even though not recommended, under certain conditions it is allowed to omit the `end` statement to terminate a function declaration ([docs](https://www.mathworks.com/help/matlab/matlab_prog/create-functions-in-files.html#bvf7wat)). This will lead to an accumulation of `meta.function` scopes, if a function file contains many functions without `end` statement. It has no implications for syntax highlighting, and it should occur very rarely or probably only for old files. But here is an example file I found, where this happens: [link](https://github.com/petercorke/robotics-toolbox-matlab/blob/master/private/arrow.m)
3. If a function is called in [command syntax](https://www.mathworks.com/help/matlab/matlab_prog/command-vs-function-syntax.html), its arguments are unquoted strings (actually: "character arrays"), but - in contrast to the Matlab IDE - not highlighted as such. For example in `foo bar baz`, `foo` is the function name and `bar` and `baz` are string arguments. For a handful of commonly used functions with predefined arguments, I added some special rules so that these arguments are recognized and highlighted accordingly, and I added the `meta.string` scope in that case (see [syntax tests](https://github.com/sublimehq/Packages/pull/2650/files#diff-c7f9d4a8068f24fb7f10f13b43513856d7dbd36ec6eed611eadc5b98b491fa40R925-R934)). But I think in general it's impossible to always apply the `string` scope for the arguments, because the same syntax can also occur e.g. within an array (`[foo bar baz]`), and in that case these identifiers are just normal variables.
4. I wasn't really sure which sub-scope to use for [a few operators](https://github.com/sublimehq/Packages/pull/2650/files#diff-721e75fbe18ecb70c64982e61a40c8046331d61e98421a389ca03270ceab9cf7R1097-R1108), because they don't fit into one of the common categories, so I chose some new names for them based on specific meaning or how they are named in the Matlab docs.

A few more general information for purpose of clarification:

* The scope `variable.function` isn't used in this syntax at all, because it's impossible with a regex-based highlighting engine to distinguish between function calls and array access in Matlab (both use the exact same syntax and are scoped `variable.other` now).
* There is no highlighting for built-in function names if immediately followed by a transpose operator (see [syntax tests](https://github.com/sublimehq/Packages/pull/2650/files#diff-c7f9d4a8068f24fb7f10f13b43513856d7dbd36ec6eed611eadc5b98b491fa40R767-R772)), because there's probably no function where a transpose would make sense, so it can be assumed that the identifier was redefined as a variable name. This behavior was not changed from the current syntax definition.
* Btw, if anyone wants to try out some code snippets without access to the Matlab software, you can use the Live Editor from the Matlab docs, e.g. [here](https://www.mathworks.com/help/matlab/learn_matlab/matrices-and-arrays.html), click on "Try This Example" and then add arbitrary code.

I think the only question left is where the current file extension `matlab` comes from and whether I should change it to `m`.
And perhaps whether I should add the missing keywords to the indentation rules .tmPreferences file in this PR and remove the Octave-specific ones.

Matlab/Matlab.sublime-syntax
```diff
@@ -1,36 +1,61 @@
 %YAML 1.2
 ---
-# http://www.sublimetext.com/docs/3/syntax.html
+# https://www.sublimetext.com/docs/syntax.html
+# https://www.mathworks.com/help/matlab/index.html
 name: MATLAB
+scope: source.matlab
```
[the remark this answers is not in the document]
jwortmann: I've updated the syntax in this PR to use version 2 now.

This could indeed be useful to allow potential inheritance for Octave, or if someone e.g. wants to add highlighting for Matlab toolbox functions.

Matlab/Matlab.sublime-syntax
```diff
@@ -1,36 +1,572 @@
...
+version: 2
+
 file_extensions:
   - matlab
```
deathaxe, line 10: As `matlab` is the only associated file extension, syntax test file should probably be renamed to be opened using that syntax automatically.

jwortmann: I must say that I've only seen Matlab source code with the `.m` file extension. So I wonder if we shouldn't better change `matlab` to `m` here in the syntax? I just was reluctant to do that yet, because I'm unsure whether there would be possible unwanted implications caused by that, e.g. for Objective-C users.

deathaxe: I guess those implications are the reason for `*.m` not being bound. I just want the test file to open as Matlab files out of the box, instead of it being highlighted as Objective-C. It wouldn't change anything for real life use.

michaelblyons: [Wikipedia](https://en.wikipedia.org/wiki/List_of_filename_extensions_(M%E2%80%93R)#M) says Objective-C, Matlab, Mathematica, and a compiled Prolog variant called Mercury all share `.m`.

Shoutout to PackageDev for automatically switching the syntax of test files to the one that they're testing. 😉

jwortmann: Okay. I've changed the file extension in 56bffd0.

4086606:
Thanks for doing this! I am working on the original syntax and perhaps our changes can feed into each other.

* Improving token naming for the original syntax
* Compiling the entirety of R2021a into `vscode-matlab`'s textmate grammar (toolboxes to follow)

I've already got a local fork on my laptop with all the key tokens corrected.

jwortmann:
> Thanks for doing this! I am working on the original syntax and perhaps our changes can feed into each other.

Thanks, I'd certainly be open/happy to collaborate if we can help each other.

But that said, I think there is unfortunately no way, for example, to derive the syntaxes from a single source or so, due to the following reasons:

* the (YAML-based) .sublime-syntax format including the latest additions from Sublime Text 4 has various features which are not supported by the .tmLanguage format (I think)
* the conventions for token scope naming in Sublime Text may have slightly deviated, compared to VS Code or to what was initially used in the TextMate grammars

But feel free to write if you have any suggestions for collaborating!

> Compiling the entirety of R2021a into vscode-matlab's textmate grammar (toolboxes to follow)

This was probably the most cumbersome part when I wrote the syntax in this PR, and I would certainly welcome an automated way to compile the list of MATLAB functions. What I did here was to copy the sorted list of functions from https://www.mathworks.com/help/matlab/referencelist.html?type=function and then I manually
* removed the duplicates (with a simple regex search) and entries with spaces in it
* removed the entries that should not get scoped as a "function", e.g. classes, keywords like `if`, `while`, constants like `eps`, or other things that don't belong there like `LineSpec`, etc.
* removed most deprecated (unless widely used in code like `csvread`) and already removed functions from very old Matlab versions
* removed the object functions listed there with multiple namespaces in it, because they get treated differently here in the syntax

This process is probably error prone and isn't something I'd want to repeat twice a year for every new Matlab version. A project I've found with a similar goal to generate a list of Matlab functions is at https://github.com/StackOverflowMATLABchat/MATLABfcnscrape, but when I tried it, it didn't work well for me and its blacklist wasn't sufficient enough.

I chose to not include functions from toolboxes in this PR, because there are just too many of them and they easily can result in wrong highlighting if used as variable names otherwise. And there are some simplifications regarding command syntax for functions, as described in the comment(s) above.

Anyways, I think this PR here should be quite "ready" and it would certainly be helpful if there are other Matlab users who would like to try this syntax update and find possible improvements.

4086606:
I noticed the use of `punctuation.accessor.dot` - I suggest `punctuation.accessor.period` because that's more common.

jwortmann:
> 
> 
> I noticed the use of `punctuation.accessor.dot` - I suggest `punctuation.accessor.period` because that's more common.

I guess that's one point where scope conventions have deviated in/from other editors. A search for [punctuation.accessor.dot](https://github.com/sublimehq/Packages/search?q=%22punctuation.accessor.dot%22&type=code) in this repo shows 40 files with results, while `punctuation.accessor.period` shows none. :shrug:

4086606:
I reread this and picked up on:

> Edit: Scopes for identifiers (variables and function calls) were changed from meta.variabel.other.valid.matlab to variable.other.matlab.

As a seasoned mathematician and developer, I would consider MATLAB's matrix indexes to be a function call:
* the colon index operator makes more sense as a operator aliasing a function parameter
* the mathematical possibilities that arise from [defining matrices as functions](https://www.dhruvonmath.com/2018/12/31/matrices/)

jwortmann:
> I would consider MATLAB's matrix indexes to be a function call

Well, basically everything in Matlab is implemented as a function, but I used the common conventions for scope names, and similar to how things are scoped in other syntaxes. For example I applied the scope `constant.language` to `true`, even though it is technically a function call and can also be used as `true()` or `true(2)` (the latter returns a 2x2 Matrix of logical values 1). The same consistency argument to follow the conventions from other syntaxes applies to variables with array indexing. But yes, the "Fortran-style" indexing syntax in Matlab with round parentheses is a quite unfortunate choice regarding static syntax highlighting, because it doesn't allow to distinguish actual function calls (i.e. things that were explicitly declared as a function) from array indexing at all. And since most things in Matlab are stored as a Matrix, you can even index into "scalar" variables, e.g. if you have `x = 123`, then you can access it with either `x`, `x(1)`, `x(1,1)`, and so on. Similarly, you can omit the parentheses for function calls if there aren't any parameters. So one has to decide to either scope everything as `variable.other` (normal variable) or everything as `variable.function` (function call). And I chose the first option here, because it made more sense to me and the alternative would be a huge change for the highlighting appearance - and probably wouldn't look very good.

Also the Matlab documentation doesn't descibe [Array Indexing](https://www.mathworks.com/help/matlab/math/array-indexing.html) as "calling a variable" or so.

In general, array indexing is eventually translated to a function call in most (all?) languages by the compiler/interpreter, for example in Python, list indexing is implemented by evaluating [`__getitem__()`](https://docs.python.org/3/reference/datamodel.html#object.__getitem__), or in Julia it is [`getindex()`](https://docs.julialang.org/en/v1/base/collections/#Base.getindex). But still, the `somevariable[123]` syntax is scoped as a variable, and not as a function call, and I wouldn't want to change that. So if you'd suggest to use the scope `variable.function` for array indexing in Matlab, I guess the same should apply to all other languages too, and there is no way the scoping conventions for these will change. And what about the indexing into the contents of a cell array in Matlab (with squiggly braces), would you want to scope them as a function call too?

4086606:
I wouldn't cell indexes a function call, and the same would apply with struct indexes. I'd treat them much like `object[index]`.

I agree that functions like `true` should not be highlighted as function calls. However, I am still planning to highlight matrix indexes as function calls in the legacy Textmate bundle

deathaxe:
I'd probably suggest to

1. exclude prototype from strings
2. turn anonymous contexts into named ones in case someone wants to inherit from Matlab and for easier debugging.

Otherwise I am fine with this syntax. It's a huge improvement.

SNDST00M:
Are you all interested in supporting custom highlighting for R2021a? Gimly/vscode-matlab#132

jrappen:
@jwortmann maybe you could provide some numbers regarding performance changes using `Tools > Build with ... > Syntax Tests - Performance`?

jwortmann:
> 
> 
> I'd probably suggest to
> 
>   1. exclude prototype from strings
> 
>   2. turn anonymous contexts into named ones in case someone wants to inherit from Matlab and for easier debugging.

I've added another little tweak for strings and have done 1. now (not sure whether it's relevant though, because this syntax has no prototype context). And I will address point 2. soon(tm).

I think I also want to update the list of built-in functions to R2021a.

@SNDST00M Did you compile these function lists in your PR by hand, or did you use some kind of script for it? I see that you have a lot of duplications for the overloaded functions, so I would guess they are generated automatically? I assume that it would be better for performance to remove all the duplications.

---

> @jwortmann maybe you could provide some numbers regarding performance changes using Tools > Build with ... > Syntax Tests - Performance?

With the new syntax test file opened, I get for this PR:

> Syntax "Packages/Matlab/Matlab.sublime-syntax" took an average of 15.3ms over 10 runs

And with the syntax from master with the same file:

> Syntax "Packages/Matlab/Matlab.sublime-syntax" took an average of 4.5ms over 10 runs

But iirc when I checked this half a year ago, the syntax from this PR was faster than the original syntax, probably because the patterns for a lot of functions from the Matlab toolboxes were removed. And when I run the regex compatibility test now, I get this error:
> Packages/Matlab/Matlab.sublime-syntax:19:5: 
> FAILED: 1 pattern in "Packages/Matlab/Matlab.sublime-syntax" are incompatible with the new regex engine

It seems to be caused by the multiline regex block at
https://github.com/jwortmann/Packages/blob/d49c3b20c346ef2ce3b726065d1ba1f77396f935/Matlab/Matlab.sublime-syntax#L19
```
  matlab_functions: |-
    (?x:
```

Does anyone have an idea why that happens? I think it didn't happen in ST3 (I'm not entirely sure though).

SNDST00M:
It's done by hand, but I do believe amanya's JSON dump of MATLAB snippets could be used to do a large portion automatically. https://github.com/aminya/Matlab-Snippets/tree/1145480/snippets_vscode

jrappen:
@jwortmann Your syntax test works fine, compare the [last run](https://github.com/sublimehq/Packages/runs/3364536586). Maybe use the current dev Build 4112 locally?

deathaxe:
The regex incompatibiliy error makes no sense to me. Except sregex has some kind of limit with regards of size of alternations/patterns.

The overall syntax cache is about 200kB. So I wouldn't care much about performance - especially with all the syntax improvements this PR introduces. It is still much simpler then many others.

With regards to prototype handling: If anybody is keen enough to create an inherited syntax definition which introduces a `prototype` to inject some kind of template-variable-interpolation (see Makefile -> ShellScript), such aspects become important.

jrappen:
Thanks for this. Unfortunately I only had a little bit of time to review this today. I noticed a few things:

#### Tests

```text
.
└─ Matlab/
   ├─ tests/
   │  ├─ indentation/       # rules defined but not backed up by tests, yet
   │  ├─ symbols/           # any plans?
   │  └─ syntax/            # maybe you could move current tests here
   ├─ ...
   └─ ...
```

For syntax tests maybe move the file. Breaking up the file into logical parts might also make sense, since these are automated and it speeds up the time for finding the location target for a quick manual edit in the future. At least for me.

```text
.
└─ Matlab/
   └─ tests/
      └─ syntax/
          ├─ syntax_test_matlab.general.matlab
          ├─ syntax_test_matlab.numbers.matlab
          └─ syntax_test_matlab.operators.matlab
```

For indentation tests compare: <https://github.com/sublimehq/Packages/search?q=reindent>. Markers are `reindent`, `reindent-unchanged` or `reindent-unindented`. This functionality is **not yet** documented in the official docs.

For symbol tests compare sublimehq/sublime_text#3154 (ST4087+). This functionality is **not yet** documented in the official docs.

#### Snippets

You haven't touched the snippets. I assume you have tested them against the new scopes from your PR.

#### Metadata (i.e. `*.tmPreferences`)

I assume you have tested them against the new scopes from your PR.

michaelblyons:
My understanding was that the indentation tests were "This file should remain as-is when it is forcibly reindented." Perhaps I am mistaken.

keith-hall:
- `reindent-unchanged`: if the whole file is reindented as is, no indentation should change
- `reindent-unindented`: if the whole file has all indentation removed, the indentation should be the same again after reindenting
- `reindent`: a shortcut for both `reindent-unchanged` and `reindent-unindented`

jwortmann:
I've added tests for the indentation rules and some simple tests for symbols in b7f1c88.

The indentation tests exclude the *switch ... case* statement, because I think it is impossible to define correct indentation rules for it with ST's regex approach in .tmPreferences files. Example:
```matlab
n = input('Enter a number: ');

switch n
    case -1
        disp('negative one')
    case 0
        disp('zero')
    case 1
        disp('positive one')
    otherwise
        disp('other value')
end
```

And the symbol tests lack tests for references, because variables and function calls are indistinguishable in Matlab.

jrappen:
@jwortmann Could you resolve conflicts? That should also trigger test against the latest dev and stable, although I don't expect any changes from 4116 to 4120 and 4121.

jrappen:
@michaelblyons @deathaxe Could you re-review? Thanks.

deathaxe (approved):

jrappen (approved):

keith-hall (approved):

PR #2651, opened 2020-12-12
[Matlab] Disable auto-pair of single quote when used as transpose operator

jwortmann:
A single quote directly following a dot `.'` is the transpose operator in Matlab and a single quoted string after a dot (without whitespace in between) would be syntax error afaik.

### DIFF
--- a/Matlab/Default.sublime-keymap
+++ b/Matlab/Default.sublime-keymap
@@ -1,0 +1,10 @@
+[
+    // Disable auto-pair for single quote when used as transpose operator
+    { "keys": ["'"], "command": "insert_snippet", "args": {"contents": "'"}, "context":
+        [
+            { "key": "selector", "operator": "equal", "operand": "source.matlab" },
+            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\.$", "match_all": true }
+        ]
+    }
+]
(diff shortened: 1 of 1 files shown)


jfcherng (approved):

Matlab/Default.sublime-keymap
```diff
@@ -0,0 +1,10 @@
+[
+    // Disable auto-pair for single quote when used as transpose operator
+    { "keys": ["'"], "command": "insert_snippet", "args": {"contents": "'"}, "context":
```
deathaxe, line 3:
```suggestion
    { "keys": ["'"], "command": "insert", "args": {"characters": "'"}, "context":
```

A normal `insert` should be sufficient here.

Matlab/Default.sublime-keymap
```diff
@@ -0,0 +1,10 @@
...
+        [
+            { "key": "selector", "operator": "equal", "operand": "source.matlab" },
+            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "[\\.)}\\]]$", "match_all": true }
```
deathaxe, line 7:
```suggestion
            { "key": "preceding_text", "operator": "regex_contains", "operand": "[])}.]$", "match_all": true }
```

With little changes, no escapes are needed in the `preceding_text` pattern. Not too important though.

jfcherng (approved):

deathaxe (approved):

PR #2654, opened 2020-12-18
[Java] Rewrite Syntax

### LINKED ISSUE #611: [Java] Add support for module-info.java highlighting
Java 9 is bringing a new file specification, contained within files named `module-info.java`, which describes the metadata around Project Jigsaw modules.  You can see some examples here: http://openjdk.java.net/projects/jigsaw/quick-start  It is not correct to highlight these files using Java mode, despite the extension, and since they are an inextricable part of the language (post Java 9), we should provide scoping for them as part of the built-in Java package.

### LINKED ISSUE #736: [Java] Correctly scope fully-qualified identifiers

### LINKED ISSUE #1610: [Java] Highlighting issue
The following breaks the highlight. However, if `A` is replaced with `a` it works properly.

```java
class A{
A(){}}
```

Screenshot:

### LINKED ISSUE #1611: [Java] More issues......
```java
class a{{
  int b;
  float c;
  double d;
  a e = new a(){};
  b = e instanceof a?1__1:0b11110101;
  c = 0x.aap0f-0x.aP-1F;
  d = 0x.aap0-0x.aP+1;
  b=e.a(b<b)> b?b:b;
  b=e.a(b<b)>b?b:b;
}int a(boolean a){return 1;}}
```

A lot of issues here:

1. Line 6: `instanceof` not highlighted
2. Line 6: literal integer `1__1` not highlighted
3. The same line: literal integer `0b11110101` not highlighted
4. Line 7: literal floating point `0x.aap0f` not highlighted
5. Line 7: literal floating point `0x.aP-1F` not highlighted
6. Line 8: literal double `0x.aap0` not highlighted
7. Line 8: literal double `0x.aP+1` not highlighted
8. Line 9: operands of `<` are highlighted like a generic class

---

### Edit

One more issue:

```java
class a<b>{{
  a<a>a;
}}
```

The generic type on line 2 is not highlighted.

### LINKED ISSUE #2919: [Java] Syntax Highlighting text blocks not supported at all, it seems
Java now supports text blocks in code. Sublime Text 4 doesn't appear to know anything about this and tries to highlight it as if it's code.

(I held off on reporting this while I was still on ST3, but now seems the time... Needless to say vscode OOTB does it right. 😉 I note github does too, in the example code block below...)

- Sublime Version: 4113
- OS Version: MacOS 11.5.1 (but expect is everywhere)

This complete program compiles and runs in java 16, and likely back to whichever java version text blocks appeared in.

``` java
public class textblock {
    private static final String USAGE = """
        textblock [options] [args...]

          --help
            Show this text.
        """.stripIndent();
        System.out.println(USAGE);
    }
}
```

image

### LINKED ISSUE #2991: [Java] Support shebang line
Since Java 11, the Java runtime supports executing single-file programs with a shebang line. It skips the initial line iff the file does not have an extension.

> When the launcher reads the source file, if the file is not a Java source file (i.e. it is not a file whose name ends with `.java`) and if the first line begins with `#!`, then the contents of that line up to but not including the first newline are ignored when determining the source code to be passed to the compiler. 

https://openjdk.java.net/jeps/330

``` java
#!/usr/bin/env -S java --source 16

class HelloWorld {
    public static void main(String[] args) {
        System.out.println("Hello, World!");
    }
}
```

deathaxe:
fixes #611
fixes #736
fixes #1610
fixes #1611
fixes #2919
closes #2991 

At it current stage the syntax may be considered a rewrite, while it has actually been improved step by step from the existing implementation. All existing test cases have been maintained while being reorganized into logical sections/groups and being extended heavily.

On the way going it has turned into a proof of concept to illustrate the power of ST4's syntax engine and push it to the limits to reveal its current possibilities.

### Adds Java 16 langage features

- sealed classes
- record classes
- switch expressions
- compact constructors

### Closely follows Java 13 Language Specification

The main goal is to match types and identifiers case insensitive whenever possible to fix as many ambiguities as possible which arise from current casing heuristics.

```Java
class a<b> implements intf<? extends test<b>> {
    string c() {
        e.f.a<b> c = (e.f.a<b>) new e.f.a<b>("foo");
        return c.say();
    }
}
```

grafik

That's achieved by using ST4's branch point feature to closely follow [Java 13 Language Specification](https://docs.oracle.com/javase/specs/jls/se13/html/index.html).

### Overcome ST's next-line-blindness

All expressions can be wrapped wherever it makes sense without breaking syntax highlighting. No matter whether fully qualified identifiers, array modifiers, or function parameter groups all of them can be placed to onto following lines.

```Java
   class {
       void main() {
           h
           .
           e
           .
           l
           <
           ?
           extends
           @
           l
           .
           o
           (
           m
           =
           y
           )
           w
           >
           o
           []
           =
           new
           r
           (
           )
           .
           l
           .
           d
           (
           )
           ;
       }    
   }
```

grafik

## Meta Scopes

With help of branching we now can scope all parts of members with desired `meta.function` scopes, even if those parts are located on different lines or consist of sophisticated expressions with arbritary nested contexts needed to match them.

```Java

class ClassName {
    
    @Annotation public int<T> method(int args) {
//  ^^^^^^^^^^^^^^^^^^^ meta.function.modifier        
//                     ^^^^^^^ meta.function.return-type
//                            ^^^^^^ meta.function.identifier
//                                  ^^^^^^^^^^ meta.function.parameters        
    }

    @Annotation 
//  ^^^^^^^^^^^ meta.function.modifier        
    public 
//  ^^^^^^ meta.function.modifier        
    int<T> 
//  ^^^^^^ meta.function.return-type
    method
//  ^^^^^^ meta.function.identifier
    (int args) 
//  ^^^^^^^^^^ meta.function.parameters        
    {
    }
}

```

grafik

## Java Server Pages

The JSP syntax has been revamped using **inheritance** to replace `with_prototype` approach to inject JSP tags into HTML, CSS and JS with far less memory footprint.

```JSP
<!DOCTYPE html>
<html>
<head>
    <title></title>
    <style type="text/css">

        @media <% print(myMedia) %> {
        }

        tr.<% print(myClass); %> {
            color: <% print("<\%foo%\>"); %>;
        }
    </style>

    <script type="text/javascript">
        <% if (true) { %>
            if (true) { <% print ("hello%\>") %> }
        <% } else { %>
            console.write("yep");
        <% } %>
    </script>
</head>
<body>
    <%-- This is a comment --%>
    <%@ include file="foo.bar" %>    
    <font color="<%=color.blue()%>"><%= map.size() %></font>
</body>
</html>  
```

grafik

**Note:**

It is not possible to maintain structure of two heavily context sensitive syntaxes such as Java and JavaScript when they are merged (the one embedded into the other). The decision was made to try to keep JavaScript intact and create Java in a way it may fail gracefully with loosing some meta scopes while trying to keep overall syntax highlighting intact. This may not work perfectly for all kinds of theoretically sophisticated edge case scenarios.

### Miscellaneous

- optimized to maintain syntax highlighting in incomplete statements
- rewritten JavaDoc syntax
- ...

## Benchmarks

It's hard to make general statements about performance when branching is heavily used. Results heavily rely on content being used for benchmarks. While micro benchmarks of certain statements show significant slow downs overall impact in real world code seems to be less significant.

What does it mean in numbers?

Comparing the original syntax tests (current master) the new syntax seems 50% faster.

The largest source file of LemMinX is [TreeLineTracker.java](https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/TreeLineTracker.java): 

Description                              | before  | now    | diff
-----------------------------------------|--------:|-------:|-----:
Original file size (43kB / 1450 lines)   | 10.7ms  | 12.5ms | +17%
Content duplicated (361kB / 10.1k lines) | 66.6ms  | 82.4ms | +24%

## TODO

The syntax is at a point when it needs to be used in real world projects in every day work to find remaining issues.


deathaxe:
It appears the online syntax checker has some trouble I can't reproduce locally.

rwols:
The CI [might need an upgrade to 4094](https://github.com/sublimehq/Packages/blob/1218ebbb58c01641361402bfeb0a2633479dd5b3/.github/workflows/ci.yml#L20).

michaelblyons:
> The CI [might need an upgrade to 4094](https://github.com/sublimehq/Packages/blob/1218ebbb58c01641361402bfeb0a2633479dd5b3/.github/workflows/ci.yml#L20).

For those of you following along at home, that's now merged with #2655.

jrappen:
Will was so kind to do the fastest merge to date.

michaelblyons (approved):
I'm not primarily a Java user, but for several months I've been opening all the Java files I _do_ use with this PR, and I haven't found any problems.

FichteFoll:
Finally got the chance to test this out a bit. Here's some feedback:

1. `meta.annotation` covers the trailing newline character, but unlike before it doesn't extend until the first non-whitespace anymore (thankfully). I have a background color for that scope, which is why I notice it, and I don't think it should cover the newline.
2. Classes receive `storage.type` in all instances, even when their static methods are accessed. We've been over this in #1842, but I believe that they should only receive `storage.type` if they are used in a declaration. (`Double d = Double.valueOf("1.0");`)
3. Generic punctuation should be section, not definition.
4. The bound identifier in a catch header is scoped as `variable.parameter`. Not a big issue, but it's odd.
5. Should ternary operators use `keyword.operator.ternary` or `keyword.operator.conditional`? The latter would also allow grouping null-coalescing operators into it. You recently closed the JavaScript PR with that, iirc.
6. in `import static org.apache.commons.lang3.StringUtils.isNotBlank;`, the second-to-last identifier is scoped as `entity.name.class` despite it not being imported. Only the function is.

deathaxe:
> `meta.annotation` covers the trailing newline character, 

Fixing that would require another level of branching in annotation contexts as we might want no gap in `meta.annoation` with parameter lists on a separate line.

```Java
    @Override
    (arg)
    class ...
```

FichteFoll:
> Fixing that would require another level of branching in annotation contexts as we might want no gap in `meta.annoation` with parameter lists on a separate line.

Yes, but to me that would certainly be worth it. I haven't seen code that splits annotation arguments between the ident and the parameters, so while supporting it is the syntactically correct way, the 99% case is more important imo.

> AFAIK, this was the case before and I just kept it.

Let's keep it then until we eventually find an answer to that.

> Imports don't use branching yet, which would be required to support it properly.

Could you make use of the fact that it's an `import static` and simplify the logic or would you still have to duplicate many contexts? Not a huge issue for me, but it does look wrong when you have a prominent highlighting of `entity.name` tokens.

deathaxe:
> I haven't seen code that splits annotation arguments between the ident and the parameters

Many constructs supported by Java may not be found often, but with regards to arguments list I just have https://github.com/sublimehq/sublime_text/issues/4152 in mind. But I guess with the whole syntax being heavily based on branches one more doesn't matter.

> Let's keep it then until we eventually find an answer to that.

Java used to distinguish between `packages` (lowercase)` and `classes` (PascalCase) in fully qualified identifiers. Removing that distinction, would probably heavily simplify a couple of things as at least one level of branching could be removed. But I expect a couple of users to complain about loosing it. That's why I decided for the hard way.

> Could you make use of the fact that it's an `import static` and simplify the logic

Need to look into it. This part was one of the first things rewritten, before branching was possible. In general the whole branching stuff causes nearly everything to be duplicated for every single syntactical construct - that's what I've learned. While poerful it is a pain to develop.

FichteFoll:
Something that I noticed with imports scoping the imported classes and functions as `entity.name.class` and `entity.name.function` respectively is that it interferes with goto definition. Each import is also offered as a definition of the class. Now, that could be mitigated by an additional tmPreferences file that removes these if `meta.import.java` is also in the scope.

However, I just don't think `entity.name` is appropriate for imported names, especially when these are highlighted strongly. To give some perspective:



In https://github.com/sublimehq/Packages/issues/1842#issuecomment-476346859 we talked about using `entity.name.import`, which is maybe not proper but something general enough I can exempt from my significant highlighting scope selector. With `- meta.import entity.name` I wouldn't feel so safe. Has this been discussed before or is it used anywhere else? I do feel that we need to talk about this.

Otherwise the PR should be fine. I don't really edit Java in ST, but the code I viewed with it looked fine otherwise, except for maybe some new features that aren't supported yet, but those can always be added later.

FichteFoll:
I also noticed that switch expressions with `->` have the `->` scoped as a lambda declaration.

https://openjdk.java.net/jeps/325

deathaxe:
> I also noticed that switch expressions with -> have the -> scoped as a lambda declaration.

This is basically also intented due to lack of an idea what to call it otherwise.

We can also just scope the arrow `punctuation.separator.expression` as the normal colon, if it feels more appropriate. Would be an easy fix.

StrangeNoises:
I've hit a bug with the new Java syntax. Screenshot may be mostly self-explanatory:

Screenshot 2021-09-03 at 17 18 46

This happened when *typing* the code shown into a Sublime Text 4 window with the Java syntax selected. As you type `case null -> ` the next character typed hangs the app, and it must be force-quit.

Opening a window, selecting the Java syntax, and *pasting* the code in doesn't cause a problem, although the `n` in the `null` is still the wrong colour.

This is:

Sublime Text 4 build 4113
Java syntax from the deathaxe/sublime-packages repo, branch `pr/java/rewrite-syntax`, git-pulled as of just now.
MacOS Monterey 12.0 beta

The typed/pasted in code is from this page: (https://blogs.oracle.com/javamagazine/hidden-gems-jdk16-jdk17-jep#anchor_10), the fourth quoted code block down from that JEP406 anchor. Yes it is valid syntax now, even if only in preview mode. 😉

I usually have code ligatures on; they're off in the screenshot as I was testing if that was the problem. It appears not to be.

It also happens if it's a switch statement style syntax, ie: `case null: ` (next character typed crashes it).

Until this problem, this updated java syntax has been working very well for me; i'd not had any problems with it at all. (But needed it as various other recent java syntax changes make working with the still-current java syntax in ST almost ususable.)

mitranim:
Related to ↑. I've had ST hang when using an experimental SBNF-based implementation of SQL. Didn't bother to make a min-repro due to the complexity of the generated syntax, but can link to sources if needed.

deathaxe:
Using branch points in conjunction with multi-push/set statements can quickly confuse ST's failsafe algorithm which tries to avoid infinite loops. Or those can aggressively try to hit such situations.

At some point there are just too many `(?=\S)` involved in pushing/branching/popping.

That said, the most recent commit fixes this specific `case null` issue.

StrangeNoises:
yes, that seems to have fixed it for me. The above test works fine now. I can go on playing with ~~fire~~preview next week. 😉

deathaxe:
Any feedback or issue report is welcome to iron out the syntax definition.

michaelblyons (dismissed):

deathaxe:
Sorry, deleted the wrong branch by accident.

michaelblyons:
@djspiewak and @ismell: Any interest in giving this PR a try?

jrappen:
> **djspiewak** and **ismell**: Any interest in giving this PR a try?

also @lastsecondsave and @StrangeNoises as former contributors to the package

StrangeNoises:
I've been using it on my daily driver for java 17 dev for weeks. If I've been quiet it's because it's been working for me without issues. 😀 I don't think I have a role in accepting it (that would be a shock!) but I do look forward to an app update that includes it, hopefully sometime soon, so that's my +1 if it's worth anything.

ismell:
> @djspiewak and @ismell: Any interest in giving this PR a try?

I'm no longer coding in Java. But from a quick glance it looks great!

lastsecondsave:
I also don't work with Java at the moment. I've checked a bunch of classes from JDK and this syntax is pretty much perfect. Although `native` keyword is considered invalid for some reason:
Screenshot from 2021-10-05 20-05-00

deathaxe:
Just and oversight.

deathaxe:
Resolved a conflict with #2760.

michaelblyons (approved):

FichteFoll (approved):
I believe this is more than ready for a dev cycle. Any further edge cases can be ironed out later.

rwols (approved):

michaelblyons:
Hooray 🎉

jwortmann:
I think I found an example where the current highlighting is not optimal:

If you write a single method call or static function call on a line, and the next line is a variable declaration or assignment, then the method/function call will be scoped `storage.type.class` while typing. For the case of a variable declaration on the next line (excluding primitive types), the class name for that variable will even switch its color while you type.

Here is an example (I disabled the autocompletion for less distraction):

java

```java
import java.util.ArrayList;

public class MyClass {

    public void myMethod() {
        ArrayList<String> myList = new ArrayList<String>();

        myList.add("foo");
        myList.add("bar");
        myList.add           // <- currently typing

        String str = "test"; // "myList" and "add" are storage.type.class in the line above, "String" is not
    }
}
```

There is a pretty strong convention in Java, that class names should start with an uppercase letter, and variables should start with a lowercase letter. Maybe the shown highlighting is what the compiler would see at the particular point in time, but it is not how I as a user would expect syntax highlighting to behave. So I would say the current highlighting for this example is a bad heuristic about what the user wants to write. Because I think it is extremely unlikely that the user really intents to write a variable declaration in the form of using both a lowercase class name *and* a linebreak between the class name and variable identifier, at the same time.

Could this perhaps be improved?

jrappen:
@jwortmann Not sure why you don't open a new issue. This PR was already merged.

Issue #1716, opened 2018-09-17
:line not working for any file except the last one


gituser823:
Hello,

I have a bash script, that opens sublime in the following manner:

subl $DIR $file1 $file2 $file3:10000 $file4:100000 $file5 $file6:1000000

The Problem is now, that it only scrolls to the end of the $file6.
So :line only seems to work for the last file.
Please fix this bug, so lines can be used for every file.
I am on Linux Kernel 4.4.0, Linux Mint.
My Sublime Version is 3.1.1 Build 3176.

Thanks!

FichteFoll:
Please post this issue on the core issues tracker here: https://github.com/SublimeTextIssues/Core/issues

This repository is about the default packages only (which are open source).

wbond:
Closing since this is on the wrong repo. I believe this may have been fixed in the 4xxx build anyway.

Issue #1214, opened 2017-10-10
[Markdown] Is there a way to consume the full width of a line without matching \n?


dubeg:
This is for fenced code blocks.
I like to set their background to a different color than the buffer's bgcolor, and having the lines of the delimiters with that same color, but it's not working if the delimiter line is the last line of the buffer.

Ex:
![md](https://user-images.githubusercontent.com/6020602/31396143-9667a0f6-adb0-11e7-9b61-85d6cb9ec458.PNG)

keith-hall:
Nope, this is currently not possible - the only way to get background highlighting for the full width of the view is with a `\n` character. You may want to consider having a blank line at the end of your files. For example by setting `"ensure_newline_at_eof_on_save": true,` in your preferences.

wbond:
If you are still interested in this @dubeg, would you mind opening an issue on the core tracker, or seeing if there is an existing issue that covers this?

dubeg:
Thank you @wbond, I don't really care anymore. I'll close it.

PR #2656, opened 2020-12-18
Close issues

### LINKED ISSUE #1555: [ASP] Error loading, apparent recursion
Environment: Mac OS 10.11.6
Build: Version 3.0, Build 3170

On load of a Classic ASP file (with or without included HTML), the following error appears:

screen shot 2018-05-10 at 7 48 23 am

### LINKED ISSUE #1671: [Java] macOS Segmentation Fault
On macOS indexing [SmallSortedMap.java](https://github.com/google/protobuf/blob/593e917c176b5bc5aafa57bf9f6030d749d91cd5/java/core/src/main/java/com/google/protobuf/SmallSortedMap.java) triggers a segmentation fault in Sublime Text.  Removing the Java syntax provided by Packages and using the default Sublime Text Java syntax fixes the problem.

The relevant portion of the crash report and software versions are below.

**Packages:** https://github.com/sublimehq/Packages/commit/289782ff2e4cb58de171579c7fc86fe00d280619
**Sublime Text:** Build 3176
**OS:** Mac OS X 10.13.6

**Note:** This does not occur on Ubuntu 18.04 with Sublime Text build 3176.

```
Process:               Sublime Text [55705]
Path:                  /Applications/Sublime Text.app/Contents/MacOS/Sublime Text
Identifier:            Sublime Text
Version:               Build 3176 (3176)
Code Type:             X86-64 (Native)
Parent Process:        ??? [44715]
Responsible:           Sublime Text [55705]
User ID:               501

Date/Time:             2018-07-23 14:47:28.808 -0400
OS Version:            Mac OS X 10.13.6 (17G65)
Report Version:        12
Bridge OS Version:     3.0 (14Y664)
Anonymous UUID:        86D53BC0-A343-C9E8-8A56-7F7D25A0719D

Sleep/Wake UUID:       AF7D7861-E97F-4B83-BB49-7E5B968879DB

Time Awake Since Boot: 29000 seconds
Time Since Wake:       8500 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000108840000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [0]

VM Regions Near 0x108840000:
    MALLOC_LARGE           0000000108800000-0000000108840000 [  256K] rw-/rwx SM=PRV  
--> 
    MALLOC_LARGE           000000010885e000-00000001088a5000 [  284K] rw-/rwx SM=PRV  

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.sublimetext.3             	0x0000000106b194b0 emit_remainder(usubstring, LexerState*, std::__1::vector<TokenInfo, std::__1::allocator<TokenInfo> >&) + 368
1   com.sublimetext.3             	0x0000000106b18eb3 lex(lex_working_memory&, LexerState*, usubstring, std::__1::vector<TokenInfo, std::__1::allocator<TokenInfo> >&) + 13299
2   com.sublimetext.3             	0x0000000106bbb957 TokenStorage::tokenise(BucketContainer<tree_list<TokenBuffer, TokenBuffer::token_list_ops>, TokenStorage::TokenBufferFactory>::iterator, long long, LexerState*, usubstring) + 663
3   com.sublimetext.3             	0x0000000106bba27d TokenStorage::insertCharacters(long long, usubstring) + 2573
4   com.sublimetext.3             	0x0000000106b111e0 process_read_data(ReadData*) + 1491
5   com.sublimetext.3             	0x0000000106b10250 FileBufferSource::beginRead(TextBuffer*, TokenStorage*) + 696
6   com.sublimetext.3             	0x0000000106b6ebfc TextBuffer::associate(BufferSource*) + 242
7   com.sublimetext.3             	0x0000000106a7e0df crawl_app_loader::callback_finished() + 1829
8   com.sublimetext.3             	0x0000000106c8ea51 setup_operation::request_finished() + 39
9   com.sublimetext.3             	0x0000000106c44e98 catalogue_rescan_operation::merge_finished(catalogue::dir*) + 184
10  com.sublimetext.3             	0x0000000106c4702c catalogue_rescan_operation::scan_callback::main_scan_single_dir(substring, std::__1::vector<file_info, std::__1::allocator<file_info> > const&) + 7304
11  com.sublimetext.3             	0x0000000106c8b965 result_helper::main_work_item::operator()() + 19
12  com.sublimetext.3             	0x0000000106cc8a1e -[WorkQueueCallback processItems:] + 319
13  com.apple.Foundation          	0x00007fff58d2f2b5 __NSThreadPerformPerform + 334
14  com.apple.CoreFoundation      	0x00007fff56c09a11 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
15  com.apple.CoreFoundation      	0x00007fff56cc342c __CFRunLoopDoSource0 + 108
16  com.apple.CoreFoundation      	0x00007fff56bec470 __CFRunLoopDoSources0 + 208
17  com.apple.CoreFoundation      	0x00007fff56beb8ed __CFRunLoopRun + 1293
18  com.apple.CoreFoundation      	0x00007fff56beb153 CFRunLoopRunSpecific + 483
19  com.apple.Foundation          	0x00007fff58ce7f26 -[NSRunLoop(NSRunLoop) runMode:beforeDate:] + 277
20  com.sublimetext.3             	0x0000000106cd11f7 px_run_event_loop() + 144
21  com.sublimetext.3             	0x0000000106a0a8ca main + 2853
22  com.sublimetext.3             	0x0000000106a099d4 start + 52
```

### LINKED ISSUE #1696: [CSS] units in scientific notation

### LINKED ISSUE #1699: [Makefile] Remove or refactor shell syntax highlighting
From the [make docs](https://www.gnu.org/software/make/manual/make.html#Recipe-Syntax):

> Makefiles have the unusual property that there are really two distinct syntaxes in one file. Most of the makefile uses make syntax (see Writing Makefiles). However, recipes are meant to be interpreted by the shell and so they are written using shell syntax. **The make program does not try to understand shell syntax: it performs only a very few specific translations on the content of the recipe before handing it to the shell.**

I think we should make the syntax match what make is really doing.

    foo:
    	echo $${HELLO} $(dir src)

`$$` is actually telling make to emit a `$` to the shell script, while `$(dir ...)` is a built in make function that emits the text to the shell script. Right now the syntax tries to intermingle the two and results in them stepping over each other. 

Also, the shell the make file is using might not be `sh` so you would get strange syntax highlighting for these cases as well.

### LINKED ISSUE #1771: [JavaScript] Move JSON out of the JavaScript package.

### LINKED ISSUE #2034: SCSS CSS4 selector coloring wrong
Certain [CSS4 selectors](https://css4-selectors.com/selectors/) have no highlighting after them, or mess up highlighting for following selectors.

This code highlights `:has(...)` correctly, but the `ul.links li.dismiss` part is white (wrong).

    .contextual-links-region:has(> .contextual-links-wrapper) ul.links li.dismiss {  
      top: 0px;  
      right: -30px;  
    }

Pasting a copy of the above code after itself will cause the 2nd occurrence to highlight `ul.links li.dismiss` correctly, but the `:has(...)` part is different. Also notice that some italics is missing inside on `top` in the first block.

Image of code highlighting

The highlighting of the 2nd block is what it looks like when pasted into a CSS file. So it looks like SCSS has some improvement (styles the :has better, but only sometimes).

Most of the selectors from CSS4 have no highlighting yet, and they will eventually need to be added. While not all of the are supported by browsers yet, many of the new features are. You can check your browser's support here: https://css4-selectors.com/browser-selector-test/

### LINKED ISSUE #2329: [YAML] "on" (and related) should not be scoped as constant in key position
```yaml
  on: on
# ^^ entity.name.tag.yaml - constant
#     ^^ constant.language.boolean.yaml

```

This test fails at present because the `on` *key* is scoped the same as the `on` *value*. This also holds true for all the other boolean constants (e.g. `true`), and really *all* constants (e.g. `42`).

This comes up increasingly frequently these days because `on` is a mandatory key in any GitHub Actions workflow definition.

### LINKED ISSUE #2550: Indentation regression for the Julia language in ST4?
- Sublime Version: 4088
- OS Version: Windows 10

## Expected behavior
Re-indent the code in this MW E: https://github.com/JuliaEditorSupport/Julia-sublime/issues/90#issuecomment-707996405

## Actual behavior
Only the definition of the function without the comment is indented correctly.
ST3 can indent both functions correctly.

## Steps to reproduce

Install the Julia package.
Create a Julia source file with
```
function f(x)
y = x^2 # comment
end

function f(x)
y = x^2 
end
```
Select all, and choose "Reindent".

### LINKED ISSUE #2674: CSS is broken in ST 3
CSS from master works fine in ST 4, but it is broken in ST 3. When I open any css file in ST 3 with CSS from master from this repo, I get the entire file colored as a css class name. Can someone please confirm the issue and update the style so that it works in ST 3 as well. Thanks.

If nobody can reproduce, I can share my screen recording and I can also bisect it if you want. But save me from effort

### LINKED ISSUE #2681: Color is not getting changed
:root{
--svdeepblue:#010a51;
--svlightblue:#0213a1;
--svmainyellow:#feca07;
--svlightyellow:#b99408;
--svwhite:#fff;
}

background-image: linear-gradient(var(--svlightblue) , var(--svlightyellow));

Color of Code is not getting changed while coding in Cascading style sheet.

color changed github

deathaxe:
The following list of issues seems safe to close.

Closes #1555 (fixed)
Closes #1671 (fixed by core)
Closes #1696 (fixed)
Closes #1699 (fixed)
Closes #1771 (fixed)
Closes #2034 (fixed)
Closes #2329 (fixed)
Closes #2550 (invalid)
Closes #2674 (invalid)
Closes #2681 (fixed)


deathaxe:
The following ones may be moved to core issues.

#1214
#1716

wbond:
Thanks for the list!

PR #2661, opened 2020-12-20
[C/C++/ObjC/ObjC++] Fix floating point number scopes

deathaxe:
Addresses #2630

This PR 
1. fixes `constant.numeric.value` assignment to numeric constants
2. simplifies some number patterns by using capture groups rather than pushing into contexts
3. makes patterns more equal between C and C++


wbond:
Thank you!

PR #2667, opened 2020-12-20
[HTML] Add scope for path separators

deathaxe:
This commit aligns syntax highlighting of urls with the implementation used in CSS.sublime-syntax, which scopes `/`, etc. as path separators.


michaelblyons:
If you're going to add `?` and `&` (which I think is a good idea), maybe also put in `#`?

PR #2673, opened 2020-12-23
[Git Formats] Target the correct scope with the snippet

frou:
`source.git.config` must be outdated, because there's no occurence of it elsewhere in the repo.


michaelblyons:
Looks like it was changed at https://github.com/sublimehq/Packages/commit/a81f4e383e8e1424ea246f0abdf31a04f0b3603a.

deathaxe:
The old scope was `source.gitconfig`. Hence this snipped never worked. Good catch.

deathaxe (approved):

PR #2678, opened 2020-12-28
[ShellScript] Add key bindings for auto-pairing

### LINKED ISSUE #2677: [Bash] Auto-pair interpolations even in a string context
- Sublime Version: 4094
- OS Version: macOS 11.1

## Expected behavior

`|` denotes the caret. When typing a `{` in

```
foo="$|"
```

I would expect an auto-pairing keybinding to trigger and complete it to

```
foo="${|}"
```

## Actual behavior

ST ignores auto-pairing of curly brackets inside strings and we end up with

```
foo="${|"
```

Note that the auto-pairing of curly-brackets and parentheses works as intended when not in a string context.

Because the bash syntax is fairly good these days, I would expect such a key-binding to make use of

- `meta.interpolation.parameter` or,
- `punctuation.section.interpolation.begin`.

From what I understand the `"selector"` key of a keybinding context checks the scope *to the right* (?) of the caret, making `punctuation.section.interpolation.begin` useless for this.

deathaxe:
Fixes #2677

This PR proposes to add key bindings to...

1. Auto-pair command interpolation parentheses

   ```
   $|         -> $(|)
   "$|"       -> "$(|)"
   foo$|baz   -> foo$(|)baz
   "foo$|baz" -> "foo$(|)baz"
   ```

   Note: Removing parens via backspace is handled by default bindings.

2. Auto-pair parameter interpolation braces

   ```
   $|         -> ${|}
   "$|"       -> "${|}"
   foo$|baz   -> foo${|}baz
   "foo$|baz" -> "foo${|}baz"
   ```

   Note: Removing braces via backspace is handled by default bindings.

3. Auto-pair backtick quotes

   ```
   |       -> `|`
   "|"     -> "`|`"
   foo|baz -> foo`|`baz

   `|`      -> |
   ```

4. Expand interpolation brackets

   ```
   (|)  -> ( | )
   {|}  -> { | }
   ```

5. Collapse interpolation brackets

   ```
   ( | )  -> (|)
   { | }  -> {|}
   ```

**Notes:**

All bindings are limited to `source.shell` and `setting.auto_match_enabled`.

### DIFF
--- a/ShellScript/Default.sublime-keymap
+++ b/ShellScript/Default.sublime-keymap
@@ -1,0 +1,89 @@
+[
+    //
+    // Command Interpolation
+    //
+
+    // Auto-pair backticks: `|`
+    { "keys": ["`"], "command": "insert_snippet", "args": {"contents": "`${0:$SELECTION}`"}, "context":
+        [
+            { "key": "setting.auto_match_enabled", "operand": true },
+            { "key": "selector", "operand": "source.shell - meta.interpolation.command" },
+            { "key": "preceding_text", "operator": "not_regex_contains", "operand": "`$", "match_all": true },
+            { "key": "following_text", "operator": "not_regex_contains", "operand": "^`", "match_all": true }
+        ]
+    },
+    // Remove backticks when backspace is pressed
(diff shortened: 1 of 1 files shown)


wbond:
Looks like some reasonable quality of life improvements, thanks!

PR #2671, opened 2020-12-22
[Haskell] Fix various syntax highlighting issues

deathaxe:
Fixes #2670

According to https://www.haskell.org/onlinereport/haskell2010/haskellch10.html a comment starts with at least 2 dashes followed by `any` character but symbols with some exceptions. The relevant rules are:

```
  comment   -> -- {-} [ any⟨symbol⟩ {any} ] newline
  symbol    -> ascSymbol | uniSymbol⟨special|_|"|'⟩
  special   -> ( | ) | , | ; | [ | ] | ` | { | }
  ascSymbol -> ! | # | $ | % | & | ⋆ | + | . | / | < | = | > | ? | @ | \
             | ^ | | | - | ~ | :
  uniSymbol -> any Unicode symbol or punctuation
```

With:

```
  any       -> graphic | space | tab
  graphic   -> small | large | symbol | digit | special | " | '
```

we can say:

```
  comment   -> -- {-} [
               space | tab | small | large | digit | special | " | '
               {any} ] newline
```

This simplified pattern is implemented in this commit.

Additionally `(--)` and `(---)` are excluded from infix operators.

---

Fixes #2672

Don't want to make another rewrite, but to avoid merge conflicts fixes for issue #2672 had to go in here as well. Details about the changes may be found in the commit messages.

PR #2628, opened 2020-11-18
[Haskell] match (--*) in punctuation.definition.comment.haskell

moodmosaic:
/cc @FichteFoll 

Before

image

After

image

PR #2679, opened 2020-12-30
[Haskell] Rewrite Syntax

### LINKED ISSUE #1321: [Haskell] Function definition spread over multiple lines not correctly highlighted or lexed correctly

### LINKED ISSUE #2670: [Haskell] Syntax Highlighting problem
Perfectly valid Haskell code:

```
(--)'ab'(
  main)=putStr"ok"
```

### LINKED ISSUE #2672: [Haskell] More syntax highlighting problems
```
{-# LANGUAGE TemplateHaskell, CPP, GADTs, KindSignatures #-}
main = putStrLn "ok"
#if 0
#endif
a = 'a--'
b = 5 `f `7`f`"3 'ab'"
f = f
data A' a' :: * where
  B :: a -> A' a
class Num a => A a where
```

---

This is perfectly valid code, you can [try it online](https://tio.run/##JYrLCsIwFET39yuGdhGQBgQRQVCoDyooIqj7Xm2ioW0sTYoL8dtr1NXMnDN3dqWqqr5/yRi7dJ@d02yNk6qbir3a/G2C5eGQIEtXJ5dga2xxNDfLvmuVQyzfVLOxmKHp/NG3O4voUUYUG40hxcoWRhMHLVhKQZfQxsg18kmu82gU8EVEpAPWVLBnpAIsMJ1igOddtYqAxXcy5Pwn6Vqxc9h3dWCzwEL8nn3/AQ).

Issues:

* Lines 3,4: directives `#if` and `#endif` not highlighted
* Line 5: templated `a` followed by a comment is wrongly highlighted as a literal character
* Line 6: the two `'` that appear in the literal string are highlighted as a literal character
* Line 6: infix function call `f` not highlighted, but `7` mishighlighted as operator
* Lines 8,9: the `'` not highlighted as part of `A`
* Line 8: the two `'` are mishighlighted as a literal character
* Line 10: the `=>` context operator not highlighted

### LINKED ISSUE #2918: [Haskell] multi-line string match
- Sublime Version: 4112
- OS Version: Windows 10

``` haskell
testString = "This is a\
             \multi-line string literal\
             \example"
```

image

The modification below could solve this problem
```YAML
        - match: \\\s
          captures:
            constant.character.escape.begin.haskell
          push:
            - match: \\
              captures:
                constant.character.escape.end.haskell
              pop: true
```
image

After Modifiying
image

deathaxe:
Fixes #1321
Fixes #2670 
Fixes #2672 
Fixes #2918

Superseeds #2628, #2662, #2671

This PR is the result of some spare time caused by pendemic stay-at-home orders in our country.

With a birds eye on existing Haskell syntax definitions of other popular editors and  a quick study of [Haskell 2010 Report](https://www.haskell.org/onlinereport/haskell2010/haskellpa1.html#haskellch4.html) the proposed syntax definition evolved from the existing package. It includes all commits of #2671 and the changes from #2628.

_The major goal is to provide an as simple as possible and robust syntax definition._

### Layouts

Haskell supports two kinds of layout schemes 
a) a C style block quote layout with braces and `;` statement terminators.
b) a python like indentation based layout.

The latter one makes it hard to find proper statement and expression boundaries and seems to be the root cause of broken syntax highlighting with existing implementations.

This is resolved by simplifying contexts and scope names to a minimum.

### Patterns

Another issue with existing syntaxes is pattern matching itself. It appears none of them has proper support for unicode identifiers or unicode operators in various situations. That's addressed by this PR as well.

### Disclaimer

I am not an active Haskell developer. Thus this PR should probably be thoroughly reviewed by an Haskell expert to make sure all basic language constructs work as expected.

### Performance

File | lines | Before | After | Diff
---|---|---|---|---
[syntax_test_haskell.hs][1] | 4096 |  10ms  | 14ms | +40%
[LaTeX.hs][2] x 6 |   12070 |   53ms |  130ms | +145%
[Parsing.hs][3] x 6 | 9444 |  44ms | 83ms | +88%

The new syntax is about 2x slower depending on use case, but that doesn't surprise much as the old one holds only 407 lines of code, while the new one consists of 1388 lines, with a lot of more stuff being considdered. It uses branching to distinguish groups from tuples and to scope type contexts in order to be able to highlight declared type identifiers correctly. The list of scoped builtins is larger etc.

[1]: https://github.com/sublimehq/Packages/blob/126e65b1572d2699ec92326211a727011b32094d/Haskell/syntax_test_haskell.hs
[2]: https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/LaTeX.hs
[3]: https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Parsing.hs


mitranim:
A welcome development. Not equipped for a full review since I only dabble, but took a cursory look through https://github.com/ghc/ghc and spotted a few things.

`import A` scopes `A` as `entity.name.namespace`, introducing it into scopes. I believe this to be inaccurate, as this doesn't add `A` to the current scope (unlike Swift's `import`). That would be `import A as B`, where only `B` is in scope.

`module A (module B.C)` scopes `C` as `storage.type`, unless it's `support.type.prelude`. I believe both to be inaccurate, as those are modules (?).

Some module-related keywords, like `as|qualified|hiding`, are scoped as `storage.modifier`. I could be mistaken, but this seems unusual to me. Wouldn't they usually be keywords?

Type scoping seems to prioritize built-ins over general principles to the exclusion of the latter. For example:

```hs
relativePackageDbPath :: Stage -> FilePath
```

`Stage` is correctly scoped as `storage.type`, but `FilePath` is incorrectly not scoped as `storage.type`, instead receiving `support.type.prelude`.

In imports and exports, function names are scoped as `variable.function`, despite this not being supported for actual function calls. While somewhat logical, I'm under the impression that `variable.function` has become the de-facto scope for using functions as calls rather than values, and sometimes using it for values rather than calls is bound to cause confusion.

Declared functions are scoped as `entity.name.function` only preceding `::`, but not `=`. While understandably much simpler, this means that functions with inferred signatures are not indexed. I can't confidently say how difficult this would be to support, or how good/bad in practice, just that it seems like an omission on a cursory inspection.

Declared types such as `data A` are scoped as `meta.declaration storage.type` and not indexed. While consistent with the recently discussed/proposed idea for declared names, this is inconsistent with the established convention of `entity.name.type`, and/or seems to lack the appropriate symbol config file.

Named field getters are lacking the `entity.name.function` scope:

```hs
data A = A { getB :: B, getC :: C }
```

Some built-ins receive bizarre scopes. While scoping `True|False|Nothing` as constants is understandable, scoping `Just|Left|Right` as `support.constant` seems incorrect, as those are unary constructors. I would propose dropping such special cases.

Declared operators are scoped and indexed as `entity.name.function` together with their parens, for example `(*>)` rather than `*>`, which prevents `goto_definition` from working.

The overuse of `constant.other` in compiler pragmas seems strange to me, but I might simply be unfamiliar with pragma conventions.

`\` and `->` in lambdas are scoped as `keyword.operator` while I believe the more recent approach would be `keyword.declaration`, like in JS.

mitranim:
Addendum: modules imported via `import A` are scoped as `entity.name.namespace`, causing them to be indexed globally, even though this imports them only into the scope of the current file, not the entire project. I believe it would be more convenient if imported names were indexed only locally, for the current file. This could be done by changing the scope, or excluding them from the global index by detecting the `meta.import` scope.

deathaxe:
> import `A` scopes `A` as `entity.name.namespace`, introducing it into scopes. I believe this to be inaccurate...

1. It actually does add `A` to scope so all exported functions and types are available without qualifier. 
 
   If `A` exports `fun` you can directly use `fun` after `import A`.

2. What you describe is `import qualified A`, which does not add `A` to scope, thus any function would need the qualifier (`A.fun`).

3. `import A as B` actually adds `A` aliased as `B` very much like python handles that.

4. `import A (fun)` imports only `fun`from `A` with adding `A` to scope so that `fun` is available without qualifier very much like (1.)

We might argue about not scoping `A` `entity` in case (2.) and (3.). I actually tried that, but finally decided to keep things simple and removed that distinction. It also felt strange and broken to see the leaf part of imported identifiers highlighted differently.

---

> Some module-related keywords, like `as|qualified|hiding`, are scoped as `storage.modifier`. I could be mistaken, but this seems unusual to me. Wouldn't they usually be keywords?

I don't have a strong oppinion about it, but just followed the results of a recent discussion about `extends`, `implements`, ... keywords in Java class declarations, which ended up in the request to keep those scoped `storage.modifier` instead of using `keyword.declaration`.

We may argue about `as`, which might be scoped `keyword.declaration` as it declares a new name for the imported module, but it also feels not too  bad to be a modifier as it modifies the imported scope name.

`qualified|hiding` modify the import operation and thus feel scoped correct.`

---

> `module A (module B.C)` scopes `C` as `storage.type`, 

That's indeed wrong. 

---

> Stage is correctly scoped as `storage.type`, but FilePath is incorrectly not scoped as `storage.type`, instead receiving `support.type.prelude`.

This is intended as both scopes denote data types, which we don't want to stack on top of each other. A user defined data type receives `storage.` while a builtin one uses `support.`, but not both at the same time. 

We had this discussion on discord already with regards to user defined vs. builtin functions.

A color scheme can easily address both the same way using `"scope": "support.type, storage.type"`.

---

> In imports and exports, function names are scoped as `variable.function`, ...

I followed the concept used in Erlang as well. Means the export list contains references to the functions defined in the module and thus uses `variable.function`. Lazily reused those contexts in imports. This may need some refinement.

With your arguments it would generally mean to either scope them `entity.name.[export|import]` or just `variable.other`.

---

> Declared functions are scoped as entity.name.function only preceding `::`, but not `=`. 

Are all `ident .... =` function declarations? That is not very clear to me, whether it is a function or variable assignment.

---

> Declared types such as `data A` are scoped as `meta.declaration storage.type` and not indexed. 

With a look at more sophisticated statements (same with classes) it is very unclear to me, which of those identifiers is really the new data type identifier which needs to receive `entity`. It might be simple with `data A`, but things get complicated quickly if `context =>` comes into play (see: https://www.haskell.org/onlinereport/haskell2010/haskellch4.html#x10-680004.2)

---

> scoping `Just|Left|Right` as `support.constant` seems incorrect, as those are unary constructors. I would propose dropping such special cases.

I'd move them to `support.type` then.


---

> Declared operators are scoped and indexed as `entity.name.function` together with their parens, for example `(*>)` rather than `*>`, which prevents goto_definition from working.

This is the case with all existing syntax definitions. All of them scope the whole token including parentheses `entity.name.function` and I just kept using that due to lack of better alternatives and the lack of knowledge whether the parentheses have certain meaning. It would be easy to just scope the parentheses as `punctuation` if that feels better.

---

> The overuse of `constant.other` in compiler pragmas seems strange to me, but I might simply be unfamiliar with pragma conventions.

What are those pragmas? They are simply key value pairs in most cases. What scope would you use instead? They feel quite constant to me.

---

> `\` and `->` in lambdas are scoped as keyword.operator while I believe the more recent approach would be keyword.declaration, like in JS.

Those operators are not handled in a special way at all by this syntax.

This is by intention as especially `->` is used in various ways. Trying to distinguish those types would require to (re-)introduce a lot of context switches, probably bloat the syntax and introduche many edge cases.

That said, I find scoping `->` as `keyword.declaration` totally wrong and a mess at all, no matter which syntax.

mitranim:
Regarding `import A`, we might have a minor misunderstanding or lexicon mismatch. I also didn't phrase it well. Here goes try 2. As you noted, this statement acts as an "include", splurging all definitions from module `A` into the current scope. However, it doesn't declare the namespace `A` in the current scope; only the statement `import A as A` would have done that.

My expectation for symbol indexing is that it should index only declared names that are actually in scope. The global index should contain symbols that are project-wide or exported by some files. The local index should additionally contain symbols available in the scope of the current file. Arbitrary identifiers that aren't actually in scope shouldn't be indexed. This would mean that `import A` shouldn't cause `A` to be indexed. I wouldn't be surprised to find that some people want other things to be indexed as well; this is just my intuitive expectation, and I feel like it's a solid default. 🤷‍♀️

Regarding `as|qualified|hiding`. My intuitive understanding of `storage` is that it's best used for identifiers and operators that denote the storage properties of a declared name (such as a variable, function, type, or parameter). Storage properties are things like its memory layout, visibility (public/static), mutability (constant/mutable), volatile, C static, and so on. These modifiers tend to be an open set. Obviously there's an infinity of types, but even the keyword modifiers aren't always fixed. The C and C++ standards have accumulated more of them over time. Meanwhile, words like `as` and `hiding` are a fixed closed set, and a part of the language's syntax. Under this view, I would lean towards them being more keyword-like than storage-like, and I hope this reasoning sheds some more light on why.

Regarding `storage.type` and `support.type`. I know there's been a lot of discussions already, but I don't recall achieving consensus about keeping them mutually exclusive. There are objective arguments in favor of making built-ins purely additive. You're right that color schemes can choose to scope `storage.type` and `support.type` the same way, but one of many problems is that `support.type` is often applied where `storage.type` wouldn't be and shouldn't be. For example, in `A.B`, `A` might be scoped as `support.type`, even while acting as a namespace, not a storage modifier. There are other cases as well. Of course not everyone cares about scoping tokens mainly by their role, rather by what's "well-known", but as someone who does, I'd appreciate supporting that use case.

Regarding function declarations. Haskell allows to omit function signatures. The following are all perfectly valid, complete, self-sufficient function definitions:

```hs
main = do
  one
  two

one = print 10

two = print 20
```

It should be noted that `=` is also used in `where` and `let` clauses, all of which declare local functions. Unlike functions in file root, these probably shouldn't be indexed; or at least it should be possible to toggle.

```hs
main = do
  let prn a = print a
  prn 10
  prn 20

main = do
  prn 10
  prn 20
  where prn a = print a
```

In operator declarations, parens are definitely just punctuation. It's Haskell's way of allowing you to use an operator in the position where a normal identifier would be expected.

Regarding compiler pragmas, GHC is full of stuff like this:

```hs
unless            :: (Applicative f) => Bool -> f () -> f ()
{-# INLINABLE unless #-}
{-# SPECIALISE unless :: Bool -> IO () -> IO () #-}
{-# SPECIALISE unless :: Bool -> Maybe () -> Maybe () #-}
unless p s        =  if p then pure () else s
```

Basically they may contain types, functions, operators, all sorts of things. Not sure how much sense it makes to scope these as constants.

mitranim:
Regarding type declarations.

Haskell likes to pretend that `data` declares a single type and 1 to N "constructors". It tries as hard as possible to pretend that these "constructors" aren't types, even going as far as forbidding you from using them in type positions.

What `data` actually does is to declare a single "sum type" and 1 to N "product types" that belong to the union. "Sum type" is also known as "union type" or "variant type". "Product type" is basically a tuple. The mandatory "constructor" allows to pattern-match and deconstruct values of those types.

For example:

```hs
data Maybe a = Nothing | Just a
```

`Maybe` is officially a type and should be indexed as such. `Nothing` is simultaneously a nullary type, equivalent to `()` (Haskell pretends that it's not), and a nullary value constructor, usable where a constructor or value would be expected. `Just` is a simultaneously a generic type with a single parameter (though Haskell won't let you use this in type positions, requiring `Maybe a` instead), and an unary value constructor. All three should be indexed, `Maybe` should be scoped as `entity.name.type` when declared, and as `storage.type` in type positions. `Nothing` and `Just` should probably be scoped either as `entity.name.type` or `entity.name.function` when declared. Haskell is really hardcore about pretending that the individual variants of a sum type are just functions, so it wouldn't be too strange to reflect this in the syntax.

The `A b =>` modifiers should probably be treated like `implements`. Something like:

```hs
   class A b => C d where
-- ^^^^^ keyword.declaration
--       ^ entity.other.inherited-class
--         ^ variable.parameter.type
--              ^ entity.name.type
--                ^ variable.parameter.type
```

Hope it's helpful!

deathaxe:
Thank you very much for your imports. I guess it will take a while to go through it to see what can be implemented without trying to add contexts which are hard or even unable to maintain because nearly anything can be stacked into each other without clear boundaries of where an expression stops. Using initial indentation to try to create blocks of code is IMHO too error prone and hard to achieve.

Your code snippets and examples help very much understanding Haskell better. I am just concerned about handling more sophisticated statements containing contexts, guards and patterns reliably. This applies especially to guarded function definitions, which may probably look as followes ? 

```
fun a b (b c) 
    | pat <- expr, expr, expr = funexpr1
    | pat2 <- expr expr = funexpr2
    | let var = decls = funexpr3
   where decls
```

This is how I interpret [4.4.3 Function and Pattern Bindings](https://www.haskell.org/onlinereport/haskell2010/haskellch4.html#x10-800004.4)

mitranim:
Hmm I see what you mean... pattern matching and guards seem like they could make it unreasonably hard to _reliably_ detect all declared functions and operators, not just the names preceding `::`. The Haskell compiler manages, but that doesn't necessarily mean it's reasonable to support in a `.sublime-syntax`.

> I also guess that `scope as .. in type positions` won't be possible with satisfying quality.

Not entirely sure what you mean by this. Unlike the function name problem, type positions should be relatively easy to detect, at least to my cursory knowledge of the language. Types mostly occur in type declarations and after `::`. Of course there are probably a lot of gotchas I'm not considering, especially considering the lack of explicit delimiters...

mitranim:
Apologies for messaging about code at New Year when everyone should be hung over. 🙂

Let me observe something from the earlier discussion:

> > Stage is correctly scoped as `storage.type`, but FilePath is incorrectly not scoped as `storage.type`, instead receiving `support.type.prelude`.
> 
> This is intended as both scopes denote data types, which we don't want to stack on top of each other. A user defined data type receives `storage.` while a builtin one uses `support.`, but not both at the same time.
> 
> We had this discussion on discord already with regards to user defined vs. builtin functions.
> 
> A color scheme can easily address both the same way using `"scope": "support.type, storage.type"`.

A few paragraphs later:

> > scoping `Just|Left|Right` as `support.constant` seems incorrect, as those are unary constructors. I would propose dropping such special cases.
> 
> I'd move them to `support.type` then.

If constructors everywhere were scoped as `storage.type`, this would be a non-issue, as a color scheme could just color them all as types. But if constructors in non-type positions are not `storage.type`, scoping _some_ of them as `support.type` will force the color scheme to choose between:

  * Coloring them as non-types in type positions (inconsistent with other types).
  * Coloring them as types in non-type positions (inconsistent with other constructors).

In addition, `Just|Left|Right` are not usable in type positions due to Haskell's idiosyncrasy. But that's not even pertinent to the main point, which is the following (quote from my later post):

> You're right that color schemes can choose to scope `storage.type` and `support.type` the same way, but one of many problems is that `support.type` is often applied where `storage.type` wouldn't be and shouldn't be.

Making support scopes mutually exclusive with role scopes always tends to lead to these kinds of misconsistencies, because support tends to be detected by name, and role by position. They don't fully overlap.

Hope this makes sense!

---

Skimmed GHC some more and got reminded about the [magic hash](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/magic_hash.html):

```hs
shiftL# :: Word# -> Int# -> Word#
a `shiftL#` b   | isTrue# (b >=# WORD_SIZE_IN_BITS#) = 0##
                | otherwise                          = a `uncheckedShiftL#` b
```

The `MagicHash` language extension allows `shiftL#` to be a valid identifier which should be `entity.name.function` in this particular case. It might be useful to support identifiers ending with `#` to correctly scope and index such functions and types. Admittedly it complicates identifier parsing, but `\b` already wasn't usable because Haskell allows `'` in identifiers.

---

For operators enclosed in parens, while it's nice that parens are now punctuation, the operators seem to no longer be scoped as `entity.name.function` where appropriate, which prevents them from being indexed:

```hs
  (<$) :: a -> f b -> f a
-- ^^ should be entity.name.function, but isn't!
```

I apologize if all of this gives you more you bargained for. I suppose this is what you get when you let a committee of mathematicians design a syntax. Renews my appreciation of the Lisp syntax. Ironic that they both come from math.

deathaxe:
> I apologize if all of this gives you more than you bargained for.

Not at all. You help me much understanding the syntax better.

> I suppose this is what you get when you let a committee of mathematicians design a syntax.

I don't even want to try to express what I think about this syntax ... . I guess I would be banned from Github due to the amount of bad words. It just doesn't feel to be made neither from nor for the real world.

> ... operators ...  which prevents them from being indexed

That's an easy fix even without the current scope. See _Symbol List - Operators.tmPreferences_. I think it's useful to index those as `keywords` rather than `funcions`. Hence scope is ok. I am just in doupt with how much sense it makes to add those to the index as that means them to be globally available even though not exported by module declaration. We'll end up in anyithing looking like a function being added to the global index database.

It's basically the same issue as with imported namespaces being indexed atm, which makes them globally available without any use.

But' that's a general issue about ST's indexer, which just isn't capable to handle such internal/external module architechtures with limited export lists. The point is, you can't index locally only.

I think we should somehow add only things to the index, which are part of an export list in a module declaraion.

Not sure how to do so as ST's default _Indexed Symbol List_ is just too generous to disable it for a single syntax without disabling it at all. At least I am not aware of a way to do so. I remember of some _Symbol Ignore Lists_ but haven't found them anymore.

deathaxe:
> If constructors everywhere were scoped as storage.type, this would be a non-issue, as a color scheme could just color them all as types. But if constructors in non-type positions are not storage.type, scoping some of them as support.type will force the color scheme to choose between:

I don't get this point. The syntax does not distinguish between types and constructors neither by identifier nor by position/context, so I don't see the issue.

Known builtin (prelude) types and constructors are scoped `support.type`. I've choosen `support.class` for those known to be classes because we are asked not to scope them `support.type.class` which would probably be more consistent here. We could easily rename them to just use `support.type` as well.

All user defined classes/types/constructors/data/... are scoped `storage.type`.

The only exceptions are `Nothing | True | False | LT | EQ | GT` which still remain in constant. If we don't want them to have a special scope they can be moved to `support.type` too.

---

The point is, we won't be able to create enough context in order to be able to reliably distinguish those tokens' types depending on position they are used in. The syntax architechture just doesn't allow us to easily push/set into contexts and pop reliably. This will result in many edge cases, which may break highlighting depending on users programming style and habits.

In other words - the semantic highlighting you have in mind from your understanding of the language itself is not possible with ST's syntax engine or would at least be too complicated and error prone to try to.

None of the existing syntaxes, this old one, SublimeHaskell, nor VS Code's one is doing well in such things even though they try. According to the issues of the latter one I learned, the more they try to create contexts the more they fail. 

A basic syntax shipped with ST core should do it's best to highlight most important stuff, while keeping easy to maintain and robust to not need to fix edge cases every day.

mitranim:
Regarding functions and operators. It should be noted that in Haskell, operators _really_ are just functions. The only difference is that functions whose _name_ is syntactically an operator are usable directly in infix position, and require `()` for the prefix position. Functions whose _name_ is syntactically an identifier are usable directly in prefix position, and require backticks for the infix position. That's the only difference. You can alias an operator as a function, or vice versa:

```hs
main = do
  print ([10] ++ [20])
  print ([10] @@ [20])
  print (append [10] [20])
  print ([10] `append` [20])

append = (++)

(@@) = append
```

This also means that the notion of "operators" as "keywords" because they involve special syntax (infix position) applies to operators _and_ regular identifiers in backticks, but does not apply to operators in parens, because parens make them operate as regular identifiers. The current version of your syntax recognizes this in one direction, by scoping backticked infix functions as `keyword.operator.function`. I believe it would be logically consistent to mirror this in the opposite direction, by treating parenthesized operators like normal identifiers in every respect.

Regarding the indexing of functions and operators. I agree that identifiers local to the current file should be indexed only locally, while exported identifiers should be indexed globally. Unfortunately, Haskell's syntax doesn't allow us to detect this reliably. Because operators are functions, with the same visibility, I don't see any solid argument why they shouldn't be indexed the same as functions. Traditionally, this means indexing them globally if they're declared in root scope. The current version indexes functions globally but operators locally, which seems logically inconsistent to me.

Regarding constructors. Because all constructors are also types, it's probably good enough to scope them all as types, regardless of context. I think we can agree on that. Personally I would lean towards trying to scope types only contextually, but I'll acquiesce to your judgement of the difficulties.

Since you mentioned classes, it should be noted that Haskell's typeclasses are actually interfaces!

deathaxe:
>  I believe it would be logically consistent to mirror this in the opposite direction, by treating parenthesized operators like normal identifiers, at least when declared.

This is what they used to be. That's why I asked whether those parentheses have special meaning which I understood you declined. That's why I turned those into normal groups using `punctuation` for parentheses.

What you write here sound very much like the opposite.

Please don't misunderstand it as too negative. I can follow your idea and intention and understand what you mean.

Regardless the special meaning I wonder what's the best and most consistent solution from highlighting perspective.

To be honest, I am not very happy with scoping backticked functions as `keyword.operator` because - as you mensioned - they can also appear in normal function position. That's what I learned with `Quux` in pandoc's code base. The token is sometimes highlighted as `storage.type` and sometimes as operator.

IMHO, we should decide for one of

a) revert parenthesed operators to be scoped as functions again
b) modify backticked functions' scope to normal `variable.function` or `storage.type` depending on first character's case.

I have no strong opinion about which solution is the better one.

mitranim:
My apologies for being unclear. By treating parenthesized operators like regular identifiers, I mean the following:

```hs
   (++) :: [a] -> [a] -> [a]
-- ^  ^ punctuation.section
--  ^^ entity.name.function (negotiable)

   (++) [10] [20]
-- ^  ^ punctuation.section
--  ^^ keyword.operator | variable.other (negotiable; probably keyword?)
```

The old syntax would treat the _entire_ `(++)` as the symbol, but it's not. The `++` is the symbol. The parens aren't part of it, they're external syntax that changes in which syntactic position the symbol `++` can be used. Scoping just the `++` allows things like `goto_definition` to work properly:

```hs
[10] |++ [20]
     ^ this is the caret; activate `goto_definition`, go to `(++) :: ...`
```

> To be honest, I am not very happy with scoping backticked functions as `keyword.operator` because - as you mensioned - they can also appear in normal function position. That's what I learned with `Quux` in pandoc's code base. The token is sometimes highlighted as `storage.type` and sometimes as operator.

This sounds like a misunderstanding. Surrounding an identifier with backticks turns it into an infix operator, from Haskell syntax's perspective. Technically, they can still appear in the prefix position when performing partial application. But that's actually where different coloring could be important. Consider:

```hs
main = do
  print (ten_with [20]) -- [10,20]
  print (with_ten [20]) -- [20,10]

join     = (++)
ten_with = (join [10])
with_ten = (`join` [10])
```

`ten_with` performs partial application on the _first_ parameter, while `with_ten` performs it on the _second_ parameter. When backticked functions appear in the prefix position, they still act like operators, so scoping them as such makes a lot of sense and may reduce gotchas.

deathaxe:
I guess `goto_definition` doesn't work with `(++)` no matter which scope we use. 

While goto definition works for your 2nd example

```Haskell
[10] ++ [20]
```

ST doesn't find the definition for...

```Haskell
(++) [10] [20]
```

It appears it always searches for `(++)` even if only `++` is scoped variable or whatever. It appears some internal ST core logic prevents that from working.

mitranim:
`goto_definition` uses the global symbol index, but whether it looks for `(op)` or `op` depends on the cursor position. Using `|` to emulate cursor position, consider these:

```hs
-- Incorrectly looks for " ++".
[10] |++ 20

-- Incorrectly looks for "(++)".
(|++) [10] [20]

-- Correctly looks for "++".
(+|+) [10] [20]
```

You're right that ST's heuristic isn't perfect here. Might be worth opening an HQ issue...

In either case, the change that allows global indexing of operator declarations lets the user to search for them manually _or_ to use `goto_definition` with the right cursor position, so it's definitely worth keeping. I personally would simply prefer `entity.name.function` there, but that's probably less important than indexing working. 🙂👍

mitranim:
> IMHO only module exports should be indexed globally.

We might have differing notions of what the symbol index is for. Personally, I find it tremendously useful when `goto_definition` or ⇪⌘R goes directly to the _definition_ of the given function, type, or global variable. Currently in ST, this works only with the global index. If the given definition is private to the current file and available in the local index, `goto_definition` won't consider it, and ⇪⌘R won't add the locals to the list. And the user doesn't always know whether the definition is in the current file or not, so they'd need several attempts.

I understand that indexing only exports has some logic behind it. It avoids "polluting" the global symbol index with symbols private to those files. So, using `goto_definition` from another module, the user would be taken to that name in the export list of the other module, then use the _local_ index to find the definition. Doable, but a two-step process. Also, sometimes you don't have an export list at all! The "main" module of a program usualy has no export, yet could be large enough to warrant `goto_definition` and symbol search. I would argue that indexing exports instead of definitions would be an inferior default. And if we index definitions, indexing exports is probably not useful in the vast majority of cases.

mitranim:
Thinking about it some more, I actually appreciate that the global symbol index includes definitions private to other files, instead of being "smart" and restricting the search to only what's visible in the current scope. Sometimes you want to search among all definitions, not just what's in scope. And `goto_definition` being somewhat "dumb" and not particularly IDE/LSP-like can be handy too; for example, it works in comments, can refer to private functions mentioned in comments, and so on.

deathaxe:
It was Erlang when I studied Symbol Lists vs. Index the last time. This is also what I took as reference for scoping exports as `variable.function` and `storage.type` in the beginning. Turned to general `entity.name.[export|import]` as it was something general proposed by Will, IIRC.

In general Erlang has the same module like architechture with only defined stuff being explicitly exported. Therefore I try to follow its scheme of symbol definitions and indexing closely. It indexes all exports as well.

You are absolutely right with `goto_definition` to target the real definition instead of the export declaration being the best, maybe even both, not sure.

jrappen:
@deathaxe If you could add performance information to the leading comment, that would be nice.

moodmosaic:
I'd be very interested to help by testing this. I just tried it with https://github.com/well-typed/integrated-shrinking/blob/3f0c768f12b1bb9c839a20dfe8121475f5f98008/src/IntegratedShrinking.hs but doesn't seem to work well.
* All imports are shown in plain text (no colors)
* Everything else under imports is shown as comments

moodmosaic (commented):
I'd be very interested to help by testing this. I just tried it with https://github.com/well-typed/integrated-shrinking/blob/3f0c768f12b1bb9c839a20dfe8121475f5f98008/src/IntegratedShrinking.hs but doesn't seem to work well.
* All imports are shown in plain text (no colors)
* Everything else under imports is shown as comments

deathaxe:
You need a recent alpha-development built of ST4 to use the syntax of this branch because it makes use of several new syntax engine features. Alpha builds are available on discord for those who are willing to help finding bugs.

Imports are scoped `variable.namespace` which is meant as counter part for `entity.name.namespace` used in `package` statements. The former may not yet be addressed by color schemes. Syntaxes used to use several different and often uncommon scopes for qualifiers before.

Here's how your file looks with Mariana.sublime-color-scheme

grafik


I personally use a slightly tweaked Breackets Dark

grafik

moodmosaic:
Thank you. I'm not on discord... could you send me a link to the particular forum/thread (and then I can proceed with sign-up, etc).

deathaxe:
The invite link is listet at the bottom of the first post at https://forum.sublimetext.com/t/resources-and-bug-tracking/22985

moodmosaic:
I don't see it 

image

deathaxe:
Than just `ctrl+f` and search "Discord" :-D Or have a look at the middle of your screenshot. ... under "Discussion Elsewhere" ;-)

moodmosaic:
Oh, I found it, thanks!

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
-      captures:
-        1: constant.numeric.value.haskell
-        2: punctuation.separator.decimal.haskell
-        3: constant.numeric.value.haskell
-        4: constant.numeric.value.exponent.haskell
-        5: constant.numeric.value.exponent.haskell
-    - match: '{{operator_infix}}'
+    - match: '{{operator_double_colon}}'
+      scope: keyword.operator.double-colon.haskell
+    - match: '{{operator_symbol}}'
```
FichteFoll, line 1338: I'd appreciate a separate match for the `=` assignment operator.

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
-    - include: operator_paren
 
-  operator_paren:
-    - match: \(({{operator_parens}})\)
-      scope: variable.function.infix.haskell
-      captures:
-        1: keyword.operator.haskell
+  big-arrow-operators:
+    - match: '{{operator_big_arrow}}'
+      scope: keyword.operator.big-arrow.haskell
```
FichteFoll, line 1343: While big arrows are (afaik) only used in function signatures, I find it hard to determine the correct scope for them. They are separators, but like the normal arrows, they separate syntactically different segments where it would probably make sense to reflect that in a different scope name that doesn't differ in the 4th subscope.

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
-      scope: meta.number.float.decimal.haskell
-      captures:
-        1: constant.numeric.value.haskell
-        2: punctuation.separator.decimal.haskell
-        3: constant.numeric.value.haskell
-        4: constant.numeric.value.exponent.haskell
-        5: constant.numeric.value.exponent.haskell
-    - match: '{{operator_infix}}'
+    - match: '{{operator_double_colon}}'
+      scope: keyword.operator.double-colon.haskell
```
FichteFoll, line 1337: I wouldn't treat the double colon as a normal operator either, more like a `keyword.declaration`.

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
+    - include: tuple-end
+    - include: records
+    - include: expressions
+    - include: else-pop
+
+  empty-tuples:
+    # 3.9 Unit Expressions and Parenthesized Expressions
+    # https://www.haskell.org/onlinereport/haskell2010/haskellch3.html#x8-380003.9
+    - match: (\()\s*(\))
+      scope: meta.sequence.tuple.haskell
```
FichteFoll, line 1110: We talked in https://github.com/sublimehq/Packages/pull/2387 whether an empty tuple is supposed to be a language constant where they have special meaning (notably in functional languages), but I while we seemed to have general agreement on this, the PRs haven't been merged yet. The description in the linked reference also supports this.

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
+    - include: sequence-separators
+    - include: type-expressions
+    - include: else-pop
+
+###[ BRACES ]##################################################################
+
+  blocks:
+    # 2.7 Layout
+    # https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-210002.7
+    - match: \{
```
FichteFoll, line 985: This also matches records, which I'm not sure is fine to be matched by it or not.

https://devtut.github.io/haskell/record-syntax.html

FichteFoll (commented):
These are great changes. Thanks for taking the time to rewrite this syntax more thoroughly than I attmpted last year.

I didn't do a full review of everything, since I haven't touched many of Haskell's aspects myself yet, but I did have some concerns regarding scoping of operators/puncutation/etc. I also ran this against a couple of files I wrote back then which by far didn't use the full feature set, but were enough to paint a general picture. (After I had removed highlighting for `variable.other` scopes in my color scheme. That's a topic for another time.)

Also to be noted is the accepted but not yet implemented RecordDotSyntax extension that will allow `record.field` instead of `field record`, but that can be done later, of course.
https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0282-record-dot-syntax.rst

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
-    # GHC permits the C preprocessor to be run on a source file.
-    - match: ^\s*(#)\s*\w+
-      scope: meta.preprocessor.c
+  left-arrow-operators:
+    - match: '{{operator_left_arrow}}'
+      scope: keyword.operator.arrow.haskell
+
+  right-arrow-operators:
+    - match: '{{operator_right_arrow}}'
+      scope: keyword.operator.arrow.haskell
```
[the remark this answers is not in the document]
deathaxe: I started to implement lambdas by specially scoping backslash followed by `variable.parameter` and also scoping the arrow in a special way. I rejected that afterall as that would have been the only expression with `variable.parameter` and finally decided to keep all kinds of operators colored the same way, too, as their meaning may differ by context as you already mension without a sane way to detect it due to language design.

I am not sure whether it would look off, if we'd scope the backslash `punctuation.definition` if parameters remain `variable.other`, but I don't have a very strong meaning about it.

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
+      scope: keyword.operator.big-arrow.haskell
 
-  preprocessor:
-    # In addition to Haskell's "native" syntax,
-    # GHC permits the C preprocessor to be run on a source file.
-    - match: ^\s*(#)\s*\w+
-      scope: meta.preprocessor.c
+  left-arrow-operators:
+    - match: '{{operator_left_arrow}}'
+      scope: keyword.operator.arrow.haskell
```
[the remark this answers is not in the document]
deathaxe: Probably yes, but just kept using arrow as it is used for the right arrows, too. In general the whole concept of semantic operator scopes is not applied due to the nature of those.

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
-        (')
-        (?:
-          [\ -\[\]-~]            # Basic Char
-          | {{escape_sequence}}  # Escapes
-        )
-        ([^']*)
-        (')
-      scope: string.quoted.single.haskell
+  infix-quoted-operators:
+    - match: (`)\s*([\w'.]*(#*))\s*(`)
```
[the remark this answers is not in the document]
deathaxe: Already thought about it, yes, but honestly was too lazy to rework those at this point, because we'd need to push into context then in order to correctly scope all `.` as accessor etc.. Or is there only one level of namespace?

Haskell/Haskell.sublime-syntax
```diff
@@ -1,407 +1,1388 @@
...
-      scope: keyword.other.big-arrow.haskell
-    - match: '\b[a-z][a-zA-Z0-9_'']*\b'
-      scope: variable.other.generic-type.haskell
-    - match: '\b[A-Z][a-zA-Z0-9_'']*\b'
-      scope: storage.type.haskell
-    - match: \(\)
-      scope: support.constant.unit.haskell
+  sequence-separators:
+    - match: ',|\|(?!\|)'
+      scope: punctuation.separator.sequence.haskell
```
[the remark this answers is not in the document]
deathaxe: The syntax doesn't push into those contexts to be able to distinguish that. It was just too error prone.

Haskell/Haskell.sublime-syntax
deathaxe: We could use `punctuation.separator`. That's what Erlang does at least.

Haskell/Haskell.sublime-syntax
deathaxe: I don't want to introduce another ambiguity for word vs. symbol like keywords and therefor also don't like that scope for lambda arrows to be honest.

Haskell/Haskell.sublime-syntax
deathaxe: Also here, the syntax can't distinguish those constructs by design.

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: I certainly see value in semantically highlighting syntax operators like these arrows. Currently, this happens for the type annotations and declarations, so it would be a regression in that regard, but the others are not and can be considered suggestions or feature requests for a new ticket.

Same for `<-`.

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: > Or is there only one level of namespace?

Not confident enough to answer this, so most likely we would need a subcontext here with a look-head or branch, depending on what is easier.

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: I don't think a single pipe character is valid anywhere else. Commas we can easily defer.

Haskell/Haskell.sublime-syntax
FichteFoll: Have we had an RFC for this?

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
deathaxe: Semantic highlighting requires knowledge about context, but context boundaries are hard to maintain due to the nature of Haskells syntax. As python it relies on indentation to detect end of blocks. Even the old syntax did not correctly maintain type contexts in all situations, only for function signatures.

In order to not need to try to handle error prone context push-pops most semantic highlighting was removed with regards to operators and also type variables.

You may call it a regression, but it at least provides scope consistency for operators and variables all over the syntax in all situations even though less different scopes are used.

I tried to implement such things but stopped when I realized all kinds of implementation would end up in certain code styles to be needed for the syntax to work. And that's what I've learned from following VS Code's syntax.

Yes, semantic highlighting is nice, but not if it becomes the root cause of aweful heavy highlighting issues. That's not worth it.

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
deathaxe: They are separators in mappings if I don't miss something. That's one major reason for the scope decision.

https://github.com/sublimehq/Packages/blob/01624ba13f8fe351184f119a70c49836da2a7130/Haskell/tests/syntax_test_haskell.hs#L1295-L1344

Otherwise they are separators of a list of guarded expressions

https://github.com/sublimehq/Packages/blob/01624ba13f8fe351184f119a70c49836da2a7130/Haskell/tests/syntax_test_haskell.hs#L2068-L2085

Erlang also has that concept of guarded clauses/expressions which I use `punctuation.separator` for, too, in order to scope the separator which is `;` in that case. Erlang even calls it a list of guarded clauses. Hence my impression is that this concept works for Haskell, too.

At least it helps to distinguish normal operators used in guard expressions from guard boundaries a bit better.

Haskell/Haskell.sublime-syntax
deathaxe: There were several discussions about word vs. symbol operators in pending PRs for instance.

Disuccsion about lambda function declarations was at  https://github.com/sublimehq/Packages/pull/2617

Didn't recall the conclusion myself, hence should probably move it to an RFC to not loose it (again).

---

Erlang also has this concept of double-colon separated types, which uses `punctuation.separator` atm.

Example:

```Erlang
-spec formaterror(Tag :: term()) -> string().
```

It is just a separator between function parameters and their type definitions. So `punctuation.separator` would be an alternative with regards to how Erlang handles that construct.

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
deathaxe: Also

```Haskell
    -- Declare a list-like instance for Char
    data instance XList Char = XCons !Char !(XList Char) | XNil
```

Haskell/Haskell.sublime-syntax
deathaxe: ... but all those changes towards `punctuation.separator` cause the syntax to look quite plain in Monokai, which might result in a need to tweak in order to not loose reputation.

Haskell/Haskell.sublime-syntax
deathaxe: I haven't followed that PR.

Haskell/Haskell.sublime-syntax
deathaxe: The old syntax does scope `=` as `keyword.operator.haskell`, too. By not applying semantic scoping in general I am not sure about making such exceptions.

Haskell/Haskell.sublime-syntax
deathaxe: When scoping `()` as constant, we should apply the same patterns to all the other constructors as well, I guess?

Token  | Descr  | Scope  |
----|----|----
`()` | unit expression | `constant.language.unit`
`(,)`, `(,,)`, `(,,,)`, `(,…,)` | tuple constructors | `constant.language.tuple`
`[]`   | list constructors | `constant.language.list`
`(->)` | function constructor | `constant.language.function`

see: https://www.haskell.org/onlinereport/haskell2010/haskellch6.html#x13-1200006.1.3

Do those really have the same meaning everywhere?

deathaxe: And how about range tuples such as `(..)`?

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: It's unfortunate if this is not feasible and while I do think it's a regression, it's definitely one that I can live with for a better overall experience. There might still be ways to target the most common type signatures and lambda definitions should definitely be possible, but I'm okay with the status quo.

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: Right, they are used for sum types …
Let's track these instances that require context-awareness in (a) separate issue(s).

Haskell/Haskell.sublime-syntax
FichteFoll: Maybe @mitranim can confirm whether `=` cannot be used anywhere else outside of an assignment?

Haskell/Haskell.sublime-syntax
FichteFoll: `punctuation` is definitely better than `keyword.operator` and would certainly be the correct choice inside record type definitions.

I don't think the Erlang example fits, however, because in Haskell the `::` are integral to the type of statement on this line (not in a record type) and denote a signature declaration. A third opinion should be helpful.

Haskell/Haskell.sublime-syntax
FichteFoll: Tuple constructors are either a type or a function, not a constant.

I don't understand what you mean with the "function constructor". `->` is a syntactic token afaict.

The *empty* list constructor I agree should also be a constant, then.

Haskell/Haskell.sublime-syntax
FichteFoll: Can be resolved.

Haskell/Haskell.sublime-syntax
mitranim: To the best of my knowledge (which is probably incomplete), Haskell uses `=` for exactly one purpose: declarations. Haskell doesn't have an "assignment" as understood in other languages, because it doesn't allow to _reassign_ names. In root scope:

```hs
funcName <params_patterns_guards> = <expressions>
data TypeName = <variants>
```

In local scope:

```hs
funcName = do
  let localFunc <params_patterns_guards> = <expressions>
  localFunc

funcName = localFunc
  where
    localFunc <params_patterns_guards> = <expressions>
```

Any local "variables" are merely nullary functions.

Haskell/Haskell.sublime-syntax
deathaxe: > Tuple constructors are either a type or a function, not a constant.

I understood the docs so that `()` is equal to `(,)`. Both denote an empty tuple. So if we decide to scope the first constant, shouldn't the others have the same scope, too then? 

Or other way round, why is `()` a constant. Maybe a type or constructor, too?

All other `(<symbol>)` constructs are scoped `meta.prefix` as I found it strange to stack `variable.function keyword.operator` to somehow denote their functional nature while maintaining operator highlighting.

> I don't understand what you mean with the "function constructor"

Unfortunatelly I don't recall where, but I copied some code from a Haskell documentation page which used `(->)` while docs said it to be a function constructor.  That's what the following test cases are copied from:

https://github.com/sublimehq/Packages/blob/f16e3fc50c812e722e38afa1adeec9df95fb0d89/Haskell/tests/syntax_test_haskell.hs#L1628-L1633

But I may have misunderstood those docs or might have been confused in other ways.

FichteFoll: > I understood the docs so that `()` is equal to `(,)`

`(,)` creates a two-tuple of its arguments and can be curried like any normal function. (Similarly, `(,,)` creates a 3-tuple.)

```haskell
Prelude> (,) 1 2 == (1, 2)
True
Prelude> with_1 = (,) 1
Prelude> with_1 2
(1,2)
```

> Or other way round, why is () a constant. Maybe a type or constructor, too?

Honestly, I don't know what a "unix expression" refers to and whether `()` is part of the typing system somehow. From gut feeling, I say it's a constant. Maybe @mitranim knows?

> That's what the following test cases are copied from:

I just get a syntax error (and not a "variable not in scope") when trying that:
```
Prelude> (->)
<interactive>:30:2: error: parse error on input ‘->’
```

I also don't recall ever seeing that, but I'm not 100% sure whether it might somehow still be legal. I'm not a Haskell expert, I just find it fun.

jwortmann: With color schemes in mind, I think this indeed needs a bit more consideration when giving another scope than `punctuation` to `()`. Usually, almost all symbol and punctuation characters are scoped either `punctuation` or `keyword.operator`. One unsatisfying, but recently improved exception to this are lambda function arrows, which currently have `keyword.declaration.function.arrow` (JS, C#, Scala) or `keyword.declaration.function.anonymous` (D & Ruby) with the latest updates.

I'm not against scoping things like `()` as `constant` and have them colored by default, but I think there should at least be a way for color schemes to target these situations if they want to use a single color for all kind of symbol and punctuation characters. While the suggestions here with a `unit`, `list` etc. sub-scope would probably allow to distinguish them from other (word-based) constants, it's still annoying for color scheme authors to search for all such exceptions and inconsistencies between different syntaxes.

Btw, with `consant.language.list` here for `[]`, wouldn't the same apply e.g. for [Python](https://github.com/sublimehq/Packages/blob/366741f52ce8aaa5255bf24fa5bb00a043fff9b7/Python/syntax_test_python.py#L1217)?

Regarding the word vs. symbol operator problem mentioned in another comment above, most of it is discussed at #2090, but unfortunately there has not really been made progress. There was a suggestion from wbond to add a `meta.operator.word` scope to situations where an operator is a word. With these additional examples here, it would be great to have a general solution for such exceptions. Perhaps add something like `meta.symbol` on top of the scope for symbols/punctuations when they are not scoped as `punctuation` or `keyword.operator`, like this example here and lambda function arrows. And then maybe only introduce only `meta.word` on top of operator keywords to have them symmetric.

FichteFoll: > but I think there should at least be a way for color schemes to target these situations if they want to use a single color for all kind of symbol and punctuation characters

Would `constant.language punctuation` suffice or be too special cased?

> Btw, with consant.language.list here for [], wouldn't the same apply e.g. for Python?

No, because `[]` is a list literal and those are mutable and not constant. However, it would be valid for the empty tuple, since that is always the same.

```
>>> t = ()
>>> t is ()
<string>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
```

deathaxe: I don't like the root of this `constants` discussion, to be honest. 

> With color schemes in mind, I think this indeed needs a bit more consideration when giving another scope than `punctuation` to `()`. Usually, almost all symbol and punctuation characters are scoped either `punctuation` or `keyword.operator`

That's exectly the reason for the original decision to use `meta.sequence punctuation` or `meta.group punctuation` for all kinds of such tokens, no matter whether they may be considered constant. That's also the reason for scoping arrows as `keyword.operator` rather than `keyword.other` which was the scope before this PR.

While it may help to see those empty `()` as constant I find it uncommon and odd compared to other languages such as python which don't (and shouldn't). I'd therefore be more than happy to revert the last commit.

> One unsatisfying, but recently improved exception to this are lambda function arrows

That's why I refuce to add another one by using `keyword.declaration` for `::` which denote type expressions. 

> And then maybe introduce only `meta.word` on top of operator keywords to have them symmetric.

We added additional meta scopes here and there, but scoping "each token" with several (espeially meta) scopes ... don't we find better solutions?

FichteFoll: I don't have strong opinions on scoping `()` or `[]` as constants in languages where they are frequently used as such, but I do recon this being quite useful for Rust, for example, where it's used as "the empty value", and it seems that Lisp people liked it as well. 

Python is arguably a different case, because it represents an empty tuple there, which is still "empty" but not the conventional "empty value", which would be `None`. To me, this differentiation makes sense and would mean that we also shouldn't scope the empty list in Haskell (`[]`) as constant, since it's an empty list and not "empty value". 
Are we confident enough to scope these kind of expressions as `constant.language.empty`? They exist in quite a few languages.

mitranim: Personally I would cast a vote against scoping `() [] {}` and such as constants, for fuzzy intuitive reasons that are hard to verbalize. But that's just one subjective opinion. Some of the reasons have already been voiced, such as color jitter when typing, or "opt in" vs "opt out" for color scheme support. Another is the issue of complexity: adding more code and special cases for no clear gain, while also having opposing voices. Erring on the side of simplicity rather than complexity is always a solid choice, in my opinion.

jwortmann: > Would constant.language punctuation suffice or be too special cased?

That would work, however it could make the `constant.language` part ineffective for highlighting in some color schemes, because only the topmost colored scope "wins". Therefore it raises the question which of it is more important and should be on top; the `constant.language` or the `punctuation` property (i.e. why not use the order `punctuation constant.language`)? We had the same issue for logical operator keywords in e.g. Python, where both `keyword.operator.logical` and `keyword.operator.word` apply, but these are orthogonal properties. One argument in favor of `constant.language` in https://github.com/sublimehq/Packages/pull/2387#issuecomment-657299850 was
> that you might accidentally leave a stray () somewhere and it would be difficult to see themes which grey-out block scoping characters

So I assume it is desired here to have it colored as `constant.language` by default, but still have a way (e.g. with the uncolored meta-scope on top) for color schemes to opt-in for the "usual" punctuation coloring. Personally, I find the punctuation property more important, but I may be biased from the color scheme I use/maintain, and I don't use Haskell or Lisp-like languages myself.

> No, because [] is a list literal and those are mutable and not constant.

Okay, that makes sense. But then I wonder again if the priority really should be whether such objects are either mutable or immutable in a specific language, rather than have unified scopes for all tuples/arrays/lists etc. over all syntaxes. Most of the time `constant.language` is used for word-like constants like `true` and `null`, so that they stand out from variable identifiers, and for me it also seems uncommon to have `()` scoped as constant.

Python uses `meta.sequence.tuple.empty.python` in combination with the usual `punctuation` scopes for empty tuples, perhaps such a solution is sufficient here as well for users or color schemes that want to apply special highlighting to `()` and `[]`?

deathaxe: Considdering all arguments of this discussion I decided to implement `meta.sequence.[list|tuple].empty`.

1. It is the most consistent solution with regards to highlighting. As the whole syntax is not designed to provide heavy/detailed semantic highlighting, it doesn't make sense to add such weird exceptions.
2. Stacking `constant` and `punctuaion` onto each other is avoided and `constant` feels out of place anyway, because `()` is used as type and constructor.

I consider this the final solution here and resolve this part of the discussion.

Haskell/Haskell.sublime-syntax
deathaxe: Hence no changes needed.

FichteFoll: I don't agree with this conclusion, but I'll open a separate PR/discussion after this is merged.

user93702699:
@deathaxe Hey this PR breaks list comprehension! Look at this example

```haskell
main :: IO ()
main = print $ f [1|True]

f :: (Show a) => [a] -> String
f [] = []
f a = a >>= show
```

user93702699:
@deathaxe Here is another issue:

```haskell
main=(#)#(#)where _#_=print$1
```

The first `_` is scoped as `variable.other`, but it should be `variable.language.anonymous`. See the screenshot:

deathaxe:
Hanks for reporting! With a look at ST 4107 in SAFE MODE using official Haskell it appears list comprehensions have been broken before.

grafik

What we see is `[1| ...` being matched as quasiquotation (https://wiki.haskell.org/Quasiquotation). Is my assumption correct, quasi quotation to require the closing punctuation `|]`? If so, I can create a fix using "branches".

user93702699:
Nice! While you are there, can you also please add support for type families?

```haskell
```

user93702699:
Haskell allows identifiers directly after a number literal. Consider the following example:

```haskell
{-# LANGUAGE FlexibleInstances #-}

main :: IO ()
main = 123f

f :: Integer -> IO ()
f = print

(⊥) = undefined

instance Num ((Integer -> a) -> a) where
  fromInteger = flip id
  (+) = (⊥)
  (*) = (⊥)
  abs = (⊥)
  signum = (⊥)
  negate = (⊥)
```

Number `123` is a literal number and it is called with argument `f`. Numbers in Haskell are not algebraic data types (like Char or String), instead they are `Num` instances, which means they can also be functions and therefore can be called. `123` should be scoped as `meta.number.integer.decimal` and `f` should be `variable.other`

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: I split this into #2868.

Haskell/Haskell.sublime-syntax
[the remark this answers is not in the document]
FichteFoll: Also mentioned this in https://github.com/sublimehq/Packages/issues/2868. You may resolve the discussion if you only kept it open for tracking this.

FichteFoll:
FYI, I branched off a couple unresolved comments I had earlier regarding context-sensitive scoping of keyword punctuation into https://github.com/sublimehq/Packages/issues/2868, so that we can focus on getting this PR merged.

rwols:
Merged by who? Who’s in charge and what is the roadmap?

FichteFoll:
Ben mentioned on Discord that he's likely to start looking into the Packages PRs but that he's also focusing on something else currently (that was 1 month ago). https://discord.com/channels/280102180189634562/280157121897562114/841480928626999306

user93702699:
GHC 9 introduced new flags such as `QualifiedDo`, `LinearTypes`, etc. They all should be `constant.language.extension`

user93702699:
`QualifiedDo` and `LinearTypes` are not the only flags (I said "etc"). There is also `LexicalNegation` and there are some older flags that should also be scoped properly:

```haskell
{-# LANGUAGE Haskell98 #-}
{-# LANGUAGE Haskell2010 #-}
{-# LANGUAGE Unsafe #-}
{-# LANGUAGE Trustworthy #-}
```

Note that flag `LinearTypes` introduces the new linear types syntax, providing the support for partial applications of linear multiplicity polymorphism in function signatures. It would be nice if you add support for that too (or at least make sure the syntax does not break in an unpleasant manner).

deathaxe:
Explicit lists of keywords, constants, ... always result in continious maintanance to keep up with development of languages. I don't plan to go through all those documents to pick them up. Once this PR is merged it should be quite easy for everyone interested to extend those lists easily.

mitranim:
Just a note: any hardcoded whitelist can be a variable, which allows any end-user to inherit from the syntax and supply their own variable. Of course, it's _much_ preferable to avoid any hardcoded lists.

deathaxe:
That's true in general. Variables are the way to go with it. They are needed to distinguish builtins from user defined tokens though. (e.g. `constant.language` vs. `constant.other`). Already thought about scoping "unknown" directives and values this way, so most users might not see a difference.

On the other hand such lists provide some low-level way to quickly see whether a valid builtin token is entered.

jpe90:
Are escaped quotes in String literals highlighted correctly?

Here is a screenshot from this PR: 
subl_highlighting
And the same code with treesitter in neovim:
nvim_ts_highlighting

rwols:
That looks correct to me.

deathaxe:
Resolved a conflict with #2760.

FichteFoll (approved):
We've been through this quite a while ago already. For verification, I just tested this out on a couple Haskell files I wrote by myself a few years ago and everything looked good or at least not worse than before, so this gets a pass from me.

deathaxe:
Most significant changes in highlighting should be caused by moving from `keyword.` scopes to `punctuation.` for most brackets.

FichteFoll:
That, plus some function scoping and imports where the most obvious to me.

jrappen (approved):

pennychase:
@jrappen asked me to look at the PR. I don't consider myself a Haskell expert, but I have been programming in the language for a couple of years. Overall, I really like the syntax, but there are some things that could be tweaked. In the module declaration all of the elements of the qualified module name should be highlighted the same way.In this capture the first element is white and the last one is yellow, but they should probably all be yellow:
Screen Shot 2022-01-31 at 10 34 25 PM

And in a qualified import, the elements of the qualified name should also all be highlighted in the same color:
Screen Shot 2022-01-31 at 10 45 09 PM

In type declarations, it looks like some types are considered "keywords" and are blue, while others are purple.Thwy are all really just type constructors and should be highlighted the same way.
Screen Shot 2022-01-31 at 10 52 40 PM

In this code, we see that Just and Nothing, the value constructors of the Maybe type are treated differently; Nothing is treated as a value constructor (highlighted purple), but Just seems to be treated as a keyword (highlighted blue). And the value constructors of the Either type, both seem to be traced as keywords (highlighted blue).
Screen Shot 2022-01-31 at 10 55 32 PM

I would expect  all Type constructors to be highlighted the same way and all value constructors tp be highlighted the same way. It doesn't make sense to treat some as reserved words (you can always not import Prelude and redefine these).

I'm looking at other examples of code and am trying to sort out some things like type class definitions and using type classes in serving clauses, and will post other comments as I finish reviewing. As I said, overall I like the highlighting distinctions (e.g., types are clearly differentiated from constructors in newtype and data declarations; it's neat that turning a function to inbox with backticks is highlighted the same way as an infix operator). Thanks for doing this.

michaelblyons (approved):

deathaxe:
> And in a qualified import, the elements of the qualified name should also all be highlighted in the same color

It's up to the color scheme to choose a color for the given types.

The syntax definition scopes qualifiers as `variable.namespace`. Only the final token is the real variable or entity being imported. That's the scope naming scheme choosen in all syntaxes.

> In type declarations, it looks like some types are considered "keywords" and are blue, while others are purple.

Basic types like `String`, `Int64` are scoped as builtin data type (`support.type`), while all user defined types are scoped `storage.type`.

> ... we see that Just and Nothing ...

`Nothing` as well as `True` and `False` are scoped `support.constant`, while `Just`, `Left` and `Right` are scoped `support.type`. Should we scope all of them `support.type` to denote them being type constructors?

pennychase:
Responded to Nothing, Just, etc in Issue #3227. I think thay all need to be scoped as constructors (they're actually the value constructors for the type). There is the question of differentiating them as built-ins. GitHub appears to treat Left/Right as built-ins and Just/Nothing as ordinary constructors. One could go either way.

moodmosaic:
Great work, all! Was this [released with BUILD 4142](https://www.sublimetext.com/download)?

Haskell/tests/syntax_test_haskell.hs
```diff
@@ -0,0 +1,4209 @@
...
+-- ^ - meta.function
+--  ^^^^^^^^^^^^^^^^^^^ meta.function.identifier.haskell
+--  ^^^^^ entity.name.function.haskell
+--       ^^ punctuation.separator.sequence.haskell
+--          ^^^^^^^^^^ entity.name.function.haskell
+        :: a -> Bool
+--     ^ meta.function.identifier.haskell
+--      ^^^^^^^^^^^^^ - meta.function
+--      ^^ punctuation.separator.type.haskell
+--         ^ variable.other..haskell
```
stefanobaghino, line 2060:
```suggestion
--         ^ variable.other.haskell
```
Question: is the double dot a typo or does it have some specific meaning that I'm missing?

michaelblyons: Looks like a typo to me. Feel free to PR.

stefanobaghino: Confirmed typo — opened #4532 to fix it.

PR #2680, opened 2020-12-30
[Haskell] Add completions

deathaxe:
This PR proposes to add completions for most basic keywords, prelude functions and types as well as for some well known pragma keywords.

It requires #2679.

_Note: I am not sure whether we want to add completions in general, hence the separate PR._


Ultra-Instinct-05:
Would it make sense to add these once there was a way to disable completions coming from `.sublime-completions` file ? Snippets are fine since there is a way to disable them but currently there is no way to disable completions coming from these files via a setting.

deathaxe:
That's why the separate PR, yes. While it only completes keywords and known functions in a word completion like fashion, it may probably cause duplicate entries when language servers or other plugin based completion engines are used.

jfcherng (approved):

jrappen (approved):

FichteFoll:
> it may probably cause duplicate entries when language servers or other plugin based completion engines are used.

That's what I'm fearing as well, although an LSP/plugin can disable completions from completions files iirc. If that is the case, then we are probably fine with offering completions for built-ins since anything else providing them can be disabled

Adding keywords should generally be fine, but the longer the list becomes the more likely it is to cause conflicts somewhere (e.g. the entire prelude completions). It would be nice if we could get someone to try this PR and report their experience. I also have no idea if an LSP implementation or similar exists for Haskell and is available to ST, but having someone test that together with this PR would be nice, too. [SublimeHaskell](https://packagecontrol.io/packages/SublimeHaskell) appears to provide dynamic completions according to the readme.

Aside from that, the contents of the PR look fine to me. No approval from me though until the above is solved.

deathaxe:
A language server is available at https://github.com/haskell/haskell-language-server

jrappen (approved):

FichteFoll (approved):

michaelblyons (approved):

PR #2682, opened 2021-01-02
[Haskell] Add Cabal Syntax

deathaxe:
This PR proposes to add basic support for _Cabal Configuration Files_.

It appears Cabal to be the default package manager used by Haskell. It might therefore make sense to have basic support for Cabal
configuration files available next to Haskell syntax itself.

It is known not to be intended to add further rarely used syntax definitions to the repo in general. We might also add it as dedicated package to Package Control then.


Haskell/Cabal.sublime-syntax
```diff
@@ -0,0 +1,406 @@
...
+    - include: other-keys
+    # expressions
+    - include: function-calls
+    - include: keywords
+    - include: operators
+
+###[ COMMENTS ]################################################################
+
+  comments:
+    - match: (--+).*\n?
```
FichteFoll, line 87: Maybe turn the `.*` match into a context body?

Haskell/Cabal.sublime-syntax
```diff
@@ -0,0 +1,406 @@
...
+        1: meta.toc-list.sections.cabal
+        2: keyword.declaration.section.cabal
+        3: entity.name.section.cabal
+
+###[ BOOLEANS ]################################################################
+
+  boolean-keys:
+    - match: ^(\s*)({{boolean_tags}})(:)
+      captures:
+        2: meta.mapping.key.cabal entity.name.tag.cabal
```
FichteFoll, line 106: I've been out of RFC discussions for a while, but did we reach an agreement on how to scope keys?

Haskell/Cabal.sublime-syntax
```diff
@@ -0,0 +1,406 @@
...
+
+  version-keys:
+    - match: ^(\s*)({{version_tags}})(:)
+      captures:
+        2: meta.mapping.key.cabal entity.name.tag.cabal
+        3: meta.mapping.cabal punctuation.separator.mapping.key-value.cabal
+      push: version-values
+
+  version-values:
+    - meta_scope: meta.versions.cabal
```
FichteFoll, line 228: Would it make sense to use `meta.section.*.cabal` for these? I don't remember a particular case where this was applied in the past, but it makes sense in my mind.

Haskell/Cabal.sublime-syntax
```diff
@@ -0,0 +1,406 @@
...
+      scope: keyword.operator.comparison.cabal
+
+  sequence-separators:
+    - match: ','
+      scope: punctuation.separator.sequence.cabal
+
+###[ LITERALS ]################################################################
+
+  literal-emails:
+    - match: <?(?=\w\S*?@\S+?\.\S+?)
```
FichteFoll, line 313: Is `+?` compliant with sregex?

Would branching make sense here?

FichteFoll (commented):
I just took a brief look as I had some time to kill. I've never used cabal before, and I can't comment on its usage in the Haskell ecosystem, but generally this looks fine to me.

Haskell/Cabal.sublime-syntax
deathaxe: Why?

Whether we use this simple pattern or push into a context does no longer have any performance differences. Where's the use then?

Haskell/Cabal.sublime-syntax
deathaxe: No, I am actually not very confident of how to scope these things in such config files. But as CSS and HTML make use of `entity.name` or `entity.other` heavily it was the first thing which came to my mind here.

It also matched a recent description Will posted about tags, which was something like "A tag is a token which denotes a start of a block of text" or something along these lines.

I am happy with any better alternative.

Haskell/Cabal.sublime-syntax
deathaxe: A common 2nd level scope would probably make sense, yes.

1. I am just not sure how good `meta.section` vs. `meta.mapping` fit together to denote the  same block of code/text .
2. With regards to the example below `entity.name.section` is used for `name` in `executable name` lines. IMHO, the whole block below that line is a section of mappings. So it feels a bit strange to denote the key-value pairs as `section` as well, maybe.

**Example**

```
executable name
   compiler: ghc 8.3
   version: 3.3
   
executable othername
  compiler: ghc 8.4
  version 2.2
```

Haskell/Cabal.sublime-syntax
deathaxe: `+?` is compliant, yes. The contexts are borrowed from `Git Formats Common.sublime-syntax` and slightly modified to also highlight e-mail addresses without `<` `>`.

Yes, branching wouldn't work for plain e-mail addresses without angles.

Haskell/Cabal.sublime-syntax
FichteFoll: Not consuming the entire line in a match allows for more flexibility with `with_prototype` (and inheritance nowadays). Optional in my eyes, but wanted to mention it regardless.

Haskell/Cabal.sublime-syntax
FichteFoll: Good point. If we're already in the values part of a mapping, then maybe we want to use `meta.mapping.value.*.cabal`?

Haskell/Cabal.sublime-syntax
deathaxe: That's very correct in general. I actually don't expect someone to want to inject special highlighting into comments though.

The `with_prototype` mechanism wouldn't be very useful for that anyway. An inherited syntax could easily override the whole `comments` context in case more sophisticated highlighting is desired. Otherwise `embed` is the best way incase someone want's Cabal syntax highlighting supported in e.g. Markdown fenced codeblocks.

I'd push into a context in case comments may contain special tags such as type hints in python or docstrings with dedicated tags to highlight, but that's not the case here.

FichteFoll (approved):

Haskell/tests/syntax_test_cabal.cabal
```diff
@@ -0,0 +1,329 @@
...
+--     ^ meta.mapping.cabal
+--      ^^^^^^^^^^^^^^^^^^ meta.mapping.value.version.cabal
+-- ^^^^ entity.name.tag.cabal
+--     ^ punctuation.separator.mapping.key-value.cabal
+--               ^^^^^^^^ meta.number.version.cabal constant.numeric.value.cabal
+--                ^ punctuation.separator.decimal.cabal
+--                   ^ punctuation.separator.decimal.cabal
+--                     ^ punctuation.separator.decimal.cabal
+
+author: Jon Doe <jon.doe@nowhere.com>
```
jrappen, line 34: Freudian slip? This guy's name is John. Jon is the other guy.

deathaxe: Could also be Foo Bar. It's a test case. Content doesn't matter.

jrappen: it was ... a joke, nvm

jrappen:
Over all this seems fine.

Was there a reason, `meta.toc-list.*` wasn't added to the symbol list?

deathaxe:
Sections are displayed in Symbol List via `entity.name.section`.

Symbol tests have not been prominent and important enough.

jwortmann:
Not specific to only this PR, but maybe we can unify the scope names used for punctuations within mappings?


What's currently used in various syntaxes (on master):

| Syntax | key-value separator, e.g. `:` or `=` | entry separator, e.g. `,` or `;` | uses `meta.mapping` |
| ------ | ----------------------------- | ------------------------- | ----------------- |
| JSON | [`punctuation.separator.mapping.key-value`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/JSON/JSON.sublime-syntax#L118) | [`punctuation.separator.mapping.pair`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/JSON/JSON.sublime-syntax#L131) | :heavy_check_mark: |
| YAML | [`punctuation.separator.key-value.mapping`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/YAML/YAML.sublime-syntax#L663) | - | :x: |
| CSS | [`punctuation.separator.key-value`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/CSS/CSS.sublime-syntax#L1510) | [`punctuation.terminator.rule`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/CSS/CSS.sublime-syntax#L3079) | :x: |
| Python (dict) | [`punctuation.separator.mapping.key-value`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/Python/Python.sublime-syntax#L1497) | [`punctuation.separator.mapping`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/Python/Python.sublime-syntax#L1528) | :heavy_check_mark: |
| JavaScript (object literal) | [`punctuation.separator.key-value`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/JavaScript/JavaScript.sublime-syntax#L1691) | [`punctuation.separator.comma`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/JavaScript/JavaScript.sublime-syntax#L244) | :heavy_check_mark: |
| Lua (table)* | [`punctuation.separator.key-value`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/Lua/Lua.sublime-syntax#L666) | [`punctuation.separator.field`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/Lua/Lua.sublime-syntax#L633) | :heavy_check_mark: |
| BibTeX | [`punctuation.separator.key-value`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/LaTeX/Bibtex.sublime-syntax#L71) | [`punctuation.separator.mapping.pair`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/LaTeX/Bibtex.sublime-syntax#L80) | :heavy_check_mark: |
| Git config | [`keyword.operator.assignment`](https://github.com/sublimehq/Packages/blob/6b0eae541177e9ee8655356ae85694a4e901762e/Git%20Formats/Git%20Config.sublime-syntax#L128) | - | :heavy_check_mark: |
| Cabal (this PR) | `punctuation.separator.mapping.key-value` | - | :heavy_check_mark: |

*not sure whether the Lua tables are directly comparable to the other mappings listed here

Any more that I have forgotten?
The `mapping` part in the scope used by a few synaxes seems to be redundant to me, because it is already given by `meta.mapping`, and color schemes could hence use `meta.mapping punctuation.separator` to target separators only within mappings. So how about to establish the conventions `punctuation.separator.key-value` and maybe `punctuation.separator.pair`?

jrappen:
* There are a few other separators, so keeping `*.mapping.*` makes sense. Compare:
  https://github.com/SublimeText/PackageDev/blob/24d53896ff1532b057df73f41ad9e6ca8578c9f3/plugins/lib/scope_data/data.py#L190-L255
* I guess deathaxe just forgot it in his CSS rewrite. The YAML rewrite was quite a while back. I agree both should be fixed.
* I don't understand why we should skip a level for `mapping`, but keep it for the others (listed two points above here).

jwortmann:
My intention is mostly to have unified scope names between syntaxes, so as long as it is used consistenty, I could live with the `.mapping.` part in the scope.

It's just that it is redundant due to the `meta.mapping`, and from the table above only two syntaxes (JSON & Python) use it, while `key-value` as the third level is used in 5 syntaxes (YAML, CSS, JavaScript, Lua, BibTeX). Actually I forgot HTML which also uses `punctuation.separator.key-value` for the `=` in tag attributes, so it's 6 syntaxes already. The scope ordering from the provided link doesn't seem fully comprehensive about what is used in this repository for `punctuation.separator` scopes, for example it misses `punctuation.separator.comma` which is also extensively used. I think for the other separator scopes there is no comparable meta scope, i.e. they don't have this redundancy. What exactly do you mean with "skip a level"?

For CSS, I can understand that it wasn't changed to `meta.mapping`, because the CSS scopes are somewhat unique and a lot of color schemes depend on them. This can't be carelessly converted without ensuring that backward compatibility is provided.

deathaxe:
@jwortmann Please open a new issue for that discussion. That's really off topic here!

deathaxe:
This is the up to date scope, which was introduced together with `meta.mapping` scope scheme. It's suggested by PackageDev's auto-completion. All other's are older.

Scopes of CSS have been discussed and CSS was not judged a data mapping. Thus scopes won't change. Especially as those are targeted by many older color schemes.

jwortmann:
Oh that is good to know, then I guess we want to update the older scopes to use this convention. That would make it easier for color schemes. I was just confused because many syntaxes use `meta.mapping` but then the punctuation scope apparently wasn't updated if that is the latest convention?

jwortmann:
Regarding the punctuation scope, I tried to search whether it was discussed before, but couldn't find much about it in this repository.

But what I found was this commit https://github.com/sublimehq/Packages/commit/60e7c391deb32d1d39b5add199f827c7d3bb8050, where `punctuation.separator.mapping` was explicitly changed to `punctuation.separator.key-value` in YAML. And in https://github.com/sublimehq/Packages/pull/600#issuecomment-248554606 it is even advised to change it in JSON as well.

So are you sure that the scope with `mapping` at the third level is really the latest convention?

To be clear, if that is really the case, then I don't want to oppose using it. My intention is just to unify the scopes between syntaxes to use a common convention, and from what I could find both in the links above as well as the current usage in syntaxes, the one without `mapping` seems to be the preference.

deathaxe:
I won't commit on it any further on this PR as it is off topic!!!!!

Haskell/Cabal.sublime-syntax
jrappen: Just adding a reference for interested readers who want to read more about what FichteFoll and deathaxe are talking about:

#757 with Keith's comment there about line comments.

jrappen (dismissed):

BenjaminSchaaf (approved):

BenjaminSchaaf:
From a shallow review this looks good and this is something we want in the packages, thanks!

deathaxe:
`punctuation.separator.mapping.key-value` and `punctuation.separator.mapping.pair` may be something to clarify at some point. Thay have been introduced as being suggested by PackageDev, but @jwortmann's arguments against them are too compelling to not think about.

jrappen:
I added my comments to 3228 with regards to the latest changes here.

jrappen (approved):

jrappen:
@michaelblyons @pennychase @moodmosaic @mitranim As this is a follow up to the Haskell re-write, maybe you could review this one too by clicking [the `Files changed` tab above](https://github.com/sublimehq/Packages/pull/2682/files) & starting a review by clicking the green button at the top right?

moodmosaic:
Thanks for reaching out, @jrappen. I would love to help but don't have the bandwidth currently.

FWIW, I ended up rolling my own custom syntax. I glued parts from the original TextMate grammar, [some of my PRs here](https://github.com/sublimehq/Packages/pulls?q=is%3Apr+author%3Amoodmosaic), and some parts from the Atom's [language-haskell](https://github.com/atom-haskell/language-haskell) grammar. This works well for me, for the past 2 years.

deathaxe:
> "LICENSE" and "vidar@vidarholen.net" are highlighted as unquoted strings, the rest as text.

Those denote paths to local or remote entities, which should be distinguished from normal keys by intent - especially as punctuation is scoped. Used related scopes from HTML `<a href="">` for that (`meta.path meta.string string`).

Could also follow Markdown and use `meta.link markup.underline.link`.

... or `meta.path markup.underline.link`.

_Note: Looks like urls/links deserve a RFC as they are scoped quite differently across syntaxes._

deathaxe:
> Here hs-source-dirs, build-depends, exposed-modules, other-modules and default-language are marked as definitions.

They are tagged `meta.mapping.key entity.name.tag` as any other top-level key-value pair. What's wrong about it?

deathaxe:
I found a possibility to quickly navigate to those keys useful. If not, we can remove _Cabal Symbol List_

deathaxe:
There's no difference between top- and none-top-level keys, logically. Former ones just describe globals, while latter ones single assets or build targets. The meaning however is the same within their scope.

Syntax currently doesn't distinguish them and I am uncertain I want to add this level of complexity. So it's all or nothing.

deathaxe:
> source-repository-package is highlighted as regular text.

Only known tokens are supported as section headers (see `section_tags` variable) to avoid false positives. We could add this special keyword. If we are in doupt about others being unknown a more general pattern was to be used.

deathaxe:
All review comments are addressed.

deathaxe:
v3.4 was the up-to-date spec version at the time of writing cabal syntax. It doesn't mean it shouldn't support newer features.

moodmosaic:
Great work! @deathaxe, was this [released with BUILD 4142](https://www.sublimetext.com/download)?

FichteFoll (approved):
Since perfect is the arch-enemy of good, I'll approve of this PR in its current state. If any of these are things you want to work on, @deathaxe, be my guest, but I wouldn't block a merge on them.

1. There are still debates to be had regarding the key scope, but since we don't have a clear direction on this yet, this particular PR is not the place to discuss and decide on it.

1. Similarly, some key-value pairs definititely look like the value should be an unquoted string (much like in YAML) where they currently only have `meta.mapping.value`. Again, not something I'm hung up upon.

1. The other thing I noticed was stacking of `meta.function-call`, but also not a requirement for me for such a rarely used syntax.

deathaxe:
to 1.) I'd be happy with any suitable scope of those keys as we don't even have categorized this kind of files. We could treate them as mappings / structured key-value pairs and use json/yaml scopes. It wouldn't be the worst choise, but there's no real need to distinguish data types of keys like in YAML. Scoping it string therefore feels a little odd as everthing in those files is a string and we'd have the same kind of discussions about "everything is string scoped" as we had/have for JSON/YAML. 

As Cabal is basically Haskell's Makefile the question arises, how to scope `.PHONY:` and friends in Makefile later on. They now are scoped `function`.

... but well, that's probably a separate topic.

to 2.) Same here. We could scope them `string`. What I had in mind is HTML and Markdown, which both don't scope plain text special as well.

to 3.) Do you have any examples? Or is it `meta.function-call.arguments meta.group` which you are concerned about?

FichteFoll:
Re 3.: No, I don't. Not sure what I was seeing there exactly but I certainly don't see it anymore.

deathaxe:
Suggestions for better scopes are always wellcome.

PR #2683, opened 2021-01-02
the “#“ alone initiates an end-of-line comment, but the “!“ does not

emsspree:
Hallo And Happy New Year!

* http://alvinalexander.com/blog/post/mac-os-x/applescript-use-comments/
* https://developer.apple.com/library/archive/documentation/AppleScript/Conceptual/AppleScriptLangGuide/conceptual/ASLR_lexical_conventions.html#//apple_ref/doc/uid/TP40000983-CH214-SW8

Not the sequence ”#!“ but the single character ”#“ initiates an end-of-line comment.
”#“ is treated/handled as/like ”--“.

Ciao


wbond:
Thank you for this fix!

PR #2684, opened 2021-01-02
the “#“ alone initiates an end-of-line comment, but the “!“ does not. (Comments.tmP…)

emsspree:
hej, see
https://github.com/sublimehq/Packages/pull/2683


wbond:
Thanks for fixing this part also!

In case you weren't aware, in the future if you have multiple fixes that are to the same syntax, you can push new commits to the branch that you use when you create a PR.

emsspree:
good to know, thank you

PR #2685, opened 2021-01-06
[PHP] Named arguments syntax not working properly

Juhlinus:
Currently the syntax for name arguments do not work when using more than one argument.

Screen Shot 2021-01-06 at 10 18 34 PM

Changing `pop 1` to `pop true` seemed to have done the trick.

Screen Shot 2021-01-06 at 10 19 41 PM

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -1190,7 +1190,7 @@ contexts:
       push:
         - include: expressions
         - match: ''
-          pop: 1
+          pop: true
   heredoc:
     - match: (?=<<<\s*'?({{identifier}})'?\s*$)
       push:


jfcherng:
I can't reproduce.

image

And `pop: true` and `pop: 1` are exactly doing the same thing in ST.

rwols:
The disclaimer in the readme should be updated to state something like "this is developed against ST4" to prevent this sort of thing.

jrappen:
> @rwols: The disclaimer in the readme should be updated to state something like "this is developed against ST4" to prevent this sort of thing.

#2686 should prevent this

---

Cannot reproduce for the same reasons @jfcherng mentioned.

```yaml
Sublime_Text:
  build: 4094
  channel: dev
  platform: osx x64
    portable: false

Sublime_Merge:
  build: 2041
  channel: dev
  platform: osx x64
    portable: false

System:
  hardware: MBP 15" 2018, i9 2.9GHz, 32GB, 1TB, Radeon Pro 560X
  software: macOS Big Sur 11.1 (20C69)
    kernel_version: Darwin 20.2.0
    open_gl:
      gl_api_version: 4.1 INTEL-16.1.7
      glsl_version: 4.10
      vendor: Intel Inc.
      renderer: Intel(R) UHD Graphics 630
```

wbond:
It appears the issue here is just using the syntax with an older build of Sublime Text, which doesn't support all of the new features.

PR #2686, opened 2021-01-07
update README, only run syntax tests when needed

jrappen:


jrappen:
This PR is intended to prevent future PRs that target ST3 or older.

michaelblyons:
One comment. In _this_ ReadMe, you might make a mention of the `st3` branch so people on the much-more-widely-used version have a pointer to instructions and files that will work for them.

jrappen:
The syntax test should not be running on changes to the README.

README.md
```diff
@@ -1,25 +1,38 @@
+[![SublimeHQ Discord](https://img.shields.io/discord/280102180189634562?label=SublimeHQ%20Discord&logo=discord&style=flat-square)](https://discord.gg/D43Pecu)
+
 # Sublime Packages
 
-> **These packages are developed against the latest [Sublime Text 3 Dev Build](http://sublimetext.com/3dev). Bugs may exist on older builds, and the format used is not compatible with builds older than 3092.**
+> **These packages are developed against the latest [Sublime Text Dev Build](http://sublimetext.com/).**
+> **Bugs may exist on older builds, and the format used is not compatible with builds older than 4075.**
```
deathaxe, line 6: FWIW, some syntaxes (Java Rewrite) may even fail on builds before 4094 or even cause older versions to crash or freeze.

michaelblyons (changes requested):
Revert this commit. The `tmPreferences` files are testable, too.

_Edit:_ Or add `tmPreferences` to the allowlist. (I don't know if there are others to be added, too.)

README.md
jrappen: @deathaxe The changelog doesn't mention any syntax related changes for ST4094. Are you sure about this?

jrappen: Since #2655 bumped the test binary from 4090 to 4094 and fixed tests failing in #2654, it must be one of those builds.

keith-hall (approved):

rwols (approved):

michaelblyons (approved):

FichteFoll:
Changes look good, but I don't know if this is the right time to change it in that way, especially considering symlinking instruction, which are a bit more nuanced wrt whether a "Sublime Text 3" folder exists.

It should definitely mention the ST3-compatible branch, however.

michaelblyons:
I imagined you link to the branch and say to use THAT readme instead.

jrappen:
@FichteFoll I could just remove the whole section and link to:

<https://www.sublimetext.com/docs/packages.html>

if Will would add the docs there. Open for suggestions to make changes for the meantime here, though.

jrappen:
@michaelblyons FichteFoll probably meant using ST4 with an existing ST3 data folder.

michaelblyons:
> @michaelblyons FichteFoll probably meant using ST4 with an existing ST3 data folder.

Ah. That makes a lot of sense, thanks.

README.md
```diff
@@ -1,25 +1,38 @@
...
+and change the paths below accordingly as described in the README there:
+
+<https://github.com/sublimehq/Packages/tree/st3#readme>
+
 ### OS X
 
 ```bash
 $ git clone https://github.com/sublimehq/Packages.git
-$ ln -s `pwd`/Packages/Python ~/Library/Application\ Support/Sublime\ Text\ 3/Packages/
+$ ln -s `pwd`/Packages/Python ~/Library/Application\ Support/Sublime\ Text/Packages/
```
wbond, line 28: The directions for all three platforms here will be wrong if they upgraded from an ST3 install, since it will keep using their ST3 data dir. It really will need them to look for their data dir. Probably the easiest way is to have them use *Preferences > Browse Packages…*.

README.md
```diff
@@ -29,10 +42,10 @@ On Windows, you can use directory junctions instead of symlinks (symlinks requir
 ```powershell
 # Using PowerShell
 PS> git clone https://github.com/sublimehq/Packages.git
-PS> cmd /c mklink /J "$env:APPDATA/Sublime Text 3/Packages/Python" (convert-path ./Packages/Python)
+PS> cmd /c mklink /J "$env:APPDATA/Sublime Text/Packages/Python" (convert-path ./Packages/Python)
 ```
 
-Alternatively, download the portable version, and clone this repository directly as a subdirectory of the *Data* folder.
+Alternatively, download the portable version, and clone this repository directly as a subdirectory of the *`Data`* folder.
```
wbond, line 48: If we tell them to use *Preferences > Browse Packages…* then we can remove this paragraph since it will work for portable also.

README.md
```diff
@@ -1,25 +1,38 @@
+[![SublimeHQ Discord](https://img.shields.io/discord/280102180189634562?label=SublimeHQ%20Discord&logo=discord&style=flat-square)](https://discord.gg/D43Pecu)
```
wbond, line 1: I'd rather link to the forum Resources post than directly to our Discord server.

README.md
```diff
@@ -1,25 +1,38 @@
...
+
 # Sublime Packages
 
-> **These packages are developed against the latest [Sublime Text 3 Dev Build](http://sublimetext.com/3dev). Bugs may exist on older builds, and the format used is not compatible with builds older than 3092.**
+> **These packages are developed against the latest [Sublime Text Dev Build](http://sublimetext.com/).**
```
wbond, line 5: This should say:

```suggestion
> **These packages are developed against the latest dev build of Sublime Text, currently build 4094. These syntaxes may be incompatible, sometimes in very subtle ways, with older builds.**
```

We'll leave the link out for now since we don't want to publicly link to downloads.

wbond (changes requested):
I've outlined some changes I'd like to see in order to merge this in.

README.md
```diff
@@ -1,25 +1,35 @@
+[![SublimeHQ Discord](https://img.shields.io/discord/280102180189634562?label=SublimeHQ%20Discord&logo=discord&style=flat-square)](https://forum.sublimetext.com/t/resources-and-bug-tracking/22985)
+
 # Sublime Packages
 
-> **These packages are developed against the latest [Sublime Text 3 Dev Build](http://sublimetext.com/3dev). Bugs may exist on older builds, and the format used is not compatible with builds older than 3092.**
+> **These packages are developed against the latest dev build of Sublime Text, currently build 4094. These syntaxes may be incompatible, sometimes in very subtle ways, with older builds.**
```
michaelblyons, line 5: This is now out of date. Do you want to mention a specific version or say "4xxx series" ?

jrappen: I simply made all changes Will suggested. I assume the `Build 4094` requirement still stands, though.

michaelblyons: Fair enough. I was going off the line in the 4095 changelog that says

> - Resolve some bugs with syntax definitions that use branching

jrappen:
closing as stale

PR #2690, opened 2021-01-13
[Markdown] Add keybindings for `, _ and *

### LINKED ISSUE #2689: [Markdown] Typing * or ` does not surround selection with character
When editing Markdown files, I expect to be able to make a selection and then type `*` or backpack to surround the selection with the character I pressed in the same what that typing `"` works in code files.

- Sublime Version: 4094
- OS Version: macOS Big Sur 11.1 (20C69)

## Expected behavior
Selected text is surrounded by asterisk or backtick

## Actual behavior
Selected text is replaced by asterisk or backtick

## Steps to reproduce

1. Open a Markdown file
2. Select some text
3. Press `*`
4. Note that selected text is replaced with an `*` rather than surrounded by it

akrabat:
These keybindings will create a pair of asterisks, underscores or back ticks when you type a single one, or will surround the selection with asterisks, underscores or back ticks as appropriate.

This PR is solely concerned with making Markdown's surrounding asterisk, underscore and back tick characters behave similar to how surrounding quotation mark and brace characters work in other language files.

Fixes #2689

### DIFF
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -19,7 +19,7 @@
             { "key": "selection_empty", "operand": false, "match_all": true },
         ]
     },
-    // If we have two asterisks with the cursor betweem them, then a backspace removes both
+    // If we have two asterisks with the cursor between them, then a backspace removes both
     { "keys": ["backspace"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Delete Left Right.sublime-macro"}, "context":
         [
             { "key": "setting.auto_match_enabled" },
@@ -49,7 +49,7 @@
             { "key": "selection_empty", "operand": false, "match_all": true },
         ]
     },
-    // If we have two underscores with the cursor betweem them, then a backspace removes both
+    // If we have two underscores with the cursor between them, then a backspace removes both
     { "keys": ["backspace"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Delete Left Right.sublime-macro"}, "context":
         [
             { "key": "setting.auto_match_enabled" },
@@ -80,7 +80,7 @@
             { "key": "selection_empty", "operand": false, "match_all": true },
         ]
     },
-    // If we have two backticks with the cursor betweem them, then a backspace removes both
+    // If we have two backticks with the cursor between them, then a backspace removes both
     { "keys": ["backspace"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Delete Left Right.sublime-macro"}, "context":
         [
             { "key": "setting.auto_match_enabled" },
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -18,6 +18,16 @@
             { "key": "selection_empty", "operand": false, "match_all": true },
         ]
     },
+    // If we have two asterisks with the cursor betweem them, then a backspace removes both
+    { "keys": ["backspace"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Delete Left Right.sublime-macro"}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown - markup.raw" },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\*$", "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^\\*", "match_all": true }
+        ]
+    },
 
     // Auto-pair backticks
     { "keys": ["`"], "command": "insert_snippet", "args": {"contents": "`$0`"}, "context":
@@ -37,5 +47,15 @@
             { "key": "setting.auto_match_enabled"},
             { "key": "selection_empty", "operand": false, "match_all": true },
         ]
-    }
+    },
+    // If we have two backticks with the cursor betweem them, then a backspace removes both
+    { "keys": ["backspace"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Delete Left Right.sublime-macro"}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operator": "equal", "operand": "punctuation.definition.raw.begin.markdown" },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "`$", "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^`", "match_all": true }
+        ]
+    },
 ]
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -5,8 +5,7 @@
             { "key": "setting.auto_match_enabled"},
             { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "selection_empty", "match_all": true },
-            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\w", "match_all": true },
-            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\S" },
         ]
     },
     {
@@ -26,8 +25,7 @@
             { "key": "setting.auto_match_enabled"},
             { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "selection_empty", "match_all": true },
-            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\w", "match_all": true },
-            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\S" },
         ]
     },
     {
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -30,6 +30,36 @@
         ]
     },
 
+    // Auto-pair underscore
+    { "keys": ["_"], "command": "insert_snippet", "args": {"contents": "_$0_"}, "context":
+        [
+            { "key": "setting.auto_match_enabled"},
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "preceding_text", "operator": "not_regex_contains", "operand": "\\\\$", "match_all": true },
+        ]
+    },
+    {
+        "keys": ["_"],
+        "command": "insert_snippet",
+        "args": {"contents": "_${0:$SELECTION}_"},
+        "context": [
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
+            { "key": "setting.auto_match_enabled"},
+            { "key": "selection_empty", "operand": false, "match_all": true },
+        ]
+    },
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
...
+
+    // Auto-pair backticks
+    { "keys": ["`"], "command": "insert_snippet", "args": {"contents": "`$0`"}, "context":
+        [
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown" },
+            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
+        ]
+    },
```
rwols, line 30: In the case of `"`, you can press <kbd>backspace</kbd> and the extra inserted `"` (totalling `"|"`, where `|` is the cursor) also is removed. This behavior seems to be missing for \`.

deathaxe: I'd rather expect to toggle `bold` and `italic` style via the same key binding.

That said, Markdown may need several such input helpers which I'd like to know whether they are wanted to be part of this core package first.

Otherwise an up-to-date "SmartMarkdown" package may do the job as well. MarkdownEditing has much to offer, too, but may need to be stipped down a bit to remove no longer needed stuff.

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
+[
+    // Auto-pair asterisks
+    { "keys": ["*"], "command": "insert_snippet", "args": {"contents": "*$0*"}, "context":
+        [
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown" },
+            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
```
AmjadHD, line 7:
```suggestion
            { "key": "setting.auto_match_enabled" },
            { "key": "selector", "operand": "text.html.markdown" },
            { "key": "selection_empty", "match_all": true },
```

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
...
+        ]
+    },
+    {
+        "keys": ["*"],
+        "command": "insert_snippet",
+        "args": {"contents": "*${0:$SELECTION}*"},
+        "context": [
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown" },
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selection_empty", "operator": "equal", "operand": false, "match_all": true },
```
AmjadHD, line 18:
```suggestion
            { "key": "selector", "operand": "text.html.markdown" },
            { "key": "setting.auto_match_enabled" },
            { "key": "selection_empty", "operand": false, "match_all": true },
```

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
...
+        ]
+    },
+
+    // Auto-pair backticks
+    { "keys": ["`"], "command": "insert_snippet", "args": {"contents": "`$0`"}, "context":
+        [
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown" },
+            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
```
AmjadHD, line 27:
```suggestion
            { "key": "setting.auto_match_enabled" },
            { "key": "selector", "operand": "text.html.markdown" },
            { "key": "selection_empty", "match_all": true },
```

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
...
+        ]
+    },
+    {
+        "keys": ["`"],
+        "command": "insert_snippet",
+        "args": {"contents": "`${0:$SELECTION}`"},
+        "context": [
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown" },
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selection_empty", "operator": "equal", "operand": false, "match_all": true },
```
AmjadHD, line 38:
```suggestion
            { "key": "selector", "operand": "text.html.markdown" },
            { "key": "setting.auto_match_enabled" },
            { "key": "selection_empty", "operand": false, "match_all": true },
```

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
...
+        ]
+    }
+
```
rchl, line 41: Invalid syntax - missing closing square bracket.

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
+[
+    // Auto-pair asterisks
+    { "keys": ["*"], "command": "insert_snippet", "args": {"contents": "*$0*"}, "context":
+        [
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown" },
```
rchl, line 6: Markdown can include various code blocks and it wouldn't make sense to have those keybindings active in those (typically or always).

I'm not sure what would be the ideal selector but excluding `-source` would probably be a good start.

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
...
+        "args": {"contents": "*${0:$SELECTION}*"},
+        "context": [
+            { "key": "selector", "operator": "equal", "operand": "text.html.markdown" },
+            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+            { "key": "selection_empty", "operator": "equal", "operand": false, "match_all": true },
+        ]
+    },
+
+    // Auto-pair backticks
+    { "keys": ["`"], "command": "insert_snippet", "args": {"contents": "`$0`"}, "context":
```
rchl, line 23: When starting or ending a code block you have to insert 3  backticks. This won't make it impossible but might be a bit unexpected that the first press of `` ` `` will insert two backticks already.

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,41 @@
+[
+    // Auto-pair asterisks
+    { "keys": ["*"], "command": "insert_snippet", "args": {"contents": "*$0*"}, "context":
```
rchl, line 3: This will insert two asterisks when wanting to start a list for example. Probably unwanted.

Markdown/Default.sublime-keymap
akrabat: Thanks. Updated.

Markdown/Default.sublime-keymap
akrabat: Updated.

Markdown/Default.sublime-keymap
akrabat: Updated.

Markdown/Default.sublime-keymap
akrabat: Updated.

Markdown/Default.sublime-keymap
akrabat: Thanks. Not sure what happened there! Fixed.

Markdown/Default.sublime-keymap
akrabat: Good point. I've excluded `markdown.raw`

Markdown/Default.sublime-keymap
akrabat: Oh yes. I've set it so that the second asterisk or back tick is only added if there is at least one preceding non-whitespace character.

Markdown/Default.sublime-keymap
akrabat: Fixed. The second asterisk is only added if there there is some non-whitespace in the line.

Markdown/Default.sublime-keymap
akrabat: > In the case of `"`, you can press backspace and the extra inserted `"` (totalling `"|"`, where `|` is the cursor) is also removed. This behavior seems to be missing for `.

Fixed.

akrabat: > I'd rather expect to toggle `bold` and `italic` style via the same key binding.

I can see the argument for `cmd+I` & `cmd+b`, but that's a separate PR. In this one, I'm trying to make the surrounding asterisk and back tick characters behave similar to how surrounding quotation marks and brace characters work in other code files that I write.

FichteFoll (approved):

Markdown/Default.sublime-keymap
deathaxe: Not sure whether they can really be compared as asterisk is no real quotation only markup. We also need to keep in mind that bold and italic style may be combined.

- *italic* = `*italic*` = `_italic_` 
- **bold** = `**bold**`
- ***bold italic*** = `***bold italic***` == `**_bold italic_**`

All those combinations should proberly be added and removed when pressing corresponding key binding.

deathaxe:
Would a key binding for fenced code blocks make sense, too?

```
    // Auto-pair backticks
    { "keys": ["`", "`"], "command": "insert_snippet", "args": {"contents": "```${1:<syntax>}\n$0\n```"}, "context":
        [
            { "key": "setting.auto_match_enabled"},
            { "key": "selector", "operand": "text.html.markdown - ( markup.raw - punctuation.definition.raw.begin)" },
            { "key": "preceding_text", "operator": "regex_contains", "operand": "^[ \t]*`?$" },
            { "key": "following_text", "operator": "regex_contains", "operand": "^$" },
        ]
    },
```

Markdown/Default.sublime-keymap
```diff
@@ -0,0 +1,61 @@
...
+        [
+            { "key": "setting.auto_match_enabled"},
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\S" },
```
FichteFoll, line 8: We should add a condition that prevents it from triggering when preceded by a backslash, since backslashes are used to escape markup characters. Same for `` ` ``.

akrabat: Great point. Fixed.

Markdown/Default.sublime-keymap
akrabat: Underscore is now added.

The combinations work as you'd expect.

e.g. typing `**_` will result in `**_|_**` (where `|` is your cursor). Pressing backspace three times will remove all 6 characters.

akrabat:
> Would a key binding for fenced code blocks make sense, too?

I can make typing three back ticks followed by an enter insert the closing three back ticks, but I can't work out how to make backspace remove them again as they are on separate line so the `following_text` criteria cannot be used.

I suggest that this is out-of-scope for this particular PR and could be added in a further PR.

wbond:
Thank for your work on this @akrabat, and all of the review by everyone else.

Looks like it should be some nice improvement to the user experience while working in Markdown!

rchl:
One thing that I feel is still missing is the ability to "overwrite" the backtick character when the cursor is just beneath it.

I'm talking about the case when typing a backtick in this case:

```
A `word|`
```

The last backtick should be "typed-over" rather than another backtick being added.
That is supported in JS syntax for example and feels wrong not having it here.

The only thing that is special here is that in Markdown you also type 3 backticks often and that might be a bit more awkward with "over-writing".

PR #2693, opened 2021-01-14
[Lua] Add support for a few new syntax features introduced in Lua 5.4 and a couple improvements.

Ultra-Instinct-05:
This PR

1. Adds support for the new [const](https://www.lua.org/manual/5.4/manual.html#3.3.7) and [to-be-closed](https://www.lua.org/manual/5.4/manual.html#3.3.8) variables, introduced in Lua 5.4. The syntax is a bit weird IMO, but it is what it is.
2. Previously, Lua used to have a 4th level scope to identify `true` vs `false` i.e. `constant.language.boolean.true` for `true` & `constant.language.boolean.true` for `false` (which is anyways wrong). Now, it reduces it to only `constant.language.boolean` since there is very little value in having the previous scopes. 
3. Adds `meta.string` scope to strings to comply with the scope naming guidelines. I have also reworked the string contexts to actually `push` or `set` into named contexts instead of anonymous contexts for a better debugging experience (since it is currently not possible to distinguish between anonymous contexts in the `show_scope_name` popup).
4. Lua 5.4 also introduces `coroutine.close()`. Support is added for that function. 
5. The scopes in which snippets are triggered are now more stringent and I have excluded strings & comments since it makes least sense to expand snippets in those scopes anyways.

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -646,7 +646,7 @@ contexts:
           set:
             - match: |-
                 (?x:
-                  create|isyieldable|resume|running|status|wrap|yield
+                  create|isyieldable|resume|running|status|wrap|yield|close
                 ){{identifier_break}}
               scope: meta.property.lua support.function.builtin.lua
               pop: true
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -307,8 +307,7 @@ contexts:
   variable-modifier:
     - match: '<'
       scope: punctuation.definition.modifier.begin.lua
-      push:
-        - variable-modifier-body
+      push: variable-modifier-body
 
   variable-modifier-body:
     - meta_scope: meta.modifier.lua
@@ -417,10 +416,7 @@ contexts:
       pop: true
 
   builtin:
-    - match: true{{identifier_break}}
-      scope: constant.language.boolean.lua
-      pop: true
-    - match: false{{identifier_break}}
+    - match: (?:true|false){{identifier_break}}
       scope: constant.language.boolean.lua
       pop: true
     - match: nil{{identifier_break}}
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -272,7 +272,6 @@ contexts:
         - function-arguments-meta
         - function-arguments
 
-    - include: variable-modifier
     - include: infix-operator
     - include: accessor
 
@@ -286,6 +285,7 @@ contexts:
     - match: ','
       scope: punctuation.separator.comma.lua
       push: expression-begin
+    - include: variable-modifier
     - include: expression-end
 
   expression-begin:
@@ -305,18 +305,19 @@ contexts:
     - include: else-pop
 
   variable-modifier:
-    - match: \<(?=\b(?:const|close)\b)
+    - match: '<'
       scope: punctuation.definition.modifier.begin.lua
       push:
         - variable-modifier-body
 
   variable-modifier-body:
     - meta_scope: meta.modifier.lua
-    - match: \>
+    - match: '>'
       scope: punctuation.definition.modifier.end.lua
       pop: 1
     - match: (?:const|close){{identifier_break}}
       scope: storage.modifier.lua
+    - include: else-pop
 
   infix-operator:
     - match: (?:[=<>~]=)
@@ -498,8 +499,7 @@ contexts:
   single-quoted-string:
     - match: \'
       scope: punctuation.definition.string.begin.lua
-      set:
-        - single-quoted-string-body
+      set: single-quoted-string-body
 
   single-quoted-string-body:
     - meta_include_prototype: false
@@ -512,8 +512,7 @@ contexts:
   double-quoted-string:
     - match: \"
       scope: punctuation.definition.string.begin.lua
-      set:
-        - double-quoted-string-body
+      set: double-quoted-string-body
 
   double-quoted-string-body:
     - meta_include_prototype: false
@@ -526,8 +525,7 @@ contexts:
   multiline-string:
     - match: \[(=*)\[
       scope: punctuation.definition.string.begin.lua
-      set:
-        - multiline-string-body
+      set: multiline-string-body
 
   multiline-string-body:
     - meta_scope: meta.string.lua string.quoted.multiline.lua
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -417,10 +417,10 @@ contexts:
 
   builtin:
     - match: true{{identifier_break}}
-      scope: constant.language.boolean.true.lua
+      scope: constant.language.boolean.lua
       pop: true
     - match: false{{identifier_break}}
-      scope: constant.language.boolean.true.lua
+      scope: constant.language.boolean.lua
       pop: true
     - match: nil{{identifier_break}}
       scope: constant.language.null.lua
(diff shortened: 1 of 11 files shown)


Lua/Lua.sublime-syntax
```diff
@@ -303,6 +304,20 @@ contexts:
 
     - include: else-pop
 
+  variable-modifier:
+    - match: \<(?=\b(?:const|close)\b)
```
FichteFoll, line 308: Why escape `<`? The `\b` seems very redundant, seeing you match on every character that comes before or after. 

Also doesn't allow whitespace before the attribute, which is legal.
```
> do local t < const > = 2; print(t); end
2
```

Since you're matching almost the entire token in the lookahead already, you might as well match the entire thing in a single expression. It's not like multiple attributes are allowed either.

Lua/Lua.sublime-syntax
```diff
@@ -303,6 +304,20 @@ contexts:
...
+      scope: punctuation.definition.modifier.begin.lua
+      push:
+        - variable-modifier-body
+
+  variable-modifier-body:
+    - meta_scope: meta.modifier.lua
+    - match: \>
```
FichteFoll, line 315: I'd prefer to YAML-escape this rather than regxp-escape, i.e. use `'>'`, but @deathaxe disagrees.

Lua/Lua.sublime-syntax
```diff
@@ -272,6 +272,7 @@ contexts:
         - function-arguments-meta
         - function-arguments
 
+    - include: variable-modifier
```
FichteFoll, line 275: Attributes are only legal in `local` declarations, so there is no need to add them to this context.

Lua/tests/syntax_test_lua.lua
```diff
@@ -612,3 +612,38 @@
 --                 ^ keyword.operator.assignment
 --                   ^ meta.number.integer.decimal constant.numeric.value
 --                    ^ punctuation.terminator.statement
+
+    local x <const> = 1, y <const> = 2;
```
FichteFoll, line 616: This example is invalid. It should look as follows:
```
local x <const>, y <const> = 1, 2;
```

FichteFoll (changes requested):

Lua/Lua.sublime-syntax
Thom1729: I just checked on the Lua online demo, and it looks like there can be a newline between the `<` and the modifier:

```lua
local text <
const > = "Hello world!"

io.write(text)
```

Is the lookahead required at all? Could the `<` in `local text <` ever be anything else? If so, it might be necessary to use a branch rather than a lookahead.

Thom1729:
> & `constant.language.boolean.true` for `false` (which is anyways wrong)

I'd just like to apologize for this.

Lua/Lua.sublime-syntax
deathaxe: Anyway, I prefered to make you happy with regards to HTML. :)

Lua/Lua.sublime-syntax
```diff
@@ -476,34 +491,50 @@ contexts:
...
     - match: \'
       scope: punctuation.definition.string.begin.lua
       set:
-        - meta_include_prototype: false
-        - meta_scope: string.quoted.single.lua
-        - include: string-content
-        - match: \'
-          scope: punctuation.definition.string.end.lua
-          pop: true
+        - single-quoted-string-body
```
deathaxe, line 502: I'd prefer it to be inline with `set` as this implementation is a multi-set with just a single item.

Lua/Lua.sublime-syntax
```diff
@@ -476,34 +491,50 @@ contexts:
...
     - match: \"
       scope: punctuation.definition.string.begin.lua
       set:
-        - meta_include_prototype: false
-        - meta_scope: string.quoted.double.lua
-        - include: string-content
-        - match: \"
-          scope: punctuation.definition.string.end.lua
-          pop: true
+        - double-quoted-string-body
```
deathaxe, line 516: Same here.

Lua/Lua.sublime-syntax
```diff
@@ -476,34 +491,50 @@ contexts:
...
+  multiline-string:
     - match: \[(=*)\[
       scope: punctuation.definition.string.begin.lua
       set:
-        - meta_scope: string.quoted.multiline.lua
-        - meta_include_prototype: false
-        - match: \]\1\]
-          scope: punctuation.definition.string.end.lua
-          pop: true
+        - multiline-string-body
```
deathaxe, line 530: and here

Ultra-Instinct-05:
> > & `constant.language.boolean.true` for `false` (which is anyways wrong)
> 
> I'd just like to apologize for this.

There was no need for apologizing at all. It's just something I noticed so decided to fix it.

Lua/Lua.sublime-syntax
Ultra-Instinct-05: I just tried using `('>')` instead of `\>` but that doesn't seem to work.

Lua/Lua.sublime-syntax
Ultra-Instinct-05: In which context should I add them ? I added them in `infix-operator` so that the `<` in attributes is matched before `<` in infix operator.

Lua/Lua.sublime-syntax
Ultra-Instinct-05: > Either way, you should add a test case that a missing > recovers gracefully

`else-pop` to the rescur :) I forgot to add that so that missing `>` doesn't break the highlighting.

Lua/Lua.sublime-syntax
FichteFoll: Probably introduce one that starts on `local`. That way it can't be confused with a comparison operator.

Lua/Lua.sublime-syntax
wbond: There shouldn't be parens around the single quotes

Ultra-Instinct-05:
I believe I have addressed all of the review comments. Let me know if any of them are wrong or more corrections needs to be done.

Lua/Lua.sublime-syntax
```diff
@@ -305,18 +305,19 @@ contexts:
     - include: else-pop
 
   variable-modifier:
-    - match: \<(?=\b(?:const|close)\b)
+    - match: '<'
       scope: punctuation.definition.modifier.begin.lua
       push:
         - variable-modifier-body
```
deathaxe, line 311: I'd suggest to never use multi-push or multi-set to push/set only one context.

Lua/Lua.sublime-syntax
```diff
@@ -402,10 +418,10 @@ contexts:
 
   builtin:
     - match: true{{identifier_break}}
-      scope: constant.language.boolean.true.lua
+      scope: constant.language.boolean.lua
       pop: true
     - match: false{{identifier_break}}
-      scope: constant.language.boolean.true.lua
+      scope: constant.language.boolean.lua
```
deathaxe, line 424: If both share the same scope patterns could be merged as well, IMHO.

```
- match: (?:false|true){{identifier_break}}
```

Lua/Lua.sublime-syntax
Ultra-Instinct-05: Right, sorry. I forgot that one.

deathaxe: That's why 4 eyes. :-)

Ultra-Instinct-05:
Thanks for the review. I have addressed both of the issues.

FichteFoll (approved):

wbond:
Thanks for your fixes here @Ultra-Instinct-05! As always, thank you for the reviews @FichteFoll and @deathaxe!

PR #2694, opened 2021-01-17
[JSON] Fix auto-pair and auto-indent while typing

deathaxe:
This PR proposes to add some key bindings which fix some core functions for JSON syntax.

Some more details may be found in commit messages.


Ultra-Instinct-05 (approved):

Ultra-Instinct-05:
Thanks for the auto indent in square brackets. That should make life a bit easier for me.

FichteFoll:
I've been wondering why auto indent isn't enabled by default for square brackets globally.

PR #2695, opened 2021-01-17
[Python] Add interpolation keybinding.

Ultra-Instinct-05:
This PR adds a key binding to help with interpolation in strings in Python.

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -646,7 +646,7 @@ contexts:
           set:
             - match: |-
                 (?x:
-                  create|isyieldable|resume|running|status|wrap|yield
+                  create|isyieldable|resume|running|status|wrap|yield|close
                 ){{identifier_break}}
               scope: meta.property.lua support.function.builtin.lua
               pop: true
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -272,7 +272,6 @@ contexts:
         - function-arguments-meta
         - function-arguments
 
-    - include: variable-modifier
     - include: infix-operator
     - include: accessor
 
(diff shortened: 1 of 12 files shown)


Ultra-Instinct-05:
Not sure how did all those Lua commits got added.

deathaxe:
Wrong local base branch?

Ultra-Instinct-05:
I guess so. Is there an easy way to fix it ? If not, I'll just close this one and make a fresh PR.

deathaxe:
You should be able to just rebase interactive onto master and drop all those Lua commits.

If you use merge, you can just right click each lua commit and use "Edit Commit / Drop ... "

Ultra-Instinct-05:
So like this ? Sorry, have to ask but I haven't done a rebase before.

image

deathaxe:
I guess you worked on Lua and than just created a the branch `add_interpolation_keybinding` at the last commit of your lua branch, so it is based on `lua-5.4-features` instead of `master`.

Merge doesn't support interactive rebasing unfortunatelly, so I guess rebasing is not the proper strategy with that tool. What I had in mind when suggesting it, was GitExtensions, which is very powerful with such workflows.

Im Merge you have two options, assuming my first statement is correct.

1.) With `add_interpolation_keybindings` checked out, you can simply drop all lua commits via context menu. Not sure if Merge learned to do it in a batch or if you still need to do it for each commit separately.

2.) You could create a `backup` branch at where you are now and then "reset hard" `add_interpolation_keybindings` to `master`. After that you can cherry pick the most recent commit.

Ultra-Instinct-05:
Looks like more trouble than necessary for a simple PR. I think I'll just close this one and open a fresh PR properly. Sorry for this fiasco ;(

rwols:
You can do the following in a terminal:

```
git checkout master
git checkout -b add_interpolation_keybinding2
git cherry-pick 67fad00
```

FichteFoll:
You can also cherry-pick with SM but unfortunately only one at a time. Alternatively, you could do an interactive rebase from `master` to your current HEAD and drop all unrelated commits.

PR #2696, opened 2021-01-17
[Python] Add a helper keybinding for interpolation.

Ultra-Instinct-05:
This PR adds a helper key binding for Python interpolation within strings.

### DIFF
--- a/Python/Default.sublime-keymap
+++ b/Python/Default.sublime-keymap
@@ -22,4 +22,13 @@
             { "key": "eol_selector", "operator": "not_equal", "operand": "string.quoted.single - punctuation.definition.string.end", "match_all": true }
         ]
     },
+
+    // Helper keybinding for interpolation.
+    { "keys": ["{"], "command": "insert_snippet", "args": {"contents": "{$0}"}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operator": "equal", "operand": "source.python, meta.string", "match_all": true },
+        ]
+    },
 ]
--- a/Python/Default.sublime-keymap
+++ b/Python/Default.sublime-keymap
@@ -28,7 +28,8 @@
         [
             { "key": "setting.auto_match_enabled" },
             { "key": "selection_empty", "match_all": true },
-            { "key": "selector", "operator": "equal", "operand": "source.python, meta.string", "match_all": true },
+            { "key": "selector", "operator": "equal", "operand": "source.python string", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?<![\\\\])([\\\\]{2})*$", "match_all": true },
         ]
     },
 ]


Python/Default.sublime-keymap
```diff
@@ -22,4 +22,13 @@
             { "key": "eol_selector", "operator": "not_equal", "operand": "string.quoted.single - punctuation.definition.string.end", "match_all": true }
         ]
     },
+
+    // Helper keybinding for interpolation.
+    { "keys": ["{"], "command": "insert_snippet", "args": {"contents": "{$0}"}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operator": "equal", "operand": "source.python, meta.string", "match_all": true },
```
rwols, line 31: That comma should not be there (it's the OR operator). The selector should be `source.python meta.string`.

deathaxe, line 31: By using `,` you literly enable this key binding for all strings in all syntaxes.

On the other hand normal key bindings already work once `string` is cleared within the interpolation.

```py
s = f"before { a + b } after"
#            ^^^^^^^^^ - string -> auto-pair already works within that block and should do so like in the rest of the code.
```

Hence I'd suggest a selector such as `source.python string`.

We may also want to not trigger auto-pairing in case of typing escaped braces such as `\{` or `\\\{` etc.

```
{ "key": "preceding_text", "operator": "regex_contains", "operand": "((?<![\\\\])([\\\\]{2})*)$", "match_all": true },
```

Ultra-Instinct-05: Thanks. Not sure why I put that comma anyways :/
Good catch on the escaped braces.

deathaxe:
FWIW auto-pairing in tuples, lists and objects in front of `,` doesn't work as as well in python.

Hence I'd vote to use the following context in _Default/Default.sublime-keymap_ instead of adding a key binding for each separate syntax.

```
{ "key": "following_text", "operator": "regex_contains", "operand": "^(?:[ -+*%,:;\\/\t\\)\\]\\}\"']|$)", "match_all": true }
```

see: https://github.com/sublimehq/sublime_text/issues/3829

Ultra-Instinct-05:
If that's the case (wasn't aware that an issue existed for auto pairing in strings), Is it beneficial adding these here ?

deathaxe:
String interpolation is quite special and supported by a couple of syntaxes only, so yes it might be best placed here.

My last comment was more about auto-pairing in general. Just try to create a python object and try auto-pairing. It fails very much the same way as JSON does atm.

```py
obj = {
    key: |,   # with `|` being the caret try to add a double quoted string. -> fails
}
```

Python/Default.sublime-keymap
deathaxe, line 31:
```suggestion
            { "key": "selector", "operator": "equal", "operand": "source.python string", "match_all": true },
            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?<![\\\\])([\\\\]{2})*$", "match_all": true },
```

Ultra-Instinct-05: @deathaxe I'll commit your suggestion as currently I don't have the time to look at it.

FichteFoll:
`{` not working in Python gets me more often that I'd like to admit. To me, this is a general issue and not a Python-specific one, although I prominently notice it in Python, since it has special syntax for it that can easily break.

Ultra-Instinct-05:
Closing in favor of #2889

PR #2699, opened 2021-01-21
[XML] Enhance inheritability

deathaxe:
This PR applies some strategic changes to XML.sublime-syntax after the scheme of HTML.sublime-syntax which intend to help 3rd-party packages with inheriting XML and modifying single aspects of the syntax.

All changes are grouped and described in single commits.

Syntax tests are updated to

1. test for added `meta.string` scopes
2. test missing attribute name situations (only `="value"` present)

--- 
Note:

I wonder whether we also should rename contexts to add a `xml-` prefix. The idea is to make it obvious which contexts are inherit from XML or which are added by an extending syntax definition. It would also "look nice" next to the already existing `dtd-` contexts.

This change is **not** yet part of this PR and may be handled separately because it is a breaking change. PackageDev is known to use some parts of XML, which would break then, but might be fixable easily.

### DIFF
--- a/XML/XML.sublime-syntax
+++ b/XML/XML.sublime-syntax
@@ -534,17 +534,15 @@ contexts:
         3: entity.other.attribute-name.xml punctuation.separator.namespace.xml
         4: entity.other.attribute-name.localname.xml
         5: invalid.illegal.bad-attribute-name.xml
-      push: tag-attribute-separator-key-value
-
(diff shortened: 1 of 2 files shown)


wbond:
Thanks for these improvements!

If we were going to rename the contexts, now would probably be the time. I'm mixed on the topic. I see the benefit, but it is nice to not break things.

PR #2701, opened 2021-01-23
[CSS] Add auto-pair keybindings to /*|*/ <=> /* | */

deathaxe:
This PR proposes to improve auto-pairing user experience by automatically adding or removing space to the left and right of caret in new/empty block comments.

**Note:**

This keybinding could also be implemented in Default.sublime-package by just removing `source.css` from selector. The remaining `comment` would ensure the key binding is triggered only within block comments. This would enable this enhancement to all syntaxes which support this style of block cumments such as: C, C++, ObjC, ObjC++, Java, JavaScript, JSON and probably numerous more.

### DIFF
--- a/CSS/Default.sublime-keymap
+++ b/CSS/Default.sublime-keymap
@@ -73,6 +73,26 @@
 		]
 	},
 
+	// Expand /*|*/ to /* | */ when space is pressed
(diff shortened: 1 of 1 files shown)


FichteFoll:
+1 for adding this to the default bindings. The context (without `selector`) is specific enough to have it active for *all* syntaxes, imo.

PR #2705, opened 2021-01-28
[CSS] Update property names from referenced website

### LINKED ISSUE #2704: [CSS] support latest gap which replaces grid-gap
- Sublime Version: 4095
- OS Version: Windows 10 19042

```css
.flex-or-grid {
  gap: 1em;
}
```

image

deathaxe:
Fixes #2704

This PR...

1. Adds all property names returned by https://www.chromestatus.com at 2021 Jan. 28th except those which begin with `alias-` or `webkit-` in exact the order provided by the page.
2. Properties no longer returned by the web page are appended to that list in case those are still valid or used. They are considered to be of very low usage.

   A comment was added as devider.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -276,85 +276,144 @@ variables:
     ){{break}}
 
   # Property names are sorted by popularity in descending order.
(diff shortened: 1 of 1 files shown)


wbond:
Thank you!

PR #2707, opened 2021-01-30
[Markdown] Improve auto-pairing key bindings

deathaxe:
Please refer to commit messages for details about the changes.

### DIFF
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -6,8 +6,8 @@
         "args": {"contents": "*$0*"},
         "context": [
             { "key": "setting.auto_match_enabled"},
-            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "\\S", "match_all": true },
             { "key": "preceding_text", "operator": "not_regex_contains", "operand": "\\\\$", "match_all": true }
         ]
@@ -17,9 +17,9 @@
         "command": "insert_snippet",
         "args": {"contents": "*${0:$SELECTION}*"},
         "context": [
-            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "setting.auto_match_enabled"},
-            { "key": "selection_empty", "operand": false, "match_all": true }
+            { "key": "selection_empty", "operand": false, "match_all": true },
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" }
         ]
     },
     {
@@ -42,8 +42,8 @@
         "args": {"contents": "_$0_"},
         "context": [
             { "key": "setting.auto_match_enabled"},
-            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "preceding_text", "operator": "not_regex_contains", "operand": "\\\\$", "match_all": true }
         ]
     },
@@ -52,9 +52,9 @@
         "command": "insert_snippet",
         "args": {"contents": "_${0:$SELECTION}_"},
         "context": [
-            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "setting.auto_match_enabled"},
-            { "key": "selection_empty", "operand": false, "match_all": true }
+            { "key": "selection_empty", "operand": false, "match_all": true },
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" }
         ]
     },
     {
@@ -77,8 +77,8 @@
         "args": {"contents": "`$0`"},
         "context": [
             { "key": "setting.auto_match_enabled"},
-            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "\\S", "match_all": true },
             { "key": "preceding_text", "operator": "not_regex_contains", "operand": "\\\\$", "match_all": true }
         ]
@@ -89,8 +89,8 @@
         "args": {"contents": "`${0:$SELECTION}`"},
         "context": [
             { "key": "setting.auto_match_enabled"},
-            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
-            { "key": "selection_empty", "operand": false, "match_all": true }
+            { "key": "selection_empty", "operand": false, "match_all": true },
+            { "key": "selector", "operand": "text.html.markdown - markup.raw" }
         ]
     },
     {
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -97,7 +97,7 @@
     {
         "keys": ["`", "`"],
         "command": "insert_snippet",
-        "args": {"contents": "```${1:language}\n$0\n```"},
+        "args": {"contents": "```${0:language}\n\n```"},
         "context": [
             { "key": "setting.auto_match_enabled"},
             { "key": "selection_empty", "match_all": true },
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -77,7 +77,7 @@
         "context": [
             { "key": "setting.auto_match_enabled"},
             { "key": "selection_empty", "match_all": true },
-            { "key": "selector", "operand": "text.html.markdown - markup.raw" },
+            { "key": "selector", "operand": "text.html.markdown - markup.raw - meta.code-fence" },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "(?<![\\\\])([\\\\]{2})*$", "match_all": true }
         ]
     },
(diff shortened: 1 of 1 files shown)


Ultra-Instinct-05:
Could we perhaps have a Markdown specific setting that gates all these key bindings behind a setting. In that way users that want to use these key bindings can use them and those that don't want to use them also can disable them.

```js
{ "key": "setting.enable_default_keybindings", "operator": "equal", "operand": "true" },
```

deathaxe:
All bindings of this package belong to the "auto-pairing" category and just intend to fill some syntax specific gaps of core key bindings. All of them are gated by `{ "key": "setting.auto_match_enabled" }` already, which may be disabled in a syntax specific settings file.

I personally would prefer a more smart plugin based solution to toggle bold and italics via key binding, but that may be a topic for a 3rd party package as there are more features to implement in order to improve UX with regards to writing lists, auto-complete footnote references, etc.

Ultra-Instinct-05:
> `{ "key": "setting.auto_match_enabled" }`

That would mean, I would also have to disable auto pairing of quotes, brackets in Markdown, just to disable these.

deathaxe:
To repeat myself ...

All bindings of this package belong to the "auto-pairing" category.

Ultra-Instinct-05:
Alright, so how do I disable them (I don't want to define user specific overrides) without having to sacrifice auto pairing of quotes, brackets etc ?

deathaxe:
1. This PR just addresses some issues with already existing bindings (except adding auto-pairing for fenced code blocks).
2. Backtick quoted bindings are definitly part of `setting.auto_match_enabled` and I don't see a reason to change that.
2. As I said, I am not very confident about auto-pairing empty `*` and `_` by just typing those characters in general, as at least the first one hurts when typing mathematical equations such as `10 * 10`. I'd either remove those two or better would like to have a smarter plugin based solution. But I don't forsee when that will happen.

mitranim:
Seems quite reasonable! Interesting idea to make a fenced code block on two consecutive backtick presses. I have only one suggestion: using `${0:language}\n\n` for that particular snippet, to avoid the inconveniences associated with `$1+` placeholders. It's just about as easy to press `↓` after typing or deleting the language tag, but this doesn't take over your Tab sequence. Not a big deal though.

Markdown/Default.sublime-keymap
```diff
@@ -97,7 +97,7 @@
     {
         "keys": ["`", "`"],
         "command": "insert_snippet",
-        "args": {"contents": "```${1:language}\n$0\n```"},
+        "args": {"contents": "```${0:language}\n\n```"},
```
mitranim, line 100: Nice, thanks!

mitranim:
Realized one more thing. Markdown code fences aren't strictly 3 backticks, they're 3+. The closing fence must match the opening fence. Like this:

``````md
`````md
````md
```md
# Header

* Bullet
```
````
`````
``````

Both ST and GitHub correctly support this feature. How does this interact with double-backtick-autofence (which I do like)? Admittedly most users don't know about this, but when someone _needs_ to use it, they might be slightly inconvenienced.

**Comment by deathaxe** [comment]
The key binding only triggers at the beginning of a line and always adds 3 backticks. It doesn't trigger if backticks already exist and is not designed to add arbitary numbers of backticks, which is also tricky due to the "typing nature" of those characters.

Basically all of that can also be achieved by a snippet such as:

```xml
<snippet>
  <content><![CDATA[```${1:language}
${0:${SELECTION:code}}
```]]></content>
  <tabTrigger>mdc</tabTrigger>
  <scope>text.html.markdown.multimarkdown, text.html.markdown</scope>
  <description>Markdown Codeblock</description>
</snippet>
```

_You may easily modify it to add more backticks._

The main motivation of that binding is just that I found myself fiddeling around with those very often. It's annoying to need to quickly type `backtick->backtick->backtick->space` for opening and closing code blocks each.

Just typing `backtick->backtick` once at a beginning of a line feels much more natural than trying to remember such cryptic snippet triggers, which I didn't even know about.

Markdown/Default.sublime-keymap
```diff
@@ -61,33 +48,62 @@
...
             { "key": "selection_empty", "match_all": true },
             { "key": "selector", "operator": "equal", "operand": "punctuation.definition.raw.begin.markdown" },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "`$", "match_all": true },
             { "key": "following_text", "operator": "regex_contains", "operand": "^`", "match_all": true }
         ]
     },
+
+    // Add fenced code block
+    {
+        "keys": ["`", "`"],
```
FichteFoll, line 98: I believe bindings with two char keys to be bad UX for the most part, since they consume the first key  until another key is pressed that either triggers a multi-chord binding or matches no bindings and only then are the previously hold-back key presses handled (or a timeout happens). That means after typing the first `` ` `` ST would seemingly do nothing.


It's okay for custom bindings imo or those that don't produce a character, but for character-based bindings I'd avoid them with a `preceding_text` context. It also means they are more flexible since they key presses don't have to be input in consecutively.

Markdown/Default.sublime-keymap
```diff
@@ -1,27 +1,20 @@
 [
     // Auto-pair asterisks
-    { "keys": ["*"], "command": "insert_snippet", "args": {"contents": "*$0*"}, "context":
```
FichteFoll, line 3: There was definitely a problem with this binding that extended beyond just typing formulas, since for example trying to add emphasis on a word has already been typed i.e. `some |words` would also trigger it, resulting in `some *|*words`, which is not how quotation auto pairing works, for example.

I would say that if you intend to input a literal `*` you *should* be escaping it with a backslash, i.e. when writing a formula, but without the context fixes fro the first part of my comment, they should not be re-added as-is.

deathaxe: Markdown is quite smart in general with regards to distinguish literal `*` from bold/italics markup and I don't think it is acceptable to force everybody to escape them also in situations it is not needed.

Also the first example `some |word` is not so clear. How do you want to know whether typing an asterisk in front of the word should wrap the whole word vs. adding an empty bold/italic char/word. It might easily be someone wants to type a bold letter like `some |ord`  => `some *w*ord`.

Markdown/Default.sublime-keymap
deathaxe: The reason why nothing is printed when typing the first backtick is not the snippet, but the nature of the caracter itself.

You can also type a plain backtick with all bindings disabled and it will not be added until you hit space or another character it will then be placed on top of.

FichteFoll: > The reason why nothing is printed when typing the first backtick is not the snippet, but the nature of the caracter itself.

This is incorrect for the non-dead backtick key, as I have it on my keyboard layout (and so does the US layout iirc).

In fact, when I try this locally with ``{ "keys": ["`", "`"], "command": "new_file", },`` it doesn't even work?

Markdown/Default.sublime-keymap
FichteFoll: > How do you want to know whether typing an asterisk in front of the word should wrap the whole word

By selecting the word when I want do do that. This would match the quotation character matching, which I think of as reasonable and have gotten used to.

Markdown/Default.sublime-keymap
deathaxe: To clarify my statement, here is a screencast taken with Windows Notepad. Please see the printed key strokes to see how typing a backtick never prints the character without another key stroke following.

Animation

In fact `{ "keys": ["`", "`"], "command": "new_file", },` works very well for me. Here the proof for those who don't believe.

Animation

Markdown/Default.sublime-keymap
deathaxe: ... which is still possible.

Markdown/Default.sublime-keymap
jwortmann: This behavior happens because <kbd>\`</kbd> is a dead key in the german keyboard layout.
However, I have also configured my own keyboard layout and removed the dead key for <kbd>\`</kbd>, but I still have no problems when I try the auto-pair keybinding from this PR, i.e. it works fine and the first \` is still inserted as expected.

Markdown/Default.sublime-keymap
FichteFoll: I think we might be talking past each other.

My primary concern about removing the binding altogether was that the probably most common use case of `a word |` => `a word *|*` will not be possible anymore. However, I do believe not having the binding is better than the current situation on `master`, which showed the behavior I outlined in the first comment of this thread, which is terrible. Typing <kbd>*</kbd> with a selection is unrelated to this.

mitranim: I think we should not auto-pair `* _ ~` by default, because of many false positives. For built-in keys, it's better to lean on the side of fewer features with fewer false positives, rather than more features with more false positives, because it's easier to opt in than to opt out.

Markdown/Default.sublime-keymap
FichteFoll: I think the binding not working is just an ST bug with my configuration. If it works for you, then it's all good. The binding *is* triggered for ``["`"]``, fwiw.

mitranim:
Suggestion: it might be reasonable to add `- source` or `- markup.raw.code-fence` to the contexts to prevent them from accidentally triggering in embedded code.

deathaxe:
All of them are already gated by ` - markup.raw - meta.code-fence`, `markup.raw - markup.raw.code-fence - meta.code-fence` , or `punctuation.definition.raw.begin.markdown` which should be quite sufficient already.

PR #2708, opened 2021-01-30
[Markdown] Promote supported code fence syntax in completions

deathaxe:
This PR adds a sublime-completions file with all supported syntax names for fenced code blocks supported by this package.

The goal is to help users supported syntax names quickly, when typing after the opening backticks of a fenced code block.

### DIFF
--- a/Markdown/Code Block Syntaxes.sublime-completions
+++ b/Markdown/Code Block Syntaxes.sublime-completions
@@ -1,0 +1,362 @@
+{
+	"scope": "text.html.markdown meta.code-fence.definition.begin",
+	"completions": [
+		{
+			"trigger": "bash",
+			"contents": "bash",
+			"annotation": "Bourne Against Shell",
+			"kind": ["markup", "s", "Syntax"],
+			"details": "Specifies <code>ShellScript</code> code highlighting"
+		},
+		{
+			"trigger": "c",
+			"contents": "c",
+			"annotation": "C",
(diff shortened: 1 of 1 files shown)


mitranim:
Interesting, may help users discover what's supported. For example I had no idea that Go was supported (via `golang`), as I always simply typed `go`. This relies on the set of supported syntaxes being hardcoded in `Markdown.sublime-syntax`, but that's true for the foreseeable future.

Markdown/Code Block Syntaxes.sublime-completions
```diff
@@ -0,0 +1,362 @@
+{
+	"scope": "text.html.markdown meta.code-fence.definition.begin",
+	"completions": [
+		{
+			"trigger": "bash",
+			"contents": "bash",
```
FichteFoll, line 6: `contents` can be omitted if they are the same as the trigger.

deathaxe: I just followed the scheme of other Default Packages, whose completions were converted to support kind info. All of them explicitly contain `contents`.

FichteFoll (approved):

deathaxe:
The only thing I realized recently is ST's auto completer to become confused with `c++` values. It doesn't correctly replace the existing token as soon as a `+` is present.

wbond:
Looks nice, thanks!

PR #2716, opened 2021-02-11
Remove .DS_Store file that made its way into the repo

frou:
It got committed as part of this: https://github.com/sublimehq/Packages/pull/2645


Thom1729 (approved):
MY BAD

frou:
Hehe, I only noticed due to having a `launchd` job that runs over my entire filesystem and deletes every `.DS_Store` (except for `~/.DS_Store`). So a deletion ready to be committed showed up in `git status`

wbond:
Thanks!

FichteFoll:
I suggest everyone on mac to add `.DS_Store` to their global gitignore, btw.

PR #2722, opened 2021-02-16
[LaTeX] Add scope for gnuplottex enviroment

galvaniccoffee:
This commit adds correct scoping for the `gnuplot` enviroment from the [gnuplottex package](https://www.ctan.org/pkg/gnuplottex).

The gnuplottex package allows users to include verbatim gnuplot scripts inside a `gnuplot` enviroment in their LaTeX files, which will be run during compilation and the resulting graphs inserted into the document.
Without proper scoping, ST will try to parse these scripts as LaTeX code. This will not only result in incorrect syntax highlighting inside the enviroment, but potentially also afterwards since gnuplot can use single `$` signs, which start a LaTeX math enviroment.
This commit assigns the proper embedded scope to the code inside a `gnuplot` enviroment, similar to the source code enviroments from the listings or minted package.

Note that per default, ST does not have syntax support for gnuplot code itself, so it will be displayed as plain text (which is still better than interpreting it as LaTeX). However, such support can be installed by the user, e. g. using the  [ST Gnuplot Package](https://packagecontrol.io/packages/Gnuplot)).

<details>
    <summary>Example screenshots</summary>

**Before:**

**After:**
(with additional gnuplot syntax support curtosy of the [ST Gnuplot Package](https://packagecontrol.io/packages/Gnuplot))

</details>

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -765,6 +765,7 @@ contexts:
     - include: pkgcomment
     - include: beamer
     - include: pkgarray
+    - include: gnuplottex
 
   # listings package
   pkglistings:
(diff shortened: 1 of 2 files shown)


galvaniccoffee:
Checks are failing due to:

     Error loading embed in syntax file "Packages/LaTeX/LaTeX.sublime-syntax": Unable to find syntax file for scope "source.gnuplot"

Welp, I did not anticipate that ST not including native support for the gnuplot syntax itself would be a problem. Any ideas what to do about this?

michaelblyons:
If you're on ST4, make your own "LaTeX with GnuPlot".sublime-syntax that `extends` LaTeX.sublime-syntax to inject your extra rules.

galvaniccoffee:
I'm on ST3, but I suppose I could try to migrate to 4 and do that one day.

Still though, since gnuplottex is included in standard LaTeX distributions and using the environment once in a document can break all syntax highlighting after it, I think something should be done here.

FichteFoll:
I'm pretty sure this won't be accepted because we're only including syntaxes that ST ships by default. That's also why the tests fail.

You have two options in ST to apply this change for you locally.

1. Modify the syntax with an override file. You probably already did that by extracting the syntax file into a LaTeX package in your normal Packages folder. I suggest using [OverrideAudit](https://github.com/OdatNurd/OverrideAudit) when doing this, since it will notify you of upstream changes. Unlikely to happen before ST4, however.
2. (Requires ST4) Define a new syntax that `extends` the existing syntax like @michaelblyons suggested. ([docs](https://www.sublimetext.com/docs/syntax.html#inheritance:ver-dev))

deathaxe:
This PR is a nice example for a situation in which it was cool, if a 3rd-party package could just hook into an existing context per scope. That's what many packages of VS Code use, to extend syntax support for fenced codeblocks in Markdown for instance.

Just install any 3rd party syntax and get fenced code block highlighting for free.

deathaxe:
This PR contains 3rd-party syntax definitions. While `embed` fails gracefully in case embedded syntax doesn't exist I don't think we should start introducing dependencies to external packages.

The way to go at current state was for end users to inherit from core syntaxes and add such rules.

I'd therefore close this PR.

keith-hall:
Agreed, especially as we have no way to get CI to pass with references to third party syntaxes, and we don't want to move the syntax here and maintain it in this repo.

PR #2723, opened 2021-02-18
[HTML] Fix foreign tags

### LINKED ISSUE #2713: [HTML] ease syntax parsing of custom/foreign tags
- Sublime Version: 3.2.2 (3211)
- OS Version: Big Sur

``` html
<livewire:element>
```

In this case the colons shouldn't be marked as errors. The same used to be true for dots but was fixed for ST4 in #2407.

Related discussion: https://forum.sublimetext.com/t/how-to-remove-dot-marks-dont-know-how-to-name-it/52550
Related PR: #1784 by @deathaxe, but it doesn't actually fix the problem.

Update: For the dot this is fixed in ST4 (see below), it's still a problem with the colon (:), though.
Update2: changed title as per [this comment](https://github.com/sublimehq/Packages/issues/2713#issuecomment-780078417)

deathaxe:
Fixes #2713

This PR removes strict lower-case custom tag highlighting.

Why?

https://html.spec.whatwg.org/multipage/syntax.html#start-tags doesn't mension custom elements but foreign elements only.

> In the HTML syntax, tag names, even those for foreign elements, may be written with any mix of lower- and uppercase letters that, when converted to all-lowercase, matches the element's tag name; tag names are case-insensitive.

Actually Firefox, Chrome and VS Code accept those colons or seem to fail silently by ignoring the namespace part. Not sure how foreign tags such as SVG or MathML are handled in detail though. They may support namespaces as they are XML.

According to https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state all characters except ascii whitespace and some punctuation are valid in tag names. That's the rule HTML (Plain).sublime-syntax applies in general.

The only thing lost by this commit is illegal highlighting of uppercase letters in unrecognized tags.

**Note:**

During work an edge case which might effect templating syntaxes was found. The first character after `<` must be an ascii, so HTML doesn't push into "tag context" if `<` is directly followed by interpolations such as `{{ ... }}`. That must be handled by the templating syntax as it is the only one to know which kinds of interpolation tags are valid.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -43,25 +43,6 @@ variables:
       | text/livescript
     )
 
-  custom_element_char: |-
-    (?x:
-      # https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts
-        [-._a-z0-9\x{00B7}]
-      | [\x{00C0}-\x{00D6}]
-      | [\x{00D8}-\x{00F6}]
-      | [\x{00F8}-\x{02FF}]
-      | [\x{0300}-\x{037D}]
-      | [\x{037F}-\x{1FFF}]
-      | [\x{200C}-\x{200D}]
-      | [\x{203F}-\x{2040}]
-      | [\x{2070}-\x{218F}]
-      | [\x{2C00}-\x{2FEF}]
-      | [\x{3001}-\x{D7FF}]
-      | [\x{F900}-\x{FDCF}]
-      | [\x{FDF0}-\x{FFFD}]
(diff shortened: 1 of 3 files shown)


dakira:
I "backported" this to ST3 and it works nicely for me. Should I submit a PR for the st3 branch?

deathaxe:
I don't think that branch will receive any updates anymore.

wbond:
> I "backported" this to ST3 and it works nicely for me. Should I submit a PR for the st3 branch?

The st3 branch exists strictly for reference. There won't be any more releases of ST3.

PR #2726, opened 2021-02-21
[C++/Objective-C++] Add C++20 char8_t, consteval, and constinit

trflynn89:


rwols (changes requested):
You need to mirror these changes in Objective-C++ as well.

trflynn89:
@rwols Done

wbond:
Looks good, thanks for adding these!

PR #2729, opened 2021-02-27
[Diff] Add setting to not trim trailing white space on save

keith-hall:
When one diffs a file whose name contains a space character, the diff is generated with a trailing tab character at EOL so that the `patch` command line utility will be able to find the file properly. Prior to this PR, ST would automatically trim the trailing whitespace by default, preventing `patch` from working.

Real world example of a patch file which targets a file with a space in the name: https://github.com/sharkdp/bat/pull/1557/files#diff-e8ae3ce548991704395140abe884a9b871bd1a85759a349d50ac84f465a25683R3

### DIFF
--- a/Diff/Diff.sublime-settings
+++ b/Diff/Diff.sublime-settings
@@ -1,0 +1,3 @@
+{
+    "trim_trailing_white_space_on_save": false,
+}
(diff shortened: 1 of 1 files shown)


michaelblyons:
See also: https://github.com/SublimeText/TrailingSpaces/#ignore-scope

deathaxe:
Would it also make sense to disable tab translation etc.?

keith-hall:
Thanks for the suggestion - I was also thinking of it :) done :)

BenjaminSchaaf:
If tab translation should be disabled then indentation detection should be disabled too, otherwise it will be automatically re-enabled if the indentation is detected to use spaces.

PR #2731, opened 2021-03-01
[CSS] Add font-display constants

adamchainz:
https://www.w3.org/TR/css-fonts-4/#font-display-desc

The "fallback" was in the syntax previously but removed in #2705 - an accident?

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -436,7 +436,7 @@ variables:
     | avoid(-column|-page|-region)?
     | background(-color|-image|-position|-size)?
     | backwards|balance|baseline|below|bevel|bicubic|bidi-override|blink
-    | block(-line-height)?
+    | block-line-height
     | blur
     | border(-bottom|-left|-right|-top)?-(color|radius|width|style)
     | border-(bottom|top)-(left|right)-radius
@@ -586,7 +586,7 @@ variables:
   # https://www.w3.org/TR/css-fonts-4/#font-display-desc
   font_display_constants: |-
     \b(?xi:
-    auto | block | swap | fallback | optional
+    block | swap | fallback | optional
     ){{break}}
 
   # Generic Font Families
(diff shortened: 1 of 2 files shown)


deathaxe:
Thanks for your contribution. While `fallback`, `swap` and `optional` have not been matched as valid CSS constants so far, `auto` and `block` are part of the `unsorted_property_constants` variable already.

While `block` may be a pure font-related constant `auto` is known to be used in several situations. I hit that "issue" when creating those special purpose variables, too. One reason to not go further with it.

Not sure how smart ST is with regards to remove duplicates, but maybe we should try to keep only one copy of a token. So `block` may be removed from unsorted_property_values while `auto` should live only there.

Note:

PR #2705 addresses property names only. It does not manipulate value related patterns. 

It does remove `fallback` from the list of valid `property_names` though. Seems the comparison between the new and the old names list lacked to reveal it as previously existing. But I also haven't found any evidence for it to be valid anywhere.

adamchainz:
> Not sure how smart ST is with regards to remove duplicates, but maybe we should try to keep only one copy of a token. So block may be removed from unsorted_property_values while auto should live only there.

Okay fair enough

> But I also haven't found any evidence for it to be valid anywhere.

I don't follow - it's listed as valid in the spec? https://www.w3.org/TR/css-fonts-4/#font-display-desc

deathaxe:
> I don't follow - it's listed as valid in the spec?

Your PR and the link point to `fallback` as value such as `font-display: fallback;`

The PR blamed to have broken it was about `fallback` as a property-name. A totally different use case. I need to correct my previous comment though as it was not removed but only restricted to be matched within `@counter-style` property lists only.

adamchainz:
> The PR blamed to have broken it was about fallback as a property-name. A totally different use case. 

Ah my bad I missed the distinction.

> So block may be removed from unsorted_property_values while auto should live only there.

I've pushed a commit with this change.

deathaxe (approved):

wbond:
Thanks @adamchainz and @deathaxe!

Issue #2732, opened 2021-03-02
[TypeScript] `<=` ligature is not rendered properly


orangecoloured:
I'm using FiraCode font and in my .tsx/.ts files the `<=` sign doesn't work.
`>=` works and I haven't noticed any other non-working ligatures.

If I create a new file and no syntax is applied the `<=` works fine.

- Sublime Version: 4098
- OS Version: macOS 11.2.2

``` ts
if (a <= b) { doSmth(); }
```

PR #2733, opened 2021-03-02
[TypeScript] Fix scope for less (or equal) than operators

jfcherng:
Also noticed that, `<` is wrongly scoped as `keyword.operator.logical` while it should be `keyword.operator.relational`.

Fixes https://github.com/sublimehq/Packages/issues/2732

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -473,10 +473,9 @@ contexts:
   expression-end:
     - meta_prepend: true
     - include: ts-type-assertion
-    - match: '<<'
       scope: keyword.operator.bitwise.js
       push: expression-begin
-    - match: (?=<)
+    - match: (?=<(?![<=]))
       branch_point: ts-function-type-arguments
       branch:
         - ts-function-type-arguments
@@ -503,7 +502,7 @@ contexts:
         - ts-type-expression-begin
 
   ts-less-than:
-    - match: '<=?'
+    - match: '<'
       scope: keyword.operator.comparison.js
       set: expression-begin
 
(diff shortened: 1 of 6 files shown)


deathaxe:
JS is the only syntax using `relational` scope. Any chance to rename it to `keyword.operator.comparison` as this is the scope all other syntaxes use?

Thom1729 (approved):

Thom1729:
Good catch, thanks for the fix.

If other syntaxes use `comparison`, then so should JS/TS. Probably should be a separate PR, though.

FichteFoll (approved):

Thom1729:
I took another look and noticed that there's already special handling for the `<<` operator:

```yaml
  expression-end:
    - meta_prepend: true
    - include: ts-type-assertion
    - match: '<<'
      scope: keyword.operator.bitwise.js
      push: expression-begin
    - match: (?=<)
      branch_point: ts-function-type-arguments
      branch:
        - ts-function-type-arguments
        - ts-less-than
```

This is basically the same situation as the `<=` operator, so they should probably work the same. We could either add `<=` handling to `expression-end` or move `<<` handling to `ts-less-than`. (Also, `<<=` is missing, not that anyone would notice.)

But it might be simplest to just change the branch rule slightly:

```yaml
  expression-end:
    - meta_prepend: true
    - include: ts-type-assertion
    - match: (?=<(?![<=])
      branch_point: ts-function-type-arguments
      branch:
        - ts-function-type-arguments
        - ts-less-than
```

This should handle all four cases — `<` explicitly, and the other three because they'll never hit the branch in the first place.

Thom1729 (approved):

JavaScript/TypeScript.sublime-syntax
```diff
@@ -473,10 +473,9 @@ contexts:
   expression-end:
     - meta_prepend: true
     - include: ts-type-assertion
-    - match: '<<'
       scope: keyword.operator.bitwise.js
```
Thom1729, line 476: I think these two lines were left in by mistake.

deathaxe (approved):

PR #2736, opened 2021-03-05
[Python] Change the default executable to invoke to "python3"

rwols:
For the osx and linux platforms, we used to invoke the "python" executable. This
is for all practical purposes the same as invoking Python 2.7. Version 2 has
been EOL'd on January 1st, 2020.

### DIFF
--- a/Python/Python.sublime-build
+++ b/Python/Python.sublime-build
@@ -1,5 +1,5 @@
 {
-	"cmd": ["python", "-u", "$file"],
+	"cmd": ["python3", "-u", "$file"],
 	"file_regex": "^[ ]*File \"(...*?)\", line ([0-9]*)",
 	"selector": "source.python",
 
@@ -13,7 +13,7 @@
 	[
 		{
 			"name": "Syntax Check",
-			"shell_cmd": "python -m py_compile \"${file}\"",
+			"shell_cmd": "python3 -m py_compile \"${file}\"",
 		}
 	]
 }
--- a/Python/Python.sublime-build
+++ b/Python/Python.sublime-build
@@ -7,6 +7,13 @@
 
 	"windows": {
 		"cmd": ["py", "-u", "$file"],
+		"variants":
+		[
+			{
+				"name": "Syntax Check",
+				"cmd": ["py", "-m", "py_compile", "${file}"],
+			}
+		]
 	},
 
 	"variants":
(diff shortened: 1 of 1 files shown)


deathaxe:
Please add OS specific build systems, then. Windows doesn't know about `python3`!

jwortmann:
I thought the same first, but it's already there and calls `"cmd": ["py", "-u", "$file"]` on windows.

deathaxe:
But not the syntax check.

rwols:
Perhaps we can remove the syntax check? I see newcomers stumble over build variants from time to time. And more advanced users can modify the build or create their own anyway.

deathaxe:
Captions clearly declare those functions. Reading means winning.

That said, thanks for your tweaks. Had no luck using them though for 2 reasons.

1. it appears the platform specifier needs to be part of each variant.
2. mixing `cmd` and `shell_cmd` causes the latter one to win, which caused the windows command to be ignored (see: https://github.com/sublimehq/sublime_text/issues/3988)
   As `cmd` is used in general, I replaced `shell_cmd` to make it work.

I'd also like to use one style of variables if possible `$file` or `${file}`, but not both if not required.


```
{
    "cmd": ["python3", "-u", "$file"],
    "file_regex": "^[ ]*File \"(...*?)\", line ([0-9]*)",
    "selector": "source.python",

    "env": {"PYTHONIOENCODING": "utf-8"},

    "windows": {
        "cmd": ["py", "-u", "$file"],
    },

    "variants":
    [
        {
            "name": "Syntax Check",
            "cmd": ["python3", "-m", "py_compile", "$file"],

            "windows": {
                "cmd": ["py", "-m", "py_compile", "$file"],
            }
        }
    ]
}
```

rwols:
Alright, so this works for all platforms now I hope :)

deathaxe (approved):

michaelblyons (approved):

FichteFoll:
Should we rename the build system to "Python 3", then?

deathaxe:
That's the only python version, isn't it? Python 2 is dead and 4 not in range.

FichteFoll:
> That's the only python version, isn't it? Python 2 is dead and 4 not in range.

I wish it was, but being unable to call `python` kinda speaks otherwise. I'm fine with it either way, just wanted to hear some opinions.

PR #2737, opened 2021-03-05
[CSS] Sync properties with web

deathaxe:
This PR updates the lists of builtin property names and values using APIs of:
  https://www.chromestatus.com/metrics/css/popularity
  https://developer.mozilla.org/en-US/docs/Web/CSS

The data is used to update ...

1. syntax patterns
   
   Some garbage was found in `property_values`, such as function names which are otherwise matched already. Those have been removed.

2. css_completions
   
   Completions data is moved to a separate sub module as 2k4 lines of data seem just a bit too much for the single css_completions module.

Note:

It's a real sync, which means old property names, no longer present on mentioned web sizes is removed. As `caniuse.com` didn't return something useful for those which have been checked, it seems safe enough to do so. A second argument here is: If we keep all no longer valid legacy or maybe removed experimental property names, which have probably never been officially supported by all browsers, we can remove those exhaustive lists of builtins at all as "linting" is questionable then.


Alexey-T:
My notes at https://github.com/sublimehq/Packages/issues/2739

deathaxe:
Lists of predefined property names and values grows, syntax size explodes, parsing time doubles (with this PR) and outcome is more than questionable from highlighting perspective. Not to try to outline maintenance effort to keep those huge lists up-to-date while trying not to break old language constructs - some of them no longer documented at MDN nor anywhere else. There's just no sane way to distinguish obsolete from malformed constants. The lists IMHO contain a lot of wrong tokens but proofing that one by one is just too much.

It will always lack behind.

All the effort to update completions is worth nothing as it includes property names and some constants only. Many completions are still missing (function arguments, tag names, selectors, at-rules, ...). Effort size and required speed of development to keep up is mybe nothing for a builtin plugin. It's just too much.

PR #2752, opened 2021-03-13
[JavaScript] Expand statement tests and fix a related scoping bug.

Thom1729:
Improve and expand the syntax tests for statements (particularly if/else and do/while). Also, several control structure bodies were incorrectly specified as blocks rather than arbitrary statements, which resulted in some wrong meta scopes.


Thom1729:
Also, some of the tests that were moved out to `syntax_test_js_control.js` were still duplicated in the main syntax test file for some reason. I've removed them.

PR #2758, opened 2021-03-19
[Ruby] Add .pryrc support

halilim:
[Pry](https://github.com/pry/pry) configuration, similar to .irbrc


wbond:
Thanks for adding this!

PR #2760, opened 2021-03-21
[Common] Add or update emacs mode patterns

### LINKED ISSUE #2759: C (and other languages?) mode hint matching is super fragile
If you have a C file that doesn't have a C extension, sublime can still match it with an emacs-style mode hint.

It uses a regex here: https://github.com/sublimehq/Packages/blob/master/C%2B%2B/C.sublime-syntax#L8

`first_line_match: "-[*]-( Mode:)? C -[*]-"`

This is super fragile, and in fact is incorrect according to the [emacs docs](https://www.gnu.org/software/emacs/manual/html_node/emacs/Choosing-Modes.html)

`Mode` should be lower case, `C` could probably be lower case, and it's missing the `;`, plus extra stuff after.

This might work better: `(?i)-[*]- *(mode:)? *(c) *;?.*-[*]-`, though I don't know the flavour of regex used here
``` lang
// -*- mode: C -*-
#include <io.h>

int main() {
        printf("hello);
}
```

I don't use sublime, but use something that uses its syntax files (`bat`)
tmux

Now I don't know why you'd write C files that don't have a .c or .cc extension, I was just made aware of this when trying to figure out why another script wasn't highlighting.

deathaxe:
Fixes #2759

This PR...

1. sorts main header keys according to #2430
2. adds or update `first_line_match` patterns to all appropriate syntax definitions in order to support emacs mode tags.

   Examples:

   ```
   // -*-c-*-
   // -*- c -*-
   // -*- mode: c -*-
   // -*- mode: c++; -*-
   ```


ActionScript/ActionScript.sublime-syntax
```diff
@@ -2,9 +2,16 @@
...
 name: ActionScript
+scope: source.actionscript.2
+
 file_extensions:
   - as
-scope: source.actionscript.2
+
+first_line_match: |-
+  (?xi:
+    ^ \s* // \s* -\*- .* \baction-?script\b .* -\*-  # editorconfig
```
michaelblyons, line 12: I think this is the only one with `.*` instead of `.*?`. Intentional or no?

deathaxe: No.

ASP/ASP.sublime-syntax
```diff
@@ -2,10 +2,17 @@
...
+scope: source.asp
+
 file_extensions:
   - asa # asp is handled by HTML-ASP.sublime-syntax
   - vbs # Visual Basic Script
-scope: source.asp
+
+first_line_match: |-
+  (?xi:
+    ^ \s* \' \s* -\*- .*? \b(asp|vbs|vbscript)\b .*? -\*-  # editorconfig
```
keith-hall, line 13: I'm thinking asp should perhaps be in the HTML-ASP sublime-syntax file instead? :)

deathaxe:
I labeled this PR `significant` as it touches all syntaxes. Complexitiy of changes per file are rather low.

Batch File/Batch File.sublime-syntax
```diff
@@ -2,10 +2,18 @@
...
+
 file_extensions:
   - bat
   - cmd
-scope: source.dosbatch
+
+first_line_match: |-
+  (?xi:
+    ^ \s* (\:|rem) .*? -\*- .*? \bbat\b .*? -\*-  # editorconfig
+  | ^ \s* @echo\s+off\b
```
jfcherng, line 14: maybe also `@echo on`.

jfcherng (approved):

jrappen (approved):

jfcherng (approved):

PR #2767, opened 2021-04-04
[Ruby] Rename conditional operator scope

deathaxe:
This commit renames scopes of ternary operators `?` `:` for consistency with C, C#, D, Groovy, Java, JavaScript, PHP and ShellScript.


deathaxe:
We may also rename scopes of the other syntaxes to `keyword.operator.conditional`. Maybe this was an even more genaral approach?

michaelblyons:
You could change everyone to `keyword.operator.conditional.ternary`, so it defaults to `keyword.operator.conditional`, but more targeting is available if someone liked their old coloration (probably from `keyword.operator`).

michaelblyons (approved):

jwortmann:
I'd just like to throw in a controversial question here that might be useful to resolve beforehand:

What about Python's inline if-else statement?
```python
isnegative = True if x < 0 else False
```

Following the logic here and in the other recent PRs, `if` and `else` in the inline statement above should get `keyword.operator.conditional.ternary` as well, correct?

But it might be a bit odd to have different scopes (and potentially different colors) for this case, in comparison to when if-else is used as a block (`keyword.control.flow.conditional`):
```python
if x < 0:
    isnegative = True
else:
    isnegative = False
```

One might *possibly* argue that unlike `?` and `:`, Python's inline if-else keywords shouldn't really be described as "operators" (because they are no symbols?) and therefore could keep `keyword.control`, but then I would refer to e.g. the `keyword.operator.logical` scopes of `and`, `or`, etc. and to that the `keyword.operator.word` issue is still unresolved.

michaelblyons:
> What about Python's inline if-else statement?
> 
> ```python
> isnegative = True if x < 0 else False
> ```
> 
> Following the logic here and in the other recent PRs, `if` and `else` in the inline statement above should get `keyword.operator.conditional.ternary` as well, correct?

Yes. That is correct.


> But it might be a bit odd to have different scopes (and potentially different colors) for this case, in comparison to when if-else is used as a block (`keyword.control.flow.conditional`):

I disagree that this is odd. To the contrary, I appreciate it. See also: Python's `in` receives different scopes depending on its usage.


> &hellip; and to that the `keyword.operator.word` issue is still unresolved.

My thoughts on `keyword.operator.word` are already known. I want the scopes to reflect what a token _means_ rather than its typography.

jwortmann:
> I disagree that this is odd. To the contrary, I appreciate it.

That's why I tried to keep my wordings above quite vague, because I would assume that possibly there might be different opinions on that and maybe there is no real right or wrong here. It just could be unexpected for some users who are used to the old colors, when the inline if-else keywords in Python will change their highlighting now.

> I want the scopes to reflect what a token means rather than its typography.

This is absolutely fine and it is not the issue I was referring to. What I mean is that

1. There is currently no mechanism for color schemes to easily target these situations with a simple rule and without knowing and requiring a special exceptional rule for each individual language-token like this. This Python example here adds just *another* exception which color scheme authors have to know about and need an adjustment for, if they prefer to have a single color for all keywords and another color for all operator symbols.
2. There are several situations where it's outright impossible to have different colorization for word-like and symbol-like tokens, because both use the exact same scope.

Here is a real-world example for 1., because I believe I'm not the only one with a demand for a solution to this problem:
Imagine a VS Code user wants to try out Sublime, installs a nice [VS Code color scheme](https://packagecontrol.io/packages/Visual%20Studio%20Code%20Plus%20Scheme) from Package Control, which was created not too long ago, because he likes and is used to that scheme. And then they see that `if` and `else` in Python's inline if-else statement have no highlighting other than the default plain text color. Then they wonder whether Sublime's highlighting engine is worse, when in fact it's actually superior than VS Code's one (I think), or whether it is a bug in the Python syntax.

deathaxe:
> What about Python's inline if-else statement?

Not sure if python inline statements can be compared to ternary operators in other languages. Your suggestion would mean to also scope `for` etc. as operator when used inline.

```python.
a = [f for f in list if f = whatever]
```

I'd keep it untouched for both syntactical and sematical reasons.

> By the way, what's the purpose of the suggested conditional part at the 3. level?

Ruby was the only syntax using `conditional`, so the original PR intended to turn it into `ternary`. Basically, the former scope name is just a bit more general to be able to put several conditional like keywords into such a group.

You are right about _Batch File_ and basically `keyword.operator.logical` wouldn't be too wrong. But with a closer look the `&` calls the following command only, if the preceding one exits successfully. So it doesn't really express pure logical combination of output results but is more like a shorthand if...else... statement. I haven't double checked whether it also applies to ShellScript.

deathaxe:
I am not super confident about it matching all use cases well, but I'd vote to scope `in` as `keyword.operator.logical` instead of something like `keyword.control.loop` etc. because it just returns a boolean or object in case it was found in a data set. Yes there are basically two use cases `if a in [1, 2, 3]` which is really logical and `a for a in [1, 2, 3]` where it triggers the iterator interface, so the suggestion might not be perfect.

jwortmann:
> Not sure if python inline statements can be compared to ternary operators in other languages.

Hm, but it still means the same, right? What if Python 4 decides to replace the keywords in
```python
isnegative = "yes" if x < 0 else "no"
```
with actual symbols `?` and `:`. Would the same expression then magically become a ternary operator?
```
isnegative = "yes" ? x < 0 : "no"
```
So what is the defining property of an operator? Just that it appears as a symbol (apparently not, because we scope `and` as `keyword.operator` too)?

On the other hand the inline if-else is just syntactical sugar and I would say the keywords itself still have the same meaning as in the if-else block and therefore it would feel natural if they have the same scope. As you see I'm not really sure myself what would be the optimal approach here. But just from the highlighting perspective I thing I would prefer to keep the `keyword.control` instead of `keyword.operator`.

***

> But with a closer look the & calls the following command only, if the preceding one exits successfully.

This is exactly the short-circuit evaluation with logical operators used in other languages too, that I mean. For example https://github.com/JuliaLang/julia/blob/4996445df37e526dac2772e333caf82f1ea987f0/base/abstractarraymath.jl#L73-L76
and I believe this is often used in C-like languages as well.

***

+1 to that Python's `in` should have the same scope even when used in slightly different occasions. Or at least a specialization of the same base scope, for example `keyword.operator.logical.loop` for when it's used in a for loop.

deathaxe:
> So what is the defining property of an operator? 

I guess it wouldn't be too wrong to scope ternaries as `keyword.control.conditional`. Just haven't thought about it as they have been scoped as operators for ages. I mean, the lambda operator `->` is scoped `keyword.declaration.function.anonymous` as well.

> This is exactly the short-circuit evaluation 

You are actually right with that. Same as python's `a = fun() or ''` stuff.

So `keyword.operator.logical` would suit for batch file, too, then. Thank's for pointing me to the correct direction.

It seems to apply to `&&` and `||`. Single `&` seems just like a statement terminator, from what I realize from a short command evaluation.

Thom1729:
Just chiming in to say that in statements like `for x in y`, I don't think that `in` is really an operator. I think it makes sense to scope the `in` operator differently from the `in` keyword in loops.

deathaxe:
Suggestions?

Thom1729:
In `for … in`, I think that `keyword.control.loop` makes sense. Or generally, a similar scope to `for` in the same situation. As I see it, they're both keywords defining the loop construct. I'm not wedded to any particular scope except that I think they should be distinct from operators.

jwortmann:
Stupid question; what about `if … in`? Is `in` an operator in that case, or should it get a similar scope to `if` (`keyword.control.conditional`) in this situation. I recognize that `for ... in` in Python is special, because for-loops always have the `in`, so it kind of belongs to the loop. But both describe pretty much the same thing on the semantic level, even though one might return an iterator and another does not.

And it's not that easy in some other languages. Here are some Julia examples that all do the same:
```julia
for i = 1:3
    println(i)
end

for i ∈ 1:3
    println(i)
end

for i in 1:3
    println(i)
end
```
Which of those are operators and which are not? Or which of those should get a similar scope to the `for` keyword here?

rwols:
@jwortmann

> Imagine a VS Code user wants to try out Sublime, installs a nice VS Code color scheme from Package Control, which was created not too long ago, because he likes and is used to that scheme.

Well, there's your problem. **VSCode scope space != Sublime scope space**.

Like I said in the pull request for that package: https://github.com/wbond/package_control_channel/pull/8063#issuecomment-707341398

Someone should write a VSCode space to Sublime space color converter. As it stands, the color scheme as published on package control is in the wrong space.

jwortmann:
> Well, there's your problem. VSCode scope space != Sublime scope space.

I don't think it's my problem (yes I know you didn't mean it literally), ultimately it will end up as Sublime's problem, when the user who doesn't know anything about scopes just sees that highlighting for inline if-else works in VS Code, but not in Sublime. And "VSCode scope space != Sublime scope space" isn't the reason, the reason rather is that even if the color scheme author wanted, he simply could not specify a general rule in the color scheme for operator symbols that will work with adjusted or newly introduced scopes like potentially this one here for Python. Because there is no such concept to target these things in the Sublime scope space.

We would need for example something like a previously suggested `meta.word` stacked on top of keywords that fall under `keyword.operator`, and `meta.symbol` or so on top of the scope for tokens like `->` which have been decided to use `keyword.declaration` instead of `keyword.operator`. As it is now, color scheme authors need to closely follow the changes in this repo and add small fixes into their scheme whenever scopes like this example here get introduced or adjusted. 

Lots of older schemes are built to work for multiple editors at the same time, and I don't think something like a VSCode space to Sublime space converter is gonna be happen.

Thom1729:
> Stupid question; what about `if … in`? Is `in` an operator in that case, or should it get a similar scope to `if` (`keyword.control.conditional`) in this situation. 

Are you talking about something like `if x in y`? In that case, `in` is behaving as an ordinary binary operator.

> And it's not that easy in some other languages. Here are some Julia examples that all do the same:
>
> …
>
> Which of those are operators and which are not? Or which of those should get a similar scope to the `for` keyword here?

In principle, I would say that none of those are operators; they're all special binding syntax that's part of the `for` loop header. However, we also have an established convention of scoping equals signs used for binding as `keyword.operator.assignment`, even when they're not “really” operators. (For instance, in Python, `=` is part of an assignment statement and you have to use the “walrus operator” `:=` in expressions.) So in this special case, I think it would be most consistent to scope the equals sign as `keyword.operator.assignment`.

It may be useful to note that the behavior of the equals sign in `for x = 1:3` is fundamentally similar to the behavior of the equals sign in `x = 1`. In both cases, it's binding the name on the left to a value on the right. Obviously the two cases are not identical, but they do share a basic similarity. On the other hand, the `in` token in `for x in y` and the expression `x in y` are very different because they treat the token `x` very differently. So in those Julia examples, I would scope `in` as part of the `for` loop, even though I would probably scope the equals sign `keyword.operator.assignment`.

For similar reasons, the set membership symbol should probably be treated like `in`. I don't have strong feelings on this because it seems like an unusual case that's unlikely to come up very often.

jwortmann:
I can follow and agree to that argumentation :thumbsup:.

The following code would be the same as the if-example then.
```julia
i = 1
while i in 1:3
    println(i)
    i += 1
end
```

But with our convention with `keyword.operator.assignment` for the equal sign, wouldn't it make sense to use the same scope for the `in` token in `for x in y` as well? The `in` also acts as an assignment operation here, so it should be logical to represent this with the same scope name. This would also
* avoid having different scopes for tokens that mean exactly the same (in my three for-loop examples above, which are fully equivalent).
* ensure that both usages of `in` share the same first 2 scope levels (`keyword.operator.logical` and `keyword.operator.assignment`) and therefore prevent possible confusions when different `keyword` scopes are colored differently by default in a color scheme. But if a user wants, they can still target them individually and assign different colors

In that case I would suggest a further specialization of the assignment scope to distinguish it from the equal sign, or even better a decision for the `meta.word` suggestion :)

> For similar reasons, the set membership symbol should probably be treated like `in`.

Yes, that seems very correct (even though I personally dislike it as long as we have no sane way to distinguish between them). In fact the `∈` is just an alias for `in` (see single-line [implementation](https://github.com/JuliaLang/julia/blob/54c7002892d1b0260890848397e4ba1e2d2506d1/base/operators.jl#L1243) and [docs](https://docs.julialang.org/en/v1/base/collections/#Base.in) - used as a function in that reference but my for-loop examples are analog).

deathaxe:
> ... the `∈` is just an alias for `in` ...

This is what we find in Haskell, too, very often. Symbols being used as synonyms for word operators.

That said: Diskussion somehow moved far away from the original intend of this PR. Looks more like an RFC now.

jwortmann:
I would suggest to
> a) closing all the others and revert to `keyword.operator.ternary`

unless we find another concrete example where `keyword.operator.conditional.*` would be useful. "ternary" seems like a more descriptive and common name for `?` and `:`, so I guess it's preferable to "conditional" or to a possible duplication of "conditional.ternary".

For Python I'd vote to keep it as it is for now, at least until we have a better way to target an ajdusted scope in color schemes.

keith-hall (dismissed):

jfcherng (dismissed):

keith-hall (approved):

jfcherng (approved):

PR #2769, opened 2021-04-05
[R] Fix scopes for comparison operators

jwortmann:
Here's another tweak for operator scopes that I recently found.

Any R user here who knows what `<>` is used for and whether its scope name is correct, because I couldn't find it in the R documentation?


deathaxe (approved):

wbond:
Thanks for these fixes!

PR #2775, opened 2021-04-08
[JavaScript] Refine ternary conditional sub scope

deathaxe:
Refines `keyword.operator.ternary` to `keyword.operator.conditional.ternary` as some syntaxes may require other types of conditional keywords.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1259,14 +1259,14 @@ contexts:
 
   ternary-operator:
     - match: '\?(?=[^.]|\.[0-9])'
-      scope: keyword.operator.ternary.js
+      scope: keyword.operator.conditional.ternary.js
       set:
         - ternary-operator-expect-colon
         - expression-no-comma
 
   ternary-operator-expect-colon:
     - match: ':'
-      scope: keyword.operator.ternary.js
+      scope: keyword.operator.conditional.ternary.js
       set: expression-no-comma
     - include: else-pop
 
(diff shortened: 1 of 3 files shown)


deathaxe:
Hmm. Who said, running syntax tests by hitting `ctrl+b` in a sublime-syntax files runs tests against all found syntax_test_... files? Seems not to work here.

Test runner says everything is ok, even so it isn't.

grafik

Thom1729 (approved):

Thom1729:
> Hmm. Who said, running syntax tests by hitting `ctrl+b` in a sublime-syntax files runs tests against all found syntax_test_... files? Seems not to work here.
> 
> Test runner says everything is ok, even so it isn't.
> 
> grafik

That's really weird; it works reliably for me.

michaelblyons:
I do my editing in a different worktree than the one used by Sublime itself. If I run syntax tests on my worktree version, it seems to do something like DeathAxe. Not sure about running them in the symlinked version.

Thom1729:
I have the repo checked out somewhere random, but the JavaScript directory symlinked under Packages, and I edit via the symlinked path.

michaelblyons:
I also have a symlink for ST to load the definitions, but I do not edit there. I edit in a sibling directory to the symlink target and _deploy_ to the symlinked directory.

I have a linter going in my Edit project that compares its files against the currently deployed syndefs. That seems to work fine, since the first-line "This is a test for X" info are ST resources, not file system paths. But if I run the "Run tests" build against the syndef in Edit, it knows how many files are linked against it (or rather against its deployed equivalent?), but it says all their assertions pass.

I don't remember quite why I set this up. It had something to do with integrating long-running branches into my main work environment.

deathaxe:
The main _Packages_ repo is checked out in a data directory. For each syntax I work on, I've created a separate worktree with a sparse-checkout next to the root repository. The worktree folders are symlinked (junction) into ST's packages path, where I do all the editing work.

Issue #1301, opened 2017-11-10
[Batch] parentheses are illegal.invalid in comments for no apparent reason


rwols:
<img width="983" alt="schermafbeelding 2017-11-10 om 22 24 41" src="https://user-images.githubusercontent.com/2431823/32679288-305adc3c-c666-11e7-9a24-636c66cb1d2a.png">

```batch
REM foo (the fooest of foos).
REM     ^ - invalid.illegal
```

However, there's no `invalid.illegal` scope when using double colons:

```batch
:: foo (the fooest of foos).
::     ^ - invalid.illegal
```
[from here](https://github.com/llvm-mirror/llvm/blob/master/utils/release/build_llvm_package.bat).

I'm not an expert at batch script, so this could be the correct behavior.

FichteFoll:
Parenthesis are special, but it doesn't seem like they should affect the REM command in this instance, according to https://ss64.com/nt/syntax-brackets.html.

keith-hall:
It looks like these were added in https://github.com/sublimehq/Packages/pull/350.

I've just done a google search for `dos batch file rem pipe` and found https://technet.microsoft.com/en-us/library/bb490986.aspx?f=255&MSPPError=-2147217396:

> Restrictions on batch file comments:
> You cannot use a redirection character "(" or ")" or pipe (|) in a batch file comment.

Although in my testing, I've found no problems with doing so.

user-9826342:
@keith-hall 

> Although in my testing, I've found no problems with doing so.

What microsoft article is telling you is that if parentheses appear in a comment (either `rem` or silenced `echo`) they *may* be treated as opening or closing parenthese of the outer block instead of part of the comment. For example

```batch
(
rem ()
```

is the same as

```batch
(
rem (
)
```

You can also see it if you replace `rem` with `echo`. But anyway, this is valid batch syntax. Another example is `rem ^(^)` which is valid even if you surround it with parentheses.

thethinker990:
+1

wbond:
@thethinker990 These are open source packages - be the change you want to see!

thethinker990:
Where do I find the syntax settings? I would give it a try.

jfcherng:
> Where do I find the syntax settings? I would give it a try.

https://github.com/sublimehq/Packages/blob/3e03b6f08c68c867f69ac141137b0cbf1d65db4d/Batch%20File/Batch%20File.sublime-syntax#L399-L400

[...]

PR #2780, opened 2021-04-09
[BATCH] Bugfixing illegal comment syntax

thethinker990:
https://github.com/sublimehq/Packages/issues/1301

```
'[(|)]' should create a illegal comment statements, but it don't
https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb490986(v=technet.10)?redirectedfrom=MSDN
```

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -391,7 +391,9 @@ contexts:
         - match: \n
           scope: comment.line.rem.dosbatch
           pop: true
-        - match: '[(|)]'
+          # '[(|)]' should create a illegal comment statements, but don't
+          # https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb490986(v=technet.10)?redirectedfrom=MSDN
+        - match: ''
           scope: invalid.illegal.unexpected-character.dosbatch
 
   strings:
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -391,6 +391,10 @@ contexts:
         - match: \n
           scope: comment.line.rem.dosbatch
           pop: true
+        # '[(|)]' should create a illegal comment statements, but don't
+        # https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb490986(v=technet.10)?redirectedfrom=MSDN
+        - match: ''
+          scope: invalid.illegal.unexpected-character.dosbatch
 
   strings:
     - match: '"'
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -391,10 +391,6 @@ contexts:
         - match: \n
           scope: comment.line.rem.dosbatch
           pop: true
-          # '[(|)]' should create a illegal comment statements, but don't
-          # https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb490986(v=technet.10)?redirectedfrom=MSDN
-        - match: ''
-          scope: invalid.illegal.unexpected-character.dosbatch
 
   strings:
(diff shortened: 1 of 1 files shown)


Batch File/Batch File.sublime-syntax
```diff
@@ -391,7 +391,9 @@ contexts:
         - match: \n
           scope: comment.line.rem.dosbatch
           pop: true
-        - match: '[(|)]'
+          # '[(|)]' should create a illegal comment statements, but don't
+          # https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb490986(v=technet.10)?redirectedfrom=MSDN
+        - match: ''
           scope: invalid.illegal.unexpected-character.dosbatch
```
rchl, line 397: This is a no-op now and can just be removed.

Batch File/Batch File.sublime-syntax
```diff
@@ -391,7 +391,9 @@ contexts:
         - match: \n
           scope: comment.line.rem.dosbatch
           pop: true
-        - match: '[(|)]'
+          # '[(|)]' should create a illegal comment statements, but don't
+          # https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb490986(v=technet.10)?redirectedfrom=MSDN
```
rchl, line 395: The comment is indented so that it's inside the `\n` match. I think it should have one less indent level.

Also, I'm not sure it's that clear what it's trying to convey, especially once the original match rule is gone.

Batch File/Batch File.sublime-syntax
thethinker990: Thank you, this is my first pull request.

rchl:
The patch should be against the master branch since st3 is no longer updated.

Batch File/Batch File.sublime-syntax
thethinker990: I did remove the `match` and `scope`statement but now the syntax check fails. Should I remove the whole comment pattern? (from [383](https://github.com/sublimehq/Packages/blob/6674253653c99de02cfac68921c9a1fb3d0780d5/Batch%20File/Batch%20File.sublime-syntax#L383))

rchl: The syntax tests need to be updated since they currently expect that the ( and ) are scoped as illegal.

BTW. I'm just pointing out the obvious issues with these changes. I don't have an opinion on whether the change itself is sane and wanted. As I understand from the discussion, the ( and ) can be illegal in some situation but not the other so I guess the choice is to either never scope them as illegal as done here, or add a lot of handling that scopes it as illegal only when it's actually illegal.

deathaxe: Handling parentheses is a separate issue as it effects nearly any statement in Batch File. Scoping parens illegal in comments perse is wrong.

_Already addressed it in a local rewrite of Batch File, but got stuck at exactly this parentheses thing. It behaves so strange, that we'd probably need to write batch file twice - one version for top-level and one for parentheses. Haven't found a sane solution to do so without duplicating all the contexts. At the time work started, ST didn't support inheritance nor branching._

michaelblyons: I think we can all agree that Batch is a terrible language. 😅

Batch File/Batch File.sublime-syntax
deathaxe: All those lines from 394 to 397 can be removed. An empty match is not what we might want to have.

Batch File/Batch File.sublime-syntax
deathaxe: Same category as Perl and Bash.

deathaxe:
The change of this PR is included in PR #2794. Thus closing this one as obsolete/superseded.

Issue #2772, opened 2021-04-08
[HTML] How many html tag types do we need?


deathaxe:
I am about to update HTML.sublime-syntax by adopting the `html_tags` variable introduced in CSS some time ago to enable HTML to highlight all valid html tags as such.

### Current state

Current HTML.sublime-syntax distinguishes between following tag types at the moment:

- `meta.tag.structure.any` (e.g.: html, head, body)
- `meta.tag.block.any` (e.g.: h1, h2, p, ...)
- `meta.tag.block.form` (form, fieldset)
- `meta.tag.inline.any`
- `meta.tag.inline.form` (e.g.: option, label, textarea, ...)
- `meta.tag.inline.table` (e.g.: table, tr, td, ...)
- `meta.tag.script` (script tags)
- `meta.tag.style` (style tags)

Those scopes exist since day 1 in 2015 and I wonder wether there is a deeper sense behind them, which makes it worth maintaining those scopes.

### Step 1: Fixing inconsistencies

I personally find 

1. all kinds of `table` tags not to be inline but rather block tags assuming we want to keep this distinction.
2. Also `label` or `textarea` feel more like block tags than inline form tags.

Those are inconsistencies I'd like to fix in a step 1.

### Step 2: Restructuring tag/scope groups?

What kind of grouping do we want to use?

a) maintain current groups: `structure`, `block` and `inline`? 
b) more detailed groups using scopes matching the tag type headers of https://developer.mozilla.org/en-US/docs/Web/HTML/Element
   Means: `meta.tag.section`, `meta.tag.text`, `meta.tag.embedding`, `meta.tag.media`, etc.
c) abandon all those different kinds of scopes and just call them `meta.tag.html`?

### Step 3: Illegal self-closing tags

Firefox highlights illegal self-closing tags.

![grafik](https://user-images.githubusercontent.com/16542113/114065508-05d30380-989b-11eb-8ee6-c9eb268c8a5a.png)

Do we want to apply something like that, too? I'd suggest to only highlight `/` illegal instead of the whole tags though.

PR #2781, opened 2021-04-09
[HTML] Extend and reorganize known html tags

deathaxe:
This PR illustrates the ideas of #2772.

1. Missing builtin html tags are added by reusing html tag name variables from CSS.sublime-syntax. The `html_..._tags` variables variables are used to categorize tags according to MDN documentation.

2. Meta scope scheme of tags is changed to `meta.tag.[html|sgml|...|other].[opening|closing|selfclosing]`.
   
   The motivation behind it is, that `meta.tag.[block|inline|structure|...]` don't provide enough details for useful code intelligence features or something like that, while the information about opening or closing tags in general might help in some situations.

   The 3rd level scope `[html|sgml|...|other]` is meant to distinguish between main types of tags.
   
   Not sure if it should be mixed with opening/closing though.

3. New tag scope scheme allows illegal highlighting of self closing tags in those which don't support it (e.g.: `<html/>`).

4. Tag names use `entity.name.tag.[block|form|inline|structure|table]` to provide some possibilities to assign different teg name colors based on category. Current implementation makes it more easy to modify this scheme, if needed.

5. Script/Style tags are modified to match the new scope scheme accordingly.

#### Notes:

1. Some test files needed to be tweaked outside of HTML package
2. The changes don't impact performance.

### DIFF
--- a/ASP/syntax_test_asp.asp
+++ b/ASP/syntax_test_asp.asp
@@ -1,9 +1,9 @@
 ' SYNTAX TEST "Packages/ASP/HTML (ASP).sublime-syntax"
 <!DOCTYPE html>
 <html>
-'<- meta.tag.structure.any.html punctuation.definition.tag.begin.html - source.asp.embedded.html
-'^^^^ meta.tag.structure.any.html entity.name.tag.structure.any.html
-'    ^ meta.tag.structure.any.html punctuation.definition.tag.end.html
+'<- meta.tag.html.opening.html punctuation.definition.tag.begin.html - source.asp.embedded.html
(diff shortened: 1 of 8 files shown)


deathaxe:
I'll fix syntax tests, once we came up with an answer about meta scopes.

facelessuser:
@michaelblyons  Not sure this will directly affect BH, but thanks for the heads up. Good to be aware of these types of things.

deathaxe:
Changes of this PR have high potential to break 3rd-party packages.

jrappen:
This is useful.

Is there a known list of packages this would break? Maybe ping their maintainers and move ahead with this?

PR #2782, opened 2021-04-10
[Markdown] Rewrite MultiMarkdown

deathaxe:
This commit rewrites the little MultiMarkdown.sublime-syntax by extending Markdown. 

1. Reduces the number of lookaheads used to find the end of meta data value contexts. Only the first header key needs a lookahead now.
2. Adds named contexts for everything

### DIFF
--- a/Markdown/MultiMarkdown.sublime-syntax
+++ b/Markdown/MultiMarkdown.sublime-syntax
@@ -1,29 +1,37 @@
 %YAML 1.2
 ---
-# http://www.sublimetext.com/docs/3/syntax.html
-version: 2
 name: MultiMarkdown
-first_line_match: (?i)^format:\s*complete\s*$
 scope: text.html.markdown.multimarkdown
+
+extends: Packages/Markdown/Markdown.sublime-syntax
+
+first_line_match: (?i:^format:\s*complete\s*$)
+
 variables:
-  header: ((?=[A-Za-z0-9])[\w -]+)(:)
+  header: ([A-Za-z0-9][\w -]*)(:)
+
 contexts:
   main:
-    - match: '^{{header}}\s*'
+    - match: ^(?={{header}})
+      push: multimarkdown-header
+    - match: ^
+      set: multimarkdown-content
+
+  multimarkdown-header:
+    - match: ^$
+      pop: 1
+    - match: ^(?:{{header}}\s*)?
       captures:
         1: keyword.other.multimarkdown
         2: punctuation.separator.key-value.multimarkdown
-      push:
-        - meta_scope: meta.header.multimarkdown
-        - match: '^$|^(?={{header}})'
-          pop: true
-        - match: .+
-          comment: |
-            The reason for not setting scopeName = "string.unquoted"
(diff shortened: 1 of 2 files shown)


deathaxe:
Note: If this PR is merged earlier than #2783, MultiMarkdown.sublime-syntax needs to be tweaked in the other PR by re-adding `version: 2` marker.

Maybe it's better to merge #2783 first and tweak this PR afterwards.

jrappen:
> ... and tweak this PR afterwards ...

@deathaxe could you update this to syntax v2?

deathaxe:
No, because this syntax extends Markdown.sublime-syntax which isn't v2.

Markdown/syntax_test_multimarkdown.md
```diff
@@ -29,7 +29,8 @@ HTML Header: <style>
              </style>
 T:           ^^^^^^^^ meta.header.multimarkdown string.unquoted.multimarkdown
 
-T: <- meta.content.multimarkdown - meta.header.multimarkdown
+T:
```
jrappen, line 32: FYI, there's a trailing space here.

Markdown/MultiMarkdown.sublime-syntax
```diff
@@ -1,29 +1,37 @@
...
   main:
-    - match: '^{{header}}\s*'
+    - match: ^(?={{header}})
+      push: multimarkdown-header
+    - match: ^
+      set: multimarkdown-content
+
+  multimarkdown-header:
+    - match: ^$
+      pop: 1
```
jrappen, line 22: Since this is v1, it should probably be `pop: true`.

Markdown/MultiMarkdown.sublime-syntax
```diff
@@ -1,29 +1,37 @@
...
-      push:
-        - - meta_scope: meta.content.multimarkdown
-        - scope:text.html.markdown
+      push: multimarkdown-header-value
+
+  multimarkdown-header-value:
+    - meta_scope: meta.header.multimarkdown
+    - meta_content_scope: string.unquoted.multimarkdown
+    - match: \n
+      pop: 1
```
jrappen, line 33: Since this is v1, it should probably be `pop: true`.

jrappen:
Otherwise this seems fine.

Markdown/MultiMarkdown.sublime-syntax
deathaxe: We decided to move ont to `pop: 1` not after switching to `version 2´ syntaxes and to use the same style in the whole file. I don't see any reason why we should use `pop: true` here.

Markdown/syntax_test_multimarkdown.md
deathaxe: Good catch, while the bigger issue is `T:` to be the correct test case indicator - not `|` as assumed in the last tests.

deathaxe:
Appears I can't push anything to this branch anymore due to some casing erros with branch names.

deathaxe:
I merged this change into #3167

PR #2784, opened 2021-04-11
[JSON] Update to sublime syntax version 2

deathaxe:
This PR proposes to update JSON to sublime-syntax version 2.

It contains some cosmetic changes only, as none of the existing patterns was incompatible.


JSON/JSON.sublime-syntax
```diff
@@ -110,28 +121,31 @@ contexts:
...
     - meta_scope: string.quoted.double.json
     - meta_include_prototype: false
-    - match: '"'
+    - match: \"
       scope: punctuation.definition.string.end.json
-      pop: true
+      pop: 1
     - include: string-escape
-    - match: $\n?
+    - match: \n
```
jfcherng, line 145: I guess these two behave different at EOF but probably doesn't matter?

deathaxe: 1. We don't need to pop a string context at eof, because there is nothing which would follow.
2. Strings are leaf-contexts. They don't include itself, such as groups can. So consuming the `\n` is safe (not causing any recursive context not to pop. _All other contexts which include strings would need to make sure to match `$`._

These are the two major points I use to choose whether to `$\n?` or `\n`. As both conditions are met the latter one is safe to be used.

jfcherng (approved):

Thom1729 (approved):

keith-hall (approved):

PR #2785, opened 2021-04-11
[JSON] More detailed constants scopes

deathaxe:


Ultra-Instinct-05:
Are there any real benefits of scoping `true` & `false` specifically up to using a 4th level scope (true vs false) ? I feel most users would never want different highlighting for `true` & `false`, so having `constant.language.boolean.json` would IMHO be sufficient for both. They are anyways scoped differently than `null` (if they wanted `null` to be of a different color).

deathaxe:
Core syntax are different on that, yes. I just followed JavaScript here. It doesn't hurt neigher in performance nor implementation effort.

keith-hall (approved):

jfcherng (approved):

PR #2787, opened 2021-04-12
[JSX] Improve commenting behavior.

### LINKED ISSUE #2461: [JSX] Comment not working in HTML content

Thom1729:
Should fix #2461.

Given the following:

```js
<foo>
    bar!
</foo>;
```

The `toggle_comment` command will comment the second line as follows:

```js
<foo>
    {/*bar!*/}
</foo>;
```

But when you call it again to uncomment, you get:

```js
<foo>
    {bar!}
</foo>;
```

…instead of the original, which we want. This is due to internal implementation details of `toggle_comment` — it won't remove `{/*`/`*/}`, even though they're specified in `JSX Comments.tmPreferences`, because the comment scope doesn't extend to the braces.

This PR adds a special case to JSX interpolations so that when an interpolation contains only a single block comment, the entire interpolation including the braces will get the `comment` scope, so that the `toggle_comment` command will do what we want. This is not really the correct scoping, but the previous behavior is really annoying and we can't really change `toggle_comment`.

In addition, this PR improves the behavior of commenting attributes. E.g.:

```js
<Foo
  bar={42}
/>
```

…will become:

```js
<Foo
  // bar={42}
/>
```

### DIFF
--- a/JavaScript/Comments.tmPreferences
+++ b/JavaScript/Comments.tmPreferences
@@ -2,7 +2,7 @@
 <plist version="1.0">
 <dict>
 	<key>scope</key>
-	<string>source.js, source.jsx, source.ts, source.tsx</string>
+	<string>source.js, source.jsx, source.ts, source.tsx, meta.jsx meta.tag.name, meta.jsx meta.tag.attributes</string>
 	<key>settings</key>
 	<dict>
 		<key>shellVariables</key>
--- a/JavaScript/JSX Comments.tmPreferences
+++ b/JavaScript/JSX Comments.tmPreferences
@@ -5,7 +5,7 @@
     <key>name</key>
     <string>JSX Comments</string>
     <key>scope</key>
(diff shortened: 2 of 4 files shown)


Thom1729:
I've simplified the comments file substantially, but they should behave the same.

Note that there are still some broken corner cases due to https://github.com/sublimehq/sublime_text/issues/2152.

PR #2789, opened 2021-04-12
[PHP] Extend HTML, JS and CSS syntaxes like ASP

### LINKED ISSUE #643: Bad PHP syntax highlighting inside a HTML attribute's double quotes
When using `<?= ... ?>` and `<?php ... ?>` inside a HTML attribute's double quotes `"..."`, the syntax doesn't highlight correctly. The PHP code does get parsed however, and thus should highlight using PHP syntax, and not according to HTML syntax.

```
<?= $isImportant ? "important" : "notsoimportant" ?>
<div class="myclass <?= $isImportant ? "important" : "notsoimportant" ?>"></div>
```

In the following image, the two `<?= ... ?>` should highlight the syntax equally, as they get both parsed as PHP.

image

### LINKED ISSUE #2687: [PHP] Should we modify base scrope from embedding.php to text.html.php?
PHP is the only core syntax which uses `embedding` top-level scope for the joined HTML/PHP syntax file.

The only other 3rd-party package found doing so is _ColdFusion_.

All other templating syntaxes, which extend HTML use something like `text.html.<name>`.

ASP and Rails already follow that scheme.

Should we modify PHP to do so as well?

FichteFoll:
Also changes the base scope to `text.html.php`.

Updates #2687 (does not touch the "HTML (C#)" syntax).
Fixes #643.

---

I would like to rebase this on #2831.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -164,7 +164,7 @@ contexts:
         - include: script-javascript-content
 
   script-javascript-content:
-    - match: (?=\S)
+    - match: (?=\S|^)
       embed: scope:source.js
       embed_scope: source.js.embedded.html
       escape: '{{script_close_lookahead}}'
@@ -260,7 +260,7 @@ contexts:
         - include: style-css-content
 
   style-css-content:
-    - match: ''
+    - match: (?=\S|^)
       embed: scope:source.css
       embed_scope: source.css.embedded.html
       escape: '{{style_close_lookahead}}'
--- a/PHP/CSS (for PHP).sublime-syntax
+++ b/PHP/CSS (for PHP).sublime-syntax
@@ -12,6 +12,30 @@ contexts:
     - meta_prepend: true
     - include: php-embedded
 
+  string-content:
+    - meta_prepend: true
+    - include: php-interpolations
+
+  php-interpolations:
+    - match: <\?(?i:php|=)?(?![^?]*\?>)
+      scope: punctuation.section.embedded.begin.php
+      push: php-interpolation-block
+    - match: <\?(?i:php|=)?
+      scope: punctuation.section.embedded.begin.php
+      push: php-interpolation-line
+
+  php-interpolation-block:
+    - clear_scopes: 1
+    - meta_scope: meta.interpolation.css meta.embedded.block.php
+    - meta_content_scope: source.php.embedded.css
+    - include: php-embedded-content
+
+  php-interpolation-line:
+    - clear_scopes: 1
+    - meta_scope: meta.interpolation.css meta.embedded.line.php
+    - meta_content_scope: source.php.embedded.css
+    - include: php-embedded-content
+
   php-embedded:
     - match: <\?(?i:php|=)?(?![^?]*\?>)
       scope: punctuation.section.embedded.begin.php
--- a/PHP/CSS (for PHP).sublime-syntax
+++ b/PHP/CSS (for PHP).sublime-syntax
@@ -1,38 +1,38 @@
 %YAML 1.2
 ---
 name: CSS (for PHP)
-hidden: true
 scope: source.css.php
 version: 2
+hidden: true
 
 extends: Packages/CSS/CSS.sublime-syntax
 
 contexts:
   prototype:
     - meta_prepend: true
-    - include: php-tag-start
+    - include: php-embedded
 
-  php-tag-start:
+  php-embedded:
     - match: <\?(?i:php|=)?(?![^?]*\?>)
       scope: punctuation.section.embedded.begin.php
-      push:
-        - meta_scope: meta.embedded.block.php
-        - meta_content_scope: source.php.embedded.css
-        - include: php-end-tag-pop
-        - include: scope:source.php
-          apply_prototype: true
+      push: php-embedded-block
     - match: <\?(?i:php|=)?
       scope: punctuation.section.embedded.begin.php
-      push:
-        - meta_scope: meta.embedded.line.php
-        - meta_content_scope: source.php.embedded.css
-        - include: php-end-tag-pop
-        - include: scope:source.php
-          apply_prototype: true
+      push: php-embedded-line
 
-  php-end-tag-pop:
-    - match: (\?>)(\s*\n)?
-      captures:
-        1: punctuation.section.embedded.end.php
-        2: meta.css-newline-after-php.php
+  php-embedded-block:
+    - meta_scope: meta.embedded.block.php
+    - meta_content_scope: source.php.embedded.css
+    - include: php-embedded-content
+
+  php-embedded-line:
+    - meta_scope: meta.embedded.line.php
+    - meta_content_scope: source.php.embedded.css
+    - include: php-embedded-content
+
+  php-embedded-content:
+    - match: \?>
+      scope: punctuation.section.embedded.end.php
       pop: 1
+    - include: scope:source.php
+      apply_prototype: true
--- a/PHP/CSS (for PHP).sublime-syntax
+++ b/PHP/CSS (for PHP).sublime-syntax
@@ -1,6 +1,6 @@
 %YAML 1.2
 ---
-name: PHP CSS
+name: CSS (for PHP)
 hidden: true
 scope: source.css.php
 version: 2
--- a/PHP/CSS (for PHP).sublime-syntax
+++ b/PHP/CSS (for PHP).sublime-syntax
@@ -31,8 +31,4 @@ contexts:
     - include: php-embedded-content
 
   php-embedded-content:
-    - match: \?>
-      scope: punctuation.section.embedded.end.php
-      pop: 1
-    - include: scope:source.php
-      apply_prototype: true
+    - include: PHP.sublime-syntax#php-embedded-content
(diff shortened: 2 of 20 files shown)


rchl:
> Also changes the base scope to `text.html.php`.

Sounds scary, no?

LSP in ST4 mostly doesn't care about base scopes anymore but I would still envision some minor issues with that. For example [LSP-intelephense](https://github.com/sublimelsp/LSP-intelephense/blob/07265abef25e849256f768f1ac91e40da88ab85b/LSP-intelephense.sublime-settings#L11-L11).

Also probably there are packages out there that have keybinding targeting the `embedding.php` context?

keith-hall:
> Sounds scary, no?

If we're gonna fix it, now is the right time ;)

rchl:
I agree but only if there is still some time for testing this change before the official, final release.

FichteFoll:
The base scope name change was actually what I did first. Only then I noticed that the `with_prototype` was kinda off-putting and ended up not being able to split the commits conveniently anymore.

deathaxe:
We might just find/use file proper naming scheme to avoid confusion in the long term.

What we have atm is.:

1. `<base syntax> (<embedded>)` (e.g.: `HTML (ASP)`, `HTML (Jinja2)`, ...)
2. Normal syntax names may contain whitespace so it may not be clear in any case, whether a syntax name has a special (extending) meaning, if we just extend them by `<target syntax> <extended syntax>` (e.g.: `Makefile Shell` or `PHP CSS`).

We therefore should probably find a naming scheme for such extensions which is clear enough under all circumstances.

Following the parethnesed approach (1.) JSP syntax included int #2654 uses `<extended syntax> (for <target syntax>)` resulting in something like `CSS (for JSP)`. It is to be read as CSS for embedding in JSP.

FichteFoll:
Since "PHP CSS" and JS are hidden syntaxes, I didn't put much thought into naming them and just added something to prevent conflicts, but I agree that both `<base syntax> (<embedded>)` and `<extended syntax> (for <target syntax>)` would be better.

FichteFoll:
Technically, the PHP syntax is an extension of the HTML syntax, so should it be named "HTML (PHP)"? Or would that bee too confusing?

jfcherng:
php is a general purpose language, which is necessary to be related to html.

deathaxe:
Basically a PHP file is the same as any other HTML based templating resource so naming `HTML (PHP)` is somewhat logical, but I expect most people from becomming confused by this naming as it is indeed very uncommon.

For the other syntaxes it may make sense to use:

- `ASP` vs. `VBScript` instead of `HTML (ASP)` vs. `ASP` 
- `JSP` vs. `Java` instead of `HTML (JSP)`

rchl:
I think @jfcherng is making the point that PHP is most commonly used for HTML but it could as easily and correctly be used for a text file with PHP tags. It's just a matter of what content-type is returned for the given file.

That said, the most common use cause (or convention) could still be used as a deciding factor here rather than being fully precise.

rchl:
And after saying that, I've just realized that the PHP syntax in ST is actually specifically for the use case of HTML embedding the PHP blocks. So naming it HTML (PHP) would actually be perfectly fine.

jfcherng:
> Basically a PHP file is the same as any other HTML based templating resource so naming HTML (PHP) is somewhat logical, but I expect most people from getting confused by this naming as it is indeed very uncommon.

Yes, that's weird. I had never seen people calling PHP as "PHP for HTML" or an editor calling it like "HTML (PHP)".

> And after saying that, I've just realized that the PHP syntax in ST is actually specifically for the use case of HTML embedding the PHP blocks. So naming it HTML (PHP) would actually be perfectly fine.

Then I guess we need a PURE PHP syntax. I.e., the hidden "PHP Source" syntax.

PHP/CSS (for PHP).sublime-syntax
```diff
@@ -0,0 +1,38 @@
...
+    - include: php-tag-start
+
+  php-tag-start:
+    - match: <\?(?i:php|=)?(?![^?]*\?>)
+      scope: punctuation.section.embedded.begin.php
+      push:
+        - meta_scope: meta.embedded.block.php
+        - meta_content_scope: source.php.embedded.css
+        - include: php-end-tag-pop
+        - include: scope:source.php
```
deathaxe, line 22: PHP Source itself, pushes into HTML if it detects the `?>` tag, which results in CSS not being highlighted correctly afterwards. I guess, what we also need is a _PHP (for CSS).sublime-syntax_ which overrides the relevant contexts to make sure we push into CSS and pops back once the next `<?` is matched.

PHP/PHP.sublime-syntax
```diff
@@ -11,31 +15,93 @@ file_extensions:
...
-          push:
-            - meta_scope: meta.embedded.line.php
-            - meta_content_scope: source.php
-            - include: php-end-tag-pop
-            - include: scope:source.php
+  prototype:
+    - meta_prepend: true
+    - include: php-tag-start
+
+  tag-attribute-value-content:
```
deathaxe, line 26: The idea behind `tag-attribute-value-content` and `strings-common-content` is to enable proper string interpolation. Contexts included here should at least contain a `- clear_scopes: 1` in order to clear `string` scopes from embedded/interpolated code. Not doing so might disable any kind of auto-completion features for those code sections.

deathaxe: see: https://github.com/FichteForks/Packages/pull/8

keith-hall (approved):

PHP/CSS (for PHP).sublime-syntax
FichteFoll: Reading the code for this (in PHP Source), shouldn't the match for `?>` be redundant? Because this currently works just fine for me.

deathaxe: `?>` is not reduntant. See what happens when removing the match from CSS.

grafik

Your examples work only, because the statements used don't push into any context, thus the `?>` pattern in CSS file recovers state correctly.

As soon as you use a statement which pushes into any context in PHP, it will push into HTML when matching a `?>`, which causes the CSS syntax to break after such a statetement.

grafik

Fortunately those constructs seem quite rare in PHP.

What might be needed to fix those issues is to override the `embedded-html` context to push into CSS/JS instead of HTML, but I am not sure how good that works with regards to mixing two context sensitive syntaxes.

FichteFoll: Ah yes, I understand. I forgot that the Php syntax itself tries to maintain a context stack for blocks such as function definitions.

While I agree that these are issues, I don't want to solve them with this PR as they existed before and it is already bigger than I intended

deathaxe (approved):

jfcherng (approved):

deathaxe:
While this PR is in good state, I couldn't resist to port string interpolation support from (Rails/JSP) at https://github.com/FichteForks/Packages/pull/9

HTML/HTML.sublime-syntax
```diff
@@ -260,7 +260,7 @@ contexts:
         - include: style-css-content
 
   style-css-content:
-    - match: ''
```
deathaxe, line 263: The current implementation of HTML doesn't support wrapping CSS in html comments (`<!-- ... -->`) That's why the match is empty. It is by intent!

Those changes may fix some bugs, but are not within scope of this PR.

If we want to change the behavior it should go into a separate one as it results in further changes.

I'd like to come up with an implementation, which handles script/style tags the same way, but it should also fix some issues related with those html comments.

HTML/HTML.sublime-syntax
```diff
@@ -164,7 +164,7 @@ contexts:
         - include: script-javascript-content
 
   script-javascript-content:
-    - match: (?=\S)
```
deathaxe, line 167: This change fixes https://forum.sublimetext.com/t/st4-cant-toggle-comment-block-on-script-tags-in-html-files/57723

deathaxe: ... needs to be ported to ASP/JSP as well.

deathaxe: ... but it breaks html comments support

grafik

HTML/HTML.sublime-syntax
FichteFoll: Can you provide an example that this change breaks?

deathaxe:
```
        <style> h1 {} </style>
        ##     ^^^^^^^ source.css.embedded.html
```

... fails as `source` starts with `h1`. The space before is unscoped.

HTML/HTML.sublime-syntax
deathaxe: To be fixed via #2831

rwols:
> If we're gonna fix it, now is the right time ;)

Too late I guess?

HTML/HTML.sublime-syntax
FichteFoll: That looks good. I would like to rebase this PR on that once it is merged.

HTML/HTML.sublime-syntax
FichteFoll: Moving to the other discussion below, since it's about the same topic.

FichteFoll:
Nice find, updated OP.

jfcherng:
> Maybe worth mensioning. This PR fixes #643

It looks like the linked issue is already fixed now, somehow.

PR #2797, opened 2021-04-25
[Rails] Replace with_prototype by inheritance

deathaxe:
This PR implements Ruby on Rails using new inheritance features of ST4's syntax engine to replace the with_prototype usage.

It basically copies and adapts contexts from ASP and JSP (Java Rewrite).

Based on that new features HAML is totally rewritten with many syntax elements being added or fixed.

Embedded or interpolated Ruby code is now supported in all syntaxes HAML/HTML/CSS/JS.

**HAML**

Before:

grafik

After:

grafik

Various smaller test files are added to check most basic syntax features and make sure embedding/interpolation isn't broken by external syntax changes.

**Scopes**

Another major change is all main scopes are renamed to end with `.rails` instead of `.ruby`. 

Reasons are:

1. Ruby on Rails is an extended Ruby dialect which already used the `source.ruby.rails` scope.
2. This extended Ruby on Rails is what is embedded in CSS/JS/HTML. Thus `text.html.rails` seems more logical than `text.html.ruby`
3. The package name is "Rails".
4. All embedding syntaxes' main scopes should end with the same sub scope. So because of 1. the logical solution is `.rails`.


themilkman:
FYI, I checked this out and will test/use it. In case of noticing any issues I'll report them here.

deathaxe:
Put some effort into improving HAML, internally. Not sure if we want to address it in this or a separaate PR.

themilkman:
In `Rails/CSS (for HAML).sublime-syntax` there is a small mistake `hiden: true`

themilkman:
I liked your changes on HAML that much that I came back to something I wanted to do a long time ago: Implement HAML filters (embedded code in another language), like

```
:ruby
  foo = Hash.new
  foo[:a] = 42
  %i(a b c)

:scss
  a { color: blue }  
```


Based on the work of the Slim Package (https://github.com/slim-template/ruby-slim.tmbundle/blob/master/Syntaxes/Ruby%20Slim.tmLanguage) I added this to your HAML syntax file.
It's currently WIP but so far it works.
My commit is here:
https://github.com/themilkman/Packages/commit/ab110d3b45fa845b39defda2be472c31ac127ff0

Open points are:
* Add missing filters: plain, Escaped, preserve, CDATA (?), LESS (?)
* Add tests

I'd like to have your feedback if the way I did it is the way to go or if there is a way to do it better.
And of course, if you'd be willing to include this in your PR (or an extracted one if it goes too far as you already mentioned).

In case that you'd be open to include the changes upon your work, you may feel free to use/change/rebase/whatever in the way you like - I am everything but an experienced syntax writer.

Generally: I did not run in any issues regarding the new "Rails (Ruby)" and "HAML" syntaxes so far (but I only checked the latter against existing projects, currently I don't have any with HAML involved, only plain Ruby/Rails). 
The only thing which was a bit unexpected was the renaming. I like the names, but had to realize that it changed first and then adopt it to the BracketHighlighter etc.

Cheers and thanks for the work you have put in this PR!

deathaxe:
>  I like the names, but had to realize that it changed first and then adopt it to the BracketHighlighter etc.

The naming scheme to express an extended syntax via parentheses got to a defacto standard somehow. So the change felt natural. I am pretty sure we can prepare a PR for BracketHighlighter in advance to keep it working.

But I have no strong feelings about the name. If we expect trouble, we might change them back.

> Implement HAML filters 

I guess we can generally add basic support for filters in the default syntax, but only for syntaxes shipped with ST out of the box. So SASS/LESS can't be part of it by default.

It should be pretty simple to provide a 3rd-party package extending this package to add support though.

Your approach looks pretty well already.

In general I'd just recomment/try to use `embed` syntaxes if somehow possible, because ST4 now is smart enough to use the existing syntax definition then. Including it, means to copy all contexts, which may result in very large syntax trees, especially with Markdown being included, which itself embeds many others.

deathaxe:
What means `Escaped` and `CDATA` in detail?  CDATA is part of XML to mark text which is to be ignored by XML parser. Not sure in which way it makes sense to be used as a filter syntax. 

Sorry, if the question is dump, but I've never used Ruby nor HAML.

themilkman:
> The naming scheme to express an extended syntax via parentheses got to a defacto standard somehow. So the change felt natural. 

+1 for standards

> I am pretty sure we can prepare a PR for BracketHighlighter in advance to keep it working.

Sure, won't be a problem once this gets merged :)


> But I have no strong feelings about the name. If we expect trouble, we might change them back.

No I guess moving forward is always better.


> I guess we can generally add basic support for filters in the default syntax, but only for syntaxes shipped with ST out of the box. So SASS/LESS can't be part of it by default.

Gnarf, for LESS I also thought that this should not be included... but I missed that I installed an extra package for SCSS. And of course with coffee this is the same.


> In general I'd just recomment/try to use embed syntaxes if somehow possible, because ST4 now is smart enough to use the existing syntax definition then. Including it, means to copy all contexts, which may result in very large syntax trees, especially with Markdown being included, which itself embeds many others.


Great, thanks for this explaination! I'll take a deeper look in it soon with your changes in the commit.

> What means Escaped and CDATA in detail? CDATA is part of XML to mark text which is to be ignored by XML parser. Not sure in which way it makes sense to be used as a filter syntax.
> Sorry, if the question is dump, but I've never used Ruby nor HAML.

No, it's not dump, I didn't know about this too until yesterday.
The docs say:
https://haml.info/docs/yardoc/Haml/Filters/Escaped.html

`Works the same as Plain, but HTML-escapes the text before placing it in the document.`

https://haml.info/docs/yardoc/Haml/Filters/Cdata.html
`Surrounds the filtered text with CDATA tags.`

TBH I do not see any real reason to include CDATA. 

For Escaped I'd add it to 'filters-plain:' as this is my interpretation of what matches best.


As mentioned I see you already restructured and integrated it. A big thanks for this! Restructuring into sub-includes makes the syntax much more readable, I'll note this down for the future. 
I see that I wouldn't be able to keep up with your light speed! :)

deathaxe:
> Restructuring into sub-includes makes the syntax much more readable

The more important benefit is to provide flexibility for extending/inheriting syntaxes to replace or manipulate certain aspects of a syntax more easily.

themilkman:
Just for the records, I used the Ruby (Rails) syntax very much in the meantime and didn't see any issues.

I had to work a bit with the new HAML syntax too and really enjoyed the changes. This improves the HAML syntax massively.

themilkman:
Just as a last drop in here: I kept on working using this syntax for Rails (et al) and had no issues during the last months - it is a great improvement.

themilkman:
As the review was requested an update: I still use this  (always updated) branch for my daily work and never got any issues. A +1 from a "simple end-user perspective" for the merge.

michaelblyons (dismissed):
LGTM. Heads-up to @facelessuser for ApplySyntax and @jfcherng for AutoSetSyntax

jrappen:
@deathaxe Was there a reason for `Rails/Ruby (Rails).sublime-syntax` not being v2?

deathaxe:
Sure!

michaelblyons (approved):

keith-hall (approved):

PR #2798, opened 2021-04-26
[CSS] Add support for `:is` and `:where` pseudo class

niksy:
`:matches` is officially renamed to `:is`.

https://developer.mozilla.org/en-US/docs/Web/CSS/:is


deathaxe (approved):

willrowe:
Could we add [`:where`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) while we're at it?

niksy:
@willrowe I was just thinking about that yesterday. Will do!

wbond:
Looks good, thanks for contributing these fixes!

PR #2801, opened 2021-04-29
[Git Formats] Add support for CODEOWNERS

deathaxe:
This PR adds support for CODEOWNERS file.

It is not directly a "git metafile" but is used by github to assign code owners. It uses .gitignore syntax for file patterns with some
exceptions.

The original request comes from:
https://forum.sublimetext.com/t/codeowners-syntax-highlighter/57292/4


keith-hall (dismissed):

jrappen:
This seems fine. The missing empty line at the end of the test file was unexpected.

jrappen:
Also surprising, this new syntax is v2 but includes from a v1 base.

deathaxe:
It's only inheritance which needs consistent versions.

jrappen:
Yes, I am aware. Just my OCD to leave a comment when I come across this stuff.

deathaxe:
We could probably move all Git Formats to sublime-syntax v2 without changes.

jrappen (approved):

michaelblyons (approved):

jrappen:
@keith-hall Could you re-review this (and possibly merge afterwards)? Thanks.

Git Formats/tests/syntax_test_git_codeowners
```diff
@@ -0,0 +1,90 @@
...
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo, unless a later match takes precedence.
+*       @global-owner1 @global-owner2
+# <- meta.path.pattern.git.codeowners entity.name.pattern.git.codeowners keyword.operator.path.asterisk.fnmatch.git
+#^^^^^^^ - meta.path - meta.owners
+#       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.owners.git.codeowners
+#       ^^^^^^^^^^^^^^ meta.reference.username.git entity.name.reference.username.git
```
keith-hall, line 12: I know I approved this before, and this scope comes from a file untouched by this PR, but does it really make sense to have a `reference` sub-scope of `entity.name` - either it is an entity or a reference, but not both, no? Or is it justified so you can goto definition on a username and find every path that user is owning?

deathaxe: `reference` is what all those `@xyz` and `#123` tags are called by git's documentation, while `entity.name` is used with `entity.name.tag` in mind, IIRC. So reference and entity are not of the meaning of defining variables and references such as we do in source code.

... and yes, that's no invention of this PR. So if we want to change that scope, we should do so in a separate PR as it will effect all syntaxes.

We could probably change it to `markup.link.reference` or something along those lines to align with Markdown.

keith-hall (approved):

PR #2802, opened 2021-05-02
[Git Formats] Git Attributes maintanance

deathaxe:
This PR adds some maintanance changes to _Git Attributes.sublime-syntax_

1. Fixes the `immediately-pop` context name
2. Excludes prototype from contexts which are not meant to receive one by inheriting syntaxes
3. Adds a named context for quoted `string-content` 
4. Scopes strings `meta.string` in case interpolation is desired at any time.

### DIFF
--- a/Git Formats/Git Attributes.sublime-syntax
+++ b/Git Formats/Git Attributes.sublime-syntax
@@ -179,14 +179,10 @@ contexts:
     # double quoted value
     - match: \"
       scope: punctuation.definition.string.begin.git.attributes
-      set:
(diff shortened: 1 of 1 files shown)


wbond:
Were there any obvious errors from the old version we could test for?

deathaxe:
The changes are made just to keep inline with other syntaxes (`immediatelly-pop`) and apply some changes, which would provide some advantages in (very rare) case of someone inheriting them. That's why just "maintenance".

This PR doesn't change behavior except adding `meta.string`.

wbond:
Sounds good!

PR #2807, opened 2021-05-05
[Git Formats] Add support for Git Log HEAD line

keith-hall:
This allows Sublime to highlight the Git Log more like "regular `git log`" output in the case of the HEAD entry.
First reported at https://github.com/sharkdp/bat/issues/1632

Honestly, I wasn't too sure about what scopes to use, so if someone with more experience with the Git Formats syntaxes could chime in, it'd be much appreciated. :)

### DIFF
--- a/Git Formats/Git Log.sublime-syntax
+++ b/Git Formats/Git Log.sublime-syntax
@@ -13,24 +13,23 @@ contexts:
   main:
     # 1st header line
     # commit d9d9fb804f5d61c13ba2f8746af33a9f3c609075
-    - match: \b(commit)\s+(\h{7,})
-      scope: meta.header.git.commit markup.raw.block.git.log
+    - match: ^(commit)\s+(\h{7,})
       captures:
         1: keyword.other.commit.git.log
         2: constant.other.hash.git.log
       push:
+        - meta_scope: meta.header.git.commit markup.raw.block.git.log
         - match: $\n?
           scope: meta.header.git.commit markup.raw.block.git.log
+          pop: true
           embed: commit-header
           escape: (?=^commit\s)
-          pop: true
-        - match: \s*(\()(HEAD)\s*(->)
+        - match: (\()(HEAD)\s*(->)
           captures:
             1: punctuation.section.parens.begin.git.log
             2: support.type.git.log
             3: punctuation.separator.mapping.git.log
           push:
-            - meta_scope: meta.header.git.commit markup.raw.block.git.log
             - match: \)
               scope: punctuation.section.parens.end.git.log
               pop: true
--- a/Git Formats/Git Log.sublime-syntax
+++ b/Git Formats/Git Log.sublime-syntax
@@ -13,13 +13,33 @@ contexts:
   main:
     # 1st header line
     # commit d9d9fb804f5d61c13ba2f8746af33a9f3c609075
-    - match: (?:(commit)\s+(\h{7,}))?\s*\n
+    - match: \b(commit)\s+(\h{7,})
       scope: meta.header.git.commit markup.raw.block.git.log
       captures:
         1: keyword.other.commit.git.log
         2: constant.other.hash.git.log
-      embed: commit-header
-      escape: (?=^commit\s)
+      push:
+        - match: $\n?
+          scope: meta.header.git.commit markup.raw.block.git.log
+          embed: commit-header
+          escape: (?=^commit\s)
+          pop: true
+        - match: \s*(\()(HEAD)\s*(->)
+          captures:
+            1: punctuation.section.parens.begin.git.log
+            2: support.type.git.log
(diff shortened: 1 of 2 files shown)


Git Formats/Git Log.sublime-syntax
```diff
@@ -13,13 +13,33 @@ contexts:
...
+          embed: commit-header
+          escape: (?=^commit\s)
+          pop: true
+        - match: \s*(\()(HEAD)\s*(->)
+          captures:
+            1: punctuation.section.parens.begin.git.log
+            2: support.type.git.log
+            3: punctuation.separator.mapping.git.log
+          push:
+            - meta_scope: meta.header.git.commit markup.raw.block.git.log
```
deathaxe, line 33: If we move this up one level, the `scope` of the match in line 16 can be removed.

Git Formats/Git Log.sublime-syntax
```diff
@@ -13,13 +13,33 @@ contexts:
...
         1: keyword.other.commit.git.log
         2: constant.other.hash.git.log
-      embed: commit-header
-      escape: (?=^commit\s)
+      push:
+        - match: $\n?
+          scope: meta.header.git.commit markup.raw.block.git.log
+          embed: commit-header
+          escape: (?=^commit\s)
+          pop: true
```
deathaxe, line 26: How about moving `pop: true` between `scope` and `embed` as this would express the order those commands are executed in more exactly. The whole match first pops the context and than embeds `commit-header`.

Git Formats/Git Log.sublime-syntax
```diff
@@ -13,13 +13,33 @@ contexts:
   main:
     # 1st header line
     # commit d9d9fb804f5d61c13ba2f8746af33a9f3c609075
-    - match: (?:(commit)\s+(\h{7,}))?\s*\n
+    - match: \b(commit)\s+(\h{7,})
```
deathaxe, line 16: `\b` is not really required, is it?

`commit` is the first word in a log and `commit-header` pops only if it sees `commit`.

Git Formats/Git Log.sublime-syntax
```diff
@@ -13,13 +13,33 @@ contexts:
...
         2: constant.other.hash.git.log
-      embed: commit-header
-      escape: (?=^commit\s)
+      push:
+        - match: $\n?
+          scope: meta.header.git.commit markup.raw.block.git.log
+          embed: commit-header
+          escape: (?=^commit\s)
+          pop: true
+        - match: \s*(\()(HEAD)\s*(->)
```
deathaxe, line 27: Why the leading `\s*`? It's not needed, IMHO.

keith-hall: removed :)

Git Formats/Git Log.sublime-syntax
keith-hall: done :)

Git Formats/Git Log.sublime-syntax
keith-hall: done :)

Git Formats/Git Log.sublime-syntax
```diff
@@ -13,13 +13,32 @@ contexts:
...
+    - match: ^(commit)\s+(\h{7,})
       captures:
         1: keyword.other.commit.git.log
         2: constant.other.hash.git.log
-      embed: commit-header
-      escape: (?=^commit\s)
+      push:
+        - meta_scope: meta.header.git.commit markup.raw.block.git.log
+        - match: $\n?
+          scope: meta.header.git.commit markup.raw.block.git.log
```
keith-hall, line 23: this `scope` shouldn't be needed due to the `meta_scope`, but ST doesn't apply it to the match which pops if it also does an `embed`, it seems

deathaxe: Yes, all combined `pop-...` statements handle the match as it was a lookahead.

It is intended behavior and documentet.

Filed an [issue](https://github.com/sublimehq/sublime_text/issues/3783) about it as I found it inconsistent in situations like that.

But also found situations, it helps keeping things clean, so I don't have a very strong opinion about this behavior.

Basically it's more easy for us to add an extra line of `scope:` than to remove unwanted scopes.

This behaviour  just reduces the re-usability of such patterns as they are tight to the context they are used int.

deathaxe (approved):

Git Formats/Git Log.sublime-syntax
keith-hall: it's maybe not required, but does it do any harm? for context, I also plan to make another PR to fix `git show` diffs which are currently unhighlighted, and felt it would be clearer with the `\b`. (See https://github.com/sharkdp/bat/issues/1735 for context.)

jfcherng (approved):

PR #2810, opened 2021-05-06
[CSS] Fix @page property names

deathaxe:
This PR adds a dedicated context for @page property-lists.

The goal is to tidy common/ordinary property lists from special at-rule stuff and highlight only those properties in at-rules which are valid.


wbond:
It looks like this has become conflicted with the other CSS changes.

deathaxe:
Just because some rules added with previous commits are located next to some added by this one. Fixed.

jfcherng (approved):

CSS/CSS.sublime-syntax
```diff
@@ -1422,16 +1445,6 @@ contexts:
       pop: 1
     - include: rule-list-body
 
-  rule-list:
```
keith-hall, line 1425: this context was unused prior to this PR, right? I got confused for a while when I didn't see any references to it change, but that's because there wasn't any :sweat_smile:

keith-hall (approved):

CSS/CSS.sublime-syntax
deathaxe: Yes, it already got obsolete due to a prior PR. This one was not merged at the same time due to merge conflicts in syntax tests.

PR #2814, opened 2021-05-11
[JavaScript] add setInterval and clearInterval

silverwind:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearInterval


keith-hall (approved):

Thom1729 (approved):

wbond:
Thanks!

Issue #2073, opened 2019-08-26
[RFC] Scoping HEREDOCs


deathaxe:
While working on some fixes/improvements of several syntaxes I found HEREDOCs being scoped quite different.

Example:

```
str = <<"TAG"
the text
TAG
```

**Ruby** scopes all of `<<"TAG"` with `string.unquoted.heredoc punctuation.definition.string.begin` and the trailing `TAG` with `string.unquoted.heredoc punctuation.definition.string.end`

**Perl**'s (#2048) way of scoping `<<"TAG"` and `TAG` is inspired by Markdown, scoping `<<` with `keyword.operator` and `"TAG"` with `constant.other.language-name`.

**Bash** scopes `<<` as `keyword.operator.assignment` and uses `keyword.control.heredoc-token` for `"TAG"`.

**PHP** ...

---

I wonder whether we can find a common sense about how to scope HEREDOCs in all syntaxes.

Suggestion

1. The whole statement should be scoped with `meta.heredoc.<lang>` or `meta.string.heredoc.<lang>` due to string interpolation.
2. `<<` should probably be scoped `punctuation.definition.heredoc`
3. The `TAG` is called delimiter in some docs, but I am uncertain about whether to use the Merkdown strategy and scope it `constant.other.language-name` or use Bash's `keyword.control`. While the former looks reasonable for the opening tag, the latter one makes more sense with the closing tag. The opening tag basically defines the `keyword` for the closing tag. Scoping both differently seems odd. So I tend to use `keyword` for both of them.
4. How to handle quotations of the leading tag name? Is it a `punctuation.definition.keyword.begin` vs. `...end` (which would mean to scope them as keyword as well)  or is the whole tag a `string` (see: Perl)?
5. If the tags are not handled as pure punctuation, should they be scoped as `string.unquoted.heredoc` or should only the region between the tags be scoped as such?

keith-hall:
re point 3:
Markdown is a bit different in that the language hint doesn't need to be repeated for the closing code fence, so calling it a HEREDOC or treating it with the same rules doesn't seem right to me, though I agree that it serves a similar purpose and consistency is key.
[...]

PR #2822, opened 2021-05-18
[PHP] PHP 8.1 new syntax

jfcherng:
PHP 8.1 is scheduled to be 

- first alpha on `Jun 10 2021`
- feature freeze on `Jul 20 2021`
- stable release on `Nov 25 2021`

**References**:

- PHP 8.1 RFCs: https://wiki.php.net/rfc#php_81
- PHP 8.1 Milestone: https://github.com/php/php-src/milestone/3?closed=1
- PHP 8.1 release schedule: https://wiki.php.net/todo/php81

### Need new parsing paths

- [X] [PHP RFC: Enumerations](https://wiki.php.net/rfc/enumerations)
- [X] [PHP RFC: Explicit octal integer literal notation](https://wiki.php.net/rfc/explicit_octal_notation)

- [X] [PHP RFC: Add array_is_list(array $array): bool](https://wiki.php.net/rfc/is_list)
- [X] [PHP RFC: noreturn type](https://wiki.php.net/rfc/noreturn_type)
- [X] [PHP RFC: Add fetch_column method to mysqli](https://wiki.php.net/rfc/mysqli_fetch_column)
- [X] [PHP RFC: fsync() function](https://wiki.php.net/rfc/fsync_function)
- [x] [PHP RFC: Fibers](https://wiki.php.net/rfc/fibers)
- [x] [PHP RFC: Pure intersection types](https://wiki.php.net/rfc/pure-intersection-types)
- [X] [PHP RFC: Readonly properties 2.0](https://wiki.php.net/rfc/readonly_properties_v2)

### Others

- [X] [`MYSQLI_REFRESH_REPLICA` constant](https://php.watch/versions/8.1/MYSQLI_REFRESH_REPLICA)
- [X] [`CURLFile` and `CURLStringFile` classes](https://php.watch/versions/8.1/CURLStringFile)
- [X] [Use `keyword.declaration.*` for `class`, `interface`, ...](https://github.com/sublimehq/Packages/issues/582)
- [X] [Fix JSON heredoc scope](https://github.com/sublimehq/Packages/pull/2822/commits/a8793c9f53c7d98051c7385b2c744c4cb48ebc69)
- [X] [Add `.php8` extension](https://github.com/sublimehq/Packages/pull/2822/commits/3e3fcce285553f82eba1fa3884a317e86d2fa837)


PHP/PHP Source.sublime-syntax
```diff
@@ -560,6 +561,55 @@ contexts:
         - include: identifier-constant-pop
     - include: use-statement-common
     - include: identifier-constant-pop
+  enum:
+    - match: '(?i)\b(enum)\s+({{identifier}})\s*'
+      captures:
+        1: storage.type.enum.php
```
Ultra-Instinct-05, line 568: Maybe `keyword.declaration.enum.php` is more suitable than giving it a `storage` scope to align with https://github.com/sublimehq/Packages/issues/582 ?

jfcherng: Agree. Currently `class` (`storage.type.php`), `interface` (`storage.type.php`) and `trait` (`storage.type.trait.php`) are `storage`.

Ultra-Instinct-05: Yes, hence the suggestion to keep `storage` for backwards compatibility and also use `keyword.declaration` as a forward fill as described in https://github.com/sublimehq/Packages/issues/582#issuecomment-453231444

jfcherng: Resolved in https://github.com/sublimehq/Packages/pull/2822/commits/2a26d837e5a4ad385879573a035638ab68eed21e

PHP/PHP Source.sublime-syntax
```diff
@@ -18,10 +18,11 @@ contexts:
 
   statements:
     - include: class
+    - include: enum
     - include: use-statement
     - match: '(?i)^\s*(interface)\s+({{identifier}})\s*'
       captures:
-        1: storage.type.interface.php
+        1: storage.type.interface.php keyword.declaration.interface.php
```
deathaxe, line 25: ST4 uses a forwared fill for `keyword.declaration`. Hence you don't need to stack it onto `storage.type`. So you can remove it.

jfcherng: Is that done invisibly? I can't see it 
image

jfcherng: Okay. I checked JS' syntax definition. I guess https://github.com/sublimehq/Packages/pull/2822/commits/77a34829326adfd2f8c4647746957e716877bd27 should resolve this.

PHP/PHP Source.sublime-syntax
```diff
@@ -1307,14 +1307,14 @@ contexts:
           embed: scope:source.js
           embed_scope: meta.embedded.js source.js
           escape: (?=^\s*JAVASCRIPT\b)
-        - match: (<<<)\s*(JSON)\s*$\n?
+        - match: <<<\s*(JSON)\s*$\n?
```
deathaxe, line 1310: Maybe https://github.com/sublimehq/Packages/issues/2073

Scopeing everything punctuation seems sub optima. Already applied the RFC in several syntaxes.

jfcherng: just a note for myself: check https://github.com/deathaxe/sublime-packages/commit/9927f07a2f70801b8cef13a2d5412e9a0f3b1a6f

jfcherng:
Since today PHP 8.1 will be feature-freeze and those voting RFCs doesn't looks good. I think this PR is done now.

PHP/PHP Source.sublime-syntax
jfcherng: Changes are made in [`798c1fc` (#2822)](https://github.com/sublimehq/Packages/pull/2822/commits/798c1fc919f896739c050ec5b0dd85b000c62c9a)

deathaxe (approved):
I am fine with this PR. We just need to decide whether we want to hold it back as "feature". Not sure.

PHP/tests/syntax_test_php.php
```diff
@@ -855,8 +855,8 @@ interface MyInter {}
 // <- keyword.declaration.interface
 //        ^ entity.name.interface
 
-interface MyInter2 extends \MyNamespace\Foo {
-// <- keyword.declaration.interface
+interface MyInter2 extends \MyNamespace\Foo, /**/ \ArrayAccess {
+// <- keyword.declaration.inte
```
deathaxe, line 859: looks truncated

PHP/PHP Source.sublime-syntax
```diff
@@ -589,12 +592,61 @@ contexts:
...
+          scope: punctuation.separator.php
+          set:
+            - match: '(?={{path}})'
+              set:
+                - meta_scope: meta.path.php
+                - include: type-hint
+                - match: ''
+                  pop: 1
+        - match: '(?=\S)'
+          pop: 1
```
deathaxe, line 617: Is there a special reason for this lookahead?

Can it be modified as follows?

```suggestion
    - match: ':'
      scope: punctuation.separator.php
      push:
        - match: (?={{path}})
          set:
            - meta_scope: meta.path.php
            - include: type-hint
            - match: ''
              pop: 1
```

deathaxe: It is not needed at all, IMHO. The context is pushed only if a colon is matched. So the pattern in line 607 matches for sure and `set`s another context. So `(?=\S)` is never reached.

By removing the lookahead and `push`ing the path-context, nothing should change, if I don't miss a detail.

keith-hall: I was thinking the same

jfcherng: But what will happen if `(?={{path}})` doesn't match? Wouldn't it just kind of consume all rest of chars in the file?
```
    - match: ':'
      push:
        - match: (?={{path}})
          set:
            - meta_scope: meta.path.php
            - include: type-hint
            - match: ''
              pop: 1
```

deathaxe: Nothing else, compared to current state of the PR, because lint 616 is not indented correctly to be able to pop in case of (?={{path}}) doesn't match.

deathaxe: Agree, we should add it as alternative.

deathaxe:
```suggestion
    - match: ':'
      scope: punctuation.separator.php
      push:
        - match: (?={{path}})
          set:
            - meta_scope: meta.path.php
            - include: type-hint
            - match: ''
              pop: 1
        - match: (?=\S)
          pop: 1              
```

jfcherng: Thanks!

deathaxe (dismissed):

PHP/PHP Source.sublime-syntax
```diff
@@ -589,12 +592,59 @@ contexts:
...
     - include: identifier-constant-pop
+  enum:
+    - match: '(?i)\b(enum)\s+({{identifier}})\s*'
+      captures:
+        1: keyword.declaration.enum.php
+        2: entity.name.enum.php
+      push: enum-definition
+
+  enum-definition:
+    - match: "(?=;)"
```
keith-hall, line 603: Are the double quotes necessary here?

deathaxe: I guess this is a general issue in several locations of this and other syntaxes.

PHP/PHP Source.sublime-syntax
```diff
@@ -1217,110 +1267,152 @@ contexts:
...
+            2: meta.string.heredoc.php meta.tag.heredoc.php
+            3: punctuation.definition.tag.begin.php
+            4: entity.name.tag.heredoc.php
+            5: punctuation.definition.tag.end.php
           embed: scope:source.json
           embed_scope: meta.embedded.json source.json
           escape: (?=^\s*JSON\b)
-        - match: <<<\s*(CSS)\s*$\n?
-          scope: punctuation.section.embedded.begin.php punctuation.definition.string.php
+        - match: (<<<)\s*(()(CSS)())\s*$\n?
```
keith-hall, line 1374: Interesting technique to have empty capture groups to be able to copy/paste the scopes. Is it worth using a lookahead to check for both quotes / no quotes and make the capture groups for the quotes optional to avoid needing the almost-duplicate match pattern?

keith-hall (dismissed):

PHP/PHP Source.sublime-syntax
deathaxe: Good catch. Maybe we could use the following to reduce duplication?

```suggestion
        - match: (<<<)\s*(('?)(CSS)(\3?))\s*$\n?
```

I made the closing punctuation optional, too, to help with not breaking syntax in the middle of typing. But it could also be obligatory.

keith-hall: Doesn't using a capture group reference in a non-pop pattern cause Oniguruma to be used? Can you double check the Compatibility Tests please?

deathaxe: It does indeed trigger it and mark the rule incompatible. But it does so, too, if backrefs are used in popping patterns included in a pushed context. ST just doesn't complain then.

As PHP uses single quoted tags only, we could also use:

```suggestion
        - match: (<<<)\s*(('?)(CSS)('?))\s*$\n?
```

... to lazily handle quoted CSS.

deathaxe: We shouldn't merge those patterns atm, because one version supports PHP variable interpolation and the other does not. (see lines 1394 vs. 1406)

We may provide intermediate inherited syntaxes in future for those supporting interpolation. This is actually what I already prepared in https://github.com/deathaxe/sublime-Packages/tree/wip/php (needs some update once this one is merged, though).

keith-hall: somehow I missed that, thanks for spotting it :)

deathaxe:
Even though this PR is labeled "feature", it only adds some crucial changes to support the next minor version of PHP. I think it would be safe to merge it soon as it doesn't contain breaking changes, except changing HEREDOC tag highlighting, maybe. But the choosen scopes are already used in other syntaxes so they just increase consistency between syntaxes.

PHP/PHP Source.sublime-syntax
jfcherng: The `escape: (?=^\s*JSON\b)` can use backreference too actually.

keith-hall (approved):

deathaxe (approved):

PR #2826, opened 2021-05-22
Update Python.sublime-syntax to python 3.10

Penguin98kStudio:
Added Support for structural-pattern-matching introduced in python 3.10
Note: match wasn't marked as an invalid word because
import re
re.match() would not be legal then

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -37,7 +37,7 @@ variables:
   exponent: (?:[eE][-+]?{{digits}})
   path: '({{identifier}}[ ]*\.[ ]*)*{{identifier}}'
   sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|WITH)\b
-  illegal_names: (?:and|as|assert|break|class|continue|def|del|elif|else|except|finally|for|from|global|if|import|in|is|lambda|not|or|pass|raise|return|try|while|with|yield)
+  illegal_names: (?:and|as|assert|break|case|class|continue|def|del|elif|else|except|finally|for|from|global|if|import|in|is|lambda|not|or|pass|raise|return|try|while|with|yield)
   format_spec: |-
     (?x:
       (?:.? [<>=^])?     # fill align
@@ -234,6 +234,9 @@ contexts:
           scope: invalid.illegal.missing-in.python
           pop: true
         - include: target-list
+    - match: \bcase\b
+      scope: keyword.control.structural-pattern.case.python
+      push: case-body
     # async with ... as ...:
     - match: \b(async +)?(with)\b
       captures:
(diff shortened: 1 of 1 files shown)


Python/Python.sublime-syntax
```diff
@@ -37,7 +37,7 @@ variables:
   exponent: (?:[eE][-+]?{{digits}})
   path: '({{identifier}}[ ]*\.[ ]*)*{{identifier}}'
   sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|WITH)\b
-  illegal_names: (?:and|as|assert|break|class|continue|def|del|elif|else|except|finally|for|from|global|if|import|in|is|lambda|not|or|pass|raise|return|try|while|with|yield)
+  illegal_names: (?:and|as|assert|break|case|class|continue|def|del|elif|else|except|finally|for|from|global|if|import|in|is|lambda|not|or|pass|raise|return|try|while|with|yield)
```
FichteFoll, line 40: Since `case` is a new keyword, it shouldn't be added here as to not mess up code where `case` is used as an identifier, much like `match` is used in the `re` module. We didn't add the async keywords either for that reason.

Python/Python.sublime-syntax
```diff
@@ -303,6 +315,17 @@ contexts:
...
+  case-body:
+    - meta_scope: meta.statement.structural-pattern.case.python
+    - include: line-continuation-or-pop
+    - match: \b(as)\b
+      scope: keyword.control.structural-pattern.case.as.python
+      set: case-as
+    - match: ':(?!=)'
+      scope: punctuation.section.block.structural-pattern.case.python
+      pop: true
+    - include: expression-in-a-statement
```
FichteFoll, line 327: As far as I can tell, this doesn't follow `as`-matching into structured matches, i.e. `case [0, ('+'|'-') as op, right]:` or `case [(int() | float()) as value]:`.

https://www.python.org/dev/peps/pep-0635/#as-patterns

Python/Python.sublime-syntax
```diff
@@ -234,6 +234,9 @@ contexts:
           scope: invalid.illegal.missing-in.python
           pop: true
         - include: target-list
+    - match: \bcase\b
```
FichteFoll, line 237: Until 3.10, `case` is a normal identifier like any other. In order not to regress these cases, we will need to use branching to highlight `case` as a keyword only in match-case blocks. It doesn't need to be perfect, but it should be reasonable.

FichteFoll (changes requested):
Please add tests for the new syntax, as well as checks for the new keywords being used as identifiers where we don't want to regress.

Python/Python.sublime-syntax
Penguin98kStudio: Yes I noticed that too as an afterthought,

Python/Python.sublime-syntax
deathaxe: The easy way, was to add a context highlighting `as NAME` to `expressions-common` accepting it being highlighted as valid expression in inadequate positions.

The hard way is to add a full set of `case-pattern` contexts duplicating all group, tuple, list, set and most expression contexts just to add `as NAME` highlighting and probably exclude some contexts not required within this scope.

PR #2831, opened 2021-05-24
[HTML] Fix toggle html comments in css/js tags

### LINKED ISSUE #2322: [HTML] toggle_comment toggles html comment within CSS blocks
Forwarded from: 
- https://forum.sublimetext.com/t/toggle-comment-inconsistant-syntax-comment-for-scope-in-html-file/50413
- https://forum.sublimetext.com/t/st4-cant-toggle-comment-block-on-script-tags-in-html-files/57723/3

Environment:
- Sublime Version: 3211, 4070
- OS Version: all

deathaxe:
Fixes #2322
Fixes sublimehq/sublime_text#4701

In order to enable comment toggling using rules of embedded syntax definitions it is important to maintain its main scope at the beginning of first and last line.

The challenge is to also support CSS/JS being wrapped into HTML comment markers.

Another goal is to keep as many implementation details away from contexts which may need to be overridden by inheriting syntax definitions such as ASP/JSP/PHP to keep dependencies low. Some little changes are required though at this point to accomplish that.

_Note: This commit includes the assumption that JavaScript won't contain a standalone `-->` on a single line._

Also note this PR may have infuence on implementation details in #2654, #2789 and #2797

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -173,7 +173,7 @@ contexts:
 
   script-javascript-content:
     - meta_include_prototype: false
-    # Note: This context is to be overridden to replace embedded css syntax.
+    # Note: This context is to be overridden to replace embedded js syntax.
     #       It is stripped from as most implementation details as possible.
     #       Maybe simplified once core issue #3787 is resolved.
     - match: ''
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -43,9 +43,9 @@ variables:
       | text/livescript
     )
 
-  script_content_begin: <!--|(?=\s*(?!<!--)\S)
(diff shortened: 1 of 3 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -160,14 +162,36 @@ contexts:
...
+  script-javascript-begin:
+    - meta_include_prototype: false
+    - match: '{{script_content_begin}}'
+      scope: comment.block.html punctuation.definition.comment.begin.html
+      set: script-javascript-content
 
   script-javascript-content:
-    - match: (?=\S)
+    - meta_include_prototype: false
+    # Note: This context is to be overridden to replace embedded css syntax.
```
FichteFoll, line 176:
```suggestion
    # Note: This context is to be overridden to replace embedded JavaScript syntax.
```

FichteFoll (approved):

keith-hall (approved):

jfcherng (approved):

PR #2837, opened 2021-05-27
[Markdown] Allow indented reference links after lists

### LINKED ISSUE #2835: [Markdown] The `meta.link.reference.def.markdown` Selector Doesn't Catch All Markdown Reference Links
### Description

The `meta.link.reference.def.markdown` selector doesn't Catch all Markdown reference links.
Namely `ref_links = self.view.find_by_selector("meta.link.reference.def.markdown")` won't hold all cases in Markdwon.

### Steps to reproduce

1. Start Sublime Text (Build `4107`).
2. Create a new Markdown document (Open an empty `.md` file).
3. Copy the following text:

```
Some list...

 *  Examining Deep Learning (DL) Based Methods  
  Always some text on Deep Learning.
 *  Hyper Parameters  
  Adding an exhaustive search of the optimal Hyper Parameters.


  [001]: https://en.wikipedia.org/wiki/Root-mean-square_deviation "Wikipedia - RMSE"
  [002]: https://en.wikipedia.org/wiki/Mean_absolute_error        "Wikipedia - MAE"
```

### Expected behavior

If the caret is before `[001]: https://en.wikipedia.org/wiki/Root-mean-square_deviation "Wikipedia - RMSE"` or `[002]: https://en.wikipedia.org/wiki/Mean_absolute_error        "Wikipedia - MAE"` the scope is expected to be a Markdown reference link.

### Actual behavior

The scope is a list:

image

It seems the issue is being after a list and having spaces.

### Environment

* Build: 4107.
* Operating system and version: Windows 10 21H1.

keith-hall:
Fixes #2835

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -1034,6 +1034,8 @@ contexts:
         - include: fenced-code-block
         - match: ''
           pop: true
+    - match: '^(?=\s*\[[^]]*\]:)'
+      pop: true
     - match: \s+(?=\S)
       push:
         - match: ^\s*$
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -278,26 +278,7 @@ contexts:
         - match: ^(?![ ]{4}|$)
           pop: true
         - include: inline-bold-italic
-    - match: |-
-        (?x:
-            \s*                        # Leading whitespace
-            (\[)([^]]*)(\])(:)         # Reference name
-            [ \t]*                     # Optional whitespace
-            (?:
-              (<)([^>]+)(>)            # The url
-            | (\S+)                    # The url
-            )
-        )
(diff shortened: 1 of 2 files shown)


jwortmann:
Not sure if it makes a difference or is relevant, but wouldn't it be correct to include the match for reference link definitions into `list-paragraph`?

According to the CommonMark specs,
> Link reference definitions can occur inside block containers, like lists and block quotations. They affect the entire document, not just the container in which they are defined

This markdown code is interpreted as a single list with 3 items:
```markdown
1. first list item

1. second list item

   [001]: https://en.wikipedia.org/wiki/Root-mean-square_deviation "Wikipedia - RMSE"

1. third list item
```

keith-hall:
Probably - I was thinking about that while working on it. I'll see if I can refactor it sometime soon :)

keith-hall:
@jwortmann done :)

PR #2838, opened 2021-05-27
[C#] Support tuples whose commas are not on the same line as the open paren

### LINKED ISSUE #2832: [C#] Method expression-bodies members
- OS Version: Windows 10 19042.985

Hi, I am seeing this issue in C# where method expressions show a syntax error on deeply-nested code.

``` csharp
[Fact]
public void ApplyValidationFail() =>
  Assert.Equal(
    "[[fail_1], [fail_2]]",
    (
      Success<Seq<string>, Option<int>>(3),
      Fail<Seq<string>, Option<int>>(Seq1("fail_1")),
      Success<Seq<string>, Option<int>>(3), Fail<Seq<string>,
      Option<int>>(Seq1("fail_2"))
    )
    .Apply((a, b, c, d) =>
      from w in a
      from x in b
      from y in c
      from z in d
      select w + x + y + z)
    .FailToSeq()
    .ToString()
  ); // <-- Syntax error
```

Seems to work fine on simple expressions.

```chsarp
[Fact]
public void BindProp() =>
  Assert.Equal(
    None,
    Optional(1)
      .Bind<int>(o => o + 2)
      .Bind<int>(_ => None)
      .Bind<int>(_ => 1)
  ); // <-- No syntax error
```

Attached are screenshots of how this shows in Sublime.

image
image

keith-hall:
Fixes #2832

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1330,43 +1330,12 @@ contexts:
             2: punctuation.separator.cs
             3: punctuation.section.brackets.end.cs
             4: keyword.operator.pointer.cs
-    - match: \((?=(?:[^,)(]|\([^\)]*\))*,)
-      scope: punctuation.section.group.begin.cs
-      push:
-        - meta_scope: meta.group.tuple.cs
-        - match: \)
-          scope: punctuation.section.group.end.cs
-          pop: true
-        - match: ','
-          scope: punctuation.separator.tuple.cs
-        - match: ':'
-          scope: keyword.operator.assignment.cs
-        - match: _\b
-          scope: variable.language.deconstruction.discard.cs
-        - match: (?!{{reserved}})(?={{namespaced_name}}{{type_suffix}}\s+{{name}}\s*[:,])
-          push: var_declaration_explicit
-        - match: '({{name}})(<)'
-          captures:
-            1: support.type.cs
-            2: meta.generic.cs punctuation.definition.generic.begin.cs
-          push:
-            - meta_content_scope: meta.generic.cs
-            - match: ','
-              scope: punctuation.separator.type.cs
-            - match: '>'
-              scope: meta.generic.cs punctuation.definition.generic.end.cs
-              pop: true
-            - include: type
-        - include: line_of_code_in_no_semicolon
     - match: \(
       scope: punctuation.section.group.begin.cs
-      push:
-        - meta_scope: meta.group.cs
-        - match: \)
-          scope: punctuation.section.group.end.cs
-          pop: true
-        - include: lambdas
-        - include: line_of_code_in_no_semicolon
+      branch_point: tuple_or_group
+      branch:
+        - group
+        - tuple
     - match: \{
       scope: punctuation.section.block.begin.cs
       set:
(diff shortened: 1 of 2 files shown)


C#/C#.sublime-syntax
```diff
@@ -1377,6 +1346,49 @@ contexts:
     - match: (?=\}|\)|>|\]|,|{{reserved}})
       pop: true
 
+  tuple:
+    - meta_scope: meta.group.tuple.cs
```
deathaxe, line 1350: Would we probably want to use `meta.sequence.tuple` and corresponding punctuations to line up with python?

C#/C#.sublime-syntax
```diff
@@ -1330,43 +1330,12 @@ contexts:
...
     - match: \(
       scope: punctuation.section.group.begin.cs
-      push:
-        - meta_scope: meta.group.cs
-        - match: \)
-          scope: punctuation.section.group.end.cs
-          pop: true
-        - include: lambdas
-        - include: line_of_code_in_no_semicolon
+      branch_point: tuple_or_group
```
deathaxe, line 1335: I'd probably vote for using the same name scheme as for contexts and keep underscore use reserved for variables only. It makes it easier to give them the same name as a context, as already noted in another PR.

deathaxe: Nevermind. I just see this syntax uses underscors for contexts, too. So this name is very fine.

C#/C#.sublime-syntax
```diff
@@ -1377,6 +1346,49 @@ contexts:
...
+      scope: keyword.operator.assignment.cs
+    - match: _\b
+      scope: variable.language.deconstruction.discard.cs
+    - match: (?!{{reserved}})(?={{namespaced_name}}{{type_suffix}}\s+{{name}}\s*[:,])
+      push: var_declaration_explicit
+    - match: '({{name}})(<)'
+      captures:
+        1: support.type.cs
+        2: meta.generic.cs punctuation.definition.generic.begin.cs
+      push:
```
deathaxe, line 1366: We probably should try to introduce "named contexts for everything". Not sure if C# is a candidate, but it helps with inheriting syntaxes as well as debugging via scope popup, which now shows context names.

keith-hall: I like this point. May be worth adding to https://github.com/sublimehq/Packages/issues/757

deathaxe: I wonder if we should create a more structured kind of documentation for things like that in the long term.

I have too much such details in mind but somehow find it hard to create documentation from it.

C#/C#.sublime-syntax
keith-hall: I agree. I took a look into this, and it's a complicated change. Maybe I can do it as part of a separate PR please? :)

deathaxe: Sure.

C#/C#.sublime-syntax
deathaxe: Refactoring syntaxes to use named contexts can be done in dedicated PRs as well.

deathaxe (approved):

FichteFoll (approved):

PR #2842, opened 2021-05-29
[Ruby] Patch Monokai Color Scheme

### LINKED ISSUE #2841: [Ruby] Syntax Highlight in Monokai theme on Sublime 4
- Sublime Version: 4
- OS Version: Ubuntu 18.04.5 LTS

Issue is only with Monokai theme -
Incorrect blue color of keywords `module`, `class`, `def`. It was red on previous sublime versions.

```ruby
module Foo
  class Bar
    def method
    end
  end
end
```

I fixed this for myself with this config (Monokai.sublime-color-scheme):
```json
// Documentation at https://www.sublimetext.com/docs/color_schemes.html
{
  "variables":
  {
  },
  "globals":
  {
  },
  "rules":
  [
    {
        "name": "Storage type",
        "scope": "storage.type",
        "foreground": "var(red2)",
        "font_style": "italic"
    }
  ]
}
```

deathaxe:
Fixes #2841

This commit adds syntax specific patch to bring back old tinting of declaration keywords, which used to be scoped `keyword.control` in ST3.

### DIFF
--- a/Ruby/Monokai.sublime-color-scheme
+++ b/Ruby/Monokai.sublime-color-scheme
@@ -1,0 +1,10 @@
+{
+	"rules":
+	[
+		{
+			"name": "Declaration Keywords",
+			"scope": "source.ruby keyword.declaration",
+			"foreground": "var(red2)"
+		}
+	]
+}


keith-hall (approved):

jwortmann:
Please see my comment at https://github.com/sublimehq/Packages/pull/2844#issuecomment-850808841.

There should be no fragmentation of color schemes.

deathaxe:
Agree in general, but those patterns should keep "Ruby specific" as it is quite special to move from `keyword.control` to `keyword.declaration`.

jwortmann:
Yes, which would be the case if the rule from this PR is copied into the Monokai fiile. Monokai already has syntax-specific rules for C/C++/Obj-C/Obj-C++, JavaScript, YAML, Diff and an outdated rule for JSON, so it would be easy to add this Ruby rule to restore the former highlighting style.

deathaxe:
Hmm. Thanks for clarifying. I thought something like that was impossible. SARCASM OFF (sorry I couldn't resist). 😈

deathaxe:
Fixed in ST 4109.

PR #2844, opened 2021-05-29
[JSON] Patch Monokai Color Scheme

deathaxe:
This commit adds syntax specific patch for Monokai.sublime-color-scheme to highlight JSON keys red to make them stand out compared to values.

**Before**

grafik

**After**

grafik

### DIFF
--- a/JSON/Monokai.sublime-color-scheme
+++ b/JSON/Monokai.sublime-color-scheme
@@ -1,0 +1,9 @@
+{
+	"rules": [
+		{
+			"name": "JSON Key Names",
+			"scope": "source.json meta.mapping.key string - punctuation",
+			"foreground": "var(red2)"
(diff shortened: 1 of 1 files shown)


jwortmann:
While I would welcome this change from the highlighting perspective, it should be made in the *Color Scheme - Default* package instead. Otherwise it would be very confusing for users, if they have an override/customization with `meta.mapping-key string`, but they wonder why it doesn't work for JSON, and the reason is unexplainable when looking at the original Monokai file on the left-side when using the *Preferences > Customize Color Scheme* menu.

*Color Scheme - Default* is not contained in this repository, but it would be a trivial change to do for the Sublime HQ members.

Related: https://github.com/sublimehq/sublime_text/issues/3590

PR #2845, opened 2021-05-29
[CSS] Patch Monokai Color Scheme

deathaxe:
Resolves https://forum.sublimetext.com/t/monokai-color-scheme-of-st3-how-to-setup-in-st4/58159

This commit adds syntax specific patch of Monokai.sublime-color-scheme to bring back old tinting of numeric units which used to be scoped `keyword.other.unit` in ST3.

### DIFF
--- a/CSS/Monokai.sublime-color-scheme
+++ b/CSS/Monokai.sublime-color-scheme
@@ -1,0 +1,10 @@
+{
+	"rules": [
+		{
+			"name": "Numeric Units",
+			"scope": "constant.numeric.suffix.css",
+			"foreground": "var(red2)",
+			"font_style": ""
+		}
+	]
+}


SohailAhmad191 (commented):
Thank You

CSS/Monokai.sublime-color-scheme
```diff
@@ -0,0 +1,10 @@
+{
+	"rules": [
+		{
+			"name": "Numeric Units",
+			"scope": "constant.numeric.suffix.css",
+			"foreground": "var(red2)",
+			"font_style": ""
```
FichteFoll, line 7: Should it specify a font style?

FichteFoll (commented):
I'm undecided whether it is a good idea to introduce syntax-specific scheme overrides inside the respective package instead of adjusting the color scheme as it is shipped. Not like it is in our power to do so, but it's something worth discussing.

CSS/Monokai.sublime-color-scheme
deathaxe: Copied it from forum and added name. Would remove font_style if one was set by `constant.numeric`. Maybe not needed?

jwortmann:
Is there any advantage that I missed if using the approach from this PR, instead of patching the main color scheme file itself? As I see it, this fragmentation of color schemes as it is suggested here has zero advantages, but only the disadvantage I pointed out in the other PR.

In general I think that it would probably have been better to adjust the built-in color schemes to the new scopes, if the old styles should be preserved, instead of this "forwad fill" for certain scopes. Custom color schemes can easily be updated to the new scopes by their authors or tweaked by users if styles shouldn't change. As it is now, this forward fill has not only unwanted implications for the declaration keywords in Ruby, but also for third-party syntaxes if they already used `keyword.declaration` before, see for example https://github.com/JuliaEditorSupport/Julia-sublime/issues/103.

Btw, some of the styles did not only change in Monokai, but also in the other built-in color schemes like Mariana (e.g. declaration keywords are now italic instead of regular in Ruby). It's just not as noticeable there, because coincidentally the colors for `storage.type` and `keyword` are the same in Mariana.

CSS/Monokai.sublime-color-scheme
jwortmann: The empty string is required here to reset the italic font style which would otherwise be applied due to the `storage.type` forward fill. (Assuming the old style from ST3 should exactly be restored)

deathaxe:
> Is there any advantage that I missed if using the approach from this PR

It is of same value as syntax specific key bindings in this repo. All of them could be placed into core as well. They exist here, because it is the only place the community can contribute to. Also it makes somewhat sense to place stuff which is valid for a single syntax only into a dedicated package and treat it as any other 3rd-party extension. So if the package is disabled all related files/binding/... are ignored as well.

_TBH, I'd prefer default/main color schemes to be patched as most rules might be applied to 3rd-party packages as well._

So maybe those PRs might be interpreted as template to patch default color scheme.

> In general I think that it would probably have been better to adjust the built-in color schemes to the new scopes, if the old styles should be preserved, instead of this "forwad fill" for certain scopes.

Forward fills mainly target backward compatibility with existing 3rd-party color-schemes most of them no longer maintained, which means "can easily be updated" doesn't really apply - at least it is quite unlikely to happen.

CSS units would have been effected the one way or the other. The original scope doesn't matter. Forward fills are used to reduce such effects.

_Default color schemes should of course contain explicit rules for new scopes to prevent forward fills from being applied at all._

rwols:
Honestly I don't think this should be merged as I consider Monokai to be deprecated. Not to mention that language-specific tweaks are not maintainable in the long run.

deathaxe:
Those patches intend to maintain old colors, which is required due to several scope changes during last dev cycle. Monokai being deprecated or not, it still seems quite popular and prefered by several users, which also want their old colors back.

That's why those changes should find their way to the main Monokai.sublime-syntax.

PR #2851, opened 2021-06-01
[Markdown] Fix YAML front matter ending to allow `...`.

Ultra-Instinct-05:
This PR fixes the markdown YAML front matter ending to allow the use of `...` as well.

This is as per the YAML spec where the document ending could either be a `---` or a `...`

See: https://yaml.org/spec/1.2/spec.html#id2760395

P.S. I guess I can't write any tests for it because as per the current syntax definition, YAML front matter needs to be at the start of a file (which is also the case for the syntax test header, so conflicts).


keith-hall (approved):

FichteFoll (approved):
Never seen it used like that, but it makes sense semantically. And yes, this cannot be tested.

PR #2853, opened 2021-06-01
[PHP] Fix AC scope

jfcherng:
Fixes https://forum.sublimetext.com/t/st4-php-autocompletion-workflow/58405

We don't need AC for built-in functions when in `string` or `comment`.


wbond (approved):
LGTM

deathaxe (approved):

PR #2855, opened 2021-06-02
[LaTeX] Fix duplicated backslashes inserted by snippets

### LINKED ISSUE #2730: [LaTex] Snippets insert double backslash
- Sublime Version: 4098
- OS Version: Windows 10

## Expected behavior
Only one backslash is inserted

## Actual behavior
When typing `\begin`, pressing Tab
```latex
\\begin{env}
        
\end{env}
```
is inserted.
Same for item, equation, and any other environment
## Steps to reproduce

1. Start ST in safe mode
2.  See actual behavior

deathaxe:
Fixes #2730

This commit modifies `tabTrigger` of those snippets, whose committed content starts with exactly the same characters.

As a result both, typing:

a) `begin`
b) `\begin`

result in

```latex
  \begin{env}
    |
  \end{env}
```

when hitting tab (twice?!)

Also removes escaped backslashes from content as those seem not to be needed. So \\begin becomes \begin.

### DIFF
--- a/LaTeX/Snippets/Chapter.sublime-snippet
+++ b/LaTeX/Snippets/Chapter.sublime-snippet
@@ -3,7 +3,7 @@
 \label{cha:${2:${1/\\\w+\{(.*?)\}|\\(.)|(\w+)|([^\w\\]+)/(?4:_:\L$1$2$3)/g}}}
 ${0:$TM_SELECTED_TEXT}
 % chapter $2 (end)]]></content>
-	<tabTrigger>cha</tabTrigger>
+	<tabTrigger>\cha</tabTrigger>
 	<scope>text.tex.latex</scope>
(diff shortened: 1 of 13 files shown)


wbond (approved):
Generally this fix looks fine.

It is rather unfortunate that ST appears to strip leading non-alphanumeric in the tabTrigger. That feels inconsistent, but I suppose the length of time that has happened probably means it is set in stone until a new format/version for snippets happens.

deathaxe:
It appears to be treated consistently in ST4, as some of those triggers had different formats before but were both treated equally, IIRC.

But I agree, ST is probably too restrictive in some situations with regards to heuristics, which can be a burdon for some syntaxes.

jfcherng (approved):

Issue #1862, opened 2019-02-02
File named "build" always opens as "NAnt Build FIle" type


madneon:
How to reproduce:
Create file named "build", put anything into it, eg. `#!/bin/bash`, and then save it.
Closing and opening the file switches to "NAnt Build File", also on Sidebar file is wrongly detected.

deathaxe:
1. The _NAnt Build File_ is defined in C#/Build.sublime-syntax. 
2. This syntax is associated with the _build_ file extension.
3. ST uses file extensions to choose the syntax first. If none matches, the `first_line_match` rule is used. That's why the shebang is ignored.
4. ST uses the last part of a file path as "extension", which means it matches both _foo.build_ and _build_. We could argue about a file named _build_ is to be a file extension, but changing the behavior would break syntax assignment for files like _.gitignore_.

With that said this issue is not related with the _Default Packages_ and can not be fixed in the syntax defintions without changing ST core behavior.

A backward compatible solution would probably be to introduce a new `file_pattern` context which was used instead of `file_extensions` if exists. It would then need to support simple file patterns.

```
file_patterns:
  - *.build     // file extension
  - build       // file name
```

keith-hall:
see https://github.com/SublimeTextIssues/Core/issues/1353 for a prior discussion

michaelblyons:
FWIW, I use [ApplySyntax](https://packagecontrol.io/packages/ApplySyntax) for more fine-grained control until things like @deathaxe's wish are implemented.  Here's the main problem I use it for:

- `~/.ssh/config`
- `~/projects/foo/.git/config`

Both files have the same filename. Neither one has a shebang or other guaranteed first line stuff. How do you distinguish them? Even with @deathaxe's system, can the patterns test the full path?

Even ApplySyntax comes with a caveat: I have to disable its default rules by setting `default_syntaxes: []` to avoid interfering with the [PackageDev](https://packagecontrol.io/packages/PackageDev)-defined languages for special XML files (`.tmPreferences`, etc.).

FichteFoll:
As a side note, I'm not sure `build` is a specific enough extension to justify having it opened with the C++ build system thingy by default. Meson, for example, also uses the `.build` extension, though that is always written as `meson.build`.
Can you think of other examples?

PR #2858, opened 2021-06-04
Highlight BUILD files as python code

jlisee:
Highlight files named `BUILD` as if they were Python.  These files will almost
universally contain code written in starlark which is a dialect of Python. This
language is used in the popular build system Bazel and it's derivatives like
Pants and Please.

While some build systems, like Bazel, are trying to recommend using a suffix
bazelbuild/bazel#4517 like `BUILD.bazel`, consensus is not universal and there
is a large body of code as well as other build systems like Pants which also
default to the extensionless `BUILD` name.

The motivation here is  provide proper syntax highlighting with the `bat` tool.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -23,6 +23,7 @@ file_extensions:
   - Snakefile
   - vpy
   - wscript
+  - BUILD
   - bazel
   - bzl
 first_line_match: ^#!\s*/.*\bpython(\d(\.\d)?)?\b


FichteFoll:
Unfortunately, `build` (or `BUILD`, apparently) is a rather contested name. https://github.com/sublimehq/Packages/issues/1862

jlisee:
> Unfortunately, `build` (or `BUILD`, apparently) is a rather contested name. #1862

Yeah that is a bummer, but I thought these extensions were case sensitive.  The current file has both `SConstruct` and `Sconstruct` indicating that is the case.

Assuming that is the case we can set aside the `build` (lower case) dispute, and focus on if there is consensus on `BUILD` (upper case).  I think it's pretty clear that the all caps `BUILD` files are in python syntax so this change would help users of Sublime as well as downstream dependencies like `bat`.

FichteFoll:
Do these files happen to have a shebang line or other identifying information in the first line of the file?

jfcherng:
On Windows, it won't be case-sensitive. That is, `mAkEfIlE` will use the `Makefile` syntax.

Ultra-Instinct-05:
One can also have `"extensions": ["BUILD"]` in `Python.sublime-settings` to have files that are named `BUILD` to have Python's syntax highlighting.

jlisee:
> Do these files happen to have a shebang line or other identifying information in the first line of the file?

Nope!  They are hold over from the Google's internal monorepo where they didn't have to care about global conventions getting in the way of simple naming.

So I think it best to shelve this and thanks for the workarounds.  In theory even on windows you could make this system case sensitive, but I understand that probably doesn't make sense.

PR #2862, opened 2021-06-08
[Makefile] [Perl] [Ruby] [ShellScript] Do not require only whitespace before -*-

scop:
Any characters are valid before the first `-*-`. Another `-*-` before the one we're interested in may throw detection off, but that's not easy to avoid with a regex, and it's unlikely to cause blatantly false positives, so accept everything there.

Lots of sample files that are affected and associate fine with shell script mode in Emacs are available at https://github.com/scop/bash-completion/tree/master/completions

### DIFF
--- a/Makefile/Makefile.sublime-syntax
+++ b/Makefile/Makefile.sublime-syntax
@@ -24,7 +24,7 @@ file_extensions:
 first_line_match: |-
   (?xi:
     ^\#! .* \bmake\b |                     # shebang
-    ^\# \s* -\*- [^*]* makefile [^*]* -\*- # editorconfig
+    ^\# .* -\*- [^*]* makefile [^*]* -\*-  # editorconfig
   )
 scope: source.makefile
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


deathaxe:
Good catch! Maybe should be synced with/into https://github.com/sublimehq/Packages/pull/2760 as probably any syntax is effected.

scop:
Yup, agree. #2760 looks great BTW :+1:

deathaxe:
I'd prefer to merge #2760 as it contains everything of this PR and addresses all packages in one step.

jfcherng (approved):

jfcherng:
> I'd prefer to merge #2760 as it contains everything of this PR and addresses all packages in one step.

I assume this PR can be closed now since #2760 has been mreged.

deathaxe:
It's not yet merged, but yes the mensioned one contains all requested changes of this PR, but addresses all syntaxes.

jfcherng:
oops, somehow I thought it was merged already.

PR #2870, opened 2021-06-16
[Python] Explicitly support shell in command use

XinArkh:
Previous commit (https://github.com/sublimehq/Packages/pull/2379) changed `shell_cmd` to `cmd`,
causing some python modules to fail to interact. e.g. matplotlib cannot plot the figure after calling plt.show().
Explicitly set the shell mode will fix this issue. 
Reference: https://stackoverflow.com/a/15467387/10820028

BTW: `py` launcher is not a robust name to call python in Windows. Some python environments such as Miniconda
doesn't support this feature. So I personally hope the administrator reconsider the usage of `py`.
Reference: https://forum.sublimetext.com/t/winerror-2-in-new-sublime-text-4/57881


deathaxe:
Use of `shell: true` is deprecated and should no longer be used. If `cmd` causes issues the way to go was to use `shell_cmd` instead.

(Disclaimer) Without any further investigation - ST3 and 4 read environment variables from login shell upon application startup, which ST2 didn't. So running "normal python" scripts should work with `cmd` the same way as calling them via interactive shell. I guess if Minicoda sets up its own environment without ST being started using that shell, `shell_cmd` might be required though.


Those default packages provide what the name says - "defaults". They can't satisfy everybodies tastes.

Py launcher was introduced due to numberous complaints about python not being found on Windows. It is installed by "original" Python for Windows installer by default. Minicoda or other packages not providing it, doesn't mean it isn't robust. It just means Minicoda not providing it causing little extra work to set things up for its users.

XinArkh:
> Use of `shell: true` is deprecated and should no longer be used. If `cmd` causes issues the way to go was to use `shell_cmd` instead.
> 
> (Disclaimer) Without any further investigation - ST3 and 4 read environment variables from login shell upon application startup, which ST2 didn't. So running "normal python" scripts should work with `cmd` the same way as calling them via interactive shell. I guess if Minicoda sets up its own environment without ST being started using that shell, `shell_cmd` might be required though.
> 
> Those default packages provide what the name says - "defaults". They can't satisfy everybodies tastes.
> 
> Py launcher was introduced due to numberous complaints about python not being found on Windows. It is installed by "original" Python for Windows installer by default. Minicoda or other packages not providing it, doesn't mean it isn't robust. It just means Minicoda not providing it causing little extra work to set things up for its users.

Haha very humorous of you, and thank you for your information. 

I still have some further questions:

1. Do you recommend I change `cmd` to `shell_cmd` and resubmit a pull request?
2. And in my opinion, the word "robust" means it could work in as many environments as possible. A truth is that Miniconda is one of the popular python environments, yet it fails to run ST in Windows. Is there a way to accommodate these users, or at least leave some tips in the documentation to help them debug?

Issue #2875, opened 2021-06-18
[Lua] Shebang highlighting


rosik:
- Sublime Version: 4107
- OS Version: Ubuntu 20.04

```lua
#!/usr/bin/env lua
print("Hello, World!")
```

![image](https://user-images.githubusercontent.com/2205188/122587686-ced15900-d066-11eb-881c-9b0bcea652e2.png)

Scripts usually start with a shebang character sequence. The number sign (`#`) is not a comment marker in lua, but shebang lines are standard so the Lua interpreter still handles them.

Proof: https://www.lua.org/pil/1.4.html

> When the interpreter loads a file, it ignores its first line if that line starts with a number sign (`#´). That feature allows the use of Lua as a script interpreter in Unix systems.

PR #2876, opened 2021-06-19
[Lua] Fix shebang

jfcherng:
Fixes https://github.com/sublimehq/Packages/issues/2875

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -3,6 +3,10 @@
 file_extensions:
   - lua
 scope: source.lua
+first_line_match: |-
+  (?x:
+    ^\#! .* \b(lua)\b # shebang
+  )
 variables:
   reserved_word: |-
     (?x:(?:
@@ -45,7 +49,8 @@ variables:
 
 contexts:
   main:
-    - include: statements
+    - match: ''
+      push: [statements, shebang]
 
   else-pop:
     - match: (?=\S)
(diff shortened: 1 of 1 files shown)


Lua/Lua.sublime-syntax
```diff
@@ -82,6 +87,20 @@ contexts:
     - include: end
     - include: statements
 
+  shebang:
+    - match: ^\#!
+      scope: punctuation.definition.comment.lua
+      set: shebang-body
+    - match: ''
```
deathaxe, line 94: Immediatelly popping will cause issues once we decide to embed LUA into Markdown for instance.

grafik

That's why Bash uses

```
    - match: ^|(?=\S)
      pop: true
```

Lua/Lua.sublime-syntax
```diff
@@ -82,6 +87,20 @@ contexts:
...
+      pop: 1
```
deathaxe, line 95: Not a big deal, but I'd vote to not mix pop styles a document. Either `pop: true` for all or `pop: 1`. We might switch to latter one once syntax version 2 is applied.

Lua/Lua.sublime-syntax
```diff
@@ -82,6 +87,20 @@ contexts:
...
+
+  shebang-body:
+    - meta_scope: comment.line.shebang.lua
+    - match: \b(lua)\b
```
deathaxe, line 99: No capture group is needed here, so parens can be removed.

Lua/Lua.sublime-syntax
jfcherng: Interesting knowing!

Lua/Lua.sublime-syntax
jfcherng: There is no other Lua PR so no conflict! Let's change them to all `1`s.

deathaxe (approved):

PR #2794, opened 2021-04-22
[Batch File] Rewrite Syntax

deathaxe:
Fixes #534
Fixes #1005
Fixes #1301

Removes any kind of sregex incompatible matches (#481)

What should I say, it was quite a long journey with some challanges to implement the weirdnesses of Microsoft engineers' ideas, some of them possible to tackle only with latest syntax engine features. I don't recall any detail anymore, but I guess there is nearly no line left untouched.

It follows the batch interpreter's behavior as closely as `https://ss64.com` describes and a couple of more sophisticated scripts revealed.

It got to a point when feedback is needed to find remaining issues and possible improvements.

PR #2879, opened 2021-06-22
[Batch File] match custom parameter

jingyuexing:
match like:
* -n 
* --n 
* --nnn-nnn 
* --any 
* -any1


deathaxe:
Already fixed by #2794

PR #2882, opened 2021-06-23
[ShellScript] Fix test expressions with groups

### LINKED ISSUE #2881: [BASH] nested brackets with =~ operator
Bash highlighting broken after ST4 upgrade; nested brackets in combination with the =~ operator will cause the highlighter to stop afterwards.

- Sublime Version: 4107
- OS Version: Windows 10 Pro / version 21H1 / build 19043.1055

``` bash
#!/usr/bin/env bash

issueyear=NOT_A_YEAR

if [[ ! ($issueyear =~ ^[0-9]+$) ]]; then
    echo "invalid issue year"
fi
```

grafik

this was first posted at https://github.com/sublimehq/sublime_text/issues/4518 where @jfcherng confirmed it with build 4109

deathaxe:
Fixes #2881

This PR fixes test pattern matching enclosed in parentheses.

Before this commit a group context wasn't popped as closing parenthesis was matched as part of pattern on the right hand side of the expression.

### DIFF
--- a/ShellScript/test/syntax_test_bash.sh
+++ b/ShellScript/test/syntax_test_bash.sh
@@ -3720,6 +3720,7 @@ echo ca{${x/z/t}" "{legs,f${o//a/o}d,f${o:0:1}t},r" "{tires,wh${o//a/e}ls}}
 #                ^^^^^ meta.set.regexp.shell
 #                ^ punctuation.definition.set.begin.regexp.shell
 #                    ^ punctuation.definition.set.end.regexp.shell
+#                     ^ keyword.operator.quantifier.regexp.shell
 #                      ^ punctuation.definition.group.end.regexp.shell
 #                        ^ punctuation.section.group.end.shell
 
--- a/ShellScript/test/syntax_test_bash.sh
+++ b/ShellScript/test/syntax_test_bash.sh
@@ -3705,11 +3705,11 @@ echo ca{${x/z/t}" "{legs,f${o//a/o}d,f${o:0:1}t},r" "{tires,wh${o//a/e}ls}}
 #                        ^ punctuation.definition.group.begin.regexp.shell
 #                          ^ punctuation.definition.group.end.regexp.shell
 
-[[ ! ($line =~ ^([0-9]+)$) ]]
+[[ ! ($line =~ ^(\([0-9]+)$) ]]
 #^^^^ meta.conditional.shell - meta.group
-#    ^^^^^^^^^^^^^^^^^^^^^ meta.conditional.shell meta.group.shell
-#                         ^^^ meta.conditional.shell
-#                            ^ - meta.conditional
+#    ^^^^^^^^^^^^^^^^^^^^^^^ meta.conditional.shell meta.group.shell
+#                           ^^^ meta.conditional.shell
+#                              ^ - meta.conditional
 #  ^ keyword.operator.logical.shell
 #    ^ punctuation.section.group.begin.shell
 #     ^^^^^ variable.other.readwrite.shell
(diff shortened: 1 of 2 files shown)


ShellScript/test/syntax_test_bash.sh
```diff
@@ -3676,8 +3705,29 @@ echo ca{${x/z/t}" "{legs,f${o//a/o}d,f${o:0:1}t},r" "{tires,wh${o//a/e}ls}}
 #                        ^ punctuation.definition.group.begin.regexp.shell
 #                          ^ punctuation.definition.group.end.regexp.shell
 
+[[ ! ($line =~ ^(\([0-9]+)$) ]]
+#^^^^ meta.conditional.shell - meta.group
+#    ^^^^^^^^^^^^^^^^^^^^^^^ meta.conditional.shell meta.group.shell
```
BenjaminSchaaf, line 3710: I think these should be properly sorted like the rest of the assertions. That goes for all of the tests added.

deathaxe: Test patterns are correctly sorted and grouped by meta-tests and normal-token-tests. The position of the whole test case among the others is well choosen as well.

Speaking as the author of the whole file here.

wbond: I think Ben is implying that line 3710 should be below 3713 and 3714. Similarly 3711-3712 should be down after 3726.

deathaxe: Yes, I understand, but exactly this is against the scheme of grouping meta vs. other scopes.

I'd argue the other way round to move those `meta.pattern` etc. tests should go up to be well grouped.

I wouldn't want to give up grouping meta tests vs. normal tests as it results in huge test lines hard to maintain. Especially when checking boundaries which often requires negative selectors.

As a sidenote: 

1. This rule was just copied from the above one and modified
2. I haven't authored all of those rules from scratch during last rewrite. Many existed already and just have been extended. While already having reorganized a lot of them, I found it too tedious to touch all of them.

Boundary checks could probably be more sophisticated here.

wbond: I don't feel one way or the other about the grouping, just trying to see if I could help interpret, in case there was a different understanding of the issue.

deathaxe: I mean, we can also start to rework all test files of all packages to bring them into a defined scheme/order.

wbond: Personally, I don't think that would be a good use of time 😄

deathaxe: I am not opposed to define and follow some kind of guideline.

Those `meta.pattern` tests represent regexp-strings more or less, rather than structureal meta scopes like groups, blocks, sections, etc. Hence they are in the lower part as well as all the `meta.number` and `meta.string` tests are.

Now that I think about it, those `meta.pattern` scopes could also be changed into `meta.string.pattern` including some kind of string scope, probably. At least some other syntaxes apply string scopes onto patterns.

deathaxe:
I've added a commit to illustrate the intent behind splitting structural meta scope tests and normal scope tests and why it might be useful with regards to test pattern complexity.

ShellScript/test/syntax_test_bash.sh
BenjaminSchaaf: Seems it's already pretty inconsistent when `meta.*` checks are separated from others in that test file, so I don't think it matters at this point. We should have a standard, ideally with an auto-formatter/linter but that's a separate issue.

BenjaminSchaaf (approved):

jfcherng (approved):

PR #2886, opened 2021-06-27
[CSS] Add smart_typing setting

deathaxe:
This commit adds support for `smart_typing` setting to extend scope of `auto_match_enabled` to related but somewhat smarter typing features.

The intention is to satisfy both the lazy and convenient developer and the purist who needs control over every single key stroke.

Due to several requests some input helpers were introduced during last dev cycle, which intend to help keeping focus on content while helping with some nasty repetitive tasks like adding space after colons and move on to next line when hitting enter in front of semi-colons.

Now that those smart behaviors exist, of course, some people complain about it.

With this commit following features are enabled only if `smart_typing` is `true`:

1. Automatically add space after colons.
2. Hitting enter in front of semicolon moves to next line.
3. Hitting space in between brackets adds space to left and right position of caret.

Note:

The `smart_typing` setting is NOT added to CSS.sublime-settings as it is intended to be used for related features in all syntax packages, just like `auto_match_enabled`.

Adding it to CSS would override user defined global settings in Preferences.sublime-settings.

As a result all those smart typing features are disabled by default in case this setting is NOT added to Preferences.sublime-settings of the Default.sublime-package or if it is decided to disable such feature by default.

Note: Alternatively just remove adding space by #2912.

### DIFF
--- a/CSS/Default.sublime-keymap
+++ b/CSS/Default.sublime-keymap
@@ -1,9 +1,34 @@
 [
+	// Add a colon and semicolon after a property name,
+	// if the next following token doesn't look like a value.
+	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ":$0;"}, "context":
+		[
+			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{ "key": "setting.smart_typing", "operator": "equal", "operand": false },
+			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
+			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\w$", "match_all": false },
+			{ "key": "following_text", "operator": "regex_contains", "operand": "^\\s*($|[^:;}]+:)", "match_all": false }
+		]
+	},
+	// Add a colon after a property name,
+	// if the next character is a semicolon or the following token looks like a value.
+	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ":$0"}, "context":
+		[
+			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{ "key": "setting.smart_typing", "operator": "equal", "operand": false },
+			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
+			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
+			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\w$", "match_all": false },
+			{ "key": "following_text", "operator": "regex_contains", "operand": "^;|^[^\\s:;][^:]*($|[;}]|//|/\\*)", "match_all": true }
+		]
+	},
 	// Add a colon followed by space and semicolon after a property name,
 	// if the next following token doesn't look like a value.
 	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ": $0;"}, "context":
 		[
 			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{ "key": "setting.smart_typing", "operator": "equal", "operand": true },
 			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
 			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
 			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\w$", "match_all": false },
@@ -15,6 +40,7 @@
 	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ": $0"}, "context":
 		[
 			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{ "key": "setting.smart_typing", "operator": "equal", "operand": true },
 			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
 			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
 			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\w$", "match_all": false },
@@ -44,7 +70,7 @@
 	// This is to not bother with semicolons even though completions end up with the caret directy in front of it.
 	{ "keys": ["enter"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Add Line.sublime-macro"} , "context":
 		[
-			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{ "key": "setting.smart_typing", "operator": "equal", "operand": true },
 			{ "key": "auto_complete_visible", "operator": "equal", "operand": false },
 			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
 			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
@@ -56,7 +82,7 @@
 	// Expand {|} to { | } when space is pressed
 	{ "keys": [" "], "command": "insert_snippet", "args": {"contents": " $0 "}, "context":
 		[
-			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{ "key": "setting.smart_typing", "operator": "equal", "operand": true },
 			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
 			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
 			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\{$", "match_all": true },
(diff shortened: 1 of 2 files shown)


deathaxe:
@rwols So you think it's better to advice people to hack around with shipped plugin code leading to other issues in future once those builtin sources move on with new changes? Not to say it is very inconvinient for those who just want to use an editor.

see: 
- https://forum.sublimetext.com/t/css-auto-space-how-to-turn-it-off/59260/6
- https://forum.sublimetext.com/t/st4-css-autocomplete-space-inconvenience-issue/58454/7

I am very happy with current behavior, but think a tool shouldn't force people to certain workflows they are not comfortable with.

Ultra-Instinct-05:
> As a result all those smart typing features are disabled by default in case this setting is NOT added to Preferences.sublime-settings of the Default.sublime-package or if it is decided to disable such feature by default.

I am all for gating these behind a setting (to keep everyone happy) but should we really be encouraging adding package defined settings in `Preferences` considering only ST core settings should be defined there ?

deathaxe:
The idea behind placing it into `Preferences` is to introduce a general setting such as `auto_match_enabled` which may be used by all packages and maybe some core bindings in future. This idea is driven by two arguments:

1. I'd find it terrible user experience if each package introduces its own setting for such features.
2. I learned about pitfalls and issues about introducing syntax-specific-only settings from MarkdownEditing package. Once several packages would want to make use of such a setting, and all of them defined a default value in a syntax-specific settings file, there's not much room (in fact none) for a user to change behavior globally or per project.

The current approach instead leaves all doors open.

A syntax-specific-only setting would make sense only, if we rename it to something very special. Maybe settings to gate each binding individually. But I wouldn't want to flood settings files with too much details.

deathaxe:
I am fine with current behavior. 

I would leave the decision about gating it behind a setting via this PR or "not adding space at all" via #2916 to @sublimehq.

keith-hall (approved):
Approving to show my support for this approach

FichteFoll:
If this is going to be accepted and applied to other languages, it should be added to the default preferences as well. Adding `core-change` label.

jwortmann:
I like the approach to make it configurable via a setting, but wouldn't the already existing "auto_complete_trailing_spaces" setting fit very well for the "space after colon" case?

```json
{
	// Add a space after completions if the auto complete engine thinks
	// they're likely enough
	"auto_complete_trailing_spaces": true,
}
```

michaelblyons:
> ```jsonc
> {
> 	// Add a space after completions if the auto complete engine thinks
> 	// they're likely enough
> 	"auto_complete_trailing_spaces": true,
> }
> ```

You can call your Markdown language `jsonc` to enable comments and trailing commas in JSON. /offtopic

deathaxe:
Is it still a "trailing" space if it is followed by a semicolon when auto completing a property?

    fon|  ->  font: |;

Does "auto_complete" fit in case of normal key bindings which automatically add the space when typing the colon? It's basically no auto completion then, is it?

One user even asked to configure those two use cases with different settings.

I like the idea in general though.

Bindings to add new line when hitting enter in front of semicolon would still need a dedicated setting such as "smart_typing" in case it should be configurable.

jwortmann:
> Is it still a "trailing" space if it is followed by a semicolon when auto completing a property?

I would say yes, because it is trailing for the left side of the cursor. The semicolon can simply be ignored, because the cursor just steps over it if you manually type the `;` after the value. The "auto_complete_trailing_spaces" addresses the preference whether you want to manually add a space when you continue typing after using autocomplete, or rather have ST insert it for you. So I think it can be assumed that this setting also applies to completions from css_completions.py.

Regarding the keybinding for `:`, I indeed wouldn't really call it "auto complete". Actually, I'm on the side of the user from the other PR #2912 and would remove the space completely in this case. When you just type the colon character, I find it quite unexpected that it will automatically insert a space to the left side of the cursor, and I ended up with double spaces as well, until I overrode this keybinding in my user settings. But the average user might not know that it is caused by a keybinding for `:` and how to manually disable it, so I can understand the complains.

This keybinding is only active with "auto_match_enabled", but in general I find it a bit strange, because would you consider the `:` and `;` really a pair, similar to quotes or brackets? If you close your eyes and type blindly, you should end up with exactly what you typed (except for <kbd>Enter</kbd> or <kbd>Tab</kbd> for confirming autocompletion, of course), and not with additional spaces, in my opinion. And this is how it always worked for all the other "auto match" bindings too.

For the newline in front of semicolons or the double-space between `/* | */`, you could use the "smart_typing" setting, or just leave it as it is now with "auto_match_enabled".

deathaxe:
> This keybinding is only active with "auto_match_enabled", but in general I find it a bit strange, because would you consider the : and ; really a pair, similar to quotes or brackets?

This setting has been choosen for the same reason as you proposing `auto_complete_trailing_spaces`. It is an existing setting which is used for all those kinds of adding pairs of characters, so it seemed suitable. If the additionally added space is ignored it behaves very much like a normal pair. Who says pairs must always be equal characters?

jwortmann:
> This setting has been choosen for the same reason as you proposing auto_complete_trailing_spaces. It is an existing setting which is used for all those kinds of adding pairs of characters, so it seemed suitable. If the additionally added space is ignored it behaves very much like a normal pair.

The difference is that "auto_complete_trailing_spaces" would exactly describe what happens, while "auto_match_enabled" is "only" the best fitting one of the existing settings, for the `: ` & `;` combo. I didn't claim to remove this setting there, my personal preference would just be to remove the additional space.

> Who says pairs must always be equal characters?

Nobody, but I think for all other pairs the characters directly belong together. In CSS properties, the colon has nothing to do with the semicolon. The colon's purpose is to be a key-value separator, and the semicolon closes this expression and separates it from the next property. The semicolon can even be omitted if it is the last property in a CSS rule.

jwortmann:
An analogue would be to have a similar keybinding for when you type the `:` after the key within Python dicts, which would then insert a space and a comma at the end of the line. This would also be pretty surprising, no?

credwards27:
Commenting in support of this change with a breaking case.

CSS pseudo-selectors are also affected by this which produces broken code unless the trailing space is removed.

**This occurs when curly braces are already present, i.e. when adding pseudo-selector to an existing selector rule.**

Given the following existing code:

```css
.parent .child {
}
```

Placing the cursor immediately after `.child` and typing `:last-child`; the out of the box behavior produces the following:

```css
.parent .child: last-child {
}
```

In that scenario, a backspace is now required to produce working code:

```css
.parent .child:last-child {
}
```

For the reason that the trailing space behavior differs depending on whether or not an opening curly brace is already present; not only do I support adding a preference to toggle the new behavior, I also consider this to be an actual bug related to CSS syntax scope detection.

deathaxe:
I can't reproduce your issue with the given example. While space is added after `color` property it is not in selectors.

Animation
e

CSS/Default.sublime-keymap
```diff
@@ -1,11 +1,24 @@
 [
+	// Add a colon and semicolon after a property name,
+	// if the next following token doesn't look like a value.
+	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ":$0;"}, "context":
```
FichteFoll, line 4: Personally, I find it questionable whether we want to match a colon with a semi-colon when `smart_typing` is disabled. We don't have a clear definition of what "smart typing" entails, but to me it includes anything that the user didn't request to insert explicitly (and that isn't covered by `auto_match_enabled`). For example, in this PR, a semi-colon is added after a property value completion iff `smart_typing` is active.

In general, I suppose we have to ask the question whether smart typing *levels* would be something we find worth implementing and whether it makes sense to differentiate between completing `:` with `:;` or `: ;`. And of course whether there are other instances of this that already exist. 
In my opinion, on and off are enough configuration by default. Any smart completion needs some getting used to and if more fine-tuned completions are desired, the user can tweak key bindings to their desire. It's not the easiest place for configuration, but at this level of granularity I think it is warranted.

Since `smart_typing` will be disabled by default, this is also something that would be backwards-incompatible if removed, as far as I understood, but we should consider whether it would be worth it.

CSS/Default.sublime-keymap
```diff
@@ -56,7 +70,7 @@
 	// Expand {|} to { | } when space is pressed
 	{ "keys": [" "], "command": "insert_snippet", "args": {"contents": " $0 "}, "context":
 		[
-			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
+			{ "key": "setting.smart_typing", "operator": "equal", "operand": true },
```
FichteFoll, line 73: What is the rationale for gating this behind `smart_typing` only? Shouldn't it also require `auto_match_enabled`?

CSS/Default.sublime-keymap
deathaxe: `auto_match_enabled` covers auto adding `begin` and `end` punctuation of sections such as strings or blocks. The value section of key-value pairs fits quite well into that definition, IMHO, if we don't intent to create typing helpers which auto add semicolons everywhere.

But you are right with `smart_typing` probably being too general to cover all possible use cases.

We might also make use of the default setting `"auto_complete_trailing_spaces": true,` to control space instead of introducing `smart_typing`.

Another question might be whether to add space in completions only, or whether to use the setting to gate key bindings as well. In former case we probably should remove adding spaces via bindings at all.

CSS/Default.sublime-keymap
deathaxe: Probably due to the assumption that someone who enables `smart_typing` would not disable `auto_match_enabled`.

The setting is just an attempt to help those who want more control over when space is added or not. I am pretty heppy with this binding in general and use it in every auto-match-scenario. 

I'd be ok with keeping it gated by `auto_match_enabled` only.

PR #2889, opened 2021-06-29
[Python] Improve auto-pairing behavior

### LINKED ISSUE #2888: [Python] Auto Pair Curly Brace in F Strings
- Sublime Version: 4109
- OS Version: macOS 11

## Expected behavior

It feels like when tying `{` in an fstring that the `}` should be automatically inserted.

## Actual behavior

No auto-pairing is done.

## Steps to reproduce

1. Type `f"variable: {"

deathaxe:
Fixes #2888

This commit improves auto-pairing of quotation marks and brackets in `source.python` scope.

To avoid interference all core defaults are overridden before creating new bindings. This is needed to prevent auto-pairing after escapes such as ( `\|` -> `\(|` ).

Auto-pairing is enabled in front of common separators `,`, `:`, `;` to improve editing experience of mappings, sets and lists. It follows behavior of JSON.

Auto-pairing parens/brackets/braces now works within strings and directly in front of quotation marks.

Several other language specific tweaks ...

Note:

A general scheme to order context keys is applied.
1. settings
2. main language selector
3. selection state
4. patterns (preceding_text, following_text)

### DIFF
--- a/Python/Default.sublime-keymap
+++ b/Python/Default.sublime-keymap
@@ -1,25 +1,110 @@
 [
-    // Auto-pair quotes even after string modifiers.
-    // Copied over from the default bindings with modifications to `preceding_text`
-    // and an added selector condition.
-    { "keys": ["\""], "command": "insert_snippet", "args": {"contents": "\"$0\""}, "context":
+    // Suppress default auto-pairing bindings to avoid unwanted interference
+    { "keys": ["\""], "command": "insert", "args": {"characters": "\""}, "context":
         [
-            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
-            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
-            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
-            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?i)\\b[bfru]+$", "match_all": true },
-            { "key": "selector", "operator": "equal", "operand": "source.python" },
-            { "key": "eol_selector", "operator": "not_equal", "operand": "string.quoted.double - punctuation.definition.string.end", "match_all": true }
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selector", "operand": "source.python" },
+            { "key": "selection_empty", "match_all": true }
         ]
     },
+    // Auto-pair double quotes
+    // - not after escape \| -> \"|
+    // - not after quote "| -> ""|
+    // - after string modifiers: f| -> f"|"
+    // - in front of dots: |.join() -> "|".join()
+    // - in front of commas: |, -> "|",
+    // - in front of colons: |: -> "|":
+    // - in front of semicolons: |; -> "|";
+    { "keys": ["\""], "command": "insert_snippet", "args": {"contents": "\"$0\""}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selector", "operand": "source.python - string.quoted.double" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?i:^|[^\"\\w\\\\]|\\b[bfru]+)$", "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|\\.|,|:|;|$)", "match_all": true }
+        ]
+    },
+
+    // Suppress default auto-pairing bindings to avoid unwanted interference
+    { "keys": ["'"], "command": "insert", "args": {"characters": "'"}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selector", "operand": "source.python" },
+            { "key": "selection_empty", "match_all": true }
+        ]
+    },
+    // Auto-pair single quotes
+    // - not after escape \| -> \"|
+    // - not after quote '| -> ''|
+    // - after string modifiers: f| -> f"|"
+    // - in front of dots: |.join() -> "|".join()
+    // - in front of commas: |, -> "|",
+    // - in front of colons: |: -> "|":
+    // - in front of semicolons: |; -> "|";
     { "keys": ["'"], "command": "insert_snippet", "args": {"contents": "'$0'"}, "context":
         [
-            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
-            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
-            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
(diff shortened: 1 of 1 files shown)


Ultra-Instinct-05:
I haven't looked too closely but does this supersede https://github.com/sublimehq/Packages/pull/2696 ? If so, I'd be more than happy to close that.

deathaxe:
Wasn't aware of this PR anymore, tbh. Yes this one tries to fix all known shortcommings with regards to auto-pairing, including interpolation. It isn't too restricted with regards to preceding escape characters though as I am a bit uncertain whether we really need it.

FichteFoll:
How much of this is Python-specific and what would you say should be added globally? I haven't looked at the PR contents yet.

deathaxe:
Quotes are python specific due to `preceding_text` matching format chars.

Maybe `following_text` of brackets may be used as global default, too. Some common punctuation which are known to separate or terminate expressions are added, which may be safe in most but maybe not all syntaxes.

Not sure if each syntax supports escaped chars, which auto-pairing is disabled for by `preceding_text` patterns. Several syntaxes do so though. In fact, some overrides for default bindings are added to make sure auto-pairing is disabled correctly.

While many syntaxes may share majority of those context patterns they might need to differ in detail to handle language specific edge cases well.

Someone would need to write those rules for all syntaxes and then check for commons, if we want to answer that for sure.

keith-hall:
It would be great if a variable from a tmPreferences file could be referenced in the keybinding contexts, then each syntax could set a variable in each relevant scope and one keybinding could automatically work for all syntaxes (which define this variable, or fall back to the default value if they don't) - I think maintenence could become much easier this way...

deathaxe:
You would probably need two variables per auto-pairing binding. One for `preceding_text` and one for `following_text`. Theoretically each language can define its own arbritary set of auto-pairing characters. Not sure how this should map well to tmPreferences files.

MarkdownEditing uses such a [Bold and Italic Markers.tmPReferences](https://github.com/SublimeText-Markdown/MarkdownEditing/blob/master/Bold%20and%20Italic%20Markers.tmPreferences) file for customizing snippets. I've removed it because it is too cumbersome to both, customize and use it.

Python/Default.sublime-keymap
```diff
@@ -1,25 +1,110 @@
 [
-    // Auto-pair quotes even after string modifiers.
-    // Copied over from the default bindings with modifications to `preceding_text`
-    // and an added selector condition.
+    // Suppress default auto-pairing bindings to avoid unwanted interference
+    { "keys": ["\""], "command": "insert", "args": {"characters": "\""}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selector", "operand": "source.python" },
```
jwortmann, line 6: This keybinding also disables the "step over closing quote" functionality from the default keybindings. You would need to exclude the `punctuation.definition.string.end` scope from the selector to prevent this, similar to #3204.

Python/Default.sublime-keymap
```diff
@@ -1,25 +1,110 @@
...
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?i:^|[^\"\\w\\\\]|\\b[bfru]+)$", "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|\\.|,|:|;|$)", "match_all": true }
+        ]
+    },
+
+    // Suppress default auto-pairing bindings to avoid unwanted interference
+    { "keys": ["'"], "command": "insert", "args": {"characters": "'"}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selector", "operand": "source.python" },
```
jwortmann, line 32: Same here.

jwortmann (changes requested):

Python/Default.sublime-keymap
deathaxe: Thanks for the head up.

jrappen (approved):

Python/Default.sublime-keymap
```diff
@@ -1,25 +1,110 @@
...
-            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
-            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|>|$)", "match_all": true },
-            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?i)\\b[bfru]+$", "match_all": true },
-            { "key": "selector", "operator": "equal", "operand": "source.python" },
-            { "key": "eol_selector", "operator": "not_equal", "operand": "string.quoted.double - punctuation.definition.string.end", "match_all": true }
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selector", "operand": "source.python - string.quoted.double" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?i:^|[^\"\\w\\\\]|\\b[bfru]+)$", "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\t| |\\)|]|\\}|\\.|,|:|;|$)", "match_all": true }
```
FichteFoll, line 24: Could we reasonably change this to match for any non-alphanumeric, i.e. `^(?!\w)`? While more general, that would simplify the regex a lot and I cannot reasonably think of a situation currently where this would be too eager. Same for single quotes, of course. Maybe even parens, brackets, and braces?

A similar thing is already done for `preceding_text`.

I would generally be happier if the non-Python-specific bindings (i.e. the non-string ones) would go into the global key bindings, but it doesn't seem like there is much interest for it. I shall try adding that to my config instead and see how it goes.

deathaxe: As literly everything can be content of a quoted string, I am not convinced of such a common pattern. I am concerned about it producing too many false positives.

FichteFoll (approved):
I would be happier if we managed to get some of the more general ones (like auto-matching of various brackets in front of closing string punctuation which would be very useful in every language that uses `{}` for string interpolation), but that train appears to be way too slow, so I'm fine with going for changes to individual packages.

We can always remove the duplicate bindings should core catch up at some point.

michaelblyons (approved):

Python/Default.sublime-keymap
```diff
@@ -1,25 +1,110 @@
 [
-    // Auto-pair quotes even after string modifiers.
-    // Copied over from the default bindings with modifications to `preceding_text`
-    // and an added selector condition.
+    // Suppress default auto-pairing bindings to avoid unwanted interference
+    { "keys": ["\""], "command": "insert", "args": {"characters": "\""}, "context":
+        [
+            { "key": "setting.auto_match_enabled" },
+            { "key": "selector", "operand": "source.python - string.quoted.double" },
```
deathaxe, line 6: I am slightly uncertain about `- string.quoted` because it disables auto-pairing double quotes in triple-quoted documentation comments or multi-line strings.

deathaxe: Multiline strings may require special treatment to support auto-pairing of quotes within, but this may be part of another PR.

PR #2912, opened 2021-07-23
[CSS] Don't add space after colons automatically

deathaxe:
This PR  is an alternative for #2886 and just alters logic to not add space after colons in hardcoded way without any setting controling this behavior.

Choose whatever is prefered.

### DIFF
--- a/CSS/Default.sublime-keymap
+++ b/CSS/Default.sublime-keymap
@@ -5,7 +5,7 @@
 		[
 			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
 			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
-			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
+			{ "key": "selector", "operator": "equal", "operand": "source.css meta.property-list - meta.selector - meta.group - comment - string", "match_all": true },
 			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\w$", "match_all": false },
 			{ "key": "following_text", "operator": "regex_contains", "operand": "^\\s*($|[^:;}]+:)", "match_all": false }
 		]
--- a/CSS/Default.sublime-keymap
+++ b/CSS/Default.sublime-keymap
@@ -1,24 +1,13 @@
 [
 	// Add a colon followed by space and semicolon after a property name,
 	// if the next following token doesn't look like a value.
-	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ": $0;"}, "context":
+	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ":$0;"}, "context":
 		[
 			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
 			{ "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
 			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
 			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\w$", "match_all": false },
 			{ "key": "following_text", "operator": "regex_contains", "operand": "^\\s*($|[^:;}]+:)", "match_all": false }
-		]
-	},
-	// Add a colon followed by space after a property name,
-	// if the next character is a semicolon or the following token looks like a value.
-	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ": $0"}, "context":
-		[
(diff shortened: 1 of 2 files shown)


Ultra-Instinct-05:
I much prefer the `smart_typing` setting. These smart features in CSS is helpful (plus with the settings way, everyone is happy).

bhdzllr:
For me the main problem is the automatic space after I fully typing a CSS property. It's ok for autocompletion, but after I typed the full property followed by the colon, I don't want autocompletion (because I already typed the full property/keyword), just the semicolon is fine.

Examples with Property "position" (| shows the cursor):

1) Typing "pos" hitting enter/tab gives me "position: |;" which is fine
2) Typing "position" and then typing the colon gives me "position: |;", but what I want is this without the space"position:|;".
I think the semicolon in the second case is fine, because if you type it again, there is no double semicolon but Sublime Text just tabs over it.

deathaxe:
The intention is te create consistent result without regards to what part of an expression already existed. Current implementation is smart enough to not add duplicated punctuation or spaces (as other solutions in other plugins do).

I prefer #2886 instead of this PR to provide some settings for users to choose whether the smartness is wanted or not.

If we want to start adding settings for every single detail of how any function woks, we'll quickly end up in hundreds of thousands of settings.

It makes absolutely no sense to have different behavior between typing `pos|` hitting tab or typing `position:|`. Either both are to end up in `position: |;`  or `position:|;`  (space or not).

Either you want the space between a colon and the value or not.

bhdzllr:
> Either you want the space between a colon and the value or not.

I agree, but there is a difference between using autocompletion intentional or typing the full declaration by myself.

Maybe the problem is the behaviour of autocompletion and this is the wrong pull request to discuss this and a new issues should be created.

I tested this behaviour in Visual Studio Code and IntelliJ IDEA.
In both IDEs if the users types the full property followed by the colon, autocompletion does not step in, because the work is already done. I think Sublime Text is on the wrong path in this case.

When typing `pos|` and hitting tab/enter the user chooses to use autocompletion.
But when typing `position:|` the user already did the extra work and typed everything also the colon by himself and will also type the space himself. If autocompletion steps in, the user ends up with two spaces between the colon and the value.

I myself currently have the problem that I always have to delete a space because I often end up with two spaces because I often type the whole property.

My objection is that after a user has typed the colon, autocompletion should not step in, because the user already typed everything by himself. Automatically inserting the semicolon is a nice improvement and don't mind, because Sublime Text is smart enough to type over it when a user types it again and the user ends up with just one semicolon.

deathaxe:
> I tested this behaviour in Visual Studio Code and IntelliJ IDEA.
In both IDEs if the users types the full property followed by the colon, autocompletion does not step in, because the work is already done. I think Sublime Text is on the wrong path in this case.

I think it is the wrong path to always step behind and follow others. Innovation comes from finding new paths. ST did so and many other popular editors copied (stole) those concepts.

It is not ST's auto-completion, which causes a space to be added when typing colon. It's done very much the same way as you described. 

It's "auto-pairing" which steps in at this point. It's the same mechanism, which also causes a closing parentheses to be added when typing an opening one.

see: https://github.com/sublimehq/Packages/blob/43081b3494a3d743b9aeb4b60e971b3f1f5528eb/CSS/Default.sublime-keymap#L2-L23

So if we auto-add a semicolon, why not also the space, if it would alsoe be added by auto-completion otherwise?

> But when typing `position:|` the user already did the extra work and typed everything also the colon by himself and will also type the space himself.

This would also argument for not add the semicolon, in which case you could simply disable `auto_match_enabled` setting.

If space is really an issue it was easy to just override the key binding liked above.

bhdzllr:
Thank you for pointing this out, I didn't know about Keymap.

> This would also argument for not add the semicolon, in which case you could simply disable auto_match_enabled setting.

I like Sublime Text auto match, just the space after the colon bugs me after i typed the colon myself. I like it when using autocomplete. In older versions of Sublime Text there was no space after typing the colon.

> If space is really an issue it was easy to just override the key binding liked above.

Thank you, did this for myself.

I also tested a little bit if I am able to find a key map that prevents double spaces and keeps the current behaviour, but the only solution I came up with was this one after the first two rules and I think it's clumsy:

```JavaScript
{ "keys": [" "], "command": "insert_snippet", "args": {"contents": ""}, "context":
		[
			{ "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
			{ "key": "selector", "operator": "equal", "operand": "source.css - meta.selector.css", "match_all": true },
			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\w:\\s", "match_all": false },
			{ "key": "following_text", "operator": "regex_contains", "operand": "^;", "match_all": false }
		]
	},
```

Anyways, for me it's ok to override the keybindings in my user preferences.

PR #2915, opened 2021-07-26
[PHP] Fix shebang is not highlighted

jfcherng:
Similar to https://github.com/sublimehq/Packages/pull/2876 but PHP already detects the shebang. This PR add scopes for the shebang line.

### DIFF
--- a/PHP/PHP.sublime-syntax
+++ b/PHP/PHP.sublime-syntax
@@ -16,6 +16,10 @@ scope: embedding.php
 contexts:
   main:
     - match: ''
+      push: [real-main, shebang]
+
+  real-main:
+    - match: ''
       push: scope:text.html.basic
       with_prototype:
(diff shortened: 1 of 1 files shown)


deathaxe:
Wouldn't this conflict with #2789?

jfcherng:
> Wouldn't this conflict with #2789?

True. But I doubt that one will be merged (any time soon).

deathaxe:
The main scope is just one little aspect of an overall large improvement. The proposed solution is more consistent with regards to the other default syntaxes but the easiest one to change back, if really in doupt. 

It doesn't help to flood the repo with possibly conflicting PRs though.

There's also #2760 which proposes a common scheme for `first_line_match`, which would also be overwritten to the bad by merging this PR.

jfcherng:
Those look like there must be some conflictions anyway. 🤔

deathaxe (approved):

Issue #1434, opened 2018-02-16
update snakemake/.Snakefile/.snakefile/.smk syntax highlighting in Python


scottkall:
Hi, I'm not sure exactly how to do the latter or I would -- could someone:

1. add lowercase `.snakemake` and `.smk` to [this document](https://github.com/sublimehq/Packages/blob/1602e19b6ae3b2c031fda6f4c7d8c32c560b8bbf/Python/Python.sublime-syntax) and then 

2. find a way so that triple-quoted shell commands in Snakefiles render as something other than a gray Python comment? 

Example in Sublime Text with sublimehq (current clone):

![pipe_smk](https://user-images.githubusercontent.com/3195509/36324600-f4e73a4c-1322-11e8-8bb4-84a892e2cfff.png)

versus the ATOM editor (1.24.0 x64):

![pipe_smk_atom](https://user-images.githubusercontent.com/3195509/36324650-2dfcd9fe-1323-11e8-8285-dee6e2b607d9.png)

I'm on Mac OS 10.13.3 and Sublime Text 3.0 build 3143.

Really, really appreciated if you can.

Tagging @wbond @karel-brinda @jskinner @guillermooo, who contributed to the Python.sublime-syntax doc.

Appendix: [details on Snakemake](https://snakemake.readthedocs.io/en/stable/)

FichteFoll:
Trying to get the python syntax to work for this special format would be impractical. It seems like you should be better off with the bash highlighting, for now at least. Ideally you would have a custom syntax for this format that could for example include the shell syntax for the quoted shell block and add a prototype match for custom variable interpolation.

karel-brinda:
@FichteFoll Snakemake is a domain-specific language built on top of Python. It would be nice to make bash highlighting work (for `shell:`), but support for highlighting Python code has, in my opinion, a higher priority.

The issue here is that `"""..."""` after `shell:` never corresponds to a doc string, but to a string carring some code. Therefore, it should not become almost invisible.

Maybe would it possible to check what's the preceding work, and if it `shell:`, don't interpret it a doc string?

karel-brinda:
This is the Snakemake grammar: http://snakemake.readthedocs.io/en/latest/snakefiles/writing_snakefiles.html. Unfortunately, the page is somehow old and doesn't include newer keywords (e.g., `conda:` or `singularity:`).

johanneskoester:
Oh yeah, I should update the grammar. Thanks for pointing this out!

[...]

PR #2916, opened 2021-07-26
[Python] Add support for Snakemake modules

karel-brinda:
Open files with the .smk and .snakefile extensions with Python highlighting

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -21,6 +21,8 @@ file_extensions:
   - gyp
   - gypi
   - Snakefile
+  - snakefile
+  - smk
   - vpy
   - wscript
   - bazel


keith-hall:
there's a discussion at https://github.com/sublimehq/Packages/issues/1434 which suggests this may not be the best idea

jfcherng:
What's the uppercased `Snakefile` at line 23? 🤔

karel-brinda:
I think there have been some changes in the best practices for Snakemake since the time of the discussion in #1434. `Snakefile` is the most common spelling for the main Snakemake file of individual workflows (~ `Makefile` for make). On systems that don't distinguish between cases, people sometimes use `snakefile`. The main workflow files often include other files, which usually have the suffix `.smk` (currently the official extension for Snakemake files to be included (~ `.mk` for Make);  see https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html for more details).

jfcherng (approved):
Since there are both `Makefile` and `makefile` in the syntax definition, I am fine with this PR.
https://github.com/sublimehq/Packages/blob/c098e8568f02322e2505d1ca30b31a095551019e/Makefile/Makefile.sublime-syntax#L15-L16

jfcherng (dismissed):

deathaxe:
Does case really matter? I am on Windows which is case-insensitive OS, so case sensitive extensions are not required.  Does it behave differently on Linux/MacOS?

My impression always has been those case sensitive extensions to be TextMate relics.

keith-hall:
Indeed, I tested on Linux which has a case sensitive file system, and ST finds the correct syntax definition even if the file has different casing. So it only makes sense to use "correct" case for default extension purposes imo

jfcherng:
Then I guess we could have another PR to remove `makefile` and close this PR?

deathaxe:
Despite all discussions about possible issues. As snakefile is already in the list of file extensions, adding `smk` only would probably be just consequent.

Another thing to consider for a future PR is to make use of `hidden_file_extensions` to prevent displaying `snakefile` as valid file extension in safe as dialogs.

jfcherng:
@karel-brinda Would you mind sharing under what circumstance the existing `Snakefile` doesn't work for you?

karel-brinda:
In my case, `Snakefile` always works. However, when I looked more closely on the configuration files for Make to update the Snakemake configuration, I realized that I had originally forgotten to include other cases while Make has them.

deathaxe:
Those duplicates in Makefile and other syntaxes seem to be relics from TextMate, which are no longer needed. We may want to go through them and tidy them up.

Python/Python.sublime-syntax
```diff
@@ -21,6 +21,8 @@ file_extensions:
   - gyp
   - gypi
   - Snakefile
+  - snakefile
```
jfcherng, line 24: > In my case, `Snakefile` always works. However, when I looked more closely on the configuration files for Make to update the Snakemake configuration, I realized that I had originally forgotten to include other cases while Make has them.

I looks like we prefer removing the lowercased `makefile` from the `Makefile` syntax rather than adding lowercase `snakemake` in this PR.

jfcherng: @karel-brinda The lowercased `makefile` extension has been removed from the syntax. If you can remove `snakefile` from this PR, it's good to go.

karel-brinda:
Should I remove also other case duplicates, such as `Sconstruct` vs. `SConstruct`?

PR #2922, opened 2021-07-29
[Markdown] Highlight task check boxes in block quotes

deathaxe:
This commit adds highlighting for `[X]` check marks in block quoted GFM tasks.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -410,11 +410,12 @@ contexts:
               push:
                 - meta_content_scope: markup.list.numbered.markdown meta.paragraph.list.markdown
                 - include: list-content
-            - match: ([ ]{,3})([*+-])(\s)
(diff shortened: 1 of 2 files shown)


deathaxe:
While this PR primarily intents to add existing syntax highlighting for checkmarks to block quotes, I'd prefer to apply `markup.checkbox` instead of `constant.language.checkbox` for all checkboxes in order to prevent confusion with embedded code.

Markdown elements probably shouldn't use scopes which are used for ordinary code, but rather rely on `markup` or `entity.name` such as HTML/XML.

We can modify those scopes in this PR or leave it to another one if we want to go this way.

An illustration can be reviewed at https://github.com/SublimeText-Markdown/MarkdownEditing/pull/607/commits/dcf0ec702ed6dcf6556236dc35e2aba8ca57f578

keith-hall (approved):

PR #2926, opened 2021-07-30
[JavaScript] Refactor arrow function branching logic.

Thom1729:
De-duplicate the arrow function branching logic. It was originally duplicated for several reasons: chiefly, sublimehq/sublime_text#3494, but also because there were issues with setting from a branch context. Those latter issues have been fixed.

The basic strategy is to add an extra layer of abstraction. Whenever we see either an open paren or an identifier character in `expression-begin`, we branch into a new `branch-possible-arrow-function` context which then checks for each of those cases individually and sets a different context for each. But since both cases are “under” the same branch point, we can re-use the contexts that contain failure rules.

This requires a slight modification for TypeScript. Only arrow functions with parenthesized argument lists may have return type annotations. Since there are no longer separate arrow-checking contexts for bare and parenthesized arrow functions, we push an extra context that handles them when we see an open paren in `branch-possible-arrow-function`.

This is all slightly simpler for existing behavior, but it also means we won't need to add a *third* copy of the arrow function logic for #2923.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -44,6 +44,7 @@ variables:
 
   non_reserved_identifier: (?:(?!{{reserved_word}}){{identifier_name}})
 
+  possible_arrow_function_begin: (?:\(|{{identifier_start}})
   either_func_lookahead: (?:{{func_lookahead}}|{{arrow_func_lookahead}})
   binding_pattern_lookahead: (?:{{identifier_name}}|\[|\{)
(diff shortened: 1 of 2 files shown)


deathaxe:
Does it mean to be an alternative implamentation for #2923?

Thom1729:
No, #2923 already includes this.

deathaxe (approved):

deathaxe:
Closing in favor of #2923, which includes this one already.

PR #2938, opened 2021-08-07
[Rust] fix writeln! macro with one argument, semicolon after where

### LINKED ISSUE #2786: [Rust] code following `where` is not syntax highlighted
- Sublime Version: Version 3.2.2, Build 3211
- OS Version: Arch Linux

```rust
pub trait Foo {
    fn bar()
    where
        Self: Sized;
}

fn main() {
	
}
```

image

keith-hall:
Fixes #2786 as I'm not sure if/when https://github.com/sublimehq/Packages/pull/2305 will ever see any traction.

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -193,6 +193,7 @@ contexts:
             - include: format-raw-string
             - match: '(?=\S)'
               set: group-tail
+        - include: group-tail
 
     - match: '\b[[:lower:]_][[:lower:][:digit:]_]*!(?=\s*(\(|\{|\[))'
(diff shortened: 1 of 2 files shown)


deathaxe:
Yeah, first complain about Rust community not willing to cooperate and now contribution pending for ever.

keith-hall:
Not convinced I was complaining, but incremental improvements may be the way to go

deathaxe:
Will complained about Rust community not being very cooperative on discord month ago, which was the reason for #2305 being opened.

deathaxe:
Superseded by #2305

PR #2939, opened 2021-08-08
[Python] use branching to detect set vs dict and for tuples vs expression

keith-hall:
This allows us to remove the complicated lookaheads / `simple_expression` variable (which have terrible performance in Oniguruma, for non-ST use cases).

To keep the existing approach of detecting set vs dict by/at the first comma, I set away from the context with the `fail` instruction when we reach the comma or colon, which allows us to "succeed" the branch and keep the existing invalid highlighting hints when there's a colon in a set or a comma without a colon in a dict etc.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -929,9 +929,6 @@ contexts:
         1: punctuation.section.sequence.begin.python
         2: punctuation.section.sequence.end.python
       push: after-expression
-    - match: \((?=\s*\*{{path}})
-      scope: punctuation.section.sequence.begin.python
-      push: inside-tuple
     - match: (?=\()
       branch_point: generators-groups-and-tuples
       branch:
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1004,34 +1004,21 @@ contexts:
         1: punctuation.section.mapping.begin.python
         2: punctuation.section.mapping.end.python
       push: after-expression
-    - match: \{(?={{simple_expression}}:|\s*\*\*)
-      scope: punctuation.section.mapping.begin.python
-      push: inside-dictionary
-    - match: \{(?={{simple_expression}}[,}]|\s*\*)
-      scope: punctuation.section.set.begin.python
-      push: inside-set
-    # If the expression is "more complex" or on the next line,
-    # fall back to default and determine later.
+    - match: (?=\{)
+      branch_point: dictionaries-and-sets
+      branch:
+        - try-set
+        - try-dictionary
+
+  try-set:
     - match: \{
-      scope: punctuation.section.mapping-or-set.begin.python
-      push:
-        - meta_scope: meta.mapping-or-set.python
-        - match: \}
-          scope: punctuation.section.mapping-or-set.end.python
-          set: after-expression
-        - match: (?={{simple_expression}}:|\s*\*\*)
-          set: inside-dictionary
-        - match: (?={{simple_expression}}[,}]|\s*\*)
-          set: inside-set
-        - match: ','
-          scope: punctuation.separator.set.python
-          set: inside-set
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -956,11 +945,30 @@ contexts:
         1: punctuation.section.sequence.begin.python
         2: punctuation.section.sequence.end.python
       push: after-expression
-    - match: \((?={{simple_expression}},|\s*\*{{path}})
+    - match: \((?=\s*\*{{path}})
```
deathaxe, line 948: Can someone elaborate on `{{path}}`? Removing this whole rule doesn't break any tests. Includes use their own contexts, so I am not sure what this pattern is finally used for. May it be removed?

Python/Python.sublime-syntax
```diff
@@ -1004,34 +1014,21 @@ contexts:
...
-      scope: punctuation.section.mapping.begin.python
-      push: inside-dictionary
-    - match: \{(?={{simple_expression}}[,}]|\s*\*)
+    - match: (?=\{)
+      branch_point: dictionaries-and-sets
+      branch:
+        - try-set
+        - try-dictionary
+
+  try-set:
```
deathaxe, line 1023: - `try-...` feels like a try-catch statement related context.
- Naming of branch contexts should be consistent.

deathaxe (changes requested):

deathaxe:
Couldn't resist to add my 50ct at https://github.com/forkeith/Packages/pull/2 ;-)

Python/Python.sublime-syntax
keith-hall: I'm also not sure what it is used for. As it breaks no syntax tests to remove it, maybe let's do so. We can always reintroduce it later with an appropriate test assertion if it turns out to be necessary.

deathaxe: I am under the impression it had to do with import statements somehow. Those have been using their own lists for a while now.

keith-hall: It was introduced in https://github.com/sublimehq/Packages/pull/999/files#diff-3038ea28e1ccdd590636eb111ab962d610e4e9cae9498f1dd9a852dc347a2d55R890 - dict and set had the same `{{path}}` logic which got relaxed in https://github.com/sublimehq/Packages/pull/2174/files

deathaxe: As imports already used their own lists at this point I can't imagine any use case in which a path would point to a tuple rather than a group. Anyway, as a `,` being present within parentheses is used to indicate a tuple, which also works in case it appears on a subsequent line, I don't think this path-heuristic is still useful.

I just think of old format strings such has `"foo: %s" % (var,)` or even tuple initializations such as `var = (1,)`. Python only expects a comma to be present to recognize them as tuple.

The only one who can tell for sure is @FichteFoll as he added this heuristics, but I don't think we need it any longer.

FichteFoll: I don't think we need it either. I know I added it 4 years ago, but I couldn't tell you why. 

Syntatically, the only thing that matters is whether you have a single- or a double-star expansion, since that distinguishes the set or dict literal. Requiring a `{{path}}` to match there is unnecessarily restrictive.

FichteFoll: Actually, this is purely theoretical and irrelevant to begin with since you cannot start a tuple with a starred expression.

```
In [1]: (*range(1, 5))
  File "<ipython-input-1-b52559586778>", line 1
    (*range(1, 5))
     ^
SyntaxError: can't use starred expression here
```

keith-hall: removed it :)

deathaxe (approved):

jfcherng (approved):

PR #2941, opened 2021-08-08
[C/ObjC/C++/ObjC++] improved syntax, typedef detection improvement

dannftk:
`typedef`s of `struct|class|enum|union` are detected wiser, target defined types are recognized as types rather than nothing

Changes are applied to C, C++, ObjC, ObjC++ syntaxes

### DIFF
--- a/C++/syntax_test_c.c
+++ b/C++/syntax_test_c.c
@@ -294,6 +294,14 @@ typedef struct mystruct {
 } mystruct;
 /* ^ entity.name.type.typedef.c */
 
+typedef struct MyStructure {} MyStructure_t;
+/* <- keyword.declaration.type.c */
+/*      ^^^^^^ keyword.declaration.struct.c */
+/*             ^^^^^^^^^^^ entity.name.struct.c */
+/*                         ^ punctuation.section.block.begin.c */
+/*                          ^ punctuation.section.block.end.c */
+/*                            ^^^^^^^^^^^^^ entity.name.type.typedef.c */
+
 /////////////////////////////////////////////
 // Data structures and return values
 /////////////////////////////////////////////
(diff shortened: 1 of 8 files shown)


dannftk:
Is anyone to react on the PR?

rchl:
When the ST developer has time, he will come and merge multiple PRs in one swipe.
It's not uncommon for PRs to hang for months before that happens though so be patient.

In the meantime, if you want to use your changes, your best bet is too have local override for that syntax.

dannftk:
How to request someone to run CI on the PR?

C++/syntax_test_c.c
```diff
@@ -1028,3 +1029,11 @@ int control_keywords()
   return 123;
   /* <- keyword.control.flow.return */
 }
+
+typedef struct MyStructure {} MyStructure_t;
```
deathaxe, line 1033: I'd try to keep test cases logically grouped. So moving this one before or after the one in line 290 would be suitable.

C++/syntax_test_cpp.cpp
```diff
@@ -2753,3 +2753,11 @@ void sayHi()
 /**
       *
 /*    ^ comment.block.c punctuation.definition.comment.c */
+
+typedef struct MyStructure {} MyStructure_t;
```
deathaxe, line 2757: The same here. There's a test case at line 410, which this one could moved to for better grouping. The current one at 410 could also be tweaked to test your change.

deathaxe (changes requested):
The intention of my request is to keep fragmentation of test cases low. Maybe already too late in some syntaxes, I know.

Otherwise I find it a simple but effective improvement.

deathaxe (dismissed):

dannftk:
@deathaxe, I've done everything how you asked me to do

deathaxe (approved):

keith-hall (approved):

PR #2945, opened 2021-08-10
[Python] Fix import statements

### LINKED ISSUE #2944: [python] syntax highlight problem
in version Build 4113
there's a syntax highlight problem for imports although this kind of importing works 

Screen Shot 1400-05-19 at 15 23 05

deathaxe:
Fixes #2944

This PR implements import statement according to https://docs.python.org/3.8/reference/grammar.html

Related BNF is:

```
import_stmt: import_name | import_from
import_name: 'import' dotted_as_names
import_from: ('from' (('.' | '...')* dotted_name | ('.' | '...')+)
              'import' ('*' | '(' import_as_names ')' | import_as_names))
import_as_name: NAME ['as' NAME]
dotted_as_name: dotted_name ['as' NAME]
import_as_names: import_as_name (',' import_as_name)* [',']
dotted_as_names: dotted_as_name (',' dotted_as_name)*
dotted_name: NAME ('.' NAME)*
```

Before this PR ...

1. an `as` keyword set the alias-list context onto stack assuming all following commas to be a list of alias names, which is obviously wrong resulting in #2944.
2. line continuation operators (`\`) broke parenthesized import lists `from foo import ( bar, bat \` by popping corresponding context off stack.
3. imports were allowed to follow asterisk imports (`*`) `from import *, foo` was ok, but is not.

To fix 1.) only one token is allowed after an `as` keyword. Corresponding context is pushed onto stack, to allow multiple
`name as alias` expressions to follow each other.

To fix 2.) some contexts needed to be split in order to match line continuation operator outside of parenthesized import lists only.

To fix 3.) asterisk pattern is moved to `imports-from-import-body` context.

Some named contexts are introduced for `name-content`, `dotted-name-content` and `qualified-name-content`. The first one is needed to pop after a single identifier was matched after `as` keyword. The ladder two are introduced to make those name
contexts follow the same implementation scheme and help with debugging by reducing amount of "anonymous contexts", which are hard to identify in the scope popup.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -113,13 +113,15 @@ contexts:
     - match: \b(pass)\b
       scope: keyword.control.flow.pass.python
 
+###[ IMPORTS ]################################################################
+
   imports:
-    - match: \b(import)\b
+    - match: \bimport\b
       scope: keyword.control.import.python
       push:
         - imports-import-body
         - expect-absolute-import
-    - match: \b(from)\b
+    - match: \bfrom\b
       scope: keyword.control.import.from.python
       push:
         - imports-from-body
@@ -131,24 +133,18 @@ contexts:
     - match: ','
       scope: punctuation.separator.import-list.python
       push: expect-absolute-import
-    - match: (?=\bas\b)
-      set: import-alias-list
+    - include: imports-as
     - include: qualified-name
-    - match: (?=\S)
-      pop: true
+    - include: import-illegal-names
 
   imports-from-body:
     - meta_scope: meta.statement.import.python
     - meta_content_scope: meta.import-source.python
     - include: line-continuation-or-pop
     - match: (?=\bas\b)
-      set: import-alias-list
+      set: imports-from-import-names
     - match: (?=\bimport\b)
-      set:
-        - meta_include_prototype: false
-        - match: import
-          scope: keyword.control.import.python
-          set: imports-from-import-body
+      set: imports-from-import
     - match: '{{illegal_names}}\b'
       scope: meta.import-path.python invalid.illegal.name.python
     - match: '{{identifier}}'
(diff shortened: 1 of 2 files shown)


Python/syntax_test_python.py
```diff
@@ -47,7 +47,28 @@
...
+#      ^ invalid.illegal.name.import.python
 
+from os import *, path
+#^^^ meta.statement.import.python - meta.import-source - meta.import-path
+#   ^ meta.statement.import.python meta.import-source.python - meta.import-path
+#    ^^ meta.statement.import.python meta.import-source.python meta.import-path.python
+#      ^ meta.statement.import.python meta.import-source.python - meta.import-path
+#       ^^^^^^^^ meta.statement.import.python - meta.import-source - meta.import-path
+#               ^^^^^^^ - meta.statement
+#              ^ constant.language.import-all.python
```
BenjaminSchaaf, line 60: Minor nitpick, but this should be one line higher.

Python/Python.sublime-syntax
```diff
@@ -131,24 +133,18 @@ contexts:
...
     - include: qualified-name
-    - match: (?=\S)
-      pop: true
+    - include: import-illegal-names
 
   imports-from-body:
     - meta_scope: meta.statement.import.python
     - meta_content_scope: meta.import-source.python
     - include: line-continuation-or-pop
     - match: (?=\bas\b)
```
BenjaminSchaaf, line 144: This seems to allow `from as foo` and `import as foo`. It's also broken in master but worth fixing here.

BenjaminSchaaf (changes requested):
Some minor things, otherwise looks good.

Python/syntax_test_python.py
deathaxe: No, it shouldn't. First set of test cases addresses (primarily) meta scopes while the second chunk addresses "normal" scopes. This has become some kind of common scheme in many test files and may cause something like that.

Normally, those second parts should also start with the first token to be complete, like in line 79, but that's out of scope of this PR, IMHO.

Python/Python.sublime-syntax
deathaxe: I wouldn't call it broken in the first place, but rather a question of how to support/handle incomplete statements.

With regards to

https://github.com/sublimehq/Packages/blob/3783eebcc7416e08056df1c982a256094109b802/Python/syntax_test_python.py#L105-L114

... it is expected behavior.

We can either scope everything illegal, just because some part of a statement is missing, or we can silently accept it and try to recover as early as possible to maintain editing experience pleasent.

I've seen both strategies all over syntaxes so far. 

We also need to consider which of those illegal highlightings are better handled by linters. While illegals might be helpful in absence of linters, user experience might quickly become inconsistent if those are present, because static syntax definitions never can be as exact as linters.

Also seeing whole bunches of code in red or flickering single chars/words all the time while editing is a pain. It's quite normal for expressions and statements to be incomplete or invalid during authoring code, so I personally tend to argue for highlighting things illegal only, if it is sure to be wrong at a certain position not being likely to caused by incomplete statements. Otherwise just recover as early as possible continuing highlighting. We want to write syntax definitions, but no linters. Sophisticated illegal handling often introduces much more complexity.

That's also what happens in syntaxes which use multi-push often. If a token is not found at expected position, the context is just popped continuing with the next one. It would end up in exactly this behavior of supporting `import as foo`.

FichteFoll: Since `import  as foo` is something than can happen during typing, I agree with @deathaxe here in that it should not be marked as invalid. The flickering during typing is indeed annoying. However, we should still mark other keywords as illegal names in this place, e.g. `import from as foo`.

rwols: I don’t agree with the idea that syntaxes should act as static analyzers. That’s what linters and language servers are for.

FichteFoll (approved):

BenjaminSchaaf (approved):

PR #2954, opened 2021-08-13
[JavaScript] [TypeScript] Implement static blocks.

Thom1729:
Static blocks are a Stage 3 proposal with an experimental implementation in Firefox. In addition, they're part of TypeScript 4.4, which is at the RC stage.

I use `keyword.declaration` for the `static` keyword in these blocks. I think this is more appropriate than `storage.modifier`, but I could be persuaded otherwise.

- https://github.com/tc39/proposal-class-static-block
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-rc/

As part of this PR, I slightly reshuffled how class element modifiers work. This was necessary to avoid code duplication due to https://github.com/sublimehq/sublime_text/issues/3494.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1376,14 +1376,16 @@ contexts:
       set:
         - match: \{
           scope: punctuation.section.block.begin.js
-          set:
-            - meta_scope: meta.block.js
-            - match: \}
-              scope: punctuation.section.block.end.js
-              pop: true
-            - include: statements
+          set: static-block-body
         - match: (?=\S)
           fail: static-block
+
+  static-block-body:
+    - meta_scope: meta.block.js
+    - match: \}
+      scope: punctuation.section.block.end.js
+      pop: true
+    - include: statements
 
   class-element-modifiers:
     - match: (?={{modifier}})
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1338,7 +1338,14 @@ contexts:
         - function-declaration-expect-body
         - function-declaration-expect-parameters
 
-    - include: class-element-modifiers
+    - match: (?=static{{identifier_break}})
+      branch_point: static-block
+      branch:
+        - static-block
+        - class-element-modifiers
+
+    - match: (?={{modifier}})
+      push: class-element-modifiers
 
     - match: |-
         (?x)(?=
@@ -1363,8 +1370,24 @@ contexts:
     - match: (?={{class_element_name}})
       push: class-element
 
+  static-block:
+    - match: static{{identifier_break}}
+      scope: keyword.declaration.other.js
+      set:
+        - match: \{
+          scope: punctuation.section.block.begin.js
+          set:
+            - meta_scope: meta.block.js
+            - match: \}
+              scope: punctuation.section.block.end.js
+              pop: true
+            - include: statements
+        - match: (?=\S)
+          fail: static-block
+
   class-element-modifiers:
     - match: (?={{modifier}})
+      pop: true
       branch_point: class-element-modifier
       branch:
         - class-element-modifier
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1372,7 +1372,7 @@ contexts:
 
   static-block:
     - match: static{{identifier_break}}
-      scope: keyword.declaration.other.js
+      scope: storage.modifier.js
       set:
         - match: \{
           scope: punctuation.section.block.begin.js
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1388,7 +1388,7 @@ contexts:
     - include: statements
 
   class-element-modifiers:
-    - match: (?={{modifier}})
+    - match: ''
       pop: true
       branch_point: class-element-modifier
       branch:
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -234,7 +234,8 @@ contexts:
     - match: '\+|-'
       scope: storage.modifier.js
 
-    - include: class-element-modifiers
+    - match: (?={{modifier}})
+      push: class-element-modifiers
 
     - match: (?=[<(])
       push:
--- a/JavaScript/tests/syntax_test_js_class.js
+++ b/JavaScript/tests/syntax_test_js_class.js
@@ -98,7 +98,7 @@ class MyClass extends TheirClass {
 //             ^^^^^^^^^^^^^ meta.function
 
     static {
-//  ^^^^^^ keyword.declaration.other
+//  ^^^^^^ storage.modifier
 //         ^ meta.block punctuation.section.block.begin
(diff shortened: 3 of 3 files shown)


deathaxe:
Is it possible just match a `static` keyword within a class body block as well as a statement block by just matching `{`? 

Is this lookahead-push-branch-fail-push-fail-set-lookahead-... fetish needed here?

Thom1729:
If we highlight the `static` token differently in static blocks versus as a modifier, then we have to branch. For the same reason, we have to branch to detect methods.

If we used the `storage.modifier` scope even in static blocks, then we could combine the static block case and the modifier case. However, the modifier behavior is also used in TypeScript for tokens like `private`, which do not need the block behavior. In principle, we could allow any modifier to define a block, but that seems pretty weird. Alternately, we could remove `static` from the regular modifier handling and implement it separately with logic for both blocks and modifiers, but then we'd be duplicating the modifier logic, which would probably require some copy-pasting due to https://github.com/sublimehq/sublime_text/issues/3494.

For these reasons, I thought it made the most sense to implement static blocks separately. If TypeScript were to add blocks for several other modifiers, then it might make more sense to combine functionality (assuming that we also wanted to change the scope).

deathaxe:
Why do we want `static` to get scoped different? Doesn't it just denote following block to be of static nature? Java has exactly the same construct and I find `storage.modifier` very sufficient. I wouldn't want and won't change that.

My primary point is just we should avoid to cause any token to need to be parsed 2 or 3 times, just because matching several lookaheads befor the real keyword is matched. Duplication of some patterns or not. This just slows down parsing.

It's probably not a good comparison, but I did some benchmarks about Java vs. JavaScript to see how all those branching stuff performs. I found the Java Rewrite to be 20-40% faster for most tested constructs.

Yes, context switching is quite cheap, but if it is 99% of the code it counts anyway.

I just get alerted, if I see a lookahead pushing a context which  just matches a token and pops.

Thom1729:
In terms of intent, I don't think that `static` in a block is really a modifier. It's not modifying anything — it's a keyword declaring a block. But the Java precedent is persuasive; I don't see the need to break with precedent.

However, I think that avoiding the additional branch would still require copy-pasting code due to sublimehq/sublime_text#3494.

I'm not surprised that JavaScript is slower than Java because the syntax is much more complicated. There are a ton of unavoidable hard cases. That doesn't mean that there aren't also avoidable hard cases. One of these days I need to dig into benchmarks to see what the low-hanging fruit are, but I suspect — and I could certainly be wrong — that the worst offenders are the capture groups for the special dollar sign identifiers and the lookaheads for things like function name detection. Of course, this might depend heavily on what the benchmark is stressing.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -1363,8 +1370,24 @@ contexts:
...
+            - meta_scope: meta.block.js
+            - match: \}
+              scope: punctuation.section.block.end.js
+              pop: true
+            - include: statements
+        - match: (?=\S)
+          fail: static-block
+
   class-element-modifiers:
     - match: (?={{modifier}})
```
deathaxe, line 1389: The way this context is used, would allow to use and empty pattern `- match: ''` instead.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -1363,8 +1370,24 @@ contexts:
     - match: (?={{class_element_name}})
       push: class-element
 
+  static-block:
+    - match: static{{identifier_break}}
```
deathaxe, line 1374: How about `- meta_include_prototype: false` here (and in the next context)?

JavaScript/JavaScript.sublime-syntax
```diff
@@ -1363,8 +1370,24 @@ contexts:
...
+      scope: storage.modifier.js
+      set:
+        - match: \{
+          scope: punctuation.section.block.begin.js
+          set:
+            - meta_scope: meta.block.js
```
deathaxe, line 1380: Especially as this syntax is heavily used as base for inheriting, I'd vote to avoid anonymous contexts whenever possible. It also helps debugging now that context names are printed to the scope popup.

Thom1729: I've added `static-block-body` for the most deeply-nested context.

JavaScript/JavaScript.sublime-syntax
Thom1729: Here, I think it wouldn't do anything. In the next context, adding `meta_include_prototype` could break things:

```js
class C {
  static /**/ {}
}
```

JavaScript/JavaScript.sublime-syntax
Thom1729: In order to do that, we'd have to consume and scope static in the rule in `class-body-contents` that branches to `static-block` and `class-element-modifiers`. But that would break in this case:

```js
class C {
  static() {}
}
```

JavaScript/JavaScript.sublime-syntax
deathaxe: Only intented to propose to add it before 1374 as this context is pushed by lookaheads. So any injected/prepended rule would be nonsense.

JavaScript/JavaScript.sublime-syntax
deathaxe: Not sure about that. 

1. I don't see anything changing when turning that pattern into an empty one.
2. `class-element-modifiers` is pushed on stack only by the branch_point `static-block` and the following pattern which both contain `static`. So this pattern must match anyway. Otherwise you'd never pop this context off stack. The first branch `class-element-modifier` consumes `{{modifier}}`. If it fails `static` will be matched as method declaration. But this happens with the current pattern as well.

Thom1729: Maybe I'm misunderstanding. Can you post the revised `static-block`?

deathaxe: The only thing I did is ...

```
  class-element-modifiers:
    - match: ''
      pop: true
      branch_point: class-element-modifier
      branch:
        - class-element-modifier
        - class-element
```

Java contains several such contexts, which are pushed/branched onto stack by a lookahead and immediately need to hand over treatment to other ones, like this one. It's just the lookahead which is technically not needed in this situation, if I don't miss something.

Thom1729: Ah, I thought you were talking about the `static-block` context. I've changed `class-element-modifiers` accordingly.

JavaScript/JavaScript.sublime-syntax
deathaxe: While suggesting this, I have use cases in mind, which inherit JavaScript to inject variable interpolation via `prototype` such as https://github.com/sublimehq/Packages/pull/2797/files#diff-450eddd7ff448ef509c1ae063ca8f11b631dfe871ccc2ecd6bf6656416159260

That's probably a general thing to take into account, which goes beyound the scope of this PR.

Thom1729: I think that if we needed a change here, we'd probably need it in other places, and I'd rather burn that bridge when we come to it.

deathaxe (approved):

jfcherng (approved):

PR #2955, opened 2021-08-13
[HTML] Add detection of boolean attributes to avoid empty attribute values

bhdzllr:
If you type HTML attributes and these are inserted by the auto completion an empty attribute value is added automatically and the cursor jumps to the value so users can start typing.

For boolean attributes like "disabled" this is valid, but not necessary and I think most people use boolean attributes without a value.

This PR removes the value for boolean attributes. I got the list of boolean attributes from this Gist: https://gist.github.com/ArjanSchouten/0b8574a6ad7f5065a5e7#gistcomment-3231272

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -18,6 +18,14 @@ KIND_TAG_MARKUP = (sublime.KIND_ID_MARKUP, 't', 'Tag')
 ENABLE_TIMING = False
 
 
+boolean_attributes = {
+    'async', 'autofocus', 'autoplay', 'checked', 'contenteditable', 'controls',
+    'default', 'defer', 'disabled', 'formNoValidate', 'frameborder', 'hidden',
+    'ismap', 'itemscope', 'loop', 'multiple', 'muted', 'nomodule', 'novalidate',
+    'open', 'readonly', 'required', 'reversed', 'scoped', 'selected',
+    'typemustmatch'
+}
+
 def timing(func):
     @wraps(func)
     def wrap(*args, **kw):
@@ -396,7 +404,6 @@ def get_tag_attributes():
 
     return tag_attr_dict
 
-
 class HtmlTagCompletions(sublime_plugin.EventListener):
     """
     Provide tag completions for HTML
@@ -588,7 +595,7 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
             [
                 sublime.CompletionItem(
                     trigger=attr,
-                    completion=f'{attr}="$1"{suffix}',
+                    completion=f'{attr}{suffix}' if attr in boolean_attributes else f'{attr}="$1"{suffix}',
                     completion_format=sublime.COMPLETION_FORMAT_SNIPPET,
                     kind=KIND_ATTRIBUTE_SNIPPET
                 )


HTML/html_completions.py
```diff
@@ -396,6 +396,14 @@ def get_tag_attributes():
 
     return tag_attr_dict
 
+def get_boolean_attributes():
```
deathaxe, line 399: As those boolean attributes are not needed anywhere else, you could probably directly return a boolean result about whether a given attribute is a boolean or not.

Using `set`s is also slightly faster.

```suggestion
def is_bool(attr):
    return attr in {
        'async', 'autofocus', 'autoplay', 'checked', 'contenteditable',
        'controls', 'default', 'defer', 'disabled', 'formNoValidate',
        'frameborder', 'hidden', 'ismap', 'itemscope', 'loop', 'multiple',
        'muted', 'nomodule', 'novalidate', 'open', 'readonly', 'required',
        'reversed', 'scoped', 'selected', 'typemustmatch'
    }
```

HTML/html_completions.py
```diff
@@ -588,7 +602,7 @@ def attribute_completions(self, view, pt, prefix):
             [
                 sublime.CompletionItem(
                     trigger=attr,
-                    completion=f'{attr}="$1"{suffix}',
+                    completion=self.generate_attribute(attr, suffix),
```
deathaxe, line 605: With the `is_bool()` it was possible to avoid one level of function calls, which also slightly improves performance.

```suggestion
                    completion=f'{attr}{suffix}' if is_bool(attr) else f'{attr}="$1"{suffix}',
```

deathaxe (commented):
Thanks for contribution. An overall good and simple solution for a topic which probaly annoys a lot of users - me, too.

The following are just some highlevel suggestions.

HTML/html_completions.py
rchl: It might not be a convention in this file but creating a top-level Set variable would also be more performant (no need to call a function and recreate the set on each call).

deathaxe: I haven't seen any timing improvements when using global set of boolean attributes. Maybe my CPU is to fast and set to small.

Everyone interested can test it by setting the global `ENABLE_TIMING = True`.

rchl: I guess it won't really make any noticeable difference for this case unless this code is triggered a 10000 times a second or something.

Still, avoiding an extra function call and an initialization of Set should be faster (in general).

bhdzllr:
@deathaxe @rchl Thank you for the review and your suggestions, I updated my commit. I added the Set with the attributes on top of the file.

HTML/html_completions.py
```diff
@@ -396,7 +404,6 @@ def get_tag_attributes():
 
     return tag_attr_dict
 
-
```
deathaxe, line 399: Would you mind adding back this line to satisfy flake8?

Thanks

bhdzllr: No problem. Is there also a convetion for the `boolean_attributes` declaration on top of the file?

bhdzllr: ... I feel like I should also add a extra break after the `boolean_attributes` declaration?

bhdzllr: Done.

deathaxe (approved):

jfcherng (approved):

jfcherng (approved):

PR #2961, opened 2021-08-15
[PHP] Refactor string contexts

deathaxe:
This PR is part 2 to address https://github.com/sublimehq/sublime_text/issues/4763.

It refactors string related contexts in order to 

1. reduce application of branch_points and avoid as many context switches as possible.  
   It results in huge performance improvements when recovering from incomplete quoted strings in large files.
   While closing a quoted string at the beginning of a file of 500k lines took 16-19s before, it completes within 200-300ms with this PR merged.
2. apply proper `meta.string meta.interpolation` scopes as already done in other syntaxes.

Also loading such a huge file completes much faster as switching back to ST does after modifying the file externally.

While it required 16 seconds for ST do display again, it now completes nearly immediatelly.

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -1592,15 +1592,25 @@ contexts:
         - string-double-quoted-regexp
         - string-double-quoted-plain
     - match: \"
-      scope: punctuation.definition.string.begin.php
+      scope: string.quoted.double.php punctuation.definition.string.begin.php
       push: string-double-quoted-content
+
+  string-double-quoted-end:
+    - match: \"
+      scope: string.quoted.double.php punctuation.definition.string.end.php
+      pop: 1
 
   string-double-quoted-content:
     - meta_include_prototype: false
-    - meta_scope: meta.string.php string.quoted.double.php
+    - meta_scope: meta.string.php
+    - meta_content_scope: string.quoted.double.php
     - include: string-double-quoted-end
     - match: (?={{sql_indicator}})
-      set: string-double-quoted-sql-content
+      push: string-quoted-sql-content
+      with_prototype:
+        - match: (?=")
+          pop: 1
+        - include: interpolation
     - match: (?=\S)
       set: string-double-quoted-plain-content
 
@@ -1643,11 +1653,6 @@ contexts:
       scope: string.quoted.double.php punctuation.definition.string.begin.php
       set: string-double-quoted-plain-content
 
-  string-double-quoted-end:
-    - match: \"
-      scope: string.quoted.double.php punctuation.definition.string.end.php
-      pop: 1
-
   string-double-quoted-plain-content:
     - meta_include_prototype: false
     - meta_scope: meta.string.php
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

willrowe:
Is there a specific reason that you removed the `meta.string-contents.quoted` scopes?

I have been building my own syntax and looking at different language syntaxes to learn how to do certain things. I noticed that some apply string scopes differently. I like how PHP adds `meta.string-contents.quoted` to just the contents (excluding the quotes themselves) because when the cursor is inside of a string and I expand the selection to the scope, it will select only the text itself and not the quotes. Usually I am doing this in order to replace the entire contents of the string, so it is very useful. Not many other language syntaxes do this that I could find, so maybe I will open an issue to discuss consistency and the preferred way to do this.

All that to say that it seems that removing the `meta.string-contents.quoted` scope will not allow me to expand the selection to just the string content anymore.

deathaxe:
First of all, you can still `ctrl+shift+a` to select string's content only.

During last dev cycle it was discussed how to handle interpolated strings to solve the problem `string` scopes manipulating colors of interpolated code in various situations. The goal was to highlight interpolations as normal code out of the box without color schemes needing to create special rules for them to look well.

We came up with `meta.string string.quoted` as general scope for quoted strings. An interpolating context would clear the `string` scope and add `meta.interpolation`, so any interpolation would result in `meta.string meta.interpolation - string`.

Most prominent examples are CSS and JS highlighting within HTML tag attributes. But also ShellScript in Makefile or Ruby on Rails, etc. to mension only few examples.

Much effort was also put into syntaxes to find and apply scope names to make them look more equal even with simple color schemes. The goal is to reduce the need of syntax specific color schemes.

With that goal in mind, the `meta.string string` scope scheme is applied to PHP by this PR, as this is how most other default syntaxes are implemented already.

PHP is actually the syntax I saw something like `meta.string-content` being applied to string content only, while working on this PR.

willrowe:
> First of all, you can still ctrl+shift+a to select string's content only.

Fantastic! Didn't know about this.

> With that goal in mind, the meta.string string scope scheme is applied to PHP by this PR, as this is how most other default syntaxes are implemented already.

Glad this has already been discussed, makes sense that this will bring PHP more in line and consistent with the others.

Thanks for all the info, all makes sense to me, carry on!

deathaxe:
@jfcherng: While writing last comment I noticed `meta.interpolation` is not yet properly handled. It does not yet clear string scopes. It seems a solution would look very much like the one's already implemented in Bash or Perl.

I could prepare required changes.

Question is: Do we want to add it to this PR or handle it separately, once this one is merged?

jfcherng:
> Question is: Do we want to add it to this PR or handle it separately, once this one is merged?


If you don't mind spending some time to split it out, I would like to have more good PRs to be merged (and thus we can resolve potential conflicts) ASAP.

deathaxe:
Agree. There are still to many potential conflicting PRs pending, which all need another approval after conflicts have been resolved.

Interpolation may probably need some time to mature anyway. So maybe we get another approval on this and probably others earlier.

I'll prepare it locally and hold it back.

keith-hall (approved):

Issue #2956, opened 2021-08-15
[PHP] SELECT string highlighting


amiklny:
### Description

Incorrect highlight

### Steps to reproduce

1. Start Sublime Text
2. Paste code
```
<?
$res = CUser::GetList(
	'LAST_NAME',
	'ASC',
	$userFilter,
	[
		'SELECT' => ['UF_DEPARTMENT'],
		'FIELDS' => ['ID', 'LAST_NAME', 'WORK_POSITION']
	]
);
?>
```

### Expected behavior

Correct highlight ('SELECT' is highlighted in green)

### Actual behavior

Incorrect highlight ('SELECT' is highlighted in magenta)

### Environment

4113, windows 11 (10)

BenjaminSchaaf:
Looks like it's highlighting the SELECT as SQL, which generally makes sense except in the case where the string is just "SELECT".

keith-hall:
partially relevant to the case where the string is just "SELECT" for example: https://github.com/sublimehq/Packages/issues/572

PR #2962, opened 2021-08-16
[PHP] Do not highlight plain SQL indicator as SQL

jfcherng:
```php
    $sql = "SELECT";
    //      ^^^^^^ - keyword.other.DML
```

Fixes https://github.com/sublimehq/Packages/issues/2956

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -9,7 +9,7 @@ variables:
   identifier: '\b{{identifier_start}}[[:alnum:]_]*\b'
   path: '\\?({{identifier}}\\)*{{identifier}}'
   regex_modifier: '[eimsuxADJSUX]*'
-  sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER)\b
+  sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER)(?=$|\s)
   visibility_modifier: '\b(?:var|public|private|protected)\b'
 
 contexts:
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

Thom1729 (dismissed):

PHP/PHP Source.sublime-syntax
```diff
@@ -9,7 +9,7 @@ variables:
   identifier: '\b{{identifier_start}}[[:alnum:]_]*\b'
   path: '\\?({{identifier}}\\)*{{identifier}}'
   regex_modifier: '[eimsuxADJSUX]*'
-  sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER)\b
+  sql_indicator: \s*(?:SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER)(?=$|\s)
```
deathaxe, line 12: I don't think we need to highlight `SELECT` as keyword in an EOF situation, if we don't do so in situations of it being the only keyword in a quoted string.

deathaxe:
I wonder how long it takes until someone complains about `"SELECT" . cols . "FROM"` not being correctly highlighted.

PHP/PHP Source.sublime-syntax
jfcherng: @deathaxe that would fail one of added tests.

jfcherng:
> I wonder how long it takes until someone complains about `"SELECT" . cols . "FROM"` not being correctly highlighted.

It has never been done iirc. And unlikely will be done.

PHP/PHP Source.sublime-syntax
deathaxe: `(?=\s)` would cover that use case as `\n` is part of `\s`. In eof the lookahead wouldn't match and thus `SELECT` would not be highlighted as keyword until a space or newline is added. With current solution SELECT is highlihted as keyword at the end of file.

jfcherng: Okay. I see. `\s` auto covers `$` for non-EOF situation.

deathaxe (approved):

Thom1729 (approved):

PR #2964, opened 2021-08-17
[CI] do not run syntax tests on every push or PR

jrappen:


.github/workflows/ci.yml
```diff
@@ -5,9 +5,17 @@ on:
       - '*'
     tags-ignore:
       - '*'
+    paths:
+      - '**/syntax_test_*'
+      - '**/*.sublime-syntax'
+      - '**/*.tmPreferences'
```
jfcherng, line 11: `.github/workflows/ci.yml` should be added.

.github/workflows/ci.yml
```diff
@@ -5,9 +5,17 @@ on:
...
   pull_request:
     branches:
       - '*'
+    paths:
+      - '**/syntax_test_*'
+      - '**/*.sublime-syntax'
+      - '**/*.tmPreferences'
```
jfcherng, line 18: `.github/workflows/ci.yml` should be added.

jfcherng (changes requested):

jfcherng (dismissed):

jfcherng (approved):

deathaxe (approved):
Not sure if we really need to track `*.hidden-tmLanguage`, but it doesn't cause any harm.

Thom1729 (approved):

deathaxe:
Already have been using this kind of filter in MarkdownEditing without noticing any issues.

PR #2968, opened 2021-08-18
[CSS] Add function argument completions

### LINKED ISSUE #2771: [CSS] Don't work auto-fill grid in Sublime Text 3
Hello. I am making a website where I need to arrange blocks of 4 in an unlimited number of lines. And I need the blocks to be transferred if there is no space, but the auto-fill does not work. Why?

- Sublime Version: Version 3.2.2 Build 3211
- OS Version: Windows 10 Home
Screenshot_1

deathaxe:
Fixes #2771

This commit adds support for function arguments completions found on https://github.com/mdn/data

A `get_common_values()` function helps sharing common values between functions and properties.

Common value keys use dashes as sub-word separators for easier maintenance as this is the style used on MDN.


CSS/css_completions.py
```diff
@@ -714,14 +925,19 @@ def complete_property_value(self, view, prefix, pt):
...
-            args_region.a - 2, sublime.CLASS_WORD_START | sublime.CLASS_WORD_END)
-        func_name = view.substr(func_region)
-
-        # TODO: provide more function specific argument completions
+    def complete_function_argument(self, view: sublime.View, prefix, pt):
+        func_name = ""
+        nest_level = 1
+        for i in range(pt - 1, pt - 32 * 1024, -1):
+            ch = view.substr(i)
+            if ch == ")" and not view.match_selector(i, "string, comment"):
```
keith-hall, line 933: I'd vote for moving the `view.match_selector(i, "string, comment")` call to a line above to avoid it needing to be repeated, and `continue`ing if it's `True`.

Can we have some comments or more explanation of how this method works, please? I understand it looks backwards from the given point one char at a time to find where the function name is, but I don't entirely understand how/the nesting level stuff as it's slightly backwards from what I'm used to (probably because it looks backwards?). It's s shame we don't have unit tests because that would likely clarify it enough for me.

CSS/css_completions.py
```diff
@@ -84,19 +79,19 @@ def get_properties():
             'bisque', 'black', 'blanchedalmond', 'blue', 'blueviolet', 'brown',
             'burlywood', 'cadetblue', 'chartreuse', 'chocolate', 'coral',
             'cornflowerblue', 'cornsilk', 'crimson', 'cyan', 'darkblue',
-            'darkcyan', 'darkgoldenrod', 'darkgray', 'darkgrey', 'darkgreen',
+            'darkcyan', 'darkgoldenrod', 'darkgray', 'darkgreen',
```
keith-hall, line 82: do I understand correctly that the English spelling of `grey` is deprecated? :)

keith-hall (commented):
This looks great - nice work! I have one minor concern about how the `complete_function_argument` function works as it is a little unclear to me, but this is a great improvement!

CSS/css_completions.py
deathaxe: The current implementation calls `view.match_selector()` only if `ch` is a parenthesis. Splitting that statement and moving `view.match_selector()` would cause it to be called on each charachter, which would significantly slow down the function.

If someone can come up with a more pythonic solution, I'd be happy to use that instead.

The main issue is a function can be called in another function's arguments list. The level of nesting is undefined. So we need to make sure to ignore any function call arguments list located in front of the caret.

Example

```
.text-calc-clamp-max-min {
    top: clamp(1*5pt, calc(12/5), | 100rem);
}
```

The `|` denotes caret. We want to make sure to return `clamp` rather than `calc`

and do so even if parentheses occure in strings. Not sure if it can happen in CSS, but wanted to make sure it works with it.

```
.class {
    top: any(foo(bar(baz("(")),  ) |  );
}
```

The function basically finds the next opeing parentheses while ignoring any balanced parens.

CSS/css_completions.py
deathaxe: I don't recall where I got that from. Too long ago.

CSS/css_completions.py
keith-hall: > The current implementation calls `view.match_selector()` only if `ch` is a parenthesis. Splitting that statement and moving `view.match_selector()` would cause it to be called on each charachter, which would significantly slow down the function.

I forgot about short circuiting here, thanks for correcting me :)

Nice, thanks for the explanation and the comments

keith-hall (approved):

jfcherng (approved):

PR #2969, opened 2021-08-18
[C#] rework tuple scopes to match Python

keith-hall:
Change the meta and punctuation scopes used for tuples in the C# syntax to match the Python syntax definition - from `meta.group.tuple punctuation.section.group.begin` to `meta.sequence.tuple punctuation.section.sequence.begin` etc.
Suggested in https://github.com/sublimehq/Packages/pull/2838#discussion_r690632925

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1366,8 +1366,8 @@ contexts:
     - match: (?=\()
       branch_point: tuple_or_group
       branch:
-        - group-begin
-        - tuple-begin
+        - group_begin
+        - tuple_begin
     - match: \{
       scope: punctuation.section.block.begin.cs
       set:
@@ -1378,13 +1378,13 @@ contexts:
     - match: (?=[]}>,]|{{reserved}})
       pop: true
 
-  group-begin:
+  group_begin:
     - meta_include_prototype: false
     - match: \(
       scope: punctuation.section.group.begin.cs
       set: group
 
-  tuple-begin:
+  tuple_begin:
     - meta_include_prototype: false
     - match: \(
       scope: punctuation.section.sequence.begin.cs
@@ -1457,7 +1457,7 @@ contexts:
 
   function_call_arguments:
     - meta_content_scope: meta.function-call.cs meta.group.cs
-    - match: '\)'
+    - match: \)
       scope: meta.function-call.cs meta.group.cs punctuation.section.group.end.cs
       pop: true
     - match: ''
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1366,12 +1366,8 @@ contexts:
     - match: (?=\()
       branch_point: tuple_or_group
       branch:
-        - - match: \(
-            scope: punctuation.section.group.begin.cs
-            set: group
-        - - match: \(
-            scope: punctuation.section.sequence.begin.cs
-            set: tuple
+        - group-begin
+        - tuple-begin
     - match: \{
       scope: punctuation.section.block.begin.cs
       set:
@@ -1381,6 +1377,18 @@ contexts:
         - include: code_block_in
     - match: (?=[]}>,]|{{reserved}})
       pop: true
+
+  group-begin:
+    - meta_include_prototype: false
+    - match: \(
+      scope: punctuation.section.group.begin.cs
+      set: group
+
+  tuple-begin:
+    - meta_include_prototype: false
+    - match: \(
+      scope: punctuation.section.sequence.begin.cs
+      set: tuple
 
   tuple:
     - meta_scope: meta.sequence.tuple.cs
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1525,9 +1525,8 @@ contexts:
       scope: punctuation.section.sequence.begin.cs
       set:
         - meta_scope: meta.sequence.tuple.cs
-        - match: (\))
-          captures:
-            1: punctuation.section.sequence.end.cs
+        - match: \)
+          scope: punctuation.section.sequence.end.cs
           pop: true
         - match: ','
           scope: punctuation.separator.sequence.cs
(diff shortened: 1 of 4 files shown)


C#/C#.sublime-syntax
```diff
@@ -1361,29 +1363,32 @@ contexts:
...
+            set: tuple
     - match: \{
       scope: punctuation.section.block.begin.cs
       set:
         - match: \}
           scope: punctuation.section.block.end.cs
           pop: true
         - include: code_block_in
-    - match: (?=\}|\)|>|\]|,|{{reserved}})
+    - match: (?=[]}>,]|{{reserved}})
```
jrappen, line 1382: This regex irritated me for a second. Maybe readability can be improved?

jrappen:
I assume you could also move to `version: 2` and replace all instances of `pop: true` with `pop: 1` while at it.

C#/C#.sublime-syntax
deathaxe: I don't think lines need to be split. There are lots of such patterns everywhere. Maybe the `]` at first position is a quite nifty trick to prevent escaping, but otherwise its a normal character class alternated by a variable.

deathaxe:
Moving to syntax version 2 would be a separate step/PR, IMHO. This is just a follow up to tidy meta scopes of sequences.

C#/C#.sublime-syntax
```diff
@@ -1361,29 +1363,32 @@ contexts:
...
             3: punctuation.section.brackets.end.cs
             4: keyword.operator.pointer.cs
-    - match: \(
-      scope: punctuation.section.group.begin.cs
+    - match: (?=\()
       branch_point: tuple_or_group
       branch:
-        - group
-        - tuple
+        - - match: \(
```
deathaxe, line 1369: I'd vote to avoid such anonymous constructs whenever possible.

We can easily use `tuple` vs. `tuple-content` or `tuple-body`. The first one containing the opening bracket and the latter ones could be created by renaming the current `tuple`  context.

Also `prototypes` could be excluded here as the `(?=\()` lookahead already implies the next token must be matched as opening parenthesis. So no any injected/prepended pattern would be nonsense. (I have templating languages in mind, which like to inject their variables to interpolate with syntaxes)

C#/C#.sublime-syntax
```diff
@@ -1498,15 +1514,15 @@ contexts:
...
     - match: \(
-      scope: punctuation.section.group.begin.cs
+      scope: punctuation.section.sequence.begin.cs
       set:
-        - meta_scope: meta.group.tuple.cs
+        - meta_scope: meta.sequence.tuple.cs
         - match: (\))
           captures:
-            1: punctuation.section.group.end.cs
+            1: punctuation.section.sequence.end.cs
```
deathaxe, line 1522: I guess the capture group could be removed?

C#/C#.sublime-syntax
keith-hall: I like your level of thinking - I wonder if such an optimization to skip prototypes in such a case could be applied automatically by the syntax engine. If the parent context doesn't skip the prototype, the match is a lookahead and the child context uses the same regex pattern from inside the lookahead: then no need to apply the prototype. I guess it could get complicated where a context is referenced in multiple places, so maybe we can somehow work on a linter to suggest such changes for a potential small perf boost? :D

C#/C#.sublime-syntax
keith-hall: nice spot, done :)

C#/C#.sublime-syntax
keith-hall: looking at the regex pattern in ST (using PackageDev's highlighting) compared to GitHub's standard YAML highlighting makes quite a big difference to readability also. Happy to gather more opinions and split into two match patterns even if it could help ease of maintenance.

C#/C#.sublime-syntax
deathaxe: Issues like the one resolved by https://github.com/sublimehq/Packages/pull/2910 teach you this.

C#/C#.sublime-syntax
```diff
@@ -1422,19 +1435,31 @@ contexts:
...
+    - match: (?=[^\s{])
       pop: true
-    - match: '\{'
+    - match: \{
       scope: punctuation.section.braces.begin.cs
       set: initializer_constructor
 
   function_call_arguments:
     - meta_content_scope: meta.function-call.cs meta.group.cs
+    - match: '\)'
```
deathaxe, line 1460: I don't want to be nagging, but it's funny to see you removing quotes in line 1454 while adding a quoted pattern here. ;)

deathaxe (dismissed):

C#/C#.sublime-syntax
```diff
@@ -1361,29 +1363,40 @@ contexts:
...
       set:
         - match: \}
           scope: punctuation.section.block.end.cs
           pop: true
         - include: code_block_in
-    - match: (?=\}|\)|>|\]|,|{{reserved}})
+    - match: (?=[]}>,]|{{reserved}})
       pop: true
 
+  group-begin:
```
Ultra-Instinct-05, line 1381: Nearly all of the sub words in context names in C# are separated using an `_`. Maybe this could be `group_begin` & `tuple_begin` to maintain some consistency with context naming style followed here ?

C#/C#.sublime-syntax
keith-hall: it's not nagging, you're quite right that I'm trying to remove quotes in general and somehow let them slip in there :D I guess at some point a clean-up PR to rename contexts to ensure they match our style guidelines and remove unnecessary quotes is in order ;)

C#/C#.sublime-syntax
keith-hall: fixed, thanks :)

deathaxe (approved):

Ultra-Instinct-05 (approved):

jfcherng (approved):

PR #2980, opened 2021-08-19
[CSS] Add indentation rules

### LINKED ISSUE #1267: Indendation wrong for CSS close brace with comment
### Summary
Automatic indentation of CSS files correctly handles open and close braces.  However typing characters after the close brace gives wrong indentation.

The real world scenario where I hit this was a comment after the close brace of a length media query, where the comment describes the query.

It seems to work fine in other modes - only CSS has the problem.

### Expected behavior
Indent should decrease on a close brace even if there is a comment following.

### Actual behavior
Indent does not decrease on a close brace with a comment following.

### Steps to reproduce
1. Create a new file.
2. Set CSS mode.
3. Type (don't paste) the code below.  Note that typing the } does correctly un-indent.
4. When you type the '/' it re-indents. 

```
a {
  font-size: 100%;
} /
```
### Environment
- Windows 7
- Sublime Text 3.0 build 3143 64-bit

deathaxe:
Fixes #1267

This PR adds _Indentation Rules.tmPreferences_ to fix some issues with regards to comments after opening and closing brackets breaking correct indentation.

It borrows some strategies from C/C++ and Java to fully exclude block comments so normal CSS rules are indented correctly even after 

```
  /**
   * comment
   **/
``` 

Before this PR all following rules got vertically aligned to `**/` which resulted in one column off.

It also overrides ST's default bracket indentation to make sure to indent after an opening bracket, even if it is followed by arbritary number of probably nested brackets, while ignoring any bracket in quoted strings.

Maybe some of those rules are rarely needed and probably won't appear in real life code, but the intention is to hopefully find some kind of common pattern, which may be applied to all syntaxes using brackets for indentation.

Furthermore it makes sure to correctly detent closing brackets even if they are followed by comments or other kinds of code.

### DIFF
--- a/CSS/Indentation Rules.tmPreferences
+++ b/CSS/Indentation Rules.tmPreferences
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>scope</key>
+	<string>source.css</string>
+	<key>settings</key>
+	<dict>
+		<key>decreaseIndentPattern</key>
+		<string>^\s*[)\]}]</string>
+		<key>increaseIndentPattern</key>
+		<string>^.*[(\[{](?:\s*/\*.*\*/)*\s*(?://.*)?$</string>
+		<key>unIndentedLinePattern</key>
+		<string>^\s*((/\*|\*/).*)?$</string>
+	</dict>
+</dict>
+</plist>
--- a/CSS/Indentation Rules.tmPreferences
+++ b/CSS/Indentation Rules.tmPreferences
@@ -41,7 +41,7 @@
 			#   ST doesn't apply `Indentation Rules - Comments`, because
 			#   comment scope doesn't cover the whole line. So they need
 			#   to be excluded from indentation engine here.
-			^\s* (?: /\* | \*+/ )
+			^\s* (?: /\*(?!.*\*/) | \*+/ )
 		]]></string>
 	</dict>
 </dict>
--- a/CSS/Indentation Rules.tmPreferences
+++ b/CSS/Indentation Rules.tmPreferences
@@ -41,7 +41,7 @@
 			#   ST doesn't apply `Indentation Rules - Comments`, because
 			#   comment scope doesn't cover the whole line. So they need
 			#   to be excluded from indentation engine here.
-			^\s* (?: (?: /\* | \*+/ ) .* )?$
+			^\s* (?: /\* | \*+/ )
 		]]></string>
 	</dict>
 </dict>
--- a/CSS/Indentation Rules.tmPreferences
+++ b/CSS/Indentation Rules.tmPreferences
@@ -8,9 +8,35 @@
 		<key>decreaseIndentPattern</key>
 		<string>^\s*[)\]}]</string>
 		<key>increaseIndentPattern</key>
-		<string>^.*[(\[{](?:\s*/\*.*\*/)*\s*(?:/\*.*)?$</string>
+		<string><![CDATA[(?x)
+			# any text
+			^.*
+			# follwed by opening bracket
+			[(\[{]
+			# optionally followed by...
+			(?<balanced>
+			# balanced parenthesis
+			  \( (?: \g<balanced> | [^)] )* \)
+			# balanced brackets
+		  	| \[ (?: \g<balanced> | [^]] )* \]
+			# balanced braces
+			| \{ (?: \g<balanced> | [^}] )* \}
+			# double quoted string with ignored escaped quotation marks
+			| \".*(?<![^\\]\\)(?<![\\]{3})\"
+			# single quoted string with ignored escaped quotation marks
+			| \'.*(?<![^\\]\\)\'
+			# block comment
+			| /\*.*?\*/
+			# anything but closing brackets
+			| [^])}]
+			)*
+			# ... until end of line
+			$
+		]]></string>
 		<key>unIndentedLinePattern</key>
-		<string>^\s*((/\*|\*/).*)?$</string>
+		<string>^\s*((?:/\*|\*/).*)?$</string>
+		<key>preserveIndent</key>
+		<true/>
 	</dict>
 </dict>
 </plist>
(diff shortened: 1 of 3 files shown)


CSS/Indentation Rules.tmPreferences
```diff
@@ -0,0 +1,16 @@
...
+	<key>scope</key>
+	<string>source.css</string>
+	<key>settings</key>
+	<dict>
+		<key>decreaseIndentPattern</key>
+		<string>^\s*[)\]}]</string>
+		<key>increaseIndentPattern</key>
+		<string>^.*[(\[{](?:\s*/\*.*\*/)*\s*(?:/\*.*)?$</string>
+		<key>unIndentedLinePattern</key>
+		<string>^\s*((/\*|\*/).*)?$</string>
```
keith-hall, line 13: is the `.*`...`?$` really necessary?

CSS/Indentation Rules.tmPreferences
```diff
@@ -0,0 +1,16 @@
...
+<plist version="1.0">
+<dict>
+	<key>scope</key>
+	<string>source.css</string>
+	<key>settings</key>
+	<dict>
+		<key>decreaseIndentPattern</key>
+		<string>^\s*[)\]}]</string>
+		<key>increaseIndentPattern</key>
+		<string>^.*[(\[{](?:\s*/\*.*\*/)*\s*(?:/\*.*)?$</string>
```
keith-hall, line 11: do you think it is worth making it an extended regex so we can have it on multiple lines for readability?

deathaxe:
Thise are two lines of code which could have been written years ago and noone cared.

deathaxe:
Added some lines of code to think about.

CSS/Indentation Rules.tmPreferences
```diff
@@ -0,0 +1,42 @@
...
+			# anything but closing brackets
+			| [^])}]
+			)*
+			# ... until end of line
+			$
+		]]></string>
+		<key>unIndentedLinePattern</key>
+		<string>^\s*(?:(?:/\*|\*/).*)?$</string>
+		<key>preserveIndent</key>
+		<true/>
```
deathaxe, line 39: While `preserveIndent` helps with multi-line block strings I have a feeling it breaks detenting text somehow.

```
a {
    color: red;
    }| <- not detended when entered manually, if `preserveIdent = true`
```

deathaxe: The following commit should fix this issue.

CSS/Indentation Rules.tmPreferences
```diff
@@ -36,7 +36,13 @@
 		<key>preserveIndent</key>
 		<false/>
 		<key>unIndentedLinePattern</key>
-		<string>^\s*(?:(?:/\*|\*+/).*)?$</string>
+		<string><![CDATA[(?x)
+			# standalone begin or end of block comments
+			#   ST doesn't apply `Indentation Rules - Comments`, because
+			#   comment scope doesn't cover the whole line. So they need
+			#   to be excluded from indentation engine here.
+			^\s* (?: (?: /\* | \*+/ ) .* )?$
```
keith-hall, line 44: perhaps a nitpick, so feel free to ignore, but I believe for an unclosed `/*`, the `\n` would have the comment scope. And for `*/`, it seems a little odd (and redundant, because the regex pattern doesn't need to match the whole line) to check for `.*$` after it because it might be valid CSS code following a close comment marker which should still affect indentation, no?

Suggesting to use `^\s*$|\*/` instead, so that a comment opened on a prior line and which is closed on the current line would match this rule, with the caveat that it is untested, and perhaps the `^` is implied, and thus `^\s*$|.*\*/` would be necessary.

deathaxe: It took that rule from...

https://github.com/sublimehq/Packages/blob/a423ed195aa8e39566ab84fca16b953b9ae78307/C%23/Indentation%20Rules.tmPreferences#L39-L40

The only thing I found not working was `**/`.

C++ seems to use something like that, too, even though it differs a little bit

https://github.com/sublimehq/Packages/blob/a423ed195aa8e39566ab84fca16b953b9ae78307/C%2B%2B/Indentation%20Rules.tmPreferences#L27-L28

I read about discussions how to solve indentation issues after block comments. Those pointed to this pattern, too.

The main issue with opening comment punctuation seems to be the non-comment scope of the whitespace in front of it, which causes indentation engine to apply `Indentation Rules` rather than `Indentation Rules - Comments` to that line. The newline and maybe the whole comment content probably doen't matter in that case, but due to existing implementations I took it as is.

---

What would you expect the last comment line to be indented like?

```css
/**
 * Test: Trailing comments spanning to next line
 * and containing closing brackets
 **/ .class { color: red; }

.class { color: red; }          /* }
*/
```

I think it needs to be ignored to keep comment formatation intact. As the whole block comment is not indented how should the indentation engine know know what the correct indentation level of the last line is, otherwise?

>  CSS code following a close comment marker which should still affect indentation, no?

Both Java and CSS contain several tests with block comments in between expressions. All of them seem to work.

> Suggesting to use `^\s*$|\*/` instead,

This breaks indentation tests.

```css
/**
 * Test: Complete trailing comment with braces
 **/

.class { color: red; }          /* ; } */
```

The content after block comments would be aligned with the closing `**/` and thus be indented one char too much.

---

I found several indentation rules which all treat the closing `*/` at the beginning of a line, but I am very suspicious about its meaning, function and usefullness.

https://github.com/sublimehq/Packages/blob/a423ed195aa8e39566ab84fca16b953b9ae78307/C%2B%2B/Indentation%20Rules.tmPreferences#L10

Neither indentation tests of Java nor CSS revealed any situation the leading `^ (.*\*/)?` has any useful effect. Maybe I miss some special situations, so we should add them to the indentation tests.

deathaxe: A4 paper filled.

keith-hall (approved):

CSS/Indentation Rules.tmPreferences
keith-hall: Thanks for looking into it and the explanation :)

deathaxe: It appears we can simplify the rule to `^\s* (?: /\* | \*+/ )`. Removing the trailing `.*$` doesn't break tests.

keith-hall (approved):

jfcherng (approved):

PR #2983, opened 2021-08-21
[LaTeX] Apply meta.mapping scopes in BibTeX entries

jwortmann:
To follow the convention from JSON, Git config, and other syntaxes.

### DIFF
--- a/LaTeX/Bibtex.sublime-syntax
+++ b/LaTeX/Bibtex.sublime-syntax
@@ -56,12 +56,13 @@ contexts:
           captures:
             0: punctuation.section.entry.end.bibtex
           pop: true
-        - match: '([a-zA-Z]+)\s*(\=)'
+        - match: '([a-zA-Z]+)(\s*(\=)\s*)'
           captures:
-            1: string.unquoted.key.bibtex
-            2: punctuation.separator.key-value.bibtex
+            1: meta.mapping.key.bibtex string.unquoted.key.bibtex
+            2: meta.mapping.bibtex
+            3: punctuation.separator.key-value.bibtex
           push:
-            - meta_scope: meta.key-assignment.bibtex
+            - meta_content_scope: meta.mapping.value.bibtex
             - match: "(?=[,}])"
               pop: true
             - include: string_content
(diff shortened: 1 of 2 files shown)


LaTeX/Bibtex.sublime-syntax
```diff
@@ -78,12 +79,13 @@ contexts:
           captures:
             0: punctuation.section.entry.end.bibtex
           pop: true
-        - match: '([a-zA-Z]+)\s*(\=)'
+        - match: '([a-zA-Z]+)(\s*(\=)\s*)'
```
jfcherng, line 82: - We don't need to escape `=` in regex.
- I checked [how JSON uses `meta.mapping.json`](https://github.com/sublimehq/Packages/blob/ff909a6922f1407dc10a1424a28af1cf5ed3f126/JSON/JSON.sublime-syntax#L87-L94), but it looks like quite different from this PR. JSON's `meta.mapping.json` applies to the whole object but this PR only applies to `\s*=\s*` which is quite weird.

deathaxe: JSON differes due to the approach being used to apply `meta.mapping` vs. `meta.mapping.[key|value]`. JSON just avoids complex patterns with capture groups to not break highlighting by intermediate newlines.

It just looks like this PR is just to adjust scopes rather than reworking the whole syntax to apply other match strategies.

The key value separator `:` is scoped `meta.mapping.json`, which is the same as we see here.

I am fine with it here.

deathaxe (commented):
What I miss in general is 

1. `meta.mapping` covering the whole section between the `@hook(key,` and the closing `)`.
2. comma between key-value pairs being scoped as `punctuation.separator.mapping.pair`.

LaTeX/Bibtex.sublime-syntax
jfcherng: Thanks for the explanation.

jwortmann:
Yes, this is supposed to be the smallest change required to update the key-value meta scope to the common convention. Therefore I didn't touch the escaped `=` and the meta scope still spans the same region as it did before here.

I guess the different approach for the `meta.mapping` section is required in JSON, because JSON values can have nested arrays/objects and it is common to use linebreaks within those. Probably that approach would be useful here too, but maybe out of scope for this PR?

jfcherng (approved):

deathaxe (commented):
I'd prefer to cover main aspects of `meta.mapping`, which includes the scope to cover the whole section beginning with or after the first comma until closing bracket as well as adding a pattern to scope the `,` accordingly.

Many older syntaxes miss those punctuation scopes and we should add them when appropriate.

deathaxe (approved):

deathaxe:
Thanks

jfcherng (approved):

PR #2987, opened 2021-08-24
[TypeScript] [TSX] Fix arrow function type parameter tests for TSX.

Thom1729:
One of the new tests added in #2923 is correct for plain TypeScript, but wrong for TSX. (#3998 would have caught the bug.)

This PR modifies that test to also be valid for TSX and adds new tests for plain TypeScript and for TSX.

It also deletes some commented code that doesn't need to exist.

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -551,26 +551,6 @@ contexts:
       scope: keyword.operator.comparison.js
       set: expression-begin
 
-  # expression-begin:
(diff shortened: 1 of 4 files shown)


deathaxe (approved):

jfcherng (approved):

keith-hall:
> (#3998 would have caught the bug.)

is that number a typo by any chance?

Thom1729:
Right number, wrong repo. Fixed.

PR #2989, opened 2021-08-24
[Git Formats] highlight diffs in Git Log syntax for `git show`

keith-hall:
first reported at https://github.com/sharkdp/bat/issues/1735

### DIFF
--- a/Git Formats/Git Log.sublime-syntax
+++ b/Git Formats/Git Log.sublime-syntax
@@ -44,7 +44,7 @@ contexts:
       escape: $
     - match: (?=^index\s)
       embed: scope:source.diff
-      escape: (?=^diff\s--git)
+      escape: (?=^(?:commit\s+\h{7,}|diff\s+--git\s))
 
   commit-header:
     # All header attributes are mappings of `key: value` format.
--- a/Git Formats/Git Log.sublime-syntax
+++ b/Git Formats/Git Log.sublime-syntax
@@ -71,6 +71,7 @@ contexts:
 
   extended-patch-headers:
     # https://git-scm.com/docs/git-show#_generating_patch_text_with_p
+    - meta_scope: meta.diff.header.extended.git.log
     - match: ^(index)\s+
       captures:
         1: keyword.context.git.log
--- a/Git Formats/Git Log.sublime-syntax
+++ b/Git Formats/Git Log.sublime-syntax
@@ -23,7 +23,7 @@ contexts:
           scope: meta.header.git.commit markup.raw.block.git.log
           pop: true
           embed: commit-header
-          escape: (?=^commit\s)
+          escape: (?=^(?:commit|diff)\s)
         - match: (\()(HEAD)\s*(->)
           captures:
             1: punctuation.section.parens.begin.git.log
@@ -39,6 +39,12 @@ contexts:
               scope: entity.other.branch-name.git.log
             - match: ','
               scope: punctuation.separator.sequence.git.log
+    - match: (?=^diff\s--git)
+      embed: scope:source.shell.bash
+      escape: $
+    - match: (?=^index\s)
+      embed: scope:source.diff
+      escape: (?=^diff\s--git)
 
   commit-header:
     # All header attributes are mappings of `key: value` format.
--- a/Git Formats/Git Log.sublime-syntax
+++ b/Git Formats/Git Log.sublime-syntax
@@ -2,9 +2,13 @@
 ---
 name: Git Log
 file_extensions:
-  - gitlog   # no real extention but why not?
+  - gitlog   # no real extension but why not?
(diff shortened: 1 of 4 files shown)


Git Formats/Git Log.sublime-syntax
```diff
@@ -39,6 +39,12 @@ contexts:
               scope: entity.other.branch-name.git.log
             - match: ','
               scope: punctuation.separator.sequence.git.log
+    - match: (?=^diff\s--git)
+      embed: scope:source.shell.bash
+      escape: $
+    - match: (?=^index\s)
+      embed: scope:source.diff
+      escape: (?=^diff\s--git)
```
deathaxe, line 47: We probably want to escape if we match the next `commit ...` line`?

Git Formats/tests/syntax_test_git_log
```diff
@@ -74,3 +74,45 @@ Date:   Thu Sep 21 22:53:04 2017 +0200
...
++++ b/JavaScript/tests/syntax_test_typescript_not_tsx.ts
+#^^ source.diff meta.diff.header meta.header.to-file punctuation.definition.to-file
+@@ -25,3 +25,12 @@
+ let strLength: number = (<string>someValue).length; // </string>
+ //                       ^^^^^^^^ meta.assertion - meta.tag
+ //                                                     ^^^^^^^^^ comment - meta.tag
++
++    <T>() => {};
++//  ^^^^^^^^^^^ meta.function
+#^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ source.diff markup.inserted
```
jfcherng, line 118: Not important. But it would be good to add  new line at EOF?

Git Formats/Git Log.sublime-syntax
deathaxe: Maybe we should also be abit lazier with regards to whitespace?

```suggestion
      escape: (?=^(commit\s+\h{7,}|diff\s+--git\s))
```

keith-hall: good idea, done :)

Git Formats/tests/syntax_test_git_log
keith-hall: added :)

Git Formats/tests/syntax_test_git_log
```diff
@@ -74,3 +74,45 @@ Date:   Thu Sep 21 22:53:04 2017 +0200
...
+
+    [TypeScript] [TSX] Fix arrow function type parameter tests for TSX. (#2987)
+
+diff --git a/JavaScript/TypeScript.sublime-syntax b/JavaScript/TypeScript.sublime-syntax
+#^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ source.shell
+#                                                                                       ^ - source.shell
+#^^^ source.shell meta.function-call.identifier variable.function
+#    ^^ punctuation.definition.parameter
+#      ^^^ variable.parameter.option
+index ae1b1061..c1a585d0 100644
```
deathaxe, line 90: `index ...` is not covered by `source.diff`. Would we want to scope it in this syntax?

Git Formats/Git Log.sublime-syntax
```diff
@@ -39,6 +39,12 @@ contexts:
               scope: entity.other.branch-name.git.log
             - match: ','
               scope: punctuation.separator.sequence.git.log
+    - match: (?=^diff\s--git)
```
deathaxe, line 42: Lazy whitespace would also apply here.

```suggestion
    - match: (?=^diff\s+--git)
```

Git Formats/Git Log.sublime-syntax
jfcherng: `\h`?

deathaxe: Would probably enough.

keith-hall: https://github.com/kkos/oniguruma/blob/master/doc/RE#L79

Git Formats/tests/syntax_test_git_log
keith-hall: yep, makes sense to :) done

Git Formats/Git Log.sublime-syntax
keith-hall: done

Git Formats/tests/syntax_test_git_log
deathaxe: Would it make sense to add a meta scope such as `meta.diff.header.extended.git.log` to the whole extended header section?

jfcherng (approved):

PR #2997, opened 2021-08-26
[meta] update templates, use GitHub issue forms

jrappen:
Links to relevant GitHub docs:

- [Syntax for issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)
- [Syntax for GitHub's form schema](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema)


deathaxe:
Thank's for clarification.

I am fine with the current state of those templates and find them a great UX improvement.

One last note/question: I just noticed the "Syntax Highlighting problem" for to be the only one which pre-fills the title field with `[PackageName]: Title`. Would we probably want to do so for all issue types?

jrappen:
Will do on Monday, unless someone beats me to it.

deathaxe (approved):

jrappen:
updated temporary test repo linked above to match current state of this PR

Thom1729:
I'm not sure it's necessary to have all of those checkboxes. Yeah, people should do those things, but it's not like the Packages repo is deluged by low-quality issues. If the checkboxes prevented five spurious reports, but also discouraged one good one, then I think that would be a bad trade-off.

I also think the syntax highlighting one should have a field for the code that's mis-highlighted. I think this is worth adding complexity because it's nearly impossible to address an issue without it. But I suppose it's not absolutely necessary.

jrappen:
> @Thom1729 : I'm not sure it's necessary to have all of those checkboxes.

Got rid of those.

> @Thom1729 : I also think the syntax highlighting one should have a field for the code that's mis-highlighted.

The placeholder for the "what happened?" textarea is markdown with an example that shows how to paste code.

If I would split up the section into markdown and add another multi-line textarea to specify **that** [textarea's `render` option](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#textarea) there it cannot be changed during fill-in to whatever the person reporting wants to use.

jrappen:
Back to using the original Discord link, due to broken redirects.

---

@TheSecEng FYI, from my tests just now:

* <https://discord.sublimetext.io> is broken
* <http://discord.sublimetext.io> is fine

deathaxe (approved):

Ultra-Instinct-05 (approved):

BenjaminSchaaf (approved):

PR #3004, opened 2021-08-27
[Common] Remove duplicate case-insensitive extensions

jfcherng:
As per discussions in https://github.com/sublimehq/Packages/pull/2916

Even on Linux/Mac, ST detects file extension in a case-insensitive manner.

### DIFF
--- a/R/R.sublime-syntax
+++ b/R/R.sublime-syntax
@@ -3,7 +3,7 @@
 # http://www.sublimetext.com/docs/3/syntax.html
 name: R
 file_extensions:
-  - r
+  - R
   - Rprofile
 scope: source.r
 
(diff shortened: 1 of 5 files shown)


R/R.sublime-syntax
```diff
@@ -3,7 +3,6 @@
 # http://www.sublimetext.com/docs/3/syntax.html
 name: R
 file_extensions:
-  - R
   - r
```
keith-hall, line 6: I think the capital version should be kept here, so it would be the default when saving a new file: https://cran.r-project.org/doc/manuals/r-release/R-intro.html#Scripting-with-R

keith-hall (approved):

deathaxe (approved):

Issue #3010, opened 2021-08-30
[HTML] Syntax highlighting misses invalid "-- >"


hijomojo:
Just found that currently if you have a blank space character between the "--" and the ">" in the end tag of an html comment, ST's syntax highlight treats it as a valid end tag, and stops the comment colourisation at that point. This is out of step with the W3 spec, and how browsers treat comment tags, which can result in difficult to find bugs.

Extra note: If the space is between the two "-" then it gets treated as invalid as it should, as does putting any other keyboard character in between the "--" and ">". Bolt result in the tag being treated as invalid and so the comment colourisation continues through to the next valid end tag.

``` html
<!-- 1st comment -- >
<p>Not a comment</p>
<!-- 2nd comment -->
```

![ST HTML comment issue trimmed](https://user-images.githubusercontent.com/38655721/131301044-c8c37c85-9811-4fb4-b0a9-fb1eff1b0470.png)

One solution I've found is changing the following in line 129 of "HTML (Plain).sublime-syntax", under "comment-content:"
Original:
    - match: (<!-)?(--\s*>)
Solution:
    - match: (<!-)?(--*>)

This solution of removing the '\s' works in ST4, and ST3 (where I found the issue). Not sure if it will introduce issues elsewhere, though...I've very little experience with this regex structure.

- Sublime Version: verified in ST4 Portable b4113, found in ST3.2.2 b3211
- OS Version: Windows 10

PR #3011, opened 2021-08-30
[HTML] Fix false positive comment ending punctuation

jfcherng:
Fixes https://github.com/sublimehq/Packages/issues/3010

### DIFF
--- a/HTML/syntax_test_html.html
+++ b/HTML/syntax_test_html.html
@@ -349,7 +349,7 @@
 ##                      ^^^^^^^^^^^^^^^ - invalid.illegal.bad-comments-or-CDATA
 
         <!-- 1st comment -- >
-##                       ^^^^ - punctuation.definition.comment.end.html
+##                       ^^^^ - punctuation
         <p>Is a comment</p>
 ##      ^^^^^^^^^^^^^^^^^^^ comment.block
         <!-- 2nd comment -->
--- a/HTML/syntax_test_html.html
+++ b/HTML/syntax_test_html.html
@@ -348,6 +348,14 @@
 ##          ^^ invalid.illegal.bad-comments-or-CDATA
 ##                      ^^^^^^^^^^^^^^^ - invalid.illegal.bad-comments-or-CDATA
 
+        <!-- 1st comment -- >
+##                       ^^^^ - punctuation.definition.comment.end.html
+        <p>Is a comment</p>
+##      ^^^^^^^^^^^^^^^^^^^ comment.block
+        <!-- 2nd comment -->
+##      ^^^^ invalid.illegal.bad-comments-or-CDATA
+##                       ^^^ punctuation.definition.comment.end.html
+
         <script type=html/text/>
         ##     ^ - meta.attribute-with-value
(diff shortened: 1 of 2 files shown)


HTML/syntax_test_html.html
```diff
@@ -348,6 +348,14 @@
 ##          ^^ invalid.illegal.bad-comments-or-CDATA
 ##                      ^^^^^^^^^^^^^^^ - invalid.illegal.bad-comments-or-CDATA
 
+        <!-- 1st comment -- >
+##                       ^^^^ - punctuation.definition.comment.end.html
```
deathaxe, line 352: It may be a good idea to keep negative test selectors a bit more generic in general to reduce the risk of those not applying after scope changes.

Basically just `- punctuation` should be enough here. So unless we change top-level scope the test keeps valid after all kinds of scope name changes.

It won't probably happen here though, but I left too many obsolete test cases behind during larger refactorings in the past to not apply this guideline whenever possible.

Positive tests should be as precise as possible, to find any renamings.

hijomojo:
Cheers for the quick action ✌
The change to the .sublime-syntax looks good to me, given it matches my suggestion. I'm not familiar with this process, or the other file changed, so I'll hold off on reviewing.

michaelblyons:
> Cheers for the quick action ✌
> The change to the .sublime-syntax looks good to me, given it matches my suggestion. I'm not familiar with this process, or the other file changed, so I'll hold off on reviewing.

The other file [is a unit test](https://www.sublimetext.com/docs/syntax.html#testing) to make sure there aren't regressions. Now that there's a test, it won't come back. I think the bug _may_ have been introduced in https://github.com/sublimehq/Packages/commit/015472d72d1e3ebfaf81b4a681b85967d0c8fc1f, but I'd have to bisect to be sure. (This was just from looking at diffs.)

Thanks for a great bug report.

deathaxe:
This bug dates back to the "initial import". It existed since day one.

deathaxe (approved):

HTML/syntax_test_html.html
```diff
@@ -348,6 +348,14 @@
 ##          ^^ invalid.illegal.bad-comments-or-CDATA
 ##                      ^^^^^^^^^^^^^^^ - invalid.illegal.bad-comments-or-CDATA
 
+        <!-- 1st comment -- >
+##                       ^^^^ - punctuation
+        <p>Is a comment</p>
+##      ^^^^^^^^^^^^^^^^^^^ comment.block
+        <!-- 2nd comment -->
+##      ^^^^ invalid.illegal.bad-comments-or-CDATA
```
keith-hall, line 356: Is this really illegal syntax in HTML5? I know that in XML, `--` is illegal inside of a comment unless it is part of the opening or closing token, but never heard of `<!` being illegal in comments.

jfcherng: https://html.spec.whatwg.org/#comments

deathaxe: This is what Firefox has to say about it ...

grafik

keith-hall (approved):

Issue #630, opened 2016-09-22
[SQL] Incorrect highlighting of a string containing only a \


ciaccia:
### Wrong highlighting with SQL (build 3124)

Consider the following (PostgreSQL) SQL script:

```
CREATE TABLE tab
(
  id bigserial,
  data text
);

INSERT INTO tab(id, data) values (1, '_');
-- CORRECT: This line is grey

INSERT INTO tab(id, data) values (2, '\');
-- WRONG: This line is yellow, it should be *GREY*
```
### Expected behaviour

The last line should be considered a comment an be grey.
### Actual behaviour

The last line is yellow, since the SQL parser does not correctly close the '\' string (it thinks the last single quote is escaped, which in SQL is not).
### Steps to reproduce
1. Open Sublime Text 3124 with standard settings
2. Create a new file, paste the above content in it and save it as `bug.sql`.
3. The last color is grey instead yellow
### Environment
- Operating system and version:
  - Mac OS X 10.11.6
- Sublime Text:
  - Build 3124

keith-hall:
Agreed, in ANSI SQL, single quotes should be doubled up to be escaped (this is working as expected), and a preceding ``\`` character shouldn't affect them.

wbond:
With the current architecture, this will be impossible to support.

Unfortunately one of the most popular databases is MySQL, and it uses `\'` unless certain options are set. Switching to only ANSI escapes and allowing `'\'` would break string quoting in many user's MySQL. Rather than change which flavor of SQL is broken, for now I am deferring to leaving the broken thing the same as it has been for the past few years.

The only way I see around this would be to have a way for a user to switch between a MySQL package and an ANSI SQL package. Since various other syntaxes include the SQL syntax in strings, they would both need to use `source.sql`, and thus only one would be able to be enabled at the same time.

The other downside here is that packages are enabled/disabled globally in Sublime Text, so you would not be able to have MySQL enabled for one project and ANSI SQL for another.

wbond:
@ciaccia Yes, that is a possibility also. MySQL, PostgreSQL, SQLite, SQL Server and Oracle all have enough differences that different flavors would be nice.

This is a similar situation to the shell syntax, where having variations for bash, sh, zsh and tcsh would be handy.

lanrion:
Also wrong in Ruby when use `/`,
[...]

PR #3016, opened 2021-09-04
[SQL] Improve handling of LIKE keyword

keith-hall:
Scope escape characters inside single quoted strings after a LIKE keyword. Also, scope `ESCAPE` and the character after it. Fixes https://github.com/sublimehq/Packages/issues/630#issuecomment-910532251.

We could improve on this in a follow-up PR to use branching to detect a strict set of escape characters (specifically, `\`) so that the scoping inside the `LIKE` string would be correct.

### DIFF
--- a/SQL/SQL.sublime-syntax
+++ b/SQL/SQL.sublime-syntax
@@ -96,8 +96,14 @@ contexts:
         1: keyword.operator.logical.sql
         2: keyword.operator.logical.sql
         3: constant.language.sql
-    - match: (?i:\b(and|or|like|having|exists|between|in)\b)
+    - match: (?i:\b(and|or|having|exists|between|in)\b)
       scope: keyword.operator.logical.sql
+    - match: (?i:\blike\b)
+      scope: keyword.operator.logical.sql
+      push: like_string
+    - match: (?i:\bescape\b)
+      scope: keyword.operator.sql
+      push: like_escape
     - match: (?i:\bvalues\b)
       scope: keyword.other.DML.II.sql
     - match: (?i:\b(begin(\s+work)?|start\s+transaction|commit(\s+work)?|rollback(\s+work)?)\b)
@@ -243,3 +249,31 @@ contexts:
       # Handle situations where the schema and . 
     - match: '{{end_identifier}}'
       pop: true
+  like_escape:
+    - match: (\')([^'])(\')
+      captures:
+        0: meta.string.escape.sql string.quoted.single.sql
+        1: punctuation.definition.string.begin.sql
+        2: constant.character.escape.sql
+        3: punctuation.definition.string.end.sql
+      pop: true
+    - match: (?=\S)
+      pop: true
+  like_string:
+    - match: \'
+      scope: punctuation.definition.string.begin.sql
+      set: inside_like_single_quoted_string
+    - match: (?=\S)
+      pop: true
+  inside_like_single_quoted_string:
+    - meta_scope: meta.string.like.sql string.quoted.single.sql
+    - match: \'
+      scope: punctuation.definition.string.end.sql
+      pop: true
+    - match: (\[)(?:.|[^]']+?)(\])
+      scope: meta.set.like.sql
+      captures:
+        1: keyword.control.set.begin.sql
+        2: keyword.control.set.end.sql
+    - match: '[%_]'
+      scope: keyword.other.any.sql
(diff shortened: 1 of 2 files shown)


michaelblyons:
> We could improve on this in a follow-up PR to use branching to detect a strict set of escape characters (specifically, `\`) so that the scoping inside the `LIKE` string would be correct.

The other escape character I have seen in more-than-niche use is `^`.

SQL/SQL.sublime-syntax
```diff
@@ -96,8 +96,14 @@ contexts:
...
         2: keyword.operator.logical.sql
         3: constant.language.sql
-    - match: (?i:\b(and|or|like|having|exists|between|in)\b)
+    - match: (?i:\b(and|or|having|exists|between|in)\b)
       scope: keyword.operator.logical.sql
+    - match: (?i:\blike\b)
+      scope: keyword.operator.logical.sql
+      push: like_string
+    - match: (?i:\bescape\b)
+      scope: keyword.operator.sql
```
deathaxe, line 105: Word like operators may wanted to be highlighted differently from symbol like operators. 

I have Breackets Color Scheme in mind for instance, which might want to highlight ESCAPE like the other keywords (in blue).

grafik

A third-level scope may help. Not sure which one as I haven't seen such a use case yet.

SQL/SQL.sublime-syntax
```diff
@@ -243,3 +249,31 @@ contexts:
...
+    - match: \'
+      scope: punctuation.definition.string.end.sql
+      pop: true
+    - match: (\[)(?:.|[^]']+?)(\])
+      scope: meta.set.like.sql
+      captures:
+        1: keyword.control.set.begin.sql
+        2: keyword.control.set.end.sql
+    - match: '[%_]'
+      scope: keyword.other.any.sql
```
deathaxe, line 279: Looks like `%` and `_` are wildcards just like `*` and `?` in path patterns?

see: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver15

Maybe one of the following scopes would be appropriate then:
- `variable.language.wildcard`
- `keyword.operator.wildcard`

Both are used in several syntaxes, already. Not sure which one would be the better choise, maybe the latter?

SQL/SQL.sublime-syntax
```diff
@@ -243,3 +249,31 @@ contexts:
...
+      scope: punctuation.definition.string.begin.sql
+      set: inside_like_single_quoted_string
+    - match: (?=\S)
+      pop: true
+  inside_like_single_quoted_string:
+    - meta_scope: meta.string.like.sql string.quoted.single.sql
+    - match: \'
+      scope: punctuation.definition.string.end.sql
+      pop: true
+    - match: (\[)(?:.|[^]']+?)(\])
```
deathaxe, line 273: It seems, sets behave pretty much like regexp character groups. Do we want to honor it?

keith-hall: They do indeed, though much more limited as I understand. It seems it is just an optional negation operator `^` (which I need to update this PR to scope correctly) followed by a set of characters. Nothing fancy with `-` or `&` etc.
https://codingsight.com/t-sql-regular-expression-like-operator-and-its-use-cases/

SQL/SQL.sublime-syntax
keith-hall: I've made it a `keyword.operator.word` for now in case it helps, but happy to revise.

SQL/SQL.sublime-syntax
keith-hall: FWIW, I took `keyword.other.any` from https://github.com/sublimehq/Packages/blob/cd881081814883bdceecdcb8bdc2343cc46bc639/Regular%20Expressions/RegExp.sublime-syntax#L352 - maybe it would also make sense to standardize this one to `keyword.operator.wildcard`.

SQL/SQL.sublime-syntax
keith-hall: It would be nice to be able to just re-use the Regex syntax def for this, but I guess we'd need to wait for https://github.com/sublimehq/Packages/issues/1942#issuecomment-905777773 first.

deathaxe (approved):

SQL/SQL.sublime-syntax
deathaxe: There are so many things to doooo....

SQL/SQL.sublime-syntax
deathaxe: Let's go with it for now.

jfcherng (approved):

tjasko:
Hey @keith-hall, thanks for knocking this out. 🙂

PR #3018, opened 2021-09-05
[JavaScript/TypeScript] Improve shebang and comments highlighting

### LINKED ISSUE #2880: [TypeScript] Can't match triple slash directives in the Typescript
- Sublime Version: 4100
- OS Version: win 10

``` typescript
/// <reference path="..." />
/// <reference path="..." />
```
before
图片
after:
图片

```yaml
ts-triple-slash-directives:
    - match: '///'
      scope: punctuation.definition.comment.ts
      push:
        - include: Packages/XML/XML.sublime-syntax#tag
        - match: $
          pop: true
```

deathaxe:
Fixes #2880

This PR implements  ...

1. shebang highlighting using the scheme of PHP/Java/ShellScript to ensure it being highlighted at the very first line, only.
2. xml highlighting in triple-slash comments in TS/TSX syntax according to Typescript specs, but with a more lazy/forgiving xml highlighting.
3. more forgiving comment punctuation highlighting in order to help ligatures for `//` or `///` to be displayed and to improve UX in case color schemes apply special highlighting to comment punctuation.
4. scopes trailing `//` in line comments `punctuation` for consistent highlighting.
5. adds various comment related test cases to illustrate changes and differences between syntaxes.

It feels odd to only highlight

-  `/*` and `*/` as punctuation in delimiters like `/***************************/` 
-  `//` in `/// comment` or `/////////////////////////////////////`

Note:

If `double-slash` and `triple-slash` sub-scopes are not needed, all line comment related contexts could probably be simplified/merged in some way.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -124,33 +124,38 @@ contexts:
     - include: comments
 
   comments:
-    - match: /\*\*(?!/)
+    - include: line-comments
+    - include: block-comments
+
+  block-comments:
+    # empty block comments
+    - match: (/\*\*+/)
+      scope: comment.block.empty.java punctuation.definition.comment.java
+    # documentation block comments
+    - match: /\*\*+
       scope: punctuation.definition.comment.begin.js
       push:
-        - jsdoc-comment
+        - jsdoc-comment-body
         - jsdoc-block-tag
+    # normal block comments
     - match: /\*
       scope: punctuation.definition.comment.begin.js
-      push:
-        - meta_include_prototype: false
-        - meta_scope: comment.block.js
-        - match: \*/
-          scope: punctuation.definition.comment.end.js
-          pop: true
-    - match: //
-      scope: punctuation.definition.comment.js
-      push:
-        - meta_include_prototype: false
-        - meta_scope: comment.line.double-slash.js
-        - match: \n
-          pop: true
+      push: block-comment-body
 
-  jsdoc-comment:
+  block-comment-end:
+    - match: \*+/
+      scope: punctuation.definition.comment.end.js
+      pop: true
+
+  block-comment-body:
+    - meta_include_prototype: false
+    - meta_scope: comment.block.js
+    - include: block-comment-end
+
+  jsdoc-comment-body:
     - meta_include_prototype: false
     - meta_scope: comment.block.documentation.js
-    - match: \*/
-      scope: punctuation.definition.comment.end.js
-      pop: 1
+    - include: block-comment-end
     # JSDoc "block" tags (i.e. @param) are only accepted at the beginning of the documentation
     # line so directly after the '/**' or after the '*' marker on the next lines.
     - match: ^\s*(\*)(?!/)
@@ -163,6 +168,35 @@ contexts:
       scope: entity.other.attribute-name.documentation.js
       pop: 1
     - match: (?=\S)|$
+      pop: 1
+
+  line-comments:
+    - match: /{4,}
+      scope: punctuation.definition.comment.js
+      push: line-comment-other-body
+    - match: /{3}
+      scope: punctuation.definition.comment.js
+      push: line-comment-triple-slash-body
+    - match: /{2}
+      scope: punctuation.definition.comment.js
+      push: line-comment-double-slash-body
+
+  line-comment-double-slash-body:
+    - meta_include_prototype: false
+    - meta_scope: comment.line.double-slash.js
+    - match: \n
+      pop: 1
+
+  line-comment-triple-slash-body:
+    - meta_include_prototype: false
+    - meta_scope: comment.line.triple-slash.js
+    - match: \n
+      pop: 1
+
+  line-comment-other-body:
+    - meta_include_prototype: false
+    - meta_scope: comment.line.other.js
+    - match: \n
       pop: 1
 
   shebang:
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -109,7 +109,7 @@ variables:
 
 contexts:
   main:
-    - meta_include_prototype: false # don't match before shebang
+    - meta_include_prototype: false # don't match comments before shebang
     - match: ''
       push: [javascript, shebang]
 
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -181,23 +181,26 @@ contexts:
       scope: punctuation.definition.comment.js
       push: line-comment-double-slash-body
 
+  line-comment-end:
+    - match: (//+)?\n
+      captures:
+        1: punctuation.definition.comment.js
+      pop: 1
+
   line-comment-double-slash-body:
     - meta_include_prototype: false
     - meta_scope: comment.line.double-slash.js
-    - match: \n
-      pop: 1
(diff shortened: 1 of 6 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -162,11 +170,51 @@ contexts:
...
+    - meta_scope: comment.line.other.js
+    - match: \n
+      pop: 1
+
+  shebang:
+    - meta_include_prototype: false
+    - match: ^\#!
+      scope: punctuation.definition.comment.js
+      set: shebang-body
+    - match: ^|(?=\S)  # Note: Ensure to highlight shebang if Erlang is embedded.
```
wbond, line 207: Not sure I understand this comment about Erlang

wbond (dismissed):
Looks good to me. Only thing I wasn't sure about was the Erlang comment, which I am guessing was a copy-paste error?

JavaScript/JavaScript.sublime-syntax
deathaxe: I guess it's just a copy and paste issue as this construct is used in several syntaxes.

deathaxe:
Fixed the typo. While doing so I found Java (PR) to also highlight trailing slashes in line comments. Added it as it might look better to have all slashes scoped the same in `// title ///////`.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -109,8 +109,11 @@ variables:
 
 contexts:
   main:
-    - include: comments-top-level
+    - meta_include_prototype: false # don't match comments before shebang
+    - match: ''
+      push: [javascript, shebang]
 
+  javascript:
```
Thom1729, line 116: Maybe just `script` for the name of this context? Might be less confusing in the inherited syntaxes.

JavaScript/tests/syntax_test_jsx.jsx
```diff
@@ -1,5 +1,86 @@
 // SYNTAX TEST "Packages/JavaScript/JSX.sublime-syntax"
 
+// comment
```
Thom1729, line 3: Why duplicate the tests in other files?

Thom1729 (approved):

JavaScript/tests/syntax_test_jsx.jsx
deathaxe: I just copied the whole bunch of comments from typescript to the other ones and adjusted rerequired stuff. May probably not needed in the ...X files at the moment, but it may not hurt to make sure not to regress something in case special treatment in react comments is needed in future.

I even prevented myself by regressing something by doing so (see the most recent force-pushes) ;-)

keith-hall:
> Fixed the typo. While doing so I found Java (PR) to also highlight trailing slashes in line comments. Added it as it might look better to have all slashes scoped the same in `// title ///////`.

Just a quick check - I recall that `wrap_line` functionality relies on the comment punctuation characters being scoped. Have you tested to ensure that nothing breaks when more characters in the comment receive these scopes (both leading and trailing punctuation)?

JavaScript/JavaScript.sublime-syntax
```diff
@@ -121,33 +124,38 @@ contexts:
...
 
   comments:
-    - match: /\*\*(?!/)
+    - include: line-comments
+    - include: block-comments
+
+  block-comments:
+    # empty block comments
+    - match: (/\*\*+/)
+      scope: comment.block.empty.java punctuation.definition.comment.java
```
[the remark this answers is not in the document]
deathaxe: Review test passed! 🤣

JavaScript/JavaScript.sublime-syntax
```diff
@@ -121,33 +124,38 @@ contexts:
     - include: comments
 
   comments:
-    - match: /\*\*(?!/)
+    - include: line-comments
+    - include: block-comments
+
+  block-comments:
+    # empty block comments
+    - match: (/\*\*+/)
```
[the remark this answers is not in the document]
deathaxe: Yep. Resident of another quick test. Removed.

deathaxe:
Tried `wrap_line` in C and C# in SAFE MODE and it doesn't add another `//` when hitting enter at the end of a line comment. The only thing I've seen is `* ` being added  in block comments.

I have no strong opinion in applying such stuff to line comments. I just added it here, because there was another syntax (forgot which one), I had a discussion about different types of line comments, which ended up in exactly such rules.

Scoping trailing slashes or all is just to highlight "delimiting comments" which span the whole line, consistently. Maybe it's not worth it. Thoughts?

We should at least distinguish double-slash and triple-slash in a sane and consistent way as both have different meaning in TypeScript and most fonts provide ligatures for them.

keith-hall:
> Scoping trailing slashes or all is just to highlight "delimiting comments" which span the whole line, consistently. Maybe it's not worth it. Thoughts?

I personally like the idea, there may be color schemes which highlight a comment's punctuation differently from the comment text

> Which one looks better??

Consistency wins IMHO i.e. your current JavaScript :)

keith-hall (approved):

JavaScript/JavaScript.sublime-syntax
```diff
@@ -162,11 +170,54 @@ contexts:
...
+      push: line-comment-other-body
+    - match: /{3}
+      scope: punctuation.definition.comment.js
+      push: line-comment-triple-slash-body
+    - match: /{2}
+      scope: punctuation.definition.comment.js
+      push: line-comment-double-slash-body
+
+  line-comment-end:
+    - match: (//+)?\n
```
Thom1729, line 185:
```suggestion
    - match: (?://+)?\n
```

JavaScript/JavaScript.sublime-syntax
```diff
@@ -162,11 +170,54 @@ contexts:
...
+      scope: punctuation.definition.comment.js
+      set: shebang-body
+    - match: ^|(?=\S)  # Note: Ensure to highlight shebang if the syntax is embedded.
+      pop: 1
+
+  shebang-body:
+    - meta_include_prototype: false
+    - meta_scope: comment.line.shebang.js
+    # Note: Keep sync with first_line_match!
+    - match: \b(node|js)\b
```
Thom1729, line 217:
```suggestion
    - match: \b(?:node|js)\b
```

Thom1729 (approved):

JavaScript/JavaScript.sublime-syntax
deathaxe: 1. I am not sure about whether we should scope the newline character `punctuation`
2. This change would break syntax due to following `caputures 1`

JavaScript/JavaScript.sublime-syntax
deathaxe: All shebang using syntaxes including `first_line_match` have this capture group. Changing it here would mean to also go though all the others as well. Not sure whether it's worth it as it is only the first line which is handled here. It doesn't have any effect on performance.

JavaScript/JavaScript.sublime-syntax
Thom1729: Oops, my bad.

JavaScript/JavaScript.sublime-syntax
Thom1729: Makes sense.

deathaxe:
Thank's for having a look on it.

PR #3026, opened 2021-09-07
[SQL] use branch points to handle LIKE ... ESCAPE with \ or ^ chars

keith-hall:


deathaxe:
Looks like `comments` evolve to a candidate for `prototype`.

SQL/SQL.sublime-syntax
```diff
@@ -249,7 +251,52 @@ contexts:
...
+  like_string_followed_by_unknown_escape:
+    - include: comments
+    - match: \'
+      scope: punctuation.definition.string.begin.sql
+      push: inside_like_single_quoted_string
+    - match: (?i:\bescape\b)
+      scope: keyword.operator.word.sql
+      set: like_escape_any
+    - match: (?=\S)
+      pop: true
```
deathaxe, line 297: Would it make sense to use the same structure as for the other contexts?

```suggestion
  like_string_followed_by_unknown_escape:
    - include: comments
    - match: \'
      scope: punctuation.definition.string.begin.sql
      set: [like_escape_any, like_escape_pop, inside_like_single_quoted_string]
    - match: (?=\S)
      pop: true
```

SQL/SQL.sublime-syntax
```diff
@@ -259,12 +306,28 @@ contexts:
       pop: true
     - match: (?=\S)
       pop: true
-  like_string:
-    - match: \'
-      scope: punctuation.definition.string.begin.sql
-      set: inside_like_single_quoted_string
+  like_escape_slash:
```
deathaxe, line 309: The name of those contexts conflicts with `like_escape_fail` and `like_escape_pop` somehow as those handle behavior on `ESCAPE` keywords while this one is about the ecape character.

Maybe this difference in meaning should be reflected by the contexts' names?

SQL/SQL.sublime-syntax
```diff
@@ -259,12 +306,28 @@ contexts:
...
+    - match: (\')(\^)(\')
+      captures:
+        0: meta.string.escape.sql string.quoted.single.sql
+        1: punctuation.definition.string.begin.sql
+        2: constant.character.escape.sql
+        3: punctuation.definition.string.end.sql
       pop: true
+    - match: (?=\S)
+      fail: like_string_branch
   inside_like_single_quoted_string:
```
deathaxe, line 331: I found it useful to keep those body contexts grouped together with those which push them on stack, especially if they are used once or a few times only.

SQL/SQL.sublime-syntax
```diff
@@ -249,7 +251,52 @@ contexts:
...
+      fail: like_string_branch
+    - match: (?=\S)
+      pop: true
+  like_escape_pop:
+    - include: comments
+    - match: (?i:\bescape\b)
+      scope: keyword.operator.word.sql
+      pop: true
+    - match: (?=\S)
+      pop: true
```
deathaxe, line 273: I'd move those contexts down to the other `like_escape_...` contexts to keep `like_string_...` contexts logically grouped together and sorted by "complexity".

A syntax should probably always start with some managing contexts containing only includes, followed by complex contexts representing statements  etc. and finalized by simple contexts (prototypes?) which are re-used often - just like a tree. The `like_escape_fail` are leaf contexts as they don't push others onto stack nor do they include any other contexts (if we ignore the comments ;-) )

deathaxe:
A quick edit on what's in my mind https://github.com/deathaxe/sublime-Packages/tree/sql_like_escape

keith-hall:
I integrated your reorganization improvements, thanks @deathaxe!

About a fallback pattern to highlight the ESCAPE keyword, I'm not sure - it's technically not a valid keyword in that position and shouldn't occur while typing so I'm tempted to say we should leave it unscoped.

deathaxe:
I just asked, because the other keywords are scoped as such in any position. I haven't double checked output of an sql server and whether it would interprete ESCAPE in that position as unquoted text. If in doupt, then we should leave it as is.

SQL/SQL.sublime-syntax
```diff
@@ -181,15 +196,18 @@ contexts:
...
       scope: punctuation.definition.string.begin.sql
       push:
+        - meta_include_prototype: false
         - meta_scope: string.regexp.modr.sql
         - match: '\}'
           scope: punctuation.definition.string.end.sql
           pop: true
         - include: string_interpolation
   string_escape:
+    - meta_include_prototype: false
```
deathaxe, line 206: `meta_include_protoype` is not crucial for contexts which are not pushed/set on stack directly, but only included via `- include`.

SQL/SQL.sublime-syntax
```diff
@@ -249,23 +271,59 @@ contexts:
...
     - match: '{{end_identifier}}'
       pop: true
-  like_escape:
-    - match: (\')([^'])(\')
-      captures:
-        0: meta.string.escape.sql string.quoted.single.sql
-        1: punctuation.definition.string.begin.sql
-        2: constant.character.escape.sql
-        3: punctuation.definition.string.end.sql
+  create-condition:
```
deathaxe, line 274: Do we want to rename all context so they use dashes any time soon? Maybe this and adding an empty line between each, to make the syntax definition look more modern.

Doesn't need to happen in this PR, but maybe in a "maintanance follow-up"?

keith-hall: sounds good to me :)

SQL/SQL.sublime-syntax
keith-hall: yep, I was just being defensive here in case this context is somehow pushed onto the stack directly in future or from a different syntax. Unlikely, I know :)

SQL/SQL.sublime-syntax
keith-hall: I renamed the contexts as part of this PR, most of them were introduced here anyway ;)

deathaxe (approved):

jfcherng (approved):

PR #3031, opened 2021-09-09
[OCaml] Fix bug when ocamlyacc file starts with comment

iamkroot:
If the ocamlyacc file begins with a multiline comment (`/*`), the `include: ocaml` rule gets activated, and parses it as a `/` operator followed by more pure ocaml syntax. By moving the `include: comments` rule higher up, it will be given precedence.

Before:
before
After:
after

Test file: https://github.com/roehst/tapl-implementations/blob/master/recon/parser.mly

(Side note: It seems even the github syntax highlighter is broken - the comments aren't getting highlighted properly)

### DIFF
--- a/OCaml/OCamlyacc.sublime-syntax
+++ b/OCaml/OCamlyacc.sublime-syntax
@@ -35,8 +35,8 @@ contexts:
           pop: true
         - include: comments
         - include: rules
-    - include: scope:source.ocaml
     - include: comments
+    - include: scope:source.ocaml
     - match: (’|‘|“|”)
       scope: invalid.illegal.unrecognized-character.ocaml
   comments:
--- a/OCaml/syntax_test_mly.mly
+++ b/OCaml/syntax_test_mly.mly
@@ -1,0 +1,10 @@
+/* SYNTAX TEST "Packages/OCaml/OCamlyacc.sublime-syntax" */
+/*  This is a comment
+ *  at the beginning of the file
+/* ^ comment */
+ */
+
+%{
+open Support.Error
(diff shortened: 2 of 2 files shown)


michaelblyons:
@keith-hall Can you do syntax tests on the first line? Or does this actually work because the first line wants to be a comment?

iamkroot:
> Can you do syntax tests on the first line? Or does this actually work because the first line wants to be a comment?

I guess the following para of the docs answers this:
> Each test in the syntax test file must first start the comment token (established on the first line, it doesn't actually have to be a comment according to the syntax), and then either a ^ or <- token. 
 
If I understand this correctly, this means that whatever I use to begin the first line is considered the comment specifier for the rest of the syntax test file.

iamkroot:
It isn't a first-line-only problem though. 
1. The problem occurs whenever the doc contains a `/*` (which activates `source.ocaml` in unchanged syntax) _before_ the `%{` (which activates the ocamlyacc node). 
2. If there are `/*` comments _after_ the `%}` token, even the existing syntax can parse them correctly. (Not exactly sure how this worked)
3. I ran the current test with and without the proposed syntax change, and it passed/failed as expected.

michaelblyons:
> whenever the doc contains a `/*` (which activates source.ocaml in unchanged syntax) before the `%{`

My mistake. Thanks for clarifying.

iamkroot:
Yeah, my original PR description does seem to hint that it is a first-line-only bug.

keith-hall (approved):

deathaxe (approved):

PR #3034, opened 2021-09-10
[Regular Expressions] Introduce Regex Common syntax and extend it

keith-hall:
Initial work on https://github.com/sublimehq/Packages/issues/1942#issuecomment-905777773

I tried to separate out anything that is specific to Boost, but it may make sense to have more contexts in "common" for re-use by other dialects, even if they are unreferenced as part of the common syntax itself.  It's also possible left something specific to Boost in the base syntax by accident. 

Would love to find a way to avoid the duplication of match patterns in the `group-start` and `group-start-extended` contexts...


keith-hall:
Interesting that I had to change the reference to `#base-literal` to the common/base syntax in Clojure - I kinda expected the extended syntax to contain all the contexts of the base syntax.

Clojure/Clojure.sublime-syntax
```diff
@@ -255,7 +255,7 @@ contexts:
       scope: punctuation.definition.string.end.clojure
       pop: true
     - match: ''
-      push: scope:source.regexp#base-literal
+      push: scope:source.regexp-common#base-literal
```
deathaxe, line 258: Maybe I am missing something, but this line doesn't change anything on my box. With or without the change, Clojure syntax tests pass. I've even renamed the shipped "Regular Expressions.sublime-package".

keith-hall: It seemed to fail on the CI runner, though...

deathaxe: Even runs fine on CI (https://github.com/deathaxe/sublime-packages/tree/debug_CI)

keith-hall: My bad, the CI failure I observed was when I had pushed my changes to `Regexp.sublime-syntax` as the "common" syntax, with the Boost specific stuff in a separate definition, in an attempt to get a nicer diff. Sorry for the confusion and misinformation. I'll revert that commit then because it would be a breaking change for Clojure regexps to only highlight common syntax.

jrappen:
Looks fine at first look.

- Was there a reason for **not** moving to syntax `version: 2` and `pop: 1`?
- Maybe it would make sense to split up the variables regexes into lines and add comments, since you add comments everywhere else. Reference pointers as in links to docs might help here. Maybe not, though.

keith-hall:
>     * Was there a reason for **not** moving to syntax `version: 2` and `pop: 1`?

Mainly I just forgot, the new syntax template doesn't specify version 2 by default, even though it probably should.

Regular Expressions/RegExp Common.sublime-syntax
```diff
@@ -0,0 +1,275 @@
...
+      scope: keyword.operator.alternation.regexp
+      push: unexpected-quantifier-pop
+
+  literal:
+    - include: quantifiers
+    - match: \.
+      scope: keyword.other.any.regexp # https://github.com/sublimehq/Packages/issues/314
+    - match: \)
+      scope: invalid.illegal.unmatched-brace.regexp
+    - match: \w+|.
```
deathaxe, line 274: I'd avoid matching arbritary single characters like `|.` does, because the creates a token for every single - in this case non-word - character then.

See:

grafik

By commenting out that pattern all `#` form a single token and the test file parses in 0,9ms instead of 1,2ms. It's not enough for a fair benchmark, I know.

grafik

Regular Expressions/RegExp Common.sublime-syntax
```diff
@@ -0,0 +1,275 @@
+%YAML 1.2
+---
+name: Regular Expression
+hidden: true
+scope: source.regexp-common
```
deathaxe, line 5: Can we avoid using dashes in scope names? It's barely used and doesn't feel that good.

We used `common` in Git Formats to specify a file which is re-used by many but does not work as standalone syntax.

If I understand your PR comment, this one is a standalone PCRE syntax without boost specific stuff?

If so, I'd suggest `source.regexp.pcre` for this file.

Once we've learned about commonly used stuff between several dialects (python, php, c#, ...) we might introduce a `source.regexp.basic` or even a `source.regexp.common` for shared contexts?

Regular Expressions/RegExp Common.sublime-syntax
```diff
@@ -0,0 +1,275 @@
...
+      captures:
+        1: storage.modifier.mode.regexp
+        2: keyword.control.group.regexp
+      set: [base-group, unexpected-quantifier-pop]
+    # Groups without 'x' mode
+    - match: '\?[ims]*-[ims]*x[imxs]*:'
+      scope: storage.modifier.mode.regexp
+      set: [group-body, unexpected-quantifier-pop]
+    # Other modifiers
+    - match: \?(?:[ixms]*-)?[ims]+:'
```
deathaxe, line 145:
```suggestion
    - match: '\?(?:[ixms]*-)?[ims]+:'
```

Regular Expressions/RegExp Common.sublime-syntax
keith-hall: > Once we've learned about commonly used stuff between several dialects (python, php, c#, ...) we might introduce a `source.regexp.basic` or even a `source.regexp.common` for shared contexts?

This is actually already what I was trying to do, sorry if something I wrote was misleading - `Regexp Common.sublime-syntax` is supposed to represent what is common in all regex engines, so I think `source.regexp.basic` would perhaps make sense.

My idea was to maybe also have *unused* contexts for some things which are not common for all dialects, but shared in some. These could then be included/referenced by the extended syntax for the dialects which use them.

Regular Expressions/RegExp Common.sublime-syntax
keith-hall: makes sense. Should we perhaps remove the `meta.literal` altogether then? If it is needed by a plugin, anything unscoped is a literal ;)

deathaxe: A pattern is normally scoped `string.regexp`, `meta.string.regexp string.quoted` or `string.quoted source.regexp`, which is some kind of literal with all regexp specific operators stacked on top. 

Hence it's not crucial.

I'd vote to remove it.

Regular Expressions/RegExp.sublime-syntax
```diff
@@ -220,90 +99,20 @@ contexts:
     - match: '\?(?=\(\?)'
       scope: keyword.other.conditional.regexp
       set: [group-body-extended, base-group-extended]
```
deathaxe, line 101: I don't think this multi-set is required here.

```suggestion
      set: group-body-extended
```

Regular Expressions/RegExp.sublime-syntax
```diff
@@ -168,29 +67,9 @@ contexts:
     - match: '\?(?=\(\?)'
       scope: keyword.other.conditional.regexp
       set: [group-body, base-group]
```
deathaxe, line 69: I don't think this multi-set is required here, because they are literly the same, except `group-body`  consumes the closing paren when popping.

```suggestion
      set: group-body
```

Regular Expressions/RegExp (Basic).sublime-syntax
```diff
@@ -0,0 +1,280 @@
...
+    - match: '-'
+      scope: punctuation.separator.sequence.regexp-common
+      pop: 1
+    - include: escaped-chars
+
+  charset-range-end:
+    - meta_scope: constant.other.range.regexp
+    - include: escaped-char
+    - match: .
+      pop: 1
```
deathaxe, line 201: With a look at PHP and Python, those use a single pattern with capture groups to match charset-ranges.

Do you think it makes sense to do so, too?

I've created a prototype for it and it performs slightly faster than this lookahead construct. It would make the popping `escaped-char` context obsolete, but result in some more scope assignments due to capturing groups.

Thoughts?

Regular Expressions/RegExp (Basic).sublime-syntax
```diff
@@ -0,0 +1,280 @@
...
+    - match: '{{invalid_char_escape}}'
+      scope: invalid.illegal.character.escape.regexp
+      pop: 1
+    - match: '{{char_escape}}'
+      scope: constant.character.escape.regexp
+      pop: 1
+
+  anchors:
+    - match: \\[bBAZzG]|[\^$]
```
deathaxe, line 227: Do we want to create a variable for this pattern, too? Might help inherited syntaxes to override just the pattern without the need to address the whole context.

Regular Expressions/RegExp (Basic).sublime-syntax
```diff
@@ -0,0 +1,280 @@
...
+      scope: keyword.control.anchors.regexp
+      push: maybe-unexpected-quantifiers
+
+  backrefs:
+    - match: \\([1-9]\d*)
```
deathaxe, line 232: Same here.

Regular Expressions/RegExp (Basic).sublime-syntax
keith-hall: I'm of two minds about this - having a variable that is only used in one place doesn't add much value to the base syntax. And overriding it in the extended syntax makes it less obvious what was changed/added - one would have to compare it to the variable in the base syntax. So generally I prefer to append/prepend rules to the context, especially as then one can change the scope suffix to make it clear which syntax it came from.

But making things easy to modify with variables simplifies extending - no need to understand the contexts, or anything in the base syntax.

deathaxe: That's the general issue with variables, yes. Especially if they contain capture groups. The context is small enough to be easily replaced by an inherited syntax. Maybe just appended/prepended to add more stuff.

I have no strong opinion about it, while sharing your concerns. It was just an idea due to all the other variables already existing.

Regular Expressions/RegExp (Basic).sublime-syntax
keith-hall: No strong opinions here, just a few questions to help us decide which approach to take:

- is (slightly) better performance our main goal?
- is easier maintenance via less repeated scopes (and more re-used contexts) our main goal?
- if it is a mix between the two, does the performance difference outweigh the "duplication"?
- which is easier to understand - a complex lookahead which is then broken down into match patterns / context includes, or a consuming regex with many capture groups?

I'd personally even want to consider using a branch point instead of a lookahead or a single regex, to make it easier to follow what is happening, prevent duplication of scope assignment in various places etc. But that may make performance even worse than the current solution, not sure.

deathaxe: That's correct. I have the same questions in mind.

Additionally I think about the theoretical fact of what' happens happens if the syntax is included via `with_prototype`.

My prototype lookes as follows:

```suggestion
  charset-ranges:
    - match: |-
        (?x)
        (?: ({{known_char_escape}}) | ({{invalid_char_escape}}) | ({{char_escape}}) | ((?!\\-)[^\]]) )
        (-)
        (?: ({{known_char_escape}}) | ({{invalid_char_escape}}) | ({{char_escape}}) | ([^\]]) )
      captures:
        0: meta.range.regexp
        1: constant.character.escape.regexp
        2: invalid.illegal.character.escape.regexp
        3: constant.character.escape.regexp
        4: constant.character.literal.regexp
        5: punctuation.separator.sequence.regexp-common
        6: constant.character.escape.regexp
        7: invalid.illegal.character.escape.regexp
        8: constant.character.escape.regexp
        9: constant.character.literal.regexp
```

1. While I saw some differences on smaller test files, the performance impract in 5000x lines of following pattern negligible. 

   ```
   [b-c] [\x00-\x{A}] [[a-z]&&[:ascii:]] [^abc\w0-9] [a-c-x] []-] [-a-n] [\-a] [a-z\-9] [\--z] [a-\z] [a-\-]
   ```

2. With regards to maintenance and easier to understand:

   The capture groups are not more complicated than all the numeric constants stuff. It's probably even more straight forward compared to pushing those contexts. The one way or another, I don't see much difference in complexity or readability.

   _Note: `escaped-char` would be obsolete._

3. `with_prototype` impact

   Those capture groups are atomic and can't be broken by injected extra patterns, while those `charset-range-begin` and `charset-range-end` contexts would be prepanded by patterns, which would never match due to the lookahead. This could even slow down the syntax in such scenarios.

   Maybe this can be avoided in future changes, by just not using `with_prototype` though.

The downside of this approach is, it wasn't possible to add further detailed scopes/highlighting to those `known_char_escape` tokens without bloating the capture groups.

Using branch points would be an elegant way, but requires context switches for each character or escape sequence. The prototype I've created performed 15% slower on those lines above.

keith-hall: After thinking about it some more, I realised that maybe I *do* have stronger opinions than I thought. I really like that you are asking these questions btw!

Optimising for speed makes perfect sense. Us maintainers don't look at the syntaxes *that* often, we can deal with the added complexity, and saving a few CPU cycles could cumulate across all the users of ST. But is it my preferred option? Honestly, no. One thing I really like is that each context has its own purpose, and should be re-used where possible. Something not highlighting correctly? Only one place to look. New version of the language to support? Easy to find where to make a change. Making a change in one place and having to (remember to) apply it in other places is not my idea of fun (whether it be a scope, a regex etc. I guess for me, its not about capture groups being harder to understand (PackageDev makes it easy when no variables are involved!) but the same construct being scoped in multiple different places) :D I even [had a go at removing the duplication in `group-start` and `group-start-extended`](https://github.com/forkeith/Packages/tree/regex_common_remove_duplicate_group-start-contexts) because I don't like it, and even though the result is a horrible mess of popping multiple contexts at once, I like that it somehow feels easier to understand. My personal opinion ofc! (alas, I didn't get it working for all cases yet, however.)

Nice to know about performance, I'd be tempted to stick to the current approach knowing those numbers, but you raise good points about `with_prototype` - maybe here is where a variable should be used for easy overriding?

Anyway, it feels like we are trying to get a perfect PR, when it may be easier for others to review smaller PRs ;)

deathaxe: I am fine with keeping the current approach. I can't really prefer the one or the other solution. So lets keep going with what we have.

Regular Expressions/RegExp (Basic).sublime-syntax
```diff
@@ -0,0 +1,280 @@
...
+    - include: character-classes
+    - include: escaped-chars
+
+  charset-ranges:
+    - match: (?=({{known_char_escape}}|{{char_escape}}|(?!\\-)[^\]])-({{known_char_escape}}|{{char_escape}}|[^\]]))
+      push: [charset-range-end, charset-range-begin]
+
+  charset-range-begin:
+    - match: '-'
+      scope: punctuation.separator.sequence.regexp-common
```
deathaxe, line 193: PD completed unwanted scope here

```suggestion
      scope: punctuation.separator.sequence.regexp
```

deathaxe (approved):

michaelblyons (approved):

PR #3041, opened 2021-09-12
[Markdown] Fix setext headings in symbol list

deathaxe:
This PR adds symbol transformations to...

1. fix underline markers of SETEXT headings from being displayed after the heading's text.

   Before this commit they looked like:

       heading======

   With this commit:

       heading

2. strip leading whitespace before starting any further transformation.  
   This is required because headings can be indented by up to 3 chars.  
   The last pattern of the old symbol transformation didn't match for those.

3. adds some patterns from MarkdownEditing, which convert links and references to normal text.

### DIFF
--- a/Markdown/Symbol List - Heading.tmPreferences
+++ b/Markdown/Symbol List - Heading.tmPreferences
@@ -9,9 +9,14 @@
 		<integer>1</integer>
 		<key>symbolTransformation</key>
 		<string><![CDATA[
-		s/\s*#*\s*\z//g;         # strip trailing space and #'s
-		s/(?<=#)#/ /g;           # change all but first # to m-space
-		s/^#( *)\s+(.*)/$1$2/;   # strip first # and space before title
+			s/\[([^]]+)\]\([^)]+\)/$1/g; 		# beautify links
+			s/\[([^]]+)\](?:\[[^]]+\])?/$1/g; 	# beautify references
+			s/^\s*//g;							# strip leading whitespace
+			s/\s*#+\s*\z//g;					# strip trailing hashes
+			s/(?<=#)#/ /g;						# change all but first # to m-space
+			s/^#( *)\s+(.*)/$1$2/;				# strip first # and space before title
+			s/^(.+?)=+\s*\z/$1/g;				# indent SETEXT heading 1
+			s/^(.+?)-+\s*\z/  $1/g;				# indent SETEXT heading 2
 		]]></string>
 	</dict>
 </dict>


keith-hall:
One problem with symbol transformations is there are no tests to show the behavior. I therefore tend to want to avoid them where possible by using/adding more specific scopes. But anyway, apart from the trailing `#`s etc which were already stripped in this step, there's probably not much else that can be done. So, any chance you could show some examples of the transformations done - the comments are nice but I don't entirely know what "beautify links" entails. It may not make sense to update the comments, just pasting here is fine to help with the review, so we can test it locally with various strings. (Do people really commonly put links in setext headings?)

deathaxe:
I agree in general and it was the first thing I've tried. 

Excludingheadings punctuations would drop the ability to indent headings by level due to missing `#` or `===` vs. `---` for transformation. Indentation is quite useful to distinguish the level of headings though. An alternative would to keep rendering the hash tags, which causes some noise, but still would fail for setext headings.


```markdown
# Heading 1

## Heading 2

 ## With [Link](https://heading.com) text

  ## With [Reference][ref] text
  ## With [Reference] text ##

### # ###


SeText Heading 1
================

SeText Heading 2
----------------
```

... shows up in Goto Symbols panel as ...

grafik

Please note: SeText headings not receiving kind information is caused by missing `entity.name.section` in the syntax.

jfcherng:
maybe you will be interested in https://gist.github.com/jfcherng/9673def3255f2a3257a2e2d59c40cdd9

Markdown/Symbol List - Heading.tmPreferences
```diff
@@ -9,9 +9,14 @@
...
 		<string><![CDATA[
-		s/\s*#*\s*\z//g;         # strip trailing space and #'s
-		s/(?<=#)#/ /g;           # change all but first # to m-space
-		s/^#( *)\s+(.*)/$1$2/;   # strip first # and space before title
+			s/\[([^]]+)\]\([^)]+\)/$1/g; 		# beautify links
+			s/\[([^]]+)\](?:\[[^]]+\])?/$1/g; 	# beautify references
+			s/^\s*//g;							# strip leading whitespace
+			s/\s*#+\s*\z//g;					# strip trailing hashes
+			s/(?<=#)#/ /g;						# change all but first # to m-space
+			s/^#( *)\s+(.*)/$1$2/;				# strip first # and space before title
```
keith-hall, line 17: can you explain this one a bit please? It looks like it could cause catastrophic backtracking because it matches space star followed by any whitespace plus followed by anything star...
It takes the text where all the leading hashes except the first one were turned into spaces, and then I'm a bit lost. The first capture group will take all but the last space before the title text starts, so if the heading was `## Hello`, then it will take the space which was converted from the second hash. And it is kept in the output, so is the comment correct? Or am I missing something?

deathaxe: Thes are the lines, which existed in this file before, so I kept them in place. If it is suspicious it was so before this PR, too. We may need to forward your question to the original author - or just replace it by the more explicit solution from  MarkdownEditing  3.0.0 which is basically the same as the one from @jfcherng's gist.

keith-hall (approved):

Markdown/Symbol List - Heading.tmPreferences
keith-hall: Apologies, I didn't notice it was there before. Thanks for the improvements!

deathaxe: Not at all. Existing before or not. If it's suspicious we need to at least think about it. We don't need to keep possible trouble makers.

keith-hall (approved):

PR #3045, opened 2021-09-16
[Markdown] Restrict image/link/reference highlighting

deathaxe:
See: https://github.com/SublimeText-Markdown/MarkdownEditing/issues/633

Space between description and metadata is not allowed in inline links, images or references. Markdown must therefore not claim it to be such an entity as it causes otherwise valid markdown files to get highlighted wrong.

Example:

    [description] (no more url)

This is no inline link according to CommonMark or GFM.


keith-hall (approved):

jrappen (approved):

jfcherng (approved):

jfcherng:
Do you want to create a PR for that? @ISSOtm

Issue #2899, opened 2021-07-14
[SQL] missing Postgres data types


basaran:
- Sublime Version: 3.2
- OS Version: Linux

```sql
CREATE TABLE IF NOT EXISTS public.dropzone_details
(
    id uuid NOT NULL DEFAULT uuid_generate_v4(),
    name text COLLATE pg_catalog."default",
    doc jsonb,
    poc json,
    size integer,
    CONSTRAINT id_name_uix UNIQUE (id, name)
)
```

Postgresql has some new data types introduced since version 9.5. It appears some of them are missing in `SQL.sublime-syntax`.

These data types should be added to `SQL.sublime-syntax`:

uuid
jsonb
json
xml
pg_snapshot
pg_lsn
txid_snapshot

Reference: https://www.postgresql.org/docs/13/datatype.html

michaelblyons:
Postgres isn't (yet) a part of #3046. Someone could add an extra extended file and start adding one, though. A handful of data types would be an easy place to start.

jrappen:
From my understanding @keith-hall is still working on it, so now would be a good time to add this to the TODOs there I guess.

jrappen:
Partly addressed by #3046 now.

deathaxe:
Note: Postgre is still incomplete right after merging PR #3046.

Issue #1764, opened 2018-11-09
[SQL] ANSI SQL


R0bk:
Back at the end of 2016 in #630 there was a brief discussion of different flavors of SQL in sublime. Currently writing code for non mySQL flavors can be quite painful e.g. temp tables (and following text) is commented out.

I was wondering if there ever was a decision surrounding adding an ANSI (or any others) compatible SQL syntax highlighter?

keith-hall:
I believe we are waiting for a future update to Sublime which will allow users to configure (on a global basis - not a project basis due to how the syntax graphs are stored) which concrete SQL syntax should be used, so that this repo can ship multiple SQL syntaxes for each dialect, then each user can choose which one should be used when a syntax references `scope:source.sql`.

jrappen:
@R0bk Maybe you could have a look at and comment on #3046 with some reference links and maybe links to tests for edge cases.

deathaxe:
Fixed by #3046

PR #3046, opened 2021-09-16
[SQL] use inheritance to support different dialects

### LINKED ISSUE #1742: [SQL] MySQL and Goto Symbol
Hello, I'm facing an issue:
image
As you can see, symbols index is wrong around `CREATE TABLE IF NOT EXISTS table_name` — ST thinks that table name is `IF`.
Also I've just found that `CREATE ALGORITHM=MERGE VIEW view_name` is out of index at all.
When I skip the `ALGORITHM=MERGE` declaration, it works fine.
Is seems, that > the SQL syntax that ships with Sublime is a bit old and perhaps doesn't cover the syntax for MySQL.
Then i've found the _SQL.sublime-syntax_ file and played with its copy. It's not very hard to fix my problem with Symbols, but generally there are lots of syntax issues with MySQL.

Proposal: create separate syntax-highlighting config containing actual syntax definitions (as of MySQL 8.0 or at least version 5.7).

keith-hall:
Initial work on supporting different SQL dialects using inheritance. Defaults to MySQL like before, but users can easily override the `SQL.sublime-syntax` to point to their preferred dialect. [Idea originally proposed by SublimeHQ on the forum](https://forum.sublimetext.com/t/inheritance-for-sql-syntax-variants/31659/4?u=kingkeith). 

Fixes #1742, #1764, #2899, #3072, #3072, closes #2631

***Note: breaking change:***
- the scope `keyword.other.DML` has been lowercased to `keyword.other.dml`, so color schemes may need a small change if they target this scope specifically


SQL/SQL (basic).sublime-syntax
```diff
@@ -0,0 +1,221 @@
...
+contexts:
+  prototype:
+    - include: comments
+
+  main:
+    - match: |-
+        (?xi)
+        \b(create(?:\s+or\s+replace)?)\s+
+        ({{dml_targets}})
+        \b\s*
```
keith-hall, line 21: Probably it makes sense to have a separate context for this instead of (or as well as) a variable, to improve the highlighting while typing experience.

jrappen:
Thanks for working on this. I noticed a few things:

- The indentation for `context: types: match: ...` is inconsistent with the rest of this PR.
- You use
  ```yaml
  captures:
    0: ...
    1: ...
  ```
  while some others use
  ```yaml
  scope: ...
  captures:
    1: ...
  ```
  which naturally both works but I thought you might want to use one over the other for readability. But you knew that.
- While requiring word borders via regex for correctly being able to scope words I believe the word scope should be only applied to the word without the borders.
- AFAIK `*.sublime-syntax` does not define a `comment` key, although `PackageDev` marks them as such. Maybe use standard YAML comments?

michaelblyons:
- RE `captures: 0:`: Someone (not me) is very determined to convert all of these to `scope:`.
- RE `comment:` mapping: I am pretty sure this is a product of the `<comment>` tag in tmLanguage that people used to use. To avoid dropping the comments, the sublime-syntax converter brought them over, and PD was kind enough to mark them as comments.

     I think we may as well convert to `#`-comments in the future as I mentioned in #2631, but the `comment:`-style ones were already there in the original.

> Thanks for working on this.

Big +1.

SQL/TSQL.sublime-syntax
```diff
@@ -0,0 +1,231 @@
...
+        - like-string-followed-by-escape-slash
+        - like-string-followed-by-escape-caret
+        - like-string-followed-by-unknown-escape
+
+  variables:
+    - match: (@)\w+
+      captures:
+        0: variable.other.readwrite.sql
+        1: punctuation.definition.variable.sql
+    - match: (@@)(?i:rowcount|trancount|error)
```
michaelblyons, line 43: Also `IDENTITY`

michaelblyons:
```sql
INSERT INTO my_table (foo, bar)
VALUES (1, 'two')
```

The table is scoped as a function.

SQL/TSQL.sublime-syntax
michaelblyons: https://github.com/michaelblyons/SublimeSyntax-Defaults/commit/1baf2de5b40893045301b6fadacfe97c9adc298c#diff-2a1a635547fcbf7145b49d2c363fd7a0c1fa6488732bf7617e868409ff622c2cR12-R13

keith-hall: Yep, thanks, lots of known variables and functions still to add :)

keith-hall:
> ```sql
> INSERT INTO my_table (foo, bar)
> VALUES (1, 'two')
> ```
> 
> The table is scoped as a function.
> 
> I'm finding that the biggest thing I miss from [my janky T-SQL definition](https://github.com/michaelblyons/SublimeSyntax-Defaults/commits/sql/select-db) is [the scoping for identifiers](https://github.com/michaelblyons/SublimeSyntax-Defaults/blob/sql/select-db/SQL/SQL%20Server.sublime-syntax#L190-L205) within (eg) `SELECT` statements. It's not at all perfect, and tailored to conventions I use (vs. the more-complete, but less specifically-scoped `identifier` context here). But it has been very nice to have table-name/alias marked a different color than column-name.

Indeed, I am aware of this (though thanks for pointing it out) and am trying to decide what approach is best to use - I'm thinking of specific, small contexts which will allow me to match column names where column names are expected, table names where table names are expected etc, while allowing fully/partially qualified names. I think the alternative is to just match anything left which is not a keyword as an identifier - it could still support qualified names, but without being context aware, it won't know if it is a table or column name when unqualified, which is why I prefer the former approach, despite it being more complicated and potentially brittle if I don't think of/cover all cases/syntax... So I had the idea to work on syntax test coverage first, to make that easier to implement in terms of knowing when it is correct.

keith-hall:
>     * While requiring word borders via regex for correctly being able to scope words I believe the word scope should be only applied to the word without the borders.

can you clarify what you mean here with an example please @jrappen ?

I believe I addressed your other comments :)

jrappen:
Will do when I'm back home.

keith-hall:
> Regression: Having a function in the `SELECT` breaks downstream highlighting.

Sorry about that, it seems every change I make breaks something :sweat_smile: but at least each time we get some more syntax test assertions to prevent it in future :D so thanks for the testing and reporting of issues! It is fixed now :)

jrappen:
I did some "comment"s research.

The built-in converter had originally brought over the comment tags from tmLanguage, when I did a PR to convert all files to sublime-syntax. I haven't tested it, but I assume the built-in converter works unchanged.

The docs do not specify a comment key, though. Whether by ommission or design, only Jon or one of the others could say.

---

With regards to word borders, I meant something like this random example from your code:

```yaml
  expressions:
    - meta_append: true
    - match: (?i)\b(CONCATENATE|CONVERT|LOWER|SUBSTRING|TRANSLATE|TRIM|UPPER)\b
      scope: support.function.string.sql
```

should've probably been this, yes? :

```yaml
  expressions:
    - meta_append: true
    - match: (?i)\b(CONCATENATE|CONVERT|LOWER|SUBSTRING|TRANSLATE|TRIM|UPPER)\b
      captures:
        1: support.function.string.sql
```

---

There still some `pop: true` left. These should be `pop: 1` when using v2 syntax definitions.

michaelblyons:
> ``` yaml
>  expressions:
>    - meta_append: true
>    - match: (?i)\b(CONCATENATE|CONVERT|LOWER|SUBSTRING|TRANSLATE|TRIM|UPPER)\b
>       captures:
>         1: support.function.string.sql
> ```


Haha, and I favor _this_ way, but 🤷 

``` yaml
  expressions:
    - meta_append: true
    - match: \b(?i:CONCATENATE|CONVERT|LOWER|SUBSTRING|TRANSLATE|TRIM|UPPER)\b
      scope: support.function.string.sql
```

deathaxe:
> RE `captures: 0:`: Someone (not me) is very determined to convert all of these to scope:.

Note: `captures: 0` may be somewhat faster than pairing captures with `scope`.

SQL/TSQL.sublime-syntax
```diff
@@ -275,23 +283,34 @@ contexts:
...
       scope: punctuation.section.group.begin.sql
       set: inside-with-group
     - match: \b(?i:nowait)\b
       scope: keyword.other.tsql
+
+  with-paren-or-pop:
+    - include: with-paren
     - match: (?=\S)
-      pop: true
+      set: table-alias
```
michaelblyons, line 313: I'm not sure exactly what's going on, but `WITH cte_table AS ( SELECT ... )` does weird stuff. It thinks the `(` is an identifier, not the opening of a sub-select. Also see `WITH cte_table (col_name) AS ( SELECT ... )`.

SQL/SQL (basic).sublime-syntax
```diff
@@ -313,21 +350,24 @@ contexts:
...
     - match: (?i)\b(asc|desc)\b
       scope: keyword.other.order.sql
 
+  distinct:
+    - match: \b(?i:distinct)\b
+      scope: keyword.other.DML.sql
+
   joins:
-    - match: (?i)\b(?:inner|(?:full\s+)?outer|cross|left|right)\s+join\b
+    - match: (?i)\b(?:inner|(?:full|left\s+)?outer|cross|left|right)\s+join\b
```
michaelblyons, line 368: T-SQL needs to add `APPLY` to the list of `JOIN`s.

michaelblyons: Also, bare `JOIN` is okay, right? It implies `INNER JOIN`?

michaelblyons (commented):
The scope-everything-as-a-constant for now is a little wild, but I can edit-color-scheme my way out of it.

Two notes on T-SQL types: 
- You're missing `text` (intentional?)
- You can wrap them in `[]` in some places, including the somewhat odd `[VARCHAR](1)` style.

SQL/SQL (basic).sublime-syntax
keith-hall: I believe I have fixed this, though I haven't added syntax tests for it yet

SQL/TSQL.sublime-syntax
keith-hall: I think I have fixed this now :)

SQL/SQL (basic).sublime-syntax
```diff
@@ -8,21 +8,19 @@ version: 2
 variables:
   string_escape: (?:\\.)
   simple_identifier: (?:\w+)
-  dml_targets: (?i:aggregate|conversion|database|domain|function|group|((?:fulltext|spatial|unique)\s+)?index|language|operator class|operator|procedure|rule|schema|sequence|table(?:space)?|trigger|type|user|view)
-  reserved: (?i:\b(?:from|order|group|select|where|inner|outer|left|right|join|on|set|;|with)\b)
+  reserved: (?i:\b(?:from|order|group|select|where|inner|outer|left|right|join|on|set|;|union|insert|delete|truncate|create|alter|drop)\b)
```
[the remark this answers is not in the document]
keith-hall: done, thanks for the suggestion

SQL/TSQL.sublime-syntax
```diff
@@ -337,11 +368,28 @@ contexts:
...
   collation:
     - match: \w+
       scope: support.constant.tsql
       pop: true
     - match: (?=\S)
       pop: true
+
+  joins:
+    - meta_append: true
+    - match: (?i)\b(?:cross\s+)?apply\b
```
michaelblyons, line 393: Or `OUTER`. Maybe others, but `CROSS` and `OUTER` are the only ones I've seen.

michaelblyons:
Looking very good. I will probably now run this at work, rather than my own def.

The current weird stuff:
- [ ] I can't seem to `UPDATE` after a CTE.
- [ ] `FOR XML RAW` 🙄 
- [ ] `UNPIVOT`
- [ ] `INSERT` rather than `INSERT INTO`: I'm not exactly sure what this does, but the highlighting doesn't recognize it.
- [ ] I'm not sure the intention of `constant.language.with`. The places I'm seeing it are not constants. They're naming columns in a CTE. Is that what you're trying to do with it?

keith-hall:
> * I'm not sure the intention of `constant.language.with`. The places I'm seeing it are not constants. They're naming columns in a CTE. Is that what you're trying to do with it?

Hmm, I thought I had covered named CTE columns with syntax tests and scoped them as `meta.column-name`. https://github.com/sublimehq/Packages/blob/be7d7fcdf5a4080bb3ffd525cdffecf0eadd5c2d/SQL/syntax_test_tsql.sql#L634

The `WITH` keyword is too overloaded in TSQL imo - we may need to get smarter about which scenarios we handle it in different ways. It can be used after `RAISERROR` "`WITH NOWAIT`", it can be used with parens after a `FROM tablename` or a `JOIN tablename` (or subquery?) (and before or after a table alias with an optional `AS`) to give table-level hints like `(NOLOCK)` - maybe I should take these known hints from your syntax def instead of scoping all words inside parens as `constant.language.with` (and perhaps changing the scope to mention "table hint") and finally at the beginning of a statement to declare a CTE. (there may be more cases I'm missing).

I am thinking that because semi-colons to end statements are mostly optional, it is going to be too brittle to try to accurately detect if it is the beginning of a (top level?) statement or not. But I may be able to improve it some more. Do you have any concrete examples you can share where it doesn't work please?

I will work on the other items, thanks for notifying about them. I may just add them into the bottom of the syntax test file without assertions to start with (like I've done for `merge`) to remind myself what still needs to be done.

keith-hall:
I found a case where a CTE wasn't scoped/identified correctly and fixed it. And I found another overload of `WITH` and added syntax tests for it. I believe I fixed `FOR XML RAW` and `INSERT` without `INTO`. `PIVOT` and `UNPIVOT` is still a WIP, ran out of time for now. I found a few other things which need fixing too, and added comments for them in the syntax tests file.

michaelblyons:
> Hmm, I thought I had covered named CTE columns with syntax tests and scoped them as `meta.column-name`. 

```sql
SELECT  *
FROM    table_name AS t1
        INNER JOIN  (SELECT foo FROM bar) AS t2(id)
--                                              ^^ constant.language.with
                    ON t2.ID = t1.ID
```
which I think is equivalent to 
```sql
SELECT  *
FROM    table_name AS t1
        INNER JOIN  (SELECT foo AS id FROM bar) AS t2
                    ON t2.ID = t1.ID
```

keith-hall:
> > Hmm, I thought I had covered named CTE columns with syntax tests and scoped them as `meta.column-name`.
> 
> ```sql
> SELECT  *
> FROM    table_name AS t1
>         INNER JOIN  (SELECT foo FROM bar) AS t2(id)
> --                                              ^^ constant.language.with
>                     ON t2.ID = t1.ID
> ```
> 
> which I think is equivalent to
> 
> ```sql
> SELECT  *
> FROM    table_name AS t1
>         INNER JOIN  (SELECT foo AS id FROM bar) AS t2
>                     ON t2.ID = t1.ID
> ```

Thanks, I hadn't heard of that before. I will try to implement everything from https://docs.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql?view=sql-server-ver15, which indeed includes this syntax.

michaelblyons:
> Thanks, I hadn't heard of that before.

I've never used it, either. 😅 I'm just clicking around in our SQL files looking for stuff that seems odd.

> I will try to implement everything from [T-SQL v15]

You're doing great. If you need to take a break, or if you want me to stop finding bugs just tell me to stop for a bit. 😁

keith-hall:
> You're doing great. If you need to take a break, or if you want me to stop finding bugs just tell me to stop for a bit. :grin:

Thanks! Pray, please continue - though I probably should also try to concentrate on other dialects (like MySql) to be sure the base syntax is as useful/correct as it should be, rather than just concentrating on T-SQL :sweat_smile:
but I find T-SQL most relatable and personally useful for me, and still lots to do/fix... so, if someone else wants to contribute a SQL dialect they are familiar with, it'd be most welcome.

basaran:
thank you for working on this, it appears newer postgresql data types haven't made it to the syntax file yet. https://github.com/sublimehq/Packages/issues/2899#issue-944713153.

michaelblyons:
T-SQL regressions:
```sql
CREATE UNIQUE NONCLUSTERED INDEX IX_some_index ON dbo.some_table(
    some_column ASC
)
```

```sql
SET ANSI_NULLS ON
```

SQL/MySQL.sublime-syntax
```diff
@@ -0,0 +1,173 @@
...
+    - meta_include_prototype: false
+    - match: '--(?=\s)'
+      scope: punctuation.definition.comment.sql
+      push: inside-double-dash-comment
+
+  strings:
+    - meta_append: true
+    - match: '"'
+      scope: punctuation.definition.string.begin.sql
+      push:
```
Ultra-Instinct-05, line 45: If you are going ahead with `push|set`ing into named contexts, perhaps it could be done here as well ?

SQL/MySQL.sublime-syntax
```diff
@@ -0,0 +1,173 @@
...
+        - match: '"'
+          scope: punctuation.definition.string.end.sql
+          pop: true
+        - include: string-interpolation
+    - include: begin-interpolation
+
+  begin-interpolation:
+    - match: '%\{'
+      scope: punctuation.definition.string.begin.sql
+      push:
```
Ultra-Instinct-05, line 59: Same as previous comment.

SQL/MySQL.sublime-syntax
```diff
@@ -0,0 +1,173 @@
...
+    - match: '(#\{)([^\}]*)(\})'
+      scope: string.interpolated.sql
+      captures:
+        1: punctuation.definition.string.begin.sql
+        3: punctuation.definition.string.end.sql
+
+  regexps:
+    - match: /(?=\S.*/)
+      scope: punctuation.definition.string.begin.sql
+      push:
```
Ultra-Instinct-05, line 78: Same as previous comment.

jrappen:
v2 syntaxes should use `pop: 1` instead of `pop: true`

SQL/MySQL.sublime-syntax
keith-hall: done

SQL/MySQL.sublime-syntax
keith-hall: done

SQL/MySQL.sublime-syntax
keith-hall: done

keith-hall:
> v2 syntaxes should use `pop: 1` instead of `pop: true`

indeed, I have changed this now, and logged an issue on PackageDev's repo for version-aware completions ;)

SQL/MySQL.sublime-syntax
```diff
@@ -0,0 +1,179 @@
...
+  strings:
+    - meta_append: true
+    - match: '"'
+      scope: punctuation.definition.string.begin.sql
+      push: inside-double-quoted-string
+    - include: begin-interpolation
+
+  inside-double-quoted-string:
+    - meta_include_prototype: false
+    - meta_scope: string.quoted.double.sql
```
Ultra-Instinct-05, line 50: Haven't done much SQL but I can see some string interpolation contexts. So should we get the `meta.string.sql` context in here to adhere with scope naming guidelines ?

SQL/MySQL.sublime-syntax
```diff
@@ -0,0 +1,179 @@
...
+    - match: '""'
+      scope: constant.character.escape.sql
+    - match: '"'
+      scope: punctuation.definition.string.end.sql
+      pop: 1
+    - include: string-interpolation
+
+  begin-interpolation:
+    - match: '%\{'
+      scope: punctuation.definition.string.begin.sql
```
Ultra-Instinct-05, line 60: Correct me if I am wrong but does string interpolation in SQL behave the same as normal string interpolations like Python, C# etc. ? If so, then shouldn't this be `punctuation.definition.interpolation.begin.sql` (If so perhaps have a separate scope for leading `%` as `punctuation.definition.interpolation.sql` ?)

SQL/MySQL.sublime-syntax
```diff
@@ -0,0 +1,179 @@
...
+  ddl-statements:
+    - meta_prepend: true
+    - match: \b(?i:create\s+or\s+replace)\b
+      scope: keyword.other.ddl.sql
+      push: [ddl-create-target, create-condition, ddl-target]
+
+  dml-statements:
+    - meta_append: true
+    - match: (?i:\binsert(\s+(?:ignore\s+)?into)?\b)
+      scope: keyword.other.DML.sql
```
Ultra-Instinct-05, line 131: Why is `DML` capitalized ? Relic of the old syntax ?

keith-hall: yup :)

jrappen:
* Could you check if this fixes #1742 and #1764 and add that to the opening comment (i.e. the commit message)?
* Were you going to add PostgreSQL as well and address #2899 while at it?

keith-hall:
>     * Could you check if this fixes [[SQL] MySQL and Goto Symbol #1742](https://github.com/sublimehq/Packages/issues/1742) and [[SQL] ANSI SQL #1764](https://github.com/sublimehq/Packages/issues/1764) and add that to the opening comment (i.e. the commit message)?
> 
>     * Were you going to add PostgreSQL as well and address [[SQL] missing Postgres data types #2899](https://github.com/sublimehq/Packages/issues/2899) while at it?
>       
>       * Docs at: https://www.postgresql.org/docs/

Yes it does fix those - done. I don't really have much interest in doing much for Postgres, I think it would be nice if we can merge with minimal support and someone can contribute further in separate PRs.

jrappen:
```text
.
└─ SQL
   ├─ tests
   │  ├─ indentation       # rules defined but not backed up by tests, yet
   │  ├─ symbols           # any plans? explained in sublimehq/sublime_text#3154 for ST4087+
   │  └─ syntax            # maybe you could move current tests here
   ├─ ...
   └─ ...
```

SQL/MySQL.sublime-syntax
keith-hall: good point, I think we should. Though, I'd personally rather leave this for a separate PR. This one is already getting too big - this context hasn't changed from the current one.

SQL/MySQL.sublime-syntax
keith-hall: I have no idea about MySQL interpolations, so I have left them as they are before this PR. I'm hoping someone with more knowledge can improve upon it after this PR is merged.

SQL/Indentation Rules - Comments.tmPreferences
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
```
jrappen, line 2: this line is redundant

keith-hall:
I know this PR isn't perfect (still lots of TODO comments etc.), but I think it proves the basics and (I hope others will agree!) is a nice base to work from for further improvements. The bigger it gets, the harder it will be to review - perfect is the enemy of good ;)

so maybe we can consider merging to master and continuing with any further improvements as separate PRs from there? I don't think there should be any regressions from previous behavior, based on the syntax tests, so we should already be in a much better state that it was in previously, even if these changes were to be included in a ST/SM release as is.

SQL/tests/indentation/syntax_test_indentation_tsql.sql
```diff
@@ -0,0 +1,65 @@
...
+c
+FROM foo
+
+SELECT a
+,b
+,c
+FROM foo
+
+CREATE TABLE hello (
+    --       @@@@@ definition
```
jrappen, line 36: Not sure why you mix symbol tests into a an indentation test file. Sure, anything is possible, but nobody will look here for those.

keith-hall: I guess I got overwhelmed with how many test files there are/can be, and how much they repeat the same code being tested. Personally I don't consider symbol tests super beneficial when there are no `tmPreferences` files changing the indexing rules in the package. The scope assertions should be enough...

SQL/MySQL.sublime-syntax
keith-hall: To be honest, I have no idea how interpolations work in MySql. My plan was just to leave it as it is (i.e. before this PR) in the hope that someone with more experience can improve upon it later.

SQL/tests/indentation/syntax_test_indentation_tsql.sql
Ultra-Instinct-05: Maybe symbol & indentation tests could be done at a later stage as well. The other packages are yet to have even indentation tests in them, save PHP & Java, I believe.

jrappen: Gotta start somewhere. Should I send you a PR to the PR?

keith-hall: Sure, please do :) thanks

SQL/Indentation Rules - Comments.tmPreferences
jrappen: fixed in forkeith/Packages#10

michaelblyons (approved):

jrappen (dismissed):

keith-hall:
> `TOP` freaks out if the `(` comes immediately after it: `SELECT TOP(1)`...

fixed now :)
image

SQL/SQL (basic).sublime-syntax
```diff
@@ -445,6 +450,11 @@ contexts:
       scope: storage.modifier.sql
     - include: expressions-or-column-name
 
+  use-db:
+    - match: \b(?i:use)\b
+      scope: keyword.context.tsql
+      push: [database-name, single-identifier-after-whitespace]
```
jrappen, line 456: Maybe rather?

```yaml
      push:
        - database-name
        - single-identifier-after-whitespace
```

It's easier to (1) see the depth and (2) read/review future diffs.

keith-hall: fixed, thanks for pointing it out. It is going to take me a while to get used to this style, maybe I'll write myself a linter ;)

jrappen:
@keith-hall could you fix the conflicts while reviews are pending anyway?

michaelblyons (approved):

deathaxe (approved):

mitranim:
Checking out the Postgres syntax and comparing it to `github.com/mitranim/sublime-sql`. Have no comments on other syntax variants since I only use PG. These comments might be somewhat misaddressed since I don't know how much of the syntax Keith is responsible for.

---

In `/* */` comments, indented leading `*` is `punctuation.definition.comment.sql`. That's technically incorrect and produces false positives when it's used as a regular bullet.

`create extension X` incorrectly gives `entity.name` to the keyword `extension`. Instead, `extension` should be a contextual keyword, while `X` does not need any special scope.

Identifiers in double quotes are incorrectly scoped as strings. In Postgres (and I believe in standard SQL as well) double quotes are used for identifiers. They're not strings and should not be scoped as such.

Single-quoted strings are not always scoped as strings. This shows up in check constraints, where I'm seeing `value ~ 'some_regex'` without string scoping (should actually be regex scoping in this case).

Lacks scoping of constraint names. In my practice, symbol search and goto reference is extremely useful for constraints. An identifier following the keyword `constraint` in table definitions and `create constraint` should be given a declaration scope and indexed.

Lacks scoping of type names following `create domain A as`.

Lacks scoping of certain reserved keywords. For any given database, some keywords are reserved and should be scoped as such in most contexts (but not when aliasing via `as`). This is the references I used for Postgres:

  * https://www.postgresql.org/docs/current/sql-keywords-appendix.html
  * https://github.com/postgres/postgres/blob/master/src/include/parser/kwlist.h

Lacks scoping of operators. Postgres supports an _open set_ of operators, allowing user-defined operators. So should the syntax. My syntax uses the following pattern, but I forgot to save the documentation reference: ``[\*\/\<\>\=\~\!\@\#\%\^\&\|\`\?\+\-]+``.

Gets confused by function definitions with `$$` quoted bodies. This breaks subsequent syntax highlighting.

Has non-contextual scoping of built-in types such as `text`. This should be _completely_ removed. Type names are _not_ reserved identifiers, and should be scoped only contextually.

Lacks support for `returns A` in function definitions.

Lacks support for various modifier keywords such as `parallel safe` in function definitions.

Lacks support for ordinal `$1` and named `:blah` parameters. Ordinal parameters are supported by functions and prepared statements. Named parameters are supposed by `psql` and commonly used in SQL query builders, such as `github.com/mitranim/sqlb`. They're extremely useful when embedding the SQL syntax in another, such as Go.

Lacks support for regular named parameters in functions.

Seems to lack support for floats with `e` exponents.

Column names lack a "declaration" scope. My syntax uses `variable.other.member.declaration`, which allows color schemes to automatically apply "member" styling. Tables are structs, and columns are struct fields, and should be scoped and styled as such.

Fails to scope column types as types. Has a whitelist of built-ins, but fails to scope user-defined types.

Uses bizarre "storage" scopes for various column keywords including `primary key`, `default`, and `check`. Fails to scope `unique` and other column-modifying keywords that occur after a column type.

Lacks support for contextual keywords in `create trigger`.

Incorrectly scopes `null` as constant rather than a keyword in compound keywords such as `not null`.

Lacks support for non-standard `::` cast operator, which is followed by a type.

Uses `entity.name.function` in places where a more specialized scope could be used, such as `entity.name.index`.

Has inconsistent scoping of the `select` keyword. Sometimes I see it highlighted, and sometimes not, in seemingly equivalent contexts.

deathaxe:
This feedback is very useful.

My understanding is this PR to intend to introduce the ability to support different aspects by introducing a `SQL (basic)` syntax.

The `T-SQL` dialect got most attention up to this poin.

Changes with regards to MySQL and PostgresSQL are rather rudimentary and target support for some special builtin functions only, which is reflected by the few numbers they have.


> In `/* */` comments, indented leading `*` is punctuation.definition.comment.sql. 

Might be correct, but scoping leading `*` is a general change which has been applied to all syntaxes a while ago as many comment generators such as DoxyGen, JsDoc, JavaDoc etc. use this style of starting each line with an asterisk.

> create extension X incorrectly gives entity.name to the keyword extension. Instead, extension should be a contextual keyword, while X does not need any special scope.

If `create extension` is the keyword, I'd say `X` is the `entity.name.extension`, no?

> Identifiers in double quotes are incorrectly scoped as strings. 

That's something which might probably be applied to `SQL (basic)`?

michaelblyons:
I found a MySQL bug with `TEMPORARY`.

```sql
CREATE TEMPORARY TABLE IF NOT EXISTS foo (
    bar NVARCHAR(400),
    baz INT
);
```

deathaxe:
Unil someone has gone though those dialects in detail using specs, there are dozens of dozens such bugs to expect. But those should have existed before, too.

The question is: Does anybody want to rewrite all dialects using this PR? Or is it a dedicated task?

michaelblyons:
I'm totally fine with merging as-is. I just want to mark down the deviations somewhere.

I do a lot of T-SQL and a little MySQL, so those are the only bits I'm competent to comment on.

If all of this is eventually rewritten in an even-better format later, I'm happy with that, too. But I think what Keith has done is good, and people using T-SQL/MySQL can benefit until such a rewrite is to occur.

jrappen (approved):

jrappen:
Let us know in case you want help  with PRs against your PR to do this now. Or just merge and do this in the next step.

keith-hall:
Well tbh I'm a bit unsure of what to do after seeing Benjamin's comment on discord... It would be nice to merge these changes somewhere - even if it's just into a feature branch on this repo for now if we can't merge to `master`, so we can make smaller, easier to review PRs to fix things.

jrappen:
That single line comment without any further explanation after you're done working on it for two months was a weird move for sure.

keith-hall:
I started doing some work to improve PostgreSQL based on suggestions here:
https://github.com/forkeith/Packages/pull/11

mitranim:
> If `create extension` is the keyword, I'd say `X` is the `entity.name.extension`, no?

To the best of my knowledge no, because it doesn't introduce `X` into the scope. Instead, various things _from_ `X` become available in the schema where the extension was "created".

keith-hall:
> Has non-contextual scoping of built-in types such as `text`. This should be _completely_ removed. Type names are _not_ reserved identifiers, and should be scoped only contextually.

done in the PR on my fork referenced above. Well, still need to fix a few Cassandra and TSQL syntax tests as a result.

jrappen:
@keith-hall Sorry, I lost track. What's the current status? Did you fix all of the points Nelo addressed? Is this ready for re-review?

keith-hall:
I believe I addressed most of them, but certainly not all. It'd still be nice to merge this and make iterative improvements as we go.

SQL/tests/syntax/syntax_test_postgres.sql
```diff
@@ -189,3 +196,99 @@ SELECT schedule[1:2][2] FROM sal_emp WHERE name = 'Bill';
...
+--         ^^^^^^^^^^ meta.column-name
+--                    ^^ keyword.operator.comparison
+--                       ^^^^^^^^^^ meta.string.regexp
+--                        ^^^^^^^^ source.regexp
+--                       ^ punctuation.definition.string.begin
+--                                ^ punctuation.definition.string.end
+--                                 ^ punctuation.section.group.end - meta.string
+);
+
+ALTER TABLE mytable ADD CONSTRAINT verif_code CHECK (code IN ('A', 'AG', 'AL', 'AS', 'B', 'C', 'D', 'DA')); -- TODO: scope CHECK correctly
```
jrappen, line 226: What about this?

SQL/tests/syntax/syntax_test_postgres.sql
```diff
@@ -189,3 +196,99 @@ SELECT schedule[1:2][2] FROM sal_emp WHERE name = 'Bill';
...
+--                                      ^ punctuation.section.brackets.end
+--                                       ^ punctuation.terminator.statement
+
+
+CREATE TRIGGER blah AFTER INSERT OR UPDATE
+-- ^^^ keyword.other.ddl
+--     ^^^^^^^ keyword.other
+--             ^^^^ meta.toc-list.full-identifier entity.name.function
+--                  ^^^^^^^^^^^^^^^^^^^^^^ keyword.other
+    ON some_table FOR EACH ROW EXECUTE PROCEDURE some_procedure(); -- TODO:
```
jrappen, line 273: What about this TODO?

jrappen:
Do you plan on postponing to fix the TODOs you wrote down in comments later, too?

keith-hall:
Good question - thanks for reminding me about those. I probably should fix those first - we don't really want to merge PRs with TODO comments 😅

mitranim:
> Might be correct, but scoping leading `*` is a general change which has been applied to all syntaxes a while ago as many comment generators such as DoxyGen, JsDoc, JavaDoc etc. use this style of starting each line with an asterisk.

It could and should be done only when the comment begins with `/**`, which indicates the beginning of a javadoc comment. Comments that begin with just `/*` shouldn't be treated as javadoc. Our JS syntax does it like this, as far as I can tell.

michaelblyons:
The conversation at #3168 got me thinking, especially after DeathAxe's mention of SQL (for Ruby).sublime-syntax which `extends` your SQL.sublime-syntax shim. Should the shim `extends` as well, rather than `include: scope:source.sql.xxx`? Then other child syntaxes for SQL embeds with interpolation get access to the named contexts rather than relying on `prototype` for everything.

If there are more arguments for/against it, I'd be interested. I haven't really thought through the ramifications.

keith-hall:
> > Might be correct, but scoping leading `*` is a general change which has been applied to all syntaxes a while ago as many comment generators such as DoxyGen, JsDoc, JavaDoc etc. use this style of starting each line with an asterisk.
> 
> It could and should be done only when the comment begins with `/**`, which indicates the beginning of a javadoc comment. Comments that begin with just `/*` shouldn't be treated as javadoc. Our JS syntax does it like this, as far as I can tell.

fixed :)


> Should the shim `extends` as well, rather than `include: scope:source.sql.xxx`? Then other child syntaxes for SQL embeds with interpolation get access to the named contexts rather than relying on `prototype` for everything.

When the ST Core feature is added to allow user-preferences to map a base scope to a specific syntax definition, and if the shim will be removed as unnecessary/unused\*, anything targeting `scope:source.sql` will get access to the named contexts. So using `embed` now could make sense for future consistency. 

I think the reason I hesitated to do this initially is because, having another syntax definition which is effectively an exact duplicate of another seems like it would be a waste of resources (I guess mainly RAM).

\* However, one tricky situation here is that `Packages/SQL/SQL.sublime-syntax` would no longer exist - anything referencing it by path would need to pick a specific dialect - as there is no support to do an embed by base scope. So I like your idea of a command to update the shim to point to a different dialect, it would sidestep this problem nicely.

deathaxe:
I am working on a more or less generic approach to display a quick panel, which lets user choose the desired dialect for a given "default" syntax.

It uses information from SQL.sublime-syntax to display a list of available dialects (sub-scoped syntaxes) and replaces the "extends: ..." directive accordingly.

deathaxe:
Here's the syntax chooser: https://github.com/deathaxe/DefaultSyntaxChooser

keith-hall:
> Here's the syntax chooser: https://github.com/deathaxe/DefaultSyntaxChooser

works beautifully, nice job! I guess we hope SublimeHQ would add it as a plugin that ships by default when this PR is merged, right?

deathaxe:
At least it can be an inspiration for a final solution. It could also be added to Package Control, for those who need it.

It would also be cool to have some mechanism to automatically find such generic syntaxes, so we wouldn't need to add an explicit entry in _Default.sublime-commands_ for each. Maybe a template file to look for or an YAML key such as `template: true` or `generic: true`. The latter one would need to parse each file though, which might be better done in core, like the `hidden` flag.

jrappen (dismissed):

jrappen:
@michaelblyons @deathaxe Maybe you guys could re-review? That would give us a chance to move forward with this.

The syntax choosing could be done in another PR.

michaelblyons (dismissed):

jrappen:
@anticultist As you have contributed to the SQL package in the past, do you have any feedback for Keith in this PR? Could you give it a test run?

anticultist:
First of all I want to say, that it is amazing to see such a progress on the SQL syntax! Great job.
I like the better and generalized approach to detect functions. (In the past I tried to go with the old approach and extend a predefined list of known functions. But at some point I realized that it is a dead end. So I stopped to improve my PR #2631... but I kept it for others as a starting point... maybe I should have pointed that out in that PR)

I believe it is a huge improvement on the SQL syntax - including the extended tests. Of cause you can find SQL code which is not good highlighted (especially procedure/function declarations). But this has nothing to do with this PR I believe. It is more of an issue on the current status of the SQL syntax.
Regarding to this PR I do not see any regressions at the moment. (I just tested the new highlighting but without the syntax definition itself...)

SQL/tests/syntax/syntax_test_mysql.sql
```diff
@@ -1,4 +1,4 @@
--- SYNTAX TEST "Packages/SQL/SQL.sublime-syntax"
+-- SYNTAX TEST "Packages/SQL/MySQL.sublime-syntax"
```
jrappen, line 1: nice

jrappen (approved):

michaelblyons (approved):

keith-hall:
> The following file on master may need an update even though not listed as conflict here:
> 
> https://github.com/sublimehq/Packages/blob/b2553fcfd99ff95ebfda975358c37ae6e43d6505/Rails/SQL%20(Rails).sublime-syntax#L1-L6

seems fine - merged `master` into the `sql` branch/this PR to be sure.

michaelblyons (approved):

jrappen (dismissed):

SQL/SQL (basic).sublime-syntax
```diff
@@ -0,0 +1,751 @@
...
+---
+name: SQL (Basic)
+scope: source.sql.basic
+hidden: true
+version: 2
+
+variables:
+  string_escape: (?:\\.)
+  simple_identifier: (?:\w+)
+  reserved: (?i:;|\b(?:from|order|group|select|where|inner|outer|left|right|join|on|set|union|insert|delete|update|truncate|create|alter|drop|return)\b)
```
deathaxe, line 11: No need to match semi-colon case insensitive.

```suggestion
  reserved: (?:;|\b(?i:from|order|group|select|where|inner|outer|left|right|join|on|set|union|insert|delete|update|truncate|create|alter|drop|return)\b)
```

SQL/SQL (basic).sublime-syntax
```diff
@@ -0,0 +1,751 @@
...
+  additional_reserved: (?!)
+  simple_types: (?i:\b(?:bit|bool|boolean|datetime|int)\b)
```
deathaxe, line 13: Are there reasons for this extra grouping here?

Would the following be enough?

```suggestion
  simple_types: \b(?i:bit|bool|boolean|datetime|int)\b
```

deathaxe (commented):
Again I can make some more or less useful formal suggestions.

SQL/SQL (basic).sublime-syntax
keith-hall: It would, but generally I thought we try to make a variable a group https://github.com/sublimehq/Packages/issues/757#issuecomment-544136278

deathaxe: Reasonable. 

I've never really paid attention to it, tbh. Only had a discussion about `(?x) ...` vs. `(?x: ... )` in mind. Other than that I've always have used a style suitable for applied use cases. The atomic style is more bullet prove though. Maybe worth keeping it this way especially with complexity of inheritance in mind.

keith-hall:
hmm, not sure about the punctuation `.parens.` -> `.group.` scoping change. Means can no longer color function call parens separately from other parens. Which to me, is quite useful in SQL syntax, as parens are used a lot :D The [scope naming guidelines](http://www.sublimetext.com/docs/scope_naming.html) are a little hazy on this, but it seems to suggest we can do it this way (search for `punctuation.section.parens.begin`).

(look at `newid()` here)
image

deathaxe:
I suggested `group` because

1. it appears most syntaxes use `punctuation.section.group` for function arguments,
2. `meta.group` is applied to those sections. Hence punctuation should follow.

Some syntaxes `meta.function-call.arguments punctuation.section.arguments`. I decided for `meta.function-call.arguments meta.group punctuation.section.group` in Java because of being uncertain.

Not sure which way we want to go with that to allow simple color schemes on the one hand and the opportunity for unique highlighting on the other.

SQL/Cassandra.sublime-syntax
```diff
@@ -0,0 +1,178 @@
...
+
+  joins:
+    - match: \bjoin\b
+
+  copy-from:
+    - match: \b(?i:from)\b
+      scope: keyword.other.cql
+      pop: 1
+    - match: ''
+      pop: 1
```
deathaxe, line 167: Due to the change of `begin-group-or-pop` the following change might be required.

```suggestion
    - match: (?=\S)
      pop: 1
```

keith-hall:
> Some syntaxes `meta.function-call.arguments punctuation.section.arguments`.

I saw that one of those syntaxes is Python, so I went with it here also :)

SQL/SQL (basic).sublime-syntax
```diff
@@ -0,0 +1,751 @@
...
+    - include: pop-on-top-level-reserved-word
+
+  ddl-target-on:
+    - match: \b(?i:on)\b
+      scope: keyword.other.sql
+      push:
+        - table-name
+        - single-identifier-after-whitespace
+    - match: (?=\S)
+      pop: 1
```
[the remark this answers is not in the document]
keith-hall: done

deathaxe (approved):

michaelblyons (dismissed):

jrappen (dismissed):

michaelblyons (dismissed):

jrappen:
@keith-hall I noticed a few things:

- I assume Michael's (@michaelblyons) TODO points from 2021-09-24 (in the discussions above) have been addressed?
    - `UNPIVOT`
    - `constant.language.with`
- I assume Ashwin's (@Ultra-Instinct-05) comments from 2021-10-03 (in the discussion above) with regards to string interpolation remain a TODO for later?
- I was looking at <https://github.com/sublimehq/Packages/pull/3046/files> and did a quick search for `keyword.other` in my browser while on that page and reviewing the points above. Is there no further distinction for so many parts of all those SQL flavors? I had 516 matches and something seemed off.
  
  image

keith-hall:
> * I assume Ashwin's (@Ultra-Instinct-05) comments from 2021-10-03 (in the discussion above) with regards to string interpolation remain a TODO for later?

correct

> * and did a quick search for `keyword.other` in my browser while on that page and reviewing the points above. Is there no further distinction for so many parts of all those SQL flavors?

most are `keyword.other.ddl` or `keyword.other.dml` etc., no? I'm not sure how to distinguish them even more tbh, but I'm open to suggestions even just for specific instances :)

jrappen:
Keith, maybe you could:

1. Change the remaining instances of:

   ```yaml
   - name: ...
     match: (?i)...
   ```

   to the more explicit:

   ```yaml
   - name: ...
     match: (?i:...)
   ```

   as well as `(?x)...` etc. etc.
2. Choose between `(?i:UPPERCASE)` and `(?i:lowercase)`. Currently it's inconsistently either one or the other.
3. Maybe you could support syntax based folding? It looks like that hasn't been added, yet.
4. Did you need anything from us apart from a review to get this moving again?

Ultra-Instinct-05:
> Maybe you could support syntax based folding? It looks like that hasn't been added, yet.

Perhaps that could be done later in a separate PR

deathaxe:
I don't see compelling reasons why we should enforce certain styles of `(?x)` vs. `(?x:...)` in normal patterns. The latter one is important in variables only, to avoid unexpected results when combining them.

jrappen:
@keith-hall I assume, clicking the button on <https://github.com/forkeith/Packages/compare/sql...sublimehq:master> and merging should fix this.

jrappen (approved):

jrappen:
@keith-hall Could you fix conflicts? Is this done for now?

michaelblyons (dismissed):

jrappen (approved):

deathaxe:
Do we really want to scope table name `entity.name.function`?

```SQL

CREATE TABLE version
--             ^ entity.name.function ???

```

keith-hall:
> Do we really want to scope table name `entity.name.function`?
> 
> ```sql
> CREATE TABLE version
> --             ^ entity.name.function ???
> ```

nope, not really :sweat_smile: fixed :)

buildersbrewery (approved):

buildersbrewery (changes requested):
.

deathaxe (commented):
After merging `master` into this branch, `WHERE` is not scoped `keyword` in the following line of PHP code

https://github.com/sublimehq/Packages/blob/11966dc3a07e0a7a7e5f9a01dbe9c3a8c5195072/PHP/tests/syntax_test_php.php#L4501

SQL/tests/syntax/syntax_test_postgres.psql
```diff
@@ -0,0 +1,311 @@
...
+-- ^^^^ keyword.declaration.variable
+    output varchar[];
+--  ^^^^^^ variable.other
+--         ^^^^^^^ storage.type
+--                ^ punctuation.section.brackets.begin
+--                 ^ punctuation.section.brackets.end
+--                  ^ punctuation.terminator.statement
+BEGIN
+-- ^^ keyword.other.LUW
+    FOR I IN array_lower(vals, 1)..array_upper(vals, 1) LOOP
```
deathaxe, line 297: Looks pretty broken.

SQL/tests/syntax/syntax_test_postgres.psql
```diff
@@ -0,0 +1,311 @@
...
+--  ^^ keyword.other
+--     ^^^^^^^^^^ meta.table-name
+--                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ keyword.other
+--                                               ^^^^^^^^^^^^^^ meta.function-call support.function
+
+
+CREATE FUNCTION highlight_result_array(vals varchar[], something tsquery, something_else boolean) RETURNS varchar[] AS $$
+DECLARE
+-- ^^^^ keyword.declaration.variable
+    output varchar[];
```
deathaxe, line 289: Is it really only one declaration, which is supported within `DECLARE`?

A second one will break highlighting.

SQL/TSQL.sublime-syntax
```diff
@@ -0,0 +1,841 @@
...
+    - match: ''
+      pop: 1
+
+  cursor-name:
+    - meta_content_scope: meta.cursor-name.sql
+    - match: ''
+      pop: 1
+
+  after-declare:
+    - match: (?=\w+\s+(?i:cursor)\b)
```
deathaxe, line 426: Are cursor names suppost to be simple unquoted identifiers only? `\w+` restricts it this way, while `single-identifer` indicates the opposite.

keith-hall: good question - I had only ever seen unquoted identifiers used, but it seems like any identifier is possible. I guess I'll have to fix this. From https://docs.microsoft.com/en-us/sql/t-sql/language-elements/declare-cursor-transact-sql?view=sql-server-ver16#arguments:

> cursor_name
Is the name of the Transact-SQL server cursor defined. cursor_name must conform to the rules for identifiers.

keith-hall: actually after testing on SQL Fiddle, it seems that it refers to unquoted identifiers only

deathaxe: So the whole pattern could be simplified to directly consume the identifer rather than pushing into simple-identifier followed by meta context.

jrappen:
are we back to reviewing?

deathaxe:
Maybe lower priority, but would we want to fill some gaps with regards to TSQL query hints? (see: https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver16)

deathaxe:
Would it make sense to finalize all scopes with `.sql` rather than `.mysql`, `.psql` or `.tsql` ?

jrappen:
Finalizing all scopes with `.sql` would imply it's SQL where it's not. Sure you have the base scope as a way to know where you are at, but you'd always have to double-check. I'd rather have a reference where parts of the code are taken from, though.

In other words I'd want to know that being at a certain point in my code, the scope tells me (as an example: ...) it's TSQL syntax within a TSQL file vs. SQL syntax within TSQL file.

deathaxe:
> ... would imply it's SQL where it's not.

???

The whole PR is about SQL, no? Just some different dialects with different feature sets, no? Everythhing is SQL here, no?

_My concern is just to end up in a wild and inconsistent mix of final scopes and unneccesary duplication of patterns which could otherwise be shared accross different syntaxes with little to no value compared to more code to maintain._

---

So if a signle dialect supports a distinct attribute in an otherwise very standard SQL statement, we want to scope that single keyword different?

SQL: 

```sql
CREATE INDEX ON table
--           ^^ keyword.other.sql
```

PostreSQL: 

```sql
CREATE INDEX ON ONLY table
--           ^^^^^^^ keyword.other.psql
```

---

How about expressions, which are shared across different but not all dialects? Do we want to define them explicitly in each dialect even though they could otherwise be located in _SQL (basic)_, just because we need to maintain dialect specific `mysql`, `psql`, ...? 

An example is `SCHEMA` keyword. Corresponding `expect-schema-name` context is located in _SQL (basic)_ with scopes terminated by `.sql`. Schema statements are supported by PostgreSQL and T-SQL (not yet implemented?), but not by MySQL or Cassandra.

So we double up all SCHEMA related context in Postgre and T-SQL instead of sharing them just because of the final scope?

There are probably more expressions like that to take into account.

jrappen:
I meant, I'd want a `tsql` suffix for those parts of the `tsql` code that are not basic `sql`.

To extend your example code, if there were meta scopes for the whole line, I guess they should both have the exact same scope with a `sql`-suffix. Those `keyword` scopes should differ though, as you said.

---

Using multiple inheritance would work for features only supported by some dialects?

jrappen:
```text
1         # base syntax
|\
| \
2  2      # sql , feature
| /
|/
3         # sql dialect with feature
```

jrappen:
I'm not sure it's worth an extra RFC issue, but I'll try to sum up the conflicts of interest here:

---

> #### 1

We can either have a base syntax which we can extend to the other dialects one by one.

**This gives us the ability to mark additional "features", so we know which part of (for example) a TSQL file comes from where (base, mysql, sql, tsql).**

---

> #### 2

We can put all features in the "feature base syntax", not use certain parts there but only in the extending "base syntax".

`feature base (features only) > base (with structure) > ...`

Which saves code duplicates, makes the "feature base syntax" unreadable and confusing, marks all scopes with `.sql` suffix.

Not ideal in my opinion.

---

> #### 3

We can have a base syntax. Then move features to "feature syntaxes" (that extend the base):

```yaml
extends:
  - SQL (basic).sublime-syntax
  - sql-feature-x.sublime-syntax
```

Which saves code duplication, adds a lot of feature syntaxes and might be confusing.

Problem here is how to scope certain features that (for example) PostgreSQL (`base > mysql > postgresql`) and TSQL (`base > tsql`) share? Back to scoping everything with sql suffix, I guess.

It's easier with features MySQL (`base > mysql`) and SQL (`base > mysql > sql`) share, because you'd automatically assume it's a MySQL feature that SQL inherited.

Not ideal in my opinion.

--

I guess it all comes down to the number of syntaxes.

Assuming there wont be any other SQL dialects added in the future, I'd go with number 1.

Assuming there will be a lot more added in the future, number 3 might make sense.

Assuming there will only be a few added in the future, it might be worth a good estimated guess, which way is better in the long run.

---

I hope this made sense.

deathaxe:
I am totally aware of how multi-inheritance can be used in general. It was one of the main reasons to ask for a common final scope `.sql` to use across all dialects. JS/JSX/TS/TSX scopes everything `.js` in all syntax dialects for instance.

But all of that is worth nothing if each dialect specific feature, which is not part of SQL-92 but shared accross n of m syntaxes must be scoped with dialect specific final scopes such as `.tsql` or `.psql`. Using dialect specific final scopes rules out shared features!

I guess, I even need to remove Postgre being inherited from MySQL to avoid `.mysql` scopes in Postgres sources.

How would you scope certain "non-standard" sql features shared across T-SQL and Postgre or Cassandra.

SQL/SQL (basic).sublime-syntax
```diff
@@ -0,0 +1,890 @@
...
+        - ddl-target-on
+        - single-identifier
+
+###[ DML STATEMENTS ]##########################################################
+
+  dml-statements:
+    - match: \b(?i:select)\b
+      scope: keyword.other.dml.sql
+    - match: \b(?i:union(?:\s+all)?)\b
+      scope: keyword.other.dml.sql
```
michaelblyons, line 223: `keyword.operator.set`? 😜

michaelblyons: My mistake. `keyword.operator.logical`

deathaxe: The whole scope naming of various statements/expressions is probably worth some refactoring. Historically many keywords are just scoped `keyword.other.dml` or `keyword.other.ddl` which seems rather over used.

It's sometimes not even very clear whether a keyword is a keyword or a constant or even some kind of function being called.

Finding good scope names may however require to finish implementation of main supported dialects first to get a better overview about what's needed. With each statement requireing its own syntax definition tree, this is a rather comprehensive task though.

michaelblyons:
Resolve conflicts? 😍

keith-hall:
> Resolve conflicts? heart_eyes

I had a quick go at it, but I don't have time to investigate what needs to change to fix all the embedded SQL inside interpolated PHP strings etc. right now.

deathaxe:
Seems some T-SQL changes conflict with https://github.com/deathaxe/sublime-packages/tree/wip/sql-dialects.

keith-hall:
> Seems some T-SQL changes conflict with https://github.com/deathaxe/sublime-packages/tree/wip/sql-dialects. SQL is soo massive.

Oh no :( I should have thought about that and perhaps merged your changes in first, sorry!

keith-hall:
@deathaxe probably it makes sense for me to undo my recent changes and re-apply/re-work them on top of yours, what do you think?

deathaxe:
My WIP branch contains various refinements with regards to context structure/architecture in order to be able to treat every statement individually in every dialect as it turned out there are many differences in detail. Shared contexts are implemented separately and included into the `alter-target`, `create-target`, etc. contexts.

MySQL already contains `grant` and `revoke` and some more privilege related statements. Some of those contexts would need to be moved to SQL (Basic) and layed out in a way the dialects can modify them to their needs. Originally I wanted to do that for all dialects in parallel statement by statement, but that's just too much. So started with MySQL with the idea of backporting it in a separate step.

This WIP branch is anything but complete or perfect, but probably pushes those syntaxes into the hopefully correct direction.

So yes, continue work from that branch would probably be the best strategy.

keith-hall:
Now includes @deathaxe 's MySQL work

michaelblyons (dismissed):

deathaxe:
I wonder how good the MySQL changes match PostgreSQLs syntax as it is based on MySQL.

michaelblyons:
The endless rebase continues... 😣

michaelblyons (dismissed):
Thanks. I didn't break up the TSQL `LIKE` placeholders when I did mine. Yours is more detailed.

SQL/PostgreSQL.sublime-syntax
```diff
@@ -0,0 +1,290 @@
+%YAML 1.2
+---
+name: PostgreSQL
```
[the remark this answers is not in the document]
deathaxe: Postgre is in early WIP state. Reading though its syntax specs and applying all the rules has not yet started.

michaelblyons:
I went to check on DeathAxe's PR in some of my user management files, and found this snippet breaks the rest of the file. (Though easily fixed with a `;` before `GO`)

```sql
DROP USER IF EXISTS [foo]
CREATE USER [foo] WITHOUT LOGIN WITH DEFAULT_SCHEMA=[bar]
GO
```

`WITHOUT` gets no scope, and the pop for `WITH`s might want `GO` and `END` as well as `;`.

SQL/Cassandra.sublime-syntax
```diff
@@ -0,0 +1,238 @@
+%YAML 1.2
+---
+# https://cassandra.apache.org/doc/latest/cassandra/cql
```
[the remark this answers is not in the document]
keith-hall: Fixed, thanks

SQL/Cassandra.sublime-syntax
```diff
@@ -0,0 +1,238 @@
...
+name: Cassandra Query Language
```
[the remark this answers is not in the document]
keith-hall: It would indeed make it easier to know what's changed and needs support added for but I haven't seen any easy way to correlate which version of Apache Cassandra ships with which version of Cassandra Query Language support. As the aim is always to support the latest version and we may not have covered all features yet, I vote to leave any comments about versioning out for now.

michaelblyons:
I use this all the time (though I actually further extend T-SQL for my main `source.sql` to add Hibernate style variables and some personal conventions).

I think the main obstacle to merge is disagreement in principle exhibited by the Discord comment from 2021:

> I don't think it makes sense to have specialised SQL syntaxes in the default packages. There's approximately too many SQL dialects and having to choose which one your particular files use every time you open them would be awful.

I'm curious whether it's an impasse or if there's an acceptable way forward.

keith-hall:
> I use this all the time (though I actually further extend T-SQL for my main `source.sql` to add Hibernate style variables and some personal conventions).
> 
> I think the main obstacle to merge is disagreement in principle exhibited by the Discord comment from 2021:
> 
> > I don't think it makes sense to have specialised SQL syntaxes in the default packages. There's approximately too many SQL dialects and having to choose which one your particular files use every time you open them would be awful.
> 
> I'm curious whether it's an impasse or if there's an acceptable way forward.

I guess the simple solution is to publish a package on package control called SQL, as we know that an installed package with the same name as a built-in one completely replaces the built in one. But then there'd be the useless built in one left in this repository which we wouldn't want to maintain...

I think the argument about there being too many dialects and having to choose one every time you open a file is moot, it would be configured per project at least if Sublime would support it. Otherwise the current solution of overriding the SQL.sublime-syntax file to extend whichever dialect the user is working with at the time seems fine, with the syntax chooser plugin.

If we do go with a separate package/repo, we could potentially incorporate Hibernate there.

If the main argument was about the number of dialects supported by default, we could theoretically cut down to MySQL (the currently only supported dialect shipped with ST) and TSQL and leave the rest for a third party package but to me it makes sense to keep them all together for easier maintenance .

deathaxe:
IMHO, the blockers are

1. to make sure all three dialects produce proper highlighting quality and completeness. According to specs, MySQL is at about 60-70% completeness. Can't say much about Postgre as I haven't studied its specs and therefore don't know how much it really shares with MySQL. Hence dropping it for the time being, is probably a good idea to reduce the distance to a release candidate.

2. scoping in general. Many keywords use scopes like `keyword.other.dml`, which leaves little opportunity to highlight them different. In some other cases (see MySQL) its unclear, whether some terms should be scoped keyword or constant. Scope naming is a question in general.

A 3rd-party SQL package could move forward in smaller steps as needed, but I'd be concerned about the way sql is embedded in all sorts of default packages. Some of them needed some adjustments to keep working with this PR. A 3rd-party package couldn't do that.

keith-hall:
> IMHO, the blockers are
> 
> 1. to make sure all three dialects produce proper highlighting quality and completeness. According to specs, MySQL is at about 60-70% completeness. Can't say much about Postgre as I haven't studied its specs and therefore don't know how much it really shares with MySQL. Hence dropping it for the time being, is probably a good idea to reduce the distance to a release candidate.
> 2. scoping in general. Many keywords use scopes like `keyword.other.dml`, which leaves little opportunity to highlight them different. In some other cases (see MySQL) its unclear, whether some terms should be scoped keyword or constant. Scope naming is a question in general.
> 
> A 3rd-party SQL package could move forward in smaller steps as needed, but I'd be concerned about the way sql is embedded in all sorts of default packages. Some of them needed some adjustments to keep working with this PR. A 3rd-party package couldn't do that.

Good point about third party package, that's why ideally it would be great to get this PR merged into this repo.

I'm not convinced that number 1 should be a blocker. As long as the highlighting is an improvement over what is currently in `master` branch, I don't see why it needs to be perfect. From the syntax test coverage alone we can be very confident that it is better for all dialects. Though it probably only matters for MySQL to satisfy user expectations as that is all that was previously supported.

In terms of number 2, it's a sad fact that almost everything in SQL is a keyword that doesn't fit into any other existing subscope. I tried to scope some things as constant in TSQL etc where possible, like for the permission type after `GRANT` but I feel like we can easily tweak this later based on feedback. In the end, comparing to other applications which highlight SQL, they also don't have many distinct colors and I believe it would look odd if Sublime did. We already ensure column names and aliases etc are correctly scoped and easily distinguished from "keywords".

michaelblyons:
> In terms of number 2, it's a sad fact that almost everything in SQL is a keyword that doesn't fit into any other existing subscope.

This may not be the place for it, but I think it's worth considering a new second-level scope on `keyword`, especially since languages (like C#) have SQL-like expressions that could use it. I don't know what the level should be called (`keyword.relational`? `keyword.query`?), but I think it'd be more useful than `keword.other.dml`.

Also, I've been using this branch for years. It'd be nice to have its improvements available for everyone without hunting for it.

deathaxe:
Besides finalizing support for various statements and dialects (e.g. Postgre) scope naming in general is still one of the major todos. I agree with `keyword.other.[dml|ddl|...]` not really fitting well into the general scope naming scheme.

deathaxe (approved):
SQL specs are too heavy to get everything done perfectly well. 

This branch/PR may not be perfect, but is still lightyears ahead of master.

I'd say: Just go with it and improve based on user feedback.

michaelblyons (approved):
Not my first time approving this. 😜

deathaxe:
### Summary

All supported SQL syntax dialects share a common _SQL (basic).sublime-syntax_ which provides common patterns/contexts and acts as a library. 

Supported dialects are: MySQL, PostgreSQL, T-SQL, Cassandra

Basically all syntaxes are completely rewritten, following SQL specs.

Primarily most important DML/DDL parts of each language are covered. As SQL requires dedicated syntax rules for each statement/feature, it is however hard to cover everything. So there are still unsupported statements in each dialect. 

All sorts of identifiers, such as tables, index, columns, users, constraints, ... are scoped to enable goto definition or goto reference as well as symbol lists. Not all of those have been added to the ladder though.

Note: Postgre is still in an early WIP state and has not seen much work. As the primarily supported syntax of the old SQL package however was MySQL, this PR doesn't step backwards.

jtojnar:
Just to make sure, is overriding supposed to be done by copying https://github.com/sublimehq/Packages/blob/master/SQL/SQL.sublime-syntax to `~/.config/sublime-text/Packages/SQL/SQL.sublime-syntax` and replacing the `extends` line?

michaelblyons:
Yes. Or install DefaultSyntaxChooser from [this DeathAxe comment](https://github.com/sublimehq/Packages/pull/3046#issuecomment-997235126).

PR #3068, opened 2021-10-01
[JavaScript] [TypeScript] Improve highlighting of declarations.

Thom1729:
Cleanup after fixes in #3061. Overall, simplify how declarations are handled, update scopes, and fix some corner cases.

- Add a new `declarations` context for class and function declarations in JavaScript, and all of TypeScript's new declarations. Use it in `statements` and in the export contexts instead of manually including the individual contexts.
- Use the proper `keyword.declaration` scopes for `type` and `enum`, and `storage.modifier` for `const` and `declare`.
- Use branching to correctly handle code like `declare()` and `abstract()`, and make sure that `declare const enum` works.

Un-special-casing `export default` means that e.g. `export default const foo = 42` is highlighted differently, but that's invalid code anyway so I don't think we care.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -337,15 +337,12 @@ contexts:
     - include: immediately-pop
 
   export-extended:
-    - include: variable-declaration
-    - include: class
-    - include: regular-function
+    - include: declaration
 
     - match: 'default{{identifier_break}}'
       scope: keyword.control.import-export.js
       set:
-        - include: class
-        - include: regular-function
+        - include: declaration
         - match: (?=\S)
           set: expression-statement
 
@@ -404,12 +401,12 @@ contexts:
       scope: punctuation.terminator.statement.empty.js
       pop: true
 
+    - include: declaration
     - include: import-statement-or-import-meta
     - include: export-statement
     - include: conditional
     - include: block
     - include: label
-    - include: variable-declaration
 
     - match: break{{identifier_break}}
       scope: keyword.control.flow.break.js
@@ -435,12 +432,14 @@ contexts:
       scope: keyword.control.flow.throw.js
       set: restricted-production
 
-    - include: class
-    - include: regular-function
-
     - include: decorator
 
     - include: expression-statement
+
+  declaration:
+    - include: variable-declaration
+    - include: class
+    - include: regular-function
 
   expect-semicolon:
     - match: \;
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -136,14 +136,6 @@ contexts:
         - ts-import-type
         - import-check-branch
 
-  export-extended:
-    - meta_prepend: true
-    - include: ts-type-declaration
-    - include: ts-interface-declaration
-    - include: ts-namespace-declaration
-    - include: ts-declare-enum
-    - include: ts-abstract-class
-
   property-access:
     - meta_prepend: true
     - match: \!\.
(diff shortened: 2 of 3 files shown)


deathaxe (approved):

rchl:
@Thom1729 This breaks some declaration files badly. Here is a full source of a file where I saw the problem:

```ts
/*@internal*/
namespace ts.server {
    declare const addEventListener: any;
    declare const removeEventListener: any;
    function findArgumentStringArray(argName: string): readonly string[] {
        const arg = findArgument(argName);
        if (arg === undefined) {
            return emptyArray;
        }
        return arg.split(",").filter(name => name !== "");
    }

    export function getLogLevel(level: string | undefined) {
        if (level) {
            const l = level.toLowerCase();
            for (const name in LogLevel) {
                if (isNaN(+name) && l === name.toLowerCase()) {
                    return LogLevel[name] as any as LogLevel;
                }
            }
        }
        return undefined;
    }

    export interface StartInput {
        args: readonly string[];
        logger: Logger;
        cancellationToken: ServerCancellationToken;
        serverMode: LanguageServiceMode | undefined;
        unknownServerMode?: string;
        startSession: (option: StartSessionOptions, logger: Logger, cancellationToken: ServerCancellationToken) => void;
    }
    function start({ args, logger, cancellationToken, serverMode, unknownServerMode, startSession: startServer }: StartInput, platform: string) {
        const syntaxOnly = hasArgument("--syntaxOnly");

        logger.info(`Starting TS Server`);
        logger.info(`Version: ${version}`);
        logger.info(`Arguments: ${args.join(" ")}`);
        logger.info(`Platform: ${platform} NodeVersion: ${getNodeMajorVersion()} CaseSensitive: ${sys.useCaseSensitiveFileNames}`);
        logger.info(`ServerMode: ${serverMode} syntaxOnly: ${syntaxOnly} hasUnknownServerMode: ${unknownServerMode}`);

        setStackTraceLimit();

        if (Debug.isDebugging) {
            Debug.enableDebugInfo();
        }

        if (sys.tryEnableSourceMapsForHost && /^development$/i.test(sys.getEnvironmentVariable("NODE_ENV"))) {
            sys.tryEnableSourceMapsForHost();
        }

        // Overwrites the current console messages to instead write to
        // the log. This is so that language service plugins which use
        // console.log don't break the message passing between tsserver
        // and the client
        console.log = (...args) => logger.msg(args.length === 1 ? args[0] : args.join(", "), Msg.Info);
        console.warn = (...args) => logger.msg(args.length === 1 ? args[0] : args.join(", "), Msg.Err);
        console.error = (...args) => logger.msg(args.length === 1 ? args[0] : args.join(", "), Msg.Err);

        startServer(
            {
                globalPlugins: findArgumentStringArray("--globalPlugins"),
                pluginProbeLocations: findArgumentStringArray("--pluginProbeLocations"),
                allowLocalPluginLoads: hasArgument("--allowLocalPluginLoads"),
                useSingleInferredProject: hasArgument("--useSingleInferredProject"),
                useInferredProjectPerProjectRoot: hasArgument("--useInferredProjectPerProjectRoot"),
                suppressDiagnosticEvents: hasArgument("--suppressDiagnosticEvents"),
                noGetErrOnBackgroundUpdate: hasArgument("--noGetErrOnBackgroundUpdate"),
                syntaxOnly,
                serverMode
            },
            logger,
            cancellationToken
        );
    }

    setStackTraceLimit();
    // Cannot check process var directory in webworker so has to be typeof check here
    if (typeof process !== "undefined") {
        start(initializeNodeSystem(), require("os").platform());
    }
    else {
        // Get args from first message
        const listener = (e: any) => {
            removeEventListener("message", listener);
            const args = e.data;
            start(initializeWebSystem(args), "web");
        };
        addEventListener("message", listener);
    }
}
```

Thom1729:
Whoops; I didn't account for `declare const` not being an enum in declaration files. Should be fixed in #3085.

PR #3078, opened 2021-10-06
[Regular Expressions] Add modifier/mode meta scopes

deathaxe:
This PR ...

1. scopes pattern sections `meta.mode.[basic|extended]` depending on the active mode (see: `(?x)` or `(?x: ...)`
2. fixes depending syntaxes and their tests (Clojure and Perl)
3. fixes a missing bailout in Perl's regexp patterns, which caused syntax to break in case of mode changes. _(thank's Keith!)_
4. adds variable interpolation support to Perl's regexp patterns.
5. optimizes some syntax patterns so performance impact of this change becomes negligible.

### DIFF
--- a/Clojure/tests/syntax_test_clojure_old.clj
+++ b/Clojure/tests/syntax_test_clojure_old.clj
@@ -96,9 +96,9 @@
   (def hello (fn [] #"(?=unclosed_paren"))
 ;                    ^^^^^^^^^^^^^^^^^^^ string.regexp.clojure
 ;                    ^ punctuation.definition.string.begin.clojure
-;                     ^ meta.group.regexp keyword.control.group.regexp
-;                      ^^ meta.group.regexp constant.other.assertion.regexp
-;                        ^^^^^^^^^^^^^^ meta.group.regexp
+;                     ^^^^^^^^^^^^^^^^^ source.regexp meta.group.regexp
+;                     ^ punctuation.section.group.begin.regexp
+;                      ^^ constant.other.assertion.regexp
 ;                                      ^ punctuation.definition.string.end.clojure
 ;                                       ^ - string.regexp.clojure
 ;                                       ^^ punctuation.section.parens.end.clojure


deathaxe:
Hmm. Applying `meta.mode` requires including syntaxes to take care off that new meta scope. I doupt that's a robust solution. Do we need that meta scope? Also those introduced lookaheads decrease performance by about 5%.

Regular Expressions/syntax_test_regexp.re
```diff
@@ -135,14 +135,14 @@ hello++
...
-#    ^ keyword.control.set
+#^ punctuation.definition.set
+#    ^ punctuation.definition.set
 # ^^^ constant.other.range
 #  ^ punctuation.separator.sequence
 
 [\x00-\x{A}]
-# <- keyword.control.set
-#          ^ keyword.control.set
+# <- punctuation.definition.set
```
keith-hall, line 144: I think it would be nice to have a `.begin` scope here

Regular Expressions/syntax_test_regexp.re
```diff
@@ -135,14 +135,14 @@ hello++
...
+#          ^ punctuation.definition.set
```
keith-hall, line 145: And a `.end` scope here :)

Regular Expressions/syntax_test_regexp.re
```diff
@@ -434,7 +439,8 @@ where escape characters are ignored.\).
 #                 ^^ variable.other.backref-and-recursion.regexp
 
 (?x)(?<named_group>test)(?&named_group)(?-x)
-#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.group.extended.regexp
+#   ^^^^^^^^^^^^^^^^^^^^ meta.group.regexp
+#                       ^^^^^^^^^^^^^^^ meta.modifier.regexp
```
keith-hall, line 443: Why do we need a meta modifier scope here?

Perl/RegExp (Perl angle brackets).sublime-syntax
```diff
@@ -0,0 +1,19 @@
+%YAML 1.2
```
michaelblyons, line 1: I had been confused about how to do interpolation without `push`/`with_prototype`, but I forgot that the `extends` pattern is available now. Exciting!

Regular Expressions/RegExp.sublime-syntax
```diff
@@ -67,6 +55,25 @@ contexts:
       scope: keyword.other.conditional.regexp
       pop: 1
 
+  modifiers:
+    - meta_prepend: true
+    - match: (\()(\?([+-]?\d+))(\))
```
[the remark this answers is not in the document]
keith-hall: regarding the `pop: 3`[...] logic, maybe it could help to just use a lookahead for the closing `)`, and let the contexts be popped off naturally?

PR #3081, opened 2021-10-09
[Python] Add more markers for SQL strings

michaelblyons:
I'm happy to add to `DECLARE` to make it `DECLARE\s+@` if that makes people happier.


keith-hall:
`DECLARE` works for me, as it is case-sensitive. Maybe you can also add some syntax tests for it? :)

PR #3085, opened 2021-10-10
[TypeScript] Fix namespace names and declare const.

Thom1729:
#3068 had a bug where `declare const` expected `enum` to follow and didn't handle variable declarations correctly. This PR fixes that.

It also fixes namespace names with dots in them not being handled correctly, which I thought was the problem originally.

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -181,7 +181,7 @@ contexts:
     - match: declare{{identifier_break}}
       scope: storage.modifier.js
       set:
-        - include: ts-const-enum
+        - include: ts-const
         - include: ts-enum
         - match: (?=\S)
           fail: ts-declare-enum
@@ -194,6 +194,14 @@ contexts:
         - match: (?=\S)
           fail: ts-abstract-class
 
+  ts-const:
+    - match: (?=const{{identifier_break}})
+      pop: true
+      branch:
+        - ts-const-declaration
+        - ts-const-enum
+      branch_point: ts-const-enum
+
   declaration:
     - meta_prepend: true
     - include: ts-enum
@@ -201,11 +209,7 @@ contexts:
     - include: ts-interface-declaration
     - include: ts-namespace-declaration
 
-    - match: (?=const{{identifier_break}})
-      branch:
-        - ts-const-declaration
-        - ts-const-enum
-      branch_point: ts-const-enum
+    - include: ts-const
 
     - match: (?=declare{{identifier_break}})
       branch:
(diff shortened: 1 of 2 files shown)


Thom1729:
Yeah, I saw that then forgot it. Thanks. Looks like `declare module` was just never properly implemented in the first place. Should be fixed in the latest commit.

I've also pushed a minor tweak to improve highlighting of the `declare` token while typing. This shouldn't affect any valid code such as `declare()`.

JavaScript/TypeScript.sublime-syntax
```diff
@@ -202,16 +232,17 @@ contexts:
...
         - ts-const-enum
       branch_point: ts-const-enum
 
     - match: (?=declare{{identifier_break}})
       branch:
-        - ts-declare-enum
+        - ts-declare
         - expression-statement
-      branch_point: ts-declare-enum
+      branch_point: ts-declare
```
deathaxe, line 245: Little nitpick, but I'd suggest to move `branch_point` before `branch` in all cases as this is the most common order of keys used so far.

JavaScript/tests/syntax_test_typescript.ts
```diff
@@ -265,12 +265,23 @@ import foo;
...
 //                     ^^^ entity.name.enum
 
+    declare const foo: any;
+//  ^^^^^^^ storage.modifier
+//          ^^^^^ keyword.declaration
+//                ^^^ meta.binding.name variable.other.readwrite
+//                   ^ punctuation.separator.type
+//                     ^^^ meta.type support.type.any
+//                        ^ punctuation.terminator.statement
+
```
deathaxe, line 275: Why is that not being added to the newly created `syntax_test_typescript_declarations.d.ts` where most other `declare` tests have been added?

JavaScript/TypeScript.sublime-syntax
```diff
@@ -202,16 +232,17 @@ contexts:
     - include: ts-namespace-declaration
 
     - match: (?=const{{identifier_break}})
+      pop: true
       branch:
         - ts-const-declaration
         - ts-const-enum
       branch_point: ts-const-enum
 
     - match: (?=declare{{identifier_break}})
```
deathaxe, line 241: Is there any reason for the absence of `pop: true` here, while it is part of the former one in line 235?

JavaScript/TypeScript.sublime-syntax
```diff
@@ -202,16 +232,17 @@ contexts:
...
       branch:
-        - ts-declare-enum
+        - ts-declare
         - expression-statement
-      branch_point: ts-declare-enum
+      branch_point: ts-declare
 
     - match: (?=abstract{{identifier_break}})
```
deathaxe, line 247: Same question with regards to `pop: true` here.

PR #3090, opened 2021-10-13
[meta] update github action workflow

jrappen:
to run parallel tests with latest stable & dev channel syntax_tests
binary

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
-name: CI
+name: ci
+
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -24,28 +25,28 @@ on:
...
-      - name: Move syntax_tests
-        run: mv st_syntax_tests/* ./
+      - name: Git checkout
+        uses: actions/checkout@master
 
-      - name: Cleanup syntax_tests archive
-        run: rm -R st_syntax_tests st_syntax_tests.tar.xz
+      - name: Get syntax_tests binary for Sublime Text Build ${{ matrix.latest-sublime-build }}
+        run: |
+          wget -O st_syntax_tests.tar.xz https://download.sublimetext.com/st_syntax_tests_build_${{ matrix.latest-sublime-build }}_x64.tar.xz
```
keith-hall, line 41: I seem to recall that there used to be an endpoint which always returned the latest version - I wonder if it is still around and if we want to use it. It would save us having to update it here after every release...

jrappen: I guess you meant these?

- <https://www.sublimetext.com/updates/4/stable_update_check>
- <https://www.sublimetext.com/updates/4/dev_update_check>

Ultra-Instinct-05: One one hand, we could have a bot (like dependabot ?) that opens a PR if the dev and stable version increments. But on the other hand, once the CI workflow is well defined, it's just the matter of updating version numbers. I wonder if automation is worth it for this case.

keith-hall (dismissed):

keith-hall (dismissed):

michaelblyons (dismissed):

jrappen:
Adds ability to run workflow manually from button GitHub. Will look (kinda sorta) like this:

image

jrappen:
I don't know. Will have to check.

jrappen:
Looking at <https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix> it seems possible to set variables with `matrix.include` for a list of dictionaries. I don't see this helping us here, though.

michaelblyons (approved):

BenjaminSchaaf (approved):

PR #3092, opened 2021-10-14
[C++] Add support for Qt-specific visibility modifiers Q_SIGNALS & Q_SLOTS

### LINKED ISSUE #3091: [C++] Treat Qt macros as built-in visibility modifiers
### What happened?

Feature request to support Qt-specific visibility modifiers macros. They act as hints/anchors for MOC (meta-object compiler), and resolve either to plain old C++ visibility modifiers, or to nothing at all:

```c++
# define Q_SLOTS
# define Q_SIGNALS public
```

There are also lowercase prefix-less `signals` and `slots` versions, but they are not advised to use except for the sole reason of looking cool.

It would be nice to have at least uppercase versions treated in the same way as `private|protected|public` triple, including auto-unindenting when a colon `:` is inserted after one of them. Note that Qt Creator IDE does unindent automatically.

To implement this, we'll have to expand this list of `visibility_modifiers` right here

https://github.com/sublimehq/Packages/blob/4bd58e57da149eb2fda72af932f8e911db6d9182/C%2B%2B/C%2B%2B.sublime-syntax#L63

and probably modify some scripts elsewhere.

ratijas:
In addition to widely used in Qt code `Q_SIGNALS:`, `Q_SLOTS:` and
`public Q_SLOTS:` visibility modifiers, this patch makes syntax
definition also accept potentially illegal code like

    public private:

or even

    Q_SIGNALS Q_SLOTS:

But that's a compromise I'm willing to take, in order to not
over-complicate things. At the very least, there's an "only one or two
items" restriction on quantifiers.

Closes #3091

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -67,7 +67,7 @@ variables:
   type_qualifier: 'const|mutable|typename|volatile'
   constant_expression_specifiers: 'consteval|constexpr|constinit'
   compiler_directive: 'inline|restrict|__restrict__|__restrict'
-  visibility_modifiers: 'private|protected|public'
+  visibility_modifiers: '(?:(?:public|private|protected|Q_SIGNALS|Q_SLOTS)\s*){1,2}'
   other_keywords: 'typedef|nullptr|{{visibility_modifiers}}|static_assert|sizeof|using|typeid|alignof|alignas|namespace|template'
   modifiers: '{{storage_classes}}|{{type_qualifier}}|{{compiler_directive}}|{{constant_expression_specifiers}}'
   non_angle_brackets: '(?=<<|<=)'
--- a/C++/Indentation Rules.tmPreferences
+++ b/C++/Indentation Rules.tmPreferences
@@ -17,7 +17,7 @@
 			# detent `default:`
 			| default\b
 			# class member access attributes
-			| ( public | private | protected ):
+			| (?:(?:public|private|protected|Q_SIGNALS|Q_SLOTS)\s*){1,2}:
 			# objc/objc++ attributes
 			| @( public | private | protected )\b
 			)
@@ -36,7 +36,7 @@
 			# indent after `default:`
 			| default\b
 			# class member access attributes
-			| ( public | private | protected ):
+			| (?:(?:public|private|protected|Q_SIGNALS|Q_SLOTS)\s*){1,2}:
 			# objc/objc++ attributes
 			| @( public | private | protected )\b
 			)


ratijas:
This PR is still relevant for me, as I'm working with Qt in Sublime Text on a daily basis.

> You should probably use inheritance to make a Qt specific dialect, compare recent PRs on this repo. This seems more like a dirty quick fix than a robust solution. 

IMHO Introducing a special "dialect" specifically for Qt and then associating all .cpp files with it (is that how you imagine it should work) would be highly overdoing it. Not to mention, C++ can be combined with obj-c and friends.

It is possible to create a "mix-in" plugin, such that, when installed, it loads along with base syntax and extends/overrides some rules?

michaelblyons:
> It is possible to create a "mix-in" plugin, such that, when installed, it loads along with base syntax and extends/overrides some rules?

This is what an extended syntax does. It just has a new name so that you can load the other one without extensions if you want to. DeathAxe is telling you the right thing. You're just concerned because it's not called the thing you want.

ratijas:
How about we reconsider inclusion of these fairly well-known modifiers in the package? Sublime Text and plugins have just upgraded, and now my local override is all broken, and I have to spend time again figuring out which rules/identifiers have changes. This wouldn't have happened if the modifications were upstreamed. I hate dealing with upstream changes, in QML plugin too -- which depends on JavaScript package and has been broken twice by unstable "API" already.

deathaxe:
The way forward is to bring C/C++ in shape for being inheritable, so such framework specific extensions can be added by inherited syntax definitions (e.g.: `C++ (Qt)`)

ratijas:
Maintaining the whole separate suntax plugin, enforcing API stability rules on the upstream -- all for the sake of changing one variable. What's even your point?

deathaxe:
Just recently someone was complaining about odd variable highlighting, which was caused by the tons of framework specific patterns in C++. So yes, this is the point.

deathaxe:
It is unnecessary to maintain a full copy of C++ syntax for what you want to achieve.

If adding just those variables, you can create your own local _C++ (Qt)_ package.

1. Add a _Packages/C++ (Qt)/C++ (Qt).sublime-syntax_

   ```yml
   %YAML 1.2
   ---
   name: C++ (Qt)
   scope: source.c++.qt
   
   extends: Packages/C++/C++.sublime-syntax
   
   file_extensions:
     - cpp
     - cc
     - cp
     - cxx
     - c++
     - C
     - h
     - hh
     - hpp
     - hxx
     - h++
     - inl
     - ipp
     - ixx
     - cppm
   
   variables:
     visibility_modifiers: '(?:(?:public|private|protected|Q_SIGNALS|Q_SLOTS)\s*){1,2}'
   ```

2. Copy indentation rules from normal C++ package, adjust `scope` to `source.c++.qt` 

Done.

PR #3096, opened 2021-10-19
[C#] allow member variables to have their assignment on a separate line

### LINKED ISSUE #3094: [C#] Unbalanced brackets wrongfully highlighted based on position of assignment operator
### What happened?

when using the following snippet:
```csharp
public class MyClass
{
    Dictionary<MyEnum, List<uint>> myDict
        = new Dictionary<MyEnum, List<uint>>
        {
            [One] = new List<uint>
            {
                1, 2, 3
            },
            [Two] = new List<uint>
            {
                4, 5, 6
            }
        };
}
```

both the `;` and the final closing bracket are highlighted in red by sublime. When moving the `=` operator in the top line it looks fine:

```csharp
public class MyClass
{
    Dictionary<MyEnum, List<uint>> myDict =
        new Dictionary<MyEnum, List<uint>>
        {
            [One] = new List<uint>
            {
                1, 2, 3
            },
            [Two] = new List<uint>
            {
                4, 5, 6
            }
        };
}
```

I would expect the same behaviour in both cases.

keith-hall:
Fixes #3094 by using branching

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -632,6 +632,8 @@ contexts:
     - match: ';'
       scope: punctuation.terminator.statement.cs
       pop: true
+    - match: (?=\})
+      pop: true
     - match: (?=\S)
       fail: method_or_member_variable_name
 
--- a/C#/tests/syntax_test_GeneralStructure.cs
+++ b/C#/tests/syntax_test_GeneralStructure.cs
@@ -1404,7 +1404,7 @@ public class MyClass
 
 public class MyClass
 {
-    bool var
+    bool var // missing semi-colon
 /// ^^^^ storage.type
 ///      ^^^ variable.other.member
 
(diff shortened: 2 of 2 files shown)


C#/C#.sublime-syntax
```diff
@@ -606,31 +606,41 @@ contexts:
...
+        2: keyword.declaration.function.accessor.get.cs
       set:
         - meta_scope: meta.property.cs
         - meta_content_scope: meta.method.cs
         - include: line_of_code_in
-    - match: '({{name}})\s*(?==[^>])'
-      captures:
-        1: variable.other.member.cs
+    - match: '{{name}}'
+      scope: variable.other.member.cs
```
deathaxe, line 627: Any chance not to break a lambda declaration if the previous attribute decl is not terminated?

```
public class MyClass
{
    bool var

    bool var => return 0;
}
```

grafik

deathaxe (approved):

C#/C#.sublime-syntax
deathaxe: May only one name token be valid in this position and branch?

Found another issue:

grafik

The closing brace is unscoped due to missing semi-colon.

Issue #2852, opened 2021-06-01
[RFC] Use punctuation.definition for brackets defining collection literals rather than punctuation.section


Thom1729:
Motivating example in JavaScript:

```js
// Block
    { }
//  ^ punctuation.section.block.begin
//    ^ punctuation.section.block.end

// Object literal
   +{ }
//  ^ punctuation.definition.mapping.begin
//    ^ punctuation.definition.mapping.end

```

# Overview

For brevity's sake, by “brackets” I mean curly braces, square brackets, parentheses, and other paired punctuation characters as appropriate.

Many languages use the same brackets to denote both sections of code (e.g. parenthesized expressions) and literal collections (e.g. mappings or sequences). Depending on the context and/or the text between the delimiters, the delimiters may represent very different syntactic constructs.

The current practice is to scope those brackets as `punctuation.section.*` in either case. This proposal suggests using `punctuation.definition` instead for collection literals, while keeping `punctuation.section` for all other purposes.

# Considerations

## Scope naming guidelines.

The [scope naming guidelines](https://www.sublimetext.com/docs/scope_naming.html#punctuation) are — somewhat surprisingly — silent on the issue.

They do say that “Sections of code delineated by” brackets should use certain meta scopes, and the brackets should use `punctuation.section.<section type>.begin|end`, where the section type might be either the bracket type (`braces`, `parens`, or `brackets`) or something semantic (`block` or `group`). They also specify `punctuation.section.interpolation.begin|end` where appropriate.

However, the guidelines do not mention collection literals such as mappings, lists, or tuples. They might be considered “Sections of code delineated by” brackets, but I think that what the guidelines had in mind were code blocks, parenthesized expressions, and the like. The given semantic scopes are `block` and `group`, not `mapping`, `sequence`, or `tuple`. Moreover, the scope naming guidelines have always been rather C-centric, and C has no true collection literals. C does have array initializers — most commonly strings.

The scope naming guidelines specify `punctuation.definition.string.begin|end` for string delimiters. This seems to be the only explicit guideline for punctuation defining literals. By analogy, brackets defining (e.g.) a mapping might be `punctuation.definition.mapping`.

I interpret the scope naming guidelines to be compatible with either `punctuation.section` or `punctuation.definition` for collection literals.

## Ergonomics

[...]

Issue #2711, opened 2021-02-06
[RFC] "Empty" values


FichteFoll:
Many languages define a language constant that is used to represent an "empty" value. For C-like languages that is `null`, while for JavaScript that may also be `undefined`. Python has `None` and functional languages like to use an empty pair of parentheses (`()`). Rust also uses `()` to represent "nothing", as opposed to an unset optional.

I want to propose scoping these constructs as `constant.language.empty`, where potential punctuation scopes follow that on the stack, since the constant is composed from both punctuation characters.

- All instances of the word constants that I can think of right now (`null` or `None`) should be using `constant.language` as their main scope already.
- `()` (and potentially similar constructs) seems to be a hot topic for people that have highlighting for punctuation scopes. In https://github.com/sublimehq/Packages/pull/2679#discussion_r566280566, we discussed the case for `()` in Haskell and https://github.com/sublimehq/Packages/pull/2387#issuecomment-657299850 is about `()` in Lisp (as well as https://github.com/sublimehq/Packages/pull/2312).  
  My proposal would circumvent that, as punctuation scopes would score higher and the scheme author needs to make a concious choice to prefer the constant highlighting over the punctuation.

Another concern presented is the change of highlighting while typing, because a plain `()`, the result of auto-pairing `(`, would exhibit a different color when highlighted as a constant compared to when you continue typing afterwards to fill the parentheses and use them as a group instead.

jwortmann:
Can we have a precise definition what exactly is meant by "empty" value?

I assume this should not apply to [`NULL`](https://en.wikipedia.org/wiki/Null_(SQL)) in SQL, which would probably be a `constant.language.boolean` (?) then. Or likewise, 
* in R, `NULL` would be `constant.language.empty`, while `NA` is `constant.language.boolean`
* in Julia, [`nothing`](https://docs.julialang.org/en/v1/base/constants/#Core.nothing) would be `constant.language.empty`, while [`missing`](https://docs.julialang.org/en/v1/manual/missing/) is `constant.language.boolean`

Correct?

[...]

PR #3097, opened 2021-10-21
[JSON] Rewrite syntax

jrappen:
- Using inheritance split up `JSON.sublime-syntax` into:
    - `JSON.sublime-syntax` with `scope:source.json`
    - `JSONC.sublime-syntax` with `scope:source.json.jsonc`
    - `JSON5.sublime-syntax` with `scope:source.json.json5`
    - `JSON_dotNET.sublime-syntax` with `scope:source.json.json-dotnet`
- Add many more file extensions for `JSON` & `JSONC`:
    - add doc links to extensions where applicable as a reference to be
      able to more quickly verify that they (still) use said syntax
      flavor
- JSON:
    - Make use of newer syntax features including those only available
      in `version: 2` syntaxes
    - Make use of `variables` (with optimizations provided by @deathaxe
      and regex patterns provided by @Thom1729)
    - Context names now more closely match the naming scheme of other
      (recently re-written) default syntaxes
    - (correctly formatted) JSON code can now be prettified or minified
      via the context menu or the command palette. JSON code can
      optionally be auto-prettified on pre save events.
    - highlight leading, trailing & multiple commas as invalid
    - only allow exactly one structure (object, array) or value
      (constant, number, string) at top level (thanks to @keith-hall)
    - links (`meta.link.inet`) and email addresses (`meta.link.email`) are
      scoped the same as in Markdown (thanks to @deathaxe)
- JSONC:
    - highlight some files by default as `JSONC` (as decided by
      @jskinner in sublimehq/Packages#285)
    - highlight leading & multiple commas as invalid, trailing as valid
    - scope empty block comments as such
    - support syntax based folding of ST4131+,
      compare sublimehq/Packages#3291
- JSON5:
    - explicitly pos numbers, hexadecimal ints, Infinity and NaN
    - single quoted strings
    - more escape chars for strings
    - ECMA identifierName as object keys (regexes thanks to @Thom1729)
        - scoped as plain unquoted strings (thanks to @Thom1729)
        - support string interpolation (thanks to @deathaxe)
    - line continuation in strings (with tests thanks to @keith-hall)
- JSON.NET:
    - support requested by @keith-hall,
      built with feedback from @michaelblyons
- Objects:
    - Highlighting speed improvements for empty objects (thanks to
      @FichteFoll)
    - Make `mapping.*` contexts more modular
- Arrays:
    - Highlighting speed improvements for empty arrays (thanks to
      @FichteFoll)
- Numbers:
    - Correctly scope number signs with `constant.numeric.sign` instead
      of `keyword.operator.arithmetic`
    - Significantly improve number highlighting (thanks to @deathaxe)
- Completions:
    - completions have been added for language constants, including kind info
      and details (with links to docs)
        - `null`, `false`, `true` for JSON
        - `Infinity` and `NaN` for JSON5
- Settings:
    - a `default_extension` is now set for all JSON flavors
- Symbol index:
    - with an object structure at the top-level, only top-level keys
      within now show up in the index (including tests for symbols and
      syntax)
- Tests:
    - test files now test the base scope
    - Significantly extend tests to cover more parts of the syntaxes
    - Split original test file into logical parts
    - Add indentation tests for:
        - `json`, `json5` & `jsonc`
        - `mapping` (objects), `sequence` (arrays)
    - Add symbols tests for:
        - top-level keys of object structures (thanks to deathaxe)
        - languages: `json`, `json5` & `jsonc`
    - Fix tests for `meta.mapping meta.mapping.*`
- Leave `JSON` headers in `Markdown` as `json` only, but split up fenced
  code blocks into `json`, `json5` & `jsonc` to behave similarly to
  `GitHub Flavored Markdown`

BREAKING CHANGES:

- JSON does not have values that can be set via an inline calculation
  with the help of operators, but only simple number values. Scopes for
  number signs have changed from being `keyword.operator.arithmetic` to
  `constant.numeric.sign`. Color scheme authors should add this, should
  it be missing.
- The `JSON.sublime-syntax` now marks comments as `invalid`, third party
  plugin authors should instead target `JSONC.sublime-syntax` to keep
  the user experience as-is.
- Indexed symbols (i.e. top-level keys in JSON object structures) are
  scoped as `source.json meta.mapping.key - (meta.mapping.value meta.mapping.key | meta.sequence.list meta.mapping.key)`.
  Color scheme authors should add special highlighting to differentiate
  them from other keys.

- fix sublimehq/Packages#285
- address sublimehq/Packages#421 (thanks to @FichteFoll)
- address sublimehq/Packages#481 to remove incompatible regex patterns according
  to @wbond
- address sublimehq/Packages#757 to fix line comments for `JSONC` (thanks to
  @keith-hall)
- address sublimehq/Packages#2430 using sort-order (as requested by @deathaxe)
- address sublimehq/Packages#2711 with regards to `constant.language.null`
  vs. `constant.language.empty` (thanks to @FichteFoll)
- address sublimehq/Packages#2852 to fix scopes of curly braces & square
  brackets in `JSON` (thanks to @Thom1729)
- address sublimehq/Packages#3228 to fix `punctuation.separator` scopes,
  compare sublimehq/Packages#3270
- address sublimehq/sublime_text#3154 and add symbol tests

Co-authored-by: Ashwin Shenoy <Ultra-Instinct-05@users.noreply.github.com>
Co-authored-by: Jack Cherng <jfcherng@users.noreply.github.com>
Co-authored-by: Janos Wortmann <jwortmann@users.noreply.github.com>
Co-authored-by: Jon Skinner <jps@sublimetext.com>
Co-authored-by: FichteFoll <FichteFoll@users.noreply.github.com>
Co-authored-by: Keith Hall <keith-hall@users.noreply.github.com>
Co-authored-by: Michael B. Lyons <michaelblyons@users.noreply.github.com>
Co-authored-by: Rafał Chłodnicki <rchl@users.noreply.github.com>
Co-authored-by: Thomas Smith <Thom1729@users.noreply.github.com>
Co-authored-by: Will Bond <wbond@users.noreply.github.com>
Co-authored-by: deathaxe <deathaxe@users.noreply.github.com>


jrappen:
⚠️  This is a first step to show my current work in progress.

jrappen:
As @jskinner requested in #285, this PR applies `JSONC` to `*.json` files by default to not confuse beginners too much.

jrappen:
There is still a lot of work to do. Please comment below with suggestions.

michaelblyons:
What do you think about using `source.json` and `source.json.jsonc`? I worry a little that moving the primary `*.json` association to `source.jsonc` will break a bunch of people and packages' configuration scopes.

jrappen:
Ok, follow up question, should I change stuff like:

```yaml
scope: comment.line.double-slash.jsonc
```

to

```yaml
scope: comment.line.double-slash.json.jsonc
```

for `JSONC.sublime-syntax`?

michaelblyons:
> Ok, follow up question, should I change stuff like:
> 
> ```yaml
> scope: comment.line.double-slash.jsonc
> ```
> 
> to
> 
> ```yaml
> scope: comment.line.double-slash.json.jsonc
> ```
> 
> for `JSONC.sublime-syntax`?

That's disputed. One camp says "yes." One camp says "no." One camp actually says "Change the last `.` to a `-`."

In this instance, I am in the "keep it as `.jsonc`" camp. Closest thing in default packages is the [GitHub markdown extensions](https://github.com/sublimehq/Packages/blob/master/Markdown/Markdown.sublime-syntax#L248).

JSON/JSON (Basic).sublime-syntax
```diff
@@ -0,0 +1,205 @@
...
+      captures:
+        1: constant.character.escape.double-quote.json
+        2: constant.character.escape.back-slash.json
+        3: constant.character.escape.forward-slash.json
+        4: constant.character.escape.backspace.json
+        5: constant.character.escape.form-feed.json
+        6: constant.character.escape.newline.json                               # linefeed
+        7: constant.character.escape.carriage-return.json
+        8: constant.character.escape.horizontal-tab.json
+        9: constant.character.escape.unicode-symbol.json
```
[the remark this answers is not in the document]
jrappen: Thanks, good to know

jrappen:
> If so, what?

In that case I'd assume nothing.

JSON/JSON (Basic).sublime-syntax
```diff
@@ -0,0 +1,205 @@
...
+####[ Variables ]#######################################################################################################
+
+
+variables:
+  boolean: |-
+    (?x:                                                                        # ignore whitespace
+      false|
+      true
+    )
+  'null': |-
```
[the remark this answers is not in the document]
jrappen: I'll probably get rid of both variables in the end.

michaelblyons:
> Could use this instead, though:
> 
> ```yaml
> - include: scope:Packages/JSON/JSON.sublime-syntax
> ```
> 
> unless I'm mistaken. I have a vague memory seeing this somewhere.

I know you can do this within a package. I believe Git Formats does so. You _may_ be able to do it between packages (e.g. Markdown front matter).

But switching away from `source.json` would still break other people's packages [all over the internet](https://github.com/TOKUJI/sublime3/blob/51b2647858d09d24d6c29ce0b79657fb2d9bfef8/Packages/zzz%20A%20File%20Icon%20zzz/aliases/JSON%20(Terraform).sublime-syntax#L9-L10).

jrappen:
Sometimes breaking stuff is the way forward. Not in this case, though.

JSON/JSONC.sublime-syntax
```diff
@@ -0,0 +1,109 @@
...
+
+####[ Extends ]#########################################################################################################
+
+  inside-double-quoted-string:
+    - meta_prepend: true
+    - meta_include_prototype: false
+
+  inside-double-quoted-string-inside-mapping:
+    - meta_prepend: true
+    - meta_include_prototype: false
```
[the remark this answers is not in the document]
jrappen: The JSON syntax doesn't have a prototype anyway, though.

1. This logically only partly makes sense to me. As we can mark these as "should not include when being extended". But on the other hand, there isn't anything when "not being extended, yet".
2. Would that break the JSON syntax?

jrappen:
⚠️ I have addressed and resolved all comments above, except for the one about `meta_include_prototype`.

---

Decimal floats are now forced to a more explicit format.

jrappen:
Thoughts on `embed: scope:source.json` in:

- LaTeX
- Markdown
- Perl
- PHP Source

anyone?

jrappen:
For Markdown I believe the frontmatter can be JSON and the fenced blocks can be split up into json & jsonc.

jrappen:
Should I be adding JSON5 (<https://spec.json5.org/>) in this step? Or later? Or does nobody care for that anyway?

deathaxe:
I was keen enough to push my number probosals directly. Hope it's ok.

jrappen:
Thanks, I'm working on this at <https://github.com/jrappen/sublime-json>. I'll give you access there, too.

jrappen:
I'm trying to fix trailing commas right now, without using branching as that's limited to 128 lines and that's not an option with JSON.

jrappen:
@deathaxe Thanks, I believe `invalid.illegal.operator.json` should be `invalid.illegal.number-sign.json`, though. As there are no operators in JSON, just values.

deathaxe:
We can of course rename it. I've choosen `operator` as signs are scoped `keyword.operator`.

deathaxe:
Btw. Why not just working in this branch? It's a WIP, it's squashed in the end, so kind and number of commits don't matter, IMHO.

jrappen:
I was trying not to spam this thread too much with my initial work.

jrappen:
Thanks to **deathaxe** for helping with numbers. I pushed our recent work. Still some TODOs left, here a list.

---

### TODOs

#### JSON

- [ ] main: highlight stuff other than array or object at top level as invalid
- [ ] array: trailing commas in array
- [ ] object: leading separators in object
- [ ] object: trailing commas in object

#### JSON5

- [ ] main: check whitespace valid chars
- [ ] strings: add multiline with new-line-escapes
- [ ] object: add ECMAScript identifier

keith-hall:
> ####
> * [ ]  main: highlight stuff other than array or object at top level as invalid

Respectfully, I disagree. A json file containing only a boolean or only a number or only a string is perfectly valid and shouldn't be highlighted as illegal.

jrappen:
ok

jrappen:
Reading the introduction on <https://www.json.org> seems to imply that it's either array or object on the top level, though.

JSON/JSON (Basic).sublime-syntax
```diff
@@ -0,0 +1,352 @@
...
+    - match: \]
+      scope: punctuation.definition.sequence.end.json
+      pop: 1
+    - match: ','
+      scope: invalid.illegal.expected-any-value.json
+    - match: (?=\S)
+      set: sequence-first-item
+
+  sequence-first-item:
+    - meta_scope: meta.sequence.item.first.json
```
deathaxe, line 228: Do we really need `meta.sequence.item` or even `...first`? As allowed data types are quite restrictive I don't see much value of another meta scope.

We should always keep in mind ST allows limited number of scopes being stacked only. So each additional scope reduces the level of nesting being supported.

deathaxe: We should try to keep the syntax as simple as possible.

jrappen: I was using `*.first`for tests with leading commas.

jrappen: In arrays and objects, the current situation is that it is:

- `meta.sequence.list` for the content scope of the array
- which is `set` replaced with `meta.sequence.item` for items in the array (or `*.first` for the first item).

So we don't stack further with arrays. Similar with objects where it's:

- `meta.mapping`, `meta.mapping.key` and `meta.mapping.value`
- not `meta.mapping meta.mapping.key` or `meta.mapping meta.mapping.value`.

deathaxe: > ... replaced with `meta.sequence.item`...

Haven't read carefully enough with regards to stacking.

Not sure if an item should replace `.list` sub scope though as `meta.sequence` is meant to be very general scope which can be sub typed if needed. IMHO, the `meta.sequence.list` should cover the whole array/list.

It's the same for mappings. The whole mapping is (to be) scoped `meta.mapping` while keys and values then use `meta.mapping.key` or `meta.mapping.value`. Those special sub scopes are just appended.

But with regards to sequences it would probably problematic to have
- `meta.sequence.list.item`
- `meta.sequence.array.item`

as any plugin/color scheme needs to explicitly address all kinds of sequences in case it just wants to match sequence items in general.

jrappen: It's `meta.sequence.item` replacing `meta.sequence.list` currently. Not `meta.sequence.list.item`. I guess I should fix that.

jrappen: Understood. Since you fixed this upstream, I assume we can close this sub-thread. Thanks.

jrappen:
Re-used some regexes by @Thom1729 from the JavaScript syntax to save me some time with JSON5 object keys with ECMA identifierName.

jrappen:
🙈  yikes

jrappen:
Open for suggestions on how to fix this.

Otherwise I'll take a 12 hour break before reviewing this.

jwortmann:
Some more questions and comments:

Out of interest, what is the effect of TM_LINE_TERMINATOR?

---

I wonder whether "JSONC (JSON with Comments)" is a good name for the default JSON syntax. I could imagine confused users in forum threads, asking why suddenly a dubious "JSONC (JSON with comments)" syntax is applied when they open a normal JSON file. Even if it has the same behavior (regarding comments & trailing commas) as before. I feel just "JSONC" would look nicer and be sufficient for the syntax name here. Or maybe even better to show "JSON" for the default JSONC syntax, and "JSON (strict)" for JSON without comments? This would allow the "Set Syntax: JSON" entry from the command palette to result in the same syntax as it was before.

Just for comparison, VSCode uses "JSON" as the name for its default JSON syntax (which does allow comment highlighting), and has an additional "JSON with comments" syntax, which I think only triggers a different mode with adjusted diagnostics for the language server.

---

It seems that the trailing dots in numbers are still marked as illegal, the scope just changed to `invalid.illegal.expected-mapping-separator.json` or `invalid.illegal.expected-separator.json`, depending on context. Although it was also marked as `invalid.illegal` even before (on current master), it could be improved by appending `\.?` to the regex match for integer numbers and leave this part unscoped. If you change it, you could add a syntax test to ensure that it won't be regressed.

---

For the (strict) JSON syntax, trailing commas are marked as illegal now. Despite this beeing technically correct for that current state of the buffer, the same arguments as for trailing dots after numbers would apply here:

json

I'd say this isn't a good user experience and should be left to linters.

---

The use of `punctuation.definition` for brackets/braces could be dabatable. I understand that this distinction mentioned in the linked issue might be useful for JavaScript, but in JSON it would be a quite significant highlighting change in Mariana (which would look worse in my opinion). And it would increase the inconsistency for brackets/braces scopes in different syntaxes, which makes it harder for color schemes. For example, I would guess that the white color in Mariana for `punctuation.section` was deliberately chosen to target all of the parens/brackets/braces. And therefore I think I would personally prefer to use standardized subscopes of `punctuation.section.*` for it instead.

In case there will be a positive decision for it in the RFC issue, then I would agree that it should be changed in JSON too, but for now I would leave it as `punctuation.section` until there is a consensus found.

michaelblyons:
- **"JSON" / "JSON (Strict)":** I like this, but I defer to consensus. One thing that might be weird: if the `name` is "JSON" for JSONC, but the name of the syntax whose filename is `JSON.sublime-syntax` is something else... people might get confused between filename and `name` key. Maybe not, though.
- **Unscoped trailing `.`:** I agree with jwortmann
- **Trailing `,`:** I also think this could just be unscoped and change to `punctuation` when the next (valid) sequence item starts to be typed. You could highlight it as invalid if _consecutive_ commas are encountered.
- **`punctuation.definition`/`punctuation.section`:** I lean slightly to jwortmann here, but I have custom overrides for JSON, so I will never see the default behavior.

jrappen:
Thanks for feedback, Janos and Michael.

---

* I assumed ST supports the `TM_LINE_TERMINATOR` key in metadata. Maybe not? Currently it's only in Python. I'll have to research. Thanks for heads up.
* `JSONC (JSON with Comments)` might've been too verbose, I can understand the preference for rather having `JSONC`. I wouldn't label `JSONC` as `JSON`, though. Nor label `JSON` as `JSON (strict)`.
* With regards to invalid floats, I agree that having the whole number marked as invalid is too flashy. In my opinion, we should keep marking the trailing period as invalid though as we cannot differentiate between views where somebody is currently typing vs views where somebody is currently not typing:
  * if someone types `0.0` it should only show up very briefly and only for the last character
  * if someone types `0.e-1` it stays and serves as a reminder to either convert to an integer or add a missing fraction
* With regards to trailing commas in `JSON`, I would like to remind you that Jon Skinner required `*.json` files to be opened with `JSONC` **by default**. So anyone using `JSON` would have to either (1) actively choose it over the default `JSONC`, (2) have an override or (3) edit a file with one of the `file_extensions` set in `JSON`. In all three situations, I would argue that invalid code should be highlighted as invalid.
* I can change the scopes for braces/brackets until the RFC issue is resolved. I strongly believe that color schemes should follow syntax definitions, though. Not the other way araound. In this case the braces/brackets are part of a definition of a JSON structure (a value).

jwortmann:
>     * `JSONC (JSON with Comments)` might've been too verbose, I can understand the preference for rather having `JSONC`. I wouldn't label `JSONC` as `JSON`, though. Nor label `JSON` as `JSON (strict)`.

My concern is also backwards compatibility. Wouldn't everything what a user added to *Settings - Syntax Specific* only be applied to the (strict) JSON syntax after this PR? If someone used *Syntax > Open all with current extension as... > JSON* for some filetype, then they will end up with the strict JSON syntax. [LSP-json](https://github.com/sublimelsp/LSP-json/blob/67596c5d191c604ad1e4504bc363aebe5eb12bcb/LSP-json.sublime-settings#L10) has a hardcoded link for `Packages/JSON/JSON.sublime-syntax` (ok, this example probably isn't a concern because it doesn't affect the ST4 version of LSP). But who knows how many other packages or user plugins are out there, which depend on the JSON syntax?

If confusion about the filenames is a concern, then they could be named according to the syntax label `JSON.sublime-syntax` & `JSON (Strict).sublime-syntax`. I haven't tested, but I guess the filename for the default JSON syntax better shouldn't change anyway, due to the settings backwards compatibility?

jrappen:
If the only concern is backwards compatibility, we might as well close this and not move forward at all. I'd rather focus on the future not the past.

Backwards compatibility is a valid concern. But how far does that go, where does it stop?

Whoever adds custom overrides with `Settings > Syntax Specific` should be able to update her/his config. Same goes for `Open all with current extension as ...`.

Jon said he was worried about the very beginners and to address their experience. I do believe it is not too much of an ask for everyone else to update their config once every x years.

I am not worried about `LSP-json`, many top contributors here are top contributors there. Someone will fix it within a very short period of time.

All (other) third party plugins should follow core. It's the same as above, for whoever is capable of writing a plugin it should not be too much of an ask to update the plugin once every x years.

I can send a few pull requests to the top JSON third party plugins on PackageControl, if that is the only blocker. I do believe it's not my job to take care of that for those authors on top of writing this PR, though. We could also mention them **after this is merged**.

jrappen:
Maybe we could move this PR into a direction where we have a solid foundation for the future that:

- Requires a change from third party plugin authors only once (after this is merged), but not in the (foreseeable) future.
- Further extend this PR with other JSON flavors to verify the base syntaxes being solid.
  - Currently looking into [JSON Lines](https://jsonlines.org)
  - I took a brief look at [HJSON](https://hjson.github.io), but that will be a lot of effort and [usage](https://hjson.github.io/download.html) seems to be very limited.

jrappen:
Linking some searches:

* GitHub
  * [`extends: Packages/JSON/JSON.sublime-syntax`](https://github.com/search?q=%22extends%3A+Packages%2FJSON%2FJSON.sublime-syntax%22&type=code)
    * currently only my test repo for this PR
  * [`include: Packages/JSON/JSON.sublime-syntax`](https://github.com/search?q=%22include%3A+Packages%2FJSON%2FJSON.sublime-syntax%22&type=code)
    * PackageDev
  * using [`Packages/JSON/JSON.sublime-syntax`](https://github.com/search?q=Packages%2FJSON%2FJSON.sublime-syntax&type=code):
    * LSP-json
    * Pretty JSON
    * some of Terence's stuff at [`STealthy-and-haSTy/SublimeScraps`](https://github.com/STealthy-and-haSTy/SublimeScraps)
* PackageControl
  * label: <https://packagecontrol.io/browse/labels/json>
  * name: <https://packagecontrol.io/search/JSON>

jwortmann:
> ... concern is backwards compatibility, we might as well close this ...

Why is that? All what I wanted to point out with my last comment, is that the file naming scheme proposed in this PR is not compatible with existing user settings for JSON, and potentially for third-party packages. This is to prevent unpleasant surprises after it gets shipped to users in a new release, so I think it should better be discussed beforehand. And currently it is just not clear to me, that the benefit (which I haven't really recognized yet) from the proposed filename change for `JSON.sublime-syntax` outweights the drawbacks of

1. breaking all user settings for JSON
2. introducing an ugly label in the status bar for the default syntax which will be used for JSON

I think the point made from sublimehq in the original issue about whether to introduce a JSONC syntax was, that there should be strong benefits for it if it comes with additional manual work for users (like what would be required due to the backwards incompatibility with settings).

And if the majority of users appreciate bright red flashes while typing, as demonstrated in the GIF I posted above, so that the syntax definition almost acts like a built-in compiler and can replace their linter, then go for it. If I understand it correctly, your argument was that basically nobody will use the strict JSON syntax anyway, because JSONC will be the default. But then why does the strict syntax exist in the first place? Just as an error checker to which a user can switch to, after he finished editing a file?

> ... I would argue that invalid code should be highlighted as invalid.

This is not the point with trailing dots and commas. The point is that it can become perfectly legal when continuing to type another character. The analogon of your agrument, lets say for Python, would be that if you have an empy file and just started to type some keyword like `def` or `if` in it, then should this be marked `invalid.illegal` until the full expression has completely been written, because at that current state of the file with only the keyword in it, it wouldn't run?

But if my comments are of no help, maybe I better should keep quite here now :shrug:

jrappen:
I personally believe that if someone chooses `JSON` as syntax, it should be highlighted as such. The current situation does not allow this using a default installation of Sublime Text. This was the starting point for me as well as FichteFoll in the issue linked in the opening post.

Deviating from the current situation and introducing other flavors of JSON will cause breaking changes. Either one way or another.

---

We can either (1) keep JSONC be a JSON syntax and introduce a JSON Basic (to make inheritance work) and JSON Strict (as a true JSON syntax). Any monkey patching around wrongly labeled syntaxes we avoid now in core and for any third parties will simply be defered to the future where third parties will have to address JSON Strict instead of JSON. Of course they will figure it out, but maybe not right away. Maybe this is less overall cost of time. Maybe not. I see this as offloading past problems to others in the future. And continuing to do so, as this has been an issue for a while and got us here in the first place. We can agree to disagree on this, I'm just stating my point.

Or we can (2) cause a bit of a stir now but be at peace for a long while afterwards. Hence why we can test this in the `dev` release channel for a while, before this hits `stable`. I'd assume package authors of frequently used packages are on the `dev` channel anyway.

We can also wait with this, until there is a longer cycle, if now is not a good time.

---

Not sure why you need to stress that the status bar label for the JSONC syntax is so ugly currently. I already agreed to change it to just JSONC. You can auto hide it while typing if desired. No big deal.

---

The JSON syntax exists for files with patterns listed in `file_extensions`, but you knew that. Why are you asking? If we remove the JSON syntax, what is the remaining point of this PR? Just adding JSON5?

---

If the bright red colors are such a bother, we have a few options. We can (1) either keep it and you add an override to Mariana to dim your reds, or (2) SublimeHQ will do it for you (which will never happen as we both know) or I (3) remove all illegal highlighting from JSON and add a linter that runs on the `json` module which ships with the Python included with Sublime and we have underlines or outlines - or whatever choice we agree upon - to highlight invalid regions (in this case). Or (4) I just remove all illegal highlighting, but then there really is no big difference between the highlighting of JSON vs JSONC.

Color schmes should follow syntaxes and (personally) I don't mind the red highlight of (temporarily or not but nonetheless) illegal code.

Because we are discussing:

- trailing periods within floats
- and trailing commas within structures (objects and arrays)

the user experience is flashing if we assume the user types fast. For you the flashing is more important (and disturbing) than is the highlighting of illegal code for me. We can again agree to disagree here, perfectly fine.

The problem for me as an author of the syntax file is though, that I cannot differentiate between "user typing" and "user done typing". Unless there is a way of "defering the highlight until after typing".

My suggestion is having a separate JSON syntax and writing it as it should be, not as the current user with old habits is used to. We can again agree to disagree.

If most users agree with youk then we should agree to close this PR as it will not benefit you or them. Hence my suggestion.

---

Which is fine with me, I can use this locally. I don't see this as having wasted my time. I have personal use for this.

keith-hall:
My 2 cents on this one issue about trailing commas being highlighted as invalid (while typing/not typing - we have no distinction as has been said already):

They are illegal in "strict JSON", and should be highlighted as such. Yes, it is possible that it could be annoying for users while typing. But:

- the benefits outweigh the drawbacks - it will likely save lots of trouble with people new to JSON who don't realize that trailing commas are illegal. "ST allows it for it's settings, but it's not working in my [language X] application. hmm... ah, I see that the syntax is different - JSONC vs JSON. Probably that's got something to do with it..."
- if we scope multiple consecutive commas distinctly from one trailing comma, color schemes/users can choose the behavior they want for trailing commas. ST isn't a "one size fits all" out of the box - it's designed to be customized.
- users can use snippets etc. to minimize the "flashes of red" by avoiding having trailing commas while typing at all
- how many people actually write JSON (and I mean "JSON (strict)" here I guess ;)) by hand these days?

jrappen:
I would agree with you, Janos, if this was a constant flicker like if you have to surfaces in the same plane in 3D and your maschine is constantly trying to render both. This is a single flash though of one single character which goes away after typing one (within the float) or two (within an array or object) additional characters.

For this reason I find this acceptable.

jrappen:
Hello guys, I didn't forget this. Simply didn't have the time to take care of this during the last couple days. I'll update this towards the weekend, as time permits.

jrappen:
⚠️ Status update

---

Some TODOs remaining before I'll open this for review:

- JSON object: leading separators
- JSON object: trailing commas
- JSON numbers: decimal floats with leading and trailing periods
- JSON5: valid whitespace

keith-hall:
I was going to ask whether you plan to work on scoping for dates, but then realized that what I've seen in the wild is not standard in any official JSON variant: https://davidsekar.com/javascript/converting-json-date-string-date-to-date-object

michaelblyons:
> I was going to ask whether you plan to work on scoping for dates, but then realized that what I've seen in the wild is not standard in any official JSON variant: https://davidsekar.com/javascript/converting-json-date-string-date-to-date-object

Haha. The old Microsoft JSON date format, leading to [one of my Stack Overflow bookmarks](https://stackoverflow.com/q/206384/241211).

jrappen:
> I was going to ask whether you plan to work on scoping for dates ...

I could still add them, though. I believe it adds value?

JSON/JSON (Basic).sublime-syntax
```diff
@@ -0,0 +1,388 @@
...
+
+  top-level-expect-number:
+    - include: numbers
+    - match: ''
+      push: top-level-expect-string
+
+  top-level-expect-string:
+    - include: strings
+    - match: ''
+      push: top-level-expect-invalid-remainder
```
deathaxe, line 59: This causes several contexts to be pushed on stack, once "string" is used as top-level item for instance.

grafik

I don't get the use of this construct.

jrappen: In JSON only one of these (constructs or values) can be at the top level:

- object
- array
- constant
- number
- string

Anything else following that is invalid. Having:

```json
"one"
"two"
```

I would mark `"one"` as valid JSON, but anything afterwards as `invalid`.

jrappen: So this was the approach of trying them one-by-one.

keith-hall: Maybe that context stack could be avoided if it would `set` instead of `push`? Or just include those contexts if they already `set` inside them?

deathaxe: How about `"string""string"`? Second one should be invalid shouldn't it?

jrappen: So do I `push` from `main` and then use `set` to go through the rest?

deathaxe: I am experimenting with creating popping variants of array/string/... to match only one at top-level. This strategy might be useful in values as well.

jrappen: Even `"one", "two"` at the top level would be invalid from the comma until the end.

jrappen: I'll give it a try.

jrappen: Thanks for the input.

jrappen: Before I try, I assume this will mark everything as `invalid.illegal.unexpected-code-after-first-structure-or-value.json` in any case? So maybe not.

deathaxe: Added my 5ct about it to https://github.com/jrappen/sublime-json/pull/8

jrappen: Thanks, I'll take a look at this tonight.

jrappen: Part of next push thanks to deathaxe.

jrappen:
@keith-hall Could you paste me a few links with references for your last request? I'd appreciate it.

I won't be adding your request to the JSON flavors I currently have (JSON, JSONC, JSON5), but rather add another flavor. I believe it's less confusing to add another flavor than add this kind of stuff to ones we already have and where it's not expected.

keith-hall:
These links may be useful:
https://www.newtonsoft.com/json/help/html/DatesInJSON.htm#DatesAndJsonNET
https://www.developer.com/microsoft/dealing-with-json-dates-in-asp-net-mvc/
https://github.com/featurist/dotnet-json-date

deathaxe:
The current implementation causes trouble with partial json in Markdown fenced code blocks.

grafik


Hence I am not convinced of adding too much of such over eager illegal syntax highlighting stuff.

jrappen:
> causes trouble with partial json in Markdown fenced code blocks

I've had this on my checklist for a while.

I initially thought it might make sense to have a (JSON) "snippet" syntax, since JSON is a comparatively simple syntax.

deathaxe:
JSON snippet syntax? There are already enought dialects. Adding more doesn't make sense. I'd rather relax existing JSON to not complain about incomplete globals.

jrappen:
Should I meta-scope links and email addresses in value strings as in the Markdown re-write?

Markdown/syntax_test_markdown.fenced_code_json.md
```diff
@@ -0,0 +1,30 @@
+//  SYNTAX TEST "Packages/Markdown/Markdown.sublime-syntax"
```
deathaxe, line 1: Perl and Markdown prefer to keep syntax tests in one file. This should be respected by a foreign syntax.

jrappen: ok, sure.

michaelblyons:
Just so you know, the @ mention in the commit description means we're getting an email for every force push.

TerminalFi:
Not a fan of built in minify and prettify.It is adding another entry point for UI sluggishness. I imagine this being shipped and then issues being filed for Sublime being slow when prettifying or minifying json. 

This is better suited for a plug-in.

Issue #692, opened 2016-11-16
[C#] Add scoping of Regular Expressions


keith-hall:
I think it would be great if ST could scope regular expressions (and replacements) inside C# code (much like it does for Python, JS etc.), for example:

```cs
var pattern = new Regex(@"^(?<s>-?[0-9]+)(?:,(?<s>-?[0-9]+))*$");
Regex.Match("2 Words", @"\d+ [A-Z][a-z]+").Dump();
return Regex.Replace(value, "(?!^)([A-Z])", " $1");
```

wbond:
This seems like it should be addressed once #93 is finally merged

michaelblyons:
Were you intending that we detect the `Regex` object name specifically and inject the syntax into the string(s) at the correct place in the method signature?

keith-hall:
That was my thinking, yes :) but it may be that we will only scope the `new Regex(@"`... constructor and not the static methods, I'd be fine with that

michaelblyons:
Lurkers may be interested in https://github.com/sublimehq/Packages/compare/master...forkeith:Packages:cs_regex

PR #3098, opened 2021-10-22
[Python] Extend Basic Regular Expression syntax for scope consistency

keith-hall:
Working towards #1942


michaelblyons:
Next up, #692? 😜

PR #3112, opened 2021-10-31
[Binary] Add Syntax

deathaxe:
This PR adds an empty Binary.sublime-syntax which is meant to be used for hexadecimal encoded binary files.

Its only purpose is to prevent confusion.

Current representation of hex groups is a clear indicator of binary content already, but the "Plain Text" syntax name in status bar may cause confusing in case file encoding is not displayed.

see: https://github.com/sublimehq/sublime_text/issues/4948

### DIFF
--- a/Binary/Binary.sublime-settings
+++ b/Binary/Binary.sublime-settings
@@ -1,0 +1,7 @@
+{
+	"auto_complete": false,
+	"auto_indent": false,
+	"spell_check": false,
+	"trim_trailing_white_space_on_save": "none",
+	"word_wrap": false,
+}
--- a/Binary/Binary.sublime-syntax
+++ b/Binary/Binary.sublime-syntax
@@ -1,0 +1,10 @@
+%YAML 1.2
+---
+name: Binary
+scope: binary.plain
+version: 2
+
+first_line_match: ^(?:\h{4} )+\h{4}$
+
+contexts:
+  main: []


jrappen (approved):

michaelblyons (approved):

deathaxe:
I think sublimehq should decide whether to add the syntax or not.

jrappen:
@deathaxe maybe also mark it significant?

deathaxe:
Not sure if its worth calling significant. It doesn't do anything but being an alias for Plain Text in case first line looks like hexadecimal representation of binary content.

The more interesting question might be whether ST want's to assign it via core functionality rather that `first_line_match` if it detects a binary file.

Ultra-Instinct-05:
I guess this can easily lead to false positives since it's based on first line match and just checks for 2 consecutive space separated 4 char hex characters. So one can easily have plain text files with the above content, which may not be binary files at all. I feel this is something better done in the core rather than a syntax.

deathaxe:
ST could probably assign binary syntax more reliably using existing core functions. _(Those which already detect binary files and assign Hexadecimal encoding for instance)_.

Nevertheless a Binary.sublime-syntax is required to be assigned as PlainText.tmLanguage is.

BenjaminSchaaf:
We'll be fixing that issue directly in core instead of using a separate syntax.

PR #3115, opened 2021-10-31
[Go] Fix symbol list kind symbols

### LINKED ISSUE #3055: [Go] Wrong symbol kinds
### Expected behavior

For Completion Metadata for Go's symbols:
- Types should be of kind `type`
- Methods should be of kind `function`

### Actual behavior

- Types are of kind `variable`
- Methods are of kind `variable`
- Constants are of kind `variable`
- Yet package-level variables themselves have no kind at all (icon is missing)

### Steps to reproduce

Use Goto Symbol in a file with Go source code and see the symbol names and their corresponding _kind_ icons.
Be sure the file has all possible top-level declarable things in Go - type, variable, constant, function, method.

deathaxe:
Fixes #3055

This PR ...

1. renames variable scopes in declaration statements to `variable.other.readwrite.declaration` to let ST choose correct `variable` kind info. `readwrite` is just added as counterpart of `variable.other.constant.declaration`.
2. merges `const` and `var` symbol list declaration files to a single _Symbol List - Variables.tmPreferences_
3. removes the custom function symbol list declaration file as it causes some fencier declarations to not have correct kind info. This way Go follows all other syntaxes by just displaying a function's name.

Notes: 

1. Namespace scope is not changed so packages will still have `namespace` kind info with the `n` kind symbol next to it. This part of #3055 is considered "wontfix".
2. We may move forward to something like `meta.declaration variable.other` in a future PR as alternative to the `variable.other.[constant|readwrite].declaration` scopes.


deathaxe:
Looks like symbol tests can't be added at this point.

michaelblyons (approved):

mitranim:
Thanks for looking into the issue. I have objections to removing the index files.

* Variables: in Go, goto for global variables is just as important as goto for functions or constants.
* Functions: Go syntax has the unique ability to scope type name and method name together, for each method, making it possible to search for a method by a combination of type name and method name.

I have custom symbol file overrides which prepend `func`, `type`, `var` and so on in the symbol list, so my screenshot is not most representative, but here it goes. I still think removing those prefixes was a stupid idea, because they allow filtering by _kind_ which ST currently doesn't support in other ways.

image
image

deathaxe:
Indexing global variables is unchanged. This PR  just merges _Symbol List - Const_ and _Symbol List - Var_ into one file.

Removing the custom _Symbol List - Functions_ is required for the correct kind info to be displayed for functions. They are still displayed in the symbol list but without any type prefixes. That's the way all syntaxes display functions.

That's the only solution to provide correct kind info as long as https://github.com/sublimehq/sublime_text/issues/3315 is not resolved.

Several overloaded methods in C++/Java would appear as "duplicate name" in the Symbol List as well.

The snippet provided by #3055 would look as follows with this commit:

grafik


> ... I still think removing those prefixes was a stupid idea, ...

I agree with you this to be a step backward with regards to the ability of filtering.

mitranim:
I see, thanks for explaining. About functions...

While having the right kind for methods would definitely be nice, the ability to easily filter methods by  type name is also very nice. Between kind and filtering, I feel that the latter is more important. I'm unaffected by the PR due to local overrides, so I'm just theorizing what would be preferred by a larger portion of users. But I'll defer to you on that.

Orthogonally to the above, after consulting the list of special cases ([discord link](https://discord.com/channels/280102180189634562/280157067396775936/736233105295278120)), I'm perplexed why it would assign kind `v`, where by all accounts it should use kind `f`.

keith-hall (approved):

PR #3120, opened 2021-11-03
[R] Add native pipe operator support

### LINKED ISSUE #3119: [R] Native Pipe |> Ligature
### What happened?

R 4.1 ships the native pipe (`|>`) which Fira Code renders as a triangle ligature. However, the R syntax of Sublime breaks this apart. It works fine in other syntax definitions or with plain text.

I asked [a question on SO](https://stackoverflow.com/questions/69810963/sublime-text-to-display-r-native-pipe-ligature-with-fira-code) and got the recommendation to open an issue here.

Thanks for your support!

deathaxe:
Fixes #3119

### DIFF
--- a/R/R.sublime-syntax
+++ b/R/R.sublime-syntax
@@ -156,6 +156,8 @@ contexts:
       scope: keyword.operator.arithmetic.r
     - match: (<-|->)
       scope: keyword.operator.assignment.r
+    - match: \|>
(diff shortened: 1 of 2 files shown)


christophsax:
That's fantastic, thanks!

michaelblyons (approved):

PR #3121, opened 2021-11-04
[C#] Add support for the #nullable compiler directive in C# 9.

rolfbjarne:
Ref: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/nullable-reference-types-specification#nullable-compiler-directives

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -174,15 +174,11 @@ contexts:
         - include: comments
         - match: $
           pop: true
-    - match: '\b(nullable)\s+(enable|disable|restore)\s+(annotations|warnings)\b'
+    - match: '\b(nullable)\s+(enable|disable|restore)(?:\s+(annotations|warnings))?\b'
       captures:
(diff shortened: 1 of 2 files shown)


C#/C#.sublime-syntax
```diff
@@ -174,6 +174,15 @@ contexts:
         - include: comments
         - match: $
           pop: true
+    - match: '\b(nullable)\s+(enable|disable|restore)\s+(annotations|warnings)\b'
```
keith-hall, line 177: maybe it makes sense to make the last capture group optional to avoid the need to duplicate the match pattern below without it?
```suggestion
    - match: '\b(nullable)\s+(enable|disable|restore)(?:\s+(annotations|warnings))?\b'
```

rolfbjarne: Thanks for the suggestion! It seems to work fine.

keith-hall (approved):

deathaxe (approved):

PR #3125, opened 2021-11-05
[Ruby] Fix HEREDOCs

### LINKED ISSUE #3124: [Ruby] SQL Heredoc not handled correctly
### What happened?

Having a simple SQL heredoc that is indented messes up syntax highlighting:

```ruby
def test
  sql = <<-SQL
    users.username LIKE :like OR
    users.email LIKE :like
  SQL
end
```
Screenshot from 2021-11-05 16-27-32

if the end tag is not indented, the highlighting is correct:

```ruby
def test
  sql = <<-SQL
    users.username LIKE :like OR
    users.email LIKE :like
SQL
end
```
Screenshot from 2021-11-05 16-31-36

deathaxe:
Fixes #3124

This commit...

1. embeds external syntaxes in HEREDOCs to reduce Ruby's size and to ensure terminating HEREDOC tags are matched under all circumstances.
2. improves ruby interpolation in embedded syntaxes using inheritance.

### DIFF
--- a/Ruby/syntax_test_ruby.rb
+++ b/Ruby/syntax_test_ruby.rb
@@ -133,24 +133,69 @@ puts <<-HTML; # comment
 #       ^^^^ entity.name.tag.ruby
 #           ^ punctuation.terminator.statement.ruby - meta.string
 #             ^ comment.line.number-sign.ruby punctuation.definition.comment.ruby - meta.string - string
-  <body>
-# ^^^^^^ meta.string.heredoc.ruby text.html.embedded.ruby meta.tag.structure
+  <script>
+    let me = #{@ruby_null};
+#  ^^^^^^^^^^^^^^^^^^^^^^^^^ meta.string.heredoc.ruby text.html.embedded.ruby source.js.embedded.html
+#            ^^^^^^^^^^^^^ meta.interpolation.ruby
+  </script>
+  <style>
+  .class[att=#{@ruby_sel}] {
+#            ^^^^^^^^^^^^ meta.string.heredoc.ruby text.html.embedded.ruby source.css.embedded.html meta.selector.css meta.interpolation.ruby
+    font-family: #{@ruby_font};
+#                ^^^^^^^^^^^^^ meta.string.heredoc.ruby text.html.embedded.ruby source.css.embedded.html meta.property-list.css meta.property-value.css meta.interpolation.ruby
+  }
+  </style>
+  <body class="#@var" style="color: #@color" onclick="run(#@what)">
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.string.heredoc.ruby text.html.embedded meta.tag.structure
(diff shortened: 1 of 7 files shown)


michaelblyons:
Whoa! I thought you were just going to

```yaml
  heredoc-css:
    - match: <<-CSS
      scope: ...
      embed: embedded-css
      escape: ^\s*CSS

  embedded-css:
    - include: scope:source.css
    - include: embedded-ruby
    - include: escaped-chars
```

Does that not work because new line is required?

deathaxe:
It does not work because normal ruby code may follow the HEREDOC tag up to the end of line.

Ruby/syntax_test_ruby.rb
```diff
@@ -133,25 +133,70 @@
...
+    function test() { var local = #{@ruby_init}; }
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.string.heredoc.ruby source.js.embedded
+#                                 ^^^^^^^^^^^^^ meta.interpolation.ruby
+  JS
+# ^^ meta.string.heredoc.ruby meta.tag.heredoc.ruby entity.name.tag.ruby
+end
+
 class_eval <<-RUBY, __FILE__, __LINE__ + 1
   def #{sym}(*args, &block)
 # ^^^^ meta.string.heredoc.ruby - meta.interpolation
```
deathaxe, line 202: Here's an example

michaelblyons (dismissed):

stoivo:
Pulled it locally and it works better for mee too. I will be using this going forward and report any issues I find.

themilkman:
Might be smth. with my ST but even using exactly this branch (not merged into my usual) the JS HEREDOC does not terminate. Not even in the Syntax Test file. All others seem to work fine.

deathaxe:
The `embed` is known to fail `escape`ing if the embedded syntax `set`s away from main branch immediately.

That's the reason why `push` is used to support shebangs in the default syntax.

https://github.com/sublimehq/Packages/blob/4adc549b9172c59fddff2cb67941fe9a6728accb/JavaScript/JavaScript.sublime-syntax#L118-L121

deathaxe:
To clarify the the last comment:

This PR embeds ....

1. `scope:source.js.embedded.ruby` in interpolation supporting HEREDOCs, which is defined in `Embeddings/JavaScript (for Ruby).sublime-syntax and is very likely to work correctly as it inherits from builtin JavaScript syntax.
2. `scope:source.js` in literal HEREDOCs, which don't support interpolation. If a 3rd-party JavaScript syntax is installed, which `set`s away from `main` forever, the above mensioned bug is triggered. The [ECMAscript Syntax](https://packagecontrol.io/packages/Ecmascript%20Syntax) is known for that due to following lines https://github.com/bathos/Ecmascript-Sublime/blob/1d62761878518a5877371c76c8910aa0765b9402/ecmascript.sublime-syntax#L184-L194

michaelblyons (approved):

keith-hall (approved):

PR #3127, opened 2021-11-06
[Git Formats] Fix issue references

### LINKED ISSUE #3123: [Git Formats] Git Commit highlights "C#10" as an issue reference when it isn't one
### What happened?

```text
[C#] Add support for C#10 features
```

image

deathaxe:
Fixes #3123

This commit enforces user name of at least 2 chars in auto-links of the form `user#issue` or `user@hash` to make sure not to highlight `C#7` or `F#7` etc.

### DIFF
--- a/Git Formats/Git Common.sublime-syntax
+++ b/Git Formats/Git Common.sublime-syntax
@@ -10,7 +10,11 @@ hidden: true
 variables:
   comment_char: '[#;]'
   hash: \b\h{7,}\b
-  repository_ref: '(?x: \b(?: [\w-]+ (/) )? [\w-]+)'  # user (optional) / repo
+  repository_ref: |-
+    (?x:
+      [\w-]+ (/) [\w-]+   # user / repo
+    | [\w-]{2,}           # user with a heuristic to exclude C#7, F#7, ...
+    )
   pretty_formats_builtins: oneline|short|medium|fuller|full|email|raw|reference
   pretty_formats_empty_value_modifiers: '[-+ ]?'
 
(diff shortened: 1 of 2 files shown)


jwortmann:
Wait, did you use that exact format which is intended to be removed here in this commit message, and this time GitHub adds the autolink for it :laughing:

So `user#234` seems to work, but not `repo#234`... that was unexpected.

`sublimehq#3123` -> sublimehq#3123
`Packages#3123` -> Packages#3123

michaelblyons (dismissed):
Someday we should look into what GitLab uses. I think in some cases it's `!` instead of `#`. Today does not have to be that day, though.

jrappen:
Michael, you are confusing GitHub with Gitlab. Gitlab differentiates between issues and merge requests in the compact formats used. GitHub uses the same for both.

jrappen (approved):

michaelblyons (approved):

keith-hall (approved):

jrappen:
`user#123` only works if it is able to (correctly?) reference a fork of said user.

I.e. `sublimehq#123` will reference issue or pull request with the number 123 for this repo. `otheruser#123` will reference an issue or pull request of `otheruser`'s fork of this repo.

michaelblyons:
> `user#123` only works if it is able to (correctly?) reference a fork of said user.
> 
> I.e. `sublimehq#123` will reference issue or pull request with the number 123 for this repo. `otheruser#123` will reference an issue or pull request of `otheruser`'s fork of this repo.

There's no possible way that sublime-syntax can tell the difference, no? It can't know whether some user has a fork of your repo.

jrappen:
I don't think so.

Unless you start dynamically monkey-patching via Python.

PR #3130, opened 2021-11-09
[C#] fix generic function calls

### LINKED ISSUE #3128: [C#] Lambda syntax not parsed correctly
### What happened?

Test code:

```csharp
using System;
class X {
	void M ()
	{
		Z <int>(P, x => { return 1; });
	}

	string P { get => null; }
	void Z<T> (string p1, Func<int, T> func)
	{
	}
}
```

is rendered as:

Screenshot 2021-11-09 at 11 51 29

Something is wrong at the end of line 5, although csc compiles the code successfully.

### LINKED ISSUE #3129: [C#] Space in generic type arguments causes parsing differences
### What happened?

Test code:

```csharp
using System;
class X {
	void M ()
	{
		Z <int>();
		Z<int>();
		X.Z <int>();
		X.Z<int>();
	}

	static void Z<T> ()
	{
	}
}
```

Rendered as:

Screenshot 2021-11-09 at 12 13 27

The call to Z on line 7 is coloured differently than the other ones.

keith-hall:
Fixes #3128, fixes #3129

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -35,7 +35,7 @@ variables:
   visibility: \b(?:public|private|protected|internal|protected\s+internal)\b
   base_type: (?:(?:bool|byte|sbyte|char|decimal|double|float|int|uint|long|ulong|short|ushort|object|string|void)\b)
   type_suffix: (?:\s*(?:\[,*\]|\*|\?)*)
-  generic_declaration: \s*(<[^(={};]*>)?\s*
+  generic_declaration: \s*(<[^={};]*>)?\s*
 
   brackets_capture: '((\[)(,*)(\]))'
   type_suffix_capture: '(\?)?{{brackets_capture}}?(?:\s*(\*))?'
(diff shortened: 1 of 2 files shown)


C#/C#.sublime-syntax
```diff
@@ -1487,6 +1449,14 @@ contexts:
       scope: punctuation.section.braces.begin.cs
       set: initializer_constructor
 
+  function_call_begin_paren:
+    - meta_content_scope: meta.function-call.cs meta.group.cs
```
deathaxe, line 1453: This implementation causes `meta.group` to be applied to whitespace between function name and parens

```csharp
        Z<int> (P, x => { return 1; });
///           ^ meta.function-call.cs meta.group.cs
```

keith-hall: well spotted good sir! fixed :)

deathaxe (approved):

jrappen (approved):

michaelblyons (approved):

Issue #3134, opened 2021-11-19
[PHP] `)` after a closure is not scoped


jfcherng:
### What happened?

```php
<?php

$array = array_reduce(
    $items,
    function ($array, $item) {        
        return $array;
    },
    $initial
);
// <- punctuation.section.group.end
```

Ref: https://discord.com/channels/280102180189634562/280157083356233728/910897785543462992

willrowe:
Thanks @jfcherng, I was just about to submit an issue!

PR #3135, opened 2021-11-19
[PHP] Fix `)` after closure is not scoped

jfcherng:
Fixes https://github.com/sublimehq/Packages/issues/3134

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -1261,10 +1261,14 @@ contexts:
         - match: \)
           scope: punctuation.section.group.end.php
           set: after-function-call
-        - match: ','
-          scope: punctuation.separator.php
-        - include: function-call-named-parameters
-        - include: nested-expressions
+        - match: (?=\S)
+          push:
+            - match: (?=\))
+              pop: 1
+            - match: ','
+              scope: punctuation.separator.php
+            - include: function-call-named-parameters
+            - include: nested-expressions
   function-call-named-parameters:
     # https://wiki.php.net/rfc/named_params
     - match: ({{identifier}})\s*(:)(?!:)
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -26,8 +26,7 @@ variables:
 
   keywords_declaration: |-
     (?x:
-      class | declare | enddeclare | function | global | include
-    | include_once | interface | namespace | require | use
+      class | declare | enddeclare | global | interface | namespace | use
     )\b
 
   keywords_control: |-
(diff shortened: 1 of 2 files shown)


PHP/PHP Source.sublime-syntax
```diff
@@ -1261,10 +1261,14 @@ contexts:
...
-        - include: function-call-named-parameters
-        - include: nested-expressions
+        - match: (?=\S)
+          push:
+            - match: (?=\))
+              pop: 1
+            - match: ','
+              scope: punctuation.separator.php
+            - include: function-call-named-parameters
+            - include: nested-expressions
```
deathaxe, line 1271: What if `?>` appears in front of closing `)`? By just pushing into another context you break `nested-expressions`'s bailout, which is implemented in `expression-end-pop` context. The new context is popped by it, if e.g. `;`, `?>` or a keyword is matched but it is immediately pushed on stack again.

I don't think that's a suitable solution.

Just replacing `nested-expressions` by `expressions` would have the same effect but without the pop-push-loop which is just prevented by ST's fail safe mechanisms.

Shouldn't closures correctly push `function-body` onto stack to prevent `expression-end-pop` from matching `;` or `}`?

deathaxe:
The bug is caused by `function` keyword in ...

https://github.com/sublimehq/Packages/blob/4b85d3f330cd94658e0aa35ed6cdc43f1a50ffe0/PHP/PHP%20Source.sublime-syntax#L29

It causes `expression-end-pop` to pop `function-call-parameters` context off stack if it sees the closure function.

The issue is fixed by removing `function` from line 29.

Maybe should check for some other keywords which may appear in expressions to avoid accidental pops. The `expressions` context matches several ones, but I wouldn't expect most/all of them valid within parentheses.

jfcherng:
> The bug is caused by `function` keyword in ...
> 
> https://github.com/sublimehq/Packages/blob/4b85d3f330cd94658e0aa35ed6cdc43f1a50ffe0/PHP/PHP%20Source.sublime-syntax#L29
> 
> It causes `expression-end-pop` to pop `function-call-parameters` context off stack if it sees the closure function.
> 
> The issue is fixed by removing `function` from line 29.
> 
> Maybe should check for some other keywords which may appear in expressions to avoid accidental pops. The `expressions` context matches several ones, but I wouldn't expect most/all of them valid within parentheses.

I guess we need a better naming then. `function` is a keyword is perfect correct I think.

PHP/PHP Source.sublime-syntax
jfcherng: I think we still go the `nested-expressions` but something has to be change in `keywords` variable. Do you think we need to rename it or just let it be.

jfcherng:
> The bug is caused by `function` keyword in ...
> 
> https://github.com/sublimehq/Packages/blob/4b85d3f330cd94658e0aa35ed6cdc43f1a50ffe0/PHP/PHP%20Source.sublime-syntax#L29
> 
> It causes `expression-end-pop` to pop `function-call-parameters` context off stack if it sees the closure function.
> 
> The issue is fixed by removing `function` from line 29.
> 
> Maybe should check for some other keywords which may appear in expressions to avoid accidental pops. The `expressions` context matches several ones, but I wouldn't expect most/all of them valid within parentheses.

Technically, `require` can work "like" a function which returns what is returned in the required file. So I think `require` has to be removed too.

deathaxe (approved):

PHP/PHP Source.sublime-syntax
deathaxe: The variable is only used to pop nested expressions at the moment. Not even sure whether all those different keyword types need to be placed in dedicated variables. I'd change it once those are to be used in other situations as well.

keith-hall (approved):

PR #3138, opened 2021-11-23
[Java] Improve JSTL interpolation in Java Server Pages

deathaxe:
This commit...

1. highlights ${expr} variable expansions everywhere in a document, not just in tag attribute values.
2. replaces the embedded Java syntax by a dedicated and more simple "EL" syntax, which is more or less a subset of Java. 

   EL means "expression language"

   Syntax is described in 

     Book:     "Core JSTL: Mastering the JSP Standard Tag Library"
     Chapter:  "The JSTL Expression Language"
 
     see: https://www.informit.com/store/core-jstl-mastering-the-jsp-standard-tag-library-9780131001534?w_ptgrevartcl=The+JSTL+Expression+Language_30946

Implementation follows https://www.informit.com/articles/article.aspx?p=30946

Inspired by https://forum.sublimetext.com/t/import-in-custom-syntax-default-ones/61716/2 and the linked repo.

### DIFF
--- a/Java/HTML (JSP).sublime-syntax
+++ b/Java/HTML (JSP).sublime-syntax
@@ -13,6 +13,24 @@ file_extensions:
   - jspf
   - jspx
 
+variables:
+  # Identifiers
+  break: (?!{{id_char}})
+
+  id: (?:{{id_first_char}}{{id_char}}*)
+  id_first_char: '[\p{L}_$]'
+  id_char: '[\p{L}\p{N}_$]'
+
+  implicit_objects: |-
+    (?x:
+      param | paramValues | header | headerValues | initParam | cookie
+    | pageScope | requestScope | sessionScope | applicationScope | pageContext
+    ){{break}}
+
+  # digits
+  dec_digit: '\d'
+  dec_exponent: '[eE][-+]?{{dec_digit}}*'
+
 contexts:
 
   prototype:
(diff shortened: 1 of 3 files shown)


Java/HTML (JSP).sublime-syntax
```diff
@@ -12,6 +12,25 @@ file_extensions:
   - jsp
   - jspf
   - jspx
+  - jstl
+
+variables:
```
jrappen, line 17: Maybe it would make sense to add a comment line here with a link to:

<https://github.com/kkos/oniguruma/blob/master/doc/UNICODE_PROPERTIES>

for those unfamiliar with the usage of named properties...maybe not. Your call.

deathaxe: If you mean `\p{L}` and `\p{N}` ... those variables are just copied from Java.sublime-syntax which has been using them a long time. It's not the task of these syntax files to teach noops the basics of regular expressions or any other stupid basics. If someone is curious about them a quick google/duckduck search is a proper solution. Same I did, when I saw them the very first time.

jrappen: yeah ok, nvm then

jrappen (approved):

michaelblyons (approved):

keith-hall (approved):

Issue #3139, opened 2021-11-24
[PHP] Comma within an array is not scoped.


Ultra-Instinct-05:
### What happened?

The comma within an array is not scoped appropriately. It should be scoped as `punctuation.seperator.comma.php`

![image](https://user-images.githubusercontent.com/34304644/143263525-d7619157-7c09-430b-83d9-ae832166a529.png)

```php
<?php 

    var_dump([...$array_1, ...$array_2]);

 ?>
```

Issue #3140, opened 2021-11-24
[PHP] Indentation fails after the colon in a `switch..case` statement


Ultra-Instinct-05:
### Expected behavior

Here's a sample PHP snippet.

```php
<?php
    $user_role = "Admin";
    switch ($user_role) {
        case "Admin":
            echo "You are an admin !";
            break;
        case "Editor":
            echo "You are an editor";
            break;
        default:
            echo "Invalid role";
    }
?>
```
The expected behavior should be for the next line to be indented one level further when the cursor is at `case "Admin": | <- cursor`.

As an additional bonus, the `colon` is not scoped appropriately.

### Actual behavior

The line is not indented one level further. It stops right under the `case` keyword.

### Steps to reproduce

1. Copy the above PHP snippets in a sample php file in safe mode.
2. Place the cursor after `case "Admin":`
3. Press <kbd>enter</kbd>

deathaxe:
Main blocker is described at https://github.com/sublimehq/Packages/pull/3660#issuecomment-1419543038

PR #3141, opened 2021-11-24
[PHP] Fix comma scope in arrays

jfcherng:
- Fixes https://github.com/sublimehq/Packages/issues/3139
- Add colon scope for https://github.com/sublimehq/Packages/issues/3140

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -62,7 +62,7 @@ contexts:
             - match: '(?={{path}})'
               push: inherited
             - match: ','
-              scope: punctuation.separator.php
+              scope: punctuation.separator.comma.php
             - match: '(?=\S)'
               pop: 1
         - match: '(?=\S)'
@@ -88,7 +88,7 @@ contexts:
     - match: '^\s*((?!default|else){{identifier}})\s*(:)(?!:)'
       captures:
         1: entity.name.label.php
-        2: punctuation.separator.php
+        2: punctuation.separator.colon.php
     - match: \{
       scope: punctuation.section.block.begin.php
     - match: \}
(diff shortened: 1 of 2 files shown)


Ultra-Instinct-05 (dismissed):

jrappen:
@jfcherng so does this fix 3140 as well?

jfcherng:
> @jfcherng so does this fix 3140 as well?

No. It just add a supplement for the `:` scope mentioned in #3140.

jfcherng:
then i think all comma scope should be revised.

Ultra-Instinct-05:
Here is another situation where comma is not scoped.

image

```php
<?php 
    class User {

        public function __construct($username, $email) {
            $this->username = $username;
            $this->email = $email;
        }

    }
    
    $user_1 = new User("John", "a@b.com");
    $user_2 = new User("Doe", "c@d.com");
?>
```

jfcherng:
There are many scattered "comma" scope in the syntax def at this moment. Also, I don't know what scope should be use for the "colon".

deathaxe:
I've choosen `comma` and `colon` in various situations in Java as well as I find them a proper choise in syntaxes which don't make extensive use of semantic highlighting.

Probably `punctuation.separator.sequence` could be used in arrays and `punctuation.separator.arguments` in argument lists, but this would probably require to refine some meta scopes as well to keep consistent. That's probably a future task.

With a look at all currently used scopes there is no real standard to group/categorize 3rd level scopes anyway. Thus a color scheme can address `punctuation.separator` only to create consistent highlighting.

Another question is, whether someone is really interested in highlighting commas different depending on usecase.

So I am fine with the choosen scopes.

deathaxe (approved):

deathaxe (approved):

keith-hall (approved):

PR #3148, opened 2021-11-28
[C/C++/ObjC/ObjC++] Improve Indentation Rules

### LINKED ISSUE #868: [C] Wrong auto-indentation level after a loop without body

### LINKED ISSUE #3145: [C/C++] Auto-indent mistakes extern "C" {...} as indentable block after preprocessor lines
### Expected behavior

In C header files, this boilerplate is commonly included for compatibility with C++ compilers:

```c
#ifdef __cplusplus
extern "C" {
#endif

// Header file contents...

#ifdef __cplusplus
}
#endif
```

Although this construct involves curly braces, wrapping it around a header file's contents should not affect indentation.

### Actual behavior

When inserting a newline between the two halves of the extern "C" {...} statement, if the cursor is on a line with a preprocessor statement, an undesired extra tab is inserted at the beginning of the new line. Example:

```c
#ifdef __cplusplus
extern "C" {
#endif

#include <stdlib.h> // Press enter key here
	// A tab character is inserted at the beginning of this line

#ifdef __cplusplus
}
#endif
```

### Steps to reproduce

Minimal example:

```c
extern "C" {
#include // Press enter key with cursor here
}
```

deathaxe:
Fixes #868
Fixes #3145

This commit copies indentation rules from Java and adds C/C++ specific rules/keywords as well as some indentation tests.

It can't solve all problems, but may help to improve some of the known issues.

### DIFF
--- a/C++/Indentation Rules.tmPreferences
+++ b/C++/Indentation Rules.tmPreferences
@@ -6,26 +6,102 @@
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string>(?x)
-		^ (.*\*/)? \s* \} .* $
-		|   ^ \s* (public|private|protected): \s* $
-		|   ^ \s* @(public|private|protected) \s* $
-		</string>
+		<string><![CDATA[(?x)
+			# line beginning with whitespace or block comments
+			^ (.*\*/)? \s*
+			(?:
+			# dedent closing braces
+			  \}
+			# detent `case ... :`
+			| case\b
+			# detent `default:`
+			| default\b
+			# class member access attributes
+			| ( public | private | protected ):
+			# objc/objc++ attributes
+			| @( public | private | protected )\b
+			)
+		]]></string>
+
 		<key>increaseIndentPattern</key>
-		<string>(?x)
-		^ .* \{ [^}"']* $
-		|   ^ \s* (public|private|protected): \s* $
-		|   ^ \s* @(public|private|protected) \s* $
-		</string>
+		<string><![CDATA[(?x)
+			# line beginning with whitespace or block comments
+			^ (.*\*/)? \s*
+			(?:
+			# indent after opening braces (may be followed by whitespace or comments)
+			# but exclude lines such as `extern "C" {`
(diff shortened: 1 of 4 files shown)


C++/Indentation Rules.tmPreferences
```diff
@@ -6,26 +6,102 @@
...
+		]]></string>
 
 		<key>bracketIndentNextLinePattern</key>
-		<string>(?x)
-		^ \s* \b(if|while|else)\b [^;]* $
-		| ^ \s* \b(for)\b .* $
-		</string>
+		<string><![CDATA[(?x)
+			# line beginning with whitespace or block comments
+			^ (.*\*/)? (?<comment_or_whitespace> (?: \s* (?<block_comment> /\*.*\*/ ) )* \s* )
```
keith-hall, line 48: I think there is a potential for catastrophic backtracking here, with `\s*\s*` effectively if there are no block comments

deathaxe: `(?<block_comment> /\*.*\*/ )` is not optional, so the whole `(?: \s* (?<block_comment> /\*.*\*/ ) )*` wouldn't match if no comment is available. Hence the last `\s*` is the only one to stay then.

Do I miss something?

keith-hall: Nope, sorry - just me trying to review directly on GitHub (i.e. without syntax highlighting) while looking after the kids :sweat_smile:

keith-hall (approved):

michaelblyons (approved):

PR #3149, opened 2021-11-28
[Python] Add shebang

deathaxe:
This commit adds a dedicated context to scope the shebang line.

Some fonts like provide ligatures for the `#!` sign, thus need to ensure it is tokenized correctly.

The statements context is set onto stack if shebang matches but not after the next start of line or non-whitespace character. This is to make sure, the shebang is highlighted within fenced code blocks of Markdown.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -29,8 +29,8 @@ file_extensions:
 
 first_line_match: |-
   (?xi:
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

jfcherng (approved):

jrappen:
No test for the markdown codeblock?

PR #3152, opened 2021-11-29
[PHP] Improve heredoc matching and add support for highlighting Yaml in heredocs

bakerkretzmar:
This PR simplifies and improves some of the matching of heredoc tokens in PHP:

- Removes empty captures, like the two occurrences of `()` in `(()HTML())`. This has no effect on any tests or scoping, since these captures are always empty.
- Matches all heredoc tokens case-insensitively, since these are also valid PHP syntax. I updated one existing test to use `sql` instead of `SQL` to verify that this works as expected.
- Refer to the matched heredoc token using a numbered reference in the `escape` pattern, instead of repeating the token (e.g. `(?=^\s*\2\b)` instead of `(?=^\s*HTML\b)`). This is a bit less repetitive and makes it easier to update the token pattern itself if it changes.
- Add `JS` as a 'recognized' heredoc token (in addition to `JAVASCRIPT`) and apply JavaScript syntax highlighting to the embedded heredoc text.

This PR also adds support for embedded Yaml syntax highlighting inside heredocs using the tokens `YAML` and `YML`. I implemented this in exactly the same way as the other embedded syntaxes, and added a similar test.

### DIFF
--- a/PHP/tests/syntax_test_php.php
+++ b/PHP/tests/syntax_test_php.php
@@ -1925,6 +1925,9 @@ three: "$four"
 //   ^       punctuation.separator.key-value.mapping
 //     ^^^^^^^ string.quoted.double
 //      ^^^^^ variable.other.php
+Yml;
+// <- meta.embedded.yaml source.yaml string.unquoted.plain.out.yaml
+//^^ meta.embedded.yaml source.yaml string.unquoted.plain.out.yaml
 yml;
 // <- entity.name.tag.heredoc
 // ^ punctuation.terminator.expression
--- a/PHP/tests/syntax_test_php.php
+++ b/PHP/tests/syntax_test_php.php
@@ -1910,7 +1910,7 @@ CSS;
 // ^ punctuation.terminator.expression
 //  ^ meta.heredoc-end
 
-echo <<<SQL
+echo <<<sql
 //   ^^^ keyword.operator.heredoc
 //      ^^^ meta.string.heredoc meta.tag.heredoc
 //      ^^^ entity.name.tag.heredoc
@@ -1920,7 +1920,7 @@ SELECT * FROM users WHERE first_name = 'John' LIMIT $limit
 //     ^ variable.language.wildcard.asterisk
 //                                     ^^^^^^ string.quoted.single
 //                                                  ^^^^^^ variable.other.php
-SQL;
+sql;
 // <- entity.name.tag.heredoc
 // ^ punctuation.terminator.expression
 //  ^ meta.heredoc-end
--- a/PHP/tests/syntax_test_php.php
+++ b/PHP/tests/syntax_test_php.php
@@ -1910,6 +1910,26 @@ CSS;
 // ^ punctuation.terminator.expression
 //  ^ meta.heredoc-end
 
+echo <<< yml
+//   ^^^ keyword.operator.heredoc
+//       ^^^ meta.string.heredoc meta.tag.heredoc
+//       ^^^ entity.name.tag.heredoc
+one: two
+//^^^^^^ meta.embedded.yaml source.yaml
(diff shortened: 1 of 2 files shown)


deathaxe:
The more important part about `(?=^\s*\2\b)` is the closing tag to need to match the opening one case-sensitively.

keith-hall:
Thanks for your contribution.

> * Removes empty captures, like the two occurrences of `()` in `(()HTML())`. This has no effect on any tests or scoping, since these captures are always empty.

My understanding is that it was done like this to keep the capture scopes consistent/the same for each match pattern, regardless of whether it was quoted or not.

> * Refer to the matched heredoc token using a numbered reference in the `escape` pattern, instead of repeating the token (e.g. `(?=^\s*\2\b)` instead of `(?=^\s*HTML\b)`). This is a bit less repetitive and makes it easier to update the token pattern itself if it changes.

It could make sense to put it in a variable for easier re-use if it is now always the same pattern.

PHP/tests/syntax_test_php.php
```diff
@@ -1910,7 +1910,27 @@ function generate2()
...
+//^ string.unquoted.plain.out entity.name.tag
+// ^       punctuation.separator.key-value.mapping
+//   ^^^ string.unquoted.plain.out
+three: "$four"
+//^^^^^^^^^^^^ meta.embedded.yaml source.yaml
+//^^^ string.unquoted.plain.out entity.name.tag
+//   ^       punctuation.separator.key-value.mapping
+//     ^^^^^^^ string.quoted.double
+//      ^^^^^ variable.other.php
+yml;
```
deathaxe, line 1928: Maybe a test to ensure case sensitive matching of closing tags would be useful?

```suggestion
Yml;
// <- meta.embedded.yaml source.yaml string.unquoted.plain.out.yaml
//^^ meta.embedded.yaml source.yaml string.unquoted.plain.out.yaml
yml;
```

deathaxe:
> My understanding is that it was done like this to keep the capture scopes consistent

Not sure if it is really useful. I'd prefer proposed solution, tbh.

bakerkretzmar:
> The more important part about `(?=^\s*\2\b)` is the closing tag to need to match the opening one case-sensitively.

Great point. I assumed that `\2` would be exactly whatever was captured, with the same casing, but I'll make sure the tests confirm this.

bakerkretzmar:
> My understanding is that it was done like this to keep the capture scopes consistent/the same for each match pattern, regardless of whether it was quoted or not.

Do you mean the syntax definition blocks being consistent? Like every single one having exactly 5 captures? Fair enough, that just didn't seem particularly useful to me.

The actual captured scopes are identical even after my changes, because the empty capturing groups created empty captures and the corresponding scopes weren't actually added.

Both of these blocks:

```yaml
- match: (<<<)\s*(()JSON())\s*$\n?
  captures:
    1: keyword.operator.heredoc.php
    2: meta.string.heredoc.php meta.tag.heredoc.php
    3: punctuation.definition.tag.begin.php
    4: entity.name.tag.heredoc.php
    5: punctuation.definition.tag.end.php
```
```yaml
- match: (<<<)\s*(JSON)\s*$\n?
  captures:
    1: keyword.operator.heredoc.php
    2: meta.string.heredoc.php meta.tag.heredoc.php entity.name.tag.heredoc.php
```

Result in the following scoping:

```php
echo <<<SQL
//   ^^^ keyword.operator.heredoc
//      ^^^ meta.string.heredoc meta.tag.heredoc
//      ^^^ entity.name.tag.heredoc
```

> It could make sense to put it in a variable for easier re-use if it is now always the same pattern.

I love this idea but I don't know how to do that 😅

keith-hall (approved):

bakerkretzmar:
Thanks a lot for your help and feedback! 🚀

PR #3169, opened 2021-12-15
[C#] Improve pattern matching support for C#10

keith-hall:


deathaxe:
Should this work as well?

```cs
static bool IsConferenceDay(DateTime date) => date is { Year: 2020, Month: 5, Day: 19 or 20 or 21 };
```

grafik

keith-hall:
yep, it should work - and I just pushed a commit which fixes it, so it does now :)

deathaxe (approved):

michaelblyons (approved):

PR #3173, opened 2021-12-21
[Python] Fix shebang highlighting for python 3.10+

### LINKED ISSUE #3172: [Python] Shebang highlighting for python 3.10+ incorrect
### What happened?

Python's shebang line is incorrectly highlighted by Sublime Text 4126 (latest stable build released 21 December 2021) if the Python version is specific.

There was no highlight of the shebang line before, but now since version 4126, it highlights
#!/usr/bin/env `python3`.10
**instead of**
#!/usr/bin/env `python3.10`

You missed the fact that the Python version in the shebang line can be specific, like `3.10` in the example above, not just `3`, so the highlight is incorrect.

See the below screenshot to better understand the issue...
screenshot

deathaxe:
Fixes #3172

This commit matches python minor version parts of two digits of length.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -29,8 +29,8 @@ file_extensions:
 
 first_line_match: |-
   (?xi:
-    ^ \#! .* \bpython(?:\d(?:\.\d)?)?\b                        # shebang
-  | ^ \s* \# .*? -\*- .*? \bpython(?:\d(?:\.\d)?)?\b .*? -\*-  # editorconfig
+    ^ \#! .* \bpython(?:\d(?:\.\d+)?)?\b                        # shebang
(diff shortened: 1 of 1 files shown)


Thom1729 (approved):

jfcherng (approved):

PedanticHacker:
Thank you for fixing this. I hope this gets merged with master and deployed with an update soon.

Python/Python.sublime-syntax
```diff
@@ -171,7 +171,7 @@ contexts:
     - meta_include_prototype: false
     - meta_scope: comment.line.shebang.python
     # Note: Keep sync with first_line_match!
```
michaelblyons, line 173: Is there a reason not to use a variable?

keith-hall: The comment might still be necessary :
https://github.com/sublimehq/sublime_text/issues/4713

PR #3179, opened 2021-12-26
[HTML] Add syntax highlighting support for JSON scripts

### LINKED ISSUE #3176: [HTML] Missed highlight for JSON block in the HTML
### What happened?

I got the yahoo site webpage. big html file. I see that included json blocks are missing their hilite. Here is the pic.

Screenshot from 2021-12-25 19-31-36

Block starts with
```
<script id="wafer-config" type="application/json">
            {"assetsLoaded":{"/aaq/fp/css/tdv2-applet-native-ads.PencilAd.atomic.ltr.c6d68ffae4f433d3aeaa08874f29f6fe.min.css":true,"/aaq/wf/wf-core-1.49.4-modern.js":true,"/aaq/wf/wf-beacon-1.3.3-modern.js":true,"/aaq/wf/wf-lightbox-1.5.4-mod
```

deathaxe:
Closes #3176

This commit adds JSON to the list of supported syntaxes between script tags.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -43,6 +43,9 @@ variables:
       | text/livescript
     )
 
+  # https://mimesniff.spec.whatwg.org/#json-mime-type
+  json_mime_type: (?i:application/|text/|.+\+)json
+
   # Embedded script and style syntaxes may be wrapped into html comments for
   # historical reasons. The following patterns match them, while maintaining
   # correct boundaries of embedded source scopes. That's required to enable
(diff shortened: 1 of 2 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -284,6 +318,11 @@ contexts:
         - script-javascript
         - tag-generic-attribute-meta
         - tag-generic-attribute-value
+    - match: (?=(?i:application/json{{unquoted_attribute_break}}|'application/json'|"application/json"))
```
jfcherng, line 321: https://mimesniff.spec.whatwg.org/#json-mime-type

> A JSON MIME type is any MIME type whose subtype ends in "+json" or whose essence is "application/json" or "text/json".

deathaxe: With more complex patterns to be matched a variable makes very much sense.

keith-hall (approved):

jfcherng (approved):

PR #3180, opened 2021-12-26
[JavaScript] Fix bug when some class field names are not followed by a semicolon or initializer

Thom1729:
Reported by Mitranim on Discord.

Private and computed field names were not popping the `field-name` context, so if the next token were an identifier, it would highlight wrong. (Class fields are usually followed by initializers, type annotations, or semicolons, but they don't have to be.)

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1834,9 +1834,9 @@ contexts:
       captures:
         1: punctuation.definition.variable.js
         2: variable.other.readwrite.js
-
+      pop: true
     - match: (?=\[)
-      push: computed-property-name
+      set: computed-property-name
 
     - include: else-pop
 
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1719,9 +1719,7 @@ contexts:
         - object-property-name
 
   object-property-name:
-    - match: (?=\[)
-      set: computed-property-name
-
+    - include: computed-property-name
     - include: literal-string
     - include: literal-number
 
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -1834,9 +1834,9 @@ contexts:
       captures:
         1: punctuation.definition.variable.js
         2: variable.other.readwrite.js
-
+      pop: true
     - match: (?=\[)
-      push: computed-property-name
+      set: computed-property-name
```
deathaxe, line 1839: Would it make sense to directly include the context?

```yaml
    - include: computed-property-name
```

_Note: It would apply to the previous `method-name` context as well._

JavaScript/JavaScript.sublime-syntax
```diff
@@ -1834,9 +1834,9 @@ contexts:
       captures:
         1: punctuation.definition.variable.js
         2: variable.other.readwrite.js
-
+      pop: true
```
deathaxe, line 1837: How about moving the whole pattern upwards to achieve the same order as in `method-name` context?

_Both contexts seem to match the same tokens but only apply different scopes. It would therefore probably make sense to align their formal content._

JavaScript/JavaScript.sublime-syntax
Thom1729: Good call. I've also done this in `object-property-name`.

JavaScript/JavaScript.sublime-syntax
Thom1729: Done.

deathaxe (approved):

jfcherng (approved):

Issue #3193, opened 2022-01-10
[rust] No `const` highlighting in struct


ismell:
### What happened?

```
bitflags! {
    struct Flags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;
        const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
    }
}
```

There is currently no highlighting in the above code.
![Screenshot from 2022-01-10 10-20-14](https://user-images.githubusercontent.com/562978/148809922-63a5b6c6-ddc8-4e80-bac5-feec3e9957c2.png)

I would expect `A`, `B`, `C`, and `ABC` to be marked as entities and appear in the global symbol list.

deathaxe:
With PR #2305 merged the struct looks like...

![grafik](https://user-images.githubusercontent.com/16542113/148828105-81880047-6eed-4f7f-b05f-e6c6942e62cb.png)

The constants' identifiers are scoped `entity.name.module.rust` and could therefore easily be indexed.

ismell:
Yay! Thanks.

deathaxe:
With #3195 constants are indexed. Thus closing this issue as resolved.

Issue #458, opened 2016-06-06
Scopes for constant variables


FichteFoll:
(via #393)

It seems that the Ruby syntax defines `entity.name.type.constant.ruby` scopes for "definitions" to "constant variables". However, as with almost all other languages, this naming scheme is [not enforced](http://rubylearning.com/satishtalim/ruby_constants.html). Since this UPPER_CASE scheme is shared among almost all languages, it makes sense to unify this behavior for all languages. As such, I propose the following changes:

1. Highlight occurences of constants as `variable.constant` (sounds contradicting) or `constant.other.user` or similar.
1. If desired, make assignments to constant variables receive a scope that will/can be indexed, i.e. `entity.name.constant`.

Opinions?

wbond:
So far C uses `entity.name.constant`. I'll update Ruby to use this scope also since I imagine users will often want to highlight constants different than types.

Currently Ruby uses `variable.other.constant` for references to it. While C code usually uses uppercase for preprocessor macros, some classes are all uppercase also, and it would be odd to highlight those as constants. I feel like we will only be able to pull this off in languages that enforce formatting of variables versus constants, etc.

FichteFoll:
Ruby emits warnings when re-assigning to uppercase identifiers, so I guess that counts? I don't know how it behaves with all-uppercase class names though.

Python only mentiones this in PEP8.

FichteFoll:
For Python, via #1642, I made the minimum requirement of the identifier only consisting of upper-case letters, digits and underscores and having at least 2 consecutive letters. Imo this is heuristic is good enough, at least for Python code, but I believe this convention is followed among most languages that aren't C (or have macros in upper-case).

PR #3195, opened 2022-01-10
[Rust] Tweak tmPreferences

deathaxe:
This commit applies some changes to Rust's tmPreferences to re-align them with the other packages.

We might want to discuss, whether we want to add `entity.name.constant.rust` to the symbol list as requested by #3193 as it would cause all global and local constant declarations to be indexed.


keith-hall (approved):

keith-hall:
There's an RFC for the scoping and indexing of constants at https://github.com/sublimehq/Packages/issues/458

deathaxe:
I can't find something about indexing in it. What I have in mind is variables not to be scoped in general as those would cause too much noise. Rust scopes constants `entity.name.constant` in declarations only. So the only concern is whether local constant declarations (in functions) are of interest in a global index.

jrappen (approved):

michaelblyons (approved):

Rust/tests/symbols/syntax_test_symbols.rs
```diff
@@ -0,0 +1,32 @@
...
+
+struct MyStruct {
+//     @@@@@@@@ definition
+
+    const A = 0;
+//        @ definition
+
+    fn bar() {}
+//     @@@ definition
+}
```
FichteFoll, line 23: This is actually invalid syntax. What it should be is:

```rust
struct MyStruct {}
//     @@@@@@@@ definition

impl MyStruct {
    const A: i32 = 0;
//        @ definition

    fn bar() {}
//     @@@ definition
}
```

PR #3198, opened 2022-01-12
[JSON] Add GLTF to the list of file_extensions

willstott101:
GLTF is the JSON encoding of a widely used 3D scene file format. A separate extension `glb` is used for the binary format, the GLTF format is generally only useful for debugging - and is therefore opened often in my sublime.

Thought it might be helpful for others. Though I'm not sure how exhaustive these lists are intended to be.

### DIFF
--- a/JSON/JSON.sublime-syntax
+++ b/JSON/JSON.sublime-syntax
@@ -20,6 +20,7 @@ file_extensions:
   - sublime-workspace
   - ipynb
   - Pipfile.lock
+  - gltf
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

jfcherng (approved):

jrappen:
Hello Will.

If you have further feedback for the JSON package, I'm working on a re-write at #3097 currently and would appreciate any input.

PR #3205, opened 2022-01-19
[Javascript] Extra space for block comments

dsl101:
Add spaces around block comments, to match style for single-line comments. Understand this might be an individual preference, but when using for 'inline' comments such as `const { destructure, /* notUsed */ } = thing`, it's much more readable. For multi-line comments, I always use non-block (`// `) which already has a space.

Adding as an option would be ideal, but beyond me I'm afraid :), so I could define it as:
```
  { "keys": ["ctrl+f2"], "command": "toggle_comment", "args": { "block": true, "space": true } },
```

If you can point me where to look to add such an option, I'm happy to give it a go...

### DIFF
--- a/JavaScript/Comments.tmPreferences
+++ b/JavaScript/Comments.tmPreferences
@@ -17,13 +17,13 @@
 				<key>name</key>
 				<string>TM_COMMENT_START_2</string>
 				<key>value</key>
-				<string>/*</string>
+				<string>/* </string>
 			</dict>
 			<dict>
 				<key>name</key>
 				<string>TM_COMMENT_END_2</string>
 				<key>value</key>
-				<string>*/</string>
+				<string> */</string>
 			</dict>
 		</array>
 	</dict>


jwortmann:
I'm not sure if it would be a good idea to change the default style here for JavaScript, because most (but not all) of the other syntaxes also don't use additional spaces for the block comments. But I agree that the style with a space is more readable and it would be consistent with the single line comments.

The `toggle_comment` command is defined in the comment.py file from the Default package, and your proposed alternative with the optional "space" parameter for the keybinding should be possible with a small tweak, for example like this:
```diff
--- Shipped Packages/Default/comment.py	2021-04-02 13:30:54
+++ Packages/Default/comment.py	2022-01-21 14:05:02
@@ -22,7 +22,7 @@
             return c != "\n"
 
 
-def build_comment_data(view, pt):
+def build_comment_data(view, pt, space):
     shell_vars = view.meta_info("shellVariables", pt)
     if not shell_vars:
         return ([], [])
@@ -42,6 +42,12 @@
         start = all_vars.setdefault("TM_COMMENT_START" + suffix)
         end = all_vars.setdefault("TM_COMMENT_END" + suffix)
         disable_indent = all_vars.setdefault("TM_COMMENT_DISABLE_INDENT" + suffix)
+
+        if space:
+            if start and not start.endswith(" "):
+                start += " "
+            if end and not end.startswith(" "):
+                end = " " + end
 
         if start and end:
             block_comments.append((start, end, disable_indent == 'yes'))
@@ -195,11 +201,11 @@
                 # add a line comment to each line
                 self.line_comment_region(view, edit, line_comments[0], region)
 
-    def run(self, edit, block=False):
+    def run(self, edit, block=False, space=False):
         for region in self.view.sel():
-            comment_data = build_comment_data(self.view, region.begin())
+            comment_data = build_comment_data(self.view, region.begin(), space)
             if (region.end() != self.view.size() and
-                    build_comment_data(self.view, region.end()) != comment_data):
+                    build_comment_data(self.view, region.end(), space) != comment_data):
                 # region spans languages, nothing we can do
                 continue
 ```

Full file:
<details>
 <summary>click to expand</summary>

```python
import sublime
import sublime_plugin


def advance_to_first_non_white_space_on_line(view, pt):
    while True:
        c = view.substr(pt)
        if c == " " or c == "\t":
            pt += 1
        else:
            break

    return pt


def has_non_white_space_on_line(view, pt):
    while True:
        c = view.substr(pt)
        if c == " " or c == "\t":
            pt += 1
        else:
            return c != "\n"


def build_comment_data(view, pt, space):
    shell_vars = view.meta_info("shellVariables", pt)
    if not shell_vars:
        return ([], [])

    # transform the list of dicts into a single dict
    all_vars = {}
    for v in shell_vars:
        if 'name' in v and 'value' in v:
            all_vars[v['name']] = v['value']

    line_comments = []
    block_comments = []

    # transform the dict into a single array of valid comments
    suffixes = [""] + ["_" + str(i) for i in range(1, 10)]
    for suffix in suffixes:
        start = all_vars.setdefault("TM_COMMENT_START" + suffix)
        end = all_vars.setdefault("TM_COMMENT_END" + suffix)
        disable_indent = all_vars.setdefault("TM_COMMENT_DISABLE_INDENT" + suffix)

        if space:
            if start and not start.endswith(" "):
                start += " "
            if end and not end.startswith(" "):
                end = " " + end

        if start and end:
            block_comments.append((start, end, disable_indent == 'yes'))
            block_comments.append((start.strip(), end.strip(), disable_indent == 'yes'))
        elif start:
            line_comments.append((start, disable_indent == 'yes'))
            line_comments.append((start.strip(), disable_indent == 'yes'))

    return (line_comments, block_comments)


class ToggleCommentCommand(sublime_plugin.TextCommand):
    def remove_block_comment(self, view, edit, comment_data, region):
        (line_comments, block_comments) = comment_data

        # Call extract_scope from the midpoint of the region, as calling it
        # from the start can give false results if the block comment begin/end
        # markers are assigned their own scope, as is done in HTML.
        whole_region = view.extract_scope(region.begin() + region.size() / 2)

        for c in block_comments:
            (start, end, disable_indent) = c
            start_region = sublime.Region(whole_region.begin(), whole_region.begin() + len(start))
            end_region = sublime.Region(whole_region.end() - len(end), whole_region.end())

            if view.substr(start_region) == start and view.substr(end_region) == end:
                # It's faster to erase the start region first
                view.erase(edit, start_region)

                end_region = sublime.Region(
                    end_region.begin() - start_region.size(),
                    end_region.end() - start_region.size())

                view.erase(edit, end_region)
                return True

        return False

    def remove_line_comment(self, view, edit, comment_data, region):
        (line_comments, block_comments) = comment_data

        found_line_comment = False

        start_positions = [advance_to_first_non_white_space_on_line(
            view, r.begin()) for r in view.lines(region)]
        start_positions.reverse()

        for pos in start_positions:
            for c in line_comments:
                (start, disable_indent) = c
                comment_region = sublime.Region(pos, pos + len(start))
                if view.substr(comment_region) == start:
                    view.erase(edit, comment_region)
                    found_line_comment = True
                    break

        return found_line_comment

    def is_entirely_line_commented(self, view, comment_data, region):
        (line_comments, block_comments) = comment_data

        start_positions = [advance_to_first_non_white_space_on_line(
            view, r.begin()) for r in view.lines(region)]
        start_positions = list(filter(
            lambda p: has_non_white_space_on_line(view, p), start_positions))

        if len(start_positions) == 0:
            return False

        for pos in start_positions:
            found_line_comment = False
            for c in line_comments:
                (start, disable_indent) = c
                comment_region = sublime.Region(pos, pos + len(start))
                if view.substr(comment_region) == start:
                    found_line_comment = True
            if not found_line_comment:
                return False

        return True

    def block_comment_region(self, view, edit, block_comment_data, region):
        (start, end, disable_indent) = block_comment_data

        if region.empty():
            # Silly buggers to ensure the cursor doesn't end up after the end
            # comment token
            view.replace(edit, sublime.Region(region.end()), 'x')
            view.insert(edit, region.end() + 1, end)
            view.replace(edit, sublime.Region(region.end(), region.end() + 1), '')
            view.insert(edit, region.begin(), start)
        else:
            view.insert(edit, region.end(), end)
            view.insert(edit, region.begin(), start)

    def line_comment_region(self, view, edit, line_comment_data, region):
        (start, disable_indent) = line_comment_data

        start_positions = [r.begin() for r in view.lines(region)]
        start_positions.reverse()

        # Remove any blank lines from consideration, they make getting the
        # comment start markers to line up challenging
        non_empty_start_positions = list(filter(
            lambda p: has_non_white_space_on_line(view, p), start_positions))

        # If all the lines are blank however, just comment away
        if len(non_empty_start_positions) != 0:
            start_positions = non_empty_start_positions

        if not disable_indent:
            min_indent = None

            # This won't work well with mixed spaces and tabs, but really,
            # don't do that!
            for pos in start_positions:
                indent = advance_to_first_non_white_space_on_line(view, pos) - pos
                if min_indent is None or indent < min_indent:
                    min_indent = indent

            if min_indent is not None and min_indent > 0:
                start_positions = [r + min_indent for r in start_positions]

        for pos in start_positions:
            view.insert(edit, pos, start)

    def add_comment(self, view, edit, comment_data, prefer_block, region):
        (line_comments, block_comments) = comment_data

        if len(line_comments) == 0 and len(block_comments) == 0:
            return

        if len(block_comments) == 0:
            prefer_block = False

        if len(line_comments) == 0:
            prefer_block = True

        if region.empty():
            if prefer_block:
                # add the block comment
                self.block_comment_region(view, edit, block_comments[0], region)
            else:
                # comment out the line
                self.line_comment_region(view, edit, line_comments[0], region)
        else:
            if prefer_block:
                # add the block comment
                self.block_comment_region(view, edit, block_comments[0], region)
            else:
                # add a line comment to each line
                self.line_comment_region(view, edit, line_comments[0], region)

    def run(self, edit, block=False, space=False):
        for region in self.view.sel():
            comment_data = build_comment_data(self.view, region.begin(), space)
            if (region.end() != self.view.size() and
                    build_comment_data(self.view, region.end(), space) != comment_data):
                # region spans languages, nothing we can do
                continue

            if self.remove_block_comment(self.view, edit, comment_data, region):
                continue

            if self.is_entirely_line_commented(self.view, comment_data, region):
                self.remove_line_comment(self.view, edit, comment_data, region)
                continue

            has_line_comment = len(comment_data[0]) > 0

            if not has_line_comment and not block and region.empty():
                # Use block comments to comment out the line
                line = self.view.line(region.a)
                line = sublime.Region(
                    advance_to_first_non_white_space_on_line(self.view, line.a), line.b)

                # Try and remove any existing block comment now
                if self.remove_block_comment(self.view, edit, comment_data, line):
                    continue

                self.add_comment(self.view, edit, comment_data, block, line)
                continue

            # Add a comment instead
            self.add_comment(self.view, edit, comment_data, block, region)

```
</details>

This would apply to all syntaxes then, and to both line and block comments - for line comments it shouldn't have any effect though, because they already have the space. So you don't have to care anymore whether putting the space in the .tmPreferences file or not, but at the same time this change would limit the current ability to use different styles (i.e. with or without space) for different syntaxes.

The Default package is out of scope for this repository, so you'd need to create it as a feature request in https://github.com/sublimehq/sublime_text. Or use it as a personal override for the file - in that case I would recommend using the OverrideAudit package from Package Control to get a notification when overrides get outdated.

dsl101:
Thanks for the comment (yes, intended :). I think your code to allow the `space` parameter is much better, and I've also learnt about OverrideAudit, so even betterer. I've made an override for `Default/comment.py` to test that change, and if all looks good I will make a PR for the main project.

PR #3206, opened 2022-01-19
[JavaScript] Use mapping and collection bracket scopes

Thom1729:
Use `punctuation.section.mapping` and `punctuation.section.sequence` where appropriate, including:

- Array and object literals.
- Array and object destructuring.
- TypeScript tuple types.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -501,22 +501,22 @@ contexts:
 
   variable-binding-array-destructuring:
     - match: '\['
-      scope: punctuation.section.brackets.begin.js
+      scope: punctuation.section.sequence.begin.js
       set:
         - meta_scope: meta.binding.destructuring.sequence.js
         - match: '\]'
-          scope: punctuation.section.brackets.end.js
+          scope: punctuation.section.sequence.end.js
           pop: true
         - include: variable-binding-spread
         - include: variable-binding-list
 
   variable-binding-object-destructuring:
     - match: '\{'
-      scope: punctuation.section.block.begin.js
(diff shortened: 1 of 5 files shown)


jrappen (approved):

deathaxe (approved):

jfcherng (approved):

ratijas:
Hi, just a quick note: please be careful with changes like this. It broke my CI for a syntax which uses JavaScript as a base/`extends`, and I was very confused about it: tests are passing locally but failing in cloud. Not sure what's the best way to handle such situations though

FichteFoll:
It's quite natural and unavoidable to cause such assertion failures when depending on a different syntax. For your case in particular, you were probably editing the test file locally on a recent-ish ST build while your CI has version pinning for ST builds and Packages repo refs. The only way not to run into this would be to never update any scopes in the default packages, which is a non-starter.

In cases like this, you need to decide what you want to base your tests on and either make them compatible with multiple builds by adding OR operators in your selectors or only look at the result of the ST build you want to support.

jrappen:
Also compare:

<https://github.com/sublimehq/Packages/blob/master/.github/workflows/ci.yml>

where you can see that the Packages repo (i.e. this one) is tested against the latest Builds from Sublime Text in the `stable` & `dev` channels.

That test against the latest `stable` Builds is "optional".

ratijas:
> I would say your problem is that you selected the wrong test to be non-optional in your matrix.
> 
> The latest dev release should be the main (i.e. non-optional) test.
> 
> Not sure why you're testing against v4129, I guess you can remove that from the matrix.

I literally just copied the whole file from here or other repo

PR #3210, opened 2022-01-22
[PHP] Support `readonly` with constructor property promotion

bakerkretzmar:
This PR adds support for using the `readonly` keyword inside constructor-promoted properties.

This is valid PHP code:

```php
class Bar
{
    public function __construct(
        public readonly string $foo,
    ) {}
}
```

But Sublime currently doesn't scope that `readonly` correctly:

image

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -820,7 +820,7 @@ contexts:
                   scope: punctuation.section.group.begin.php
                   set:
                     - meta_scope: meta.function.parameters.php meta.group.php
-                    - match: '{{visibility_modifier}}'
+                    - match: '(?i){{visibility_modifier}}|\b(?:readonly)\b'
                       scope: storage.modifier.php
                     - include: function-parameters
             # Exit on unexpected content
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -820,7 +820,7 @@ contexts:
                   scope: punctuation.section.group.begin.php
                   set:
                     - meta_scope: meta.function.parameters.php meta.group.php
-                    - match: '(?i){{visibility_modifier}}|\b(?:readonly)\b'
+                    - match: '(?i){{visibility_modifier}}|\breadonly\b'
                       scope: storage.modifier.php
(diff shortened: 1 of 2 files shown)


PHP/PHP Source.sublime-syntax
```diff
@@ -820,7 +820,7 @@ contexts:
                   scope: punctuation.section.group.begin.php
                   set:
                     - meta_scope: meta.function.parameters.php meta.group.php
-                    - match: '{{visibility_modifier}}'
+                    - match: '(?i){{visibility_modifier}}|\b(?:readonly)\b'
```
michaelblyons, line 823: I'm curious about the case insensitivity.

bakerkretzmar: I don't mind either way, I went with what was used in [the one other place that `readonly` shows up](https://github.com/sublimehq/Packages/blob/065a2b7b27066908b202e4d0ffcd45acf097a9cc/PHP/PHP%20Source.sublime-syntax#L258-L260) and [the part for typehints](https://github.com/sublimehq/Packages/blob/065a2b7b27066908b202e4d0ffcd45acf097a9cc/PHP/PHP%20Source.sublime-syntax#L400-L405). Feels to me like the same idea as `public`/`protected`/`private` or `int`/`string`/`array`/etc. in function arguments, which as far as I know all technically _can_ be uppercase.

michaelblyons: Okay. If I were doing this, I'd change it to `'{{visibility_modifier}}|\b(?i:readonly)\b'`, but otherwise sounds good.

jfcherng (dismissed):

PHP/PHP Source.sublime-syntax
deathaxe: If all modifiers may be upper-case the proposed solution is correct as `{{visibility_modifier}}` matches case-sensitive only. I'd probably just skip the group, but that's not a big deal.

```suggestion
                    - match: '(?i){{visibility_modifier}}|\breadonly\b'
```

deathaxe (dismissed):

michaelblyons (approved):

deathaxe (approved):

PR #3217, opened 2022-01-27
[C/C++/ObjC/ObjC++] Improve comments punctuation

### LINKED ISSUE #3215: [C family] Alt + Q hard-line-wrap feature doesn't work for comments which begin with 3 slashes `///` (Doxygen comments)
### Description of the bug

Any comments in C or C++ which begin with 3 slashes (`///`) make the <kbd>Alt</kbd> + <kbd>Q</kbd> hard-line-wrap feature not work right.

### Steps to reproduce

Put this into a C or C++ file:
```c
/// This is a really long doxygen comment This is a really long doxygen comment This is a really long doxygen comment This is a really long doxygen comment 
```

Place your cursor on that line, then press <kbd>Alt</kbd> + <kbd>Q</kbd> to hard-line-wrap. It does this:
```c
/// This is a really long doxygen comment This is a really long doxygen comment This is a really
//  long doxygen comment This is a really long doxygen comment 
```

That's not right. The 2nd line has only 2 slashes, not 3, and it has an extra space in it. It should have done this instead:

```c
/// This is a really long doxygen comment This is a really long doxygen comment This is a really
/// long doxygen comment This is a really long doxygen comment 
```

### Expected behavior

see above

### Actual behavior

see above

### Sublime Text build number

4126

Ubuntu 18.04

### (Linux) Desktop environment and/or window manager

Gnome

### Additional information

My clunky work-around, which is time-consuming, is to use the multi-line cursor mode to remove 1 slash per line, making each line begin with `//`. I then hit <kbd>Alt</kbd> + <kbd>Q</kbd> to hard-line-wrap, then I use the multi-line cursor mode to add 1 slash per line back to the beginning of each line, making them all begin with `///` again.

### OpenGL context information

_No response_

deathaxe:
Fixes #3215

This PR improves comment punctuation highlighting to support arbitrary number of opening and closing punctuation chars.

This helps to keep line comment punctuation consistent when performing hard line wrapping.

The new contexts are copied from JavaScript (see #3018) and adjusted to match C's scopes and maintain the existing "banner comments".

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -278,21 +278,7 @@ contexts:
     - include: keywords-angle-brackets
     - match: '(?={{path_lookahead}}\s*<)'
       push: global-modifier
-    # Take care of comments just before a function definition.
-    - match: /\*
-      scope: punctuation.definition.comment.c
-      push:
-        - - match: \s*(?=\w)
-            set: global-modifier
-          - match: ""
-            pop: true
-        - - meta_scope: comment.block.c
-          - match: \*/
-            scope: punctuation.definition.comment.c
-            pop: true
-          - match: ^\s*(\*)(?!/)
-            captures:
-              1: punctuation.definition.comment.c
+    - include: global-block-comments
     - include: early-expressions
     - match: ^\s*\b(extern)(?=\s+"C(\+\+)?")
       scope: storage.modifier.c++
@@ -323,6 +309,38 @@ contexts:
     - match: ^\s*(?=\w)
       push: global-modifier
     - include: late-expressions
+
+  global-block-comments:
+    # NOTE: This context overrides `comments` context, which re-uses scopes
+    # from `source.c#comments`. So keep the trailing `c` for consistency reasons!
+    - match: ^(/\*) =(\s*.*?)\s*= (\*/)$\n?
+      scope: comment.block.banner.c
+      captures:
+        1: punctuation.definition.comment.begin.c
+        2: meta.toc-list.banner.block.c
+        3: punctuation.definition.comment.end.c
+    # empty block comments
+    - match: /\*\*+/
+      scope: comment.block.empty.c punctuation.definition.comment.c
+      push: global-block-comment-after
+    # documentation block comments
+    - match: (?:/\*!|/\*{2,})
+      scope: punctuation.definition.comment.begin.c
+      push:
+        - global-block-comment-after
+        - scope:source.c#block-comment-documentation-body
+    # normal block comments
+    - match: /\*
+      scope: punctuation.definition.comment.begin.c
+      push:
+        - global-block-comment-after
+        - scope:source.c#block-comment-body
+
+  global-block-comment-after:
+    - match: \s*(?=\w)
+      set: global-modifier
+    - match: ''
+      pop: true
 
   statements:
     - include: preprocessor-statements
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -317,7 +317,7 @@ contexts:
   global-block-comments:
     # NOTE: This context overrides `comments` context, which re-uses scopes
     # from `source.c#comments`. So keep the trailing `c` for consistency reasons!
-    - match: ^(/\*) =(\s*.*?)\s*= (\*/)$\n?
+    - match: ^(/\*) =\s*(.*?)\s*= (\*/)$\n?
       scope: comment.block.banner.c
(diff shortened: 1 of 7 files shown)


jwortmann:
What I had in mind for the `///` and `//!` Doxygen comment scope was rather `comment.line.documentation`, which is already used in C# and R. I would assume that such kind of comments are always meant to be used for Doxygen, so it might be useful for later, in case there will be added more highlighting for Doxygen comments at some point in the future. Doxygen isn't really part of the C or C++ language, but I think the same is true for JSDoc, PHPDoc, etc.

Optionally, Doxygen block comments could use `comment.block.documentation` then, if support for those should be added as well:
```cpp
/**
 * ... text ...
 */

/*!
 * ... text ...
 */
```

(from https://www.doxygen.nl/manual/docblocks.html#cppblock).

deathaxe:
Should we use that in JS, too, then? I've just copied those contexts including scopes from there.

Those `double-slash` etc. sub scopes don't make much sense then any longer.

jwortmann:
I think JSDoc, JavaDoc, ... only work in block comments, so they should already have the correct scopes.

> Those double-slash etc. sub scopes don't make much sense then any longer.

That is probably true and I don't see a reason why color schemes would want to target these special scopes and highlight for example `triple-slash` differently. Actually, the `documentation` part in the scope is an orthogonal property to line/block and the punctuation style, so it would make more sense to always have it decoupled from the `comment` scope, for example like `meta.documentation commen.line.triple-slash`. But I don't think that would be feasible due to backwards compatibility with color schemes, so I'd suggest to keep it at the third level of the comment scope. In case it is a line doc-comment, this means to drop the `triple-slash`/`other`, but I think documentation comments which are also line comments are quite rare anyway.

deathaxe:
> ... I don't see a reason why color schemes would want to target these special scopes ...

I can barely remember such a discussion though, which this distinction was argued as important.

michaelblyons:
There's #2629, and maybe another place. I feel like it was a Mitranim conversation.

jwortmann:
Since you have also improved the "banner" comments, what do think about adjusting the scope from `meta.toc-list.banner` to `entity.name.section`, which you already used in R? It doesn't necessarily needs to be part of this PR though.

`entity.name.section` has the advantage that it gets a "navigation" icon in the Goto Symbol panel, and we could make the scopes consistent between syntaxes this way. I just created a similar PR for Matlab #3219.
The `Symbol List - Prefix Banner Items.tmPreferences` file would need to be adjusted as well then. 

---

I think I see now what you meant with using it in JS too, with the current `triple-slash` in TypeScript for
```typescript
/// <reference no-default-lib="true"/>
```
But from what I found in the TypeScript docs, that seems to be only a compiler directive, not a documentation comment, so I would say that the current scoping is still okay there. Not sure if it would make sense to alternatively introduce another sub-scope for it instead, like `comment.line.directive`?

C++/C++.sublime-syntax
```diff
@@ -328,6 +314,38 @@ contexts:
       push: global-modifier
     - include: late-expressions
 
+  global-block-comments:
+    # NOTE: This context overrides `comments` context, which re-uses scopes
+    # from `source.c#comments`. So keep the trailing `c` for consistency reasons!
+    - match: ^(/\*) =(\s*.*?)\s*= (\*/)$\n?
```
jwortmann, line 320: I know this pattern was just copied, but this `.*?` combination looks a bit suspicious or even bad to me. Am I missing something, or isn't the `?` needed here?

deathaxe: It is required to give the trailing `\s*` precedence so `meta.toc-list` only includes the text stripped by whitespace. But the capture group includes the leading whitespace, which seems wrong.

michaelblyons (approved):

ElectricRCAircraftGuy:
I'd love to see this fix for hard-line-wrap on doxygen comments get merged. I've been fighting with it for years. 

In case it helps expedite it: I am a paying user. I have a Single User Licence registered to Gabriel Staples, and Sublime Text is my primary editor.

ElectricRCAircraftGuy:
I'm still hoping to see this feature added to the Alt + Q hardwrap in sublime text. Meanwhile, I'm moving to VSCode and the Rewrap extension to do this for me: 

image

FichteFoll (approved):

deathaxe:
see also: https://github.com/sublimehq/Packages/pull/3855#issuecomment-1771405398

PR #3219, opened 2022-01-29
[Matlab] Add code section comments to symbol list

jwortmann:
Reference from the Matlab docs: https://www.mathworks.com/help/matlab/matlab_prog/create-and-run-sections.html

The scope `entity.name.section` is added to the section title, which is already used in R and gets a "navigation" kind icon in the Goto Symbol panel. Also the pattern was adjusted so that section comments should start on their own line.

### DIFF
--- a/Matlab/Matlab.sublime-syntax
+++ b/Matlab/Matlab.sublime-syntax
@@ -467,10 +467,11 @@ contexts:
     - include: block-comments
 
   line-comments:
-    - match: (%%(?!%)).*\n?
+    - match: ^\s*(%%(?!%))\s*(.*\S)?\s*\n?
       scope: comment.line.double-percentage.matlab
       captures:
         1: punctuation.definition.comment.matlab
+        2: entity.name.section.matlab
     - match: (%+).*\n?
       scope: comment.line.percentage.matlab
       captures:
--- a/Matlab/Symbol List.tmPreferences
+++ b/Matlab/Symbol List.tmPreferences
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+    <key>scope</key>
+    <string>source.matlab entity.name.section</string>
+    <key>settings</key>
+    <dict>
+        <key>showInSymbolList</key>
+        <integer>1</integer>
+    </dict>
+</dict>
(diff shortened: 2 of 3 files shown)


Matlab/Matlab.sublime-syntax
```diff
@@ -467,10 +467,11 @@ contexts:
     - include: block-comments
 
   line-comments:
-    - match: (%%(?!%)).*\n?
+    - match: ^\s*(%%(?!%))\s*(.*\S)?\s*\n?
```
jfcherng, line 470: is `^` necessary?

jwortmann: According to the link above
> You also can enter two percent signs (%%) at the start of the line where you want to begin the new section.

(but they also work after whitespace nested within e.g. loops to create section breaks)

jrappen:
Based on your changes to the test file and that passing (without checking further), you probably:

- now match the second rule in `line-comments` with the test
- which doesn't have a capture for `entity.*`
- which makes the negative test redundant
- which makes me think you messed this up

Maybe double-check `comment.line.percentage` vs. `comment.line.double-percentage`?

jwortmann:
The negative test for `entity` is redundant, yes, but it doesn't hurt and it makes it clear that we don't want to match it as a code section comment if there is anything other than whitespace on the same line before the comment (unlike the new highlighting here on GitHub).

deathaxe (approved):

jfcherng (approved):

deathaxe:
> Maybe double-check `comment.line.percentage` vs. `comment.line.double-percentage`?

No matter how many leading percentages a comment has only section comments are of special meaning. That's exactly the strategy used in #3217 to limit the amount of different comment scopes.

jwortmann:
I noticed that the adjusted pattern for section comments accidentally resulted in applying the `comment` scope to preceding whitespace as well. This was not intended, but on the other hand it makes it possible to look a bit like in the Matlab IDE, by using an adjusted background color for that scope in the color scheme:

matlab

matlab2

Should it be "fixed", or leave it as it is? (The fix would be to add a capture group and move `comment.line.double-percentage.matlab` from "scope" to "captures".)

deathaxe:
I think it's ok this way.

PR #3220, opened 2022-01-30
[C/C++/ObjC/ObjC++] Fix loop statement indentation

deathaxe:
This commit restricts `bracketIndentNextLinePattern` to not match loops
if the loop condition expression spans multiple lines as it causes
breaks indentation of the following code block if the block's opening
brace directly follows the closing parenthesis on the same line.

The result was only the first line of the block being indented.

Fixes https://forum.sublimetext.com/t/the-indenting-of-a-c-for-loop-is-incorrect-under-certain-conditions/52634/4

### DIFF
--- a/C++/Indentation Rules.tmPreferences
+++ b/C++/Indentation Rules.tmPreferences
@@ -60,10 +60,10 @@
 			  # followed by whitespace or block comments [optional]
 			  \g<comment_or_whitespace>
 			  # top-level balanced parentheses
-			  \(
-				(?<group_body> (?:
+			  (?<group>
+			  	\( (?:
 				# nested balanced parentheses
-				  \( \g<group_body> \)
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

keith-hall (approved):

michaelblyons:
When the closing parenthesis is on a new line, I feel like I would do this:

```js
if (
    true
)
{
    foo();
}

if (
    true
) {
    foo();
}
```

instead of

```js
if (
    true
    )
{
    foo();
}

if (
    true
    ) {
    foo();
}

deathaxe:
It's ST's default `indentParens` which applies here. With only one line being taken into account by the indentation engine, I am not sure whether we can find patterns which satisfy all kinds of coding styles.

PR #3230, opened 2022-02-05
[Bash] Color escapes in strings

michaelblyons:
The color escapes do not technically work in all string contexts, but I think the false-positive percentage is probably low.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -1958,6 +1958,8 @@ contexts:
       scope: constant.character.escape.shell
 
   string-escapes:
+    - match: '\\(?:e|033)\[\d{1,3}(?:;\d{1,3}){0,2}m'
+      scope: constant.character.escape.color.shell
     - match: \\[$`"\\]
       scope: constant.character.escape.shell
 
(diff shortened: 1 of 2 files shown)


ShellScript/Bash.sublime-syntax
```diff
@@ -1958,6 +1958,8 @@ contexts:
       scope: constant.character.escape.shell
 
   string-escapes:
+    - match: '\\(?:e|033)\[\d{1,3}(?:;\d{1,3}){0,2}m'
```
jfcherng, line 1961: The color sequence can go up to 24-bit `ESC[ 38;2;⟨r⟩;⟨g⟩;⟨b⟩ m`.
https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit

michaelblyons: 🤷‍♂️ If you prefer it, you can make a suggestion. If you really want, we can even [expressly make 0-255](https://regex101.com/r/BAjN4w/1).

jfcherng: No. I mean your regex didn't highlight something like `\033[38;2;0;255;0m`

jfcherng:
```suggestion
    - match: '\\(?:e|033)\[\d{1,3}(?:;\d{1,3}){0,4}m'
```

deathaxe (approved):

jfcherng (approved):

PR #3234, opened 2022-02-10
[Haskell] Add shebang highlighting

deathaxe:
The keywords `cabal` and `stack` are borrowed from
- https://github.com/haskell/haskell-language-server/blob/4b8f9d53bed3e775ccd120d9a2471ae374a88db3/install.hs#L1
- https://github.com/haskell/haskell-language-server/blob/4b8f9d53bed3e775ccd120d9a2471ae374a88db3/GenChangelogs.hs#L1

### DIFF
--- a/Haskell/Haskell.sublime-syntax
+++ b/Haskell/Haskell.sublime-syntax
@@ -14,7 +14,7 @@ file_extensions:
 
 first_line_match: |-
   (?xi:
-    ^ \#! .* \brunhaskell\b                     # shebang
+    ^ \#! .* \b(cabal|runhaskell|stack)\b       # shebang
   | ^ \s* -- .*? -\*- .*? \bhaskell\b .*? -\*-  # editorconfig
   )
 
@@ -219,6 +219,11 @@ contexts:
     - include: preprocessor-directives
 
   main:
+    - meta_include_prototype: false
+    - match: ''
+      push: [haskell, shebang]
+
+  haskell:
     - include: modules
     - include: declarations
     - include: statements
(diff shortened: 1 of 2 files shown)


jrappen (approved):

michaelblyons (approved):

Haskell/Haskell.sublime-syntax
```diff
@@ -290,6 +295,21 @@ contexts:
...
+  shebang:
+    - match: ^\#!
+      scope: punctuation.definition.comment.haskell
+      set: shebang-body
+    - match: ^|(?=\S)  # Note: Ensure to highlight shebang if Haskell is embedded.
+      pop: 1
+
+  shebang-body:
+    - meta_scope: comment.line.shebang.haskell
+    # Note: Keep sync with first_line_match!
```
FichteFoll, line 307: Probably not unique to this syntax, but I wonder whether we should make this match more general to allow for any executable instead of only highlighting the expected values and what is used for syntax detection? It could be a wrapper script, for example. A path that ends with `env` would then be exempted.

FichteFoll (approved):

Haskell/Haskell.sublime-syntax
deathaxe: I think a generic but still reliable approach which can be applied to all syntaxes would be welcome.

jwortmann: > What do you think about a syntax or context in the Bash package that all other syntaxes could embed for their first line shebang match?

Does that still work if someone adds Bash to "ignored_packages"?

deathaxe: It certainly won't. I also don't think it is required to `embed` such a tiny context. And we should probably try to keep dependencies between packages low. Inheritance comes with a certain level of complexity so a use case should be worth it.

FichteFoll: Sure it's not "required" but it does beg the question of how to deduplicate such a frequently occurring syntax element that would ideally be scoped consistently. We could of course also include a context, which would not produce errors, as far as I remember, though I'm not exactly sure what happens when you embed something that doesn't exist.

deathaxe: An error about missing syntax is printed to console and the region up to the `escape` is unscoped.

PR #3235, opened 2022-02-12
[Lua] Add support for EmmyLua and LDoc doc comments

### LINKED ISSUE #3211: [Lua] The Lua syntax still not support comment document
## haven't comment document highlight when i write code in lua file
[EmmyLua Annotations](https://github.com/sumneko/lua-language-server/wiki/EmmyLua-Annotations)

## Sublime
image

## VSCode
image

jwortmann:
Resolves #3211.

This PR adds basic highlighting for [EmmyLua](https://github.com/sumneko/lua-language-server/wiki/EmmyLua-Annotations) and [LDoc](https://stevedonovan.github.io/ldoc/manual/doc.md.html) documentation comments. They are not really part of the Lua language, so I don't know whether it is desired to be added (having the open PR for JSDoc types in mind, which seems to be stuck for some reason), but I think it would be useful to provide a PR. Also there are at least three competing standards for the scope name of `@tag` tags already used in the default syntaxes...

It is kept pretty basic with only `@tags` being highlighted, with two exceptions for `@section` and `@param`.

Ideally this should be tested by a regular Lua user, @jingyuexing maybe do you want to try it out?

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -102,7 +102,7 @@ contexts:
       captures:
         1: entity.name.tag.documentation.lua
         2: entity.name.section.lua
-    - match: ^\s*(@param)(?:\[[\w=']*\])?\s+({{identifier}})\b
+    - match: ^\s*(@param)(?:\[[\w=']*\])?\s+({{identifier}})
       captures:
         1: entity.name.tag.documentation.lua
         2: variable.parameter.lua
@@ -122,7 +122,7 @@ contexts:
 
   line-comments:
     - match: -{3,}
-      scope: punctuation.definition.comment.lua
+      scope: comment.line.documentation.lua punctuation.definition.comment.lua
       push: line-doc-comment-body
     - match: --
       scope: punctuation.definition.comment.lua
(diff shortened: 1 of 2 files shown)


Lua/Lua.sublime-syntax
```diff
@@ -75,10 +85,34 @@ contexts:
...
+    - meta_include_prototype: false
+    - meta_scope: comment.block.documentation.lua
+    - match: \]\]
+      scope: punctuation.definition.comment.end.lua
+      pop: 1
+    - match: ^\s*(@section)\s+(.*?\S)\s*(?=\]\]|$)
+      captures:
+        1: entity.name.tag.documentation.lua
+        2: entity.name.section.lua
+    - match: ^\s*(@param)\s+({{identifier}})\b
```
jingyuexing: Maybe add a type highlight?

jwortmann: The difficulty is that the tags are not really standardized. There are also [LDoc](https://stevedonovan.github.io/ldoc/manual/doc.md.html#Commenting_Conventions) and LuaDoc, which use a slightly different convention. If we add type highlighting for the next word after the parameter name, then for LDoc it would mistakenly highlight the first word of the description as a type. By keeping it simple, we can reduce the chance of false highlighting for all users. Would you say that EmmyLua is the de-facto standard?

From LDoc:

LDoc

jingyuexing: Emmmm,not thought about that :cold_sweat:

jingyuexing: So far everything is fine :smiley:

jrappen (approved):

jwortmann:
I just pushed a small stylistic change, but I think this PR should be ready now (unless anyone finds possible things to improve).

jrappen (approved):

deathaxe (approved):

jingyuexing (approved):

jfcherng (approved):

PR #3237, opened 2022-02-14
[PHP] Suppress method completions when typing php tags

deathaxe:
Resolves https://github.com/sublimehq/sublime_text/issues/5256

The fix consists of two parts:

1. Extending the completion's selector by 

       - punctuation.section.embedded.begin

2. Making sure to highlight incomplete <?php tags

       punctuation.section.embedded.begin
    
   Note: If this isn't acceptable we might want to scope incomplete
   tags using a `meta` scope and add this to the selector (1.) as well.

   We just need to ensure ST sees the correct scope while typing the
   opening php tag in order to not add those methods to the list of
   available completions.

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -120,7 +120,7 @@ contexts:
       push:
         - meta_scope: embedding.php text.html.basic
         - clear_scopes: true
-        - match: <\?(?i:php)?
+        - match: <\?(?i:p(?:hp?)?)?
           scope: meta.embedded.block.php punctuation.section.embedded.begin.php
           pop: 1
         - match: ''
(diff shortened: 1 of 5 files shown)


jfcherng (dismissed):
This is a clever workaround!

deathaxe:
It could be. How likely are false positives due to incomplete tags?

michaelblyons:
> How likely are false positives due to incomplete tags?

I don't know. I don't write PHP. 😀 But I have [written something similar](https://github.com/michaelblyons/SublimeSyntax-ssh_config/blob/master/SSH%20Config.sublime-syntax#L197-L204) to not-highlight boolean values until you have finished typing them. (In context, anything *not* boolean [is marked invalid](https://github.com/michaelblyons/SublimeSyntax-ssh_config/blob/master/SSH%20Config.sublime-syntax#L232-L234).)

I just imagine someone writes `<?ph` and is confused when the code doesn't run. 🤷‍♂️ I'm not against what you have here. I just think it'd be nice.

bmuessig:
Awesome, great to see this merged now. Thanks everyone!

PR #3238, opened 2022-02-16
[meta] Update workflow

jrappen:
- update latest dev build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,10 +47,6 @@ jobs:
             sublime-build: 4127
             optional: false
 
-          # first build with the latest changes to the syntax engine in core
-          - sublime-channel: dev
-            sublime-build: 4095
-            optional: true
     steps:
 
       - name: Git checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -36,12 +36,21 @@ jobs:
...
+
+          # latest dev build
           - sublime-channel: dev
-            sublime-build: 4125
+            sublime-build: 4127
             optional: false
+
+          # first build with the latest changes to the syntax engine in core
+          - sublime-channel: dev
+            sublime-build: 4095
```
keith-hall, line 52: What is the motivation for this? I personally don't think we need to spend time trying to keep compatibility with older builds, even seeing if it is compatible isn't interesting for me :)

jrappen: Ok

jrappen: I thought it might make sense to check against the documented first build with the latest changes to the syntax engine in core.

So if SublimeHQ adds any changes to `master` with undocumented changes to the syntax engine that would break that test.

But maybe this is the wrong approach. I removed that part for now.

keith-hall (approved):

PR #3242, opened 2022-02-20
[Go] Fix compiler directive syntax patterns

kortschak:
This adds line and export directive syntax patterns and generalises the `//go:name` pattern for use in other contexts such as linter directives. It also fixes `//go:name` matching to avoid matching invalid directive comments.

~~It's not completely clear to me how to express the expectation in the tests (and some current tests expectation are incorrect; `//go` and `//go:` are invalid directives — so I expect tests to fail), so please let me know what I should add.~~

These should be marked as valid directives:
```
//go:linkname
//go:cgo_import_dynamic
//lint:ignore U1000 This is here for user documentation.
//lint:ignore-file Autogenerated.
//line file.rl:3732:45
//line :3732:45:45
//line :3732 :45:45
/*line file.rl:3732:45*/
/*line :3732:45:45*/
/*line :3732 :45:45*/
//export myfunc
//extern myfunc
```

These should not be marked:
```
/*line file.rl:3732:45 */
/*line :3732:45:45 */
/*line :3732 :45:45 */
//go:
//go
```

Please take a look.

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -82,9 +82,6 @@ variables:
   # Directive comment key.
   directive: \b[[:alpha:]_][[:alnum:]_]*\b
 
-  # Line directive comment key.
-  line_directive: (line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})
-
   # Cgo export and gccgo extern directive.
   export_directive: (export|extern) ({{ident}})
 
@@ -157,7 +154,7 @@ contexts:
     # position for the character immediately following the comment as
     # having come from the specified file, line and column. Both single
     # line and block comment line directives are valid.
-    - match: (//){{line_directive}}$
+    - match: (//)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})$
       captures:
         1: punctuation.definition.comment.go
         2: meta.keyword.annotation.go
@@ -171,7 +168,7 @@ contexts:
         - match: $
           set: pop-line-comment
 
-    - match: (/\*){{line_directive}}(\*/)
+    - match: (/\*)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})(\*/)
       captures:
         1: punctuation.definition.comment.go
         2: meta.keyword.annotation.go
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -80,7 +80,7 @@ variables:
   ident: \b(?!{{keyword}})[[:alpha:]_][[:alnum:]_]*\b
 
   # Directive comment key.
-  directive: \b[[:alpha:]_][[:alnum:]_]*\b
+  directive: \b[[:alpha:]_-][[:alnum:]_-]*\b
 
   # Single line only
   inline_comment: /[*](?:[^*]|[*](?!/))*[*]/
@@ -117,7 +117,7 @@ contexts:
 
   # https://golang.org/ref/spec#Comments
   match-comments:
-    # Go has magic comments in the form of:
+   # Go has magic comments in the form of:
     #
     # //go:some_directive arg0 arg1 ...
     #
@@ -132,7 +132,7 @@ contexts:
     # highlighting them as compiler pragmas could be misleading. We scope them
     # as plain comments by default, and add some detailed meta scopes for
     # enterprising users wishing to add more color.
-    - match: (//)([a-z]+)(:)({{directive}}.*)
+    - match: (//)([a-z]+)(:)({{directive}})[ ]?
       captures:
         1: punctuation.definition.comment.go
         2: meta.keyword.annotation.go
@@ -140,7 +140,7 @@ contexts:
         4: meta.variable.function.go
       push:
         - meta_scope: meta.annotation.go comment.line.go
-        - match: \S+
+        - match: .+
           scope: meta.variable.parameter.go
         # End the annotation scope at EOL, but stretch the comment scope
         # indefinitely to the right.
@@ -154,22 +154,18 @@ contexts:
     - match: (//)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})$
       captures:
         1: punctuation.definition.comment.go
-        2: meta.keyword.annotation.go
-        3: meta.variable.function.go
+        2: meta.variable.function.go
+        3: meta.variable.parameter.go
       push:
         - meta_scope: meta.annotation.go comment.line.go
-        - match: \S+
-          scope: meta.variable.parameter.go
-        # End the annotation scope at EOL, but stretch the comment scope
-        # indefinitely to the right.
         - match: $
           set: pop-line-comment
 
     - match: (/\*)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})(\*/)
       captures:
         1: punctuation.definition.comment.go
-        2: meta.keyword.annotation.go
-        3: meta.variable.function.go
+        2: meta.variable.function.go
+        3: meta.variable.parameter.go
         4: punctuation.definition.comment.go
       push:
         - meta_scope: meta.annotation.go comment.block.go
@@ -184,14 +180,10 @@ contexts:
     - match: (//)(export|extern) ({{ident}})$
       captures:
         1: punctuation.definition.comment.go
-        2: meta.keyword.annotation.go
-        3: meta.variable.function.go
+        2: meta.variable.function.go
+        3: meta.variable.parameter.go
       push:
         - meta_scope: meta.annotation.go comment.line.go
-        - match: \S+
-          scope: meta.variable.parameter.go
-        # End the annotation scope at EOL, but stretch the comment scope
-        # indefinitely to the right.
         - match: $
           set: pop-line-comment
 
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -132,7 +132,7 @@ contexts:
     # highlighting them as compiler pragmas could be misleading. We scope them
     # as plain comments by default, and add some detailed meta scopes for
     # enterprising users wishing to add more color.
-    - match: (//)([a-z]+)(:)({{directive}})[ ]?
+    - match: (//)([a-z]+[a-z-]*)(:)({{directive}})[ ]?
       captures:
         1: punctuation.definition.comment.go
         2: meta.keyword.annotation.go
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -133,52 +133,43 @@ contexts:
     # as plain comments by default, and add some detailed meta scopes for
     # enterprising users wishing to add more color.
     - match: (//)([a-z]+[a-z-]*)(:)({{directive}})[ ]?
+      scope: meta.annotation.go
       captures:
         1: punctuation.definition.comment.go
         2: meta.keyword.annotation.go
-        3: meta.punctuation.accessor.colon.go
+        3: punctuation.accessor.colon.go
         4: meta.variable.function.go
-      push:
-        - meta_scope: meta.annotation.go comment.line.go
-        - match: .+
-          scope: meta.annotation.parameters.go
-        # End the annotation scope at EOL, but stretch the comment scope
-        # indefinitely to the right.
-        - match: $
-          set: pop-line-comment
+      push: pop-line-annotation
 
     # Go also has line renumbering; line directives specify the source
     # position for the character immediately following the comment as
     # having come from the specified file, line and column. Both single
     # line and block comment line directives are valid.
-    - match: (//)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})$
+    - match: ((//)(line) )([a-zA-Z0-9. :]*(?::[0-9]+){1,2})$\n?
+      scope: comment.line.go
       captures:
-        1: punctuation.definition.comment.go
-        2: meta.variable.function.go
-        3: meta.variable.parameter.go
-      push:
-        - meta_scope: meta.annotation.go comment.line.go
-        - match: $
-          set: pop-line-comment
+        1: meta.annotation.go
+        2: punctuation.definition.comment.go
+        3: meta.variable.function.go
+        4: meta.annotation.parameters.go
 
-    - match: (/\*)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})(\*/)
-      scope: meta.annotation.go comment.block.go
+    - match: ((/\*)(line) )([a-zA-Z0-9. :]*(?::[0-9]+){1,2})(\*/)
+      scope: comment.block.go
       captures:
-        1: punctuation.definition.comment.begin.go
-        2: meta.variable.function.go
-        3: meta.variable.parameter.go
-        4: punctuation.definition.comment.end.go
+        1: meta.annotation.go
+        2: punctuation.definition.comment.begin.go
+        3: meta.variable.function.go
+        4: meta.annotation.parameters.go
+        5: meta.annotation.go punctuation.definition.comment.end.go
 
     # Finally, Cgo and gccgo have directives for exporting functions to C.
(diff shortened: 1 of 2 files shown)


jrappen:
In general, capturing regex groups should not be part of variables in the syntax. This is mainly for readability reasons.

Instead use the variables within a capturing group of the match in the contexts.

jrappen:
Additionally, you should probably add "negative tests" for the examples listed above, where markings should not be applied.

kortschak:
>Additionally, you should probably add "negative tests" for the examples listed above, where markings should not be applied.

These are included.

jrappen (approved):

jrappen:
This seems fine. I don't use Go, though.

Go/Go.sublime-syntax
```diff
@@ -114,30 +117,76 @@ contexts:
 
   # https://golang.org/ref/spec#Comments
   match-comments:
-    # Go has magic comments in the form of:
+   # Go has magic comments in the form of:
```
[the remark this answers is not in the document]
michaelblyons: Everything else seems fine to me.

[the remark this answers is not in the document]
jrappen: Nice catch, this should be fixed before merge. Workflow should've caught this, but it didn't auto-run because this is the first contribution to the repo.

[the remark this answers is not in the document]
kortschak: That was editor fumbling. Will fix.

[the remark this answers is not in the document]
kortschak: Done

michaelblyons (approved):

kortschak:
Looking at https://github.com/golang/go/issues/43776, and noting the existence of [`//go-sumtype:` as a namespace](https://github.com/BurntSushi/go-sumtype), I think it's worth generalising the pattern.

Go/Go.sublime-syntax
```diff
@@ -132,7 +132,7 @@ contexts:
     # highlighting them as compiler pragmas could be misleading. We scope them
     # as plain comments by default, and add some detailed meta scopes for
     # enterprising users wishing to add more color.
-    - match: (//)([a-z]+)(:)({{directive}})[ ]?
+    - match: (//)([a-z]+[a-z-]*)(:)({{directive}})[ ]?
```
deathaxe, line 135: Are underscores supported? They are at least part of `{{ident}}`.

kortschak: There is no formal documentation for them, but there are no underscore-containing namespace cases in the wild. There are underscore-containing directives (the compiler recognises a bunch of `//go:cgo_[^ ]+`.

Given that this covers what exists, it's probably OK to leave it until golang/go#43776 is resolved.

deathaxe: OK, just wanted to be sure we don't miss something.

Go/Go.sublime-syntax
```diff
@@ -118,26 +121,72 @@ contexts:
...
       captures:
         1: punctuation.definition.comment.go
         2: meta.keyword.annotation.go
         3: meta.punctuation.accessor.colon.go
         4: meta.variable.function.go
       push:
         - meta_scope: meta.annotation.go comment.line.go
+        - match: .+
+          scope: meta.variable.parameter.go
```
deathaxe, line 144: How about ...

```suggestion
        - meta_content_scope: meta.variable.parameter.go        
```

Otherwise should it use the same pattern as in line 172?

```suggestion
        - match: \S+
          scope: meta.variable.parameter.go
```

deathaxe, line 144: How about ...

```suggestion
        - meta_content_scope: meta.variable.parameter.go        
```

Otherwise should it use the same pattern as in line 172?

```suggestion
        - match: \S+
          scope: meta.variable.parameter.go
```

kortschak: `\S+` is incorrect; the text following the ns and directive is not restricted and can include spaces. An example of this is the `//lint:ignore U1000 This is here for user documentation.` example — all of "U1000 This is here for user documentation." is the extra information. Different tools will interpret this differently, for example `staticcheck` will interpret this is "U1000" and "This is here for user documentation." while `go:generate go run ./builder.go` will interpret the remaineder as "go", "run" and "./builder.go" in an argv. So I don't think it's up to the syntax to determine that.

deathaxe: If any kind of text may follow the directive scoping it `meta.variable.parameter` seems unsuitable at all and should be removed.

Go/Go.sublime-syntax
```diff
@@ -118,26 +121,72 @@ contexts:
...
+        3: meta.variable.parameter.go
+        4: punctuation.definition.comment.go
+      push:
+        - meta_scope: meta.annotation.go comment.block.go
         - match: \S+
           scope: meta.variable.parameter.go
         # End the annotation scope at EOL, but stretch the comment scope
         # indefinitely to the right.
         - match: $
           set: pop-line-comment
```
deathaxe, line 177: Pushing into a context after a closed block quote doesn't seem to make sense.

`/*line foo.bar::9 */ ???annotation here???` 

I don't think arbritary text after a block quote is parsed as part of a comment or annotation, is it?

Go/Go.sublime-syntax
kortschak: It should be tagged though, it is part of the directive comment. I don't really care how it's tagged, but semantically this seems the most appropriate. Directive comments are poorly specified and used by different tools with different behaviours. This captures that.

Go/Go.sublime-syntax
kortschak: Yes, this didn't feel right at the time, but I was unsure of how to achieve what I wanted.

kortschak: Hopefully this most recent change fixes this. If not, please make a suggestion.

Go/Go.sublime-syntax
deathaxe: _You are absolutely right about specification. So please just read my comments as suggestions to get the best out of your welcome efforts to help improving ST's syntaxes._ 

`meta.variable` is used to scope variable expansions or interpolations in various syntaxes and thus doesn't really fit here. I guess the original intent here was to use it for arguments _(see previous comments like `directive arg1 arg2`)_ as subtile alternative to `variable.other` which would have been highlighted by most color schemes. Scoping the whole line `meta.variable` feels off, if it doesn't really contain variables.

The whole line is already scoped `meta.annotation` and we've already used `meta.annotation.parameters` in other syntaxes. 

So would the following probably be sufficient?

```yaml
    - match: (//)([a-z]+[a-z-]*)(:)({{directive}})[ ]?
      captures:
        0: meta.annotation.go
        1: punctuation.definition.comment.go
        2: meta.keyword.annotation.go
        3: meta.punctuation.accessor.colon.go
        4: meta.variable.function.go
      push:
        - meta_scope: comment.line.go
        - meta_content_scope: meta.annotation.parameters.go
        - match: \n
          pop: true
```

As we started to scope special keywords in comments for JavaDoc, LuaDoc or C#, it would probably even be suitable to replace those `meta.` scopes of capture group 2-3 by real `keyword`, `punctuation`, ... scopes. The `variable.annotation` scope is specified and used in several syntaxes (Python's `@annotation.function()` or C#). Not sure if those would be suitable here, too.

This may be solved in another PR though.

Go/Go.sublime-syntax
```diff
@@ -118,26 +121,68 @@ contexts:
...
+    - match: (/\*)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})(\*/)
+      captures:
+        1: punctuation.definition.comment.begin.go
+        2: meta.variable.function.go
+        3: meta.variable.parameter.go
+        4: punctuation.definition.comment.end.go
+      push:
+        - meta_scope: meta.annotation.go comment.block.go
+        - match: $
+          pop: true
```
deathaxe, line 173: If the whole entity of interest is already mathed by a single pattern, you don't need to push into a separate context afterwards.

`meta.annotation` can be assigned directly via `scope: ` keyword then.

```suggestion
    - match: (/\*)(line) ([a-zA-Z0-9. :]*(?::[0-9]+){1,2})(\*/)
      scope: meta.annotation.go comment.block.go
      captures:
        1: punctuation.definition.comment.begin.go
        2: meta.variable.function.go
        3: meta.variable.parameter.go
        4: punctuation.definition.comment.end.go
```

Go/Go.sublime-syntax
kortschak: Yeah, that probably would work. Thanks for the detailed comment, it is very helpful.

kortschak: I just tried that and I don't think it's right, this marks the whole comment as `meta.annotation.parameters.go`, which does not match what I would expect here. Alternatively (and entirely possibly), I don't understand the (honestly not very friendly — no docs I could find and screeds of uncondensed output) test output. I think it would be preferable to get this merged and fix that up later.

I have made the change to identity though.

deathaxe:
I'd vote to avoid stacking `meta.annotation` and `meta.annotation.parameters`. Added a proposal to the branch accordingly. Feel free to further tweak it.

Go/syntax_test_go.go
```diff
@@ -66,84 +66,87 @@ You may have to disable Go-specific linters when working on this file.
     //go:generate one two three
 // ^ -comment -meta -punctuation
 //  ^^ punctuation.definition.comment.go
-//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation.go comment.line.go
+//  ^^^^^^^^^^^^^^ comment.line.go meta.annotation.go
```
kortschak, line 69: @deathaxe I'd like to understand why `meta.annotation.go` should not reach to the end of the line and if it doesn't why it does reach into the space that separates it from the `meta.annotation.parameters.go`.

deathaxe: It doesn't need to not do so, but this was the state of the art before my commit. So I just maintained it. Same applies to the space in fron to  `meta.annotation.parameters`.

I'd be very fine with both being changed.

One aspect to considder though is annotations being highlighted using `background` by some color schemes.

grafik

By scoping `\n` `meta.annotation` those background colors would span up to the very right edge of a view. Is it what we want?

kortschak: Sorry, but the image doesn't help me understand that answer.

I was asking two things.

1. Why not make to the end of the text like so:
    ```
        //lint:ignore U1000 Reason.
    //  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ comment.line.go meta.annotation.go
    ```
    This comes down to what the semantics of a `meta.annotation.go` are and something that I may not be understanding. But to my mind, the entire directive comment is a meta annotation.

2. If they do not make it to the end of the line as above, why cover the space after the directive name like so
    ```
        //lint:ignore U1000 Reason.
    //  ^^^^^^^^^^^^^^ comment.line.go meta.annotation.go
    ```
    rather than like so
    ```
        //lint:ignore U1000 Reason.
    //  ^^^^^^^^^^^^^ comment.line.go meta.annotation.go
    ```
   The former having ugly background highlighting as described above.

deathaxe: > 1. Why not make to the end of the text like so:

Because we want a single `meta.annotation` on stack at any point, only. Each character is scoped `meta.annotation`. `U1000 Reason.` are scoped `meta.annotation as well, but have assigned another "subscope" `parameters`. The API function `view.match_selector(pt, `meta.annotation`) would return `True` for each character.

```
    //lint:ignore U1000 Reason.
//  ^^^^^^^^^^^^^^ meta.annotation
//                ^^^^^^^^^^^^^ meta.annotation.parameters
```

Previously it was:

```
    //lint:ignore U1000 Reason.
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.annotation
//                ^^^^^^^^^^^^^ meta.annotation.parameters
```

So `U1000 Reason.` were scoped `meta.annotation meta.annotation.parameters`. This is the kind of scope duplication or "stacking of scopes", we want to avoid here.

> 2. If they do not make it to the end of the line as above, why cover the space after the directive name ...

The `[ ]?` at the end of https://github.com/sublimehq/Packages/blob/91daefae708d63281c43a65e14867286ebb9238e/Go/Go.sublime-syntax#L135 causes `meta.annotation.parameters` which is matched by https://github.com/sublimehq/Packages/blob/91daefae708d63281c43a65e14867286ebb9238e/Go/Go.sublime-syntax#L143-L144 to start at `U1000` not covering the space in front of it.

I just truncated the `meta.annotation.go` without touching boundaries of `meta.annotation.parameters`. And I applied this scheme to all the other patterns for consistency reasons. The space should (must) be in- or excluded the same way in all patterns.

_It would be absolutly correct to change that behavior, if we do so for all patterns in the same way._


> The former having ugly background highlighting as described above.

I can't really follow, because a color scheme should address `meta.annotation` rather than `meta.annotation.go` to ensure to correctly apply to all syntaxes. If it does so, an assigned background would span the whole line (as my screenshot above illustrates). As `meta.variable` is assigned to `ignore` only, any color applied to that scope would not apply to the space after `ignore`.


```
    //lint:ignore U1000 Reason.
//  ^^^^^^^^^^^^^^ meta.annotation
//         ^^^^^^ meta.variable.function
//               ^ - meta.variable.function
```

kortschak: Thanks. I'll think about this. Part of the issue wit the image is that the contrast is just not visible; I can't tell what it is saying at all.

deathaxe: grafik

kortschak: The test marks are fine, just the contrast. This is better though, thanks.

I think that given the definition of the derectives includes a punctuation space between the directive and the parameters that space+ should probably be its own capture like the colon is. I'll send a PR tomorrow with my tweaks.

kortschak: Had some time to think more clearly about this and I agree, this is good.

Go/Go.sublime-syntax
```diff
@@ -118,25 +121,55 @@ contexts:
...
     # as plain comments by default, and add some detailed meta scopes for
     # enterprising users wishing to add more color.
-    - match: (//)(go)(:)({{ident}})?
+    - match: (//)([a-z]+[a-z-]*)(:)({{directive}})[ ]?
+      scope: meta.annotation.go
       captures:
         1: punctuation.definition.comment.go
         2: meta.keyword.annotation.go
-        3: meta.punctuation.accessor.colon.go
+        3: punctuation.accessor.colon.go
```
jwortmann, line 140: I wonder why this `meta.` prefix is removed for the colon scope name here? This will result in only the colon to be highlighted and stand out by default. Shouldn't the `meta.` either be removed from the `meta.keyword`, `meta.variable.function` and `meta.variable.parameter` scopes as well then, or alternatively, if the directive comments shouldn't be highlighted, then keep it for the colon scope too?

deathaxe: Absolutely correct! As stated in one of my comments above I was (am) tempted to remove those `meta.` scope prefix to enable highlighting such annotations because this is what we started by supporting comments with special meaning like `/// <directive>` in C#. If we don't want to highlight those, we don't need to add special patterns to scope them.

This change is a left over from tests related with those considerations.

kortschak:
Please revert c41f4ead5408a8a709062254d25311ae41b646fc, it was better before that.

deathaxe:
We could scope whitespace as `punctuation` but only background colors could be applied. Not sure if it's worth it.

kortschak:
The reason I think it should be like that is if people are using background to separately mark the function from the parameters, that will bleed into the space. I don't use background, but I do mark the function differently to the parameters.

kortschak:
Sorry, clobbered your commit. The naming of the space can be changed to suit.

deathaxe (approved):

michaelblyons (approved):

kortschak:
@deathaxe Thank you for your help and patience with this.

PR #3245, opened 2022-02-23
[Python] Fix implicit tuples termination

### LINKED ISSUE #3243: [Python] Trailing commas in assignments break highlighting
### What happened?

```py
class Test:
    __slots__ = "items",
    
    def func():
        pass
```

A trailing comma in an assignment implicitly denotes a tuple, so the above would be equivalent to `__slots__= ("items",)`.

deathaxe:
Fixes #3243

Pop `allow-unpack-operators` off stack directly at the beginning of a line if it doesn't begin with optional whitespace followed by asterisk or hashtag.

This is to ensure to correctly match `def` statements whose pattern requires to be matched at bol.

The new pattern is appended to keep performance impact as low as possible. It is matched upon newlines only.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1014,6 +1014,8 @@ contexts:
       pop: true
     - match: (?=\S)
       pop: true
+    - match: ^(?!\s*[#*])
+      pop: true
 
   classes:
     - match: '^\s*(class)\b'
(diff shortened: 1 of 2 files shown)


Thom1729:
(Not important, but I don't think that rule order matters for performance. I think that all of the patterns for a context are compiled into a single state machine, and unless two patterns could overlap, the order shouldn't matter.)

FichteFoll (approved):

PR #3250, opened 2022-02-25
[CSS] Update completions from MDN

### LINKED ISSUE #3039: [CSS]: Css auto-completion doesn't have several values of 'display'
### What happened?

Proof: https://developer.mozilla.org/en-US/docs/Web/CSS/display

Paste this text, I marked few items which are NOT suggested by auto-complete, by `MISS`.
```
p{
/* legacy values */
display: block;
display: inline;
display: inline-block;
display: flex;
display: inline-flex;
display: grid;
display: inline-grid;
display: flow-root; /* MISS */

/* box generation */
display: none;
display: contents;

/* two-value syntax */
display: flow; /* MISS */
display: inline flow;
display: inline flow-root;
display: block flex;
display: inline flex;
display: block grid;
display: inline grid;
display: block flow-root;

/* other values */
display: table;
display: table-row; /* all table elements have an equivalent CSS display value */
display: list-item;

/* Global values */
display: inherit;
display: initial; 
display: revert; /* MISS */
display: unset;
}
```

deathaxe:
Fixes #3039

This PR updates completions using data from:

  https://www.chromestatus.com/metrics/css/popularity
  https://developer.mozilla.org/en-US/docs/Web/CSS

Completions data is moved to "completions" package as 2k4 lines of data seem just a bit too much for the single css_completions module.

Note: The list of added properties may still not be complete. It should improve completeness of completions significantly though.


CSS/completions/properties.py
```diff
@@ -0,0 +1,1366 @@
+from .common import get_common_values
+
+
+def get_properties():
```
keith-hall, line 4: I personally don't like the idea of prefixed properties being automatically completed/suggested, but I could well be in the minority here, so it would be interesting to hear what other people think. I'd be okay with an easy way to disable them, however.

deathaxe: I don't have a strong oppinion about it. Just copied what MDN offered. Removing is always an easy fix. :-D

deathaxe: Removed.

keith-hall (approved):

michaelblyons (approved):

PR #3255, opened 2022-03-05
[JavaScript] Decorators can't have bracketed accesses.

### LINKED ISSUE #3253: [TypeScript] Highlighting broken for strings and brackets
### What happened?

When writing Typescript using the below example, highlighting is broken for strings and brackets.

image

```typescript
import ApiService from "@/core/services/ApiService";
import { Actions, Mutations } from "@/store/enums/StoreEnums";
import { User } from "@/store/modules/AuthModule";
import { AxiosRequestConfig } from "axios";
import { Action, Module, VuexModule } from "vuex-module-decorators";

export interface UserInfo {
  errors: unknown;
  user: User;
}

@Module
export default class UserModule extends VuexModule implements UserInfo {
  errors = {};
  user = {} as User;

  /**
   * Get current user object
   * @returns User
   */
  get currentUser(): User {
    return this.user;
  }

  /**
   * Get errors
   * @returns array
   */
  get getErrors() {
    return this.errors;
  }

  @Action
  [Actions.RETRIEVE_USER](credentials: AxiosRequestConfig) {
    return ApiService.post("/v1/auth/login", credentials)
      .then(({ data }) => {
        this.context.commit(Mutations.SET_AUTH, data);
      })
      .catch(({ response }) => {
        this.context.commit(Mutations.SET_ERROR, response.data.message);
      });
  }
}
```

Thom1729:
Fix #3253.

Annotations used `left-expression-end`, but a left-expression can contain a bracketed property access while an annotation cannot.

Also slightly improve scoping for annotations with tagged templates.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -926,7 +926,7 @@ contexts:
     - include: immediately-pop
 
   decorator-name:
-    - match: '{{identifier_name}}{{left_expression_end_lookahead}}'
+    - match: '{{identifier_name}}(?!\s*[.\(`])'
       scope: variable.annotation.js
       pop: true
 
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

deathaxe:
Not sure if it should be handled separately, but is it possible to terminate `meta.annotation` earlier, so it doesn't span to the next token (including linefeeds)?

grafik
grafik

deathaxe (approved):

Thom1729:
@deathaxe Good idea. I'll do this in a separate PR, and also look at other line continuations that might benefit from similar treatment.

PR #3256, opened 2022-03-05
[Python] Fix constants highlighting in class declaration parameters

### LINKED ISSUE #3252: [Python] Wrong scope for boolean keyword arguments in inherited class list
### What happened?

For example, the following test should pass.

```py
class EventModifierKeysDict(TypedDict, total=False):
#                                            ^^^^^ constant.language.boolean
    primary: bool
    ctrl: bool
    alt: bool
    altgr: bool
    shift: bool
    super: bool
```

Now it's scoped as `entity.other.inherited-class.python`.

deathaxe:
Fixes #3252

A simple approach to fix highlighting of constants in class declarations.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1056,7 +1056,7 @@ contexts:
               captures:
                 1: variable.parameter.class-inheritance.python
                 2: keyword.operator.assignment.python
-            - match: (?={{path}})
+            - match: (?=(?!None|False|True|Ellipsis|NotImplemented|__debug__){{path}})
               push:
                 - meta_scope: entity.other.inherited-class.python
                 - match: '{{identifier}}(?: *(\.) *)?'
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1052,11 +1052,12 @@ contexts:
             - match: ','
               scope: punctuation.separator.inheritance.python
             - include: illegal-name
-            - match: ({{identifier}}) *(=)
+            - include: constants
+            - match: (?:({{identifier}}) *)?(=)
               captures:
                 1: variable.parameter.class-inheritance.python
                 2: keyword.operator.assignment.python
-            - match: (?=(?!None|False|True|Ellipsis|NotImplemented|__debug__){{path}})
+            - match: (?={{path}})
               push:
                 - meta_scope: entity.other.inherited-class.python
                 - match: '{{identifier}}(?: *(\.) *)?'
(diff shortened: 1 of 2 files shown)


Python/Python.sublime-syntax
```diff
@@ -1056,7 +1056,7 @@ contexts:
               captures:
                 1: variable.parameter.class-inheritance.python
                 2: keyword.operator.assignment.python
-            - match: (?={{path}})
+            - match: (?=(?!None|False|True|Ellipsis|NotImplemented|__debug__){{path}})
```
keith-hall, line 1059: I feel like this should use a variable or something ideally

keith-hall (approved):

Python/Python.sublime-syntax
deathaxe: An alternative might be to include `constants` context before this pattern if we are concerned about consistency in case of more of them being added in future.

As all those constants are matched in separate patterns and this negative lookahead seems just required in this special case, I wonder what value a putting them into a variable has.

In order to ensure not any of those special constants is otherwise matched the negative lookahead would probably be needed in front of `{{path}}` or even `{{identifier}}` but I am concerned about too many other implications.

deathaxe: ... which arises a general design topic. What I have in mind is just the way how predefined keywords are prevented from being scoped as identifier in Java. Such a change - if required - would have some larger impact, I guess.

keith-hall (approved):

Python/Python.sublime-syntax
keith-hall: > if we are concerned about consistency in case of more of them being added in future.

that was indeed my worry :)

FichteFoll (approved):

Python/syntax_test_python.py
```diff
@@ -1613,6 +1614,21 @@ class MyClass(Inherited, \
...
+#     ^^^^^^^^^ entity.name.class.python
+#               ^^^^^^^^^ entity.other.inherited-class.python
+#                        ^ punctuation.separator.inheritance.python
+#                          ^^^^ constant.language.null.python
+#                              ^ punctuation.separator.inheritance.python
+#                                ^^^^^ variable.parameter.class-inheritance.python
+#                                     ^ keyword.operator.assignment.python
+#                                      ^^^^^ constant.language.boolean.python
+#                                           ^ punctuation.separator.inheritance.python
+#                                             ^^^^ constant.language.boolean.python
```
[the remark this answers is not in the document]
deathaxe: > ... is meant to test that the identifier used here is reserved? ...

Eaxactly.

> introducing an `invalid_targets` variable

Would be an option as language constants are obviously not to be used as assignment targets. Just a question of how much effort to put into all the illegal scoping as it can easily become quite complex in general. pyright clearly marks it illegal, already.

PR #3257, opened 2022-03-05
[JavaScript] Use branching to improve continuation scopes.

Thom1729:
Suggested by @deathaxe  in https://github.com/sublimehq/Packages/pull/3255#issuecomment-1059741584.

When a decorator ends at the end of a line, end the annotation scope there. This requires branching to look across the line boundary.

Also reimplement expression statement line continuation in the same way. This eliminates the old set/set back pattern in favor of branch/pop. I was going to reuse this pattern in `variable-binding-list-top`, but it turns out that the line continuation code there wasn't doing anything in the first place, so I removed it.


deathaxe:
Should it also work after parens?

grafik

Thom1729:
Good catches. The problem with `@Module(parameters)` was a presumably-longstanding bug where the function call meta context was running too long, which was easily fixed. The problem with `@(whatever)` was that it was looking for a `=>` after the close paren, but after reading the [decorators proposal](https://github.com/tc39/proposal-decorators) and trying it out in AST explorer, arrow functions aren't allowed there anyway. I can probably cut way back on `decorator-expression-begin`, but decorators are still in Stage 2 and there's no formal spec yet, so I'll punt on that.

deathaxe (approved):
I am fine with the changes. Not sure how to handle failing tests of ST4126 though. It won't ever see those changes, but the failing test prevents merging.

keith-hall:
> I am fine with the changes. Not sure how to handle failing tests of ST4126 though. It won't ever see those changes, but the failing test prevents merging.

I think we can update CI to make the test run against stable optional.

jrappen:
The test against stable (in this case 4126) is being run optionally.

I assume the run was aborted, as the other (mandatory) one passed before it (the optional one) finished. Will have to double-check.

@Thom1729 @keith-hall @deathaxe You should be able to re-run the tests. There should be a button for you (as contributors) at the top right of the page.

jrappen:
... or there was a time-out (currently set to 15 mins).

deathaxe:
Already re-ran it. ST4216 gets stuck due to some infinite loop, I guess. Maybe an issue like https://github.com/sublimehq/sublime_text/issues/5280

Thom1729:
I downloaded 4126 for Mac, and the tests run fine. Not sure what's going on with the runner. I might tweak the PR anyway just to see if it helps.

deathaxe:
ST 4126 on Windows 11 fails to load `syntax_test_js_indent_common.js`. The progress bar stops at 90% and ST freezes.

JavaScript/JavaScript.sublime-syntax
```diff
@@ -937,18 +935,29 @@ contexts:
...
 
-    - include: expression-break
-
     - match: (?=`)
       push: literal-string-template
 
     - match: (?=(?:{{dot_accessor}})?\()
       push: function-call-arguments
 
+    - match: (?={{line_ending_ahead}})
```
deathaxe, line 944:
```suggestion
    - match: '{{line_ending_ahead}}'
```

deathaxe:
Rebasing this PR on current master should solve CI issues.

deathaxe (approved):

michaelblyons (approved):

PR #3259, opened 2022-03-06
[PHP] Fix SQL query concatenation 

### LINKED ISSUE #572: [PHP] SQL Syntax highlighting breaks at first appearance of matching quote
```
<?
$query_1 = "SELECT * FROM Database.dbo.test WHERE 1=1;";
$query_2 = "SELECT * FROM ".DATABASE_MAIN.".dbo.test WHERE 1=1;";
```

image

I'm not sure how feasible it would be for syntax highlighting to cover everything between the assignment operator(=) and instruction terminator(;) during an assignment but I(and my team) would greatly appreciate it.

Regardless of feasibility, thanks for an amazing product!

deathaxe:
Fixes #572

see: https://forum.sublimetext.com/t/color-of-sql-query-change-when-add-quotes/63178/2

This commit adds support for highlighting sql queries in all parts of a chained string.

```
$sql = "SELECT " . $col . " FROM " . $table . " WHERE foo is " . $bar ;
```

grafik

_Note: This PR may cause some conflicts with #3046. Depending on which one is merged first, we might need to tweak the one or the other slighly afterwards._

### DIFF
--- a/PHP/tests/syntax_test_php.php
+++ b/PHP/tests/syntax_test_php.php
@@ -1728,6 +1728,38 @@ $sql = "
 ";
 // <- meta.string.php string.quoted.double.php punctuation.definition.string.end.php - meta.interpolation - string string
 
+$sql = "SELECT " . $col . "FROM $table WHERE ( first_name =" . $name . ")" ; . "GROUP BY" ;
+//     ^ meta.string.php - meta.interpolation
+//      ^^^^^^^ meta.string.php meta.interpolation.php source.sql.embedded.php
+//             ^ meta.string.php - meta.interpolation
+//              ^^^^^^^^^^ - meta.string
+//                        ^ meta.string.php - meta.interpolation
+//                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.string.php meta.interpolation.php source.sql.embedded.php
+//                                                         ^ meta.string.php - meta.interpolation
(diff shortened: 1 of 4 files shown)


PHP/tests/syntax_test_php.php
```diff
@@ -1750,6 +1783,37 @@ function generate2()
 ';
 // <- meta.string.php string.quoted.single.php punctuation.definition.string.end.php - meta.interpolation - string string
 
+$sql = 'SELECT ' . $col . 'FROM table WHERE ( first_name =' . $name . ')' ; . 'GROUP BY' ;
```
jfcherng, line 1786: I am a little curious about the purpose of this `;`.

```
$sql = 'SELECT ' . $col . 'FROM table WHERE ( first_name =' . $name . ')' ; . 'GROUP BY' ;
//                                                                        ^
```

keith-hall (approved):

PHP/tests/syntax_test_php.php
deathaxe: It is to illustrate SQL highlighting stops on expression/statement boundaries.

deathaxe: We could also add tests with concated strings in parens etc.

jfcherng (approved):

michaelblyons:
Wild.

FichteFoll:
I can practicall *feel* the SQL injection over here.

PR #3262, opened 2022-03-06
[CSS] Add space after colon only via `auto_complete_trailing_spaces`

deathaxe:
This commit is the 3rd alternative for #2886 and #2912.

It uses the default `auto_complete_trailing_spaces` setting to decide
whether to add a space after colons or not when completing properties.

Default behavior (`auto_complete_trailing_spaces`: true):

    font-fam|    =>   font-family: |;

Custom behavior (`auto_complete_trailing_spaces`: false):

    font-fam|    =>   font-family:|;

Default bindings no longer add space when typing `:` (colon):

    font-family|    =>   font-family:|;

inspired by: https://github.com/sublimehq/Packages/pull/2886#issuecomment-922543383

### DIFF
--- a/CSS/Default.sublime-keymap
+++ b/CSS/Default.sublime-keymap
@@ -1,24 +1,13 @@
 [
 	// Add a colon followed by space and semicolon after a property name,
 	// if the next following token doesn't look like a value.
-	{ "keys": [":"], "command": "insert_snippet", "args": {"contents": ": $0;"}, "context":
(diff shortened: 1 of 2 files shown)


jrappen (approved):

FichteFoll (approved):
This seems to be a low-friction change that is easier and quicker to reason about than #2886 while still allowing that to be a follow-up and addition to this change.

michaelblyons (approved):

PR #3265, opened 2022-03-11
[YAML] Fix decimal floating point regexp.

azizk:
The official page lists a regular expression which seems incorrect.

It allows corner cases like `.`, `..`, `...` etc. to be considered decimal floats.

Python's yaml module does not parse them as a number, but as strings.
```python
import yaml
yaml.safe_load("float: ...") == {'float': '...'}
# True
```
The (n)vim editor highlights these cases as normal text.

PS.: The one thing I'm unsure about is whether to allow numbers like `._123`. Python's yaml parser doesn't...
PPS.: You can edit the commit if you don't like the style of the regexp option `(?x)`. But I would suggest to keep it and refactor the other regexps to match that style. :smile:


keith-hall (dismissed):

YAML/YAML.sublime-syntax
```diff
@@ -171,9 +171,17 @@ variables:
   _type_int_other: ([-+]?)([1-9][0-9_]*(?::[0-5]?[0-9])+) # (base 60)
 
   # http://yaml.org/type/float.html
-  _type_float_decimal: ([-+]?)((?:[0-9][0-9_]*)?(\.)[0-9.]*(?:[eE][-+][0-9]+)?)
+  _type_float_decimal: |-
+    (?x)
+    ([-+]?) (
+      (?:
+        (?:[0-9][0-9_]*) (\.) [0-9_]*
+      | (\.) (?:[0-9][0-9_]+)
```
deathaxe, line 179: I'd suggest

```suggestion
      | (\.) [0-9_]+
```

to correctly highlight floating point numbers like `.1` with only one decimal. If we are concerned about `._` being highlighted as float per accident, we might  want to use 

```suggestion
      | (\.) (?:[0-9][0-9_]*)
```

YAML/tests/syntax_test_types.yaml
```diff
@@ -133,6 +133,10 @@
    .NaN, ~, .Nan,
 #  ^^^^ meta.number.yaml constant.language.nan.yaml
 #           ^^^^ - constant
+   ., .., ...,
+#         ^^^ string.unquoted.plain.in.yaml - meta.number.float
+#     ^^ string.unquoted.plain.in.yaml - meta.number.float
+#  ^ string.unquoted.plain.in.yaml - meta.number.float
```
deathaxe, line 139: We generally prefer it like so

```suggestion
#  ^ string.unquoted.plain.in.yaml - meta.number.float
#     ^^ string.unquoted.plain.in.yaml - meta.number.float
#         ^^^ string.unquoted.plain.in.yaml - meta.number.float
```

deathaxe (dismissed):

YAML/tests/syntax_test_types.yaml
```diff
@@ -133,6 +133,11 @@
    .NaN, ~, .Nan,
 #  ^^^^ meta.number.yaml constant.language.nan.yaml
 #           ^^^^ - constant
+## Note: it's uncertain whether `._1` should be considered valid.
```
deathaxe, line 136: Highlighting helps with smoother UX while editing though. We don't neeed or shouldn't be too restrictive anyway.

azizk: In this case I'm not sure, because it makes a difference to see if something is actually (processed as) a number or a piece of text, no?

YAML/YAML.sublime-syntax
```diff
@@ -171,9 +171,17 @@ variables:
   _type_int_other: ([-+]?)([1-9][0-9_]*(?::[0-5]?[0-9])+) # (base 60)
 
   # http://yaml.org/type/float.html
-  _type_float_decimal: ([-+]?)((?:[0-9][0-9_]*)?(\.)[0-9.]*(?:[eE][-+][0-9]+)?)
+  _type_float_decimal: |-
+    (?x)
+    ([-+]?) (
+      (?:
+        (?:[0-9][0-9_]*) (\.) [0-9_]*
+      | (\.) (?:[0-9][0-9_]*)
```
FichteFoll, line 179:
```suggestion
      | (\.) (?:_*[0-9][0-9_]*)
```

YAML/tests/syntax_test_types.yaml
```diff
@@ -133,6 +133,15 @@
...
+   .1_, ._1, +._12,
+#  ^^^ meta.number.float.decimal.yaml constant.numeric.value.yaml
+#       ^^^ string.unquoted.plain.in.yaml
+#            ^^^^^ string.unquoted.plain.in.yaml
```
FichteFoll, line 140:
```suggestion
#       ^^^ string.unquoted.plain.in.yaml constant.numeric.value.yaml
#            ^^^^^ string.unquoted.plain.in.yaml constant.numeric.value.yaml
```

FichteFoll (changes requested):
Note that neither [1.2.0](https://yaml.org/spec/1.2.0/#id2604185) nor [1.2.2](https://yaml.org/spec/1.2.2/#1032-tag-resolution) mention underscores in their schema definition, while the [float type page for 1.1](https://yaml.org/type/float.html) mentions underscores as being allowed anywhere. Since we aim for the highest compatibility, we should thus

1. continue to support underscores,
2. support `._1` (see suggestions),
3. probably not support `._` and fall back to string highlighting, because I agree that the accuracy of highlighting numbers correctly is more important than highlighting while typing (and the fallback is a string anyway, not some invalid syntax). Please add a test for that.

There are always going to be differences with some parser because many of them are either non-compliant, don't use the core schema as the default (or at all), or support a mix between YAML 1.1 and 1.2 and we want to support both here anyway. Unless the majority of commonly used parsers agree on a syntax (like `._` being parsed as a float or `._1` as a string, which I didn't check).

YAML/YAML.sublime-syntax
deathaxe: In that case we can simplify the pattern to

```suggestion
      | (\.) [0-9_]+
```

deathaxe (approved):

azizk:
I agree underscores should be supported. It's strange that v1.2.2 doesn't mention them. What's interesting is the regexp actually, which looks more sane than the one given on `float.html`.
`[-+]? ( \. [0-9]+ | [0-9]+ ( \. [0-9]* )? ) ( [eE] [-+]? [0-9]+ )?`
I think I'll add a commit to make the `-+` in the exponent optional. :+1:

deathaxe:
It appears pyyaml requires the sign. Did you see other parsers not doing so?

azizk:
Yeah, looks like it. I just went with the 1.2.2 spec. Maybe parsers out there need to catch up?

azizk:
@deathaxe Btw, should we disallow `123.___`?

deathaxe:
pyyaml seems to accept it as float.

```py
Python 3.3>>> yaml.load("test: 1.___")
{'test': 1.0}
```

azizk:
Yeah I tested it. It's kinda sad if you ask me. So many strange variants are accepted. :roll_eyes: :smile:
Anyway, I think it's ready unless something else pops up. :+1:

FichteFoll:
FYI, I suggest https://github.com/yaml/yaml-editor as a helpful tool for testing specific syntax with various yaml parsers.

FichteFoll (approved):

deathaxe (approved):

PR #3287, opened 2022-03-23
[CSS] Add hyphen to list of sub-word-separators

### LINKED ISSUE #2709: [CSS] Subword navigation not working with dash in names
- Sublime Version: 4096
- OS Version: Windows build 19042.746

## Expected behavior
Subword navigation should stop at dashes in the middle of class names, property names, etc.

## Actual behavior
It skips dashes. It works with camelCase however.

Subword nav works with dashes in HTML files, so I'm guessing it's specific to the syntax definition of CSS. Am I wrong?

deathaxe:
Fixes #2709

### DIFF
--- a/CSS/CSS.sublime-settings
+++ b/CSS/CSS.sublime-settings
@@ -8,4 +8,6 @@
 
     // Default separators except `-`
     "word_separators": "./\\()\"':,.;<>~!@#$%^&*|+=[]{}`~?",
+
+    "sub_word_separators": "_-",
 }
--- a/CSS/CSS.sublime-settings
+++ b/CSS/CSS.sublime-settings
@@ -9,5 +9,6 @@
     // Default separators except `-`
     "word_separators": "./\\()\"':,.;<>~!@#$%^&*|+=[]{}`~?",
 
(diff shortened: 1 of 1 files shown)


michaelblyons (dismissed):

FichteFoll:
I'm not at a pc currently and won't for a while, but is the subword separator setting empty by default?

deathaxe:
It contains `_` by default, so I kept it and just added `-`.

keith-hall (dismissed):

CSS/CSS.sublime-settings
```diff
@@ -8,4 +8,6 @@
 
     // Default separators except `-`
     "word_separators": "./\\()\"':,.;<>~!@#$%^&*|+=[]{}`~?",
+
+    "sub_word_separators": "_-",
```
FichteFoll, line 12: A comment denoting the difference to the default would be appreciated, similar to the `word_separators` setting, so that it can be synchronized more easily in the future when the default changes. (And gives some context to the change.)

michaelblyons (approved):

keith-hall (approved):

Issue #3228, opened 2022-02-03
[RFC] Key-Value Separators


deathaxe:
## Relational

Several syntaxes such as Python, JavaScript or JSON support data objects in the form of arrays, lists or key-value pairs.

Those data objects can be grouped into:

1. sequences
   - arrays _(C, C++, ...)_
   - lists _(Python, JS, JSON, ...)_
   - vectors _(C++ std/boost library)_
   - ...
2. mappings _(aka. key-value pairs)_
   - dictionaries _(VBScript)_
   - maps _(C++ std/boost library)_
   - objects _(JSON)_
   - ...

That's pretty clearly expressed by meta scopes `meta.sequence` and `meta.mapping`.

List separators are commonly scoped `punctuation.separator.sequence`.

```jsonc
["item1", "item2", ...]
//      ^ punctuation.separator.sequence
```

The situation with mapping punctuations seems a bit more unclear at the moment which caused in several solutions having found their way into the wild.

```jsonc
{"key1": "value1", "key2": "value2", ...}
//     ^ punctuation.separator.???
//               ^ punctuation.separator.???
```

## Motivation

The current situation is reflected by a couple of discussions started in pending PRs, such as https://github.com/sublimehq/Packages/pull/2682#issuecomment-964466320 or https://github.com/sublimehq/Packages/pull/2564#discussion_r788000603 which re-introduces `punctuation.separator.mapping` to YAML, while it has been removed by https://github.com/sublimehq/Packages/commit/60e7c391deb32d1d39b5add199f827c7d3bb8050.

If this topic has been discussed somewhere the related posts seem lost. Hence it may make sense to come up with a guideline how to commonly scope punctuations in data objects such as mappings or sequences in a (this) RFC issue.

## Current Situation

The following table, started by @jwortmann, illustrates currently used scopes for separators of key-value pairs.

| Syntax | key-value separator, e.g. `:` or `=` | entry separator, e.g. `,` or `;` | uses `meta.mapping` |
| ------ | ----------------------------- | ------------------------- | ----------------- |
[...]

PR #3289, opened 2022-03-25
[Go] preliminary support for Go 1.18

### LINKED ISSUE #3281: [Go] Support go1.18
### What happened?

Some points:
`
func MyFunc[T any](s []T) 
func MyFunc2[K comparable, V int64 | float64](m map[K]V) V 
type Number interface {
    int64 | float64
}
`

mitranim:
Features:

  * Support for `~`.
  * Support for `any` and `comparable`.
  * Support for type unions.
  * Support for general interfaces (with type unions).
  * Support for type parameters in generic functions.
  * Support for type parameters in method receivers of generic types.
  * Support for type arguments following type names.
  * Support for type parameters in type declarations.
  * Support for embedded parametrized types in struct definitions.
  * Support for type arguments between function name and function arguments.
  * Support for `invalid.illegal` for dangling `)]}`.
  * Slightly better meta scopes in type definitions.

Known regressions:

  * Embedded interfaces are no longer scoped as "inherited class", due to ambiguities.
  * "Namespace" idents such as `ident.` no longer support a comment between identifier and dot.
  * Last ident in parens followed by opening paren is no longer scoped as a function call, due to false positives and ambiguities.

Known issues and limitations:

  * Detection of type parameter lists and type argument lists is incomplete. In some places it relies on regex lookahead, which doesn't work across multiple lines. There are also edge case ambiguities which are supported by the Go compiler but not by our syntax.

Misc tweaks in comments:

  * Convert anon to named contexts for compatibility with overrides.
  * Avoid false positives when scoping javadoc bullets.
  * Use `prototype` to reduce repetition.

Closes #3281.

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -238,7 +238,7 @@ contexts:
     - meta_scope: comment.block.go
     - match: \*/
       scope: punctuation.definition.comment.end.go
-      pop: true
+      pop: 1
     # Indentation, followed by exactly one space, followed by *.
     # This enables support for javadoc comments, while avoiding false
     # positives in non-javadoc comments that use "*" as simple list bullets.
@@ -365,7 +365,7 @@ contexts:
   pop-identifier-anon:
     - match: '{{ident_anon}}'
       scope: variable.language.anonymous.go
-      pop: true
+      pop: 1
 
   # https://golang.org/ref/spec#Operators_and_punctuation
   #
@@ -441,10 +441,10 @@ contexts:
   pop-selector:
     - match: '{{ident_anon}}(?!{{noise}}\()'
       scope: variable.language.anonymous.go
-      pop: true
+      pop: 1
     - match: '{{ident}}(?!{{noise}}\()'
       scope: variable.other.member.go
-      pop: true
+      pop: 1
     - match: \(
       scope: punctuation.section.parens.begin.go
       set: pop-type-assertion
@@ -474,7 +474,7 @@ contexts:
 
   pop-brackets-inner:
     - match: ''
-      pop: true
+      pop: 1
       branch_point: point-brackets-inner
       branch:
         - pop-brackets-inner-type-parameter-list
@@ -644,7 +644,7 @@ contexts:
         - meta_scope: string.quoted.other.go
         - match: '`'
           scope: punctuation.definition.string.end.go
-          pop: true
+          pop: 1
         - include: match-template-string
         - match: \%%
           scope: constant.character.escape.go
@@ -658,7 +658,7 @@ contexts:
         - meta_scope: string.quoted.double.go
         - match: '"'
           scope: punctuation.definition.string.end.go
-          pop: true
+          pop: 1
         - include: match-template-string
         - match: '{{char_escape}}'
           scope: constant.character.escape.go
@@ -674,7 +674,7 @@ contexts:
        - clear_scopes: 1
        - match: "}}"
          scope: punctuation.section.interpolation.end.go
-         pop: true
+         pop: 1
        - match: "\\s-"
          scope: keyword.operator.template.right.trim.go
        - match: "-\\s"
@@ -837,7 +837,7 @@ contexts:
   pop-package-name:
     - match: '{{ident}}'
       scope: entity.name.namespace.go
-      pop: true
+      pop: 1
     - include: pop-before-nonblank
 
   match-keyword-range:
@@ -962,11 +962,11 @@ contexts:
   pop-func-name:
     - match: '{{ident_anon}}'
       scope: variable.language.anonymous.go
-      pop: true
+      pop: 1
 
     - match: '{{ident}}'
       scope: entity.name.function.go
-      pop: true
+      pop: 1
 
     - include: pop-before-nonblank
 
@@ -1043,7 +1043,7 @@ contexts:
 
   pop-parameter-type:
     - match: (?=\)|,)
-      pop: true
+      pop: 1
     - include: match-ellipsis
     - include: pop-type
 
@@ -1114,19 +1114,19 @@ contexts:
 
   pop-now:
     - match: ''
-      pop: true
+      pop: 1
 
   pop-before-nonblank:
     - match: (?=\S)
-      pop: true
+      pop: 1
 
   pop-before-semicolon:
     - match: (?=;)
-      pop: true
+      pop: 1
 
   pop-before-newline:
     - match: (?=$)
-      pop: true
+      pop: 1
 
   pop-before-terminator:
     - include: pop-before-semicolon
@@ -1134,30 +1134,30 @@ contexts:
 
   pop-before-paren-end:
     - match: (?=\))
-      pop: true
+      pop: 1
 
   pop-on-paren-end:
     - match: \)
       scope: punctuation.section.parens.end.go
-      pop: true
+      pop: 1
 
   pop-before-bracket-end:
     - match: (?=\])
-      pop: true
+      pop: 1
 
   pop-on-bracket-end:
     - match: \]
       scope: punctuation.section.brackets.end.go
-      pop: true
+      pop: 1
 
   pop-on-brace-end:
     - match: \}
       scope: punctuation.section.braces.end.go
-      pop: true
+      pop: 1
 
   pop-before-brace-end:
     - match: (?=\})
-      pop: true
+      pop: 1
 
   match-type:
     - match: (?=\S)
@@ -1396,11 +1396,11 @@ contexts:
   pop-member:
     - match: '{{ident_anon}}'
       scope: variable.language.anonymous.go
-      pop: true
+      pop: 1
 
     - match: '{{ident}}'
       scope: variable.other.member.go
-      pop: true
+      pop: 1
 
   pop-arguments-starting-with-type:
     # Go allows functions and types to be wrapped in an arbitrary number of
@@ -1416,7 +1416,7 @@ contexts:
   # should also match \r\n due to Sublime's internal normalization.
   pop-on-eol:
     - match: $\n?
-      pop: true
+      pop: 1
 
   pop-inside-parens:
     - include: pop-on-paren-end
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -641,7 +641,7 @@ contexts:
       scope: punctuation.definition.string.begin.go
       push:
         - meta_include_prototype: false
-        - meta_scope: string.quoted.other.go
+        - meta_scope: string.quoted.backtick.go
         - match: '`'
           scope: punctuation.definition.string.end.go
           pop: 1
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -802,8 +802,7 @@ contexts:
     # A proper implementation may involve complicated branching.
     - match: (?=\([^()]*\){{noise}}{{ident}}{{noise}}[\[\(])
       set: [pop-func-signature, pop-func-method-receiver]
-    - match: (?=\S)
-      set: pop-func-signature
+    - include: pop-func-signature
 
   match-keyword-go:
     - match: \bgo\b
@@ -831,7 +830,7 @@ contexts:
 
   match-keyword-package:
     - match: \bpackage\b
-      scope: storage.type.namespace.go keyword.declaration.namespace.go
+      scope: keyword.declaration.namespace.go
       push: [meta-namespace, pop-package-name]
 
   pop-package-name:
@@ -869,8 +868,7 @@ contexts:
     - match: \(
       scope: punctuation.section.parens.begin.go
       set: pop-type-declaration-list
-    - match: (?=\S)
-      set: pop-type-declaration
+    - include: pop-type-declaration
 
   pop-type-declaration-list:
     - include: pop-on-paren-end
@@ -1271,7 +1269,7 @@ contexts:
     - include: pop-before-nonblank
 
   # Ideally we would scope embedded interfaces as `entity.other.inherited-class`
-  # rather than `storage.type` Prior to Go 1.18 and type unions, it was
+  # rather than `storage.type`. Prior to Go 1.18 and type unions, it was
   # possible to detect embedding unambiguously. With the advent of unions,
   # embeddings are syntactically indistinguishable from single-element unions.
   pop-interface-inner:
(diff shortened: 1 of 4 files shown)


Go/Go.sublime-syntax
```diff
@@ -175,37 +214,45 @@ contexts:
...
-          captures:
-            1: punctuation.definition.comment.go
+  pop-comment-line-annotation:
+    - meta_scope: meta.annotation.go comment.line.go
+    - match: \S+
+      scope: meta.variable.parameter.go
+    # End the annotation scope at EOL, but stretch the comment scope
+    # indefinitely to the right.
+    - match: $
+      set: pop-comment-line-normal
```
deathaxe, line 224: This context seems to be unused and may be removed.

Go/Go.sublime-syntax
```diff
@@ -175,37 +214,45 @@ contexts:
...
+
+  pop-comment-general:
+    - meta_scope: comment.block.go
+    - match: \*/
+      scope: punctuation.definition.comment.end.go
       pop: true
+    # Indentation, followed by exactly one space, followed by *.
+    # This enables support for javadoc comments, while avoiding false
+    # positives in non-javadoc comments that use "*" as simple list bullets.
+    - match: ^(?:\t|[ ]{2}|[ ]{4}|[ ]{8})*[ ](\*)(?!/)
```
deathaxe, line 253: Smart!

Go/syntax_test_go.go
```diff
@@ -35,22 +34,30 @@ You may have to disable Go-specific linters when working on this file.
 //                            ^^^^^^^^^^^ comment.line.go
 
     /*
+//  ^^ comment.block.go punctuation.definition.comment.begin.go
 // ^ -comment
```
deathaxe, line 38: We should probably avoid duplicated tests (of `comment.block). Instead we could ensure correct boundaries by:

```suggestion
// ^ -comment
//  ^^ comment.block.go punctuation.definition.comment.begin.go
//    ^^ comment.block.go - punctuation
```

Go/syntax_test_go.go
```diff
@@ -35,22 +34,30 @@ You may have to disable Go-specific linters when working on this file.
...
 //  ^^^^ comment.block.go
     comment
 //  ^^^^^^^^ comment.block.go
     */
-//  ^^ comment.block.go
+//  ^^ comment.block.go punctuation.definition.comment.end.go
```
deathaxe, line 43: The counter part for checking boundaries of the previous suggestion.

```suggestion
//^^ comment.block.go - punctuation
//  ^^ comment.block.go punctuation.definition.comment.end.go
```

Go/syntax_test_go.go
```diff
@@ -35,22 +34,30 @@ You may have to disable Go-specific linters when working on this file.
...
 //    ^ -comment
 
     /* * */
 // ^ -comment
+//     ^ -punctuation.definition.comment.go
 //  ^^^^^^^ comment.block.go
```
deathaxe, line 49: Just a tiny nit pick.

```suggestion
//  ^^^^^^^ comment.block.go
//     ^ -punctuation.definition.comment.go
```

jrappen:
* It might make sense to move tests to `./Go/tests/` and split them up.
* You're still using syntax v1. Was there any reason not make use of newer v2 syntax engine features?

mitranim:
> * It might make sense to move tests to `./Go/tests/` and split them up.

I have no strong opinion. It has obvious upsides but also some downsides. Search-and-replace opens N files, split-panel becomes slightly more annoying to use.

> * You're still using syntax v1. Was there any reason not make use of newer v2 syntax engine features?

No reason. Updated to v2.

deathaxe:
FWIW There's no hard restriction with regards to used syntax features and syntax version. V2 just gates some backward incompatible changes to the syntax engine, which require a syntax to opt-in. Most of them are related to how meta scopes are applied in push/set/pop scenarios.

Go/Go.sublime-syntax
```diff
@@ -70,19 +71,23 @@ variables:
...
   # https://golang.org/ref/spec#Predeclared_identifiers
   predeclared_func: \b(?:append|cap|close|complex|copy|delete|imag|len|make|new|panic|print|println|real|recover)\b
 
+  ident_blank: \b_\b
+
   # Note: this matches ALL valid identifiers, including predeclared constants,
   # functions and types.
   ident: \b(?!{{keyword}})[[:alpha:]_][[:alnum:]_]*\b
 
+  keyword_or_ident: \b[[:alpha:]_][[:alnum:]_]*\b
```
deathaxe, line 85: May probably want to use it in `ident`?

```yaml
ident: (?!{{keyword}}){{keyword_or_ident}}
```

Go/Go.sublime-syntax
```diff
@@ -296,18 +334,23 @@ contexts:
...
       scope: variable.function.go support.function.builtin.go
+    - match: '{{ident}}(?=(?:{{noise}}\))*(?:{{noise}}{{type_argument_list}})?{{noise}}\()'
+      scope: variable.function.go
+      push: pop-type-argument-list
     - match: '{{ident}}(?=(?:{{noise}}\))*{{noise}}\()'
       scope: variable.function.go
 
   # See notes on `match-call-or-type-conversion`.
   pop-call-or-type-conversion:
+    - include: pop-identifier-blank
```
deathaxe, line 353: It appears the whole context is unused.

Go/Go.sublime-syntax
```diff
@@ -388,49 +455,71 @@ contexts:
...
+      pop: true
+    - match: \(
+      scope: punctuation.section.parens.begin.go
+      set: pop-type-assertion
+    - include: pop-before-nonblank
+
+  pop-type-assertion:
+    - include: pop-on-paren-end
+    - match: (?=\S)
+      push: pop-type
```
deathaxe, line 475:
```suggestion
    - include: match-type
```

Go/Go.sublime-syntax
```diff
@@ -967,175 +1059,269 @@ contexts:
...
-  pop-on-terminator:
-    - include: pop-on-semicolon
-    - match: $
+  pop-before-brace-end:
+    - match: (?=\})
       pop: true
 
+  match-type:
+    - match: (?=\S)
+      push: pop-type
```
deathaxe, line 1176: I'd suggest the following to remove one extra context switch. I think this trivial kind of duplication doesn't hurt.

```suggestion
    - match: (?=\S)
      push: [pop-after-type, pop-type-single]
```

Go/Go.sublime-syntax
```diff
@@ -873,67 +944,89 @@ contexts:
...
+    - match: (?=\()
+      set: [pop-func-return-signature, pop-func-parameter-list]
+    - include: pop-before-nonblank
+
+  pop-func-return-signature:
+    - include: pop-before-terminator
+    - match: (?=\()
+      set: pop-func-parameter-list
+    - match: (?=\S)
+      set: pop-type
```
deathaxe, line 993: The way `pop-type` is implemented the following would work, too, and avoid one extra context switch.

```suggestion
    - include: pop-type
```

Go/Go.sublime-syntax
```diff
@@ -967,175 +1059,269 @@ contexts:
...
-      scope: punctuation.section.parens.end.go
-      pop: true
-    - include: match-comments
-    - include: match-keywords
+    - include: pop-on-paren-end
     - include: match-comma
     - include: match-ellipsis
+    - include: match-keywords
     - match: (?=\S)
       push: pop-type
```
deathaxe, line 1067: The way `match-type` is implemented the following would work, too, and avoid one extra context switch.

```suggestion
    - include: match-type
```

Go/Go.sublime-syntax
```diff
@@ -967,175 +1059,269 @@ contexts:
...
+      scope: variable.other.member.declaration.go
+      push: pop-struct-field-meta
+
+    - include: match-any
+
+  pop-struct-field-meta:
+    - include: pop-before-brace-end
+    - include: pop-before-terminator
+    - include: pop-type
+    - include: match-any
```
deathaxe, line 1272: I have a feeling about `pop-type` to need to be the last line as it starts with `(?=\S)`, but I may be wrong.

```suggestion
    - include: match-any
    - include: pop-before-brace-end
    - include: pop-before-terminator
    - include: pop-type
```

deathaxe (commented):
Here's another set of suggestions.

Go/Go.sublime-syntax
mitranim: Wasn't sure if this would work properly because unlike the previous version, this inserts `\b` into the middle of the regex. But tests pass so you might be right.

Go/Go.sublime-syntax
mitranim: The "pop-before" rules must come first, but you're right in that `match-any` wouldn't work after `pop-type`, so I'll just remove `match-any` from here. It was included for graceful degradation, like in some other places.

jrappen:
Appreciate the effort to name contexts.

TheSecEng:
Is this also going to support syntax based code folding ?

jrappen:
@TheSecEng Currently no changes for that, yet.

@mitranim For reference the [`master`](https://github.com/sublimehq/Packages/commits/master) branch already has some commits for the new code folding stuff in other languages.

jrappen:
Maybe address #3228 as well with this PR?

deathaxe:
Syntax based code folding should work pretty well, once each pushed context receives a meta scope so neiboured parens can be detected as separate folding begin/end markers.

Go/Go.sublime-syntax
```diff
@@ -799,37 +859,31 @@ contexts:
...
-                - match: (?=\))
-                  pop: true
-                - include: pop-type-alias-or-typedef
-            - include: match-any
+  pop-type-declaration-or-list:
+    - match: \(
+      scope: punctuation.section.parens.begin.go
+      set: pop-type-declaration-list
+    - match: (?=\S)
+      set: pop-type-declaration
```
deathaxe, line 869: Maybe 

```suggestion
    - include: pop-type-declaration
```

Go/Go.sublime-syntax
```diff
@@ -1202,72 +1352,83 @@ contexts:
...
+    - include: match-any
+
+  match-namespace:
+    - match: '({{ident}})\s*(\.)'
+      captures:
+        1: variable.other.go
+        2: punctuation.accessor.dot.go
+
+  meta-function-declaration:
+    - meta_scope: meta.function.declaration.go
```
deathaxe, line 1421: All those meta-only contexts should contain `- meta_include_prototype: false` to avoid `prototype`s from being prepended. Even if the language itself doesn't define any `prototype` context any inherited syntax might do so in order to inject templating patterns etc.

```suggestion
  meta-function-declaration:
    - meta_include_prototype: false
    - meta_scope: meta.function.declaration.go
```

deathaxe (commented):
The linked issue still contains some open tasks. 

Do you plan to integrate them in this PR?

In case those are planned for future PRs, would you mind adding some small reminder issues with existing limitations and possible enhancements to certain code constructs - maybe including small examples?

mitranim:
Apologies for slow responses. The linked issue had multiple progress reports which were all out of dated, and it wasn't meant to be a progress status of _this_ PR. However, there are other known defects. I'm building a list now.

mitranim:
Known defects / missing features:

* [ ] In parameter lists, types with type arguments in square brackets (inline type instantiation) should be treated as types, rather than regular parameters.

```go
interface{
  Method(Type[TypeArg])
}

func(Type, Type[TypeArg])
```

* [ ] In embedded interfaces with type arguments, type arguments should be scoped as `variable.other.type` rather than `variable.other`.

```go
interface{
  EmbeddedInterface[TypeArg]
}
```

* [ ] "Function call" detection should work with type arguments of arbitrary complexity, rather than just with identifiers.

```go
FuncName[Type[TypeArg]]()
```

* [ ] Parameter lists should support types of arbitrary complexity such as `[][]Ident`.

```go
func FuncName(param [][]Type)
```

deathaxe (dismissed):
This PR provides many improvements with regards to type support.

Not supporting block quotes everywhere in fully-qualified identifiers is acceptable. Many other syntaxes don't do so at all, even though valid by language specs.

Work on Java syntax teached me required complexity needed to handle all those typing and generics "perfect" especially with regards to multi-line support.

So I think we could add this PR as improvement and add an issue with the open "TODOs" to come back to them later. Maybe fixing them in smaller pieces?

jrappen (approved):

mitranim:
I've been using this implementation for months, and while it's still incomplete, it's a massive improvement over not having support for Go 1.18. I would suggest merging this sooner, to improve user experience in the next stable build, and then I'll open a new issue for the remaining unresolved problems.

deathaxe:
Could you probably have a look at the conflicts?

mitranim:
> Maybe you could add to the commit message (the opening comment) that you're addressing RFC issue 2569 with the scopes for doc blocks.

This confuses me. The issue seems to be about syntax highlighting inside docblocks, for stuff like `@param{type}`. Our Go syntax does not support this and hopefully never will, because Go generates documentation from source code, making Javadoc unnecessary. Not sure if this RFC applies.

deathaxe (approved):

mitranim:
For `set: [one, two]` that easily fits under 80 char, I prefer a one-liner, but that's very subjective. I wouldn't mutiny if someone changed it later. Not seeing a strong convention and don't have a strong opinion.

For long regexes that are lists of words such as `(?:one|two|three)`, I subjectively prefer a one-liner for code compactness. They tend to appear in `variables`, and you can easily scroll past a one-liner, while a multi-liner requires you to "parse" code. However, a flat multi-line structure makes Git diffs more useful:

```diff
  predeclared_type: |-
    \b(?x:
-     bool
+     any
+     |bool
      |byte
+     |comparable
      |complex64
      |complex128
      ...
    )\b
```

...as opposed to:

```diff
-  predeclared_type: \b(?:byte|complex64|complex128| ...
+  predeclared_type: \b(?:any|bool|byte|comparable|complex64|complex128 ...
```

So if we're changing variable lists, it should be by flattening them vertically, IMO.

deathaxe:
> For set: `[one, two]` that easily fits under 80 char

I agree, one-liners beeing well suited in case they fit into 80 chars boundary.

Multi-line form got a bit more popular just because of many contexts being pushed at once in syntaxes like Java or JavaScript. They may produce more readable code, if context names get longer.

I've tried to keep with one style recently, but there's no strong convention about it, nor a need for one.

> a flat multi-line structure makes Git diffs more useful

Agree. With a look at Ruby, which makes already use of it, and some experiments with applying it to PHP, I'd be a bit concerned about the huge amount of lines added - especially for syntaxes with insane long lists of builtin funcs/vars/... (such as PHP).

TheSecEng:
Been using this for a month or so. Great work!

michaelblyons:
@patrickrgaffney @jakelucas Any comments?

jakelucas:
> @patrickrgaffney @jakelucas Any comments?

I can only assume I've been pinged here by mistake
Looks good to me

michaelblyons (approved):

PR #3291, opened 2022-03-26
[JSON] Add support for syntax based folding

deathaxe:
Requires ST 4130

This commit ...

1. disables indentation folding
2. introduces a common `meta.fold.[begin|end]` scope
3. adds required `Folding Rules.tmPreferences`
4. reworks comments based on JavaScript.sublime-syntax to properly support highlighting and folding block comments. It keeps double slash line comments only though.

Note: 

1. The rules for `meta.fold.begin` and `meta.fold.end` should probably be added to _Default/Fold.tmPreferences_
2. _Folding Rules.tmPreferences_ is choosen to follow the scheme of _Indentation Rules.tmPreferences_

### DIFF
--- a/JSON/Folding Rules.tmPreferences
+++ b/JSON/Folding Rules.tmPreferences
@@ -11,9 +11,9 @@
 		<array>
 			<dict>
 				<key>begin</key>
-				<string>meta.fold.begin</string>
+				<string>punctuation.definition.comment.begin</string>
 				<key>end</key>
-				<string>meta.fold.end</string>
+				<string>punctuation.definition.comment.end</string>
 			</dict>
 			<dict>
 				<key>begin</key>
--- a/JSON/Folding Rules.tmPreferences
+++ b/JSON/Folding Rules.tmPreferences
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>scope</key>
+	<string>source.json</string>
+	<key>settings</key>
+	<dict>
+		<key>indentationFoldingEnabled</key>
+		<false/>
+		<key>foldScopes</key>
+		<array>
+			<dict>
+				<key>begin</key>
+				<string>meta.fold.begin</string>
+				<key>end</key>
+				<string>meta.fold.end</string>
+			</dict>
+			<dict>
+				<key>begin</key>
+				<string>punctuation.section.mapping.begin</string>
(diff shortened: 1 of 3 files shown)


jfcherng (approved):

jrappen (approved):

jfcherng (dismissed):

jrappen:
@deathaxe Could you comment on why you chose to add a `meta.fold` scope for beginnings and ends of block comments but not objects and arrays? You could've added `punctuation.definition.comment.*` scopes directly as you've done with the others.

---

Found a typo:

image

deathaxe:
We've started to scope all asterisks `punctuation.definition.comment.[begin|end]` in several syntaxes' block comments. So directly using that as fold-`begin` and fold-`end` markers would cause

```jsonc
/******************************
 *
 ******************************/
```

to be folded to

```jsonc
/******************************...******************************/
```

Scoping only `/*` = `meta.fold.begin` and `*/` = `meta.fold.end` ensures to end up in

```jsonc
/*...*/
```

instead.

The cost is degrated experience with ligatures as those `meta.fold` parts have different scopes.

---

I hope it's self-explaining that `{` vs. `}` and `[` vs. `]` don't suffer this problem and don't therefore need another extra scope assigned.

---

I'd prefer an additional regexp pattern as alternative to `meta.fold`, to avoid this extra-scope (with a look at being limited to 64 scopes on stack, see https://github.com/sublimehq/sublime_text/issues/5283)

jrappen:
> We've started to scope all asterisks ...

Right, sorry I missed that. I'll take back that question 😬

jrappen:
How about scoping those extras in the first and last line like those in the middle like so:

```json
    /**
//  ^^ punctuation.definition.comment.begin.json
//    ^ punctuation.definition.comment.json
     *
//   ^ punctuation.definition.comment.json
    */
//  ^^ punctuation.definition.comment.end.json
```

deathaxe:
It would remove the need to add an extra `meta.fold`, yes. The ligature issue still persists though.

Scoping all asterisks after and infront of `/` the same way is done to ensure they are highlighted the same and ligatures are not applied. What we have now is

grafik

jrappen (approved):

deathaxe:
Syntax based folding was added via Fold.tmPreferences by ST4132. Hence closing.

jrappen:
Superseded by #3404.

PR #3292, opened 2022-03-26
[XML] Add support for syntax based folding

deathaxe:
Requires ST 4130+

This PR ...

1. disables indentation folding
2. introduces a common `meta.fold.[begin|end]` scope
3. adds required `Folding Rules.tmPreferences`
4. renames `punctuation.definition.group` into
   `punctuation.section.group` as this is the one used in most other
   syntaxes.

Why introducing `meta.fold`?

1. Syntax based folding depends on pairing tokens of certain scopes.
   Hence a syntax must carefully ensure to apply scopes which are used
   for folding to single tokens only. Otherwise unexpected regions are
   folded (see https://github.com/sublimehq/sublime_text/issues/5334).

2. XML uses quite common `punctuation.definition.tag.[begin|end]`
   scopes, which may sometimes not balanced in ways needed for folding

   The CDATA tag is one example. It consists of
   - 2 tokens of punctuation.definition.tag.begin and
   - 1 token of punctuation.definition.tag.end

   ```
   <![CDATA[ string ]]>
   ^^^ punctuation.definition.tag.begin
           ^ punctuation.definition.tag.begin
                    ^^^ punctuation.definition.tag.end
   ```

   So using punctuation.definition.tag for folding would end up in
   unexpected results.

   Due to (1.) we need to ensure to only scope the second `[` as fold
   start marker.

   ```
   <![CDATA[ string ]]>
   ^^^ punctuation.definition.tag.begin
           ^ meta.fold.begin punctuation.definition.tag.begin
                    ^^^ meta.fold.end punctuation.definition.tag.end
   ```

   ---

   Another example are XML tags. If both attributes and child nodes are
   to be folded, the closing punctuation of the opening tag needs to be
   both, the fold end marker for attributes and the fold begin marker
   for child nodes.

   ```
   <tag ... > ... </tag>
    ^^^ meta.fold.begin
            ^ meta.fold.end meta.fold.begin
                  ^^ meta.fold.end
   ```

3. Using arbitrary existing scopes in folding rules, would end up in
   broken folding behavior in inherited syntaxes for sure. By relying
   on a new scope, we can ensure backward compatibility and enforce
   3rd-party syntax authors to think about this topic.

4. By adding `meta.fold` it becomes quite clear in a syntax definition,
   which tokens are meant to be used as folding boundaries.

5. Folding support can be added without changing scopes in ways which
   might effect syntax highlighting.

6. Context bailouts can be taken into account more easily. See how
   illegal tag end tokens are included into folding regions, to
   hopefully avoid breaking folding too badly, in incomplete code
   scenarios.

### DIFF
--- a/XML/Folding Rules.tmPreferences
+++ b/XML/Folding Rules.tmPreferences
@@ -5,8 +5,6 @@
 	<string>text.xml</string>
 	<key>settings</key>
 	<dict>
-		<key>indentationFoldingEnabled</key>
-		<false/>
 		<key>foldScopes</key>
 		<array>
 			<dict>
--- a/XML/Folding Rules.tmPreferences
+++ b/XML/Folding Rules.tmPreferences
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>scope</key>
+	<string>text.xml</string>
+	<key>settings</key>
+	<dict>
+		<key>indentationFoldingEnabled</key>
+		<false/>
+		<key>foldScopes</key>
+		<array>
+			<dict>
+				<key>begin</key>
+				<string>meta.fold.begin</string>
+				<key>end</key>
+				<string>meta.fold.end</string>
+			</dict>
+			<dict>
+				<key>begin</key>
+				<string>punctuation.definition.string.begin</string>
+				<key>end</key>
+				<string>punctuation.definition.string.end</string>
+			</dict>
+			<dict>
+				<key>begin</key>
+				<string>punctuation.section.brackets.begin</string>
+				<key>end</key>
+				<string>punctuation.section.brackets.end</string>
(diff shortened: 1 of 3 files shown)


jrappen (approved):

jrappen:
Maybe disable indentation folding like in the JSON PR?

deathaxe:
For the answer to this question, please refer to commit message of 4ec49f152c4fc06afed5e076413c720192088451

jrappen:
I think you should do it anyway and third-party packages must follow. Otherwise this will be delayed forever.

---

Yeah, well, but that's just like my opinion, man. [↗️](https://www.youtube.com/watch?v=1vBesOFURek)

deathaxe:
It would just cause avoidable bug reports. It's not worth it especially as both indentation and syntax based folding work well together.

jrappen:
Superseded by #3404.

Issue #689, opened 2016-11-15
[C] Go to definition to enum value?


puetsua:
```
enum {
   VALUE_1 = 0,
   VALUE_2,
   VALUE_3
}
```
It seems 'Go to definition' is not working in this case in *.c *.h files. Any help?

wbond:
Currently enum values aren't indexed by the indexer, thus "Go to definition" doesn't work with them. This is probably a pretty reasonable addition, assuming it is a C enum and not a C++ enum class.

Are you familiar with syntax definitions for Sublime Text at all? If so, I can give you some pointers that may help in creating a PR to add this functionality.

speeder:
I have no idea what I am doing, but I am trying to do this. Where I edit the indexer?

hankofficer:
@wbond Thanks! However I have no idea how to add syntax definition for sublime text. ^^;
Is there anything I should look for or I should be aware of? I'd love to try it myself.

wbond:
It going to require some learning, but here are some resources:

 - The syntax format: http://www.sublimetext.com/docs/3/syntax.html
 - How symbols are "extracted": http://docs.sublimetext.info/en/latest/reference/symbols.html
 - Scope naming guidelines: http://www.sublimetext.com/docs/3/scope_naming.html
 - Code that will be affected:
   - https://github.com/sublimehq/Packages/tree/master/C%2B%2B
   - https://github.com/sublimehq/Packages/tree/master/Objective-C
   - C++ gets all changes from C
   - Objective-C gets all changes from C
   - Objective-C++ gets all changes from C++ and Objective-C
 - For any scope name changes, be sure to add syntax tests for all four syntaxes (see the last section of the Syntax documentation)

rwols:
There's a PR to address this here: https://github.com/sublimehq/Packages/pull/976

LinuxDevon:
PR #976 has been open for far too long and no response from original poster. I opened a new PR #3299  to address this issue.

PR #3299, opened 2022-04-01
[C] Index enumeration values for c / cpp

LinuxDevon:
This is to address this issue: #689.

The previous PR: #976 is dead at the moment. There isn't any response from the original poster. This branch is intended to address the comments in the PR.

**DISCLAIMER** I haven't worked with Objective-C but was request a port to that language as well in the original PR. I did my best and based on the tests looks exactly like its counter parts C/CPP.


keith-hall (approved):

C++/syntax_test_c.c
```diff
@@ -10,6 +10,10 @@ int main(){
 }
 
 enum Foo { kFoo, kBar };
+/* <- keyword.declaration */
+/*   ^ entity.name.enum */
+/*         ^ entity.name.enum.constants */
```
deathaxe, line 15: `entity.name.enum` should be reserved for the enum data type (e.g.: `Foo`).

I'd suggest `entity.name.constant` for `kFoo` and `kBar` as this is what D and Java use.

_C# and Rust scope them `constant.other`, but this should probably be changed._

C++/C.sublime-syntax
```diff
@@ -687,6 +687,14 @@ contexts:
       push: data-structures
     - include: expressions
 
+  data-structures-body-enum:
+    - include: preprocessor-data-structures
+    - match: '(?={{before_tag}})'
+      push: data-structures
+    - match: '{{identifier}}'
+      scope: entity.name.enum.constants
```
deathaxe, line 695:
```suggestion
      scope: entity.name.constant.c
```

C++/C++.sublime-syntax
```diff
@@ -333,6 +333,13 @@ contexts:
     - include: scope:source.c#label
     - include: expressions
 
+  statements-enum:
+    - include: preprocessor-statements
+    - include: scope:source.c#label
+    - match: '{{identifier}}'
+      scope: entity.name.enum.constants.c++
```
deathaxe, line 340:
```suggestion
      scope: entity.name.constant.c++
```

Objective-C/Objective-C++.sublime-syntax
```diff
@@ -117,6 +117,13 @@ contexts:
     - include: scope:source.c#label
     - include: expressions
 
+  statements-enum:
+    - include: preprocessor-statements
+    - include: scope:source.c#label
+    - match: '{{identifier}}'
+      scope: entity.name.enum.constants.objc++
```
deathaxe, line 124:
```suggestion
      scope: entity.name.constant.objc++
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -783,6 +783,14 @@ contexts:
       push: data-structures
     - include: expressions
 
+  data-structures-body-enum:
+    - include: preprocessor-data-structures
+    - match: '(?={{before_tag}})'
+      push: data-structures
+    - match: '{{identifier}}'
+      scope: entity.name.enum.constants
```
deathaxe, line 791:
```suggestion
      scope: entity.name.constant.objc
```

deathaxe (changes requested):
I am fine with this solution in general. I'd just suggest to change scope names of enum constants to get inline with D and Java.

_The general scope naming scheme is about scopes beginning with very general names such as `entity` or `constant` and each level of scope adding more detail up to the language they refer to._

Note: `entity.name.constant` should be added to `Symbol List.tmPreferences`. `Symbol Index - Include Constants.tmPreferences` can be removed then.

C++/syntax_test_c.c
LinuxDevon: Pulled in the suggested changes for the scope change. Updated the tests to reflect that as well. Thank you for the input! Let me know if there is anything else or something I missed

deathaxe: You may probably want to pick the tmPreferences from https://github.com/LinuxDevon/Packages/pull/1 so `entity.name.constant` is properly added to symbol list and index.

LinuxDevon:
@deathaxe I think I understood what you meant about the note with the `tmPreference` files. Double check me on that please. The Objective-C didn't match so I modified that to mirror the C/C++ one. Removed the `constant` file since these are now in the symbol file.

C++/syntax_test_c.c
LinuxDevon: Just updated the `tmPreference` files as well. I missed the comment earlier.

Objective-C/Symbol Index.tmPreferences
```diff
@@ -2,11 +2,13 @@
 <plist version="1.0">
 <dict>
 	<key>scope</key>
-	<string>source.objc, source.objc++</string>
```
deathaxe, line 5: This file I also stumbled over.

It is originally meant to do some symbol translations only. Its scopes overlap with the same file in C++ package and the translation rule conflicting with the other file. Haven't looked into it any further though.

But as _C++/Symbol List.tmPreferences_ already adds `entity.name.constant` to symbol list and index for `source.objc` and `source.objc++` your change here is not needed.

LinuxDevon: Interesting. I didn't see the `source.objc` and `source.objc++`. So to be clear, revert this file but keep the `Symbol Index - Include Constants.tmPreference` deleted? Seeing as how the `entity.name.constant` is already added via ` C++/Symbol List.tmPreferences`

LinuxDevon: Complete

deathaxe (approved):

keith-hall (approved):

rwols:
Someone should really just rewrite the C-family with inheritance and stuff like that 😬

deathaxe:
I had been about starting to, several times, but than other simpler to fix issues were found again and again.

PR #3304, opened 2022-04-02
[SQL] Many fixes and improvements overall

azizk:
Hey!

My main goal was to fix the way the word-boundary escape `\b` is used throughout the syntax file. I did much more besides that.
So I wanted, for example, `1IN` to be two tokens, and I assumed that every engine recognizes them as such, but I just tested with PSQL and MySQL and it looks like it's 2 tokens for PSQL and 1 for MySQL.

```sql
-- This is valid in PSQL and returns `true`. MySQL thinks it's an undefined function called "1IN".
SELECT 1IN(1,2,3);
```

I haven't looked at what the standard says yet. Regardless, I think my changes are pretty useful as they are and the thing above can be reverted by adding a `\b` back to the `match: \d+` context. My last commit is solely for formatting. Hope you like it! :+1:

### DIFF
--- a/SQL/SQL.sublime-syntax
+++ b/SQL/SQL.sublime-syntax
@@ -283,8 +283,7 @@ contexts:
       set: identifier-create
   drop-condition:
     - include: dml-condition
-    - match: (?=\S)
-      pop: true
+    - include: if-non-space-pop
   dml-condition:
     - match: (?i:if)\b
       scope: keyword.control.flow.sql
@@ -296,26 +295,22 @@ contexts:
     - match: \'
       scope: punctuation.definition.string.begin.sql
       set: [like-escape-fail, inside-like-single-quoted-string]
-    - match: (?=\S)
-      pop: true
+    - include: if-non-space-pop
   like-string-followed-by-escape-slash:
     - match: \'
       scope: punctuation.definition.string.begin.sql
       set: [like-escape-character-slash, like-escape-pop, inside-like-single-quoted-string-slash-escape]
-    - match: (?=\S)
-      pop: true
+    - include: if-non-space-pop
   like-string-followed-by-escape-caret:
     - match: \'
       scope: punctuation.definition.string.begin.sql
       set: [like-escape-character-caret, like-escape-pop, inside-like-single-quoted-string-caret-escape]
-    - match: (?=\S)
-      pop: true
+    - include: if-non-space-pop
   like-string-followed-by-unknown-escape:
     - match: \'
       scope: punctuation.definition.string.begin.sql
       set: [like-escape-character-any, like-escape-pop, inside-like-single-quoted-string]
-    - match: (?=\S)
-      pop: true
+    - include: if-non-space-pop
   inside-like-single-quoted-string-slash-escape:
     - meta_include_prototype: false
     - meta_scope: meta.string.like.sql string.quoted.single.sql
@@ -351,14 +346,12 @@ contexts:
   like-escape-fail:
     - match: (?i:escape)\b
       fail: like-strings-branch
-    - match: (?=\S)
-      pop: true
+    - include: if-non-space-pop
   like-escape-pop:
     - match: (?i:escape)\b
       scope: keyword.operator.word.sql
       pop: true
-    - match: (?=\S)
-      pop: true
+    - include: if-non-space-pop
   like-escape-character-any:
     - match: (\')([^'])(\')
       captures:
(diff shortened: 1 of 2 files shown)


deathaxe:
Many thanks to your contribution. I wonder whether we can incooperate any valuable changes into #3046 as it already comes a long with a lot of improvements especially with regards to syntaxes for different SQL dialects based on one "root-syntax".

azizk:
Oh dear, I should have done a search before doing this. :smile:
I'll check it out later...

jrappen:
Yes, as @deathaxe said, please take a look at @keith-hall 's #3046. As you can see, his branch for that PR is <https://github.com/forkeith/Packages/tree/sql> which you can use for a diff and/or possible PRs to target and be included in #3046.

azizk:
Thanks @jrappen! Yeah I checked it out and it's looking great. There are some things I'd do differently (the leading `\b`) and I could also contribute some more stuff, which isn't included in this PR (mostly PostgreSQL). Hey @keith-hall, if you're interested I could contribute and send a PR to your branch. :+1:

Nevertheless, do you all think it's still worthwhile to merge this in? I don't mind if it's overwritten later by Keith's work, but until then these changes may be useful and deployed to users earlier? :smiley:

SQL/SQL.sublime-syntax
```diff
@@ -8,10 +8,7 @@ file_extensions:
   - ddl
   - dml
 
-first_line_match: |-
-  (?xi:
-    ^ \s* -- .*? -\*- .*? \bsql\b .*? -\*-  # editorconfig
-  )
+first_line_match: (?xi) ^ \s* -- .*? -\*- .*? \bsql\b .*? -\*-  # editorconfig
```
keith-hall, line 11: arguably, keeping this on multiple lines is better for consistency with other syntaxes https://github.com/sublimehq/Packages/commit/3629e8bce1d25fbbc66528ed212f94597cbb8800

keith-hall:
a PR to my branch improving PostgreSQL would be very welcome :)
I'm not sure I agree with the blanket removal of leading `\b` though. It would make sense to do this for the PSQL variant, sure, but it'd be nice to keep it for the others, so even though merging this PR early could be beneficial for PSQL users, it may be a regression for all others.

azizk:
@keith-hall Alright, gonna send some commits then. :smiley:
So regarding the leading word anchor `\b`, afaics it's not necessary. It's only needed if you don't consume all the word characters (the way it is now), but if you do then using a leading `\b` everywhere is pretty redundant. Or maybe I'm overlooking something? And in case you want to recognise something like `1IN` as a single token, like I described in my PR comment, then adding a *trailing* `\b` to the number regexp is all there is to make it work.

SQL/SQL.sublime-syntax
```diff
@@ -9,9 +9,8 @@ file_extensions:
   - dml
 
 first_line_match: |-
-  (?xi:
-    ^ \s* -- .*? -\*- .*? \bsql\b .*? -\*-  # editorconfig
-  )
+  (?xi)
```
jrappen, line 12: please revert this change

azizk: Ok. Thought it's consistent with the rest of the formatting I did...

jrappen: You should never use this style for variables, as it'll have unintended consequences when using them.

azizk: You're right, for variables it's probably not the best. But `first_line_match` isn't really a variable, right?
Now that you're saying that, I'll have to check whether this is an issue in ElixirSyntax. Maybe I was lucky and didn't run into issues with `(?x)` in variables...

jrappen: Sure, then let me rephrase this: Explicit is better than implicit.

jrappen:
What's the current status with this PR and the request to add your changes to Keith's PR?

So far you have only renamed some stuff on Keith's branch and not added any meaningful changes from this PR to his.

azizk:
@jrappen From my end the PR here can go in. Keith is concerned about the leading `\b` however. I asked about it in my previous comment...
I know I only contributed minor things on his branch. I haven't committed my other progress yet. Just had other priorities in the meantime.

azizk:
I just rebased the branch. If the PR is still useful please merge, otherwise let's close this. I'm lacking the time and resources to continue with this...

azizk:
I'm not interested in this anymore. If you still want the changes, feel free to merge or cherry-pick them. :+1:

PR #3306, opened 2022-04-02
[HTML] Add meta.string to CDATA content

### LINKED ISSUE #3305: [HTML/XML] Unify CDATA scope tags
### What happened?

Can you make it easier for color schemes to target the open/close tags and the content of CDATA? HTML and XML have slightly divergent scopes, which leads to a rather ugly rule in my color defs.

deathaxe:
Fixes #3305

This commit adds `meta.string.html` to CDATA content so it uses the same scopes as CDATA in XML.

### DIFF
--- a/HTML/HTML (Plain).sublime-syntax
+++ b/HTML/HTML (Plain).sublime-syntax
@@ -110,7 +110,7 @@ contexts:
 
   cdata-content:
     - meta_scope: meta.tag.sgml.cdata.html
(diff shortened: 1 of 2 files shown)


michaelblyons:
Thanks! See update to #3305. I'm also hoping for punctuation consistency.

keith-hall (approved):

michaelblyons (approved):

PR #3312, opened 2022-04-04
[TypeScript] Implement generic function types

Thom1729:
There were no tests for the `match: (?=<)` /  `pop: 3` rule. I *think* it was there for cases like `true ? (a) : <foo />`. I've added a test for that, and a new fix compatible with generic function types, and then a new fix for that fix.

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -822,8 +822,6 @@ contexts:
             - ts-type-expression-begin
 
   ts-type-expression-begin:
-    - match: (?=<)
-      pop: 3
     - match: keyof{{identifier_break}}
       scope: keyword.operator.type.js
     - match: typeof{{identifier_break}}
@@ -861,6 +859,11 @@ contexts:
     - include: ts-type-special
     - include: ts-type-primitive
     - include: ts-type-basic
+
+    - match: (?=\<)
+      set:
+        - ts-type-function
+        - ts-type-parameter-list
 
     - match: (?=\()
       pop: true
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -540,6 +540,7 @@ contexts:
             - ts-type-parameter-default
             - ts-type-parameter-constraint
         - include: comma-separator
+        - include: else-pop
     - include: else-pop
 
   ts-type-parameter-constraint:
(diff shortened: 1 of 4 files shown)


JavaScript/TypeScript.sublime-syntax
```diff
@@ -530,16 +530,24 @@ contexts:
...
-        - include: comma-separator
+        - - meta_scope: meta.generic.js
+          - match: \>
+            scope: punctuation.definition.generic.end.js
+            pop: true
+
+          - match: ','
+            scope: punctuation.separator.comma
+            push: ts-type-parameter-list-item
+          - include: else-pop
```
deathaxe, line 541: May we try to avoid anonymous contexts, especially in multi-push/set statements?

deathaxe (commented):
Seems to require some conflicts to be resolved due to recent merge.

JavaScript/tests/syntax_test_tsx.tsx
```diff
@@ -203,12 +203,24 @@ if (a < b || c <= d) {}
...
-//  ^^^^^^^^^^^^^^ meta.function meta.generic
+    <T extends "s">() => {x}; // </T>;
+//  ^^^^^^^^^^^^^^^^^^^^^^^^ meta.function
+//  ^^^^^^^^^^^^^^^ meta.function meta.generic
 //   ^ variable.parameter.generic
 //     ^^^^^^^ storage.modifier.extends
-//             ^^ meta.function meta.generic meta.mapping
+//             ^^^ meta.function meta.generic
+
+    <T extends="s">() => {x}; // </T>;
```
deathaxe, line 213: This line looks broken. Neither `()` nor `=>` have a scope assigned.

grafik

Thom1729: It is expected. The entirety of `() => {x}; // ` is JSX content, not a function. I've added a negative test to help clarify.

Thom1729: It's not surprising that it looks broken, because the test is supposed to look (to the parser) as much like an arrow function as possible without actually being one. I spent a lot of time in AST Explorer trying to find a minimally-differing pair of test cases (lines 206/213).

Thom1729: (Also, there should arguably be some kind of scope for JSX text that we could use for a positive test here, but currently there isn't. Worth considering for a future PR.)

deathaxe (approved):

michaelblyons (approved):

PR #3328, opened 2022-04-13
[Rails] Set default_extension in sublime-settings

themilkman:
Set the default file extension to "rb" for new "Ruby (Rails)" files.

See https://github.com/sublimehq/sublime_text/issues/5373 for details.

The idea comes from @Ultra-Instinct-05 and @jfcherng .

Thanks for any feedback.

### DIFF
--- a/Rails/Ruby (Rails).sublime-settings
+++ b/Rails/Ruby (Rails).sublime-settings
@@ -1,0 +1,5 @@
+{
+  // This gives precedence over the first file_extensions entry in the
+  // "Ruby (Rails).sublime-syntax" file.
+  "default_extension": "rb"
+}


deathaxe:
Applying file extensions via sublime-settings in shipped packages makes it harder to assign new ones. XML used to do something like that with some negative side effects for "Always Open with ..." functionality or if a 3rd-party package wants to specify its own bindings for certain file types, which are already listed in a Default Package.

Therefore extensions should only be assigned via `file_extensions` key in a sublime-syntax file.

Assigning .rb to Ruby (Rails) may not cause syntax errors for normal Ruby users, but Ruby (Rails)  being displayed in the status bar might be unexpected.

It shouldn't be too much of an issue for Rails devs to assign .rb to Ruby (Rails) on the users end, once via syntax specific setting in User package.

jfcherng:
> Assigning .rb to Ruby (Rails) may not cause syntax errors for normal Ruby users, but Ruby (Rails) being displayed in the status bar might be unexpected.

That is not what this PR will do. This PR prioritizes `.rb` as the default save file extension for Ruby (Rails) files by syntax-specific settings of Ruby (Rails).

Rails/Ruby (Rails).sublime-settings
```diff
@@ -0,0 +1,5 @@
+{
+  // This gives precedence over the first file_extensions entry in the
+  // "Ruby (Rails).sublime-syntax" file.
+  "default_extension": ["rb"]
```
jfcherng, line 4: This seems should be `"default_extension": "rb"` ?

deathaxe:
I see. Just found a `ruby.rail` in `Ruby.sublime-syntax`. Not sure how common it is, but it would feel better placed in Ruby (Rails).

Ultra-Instinct-05:
> Applying file extensions via sublime-settings in shipped packages makes it harder to assign new ones. XML used to do something like that with some negative side effects for "Always Open with ..." functionality or if a 3rd-party package wants to specify its own bindings for certain file types, which are already listed in a Default Package.

I believe you are talking about the `hidden_extensions` setting, which is different from `default_extension`. See https://github.com/sublimehq/sublime_text/issues/1326#issuecomment-284475601

Rails/Ruby (Rails).sublime-settings
themilkman: Oh my... too many things around, sorry. Force pushed the correct version.

themilkman:
> I see. Just found a `ruby.rail` in `Ruby.sublime-syntax`. Not sure how common it is, but it would feel better placed in Ruby (Rails).

Sounds correct to move it away (as you did), although never heard of it during the last 10 years of Ruby and Ruby on Rails development. :+1:

jfcherng (approved):

deathaxe (approved):

PR #3331, opened 2022-04-15
[PHP] Refactor Syntax

### LINKED ISSUE #625: [PHP] Apply a scope as `entity.name.variable.php` for `MY_CONSTANT` at the `define('MY_CONSTANT', 'HELLO')` variable creation function call, for symbols navigation
### [PHP] Apply a scope as `entity.name.variable.php` for `MY_CONSTANT` at the `define('MY_CONSTANT', 'HELLO')` variable creation function call, for symbols navigation

This is to allow locote those symbol within the PHP's `Symbol List.tmPreferences` file. This problem was originally discussed here: https://forum.sublimetext.com/t/why-use-ctags-while-there-is-the-built-in-sublime-text-symbols/22917/7

After this enhancement, the file would need look like this, containing the `entity.name.variable.php` as symbol selector:

``` json
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
    <key>name</key>
    <string>Symbols List: Indent Functions</string>
    <key>scope</key>
    <string>entity.name.function.php, support.function.magic.php, entity.name.class.php, entity.name.interface.php, entity.name.trait.php, entity.name.variable.php</string>
    <key>settings</key>
    <dict>
        <key>showInSymbolList</key>
        <integer>1</integer>
        <key>showInIndexedSymbolList</key>
        <integer>1</integer>
    </dict>
</dict>
</plist>
```

### LINKED ISSUE #643: Bad PHP syntax highlighting inside a HTML attribute's double quotes

### LINKED ISSUE #1168: Highlights not working in PHP within Javascript code
Hi,
I encountered a problem with a PHP file with Javascript instrucions in the <HEAD> part.
When, inside a Javascript funcion, I insert a PHP instrucion, the highligh simply stop to work corretly.

This is the line when It starts to appear:
imagen

Following the code, this is how a Javascrpit funcion is highlighted:
imagen

Here you find the full file (I cutted a little bit), so you can reproduce the problem.
Thanks!
[bok_pre_edit_test_bug.zip](https://github.com/sublimehq/Packages/files/1326450/bok_pre_edit_test_bug.zip)

### LINKED ISSUE #1665: Error loading php syntax
On sublime 3.1.1 build 3176

Without changing, or adding any plugin, suddenly sublime starts complaining this:

`Error loading syntax file "Packages/PHP/PHP.sublime-syntax": Apparent recursion within a with_prototype action: 25000 context sanity limit hit`

### LINKED ISSUE #2459: [PHP] Improve variable scope
The second dollar sign should be in the same scope as the first one or in a different scope from the string name.

The second dollar sign and string scopes are the same: `embedding.php text.html.basic meta.embedded.block.php source.php punctuation.definition.variable.php variable.other.class.php`

php-scope-issue

- Sublime Version: 4
- OS Version: 4084

See for example, Github:
``` php
$timber              = new \Timber\Timber();
$timber::$dirname    = array( 'views' );
$timber::$autoescape = false;

self::$somevar
```

### LINKED ISSUE #2625: [PHP] closure scope does not cover body
I was reviewing the syntax highlighting for PHP and noticed that the meta_scope `meta.function.closure.php` ends before `use` as shown below. Shouldn't it scope all the closure's body similar to `meta.function.php`?

- Sublime Version: 3211
- OS Version: Windows 10

``` php
<?php

class Test
{
    public function __construct($foo)
// ^-----> meta.function.php starts here
    {
        $this->query(function ($bar) use ($foo, $baz) {
                  // ^-------------^-----> meta.function.closure.php only covers these
            return $foo;
        }, $zzz);
    // ^-----> I expected meta.function.closure.php to end here
    
        return $bar;
    }
// ^-----> meta.function.php ends here, scoping the whole function
}
```

### LINKED ISSUE #3214: [PHP] End of indented heredoc not recognized if content ends with $variable
### What happened?

In `PHP Source.sublime-syntax` there is an interference between `heredoc` and `variables`/`- match: (\$+){{identifier}}` that results in wrong highlighting.

Minimal example:
```php
<?php
	<<<END
	$variable
	END;
	This line is recognized as a string.
```

Result:
grafik

Expected result:
grafik

### LINKED ISSUE #3375: [PHP] The SQL keywords `WITH` is not recognized a valid SQL.
### What happened?

The string is not formatted as SQL 

```
<?php

$sql = "WITH RECURSIVE cte as (
            SELECT 1 as x
            UNION ALL
            SELECT x+1 FROM cte WHERE x<10)
        SELECT * FROM cte;";
```

When using a string that starts with `SELECT` then syntax coloring is working, as expected, but with the keyword `WITH` it stops working.

        
This bug-report was caused by this (stackoverflow)question: [How to make PHP handeling my "WITH CTE as" SQL?](https://stackoverflow.com/questions/72158928/how-to-make-php-handeling-my-with-cte-as-sql)

deathaxe:
Fixes #625
Fixes #643
Fixes #1168
Fixes #1665
Fixes #2459
Fixes #2625
Fixes #3214
Fixes #3375

Supersedes #2789

This PR suggests a major overhaul of PHP syntax to 
1. solve known issues, such as huge syntax cache and compile times
2. improve maintainability and readability
3. apply some design patterns with regards to scope naming or syntax architecture

## Major Changes

- converts most contexts to named contexts
  - for readability reasons
  - to support inheritance

- organizes all rules into
  - declarations (classes, functions, ...)
  - statements (if, else, while, ...)
  - expressions (arithmetic, function-calls, ...)

- replaces _(nearly)_ all `push ... with_prototype` constructs
  - to significantly reduce syntax cache size
  - gain more control over string interpolation scenarios `meta.string meta.interpolation`
  - significantly reduce risk of exceeding context sanity limit (see: #1665)

- applies some scope changes to more closely follow recently discussed scope naming

## Benchmarks

Attribute | ST4131 | this PR | Diff
---       | ---    | ---     | ---
Syntax Cache Size | 38.050kB  | 7.481kB | -30.569kB (-80%)
Syntax Compile Time | 2,8s | 0,6s | -2,2ms (-78,6%)
Syntax Test -- Performance [syntax_test_php.php][syntax_test_php] | 11,2ms | 7,9ms | -3.3ms (-29%)
Syntax Test -- Performance [PHPUnit/composer.php][composer] | 557ms | 376ms | -181ms (-32%)

##### Syntax Cache Size

1. Open vanilla setup of ST4131 (cache folder cleared, no session, no 3rd-party packages)
2. Open [syntax_test_php.php][syntax_test_php]
3. determine folder size of _Data/Cache/PHP_

##### Syntax Compile Time

1. Prepare screen capturing (ScreenToGif)
2. Open vanilla setup of ST4131 (cache folder cleared, no session, no 3rd-party packages)
3. Open [syntax_test_php.php][syntax_test_php]
4. analyse screen capturing taken from clicking on Open File dialog's OK button until file content is displayed.

[composer]: https://github.com/sebastianbergmann/phpunit/blob/main/tools/composer
[syntax_test_php]: https://github.com/sublimehq/Packages/blob/master/PHP/tests/syntax_test_php.php


jwortmann:
Here are two more possible improvements worth to consider:

1. [Attributes](https://www.php.net/manual/en/language.attributes.overview.php) in PHP currently use uncommon scopes `meta.attribute` and `punctuation.definition.attribute`:

    ```php
    <?php
        #[Attribute('foo')]
    //  ^^^^^^^^^^^^^^^^^^^ meta.attribute
    //  ^^ punctuation.definition.attribute.begin
    //    ^^^^^^^^^ support.class
    //                    ^ punctuation.definition.attribute.end
    ?>
    ```

    But they should better be `meta.annotation`, `punctuation.definition.annotation` and possibly `variable.annotation`, to reduce scope inconsistencies between syntaxes:

    ```php
    <?php
        #[Attribute('foo')]
    //  ^^^^^^^^^^^^^^^^^^^ meta.annotation
    //  ^^ punctuation.definition.annotation.begin
    //    ^^^^^^^^^ variable.annotation (or maybe keep support.class here?)
    //                    ^ punctuation.definition.annotation.end
    ?>
    ```

    Unbenannt

    Also Compare with C#, which has so called [attributes](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/) too and with a similar syntax like in PHP:

    https://github.com/sublimehq/Packages/blob/2a94a919348bfff3567fb585450a1eef4883e3a7/C%23/tests/syntax_test_GeneralStructure.cs#L726-L728

2. PHP is a syntax where it is still impossible to distinguish between operator keywords and operator symbols, because they use the exact same scope name:

    ```php
    <?php 
        if ($a && $b || !$c and $d or $e);
    //         ^^ keyword.operator.logical.php
    //               ^^ keyword.operator.logical.php
    //                      ^^^ keyword.operator.logical.php
    //                             ^^ keyword.operator.logical.php
    ?>
    ```

    There is a high demand (from myself :wink:) to make it possible in color schemes to apply different highlighting colors to them. I'm aware that there is/was a certain aversion against the `keyword.operator.word` scope brought up in another PR, and an argument was made in https://github.com/sublimehq/Packages/pull/2090#issuecomment-531297832 that this scope was only meant to be used for operators with *lower precedence* than other operators. But exactly this would apply to the `and`, `or`, ... operator keywords in PHP, see https://www.php.net/manual/en/language.operators.precedence.php. I would also be fine with other possible solutions to this problem, though.

deathaxe:
Annotations are on the roadmap already, just wanted to pre-publish with changes to syntax tests kept low to proof not to have broken too much by refactoring.

deathaxe:
Attributes are classes in C#, Java and PHP. As the former ones already use `variable.annotation` lets follow them and the scope naming guidelines.

deathaxe:
https://github.com/sublimehq/Packages/pull/3354#issuecomment-1116289745

> Will PHP also move to something similar ? Currently, it also has a super huge list of builtin functions. I am assuming that also would take similar maintenance efforts of finding out which are new and which might be deprecated.

Removing those huge lists of predefined constants/classes/functions would reduce PHP's cache size from 1050kB to 400kb with further round about 10-20% performance gain.

As not being involved in maintaining PHP's builtin constants/classes/functions I have no strong objections about keeping or removing them. I can't say anything about their completeness, also.

Missing items probably bother only few users as they are well highlighted in any case, while CSS doesn't highlight missing predefined property names at all, atm.

We could probably split the syntax into a `PHP (Plain)` and a `PHP` to provide a syntax without builtins treated special for those who don't care, but it wouldn't probably be used at all if not assigned by default.

Ultra-Instinct-05:
I still appreciate the fact that PHP at least tells builtin functions v.s. user defined functions apart without the need of semantic highlighting or LSP capabilities, but I agree it can be cumbersome to maintain.

deathaxe:
Let's start review.

jfcherng:
I have a very weird question.

image

The color of `embedding.php text.html.php meta.embedded.php source.php.embedded.html meta.namespace.php keyword.declaration.namespace.php` is controlled by 
```js
        {
            "font_style": "italic",
            "foreground": "blue",
            "name": "Storage type",
            "scope": "storage.type",
        },
```
rather than 
```js
        {
            "font_style": "bold",
            "foreground": "red",
            "name": "Keyword",
            "scope": "keyword",
        },
```

jrappen:
Was there a reason for not moving those function name word lists to variables in the `PHP Source.sublime-syntax`?

jrappen:
When using regexes where case is ignored you usually:

```yaml
- match: (?i:lowercase)
  scope: whatever
```

but with `constant.*`s you:

```yaml
- match: (?i:UPPERCASE)
  scope: constant.***
```

which was unexpected. Was there a reason?

jrappen (dismissed):

jrappen:
In general this looks fine. Thanks for the effort with named contexts.

deathaxe:
> Was there a reason for not moving those function name word lists to variables

It would have ended up in dozens of variables, used only once. Correctly formatting them would also add hundreds of lines ending up in a huge wall of text at the beginning of the syntax file. Neither found it useful nor worth the effort.

It may be more useful to add a dedicated context with patterns of constants/classes/functions for each extension. Something like https://github.com/sublimehq/Packages/pull/2572, but organized at the end of the `PHP Source` syntax.

I have a local commit which prepares that strategy by moving all the `support-...` contexts to the end of the file and introducing a structure like https://github.com/deathaxe/sublime-packages/commit/fdaa3a8426f6b0e0e966a95b2dda7a754b5fa789. We could add it as a template or leave that to the contributor who wants to re-organize those patterns.

---

Tokens in patterns such as `(?i:lower)` and `(?i:UPPER)` generally follow the casing those keywords would generally use. Keywords, variables, ... are normally used lower-case, while constants are written upper-case, even though php interprets source case-insensitive.

As in all/most other syntax definition the choise of case just reflects the natural casing used in a language - often defined by style guides.

jrappen:
> ... variables, used only once.

Fair enough.

The strategy you mention for moving parts of the syntax seems slightly overkill in my eyes, maintainability would improve though, as far as I can tell.

> ... the choice of case just reflects the natural casing used ...

I guess I wanted to say that I was interested if there was some kind of performance improvement. I had [this stackoverflow question's](https://stackoverflow.com/questions/11227809/why-is-processing-a-sorted-array-faster-than-processing-an-unsorted-array) top answer in mind.

deathaxe:
Not sure what casing is related with branch prediction to, but I'd argument the choosen case to be irrelevant for performance because those patterns are compiled to regexp binary format which will end up in something like `[Uu][Pp][Pp][Ee]...` anyway.

jrappen:
FYI, former contributors to the PHP syntax for Sublime Text:

@bakerkretzmar @husanjun @CameronHall @DavidW1024 @idleberg @sptndc @exploitmik @cixtor @taylorotwell @bmatt468 @gwarnants @duncan3dc @ahsankhatri @Briles @akalongman @vagkaefer @henriquemoody

If you have some time, maybe you could test-drive this and leave a review. Thanks!

jrappen (dismissed):

jrappen:
The opening comment (commit message for the PR) misses some info from the latest commits here, i.e. `addresses #3374`.

deathaxe:
#3374 is a PR! If you mean #3375, its there.

PHP/PHP Source.sublime-syntax
```diff
@@ -811,748 +795,839 @@ contexts:
...
+    - include: constants
+
+  use-statement-const-alias:
+    - match: '{{guarded_identifier}}'
+      scope: entity.name.constant.php
+      pop: 1
+    - include: attributes
+    - include: comments
+    - include: else-pop
+
```
bakerkretzmar, line 1360:
```suggestion
```

PHP/PHP Source.sublime-syntax
```diff
@@ -811,748 +795,839 @@ contexts:
...
+  use-statement-function-alias:
+    - match: '{{guarded_identifier}}'
+      scope: entity.name.function.php
+      pop: 1
+    - include: attributes
+    - include: comments
+    - include: else-pop
+
+
```
bakerkretzmar, line 1412:
```suggestion
```

PHP/PHP Source.sublime-syntax
```diff
@@ -1567,126 +1642,444 @@ contexts:
...
+  inherited-user-class-name:
+    - match: '{{identifier}}'
+      scope: entity.other.inherited-class.php
+      pop: 1
+
+###[ CONSTANTS ]##############################################################
+
+  constants:
+    - match: (?i:TRUE|FALSE)\b
```
bakerkretzmar, line 1986: These are way more commonly lowercase.

```suggestion
    - match: (?i:true|false)\b
```

PHP/PHP Source.sublime-syntax
```diff
@@ -1567,126 +1642,444 @@ contexts:
...
+      scope: constant.language.boolean.php
+    - match: (?i:NULL)\b
```
bakerkretzmar, line 1988: Same here.

```suggestion
    - match: (?i:null)\b
```

PHP/PHP Source.sublime-syntax
```diff
@@ -2016,18 +2569,214 @@ contexts:
...
+
+  ternary-expression:
+    - match: ':(?!:)'
+      scope: keyword.operator.ternary.php
+      pop: 1
+    - include: nested-expressions
+
+  variadic-operators:
+    - match: \.\.\.
+      scope: keyword.operator.variadic.php
```
bakerkretzmar, line 2724: Not sure how to do this or if you already did, but this syntax is also used for first-class callables in PHP 8.1+, so it can appear in other contexts now: https://www.php.net/manual/en/functions.first_class_callable_syntax.php.

Maybe that should just be added as an additional scope name?

PHP/PHP Source.sublime-syntax
```diff
@@ -811,748 +795,839 @@ contexts:
...
+    - match: (?i:const)\b
+      scope: keyword.declaration.constant.php
+      set: use-statement-const
+    - match: (?i:function)\b
+      scope: keyword.declaration.function.php
+      set: use-statement-function
+    - match: (?=\S)
+      set: use-statement-ns-class
+
+
```
bakerkretzmar, line 1328:
```suggestion
```

bakerkretzmar (commented):
Most of this is way over my head but it looks great, thanks! Left some formatting comments and a couple questions.

I'm also wondering based on the changes to the tests and scope names... is this going to break lots of syntax highlighting / themes? Fine if it will if it's worth it, just wondering what to expect 😅

PHP/PHP Source.sublime-syntax
```diff
@@ -1,698 +1,698 @@
...
+    | {{namespace_modifier}}
     )
 
-  keywords_declaration: |-
+  # Statement Level Keywords
+  declaration_keywords: |-
+    (?xi: class | const | declare | enddeclare | interface | namespace | trait | use | var )\b
+  control_keywords: |-
+    (?xi: case | catch | default | do | (?: end )? (?: for(?:each)? | if | switch | while )
+    | else | elseif | finally | switch | try )\b
```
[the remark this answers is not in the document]
jfcherng: and `enum` for `declaration_keywords` btw.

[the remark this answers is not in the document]
deathaxe: All `statement level keywords` pop `expressions` context off stack (terminate it).

1. `match` is used like a function-call which returns a value and therefore looks like being valid in the middle of an expression. Adding it to `control_keywords` would cause any expression to be popped off stack, which seems wrong with the way it works.

```
return match { ... }
```

also valid

```
$var = "true";

if ( match($var) { "true" => true, "false" => false } ) { echo "match returned true"; } else { echo "or no "; }
```

3. `enum` is not a reserved keyword (according to https://www.php.net/manual/en/reserved.keywords.php). PHP treats it as ordinary constant in expressions. Hence `enum` also shouldn't pop the `expressions` syntax off stack, too.

PHP/PHP Source.sublime-syntax
```diff
@@ -1,698 +1,698 @@
...
+  flow_keywords: |-
+    (?xi: break | continue | die | exit | finally | goto | return | throw | yield )\b
+
+  # Expression Level Keywords
+  builtin_functions: |-
+    (?xi: echo | print | empty | eval | isset | list | unset )\b
+  builtin_variables: |-
+    (?xi: parent | self )\b
```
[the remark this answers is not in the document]
deathaxe: `static` is used as both a type modifier and a variable for late bindings, depending on context. Therefore it needs to be treated special, while `builtin_variables` is applied globally.

PHP/PHP Source.sublime-syntax
deathaxe: Those duplicated empty lines are used to group the different types of contexts. They exist by intent

PHP/PHP Source.sublime-syntax
deathaxe: Fixed

PHP/PHP Source.sublime-syntax
deathaxe: Fixed

PHP/PHP Source.sublime-syntax
deathaxe: It's not explicitly implemented but those constructs seem properly handled, except `...` being scoped `variadic`.

`->myMethod` is scoped as ordinary function-call, not sure it should receive any special treatment.

PHP/PHP Source.sublime-syntax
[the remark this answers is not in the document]
bakerkretzmar: Gotcha, makes sense! Like I said, way over my head 😂

PHP/PHP Source.sublime-syntax
bakerkretzmar: My bad 👍🏻

PHP/PHP Source.sublime-syntax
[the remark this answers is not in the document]
deathaxe: Using the syntax in everyday projects and reporting back any kinds of annoyances etc. would be appriciated.

Ultra-Instinct-05:
I am on the `refactor-php` branch. Looks like HTML is no longer highlighted. This problem doesn't occur on the `master` branch

image

jfcherng:
> I am on the `refactor-php` branch. Looks like HTML is no longer highlighted. This problem doesn't occur on the `master` branch
> 
> image

Work on my side. Maybe manually clean the cache?

Ultra-Instinct-05:
Cleared the cache & restarted, but still doesn't seem to highlight.

deathaxe:
You also need latest HTML package from current master (or this branch). The one with the dedicated `html` context in HTML (Plain).sublime-syntax.

Ultra-Instinct-05:
Does it make sense to scope format specifiers within strings https://www.php.net/manual/en/function.printf.php ? Currently, they are not highlighted, but I believe they are only valid within `printf()`. Would still be a nice to have.

jfcherng (approved):

jrappen (approved):

michaelblyons (approved):

jfcherng:
Thanks for your hard work. 😄

PR #3333, opened 2022-04-15
[Haskell] Add HSX support

deathaxe:
This PR proposes to add support for `[hsx|...|]` quasi-quotes in Haskell.

HSX supports XML content, but this commit embedds HTML instead as it is
what other editors seem to use. It's probably the most common use case
for Haskell Web Devs.

see: https://hackage.haskell.org/package/hsx

### DIFF
--- a/Haskell/Embeddings/HTML (for HSX).sublime-syntax
+++ b/Haskell/Embeddings/HTML (for HSX).sublime-syntax
@@ -1,0 +1,47 @@
+%YAML 1.2
+---
+scope: text.html.embedded.haskell
+version: 2
+hidden: true
+
+extends: Packages/HTML/HTML.sublime-syntax
+
+contexts:
+
+  prototype:
+    - meta_prepend: true
+    - include: haskell-interpolations
+
+  cdata-content:
+    - meta_prepend: true
+    - meta_include_prototype: false
+    - include: haskell-string-interpolations
+
+  tag-attribute-value-content:
+    - meta_prepend: true
+    - include: haskell-string-interpolations
+
+  strings-common-content:
+    - meta_prepend: true
+    - include: haskell-string-interpolations
+
+  haskell-string-interpolations:
+    - meta_include_prototype: false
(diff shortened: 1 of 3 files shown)


michaelblyons (approved):

Haskell/Embeddings/HTML (for HSX).sublime-syntax
```diff
@@ -0,0 +1,47 @@
+%YAML 1.2
+---
+scope: text.html.embedded.haskell
+version: 2
+hidden: true
+
+extends: Packages/HTML/HTML.sublime-syntax
```
FichteFoll, line 7: Can you add a comment here or at the top of the file with a short summary for why we include HTML instead of XML here? I believe that would be a lot easier to find than the commit message.

Haskell/Haskell.sublime-syntax
```diff
@@ -1254,6 +1255,19 @@ contexts:
       scope: meta.quoted.quasi.haskell punctuation.section.quoted.end.haskell
       pop: 1
 
+  hsx-quotes:
+    - match: (\[)(hsx)(\|)
+      scope: meta.quoted.quasi.haskell
```
FichteFoll, line 1260: I presume you didn't chose `meta.string` here on purpose? Do we know of any other occurances of `meta.quoted` in the current code base?

deathaxe: > Quasiquoting allows programmers to use custom, domain-specific syntax to construct fragments of their program. Along with Haskell's existing support for domain specific languages, you are now free to use new syntactic forms for your EDSLs. 
> see: https://wiki.haskell.org/Quasiquotation

That description doesn't read like quasi quotes are plain strings. They rather seem to have a special meaning. Hence special meta scope was choosen when rewriting Haskell and so is for hsx.

FichteFoll (approved):

michaelblyons (approved):

PR #3339, opened 2022-04-16
[C#] Rename declaration keywords scopes

deathaxe:
This PR suggests to rename scope of special declaration keywords from `storage.type` to  `keyword.declaration`.

_This change has been made to several syntaxes in the past. C# is one of the remaining syntaxes, which have not been updated._

Note: Maybe scopes of more keywords such as `var` or `dynamic` should be renamed as well.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -294,27 +294,27 @@ contexts:
       scope: storage.modifier.access.cs
     - match: \b(class)\s+({{name}})
       captures:
-        1: storage.type.class.cs
+        1: keyword.declaration.class.cs
         2: entity.name.class.cs
       push: [class_signature, data_type_signature]
     - match: (?:\b(readonly)\s+)?\b(record)\s+(?:(struct)\s+)?({{name}})
       captures:
         1: storage.modifier.cs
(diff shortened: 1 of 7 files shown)


michaelblyons (approved):

jfcherng (approved):

keith-hall (approved):

keith-hall:
> Note: Maybe scopes of more keywords such as `var` or `dynamic` should be renamed as well.

As `let` and `var` in JS are scoped as `keyword.declaration`, then it could make sense to scope `var` in C# the same. `dynamic` is more of a type though, so I don't think it should be re-scoped, but happy to hear some arguments.

keith-hall:
Actually it could make sense for `dynamic` also when it's used where `var` can also be used, but imo it would feel weird when its a parameter to a function... `void Foo(dynamic bar)`

deathaxe:
I share your feelings about that. Maybe we can come up with a guideline like "keyword.declaration is used to declare new entites, such as functions or structured types (class/struct/interface/...)". So keywords like `var` which just start variable definitions would generally receive `storage.type` even in dynamically typed syntaxes.

PR #3351, opened 2022-04-22
[Erlang] Simplify comment punctuation scope

deathaxe:
This commit removes `.percentage` sub scope as no other language uses this level of detail.

### DIFF
--- a/Erlang/Erlang.sublime-syntax
+++ b/Erlang/Erlang.sublime-syntax
@@ -236,7 +236,7 @@ contexts:
 
   comments:
     - match: \%+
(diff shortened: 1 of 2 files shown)


michaelblyons:
Do we still have a scope for "units?" I kind of feel like `%` is a unit.

deathaxe:
This PR just shortans comment punctuation scope a bit. Not sure how it is related with units.

michaelblyons (approved):

jrappen (approved):

jfcherng (approved):

PR #3354, opened 2022-04-23
[CSS] Remove huge lists of property names and values 

### LINKED ISSUE #3342: [CSS] Missed some new properties
### What happened?

The issue is here, partly it';s related to ST4 too.
https://github.com/atom/language-css/issues/194

The snippet is
```
div {
	/* Non-experimental properties */
	border-block: initial;
	border-block-color: initial;
	border-block-style: initial;
	border-block-width: initial;
	border-inline: initial;
	border-inline-color: initial;
	border-inline-style: initial;
	border-inline-width: initial;
	color-scheme: initial;
	content-visibility: initial;
	counter-set: initial;
	forced-color-adjust: initial;
	offset-rotate: initial;
	print-color-adjust: initial;
	text-decoration-thickness: initial;
	text-underline-offset: initial;

	/* Experimental properties */
	inherits: initial;
	math-style: initial;
	overflow-clip-margin: initial;
	syntax: initial;

	/* Behind experimantal flag */
	align-tracks: initial;
	animation-timeline: initial;
	justify-tracks: initial;
	line-height-step: initial;

	/* Not implemented in any browser */
	margin-trim: initial;
	masonry-auto-flow: initial;

	/* Deprecated */
	viewport-fit: initial;

	/* No MDN docs */ 
	block-overflow: initial; /* I think this has been renamed to `overflow-block` */
	input-security: initial;

	/* svg */
	vector-effect: initial; /* no compatibility table on MDN, but works on chrome */
}
```

deathaxe:
Fixes #3342

This PR removes all those huge lists of generic/unsorted property names and values. It keeps only those properties, which are required to maintain valid syntax highlighting (see font, font-family).

Reasons:

1. It is not the scope of a syntax to do linting!
   LSP and LSP-css do a pretty well job in marking invalid properties.

   Each of `unsorted_property_constants` has been highlighted as valid
   constant in each property, which is wrong anyway. Hence current
   value of linting by syntax is doubtful anyway.

   It causes more trouble than value, IMHO.

   Completions can be used to find valid properties and their values.
   It's already hard enough to keep those up-to-date.

2. Maintenance effort is just too high to keep up with fast moving
   CSS standard. We are frequently missing new entries causing users
   to be disappointed.

3. Current lists of property names and constants are already heavily out of date!
   Adding all known props from MDN would cause them to grow by 2x or 3x
   which in turn causes syntax to grow heavily and parsing time to
   increase significantly.

   As CSS is used in conjunction with other sophisticated syntaxes like
   HTML, PHP, ... it might become problematic at some point.

TODO: We might want to find more general scopes for property names and constants (e.g.: `support.constants` to `constant.other`).

_Disclaimer: I am aware of past discussions about this topic and the attitute to keep this distinction. If we want to keep those lists someone else should overtake maintenance._


CSS/CSS.sublime-syntax
```diff
@@ -2724,19 +2447,43 @@ contexts:
     - match: \b(ltr|rtl){{break}}
       scope: support.constant.property-value.css
 
+  font-dispaly-constants:
```
michaelblyons, line 2450: typo.

jfcherng (approved):

jfcherng (changes requested):
Sorry. I just click approveon a wrong PR.

jrappen (approved):

jrappen:
Appreciate the effort with named contexts and links to documentation.

---

Did you time the effect on the difference of parsing time? Might help to convince more reviewers.

jrappen:
Inheritance might work?

Looking at <https://www.w3.org/Style/CSS/> there seems to be a yearly snapshot.

---

I'm not a CSS guy, though. Maybe I should stop posting random ideas.

deathaxe:
If someone wants to manually work through those documents regularly to catch up changes and update syntax patterns of supported properties and values, It would be fine as well. I don't want to.

Ultra-Instinct-05:
Will PHP also move to something similar ? Currently, it also has a super huge list of builtin functions. I am assuming that also would take similar maintenance efforts of finding out which are new and which might be deprecated.

jrappen (approved):

jfcherng (dismissed):

jrappen:
@michaelblyons could you re-review?

CSS/CSS.sublime-syntax
```diff
@@ -1502,6 +1198,7 @@ contexts:
   other-property:
     # Note: Consume unknown identifiers to maintain word boundaries.
```
FichteFoll, line 1199: Comment is not accurate anymore since we now actively consume remaining properties to assign a scope.

CSS/syntax_test_css.css
```diff
@@ -2082,9 +2082,9 @@
 /*              ^ meta.property-list.css - meta.property-name - meta.property-value */
 /*               ^^^^^^^^^^ meta.property-list.css meta.property-value.css */
 /*                         ^ meta.property-list.css - meta.property-name - meta.property-value */
-/*  ^^^^^^^^^^^^ - entity - support */
+/*  ^^^^^^^^^^^^ support.type.property-name.css */
 /*              ^ punctuation.separator.key-value.css */
-/*                ^^^^^^^^^ constant.other.css - support */
+/*                ^^^^^^^^^ support.constant.property-value.css */
```
FichteFoll, line 2087: Can you also add an arbitrary/custom property match without any part being known or do we already have that? From the diff, I didn't get that impression at least.

FichteFoll (commented):
Found this PR in my open tabs, so here's a review.

CSS/syntax_test_css.css
deathaxe: I'd say this is exactly this, isn't it?

The `-font-family` starts with a hyphon, which is invalid and would have therefore not been highlight before. This test indicated syntax doesn't care about it anymore. Or is it the leading hyphon, which is misleading?

The value `sans-none` would be scoped `string.unquoted` when used in a valid `font-family` property. In any other known or unknown property it is treated as normal constant.

FichteFoll: Then it's fine. I just wanted to make sure I'm understanding the test correctly.

FichteFoll (approved):

jfcherng (approved):

deathaxe:
We should wait with it until next dev builds start. Current master may receive only smaller and/or safe commits which mainly fix regressions.

deathaxe:
As ST continues with dev builds no reason to wait with merging any longer.

PR #3357, opened 2022-04-24
[C#] Improve namespace declaration

deathaxe:
This commit improves qualified and unqualified identifier scopes in
namespace declaration statements.

1. Qualified identifiers are scoped `meta.path`.
2. Path elements are scoped `variable.namespace`.
3. Only the last element is scoped `entity.name.namespace.

Goal is to have same scoping as in Erlang, Java, Perl, PHP, ...

Note:

The current implementation still relies on normal push/pop patterns,
which works only for single-line identifiers. If support for multi-line
identifiers is desired, some contexts from Java might be of interest.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -265,28 +265,55 @@ contexts:
 
   namespace_declaration:
     # package declaration
-    - match: '\b(namespace)\b'
+    - match: \b(namespace)\b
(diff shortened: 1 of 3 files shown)


michaelblyons (approved):

jrappen (approved):

jrappen:
appreciate the effort with named contexts

jfcherng (approved):

PR #3358, opened 2022-04-24
[CSS] Move system-font names' completions

deathaxe:
This commit moves system-font names to `<generic-font-names>` and fixes
reference names to that set of values. So system and generic font names
are available in both `font` and `font-family` properties.


michaelblyons (dismissed):

CSS/completions/common.py
```diff
@@ -267,7 +267,17 @@ def get_common_values():
             ["ornaments()", "ornaments($1)"],
             ["annotation()", "annotation($1)"],
         ],
-        "generic-font-name": ["serif", "sans-serif", "cursive", "fantasy", "monospace"],
+        "generic-font-name": [
+            "serif",
+            "sans-serif",
+            "cursive",
+            "fantasy",
+            "monospace" "caption",
```
jrappen, line 275: @deathaxe typo?

jrappen (approved):

michaelblyons (approved):

keith-hall (approved):

keith-hall:
For some context, in https://www.w3.org/TR/CSS2/fonts.html#propdef-font the `caption` | `icon` | `menu` | `message-box` | `small-caption` | `status-bar` are separate from `font-family` / `generic-font-family`:

> System fonts may only be set as a whole; that is, the font family, size, weight, style, etc. are all set at the same time. These values may then be altered individually if desired. If no font with the indicated characteristics exists on a given platform, the user agent should either intelligently substitute (e.g., a smaller version of the 'caption' font might be used for the 'small-caption' font), or substitute a user agent default font. As for regular fonts, if, for a system font, any of the individual properties are not part of the operating system's available user preferences, those properties should be set to their initial values.

> That is why this property is "almost" a shorthand property: system fonts can only be specified with this property, not with ['font-family'](https://www.w3.org/TR/CSS2/fonts.html#propdef-font-family) itself.

deathaxe:
So fixing the reference name is the only required change, left.

keith-hall (approved):

PR #3377, opened 2022-05-10
[meta] update workflows

jrappen:
- add release workflow
- use `actions/checkout@v3` in ci workflow

---

This commit adds a release for every tag:

- <https://github.com/sublimehq/Packages/tags>
- <https://github.com/sublimehq/Packages/releases>

with a zip of the Packages from this repo named `Packages ${{ github.ref }}.zip`.

```text
Packages vXXXX.zip
└─ Data
   └─ Packages
      ├─ *.sublime-package
      ├─ ...
      └─ *.sublime-package
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
     steps:
 
       - name: Git checkout
-        uses: actions/checkout@master
+        uses: actions/checkout@v3
 
       - name: Get binary for Build ${{ matrix.sublime-build }} (${{ matrix.sublime-channel }})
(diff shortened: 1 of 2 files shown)


jrappen:
Maybe a contributor could request a review from Ben.

deathaxe:
Any tagged commit is part of an official ST release. What's the benefit of providing extra downloads?

FichteFoll:
Additionally, any zip can be downloaded as a tar- or zipball from GitHub already, albeit with tests, but those are part of the shipped packages as well, so I don't see the point either.

PR #3381, opened 2022-05-14
[Python] Update sql_indicator

deathaxe:
This commit ...

1. copies `sql_indicator` from PHP (PR #3331) to support same kinds of
   sql queries.
   
   `ADD`, `DROP` and `TRUNCATE` are added which are used by some sql
   dialects such as T-SQL.

2. removes `string` scope from queries, keeping only `meta.string`, so
   SQL queries are highlighted the same way as in standalone sql source
   files and to enable code completions, etc. which are normally
   disabled by `string` scope.


jfcherng (approved):

jfcherng:
It looks like the removals of `string.quoted.single.python` from meta scope break tests.

deathaxe:
I really hate that splitting of syntax tests as it simply doesn't work to run tests for all at the same time on my box. Hence its too reasy to miss them.

jfcherng (approved):

michaelblyons (approved):

FichteFoll:
> I really hate that splitting of syntax tests as it simply doesn't work to run tests for all at the same time on my box. Hence its too reasy to miss them.

Does ST not pick up all relevant test files when you run the "Syntax Test" build on the syntax file?

deathaxe:
Just found the related core issue about that: https://github.com/sublimehq/sublime_text/issues/3771

PR #3394, opened 2022-05-18
[HTML] Prevent completions in inappropriate contexts

deathaxe:
This commit...

1. prevents attribute-completions within attribute values
   (aka. strings)

       <div attrib="|">   <- don't suggest attribute names

2. prevents abbreviation expansion within tags

       <div | >  <- don't expand `tag.attrib`

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -486,13 +486,16 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
 
         # Note: Exclude opening punctuation to enable appreviations
         #       if the caret is located directly in front of a html tag.
-        if match_selector("text.html meta.tag - punctuation.definition.tag.begin"):
+        if match_selector("text.html meta.tag - meta.string - punctuation.definition.tag.begin"):
             if ch in ' \f\n\t':
                 return self.attribute_completions(view, locations[0], prefix)
             return None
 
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

jrappen (dismissed):

HTML/html_completions.py
```diff
@@ -486,13 +486,16 @@ def match_selector(selector):
...
-        if match_selector("text.html meta.tag - punctuation.definition.tag.begin"):
+        if match_selector("text.html meta.tag - meta.string - punctuation.definition.tag.begin"):
             if ch in ' \f\n\t':
                 return self.attribute_completions(view, locations[0], prefix)
             return None
 
-        # Expand tag and attribute appreviations
-        return self.expand_tag_attributes(view, locations) or self.tag_abbreviations
+        if match_selector("text.html - meta.tag"):
+            # Expand tag and attribute appreviations
```
michaelblyons, line 495: Should be "abbreviations"

michaelblyons (dismissed):

HTML/html_completions.py
deathaxe: Fixed

keith-hall (approved):

michaelblyons (approved):

PR #3398, opened 2022-05-22
[Markdown] Improve MultiMarkdown frontmatter

deathaxe:
This PR:

1. adds support for YAML like frontmatter in MultiMarkdown

   If a document starts with `---` MultiMarkdown 6 interprets it as
   beginning of meta-data for compatibility with YAML frontmatter,
   if not run in compatibility mode (-c flag).

   It does not support YAML though! Content is limited to same simple
   key-value pairs as in classical multimarkdown header format.

2. scopes all kinds of headers `meta.frontmatter.multimarkdown`
3. scopes key-value pairs `meta.mapping`
4. supports HTML in frontmatter values. Therefore `string` scope is
   removed.

Implementation uses `embed` to ensure correct bailouts without too many context duplication.

inspired by: https://github.com/SublimeText-Markdown/MarkdownEditing/issues/690

---

#### Classical metadata format

grafik

#### YAML frontmatter like metadata format

grafik

### DIFF
--- a/Markdown/MultiMarkdown.sublime-syntax
+++ b/Markdown/MultiMarkdown.sublime-syntax
@@ -13,26 +13,39 @@ variables:
 
 contexts:
   main:
-    - match: ^(?={{header}})
-      push: multimarkdown-header
+    - meta_include_prototype: false
     - match: ^
-      set: multimarkdown-content
+      set:
+        - markdown
+        - frontmatter
 
-  multimarkdown-header:
-    - match: ^$
-      pop: 1
-    - match: ^(?:{{header}}\s*)?
+  frontmatter:
+    # MultiMarkdown 6 supports frontmatter like headers for compatibility reasons
+    # Note: The content is still limited to simple key-value pairs, no YAML!
+    - match: (---)\s*\n
       captures:
-        1: keyword.other.multimarkdown
-        2: punctuation.separator.key-value.multimarkdown
-      push: multimarkdown-header-value
-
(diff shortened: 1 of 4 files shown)


jrappen (approved):

Markdown/MultiMarkdown.sublime-syntax
```diff
@@ -13,26 +13,39 @@ variables:
...
-        2: punctuation.separator.key-value.multimarkdown
-      push: multimarkdown-header-value
-
-  multimarkdown-header-value:
-    - meta_scope: meta.header.multimarkdown
-    - meta_content_scope: string.unquoted.multimarkdown
-    - match: \n
+        0: meta.frontmatter.multimarkdown
+        1: punctuation.section.block.begin.frontmatter.multimarkdown
+        2: constant.other.language-name.multimarkdown
```
jrappen, line 29: I don't see a group 2?

deathaxe: Fixed.

jrappen (approved):

deathaxe:
Added a bailout to `frontmatter`.

see: https://github.com/SublimeText-Markdown/MarkdownEditing/issues/695

jrappen (dismissed):

jrappen (approved):

michaelblyons (approved):

jfcherng (approved):

PR #3399, opened 2022-05-22
[Markdown] Add support for setext headings in list-blocks

deathaxe:
This commit adds `list-setext-headings-or-paragraphs` contexts which are a modified copy of `setext-headings-or-paragraphs` to support setext headings within list items.

Note: It may also be useful if markdown is to be embedded into 3rd-party syntaxes (e.g.: Astro) with lazy indentation handling.

#### Example

grafik

#### Rendered as 

---

1. Heading
===

1. List Item
   ===

   - nested list
     
     with heading
     ---

     and paragraph

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -259,6 +259,20 @@ variables:
        )
     )
 
+  list_paragraph_end: |-
+    (?x: # pop out of this context if one of the following conditions are met:
+      ^(?= [ \t]*
+        (?: $                           # the line is blank (or only contains whitespace)
+        |   {{block_quote}}             # a blockquote begins the line
+        |   {{atx_heading}}             # an ATX heading begins the line
+        |   {{fenced_code_block_start}} # a fenced codeblock begins the line
+        |   {{thematic_break}}          # line is a thematic beak
+        |   {{list_item}}               # a list item begins the line
+        |   {{html_block}}              # a html block begins the line
+        )
+      )
+    )
+
   # https://spec.commonmark.org/0.30/#left-flanking-delimiter-run
   bold_italic_asterisk_begin: |-
     (?x:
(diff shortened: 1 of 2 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -23,7 +23,16 @@ variables:
   atx_heading: (?:[ ]{,3}[#]{1,6}(?:[ \t]|$))         # between 0 and 3 spaces, followed 1 to 6 hashes, followed by at least one space or tab or by end of the line
   atx_heading_space: (?:(?=[ \t]+#+[ \t]*$)|[ \t]+|$) # consume spaces only if heading is not empty to ensure `atx_heading_end` can fully match closing hashes
   atx_heading_end: (?:[ \t]+(#+))?[ \t]*($\n?)        # \n is optional so ## is matched as end punctuation in new document (at eof)
-  setext_escape: ^(?=[ ]{,3}(?:=+|-+)\s*$)            # between 0 and 3 spaces, followed by at least one hyphon or equal sign (setext underline can be of any length)
+
+  setext_heading_or_paragraph: ^[ ]{,3}(?:=+|(?=\S))  # between 0 and 3 spaces, followed by non-whitespace (consume equal signs as paragraphs may start with them)
+  setext_heading_escape: ^(?=[ ]{,3}(?:=+|-+)\s*$)    # between 0 and 3 spaces, followed by at least one hyphon or equal sign (setext underline can be of any length)
```
jrappen, line 28: it's "hyphen", also same in lines below

jrappen:
I'm not quite convinced that having capturing groups in variables is a good idea.

Otherwise this seems fine.

deathaxe:
Capture groups in variables have downsides for sure. But I am heading towards a more global strategy here. The added contexts are basically a copy of existing `setext-heading-or-paragraph` with two differences:

1. indentation patterns
2. paragraph-end patterns

Same applies to many other contexts, too.

So basically we could write a syntax wich contains only inline contexts and define variables for those differences. Doing so would enable us to inherit inlines for re-use in headings, paragraphs, lists, block quotes headings, block quote paragraphs, block quote lists, etc.

It would heavily reduce some duplication while further pushing accuracy.

It will especially become important if we want to introduce branching to avoid those illegal highlightings of unclosed `emphasis` (bold/italic/...).

Markdown/Markdown.sublime-syntax
deathaxe: Thanks, fixed.

jrappen:
Sure, I just find it difficult to review syntaxes using capturing groups in variables because you always have to jump back and forth between definition and reference and double-check.

jrappen (dismissed):

deathaxe:
Note: This PR is required for further work on MathJAX support.

jrappen (approved):

michaelblyons (approved):

jfcherng (approved):

PR #3400, opened 2022-05-22
[Markdown] Add public context for embedding by 3rd-party syntaxes

deathaxe:
This commit adds an alias context meant for use by external syntaxes, which want to embed Markdown.

The motivation is to avoid using `list-block-content` as it leaks too many implementation details.

_Note: This commit doesn't change behavior of Markdown syntax._

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -317,6 +317,13 @@ contexts:
     - include: atx-headings
     - include: setext-headings-or-paragraphs
 
+  indented-markdown:
+    # This is a public context which can be embedded by 3rd-party syntaxes
+    # to support Markdown highlighting in locations of arbitrary indentation,
+    # but without reference to list blocks as this is an implementation detail.
+    # The content may diverge from list block content in future.
+    - include: list-block-content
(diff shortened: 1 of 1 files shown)


deathaxe:
It may drive Markdown highlighting in HAML filters (Ruby on Rails) one day.

grafik

michaelblyons:
So we could make a format for the package control upgrade messages where (almost) everyone uses Markdown...?

deathaxe:
Not sure how core syntaxes are related with Package Control messages, but primary intention is to enable 3rd-party packages to use or extend Markdown the same way like HTML.

It might be pretty useful with regards to all those static site generators which all use markdown as template language.

Once this repo is released with a build, we can upgrade Liquid package with the content from [Liquid ST4 branch](https://github.com/deathaxe/sublime-liquid/tree/st4) to support authoring for [Jekyll](https://jekyllrb.com/)

Other's like [Astro](https://docs.astro.build/en/getting-started/) may follow, which is one this PR might be useful for.

FichteFoll (approved):

jfcherng (approved):

PR #3405, opened 2022-05-27
[C#] Added memory-based types

NickName73:
nint and nuint are base types included in dotnet
[Documentation about this memory types](https://docs.microsoft.com/dotnet/csharp/language-reference/builtin-types/nint-nuint)


keith-hall (approved):
It may be worth adding a small syntax test assertion for the new types, but otherwise seems fine to merge :)

deathaxe (approved):

PR #3408, opened 2022-05-29
[PHP] Fix regression in PHP arrow function return type scoping

### LINKED ISSUE #3407: [PHP] Regression: short arrow functions with return types scoped/highlighted incorrectly
### What happened?

Since build 4134, PHP short arrow functions with return types are not scoped properly. I haven't dug into the scopes for every different part of the short arrow syntax yet, but it appears to affect the `fn` keyword, the arguments passed to the function, the return type of the function, and sometimes `)` characters following the function.

I would guess these come from #3331. I'll look into a fix but someone more familiar with the new syntax definition (@deathaxe 👀) might be able to find the issue faster.

Demo in safe mode:

Screen Shot 2022-05-28 at 8 25 26 PM

bakerkretzmar:
Fixes #3407 by adding back support for short arrow function return types.

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -1005,6 +1005,7 @@ contexts:
 
   arrow-function-or-fail:
     - include: comments
+    - include: function-return-type
     - match: =>
       scope: keyword.declaration.function.arrow.php
       pop: 1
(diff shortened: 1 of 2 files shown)


deathaxe:
Thanks for contrubution. Hope hijacking your PR is not too rude.

Added an `arrow-function-return-type` context to maintain `meta.function.anonymous` scope in return types. The context is added to the multi-push expression so `=>` is validated after optional return-type.

jfcherng (approved):

michaelblyons (approved):

bakerkretzmar:
@deathaxe not at all, I was kinda hoping you would 😄 I wouldn't have gotten there myself. Thanks a lot!

Issue #3395, opened 2022-05-20
[YAML] Explicit mappings are not terminated properly


FichteFoll:
### What happened?

Sets in yaml are specified using a mapping with no values for the keys, which is easily accomplished in a multi-line layout of prefixing the items with an explicit key indicator `?`. However, the `meta.mapping.key` scope is not properly terminated.

```yaml
set: !!set
  ? this
  ? is
  ? a
  ? set

list:
  - this
  - is
  - a
  - list
```

![2022-05-20_16-17-17](https://user-images.githubusercontent.com/931051/169548291-dbd47ac7-0e70-4867-bbd1-ef3a712d2bde.png)

~~The same can also be observed when there are values for the keys~~
**Edit**: The following is actually correct behavior! I even added a test for it and then forgot about the behavior.

```yaml
map:
  ? this: value
  ? is: value
  ? a: value
  ? set: value
```

![2022-05-20_16-23-09](https://user-images.githubusercontent.com/931051/169548658-d1bb3d04-fac9-45be-88b2-382103b2ce39.png)

Likely a regression from #2564.

deathaxe:
It appears at least some of those issues have existed before, already, but haven't been that obvious due to `meta.mapping.key` not being present.

There's even a test case for explicit keys not being followed by values in the same line (related with: your second screenshot)

https://github.com/sublimehq/Packages/blob/9e35426d39030880ddf952caab000ec7a5eadf24/YAML/tests/syntax_test_block.yaml#L130-L135

deathaxe:
Also found:

![grafik](https://user-images.githubusercontent.com/16542113/169645954-3025949f-0fe8-4e14-86d3-afa2869aa58e.png)

from: https://perlpunk.github.io/slides.tpcia2017/the-state-of-the-yaml/slide026.html

FichteFoll:
> There's even a test case for explicit keys not being followed by values in the same line (related with: your second screenshot)

Right, I totally forgot about that. With a line break before the colon, this example works correct (as we also have tests for that. However, empty/missing values are seemingly not supported. I only recently learned about this syntax for sets (never used it before) and thus didn't consider it during development of the syntax.

PR #3409, opened 2022-05-30
[YAML] Fix bailout from explicit keys without value

deathaxe:
This commit addresses #3395 by terminating explicit mapping keys by lines which

1. start with `:` or 
2. are indented less or equal to the opeing `?`

It doesn't introduce special handling of `!!set` values so far. Those are still scoped explicit key without value as it is what spec calls them, too.

_Note: Appends `-11` to contexts which have `-12` counter parts._

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -732,22 +732,16 @@ contexts:
     - include: block-pair
 
   block-pair:
-    - match: \?(?=\s|$)
-      scope: meta.mapping.yaml punctuation.definition.key-value.begin.yaml
+    - match: ( *)(\?)(?=\s)
+      captures:
+        2: meta.mapping.yaml punctuation.definition.key-value.begin.yaml
       push: block-key-explicit
     # Attempt to match plain-out scalars and highlight as "meta.mapping.key",
     # if followed by a colon
     - match: '{{_flow_key_out_lookahead}}'
       push: block-key-implicit
-    - match: :(?=\s|$)
+    - match: :(?=\s)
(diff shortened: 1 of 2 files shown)


deathaxe:
I am not aware of a proper solution for following constructs however.

```yml
- - - ? explicit key 
        without value

    - should be list, but is scoped key
```

deathaxe:
This PR solves linked issue for most common situations. I'd therefore vote to merge it, despite some edge cases not beeing handled.

YAML/YAML.sublime-syntax
```diff
@@ -757,10 +751,21 @@ contexts:
       captures:
         1: punctuation.separator.key-value.mapping.yaml
       pop: true
-    - match: (?=\?(?=\s|$))  # Empty mapping keys & values are allowed
+    # multi-line keys are indented by at least one more space than `?`
+    - match: ^(?!\1 +\S|\s*$)
+      pop: true
+    # explicit key of any indentation
+    - match: ^ *(?=\?\s)
       pop: true
```
FichteFoll, line 759: What does this match protect us from? Could you add a test for it? As far as I can see,
```
!!set
? a
? b
```
should already be handled by the earlier match.

FichteFoll (commented):
Sorry, I just missed this earlier in my flood of notifications.

YAML/YAML.sublime-syntax
deathaxe: Probably just for double safety out of fear to break something. It actually seems contra-productive.

FichteFoll (approved):

jfcherng (approved):

PR #3416, opened 2022-05-31
[CSS] Add support for syntax dialects

deathaxe:
This PR

1. renames CSS.sublime-syntax to CSS (Basic).sublime-syntax
2. creates an inherit CSS from CSS (Basic)

Goal
----

Support different CSS dialects (Basic, PostCSS, Tailwind CSS, ...) in embedded code scenarios including templating support.

Motivation
----------

More and more syntax packages inherit CSS.sublime-syntax to add support for interpolation or embed template tags (e.g.: ASP, JSP, PHP, Liquid, Vue, ...).

Other packages might provide certain dialects of CSS, such as PostCSS or Tailwind CSS.

It is currently not easily possible to embed Tailwind CSS in PHP or Liquid without rewriting each of those syntaxes, because all of them rely on default CSS.sublime-syntax due to inheritance depending on file names rather than scopes.

The motivation is to enable exactly this (again), to benefit from both inheritance and flexibility from 3rd-party packages.

Strategy
--------

The idea is to turn CSS.sublime-syntax into an interface, which is inherit from CSS (Basic) and can be extended by and embedded into syntaxes such as PHP by default, while being able to easily create an
override with `extends` value replaced by desired syntax dialect (e.g.: Tailwind CSS).

       CSS (PHP)
          |
         CSS
          |  \
          |  Tailwind CSS
          |   |
          |  PostCSS
          |  /
         CSS3

Note: It follows the idea being discussed in SQL (#3046).

Example
-------

To enable Tailwind CSS in PHP

1. install [Tailwind CSS](https://packagecontrol.io/packages/Tailwind%20CSS)
2. modify _Tailwind CSS.sublime-syntax_ to extend from _CSS (Basic).sublime-syntax_
3. create a _Packages/CSS/CSS.sublime-syntax_ with following content

   ```yml
   %YAML 1.2
   ---
   name: CSS
   scope: source.css
   version: 2
   
   extends: Packages/Tailwind CSS/Tailwind CSS.sublime-syntax
   
   file_extensions:
     - css

   ```

A command to choose desired CSS dialect could be provided via Command Palette by something like https://github.com/deathaxe/DefaultSyntaxChooser


Ultra-Instinct-05:
Does it make sense to rename the basic version of CSS as `CSS (Plain)` instead of `CSS (Basic)` to be inline with the HTML syntax ?

deathaxe:
The original intent was to use `Plain` for syntaxes without those long lists of pre-defined tokens. _HTML (Plain)_ doesn't know about any html tag for instance, just highlighting anything which looks like that.

I rather found the whole CSS (...) naming scheme a bit problematic as linked "dialect chooser" doesn't handle it as CSS dialect for that reason.

CSS3 or something like that would probably be the better choice.

deathaxe:
The way diamond inerhitance works, makes outlined strategy impossible. Hence maybe not worth the extra complexity. That's something which would need to be handled by core via e.g.  sublimehq/sublime_text#5004

michaelblyons:
If you still have the branch locally, can you undelete? I'd like to ask some questions about strategy if yiu don't mind.

PR #3420, opened 2022-06-01
[HTML] Reorganize Tags

deathaxe:
This PR proposes to 

1. scope html tags by semantic meaning

   syntactic:
   - maybe self closing tags (`<link ... />`)
   - opening tags `<style>`
     - which must not be self-closing
     - scoped: `meta.tag.<type>.begin`
   - closing tags `</style>`
     - which must not contain attributes nor self-closing
     - scoped: `meta.tag.<type>.end`
   
   semantic groups:
   - block
   - form
   - inline
   - script
   - style
   - structure
   - table

2. highlight invalid self-closing html tags
3. update known tag names frowm whatwg.

It is inspired by #2772 and built on lessons learned from #2781.

A requests for better semantic scoping can be found at

- https://forum.sublimetext.com/t/html-sublime-syntax-wrong-scope-name-for-tables-tags/64207/2

#### Benchmarking

1. Syntax cache increases by 30% from about 60kB to 93kB.
2. Parsing performance doesn't change.  
   _Tried with syntax_test_html.html and some real world sources._

### DIFF
--- a/ASP/syntax_test_asp.asp
+++ b/ASP/syntax_test_asp.asp
@@ -1,9 +1,9 @@
 ' SYNTAX TEST "Packages/ASP/HTML (ASP).sublime-syntax"
 <!DOCTYPE html>
 <html>
-'<- meta.tag.structure.any.html punctuation.definition.tag.begin.html - source.asp.embedded.html
-'^^^^ meta.tag.structure.any.html entity.name.tag.structure.any.html
-'    ^ meta.tag.structure.any.html punctuation.definition.tag.end.html
+'<- meta.tag.structure punctuation.definition.tag.begin.html - source.asp.embedded.html
+'^^^^ meta.tag.structure entity.name.tag.structure
+'    ^ meta.tag.structure punctuation.definition.tag.end.html
  <%@ TRANSACTION = Required %>
 '^^ punctuation.section.embedded.begin.asp
 '    ^^^^^^^^^^^ constant.language.processing-directive.asp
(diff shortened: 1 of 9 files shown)


jwortmann:
1. I could imagine that it might be useful to have a special scope for [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements), i.e. tag names with a dash, for example `<popup-info>`. Currently, they just use the scope `meta.tag.other entity.name.tag.other`, just like e.g. `<unknowntagname>` ( = incomplete tag names while typing). Maybe a better scope name would be `meta.tag.custom entity.name.tag.custom`, if color schemes wants to highlight them special. A regexp for valid custom element names could be derived from https://html.spec.whatwg.org/#valid-custom-element-name.
2. Not sure if it is worth to introduce the new groups "form" and "table" at the 3rd level of the scope name. I'm certainly no HTML expert, but the various tags used for tables and forms already seem to fall into one of the categories "block" and "inline" (see https://www.w3schools.com/html/html_blocks.asp). New tags which are not yet listed, could probably also be categorized into one of the categories, dependant on how they are rendered in the browser.
It could be useful to know what exactly the mentioned plugin wants to achieve, but maybe it would be better to keep the current categorization? The scope name for the `<table>` tag should be fixed of course.
4. The rest of this PR with begin/end for opening & closing tags, and highlighting for invalid self-closing tags looks like useful improvements to me.

jrappen (approved):

deathaxe:
I am uncertain, tbh. while the PR tries to cleanup some messed tag scope groupings, I am concerned about the change to be to invasive. It may break a lot of 3rd-party packages' syntax tests due due to renamed scopes. Also not sure whether we need all the "new" tag categories.

PR #3421, opened 2022-06-04
[Python] Adjust scope for module name in qualified base class

jwortmann:
Inspired by the syntaxes for e.g. Java and PHP.

**Before**:
before

**After**:
after

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1080,14 +1080,19 @@ contexts:
               captures:
                 1: variable.parameter.class-inheritance.python
                 2: keyword.operator.assignment.python
-            - match: (?={{path}})
+            - match: (?={{identifier}} *\.)
               push:
-                - meta_scope: entity.other.inherited-class.python
-                - match: '{{identifier}}(?: *(\.) *)?'
+                - meta_scope: meta.path.python
+                - match: '({{identifier}}) *(\.) *'
                   captures:
-                    1: punctuation.accessor.dot.python
+                    1: variable.namespace.python
+                    2: punctuation.accessor.dot.python
+                - match: '{{identifier}}'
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):

Python/syntax_test_python.py
```diff
@@ -1609,8 +1609,10 @@ class MyClass(Inherited, \
 #                      ^ punctuation.separator.inheritance
 #                        ^ punctuation.separator.continuation.line.python
               module . Inherited2, metaclass=ABCMeta):
-#             ^^^^^^^^^^^^^^^^^^^ entity.other.inherited-class
+#             ^^^^^^^^^^^^^^^^^^^ meta.path
```
michaelblyons, line 1612: Can you add a `- meta.path meta.path` here?

michaelblyons (dismissed):

Python/Python.sublime-syntax
```diff
@@ -1080,14 +1080,19 @@ contexts:
...
-                - meta_scope: entity.other.inherited-class.python
-                - match: '{{identifier}}(?: *(\.) *)?'
+                - meta_scope: meta.path.python
+                - match: '({{identifier}}) *(\.) *'
                   captures:
-                    1: punctuation.accessor.dot.python
+                    1: variable.namespace.python
+                    2: punctuation.accessor.dot.python
+                - match: '{{identifier}}'
+                  scope: entity.other.inherited-class.python
```
deathaxe, line 1091: The context may be popped once the final identifier is consumed.

```suggestion
                  scope: entity.other.inherited-class.python
                  pop: true
```

deathaxe (approved):

michaelblyons (approved):

PR #3423, opened 2022-06-05
[Git Formats] Add scope for newline at section end

jfcherng:
Basically I would like to introduce the same thing like https://github.com/sublimehq/Packages/pull/2472

### DIFF
--- a/Git Formats/Git Config.sublime-syntax
+++ b/Git Formats/Git Config.sublime-syntax
@@ -47,7 +47,11 @@ contexts:
     - meta_scope: meta.brackets.git.config
     - match: \]
       scope: punctuation.definition.brackets.end.git.config
-      pop: 1
+      set:
+        - match: \s*$
+          scope: meta.whitespace.newline.git.config
+        - match: ''
+          pop: 1
     - include: illegal-line-end-pop
     - match: \S
       scope: invalid.illegal.header-end.git.config
--- a/Git Formats/tests/syntax_test_git_config
+++ b/Git Formats/tests/syntax_test_git_config
@@ -14,6 +14,7 @@
 #^^^^^^^ meta.brackets entity.name
 #       ^ meta.brackets punctuation.definition.brackets.end
 #        ^ -meta.brackets
+#        ^ meta.whitespace.newline.ini
 [section "subsection"]
 # <- meta.brackets punctuation.definition.brackets.begin
 #^^^^^^^^^^^^^ meta.brackets
@@ -96,6 +97,7 @@
 #           ^ punctuation.definition.string.end
 #            ^ punctuation.definition.brackets.end
 #             ^ -meta.brackets
+#             ^ meta.whitespace.newline.ini
     old = red
 # <- meta.mapping
 #^^^^^^^^^^^^ meta.mapping


Git Formats/Git Config.sublime-syntax
```diff
@@ -47,7 +47,11 @@ contexts:
     - meta_scope: meta.brackets.git.config
     - match: \]
       scope: punctuation.definition.brackets.end.git.config
-      pop: 1
+      set:
+        - match: \s*$
+          scope: meta.whitespace.newline.git.config
+        - match: ''
+          pop: 1
```
deathaxe, line 54: What should be added is a dedicated meta scope for section headers. INI syntax uses `meta.section`. So we could rename `meta.brackets` to somewhat more specific:

```yml
  section-header:
    - match: \[
      scope: punctuation.definition.section.begin.git.config
      set: [expect-line-end, section-header-end, section-name]

  section-header-end:
    - meta_scope: meta.section.git.config
    - match: \]
      scope: punctuation.definition.section.end.git.config
      pop: 1
    - include: illegal-line-end-pop
    - match: \S
      scope: invalid.illegal.header-end.git.config
```

deathaxe (commented):
Markdown uses `meta.whitespace.newline` as folding marker, because atx headings don't have a crucial end marker, which consists of a single or only few chars.

Section headers in ini / git config files however are defined by leading and trailing brackets, which can be used as folding markers.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
    <key>scope</key>
    <string>text.git.config</string>
    <key>settings</key>
    <dict>
        <key>foldScopes</key>
        <array>
            <dict>
                <key>begin</key>
                <string>meta.section punctuation.definition.section.end</string>
                <key>end</key>
                <string>meta.section punctuation.definition.section.begin</string>
                <key>excludeTrailingNewlines</key>
                <true/>
            </dict>
        </array>
    </dict>
</dict>
</plist>
```

deathaxe:
P.S.:

`punctuation.definition.brackets` is probably not needed in `expect-section` context and may be removed:

```
  expect-section:
    - match: (?=^\s*\[) # start of new section
      pop: 1
    - match: \]
      scope: invalid.illegal.stray-bracket.git.config
```

deathaxe:
Do we want to use `Git Config - Fold.tmPreferneces` to make clear which syntax it belongs to?

jfcherng:
To be clarify, I don't care about folding things since I don't use it personally. The goal of this PR is to capture the newline char at EOL so that I the whole section line can be applied a bg color.

deathaxe (approved):

deathaxe (changes requested):
Then you might want to take trailing comments into account:

grafik

jfcherng:
Is it possible to use variables in another syntax without inheritance?

deathaxe:
Unfortunatelly not. Already thought about refactoring syntaxes so all use Git Common as base syntax ;-)

I guess the easiest solution was something like scoping the whole line `meta.section` and re-add `meta.brackets` for the `[name]` part. So you'd have one scope to address in your color scheme.

deathaxe (dismissed):

Git Formats/tests/syntax_test_git_config
```diff
@@ -88,14 +91,13 @@
 
 # COLOR TESTS
 [color "diff"]
```
jrappen, line 93: did you also want to test the extent of `meta.brackets`?

Git Formats/tests/syntax_test_git_config
```diff
@@ -454,8 +456,8 @@ stray-bracket]
 
 # REAL WORLD SAMPLE TESTS
 [alias]
```
jrappen, line 458: same as above

Git Formats/tests/syntax_test_git_config
```diff
@@ -9,22 +9,25 @@
...
-#^^^^^^^^^^^^^ meta.brackets
+# <- meta.section.header meta.brackets punctuation.definition.brackets.begin
+#^^^^^^^^^^^^^^^^^^^^^ meta.section.header meta.brackets
+#                     ^ meta.section.header - meta.brackets
 #^^^^^^^ entity.name
 #        ^^^^^^^^^^^^ string.quoted.double
 #        ^ punctuation.definition.string.begin
 #                   ^ punctuation.definition.string.end
 #                    ^ punctuation.definition.brackets.end
 [section \"subsection"]
```
jrappen, line 28: extent of `meta.brackets` not fully tested here

michaelblyons (approved):

jrappen (approved):

deathaxe (approved):

PR #3425, opened 2022-06-05
[Ruby]Add new `BlankSlate` snippet using `BasicObject`

toshimaru:
## Description

Since Ruby 1.9, `BasicObject` is available to create `BlankSlate` class.

> BasicObject is the parent class of all classes in Ruby. It's an explicit blank class.
> 
> [BasicObject](https://ruby-doc.org/core-1.9.3/BasicObject.html) can be used for creating object hierarchies independent of Ruby's object hierarchy, proxy objects like the Delegator class, or other uses where namespace pollution from Ruby's methods and classes must be avoided.

ref. https://ruby-doc.org/core-1.9.3/BasicObject.html

I re-create BlankSlate snippet and modify `tabTrigger` in order to distinguish from other Ruby class snippets.

image

## Changes

- Use BasicObject for Ruby BlankSlate since `BasicObject` is available as of Ruby v1.9
- Change tabTrigger from `cla` to `clab` since we already have some snippets triggered by `cla`
- Simplify snippet file name

## `BlankSlate` Example

The following is an example of `BlankSlate` class with `BasicObject`

```rb
class BlankSlate < BasicObject
end
```


deathaxe:
How about just not using that snippet?

Removed things tend to cause someone to complain for whatever reason. Just imagine maintainer of old code bases maybe relying on older ruby builds for some reason, which still might want that snippet.

jrappen:
Right. I'm saying both your points are valid.

toshimaru:
@deathaxe thanks for your feedback :)

> Removed things tend to cause someone to complain

Yes, that was my concern.

> How about just not using that snippet?

I'm happy to do that, but I have another suggestion for this, 
what about adding another trigger for that because we already have a lot of snippet for `cla`.

image

I'd like to another trigger for BlankSlate class like Ruby methods.

image

How about adding `clab` for the trigger?
What do you think about my idea?

jrappen:
Just in case you didn't know:

- You can ignore [snippets](https://www.sublimetext.com/docs/completions.html#snippets) by using [`ignored_snippets`](https://www.sublimetext.com/docs/completions.html#ignored_snippets):
  image
- [Overriding](https://www.sublimetext.com/docs/packages.html#overriding-files-from-a-zipped-package) is also an option.

toshimaru:
That's good to know!

I'm lazy, so it'll be great if nice snippets are available by default rather than customizing. 😸

deathaxe:
> How about adding `clab` for the trigger?

Looks like a reasonable change / trigger.

toshimaru:
@deathaxe Hey, I've re-added Ruby BlankSlate snippets. Could you review it ❓

deathaxe:
I hesitate approving this change because snippet's content is significantly changed. Changing trigger to reduce duplicates might be ok, but I am not sure whether changed content may cause complains.

toshimaru:
> may cause complains.

I don't think so. BasicObject is available as of Ruby 1.9.3 and [Ruby 1.9.3 EoL-ed on February 23, 2015](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/).

I know many users are still using Ruby v2.x, but I don't know the user using Ruby v1.9 in 2022.

michaelblyons (approved):

FichteFoll (approved):
Pretty much all currend `class-….sublime-snippet` files currently defining `cla` as the tab trigger makes them arguably impossible to use in a proper scenario, unless you can trust ST to always put the same `cla` at the first position (or pre-select it, which I admittedly don't remember right now). 

Outside of that, you'd likely select the one snippet from the multiple `cla`s that you want using arrow keys, so using a different tab trigger for the updated snippet is definitely appreciated.

I don't know any Ruby, so I cannot comment on the likelyhood of these snippets being used or useful in the first place, but just from a quick glance they appear very dated and probably go back by at least 10 years or the initial import of the Ruby package into ST. If the proposal is a faithful replacement, then that is fine in my book.

Still, snippets in this repo and especially the old ones are not in a good place in general and it's unclear which direction to move forward.

PR #3426, opened 2022-06-07
[PHP] PHP 8.2 new syntax

jfcherng:
PHP 8.2 is scheduled to be 

- first alpha on `Jun 09 2022`
- feature freeze on `Jul 19 2022`
- stable release on `Nov 24 2022`

**References**:

- PHP 8.2 RFCs: https://wiki.php.net/rfc#php_82
- PHP 8.2 Milestone: https://github.com/php/php-src/milestone/4?closed=1
- PHP 8.2 release schedule: https://wiki.php.net/todo/php82

### Need new parsing paths

- [x] [PHP RFC: Disjunctive Normal Form Types](https://wiki.php.net/rfc/dnf_types)

- [X] [PHP RFC: Readonly classes](https://wiki.php.net/rfc/readonly_classes)
- [X] [PHP RFC: Add `true` type](https://wiki.php.net/rfc/true-type)
- [X] [PHP RFC: Allow `null` and `false` as stand-alone types](https://wiki.php.net/rfc/null-false-standalone-types)


PHP/PHP Source.sublime-syntax
```diff
@@ -683,7 +683,7 @@ contexts:
 ###[ CLASS DECLARATIONS ]#####################################################
 
   class-declarations:
-    - match: (?i:(abstract|final)\s+)?((?i:class))\b
+    - match: (?i:(abstract|final|readonly)\s+)?((?i:class))\b
```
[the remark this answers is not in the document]
deathaxe: The easiest workaround was to remove modifiers from pattern at all. So `readonly` would be scoped correctly automatically and the declaration pattern would look exactly like those for enums and interfaces. The only downside is `meta.class` not covering those modifiers, which it currently does not for enums/interfaces/functions as well.

```yml
  class-declarations:
    - match: (?i:class)\b
      scope: meta.class.php keyword.declaration.class.php
      push:
        - class-declaration-body
        - class-declaration-name
```

Assuming modifiers are always written in the same line as the declaration keyword, an alternative (maybe for the others, too) could look like:

```yml
  class-declarations:
    - match: (?=(?:{{storage_modifiers}}\s+)*(?i:class)\b)
      push:
        - class-declaration-body
        - class-declaration-name
        - class-declaration-type

  class-declaration-type:
    - match: (?i:class)\b
      scope: keyword.declaration.class.php
      pop: 1
    - match: '{{storage_modifiers}}'
      scope: storage.modifier.php
```

In case modifiers may be on separate lines, we'd need an approach using `branch`es just like Java does.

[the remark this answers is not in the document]
jfcherng: Thanks. I think the easy workaround is probably fine.

jfcherng:
Since it's feature-frozen days ago and there is no voting RFCs. I assume this PR has been done.

PHP/PHP Source.sublime-syntax
```diff
@@ -1905,17 +1905,17 @@ contexts:
 
 ###[ TYPE HINTS ]#############################################################
 
-  type-hints:
+  type-hints-push:
```
deathaxe, line 1908: I'd prefer to keep the plural = pushing vs. singular = setting/popping style of context naming rather than adding arbitary `set`, `pop`, ... pre- or suffixes. 

`-pop` suffix is dedicated to those prototype like contexts which contain patterns which are re-used to pop contexts ( like `else-pop` ...).

PHP/PHP Source.sublime-syntax
```diff
@@ -1931,10 +1931,20 @@ contexts:
     - match: '&(?!\s*\$)'
       scope: punctuation.separator.type.intersection.php
       push:
-        - type-hint-nullable
         - type-hint-simple-type
+        - type-hint-nullable
+    - include: type-hint-body-group
     - include: else-pop
 
+  type-hint-body-group:
```
deathaxe, line 1939: `type-hint-group` should be enough. `-body` is a suffix, which should not be followed by anything else.

PHP/PHP Source.sublime-syntax
deathaxe: ... otherwise we end up adding `set` and `pop` or `push` to nearly each context, which would add much naming repetition.

jfcherng: ok. i didn't aware this naming convention at all. so get confused.

PHP/PHP Source.sublime-syntax
```diff
@@ -1934,10 +1931,20 @@ contexts:
...
         - type-hint-simple-type
+        - type-hint-nullable
+    - include: type-hint-group
     - include: else-pop
 
+  type-hint-group:
+    - match: \(
+      scope: punctuation.section.group.begin.php
+      push:
+        - match: \)
```
deathaxe, line 1943: A meta scope might help syntax based folding.

```suggestion
        - meta_scope: meta.group.php
        - match: \)
```

PHP/PHP Source.sublime-syntax
deathaxe: Sorry for my "wrong" hint. With regards to the plural vs. singular thing, this context should be `type-hint-groups` as it pushes a context onto stack.

deathaxe (approved):

michaelblyons (approved):

PR #3432, opened 2022-06-18
[JavaScript] Refactor literal strings

deathaxe:
This PR adds named contexts for literal strings and templates to better support inheritance.

It doesn't change behavior.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1088,58 +1088,75 @@ contexts:
     - include: else-pop
 
   literal-string:
-    - match: "'"
+    - include: literal-double-quoted-string
+    - include: literal-single-quoted-string
+
+  literal-double-quoted-string:
+    - match: \"
       scope: punctuation.definition.string.begin.js
-      set:
-        - meta_include_prototype: false
-        - meta_scope: meta.string.js string.quoted.single.js
-        - match: \'
-          scope: punctuation.definition.string.end.js
-          pop: true
-        - match: \n
-          scope: invalid.illegal.newline.js
-          pop: true
-        - include: string-content
-    - match: '"'
+      set: literal-double-quoted-string-content
+
+  literal-double-quoted-string-content:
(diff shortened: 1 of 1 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -1088,59 +1088,76 @@ contexts:
...
+    - include: string-content
 
   string-content:
     - match: \\\n
       scope: constant.character.escape.newline.js
     - match: \\(?:x\h\h|u\h\h\h\h|.)
       scope: constant.character.escape.js
 
+  string-interpolations:
+    - match: \$\{
```
Ultra-Instinct-05, line 1147: Does it make sense to give the leading `$` a different scope like `punctuation.definition.interpolation.js` since it signals the start of an interpolation ?

deathaxe: This kind of `$` interpolation scoping is a topic on its own, I guess. I struggled with scoping them in Bash/Perl/PHP and probably various other syntaxes when working on variables and interpolation topics. Especially interesting in syntaxes like PHP/Perl/Bash whose variables all begin with `$`. So it is normally scoped `punctuation.definition.variable`, while being scoped `punctuation.section.interpolation` in some contexts (like here).

Perl also uses it as `referencing operator` in case multiple `$$$$var` appear.

We should probably revice that with all related syntaxes being looked into.

Ultra-Instinct-05 (approved):

JavaScript/JavaScript.sublime-syntax
Ultra-Instinct-05: My main point here was scoping of `$` within strings that are considered interpolations. `$` in variables receiving `punctuation.definition.variable` makes sense, but I don't know anything that uses it apart from PHP and perhaps that's a related but separate topic than strings. But I agree, it can be tackled separately.

michaelblyons (approved):

Thom1729 (approved):

PR #3435, opened 2022-06-19
[CSS] Use auto_complete_trailing_symbols setting

deathaxe:
This PR modifies CSS auto-completions logic by appending
- `:;` after properties
- `;` after values

only if `auto_complete_trailing_symbols` is set `true`.

The setting is available out of the box.

The goals are to 

1. add more flexibility to customize completion behavior 
2. to follow ST's auto-completion behavior more closely

### DIFF
--- a/CSS/css_completions.py
+++ b/CSS/css_completions.py
@@ -91,28 +91,29 @@ class CSSCompletions(sublime_plugin.EventListener):
         return None
 
     def complete_property_name(self, view, prefix, pt):
-        text = view.substr(sublime.Region(pt, view.line(pt).end()))
-        matches = self.re_value.search(text)
-        if matches:
-            colon, space, value, term = matches.groups()
-        else:
-            colon = ""
-            space = ""
-            value = ""
-            term = ""
-
-        # don't append anything if next character is a colon
         suffix = ""
-        if not colon:
-            # add space after colon if smart typing is enabled
-            if not space and view.settings().get("auto_complete_trailing_spaces"):
-                suffix = ": $0"
+        if view.settings().get("auto_complete_trailing_symbols"):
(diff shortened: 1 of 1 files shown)


keith-hall (approved):
I think tying to `auto_complete_trailing_symbols` makes sense, and the change looks okay to me. Its unfortunate for me personally though, because although I like semi-colons completed automatically in CSS files, I have `auto_complete_trailing_symbols` turned off because I don't like buffer completions "randomly" inserting extra (non-word) characters, as it feels less predictable to me and I end up re-typing them and having to delete the duplicates. But I can accept that I can just tweak my keybindings to get my setup perfect for me if this change will benefit others :)

deathaxe:
I also have it disabled for exactly the same reason, by default. Enabled it via syntax specific settings for CSS. Haven't checked for "false positive trailing punctuation" so far, though.

I'd also be fine with another setting to control it, maybe even one which could be applied to the key binding as well. It's just that all those smart features seem to be a quite sensible part for some users.

jwortmann:
Perhaps this setting should be set to `true` in the syntax specific CSS.sublime-settings to preserve the current behavior? Because the global setting is set to `false` by default, and to not append the colon and semi-colon anymore could be an unwanted change for some or many users. On the other hand it might be confusing that they would need to adjust the syntax specific settings then, if they want to turn it off (but it would still be an improvement to the status quo which doesn't allow to turn it off at all). And I don't know if that would have any other effects on completions for e.g. CSS functions.

deathaxe:
If we are in doubt I'd prefer an alternative setting rather than introducing more syntax specific overrides. Those are hard to get rid of from a users end and thus may cause complains as well. That's the lesson learned from MarkdownEditing.

PR #3446, opened 2022-07-13
[Git Formats] Highlight "Change-Id:" in commit msg

jfcherng:
Not sure whether it's worth being added but anyway, a PR is here.

---

It's used in [Gerrit Code Review](https://en.wikipedia.org/wiki/Gerrit_(software)).

For people who use Gerrit Code Review, their commit messages usually look like the following.
The `Change-Id` attribute is necessary for submitting a code review.

```
This is the commit subject.

Here are the commit details line 1.
Here are the commit details line 2.
......

Change-Id: I3b3613e336397febf02cd8e5c14c53d493343e93
Signed-off-by: Foo Bar <foo.bar@example.com>
```

### DIFF
--- a/Git Formats/Git Commit.sublime-syntax
+++ b/Git Formats/Git Commit.sublime-syntax
@@ -81,10 +81,11 @@ contexts:
 
   # for Gerrit Code Review
   change-id:
-    - match: ^\s*(Change-Id)\s*(:)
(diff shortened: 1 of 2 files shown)


jrappen:
what about a test?

deathaxe (approved):

PR #3449, opened 2022-07-16
[JavaScript] Re-enable indentation based folding

### LINKED ISSUE #3429: [JavaScript] Folded symbols disappear in some cases
### What happened?

Folding symbols disappear
image
image

SC ST4126: No problem
DC ST4134: Exist

deathaxe:
Fixes #3429

### DIFF
--- a/JavaScript/Fold.tmPreferences
+++ b/JavaScript/Fold.tmPreferences
@@ -5,8 +5,6 @@
     <string>source.js | source.ts</string>
     <key>settings</key>
     <dict>
-        <key>indentationFoldingEnabled</key>
-        <false/>
         <key>foldScopes</key>
         <array>
             <dict>


jfcherng (approved):

FichteFoll:
What is our rule/guideline for disabling or enabling indentation-based folding for syntaxes?

deathaxe:
It would be safe to be disabled if all relevant aspects of a syntax and its posibily inherit 3rd-party syntaxes are handled by syntax based folding. As at least block comments are hard to handle without interference with ligatures and the way we scope begin/end punctuation, I guess we can't or shouldn't disable indentation based folding at all.

I tried to do so for JSON & XML, too, and quickly learned many 3rd party packages relying on it (e.g.: PackageDev) would need adoption. Otherwisw folding might not work at all in some cases. So it's not safe to disable indentation based folding.

FichteFoll (approved):

PR #3451, opened 2022-07-24
[D] Fix declaration vs. assignment detection

### LINKED ISSUE #3450: [D] Identifiers starting with `const` considered keyword for syntax highlighting
* Sublime Text Build 4131

It appears as though identifiers starting with `const` are treated as keywords and thus highlight wrong when used as variables.

```d
void main() {
  bool constant = false;
  constant = true;
}
```

example

`constant` is just one example - `constabcdefg` does this too, etc. Thanks!

deathaxe:
Fixes #3450

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -46,7 +46,7 @@ variables:
   string_lookahead: '(?=`|[rxq]?"|q{)'
 
   definitely_value_lookahead: '(?=!|~|\+|\-|\*|&|\bcast\b|\bdelete\b|\bnew\b|\bimport\b|\bis\b|\b__traits\b|\bfunction\b|\bdelegate\b|[0-9]|\[|\(|{{string_lookahead}}|\b({{language_constant}})\b|\b({{language_variable}})\b)'
-  definitely_declaration_lookahead: '(?=(?:{{name}}\.)*{{name}}(\s+|\s*\*+\s*|\**\[.*\]\s*){{name}}|\b{{type_qualifier}}\b)'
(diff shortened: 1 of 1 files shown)


D/D.sublime-syntax
```diff
@@ -46,7 +46,7 @@ variables:
   string_lookahead: '(?=`|[rxq]?"|q{)'
 
   definitely_value_lookahead: '(?=!|~|\+|\-|\*|&|\bcast\b|\bdelete\b|\bnew\b|\bimport\b|\bis\b|\b__traits\b|\bfunction\b|\bdelegate\b|[0-9]|\[|\(|{{string_lookahead}}|\b({{language_constant}})\b|\b({{language_variable}})\b)'
-  definitely_declaration_lookahead: '(?=(?:{{name}}\.)*{{name}}(\s+|\s*\*+\s*|\**\[.*\]\s*){{name}}|\b{{type_qualifier}}\b)'
+  definitely_declaration_lookahead: '(?=(?:{{name}}\.)*{{name}}(\s+|\s*\*+\s*|\**\[.*\]\s*){{name}}|\b({{type_qualifier}})\b)'
```
keith-hall, line 49: Would it make sense to have this be a non-capturing group?
```suggestion
  definitely_declaration_lookahead: '(?=(?:{{name}}\.)*{{name}}(\s+|\s*\*+\s*|\**\[.*\]\s*){{name}}|\b(?:{{type_qualifier}})\b)'
```

keith-hall (dismissed):

D/D.sublime-syntax
deathaxe: I followed the style of all the other variables for consistency reasons and don't want to overhaul all of them.

keith-hall (approved):

jfcherng (approved):

PR #3453, opened 2022-07-25
[Batch File] Fix REM multi-line comment highlighting

deathaxe:
REM comments out only the content on the same line, except being directly followed by `^`. Only the following line is commented out then.

see: https://github.com/sublimehq/sublime_text/issues/5512

### DIFF
--- a/Batch File/tests/syntax_test_batch_file.bat
+++ b/Batch File/tests/syntax_test_batch_file.bat
@@ -20,7 +20,7 @@
 :: ^^^^^^^^^^^^^^^^^ comment.line.rem.dosbatch
 
    REM ^
-   I'm a (com|ment)
+   I'm a (com|ment) ^
    not a comment
 :: ^^^^^^^^^^^^^ - comment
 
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):
What a terrible language.

jfcherng (dismissed):

Batch File/tests/syntax_test_batch_file.bat
```diff
@@ -19,13 +19,18 @@
    I'm a (com|ment)
 :: ^^^^^^^^^^^^^^^^^ comment.line.rem.dosbatch
 
+   REM ^
+   I'm a (com|ment)
+   not a comment
+:: ^^^^^^^^^^^^^ - comment
```
FichteFoll, line 25: Could you add tests for
```batch
REM ^
  line 1 ^
  line 2

REM line 1 ^
  line 2 ^
  line 3
```
? 
(Which is what https://github.com/sublimehq/sublime_text/issues/5512 reported but I don't see that having a test here)

deathaxe: I don't see a point in duplicating issue snippets.

The point is the line after REM being a comment only if REM is directly followed by `^`. 

Hence there's absolutely no point in checking with another `^` at line 2, because it doesn't change anything.

So no, I won't add something.

FichteFoll (approved):

michaelblyons (approved):

PR #3464, opened 2022-08-07
[Python] Improve case statement indentation rules

### LINKED ISSUE #3456: [Python] Case statement doesn't auto-dedent
### Expected behavior

```py
match thing := "one":
	case "one":
		stuff
	case "two":
		stuff
```

### Actual behavior

```py
match thing := "one":
	case "one":
		stuff
		case "two":
			stuff
```

### Steps to reproduce

1. Start a `match` block with whatever condition (I used a `:=` as part of it but that doesn't affect the result)
2. Enter at least two `case`s, without manually entering/deleting any indentation

### Environment
Windows 10 x64, ST Portable 4134

deathaxe:
Fixes #3456 

This PR...

1. adds an `expect-first-case-statement` context to scope first case
   statement `meta.disable-dedentation.python`.

   > **Note**
   >  An extra pattern is added which doesn't `fail` any branches.

2. adds `Indentation Rules - Case.tmPreferences` to decrease
   indentation level of all but the first case keywords.

   > **Note**
   >  As `case` can also be a variable, dedicated indentation rules
   >  are used to avoid decreasing indentation level in normal
   >  assignment expressions.

   > **Note**
   >  Indentation rules are not applied if a case pattern spans
   >  multiple lines, because `case` is not scoped keyword before final
   >  `:` is inserted. ST applies indentation rules per line only.

3. scopes whitespace after `:` `meta.statement.conditional.case` as it
   is required for ST to reliably apply indentation rules.

### DIFF
--- a/Python/Indentation Rules.tmPreferences
+++ b/Python/Indentation Rules.tmPreferences
@@ -6,7 +6,7 @@
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string>^\s*(elif|else|except|finally)\b.*:</string>
+		<string>^\s*(elif|else|except|finally)\b</string>
 		<key>increaseIndentPattern</key>
 		<string><![CDATA[(?x)
 			^\s*
--- a/Python/Indentation Rules.tmPreferences
+++ b/Python/Indentation Rules.tmPreferences
@@ -5,8 +5,6 @@
 	<string>source.python</string>
 	<key>settings</key>
 	<dict>
-		<key>decreaseIndentPattern</key>
-		<string>^\s*(elif|else|except|finally)\b</string>
 		<key>increaseIndentPattern</key>
 		<string><![CDATA[(?x)
 			^\s*
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -501,10 +501,11 @@ contexts:
     - include: allow-unpack-operators
 
   case-statement-end:
-    - match: ':(?!=)'
-      scope:
-        meta.statement.conditional.case.python
-        punctuation.section.block.conditional.case.python
+    # consume trailing whitespace so ST applies indentation rules reliably
+    - match: (:(?!=))\s*
+      scope: meta.statement.conditional.case.python
+      captures:
+        1: punctuation.section.block.conditional.case.python
       pop: true
 
   case-statement-fail:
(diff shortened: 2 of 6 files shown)


Python/Python.sublime-syntax
```diff
@@ -725,7 +726,27 @@ contexts:
...
+        meta.statement.conditional.case.python
+        keyword.control.conditional.case.python
+      set:
+        - meta-disable-detention
+        - case-statement-pattern
+        - allow-unpack-operators
+    - include: comments
+    - include: else-pop
+
+  meta-disable-detention:
```
keith-hall, line 745:
```suggestion
  meta-disable-dedention:
```

keith-hall (approved):

Python/Python.sublime-syntax
jwortmann: "dedentation" ?

keith-hall (approved):

Python/Indentation Rules.tmPreferences
```diff
@@ -6,7 +6,7 @@
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string>^\s*(elif|else|except|finally)\b.*:</string>
```
FichteFoll, line 9: Should probably disable these patterns for multi-line `string`s if we're making them less specific. For comments, the Default package already exempts them from indentation rules by setting `preserveIndent

Python/Python.sublime-syntax
```diff
@@ -501,10 +501,11 @@ contexts:
     - include: allow-unpack-operators
 
   case-statement-end:
-    - match: ':(?!=)'
-      scope:
-        meta.statement.conditional.case.python
-        punctuation.section.block.conditional.case.python
+    # consume trailing whitespace so ST applies indentation rules reliably
```
FichteFoll, line 504: Does this still work with comments after the colon?

deathaxe: Reindenting seems uneffected by comments after colons.

Indentation when hitting enter after `case "foo": # comment` completely fails. But this is unrelated to this change.

I won't work on that as all efforts to improve this kind of behaviors in other syntaxes has been completely reverted recently.

deathaxe: The next commit should improve situation with trailing comments.

Python/Indentation Rules.tmPreferences
deathaxe: Using jwortmanns approch for Matlab keeping `.*:` as is.

FichteFoll (approved):

PR #3468, opened 2022-08-10
[HTML] Fix html completion not working prior to tag

### LINKED ISSUE #3467: [HTML] snippet completions aren't suggested prior to closing tags
### Expected behavior

HTML tag snippets should be suggested when caret is within tags

### Actual behavior

No HTML tag snippets are suggested

This seems to be introduced in commit `9e35426d39030880ddf952caab000ec7a5eadf24`

### Steps to reproduce

1. Create a HTML buffer with the following contents: `<div></div>`
2. Place caret directly prior to closing tag (within div)
3. Type `tt` and attempt to auto-complete to the `<tt></tt>` snippet

jwortmann:
Fixes #3467 regression introduced in https://github.com/sublimehq/Packages/commit/9e35426d39030880ddf952caab000ec7a5eadf24.

### DIFF
--- a/HTML/html_completions.py
+++ b/HTML/html_completions.py
@@ -491,7 +491,7 @@ class HtmlTagCompletions(sublime_plugin.EventListener):
                 return self.attribute_completions(view, locations[0], prefix)
             return None
 
-        if match_selector("text.html - meta.tag"):
+        if match_selector("text.html - meta.tag, text.html punctuation.definition.tag.begin"):
             # Expand tag and attribute abbreviations
             return self.expand_tag_attributes(view, locations) or self.tag_abbreviations
 


michaelblyons (approved):

jfcherng (approved):

michaelblyons:
Just wait until we find out that someone has made an extension of HTML that sometimes has two-char tag openings. 😛

deathaxe (changes requested):
This change might trigger snippets in "tag-name" position which causes us to end up in double `<<`. 

Animation

deathaxe (approved):
The issue mensioned before seems triggered by prior statement in the method.

PR #3471, opened 2022-08-12
[PHP] Fix quoted PHP strings in tag attributes

deathaxe:
Fixes https://discord.com/channels/280102180189634562/280157121897562114/1006265433244041298


michaelblyons (approved):

jfcherng (approved):

jfcherng:
Just curious about why `with_prototype` gives 
image

when this PR works.

deathaxe:
Given that `with_prototype` also ignores `meta_include_prototype: false` it appears to work fundamentally different than normal `prototype`. Also cache size of smaller examples differ much when comparing both mechanisms.

PR #3472, opened 2022-08-12
[Git Formats] Add Git Commit Message

deathaxe:
This PR adds a separate Git Commit Message syntax which defines commit message subject and body only. 

It is meant for use with Sublime Merge. _Git Log_ makes use of it already.

The goal is consistent highlighting accross sublimehq applications, which has been regressed with SM 2076.

### DIFF
--- a/Git Formats/tests/syntax_test_git_commit_message
+++ b/Git Formats/tests/syntax_test_git_commit_message
@@ -1,0 +1,133 @@
+# SYNTAX TEST "Git Commit Message.sublime-syntax"
+# <- text.git.commit-message comment.line punctuation.definition.comment
+Add git package # no comment (#403)
+# <- meta.subject.git.commit markup.heading.subject.git.commit - meta.expect-subject
+#^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.subject.git.commit markup.heading.subject.git.commit
+#               ^^^^^^^^^^^^^^^^^^^ - comment
+#                             ^ punctuation.definition.reference.issue.git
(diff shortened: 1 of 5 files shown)


keith-hall (approved):

jfcherng (approved):

Git Formats/tests/syntax_test_git_commit_message
```diff
@@ -0,0 +1,136 @@
...
+#                         ^ punctuation.definition.reference.email.begin.git
+#                          ^^^^^^^^^^^^^^^^^^^^ entity.name.reference.email.git
+#                              ^ - punctuation.separator.email.git
+#                                   ^ punctuation.separator.email.git
+#                                          ^ punctuation.separator.domain.git
+#                                              ^ punctuation.definition.reference.email.end.git
+  This commit references a hash: 21dde213
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.message.git.commit
+#                                ^^^^^^^^ constant.other.hash.git
+  This commit does not references a hash: user@21dde213
```
FichteFoll, line 128: I suppose the "comment" on this line here is incorrect since the tests assert hash highlighting for this line.

FichteFoll (approved):

Git Formats/tests/syntax_test_git_commit_message
deathaxe: Those tests are just copies from syntax_test_git_commit, but yes you are right. IIRC, the related pattern has been updated as some code hosters supported it others not (at the time implementing that change).

Not a big deal though.

PR #3474, opened 2022-08-13
[JavaScript] Fix syntax error after img/br/... tags

### LINKED ISSUE #3473: [JavaScript] implicit self-closing html tags break highlighting of JSX
### What happened?

The `img` tag breaks highlighting by never popping `meta.jsx` off stack. Closing `))` are not scoped `punctuation`!

```jsx
    data.slice(data.length - 1).map((review) => (
      <div class="review grid-sm">
        <img src={review.avatar} alt={review.name} class="review-avatar">
      </div>
    ))
//  ^^ punctuation
```

grafik

Adding a `</img>` fixes it. So I guess special treatment for certain kinds of html tags such as `img`, `hr`, `br`, ... is needed or pushing contexts onto stack for each opening tag needs to be removed.

deathaxe:
Fixes #3473 

This commit fixes syntax highlighting after implicitly self-closing html tags such as `` by adding special treatment for well known tags.

> Note: Any more generic approach would be appriciated.

I guess the current solution would fail if someone wants to add `<script src="foo.js">` into its JSX code.

### DIFF
--- a/JavaScript/JSX.sublime-syntax
+++ b/JavaScript/JSX.sublime-syntax
@@ -19,6 +19,9 @@ variables:
   jsx_identifier_break: (?!{{jsx_identifier_part}})
   jsx_identifier: '{{identifier_start}}{{jsx_identifier_part}}*{{jsx_identifier_break}}'
 
+  # implicitly self-closing html tags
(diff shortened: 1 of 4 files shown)


deathaxe:
Closing as implicit closing `` tags are not supported by default JSX/TSX compilers.

PR #3476, opened 2022-09-01
[C++] Detect forwarding headers without extension with nothing but a single #include

ratijas:
Tried in safe mode, works for me.

Can be checked with files like `/usr/include/qt/QtCore/QSet`:

```c++
#include "qset.h"
```

This patch is important to keep LSP GoTo flow non-disruptive. Sending to C++ syntax, because plain C usually doesn't use forwarding headers like this.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -24,6 +24,7 @@ file_extensions:
 first_line_match: |-
   (?xi:
     ^ \s* // .*? -\*- .*? \b(c\+\+|cpp\b) .*? -\*-  # editorconfig
+  | ^ \s* \# \s* include \s* [<"]
   )
 
 variables:


C++/C++.sublime-syntax
```diff
@@ -24,6 +24,8 @@ file_extensions:
 first_line_match: |-
   (?xi:
     ^ \s* // .*? -\*- .*? \b(c\+\+|cpp\b) .*? -\*-  # editorconfig
+  | ^ \s* \# \s* include \s* <
+  | ^ \s* \# \s* include \s* "
```
jfcherng, line 28:
```suggestion
  | ^ \s* \# \s* include \s* [<"]
```

ratijas: makes sense. I originally wanted something like a complete match with filepath and closing bracket/quote, but quickly realized it would be an overkill.

deathaxe:
Is `#include` specific enough to ensure it's C/C++ only? At least Haskell supports C-style includes, too.

ratijas:
> Is `#include` specific enough to ensure it's C/C++ only? At least Haskell supports C-style includes, too.

According to the official documentation:

> #### first_line_match
> When a file is opened without a recognized extension, the first line of the file contents will be tested against this regex, to see if the syntax should be applied.
> [link](https://www.sublimetext.com/docs/syntax.html)

I believe in case of haskell, file extension should take precedence.

deathaxe:
I am aware of how `first_line_match` works. I've just seen several Haskell/Perl/... files in various repos, which are used as scripts without extension.

It's just a question of false positive rates caused by ambiguities.

Just have such issues with regards to file extensions like `.m`, `.r` or `.S` in mind, which are used by some assemblers. While it's easy to re-assign extensions, things get complicated if with false matching `first_line_match`s.

ratijas:
Scripts without extensions should use sheband to disambiguate things. Idk maybe it's just me, but with all due respect to functional programming C++ is still way more popular (at least in my circles), so this assumption is well worth all the possible false-positives.

michaelblyons:
Or you can install ApplySyntax or AutoSetSyntax and make your own rules. 🤷‍♂️

ratijas:
> Or you can install ApplySyntax or AutoSetSyntax and make your own rules. man_shrugging

I didn't get that. I'm trying to upstream my rule here — is there an issue with that? I'd rather avoid extra plugins and configurations unless absolutely needed.

Can anyone please show the real-world usage of extensionless files starting with `#include` lines that are not C++?

deathaxe:
I am not opposed to this proposal in general and agree with you the chance of extension-less files containing `#include` to be of C++ is very high.

Can't however judge the chance of ambiguities realisticly as there are just too many 3rd-syntaxes to know them all.

I'd at least expect possible ambiguities with Objective-C++, which might probably want to try the same. But I don't know that syntax enough to say whether it's realistic or not.

So the question is just: "what if" because a change like this applies globally for all users, even those which probably never use C++ or Qt, but may be effected in negative ways. 

That said, ST has never been batteries included for each and everybodies workflow and use case, and it probably can't.

_If ST's C/C++ syntaxes would be in proper shape already, I'd suggest to create a C++ (Qt) package with inherited syntaxes. It could be installed by users, which want Qt. All others woudn't be effected by Qt relevant special cases then._

ratijas:
First of all, I wanna say I really though that such forwarding headers are way more popular among C++ libs. But I skimmed through `/usr/include` and it turned out that it just mostly Qt and KDE libs using such convention. Weird.

But anyway, C++ standard library has this strange thing for headers without extensions too. Sometimes they are "forwarding" too, with a single `#include` directive, like this one:

_/usr/include/c++/12.2.0/climits:_

```c++
// -*- C++ -*- forwarding header.

// Copyright (C) 1997-2022 Free Software Foundation, Inc.
//
// This file is part of the GNU ISO C++ Library.  This library is free
// [...]
// <http://www.gnu.org/licenses/>.

/** @file include/climits
 *  This is a Standard C++ Library file.  You should @c \#include this file
 * [...]
 */

//
// ISO C++ 14882: 18.2.2  Implementation properties: C library
//

#pragma GCC system_header

#include <bits/c++config.h>
#include <limits.h>

#ifndef _GLIBCXX_CLIMITS
#define _GLIBCXX_CLIMITS 1

#ifndef LLONG_MIN
#define LLONG_MIN (-__LONG_LONG_MAX__ - 1)
#endif

#ifndef LLONG_MAX
#define LLONG_MAX __LONG_LONG_MAX__
#endif

#ifndef ULLONG_MAX
#define ULLONG_MAX (__LONG_LONG_MAX__ * 2ULL + 1)
#endif

#endif
```

(Which also shows that I didn't came up with that "forwarding" terminology myself.) However, neither of them starts outright with `#include` — it's always comments and licensing stuff in the way.

> I'd at least expect possible ambiguities with Objective-C++, which might probably want to try the same. But I don't know that syntax enough to say whether it's realistic or not.

AFAIK in obj-c world frameworks usually follow `#include <Something/Something.h>` naming, and I've never seen an obj-c header without an extension (would that even be properly recognized by any compiler?)

So, I'd still ask to merge this after all. It is highly unlikely to hurt anybody, but would be a nice addition and would save some devs from frustration — without extra (non-existent!) plugins.

deathaxe:
> Which also shows that I didn't came up with that "forwarding" terminology myself.

This has never been part of any concerns.

Please note, `climits` is already highlighted using C++ due to emacs mode switch `// -*- C++ -*-` in line 1.

ratijas:
>> Which also shows that I didn't came up with that "forwarding" terminology myself.
>
> This has never been part of any concerns.

Wasn't sure where did I see that term before, so kinda doubted myself, and felt like I'm saying things no one else understands which leads to this many questions and other concerns.

> Please note, climits is already highlighted using C++ due to emacs mode switch // -*- C++ -*- in line 1.

Well, that's reasonable, I guess, given this insane extension-less situation. Though, I don't think I could convince the whole Qt, and possibly ECM (CMake framework from KDE) to add extra line to their headers.

I still believe it won't do any harm to just merge it. And if there are any counterexamples out there, I'm still waiting to see them.

PR #3478, opened 2022-09-02
[HTML] Support all valid json mime-types

### LINKED ISSUE #3477: [HTML] Missed some highlight for this webpage
### What happened?

webpage: 
[itsfoss.zip](https://github.com/sublimehq/Packages/files/9473486/itsfoss.zip)

here we see white parts, where it must be CSS tokens:
line 9

s1

line 21: missed highlight for JSON block, it's white here

s2

deathaxe:
Fixes #3477

whatwg says:

1. A JSON MIME type is any MIME type whose subtype ends in "+json"
   or whose essence is "application/json" or "text/json".

2. The essence of a MIME type mimeType is mimeType’s type,
   followed by U+002F (/), followed by mimeType’s subtype.

   A MIME type’s type is a non-empty ASCII string.
   A MIME type’s subtype is a non-empty ASCII string. 

see: https://mimesniff.spec.whatwg.org/#understanding-mime-types

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -43,8 +43,13 @@ variables:
       | text/livescript
     )
 
+  # A JSON MIME type
+  # is any MIME type whose subtype ends in "+json"
+  # or whose essence is "application/json" or "text/json".
+  # https://mimesniff.spec.whatwg.org/#understanding-mime-types
   # https://mimesniff.spec.whatwg.org/#json-mime-type
-  json_mime_type: (?i:application/|text/|.+\+)json
+  json_mime_type: |-
+    (?xi: (?:[[:ascii:]]+/)?[[:ascii:]]+\+json | (?:application|text)/json )
 
   # Embedded script and style syntaxes may be wrapped into html comments for
   # historical reasons. The following patterns match them, while maintaining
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

Alexey-T:
Maybe you need [:ascii:] excluding `/` and `>` ?

michaelblyons (dismissed):
New behavior is "more correct" according to spec, but I'm confused why the old regex didn't work in Alexey's bug report. The JSON is highlighted for me.

deathaxe:
> Maybe you need [:ascii:] excluding / and > ?

Not, if we strictly follow the specs.  
_The commit message contains relevant parts of whatwg spec, the implementation follows._

I can't find any restrictions about `type` or `subtype` with regards to `/` or `>` anywhere in the spec.

1. You are right it ends up in arbritary number of `/` being valid, which drives pattern's `type` part reduntant.

2. `>` is a valid character within quoted attribute values. It just terminates a tag if consumed in an unquoted attribute value.
   grafik

---

Another look at the issue arises the question which ST version you use, because ST4134 highlights `application/ld+json` well, even without this PR.

Alexey-T:
ST 4126.

FichteFoll (approved):

deathaxe:
So the json part of #3477 is a duplicate of #3176, which has already been fixed. The only restriction, which is left, was to replace `.+` by `[[:ascii:]]` if we want to be restrictive with regards to specs.

deathaxe:
I've adjusted the pattern accordingly. Keep it simple while restrict to ascii chars.

deathaxe:
Decided to make a new PR to also batch MIME type parameters support.

PR #3481, opened 2022-09-05
[JavaScript] Allow string import/export aliases

Thom1729:
Reported by mitranim on Discord.


mitranim:
JS engines also allow this with double-quoted strings, but not with backtick-quoted strings. Not sure if the PR addresses it since the test only uses single quotes.

Thom1729:
The change works for single- and double-quoted strings, but not templates. I've added a test.

deathaxe (approved):

FichteFoll (approved):

PR #3484, opened 2022-09-20
[meta] update workflow

jrappen:


.github/workflows/ci.yml
```diff
@@ -46,13 +46,31 @@ jobs:
...
+      - name: Clean up for automated testing
+        run: |
+          # remove (some) subdirectories in root directory
+          find . -maxdepth 1 -mindepth 1 -type d -name '.git' -exec rm -rf '{}' ';'
+          find . -maxdepth 1 -mindepth 1 -type d -name '.github' -exec rm -rf '{}' ';'
+          # remove (some) files in root directory
+          find . -maxdepth 1 -mindepth 1 -type f -name '.gitignore' -delete
+          find . -maxdepth 1 -mindepth 1 -type f -name 'LICENSE' -delete
+          find . -maxdepth 1 -mindepth 1 -type f -name 'README.md' -delete
+        shell: bash
```
deathaxe, line 73: What's the benefit of cleaning up test environment?

.github/workflows/ci.yml
```diff
@@ -46,13 +46,31 @@ jobs:
...
+            sublime-build: 4137
             optional: false
 
     steps:
 
+      - name: Git config (global)
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
+          git config --global init.defaultBranch master
```
deathaxe, line 58: Is that needed for `actions/checkout@v3`?

.github/workflows/ci.yml
deathaxe: As the syntax test binary just looks for `syntax_test_...` files, I don't think that kind of extra work is needed. File traversal itself is fast enough maybe even faster than looking for and removing those files first.

.github/workflows/ci.yml
deathaxe: The primary task of actions like `actions/checkout` is to avoid us needing explicitly run shell scripts. If we'd want to be explicit we could just do a git pull via bash manually.

So I'd expect `actions/checkout` to configure whatever is needed to pull required assets for testing.

michaelblyons (approved):

deathaxe (approved):

PR #3488, opened 2022-09-22
[Java] Add support for Java 19

### LINKED ISSUE #3483: [Java] support Java 19 features: `when` (soft) keyword and record patterns
### What happened?

Java 19 [implements](https://openjdk.org/jeps/427) the following syntax:

```java
static void testTriangle(Shape s) {
    switch (s) {
        case null -> 
            { break; }
        case Triangle t
        when t.calculateArea() > 100 ->
            System.out.println("Large triangle");
        default ->
            System.out.println("A shape, possibly a small triangle");
    }
}
```

This should be added. Apart from that, there are now [deconstruction patterns](https://openjdk.org/jeps/405) like this:

```java
static void printColorOfUpperLeftPoint(Rectangle r) {
    if (r instanceof Rectangle(ColoredPoint(Point p, Color c),
                               ColoredPoint lr)) {
        System.out.println(c);
    }
}
```

deathaxe:
Closes #3483

This PR adds support for

1. JEP 394: Pattern Matching for instanceof
2. JEP 405: Record Patterns (Preview)
3. JEP 427: Pattern Matching for switch (preview 3)

JEP 394 has already been added in Java 16, but highlighting hasn't failed badly, so has not been handled in this syntax definition.

With support for JEP 405 type patterns however need a special treatment to correctly handle nested record structure patterns.

JEP 427 builds on top of JEP 394/405 and implements pattern matching for swich-case statements/expressions.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -152,12 +152,15 @@ contexts:
     - include: labeled-statements
 
   expressions:
+    - include: instantiations
+    - include: lambdas
+    - include: constant-expressions
+
+  constant-expressions:
     - include: groups
     - include: literals
     - include: punctuations
-    - include: lambdas
     - include: operators
-    - include: instantiations
     - include: primitive-types
     - include: var-types
     - include: array-modifiers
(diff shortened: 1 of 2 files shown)


ice1000:
Niiice

deathaxe:
I am aware of this one to be a bit more complicated, but is anyone keen enough to take a look into it?

The features it adds are anounced with Java 20 again and it may become more likely users wanting to use them.

michaelblyons (approved):

FichteFoll (approved):

deathaxe:
Thanks!

PR #3492, opened 2022-09-27
[Ruby] Add support for .arb view template files

ocarreterom:
Add support for [Arbre](https://activeadmin.github.io/arbre/) view template files with `.arb` extension.

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -28,6 +28,7 @@ name: Ruby
 scope: source.ruby
 
 file_extensions:
+  - arb
   - rb
   - rbi
   - rbx


ocarreterom:
I am not sure if should be part of the Ruby or the Rails package 🤔

deathaxe:
As it is Rails related, it's probably better suited to assign the extension to _Rails/Ruby (Rails).sublime-syntax_ as it provides some Rails specific syntax extensions, which may be of interest in Arbre templates as well.

ocarreterom:
Changed!

Rails/Ruby (Rails).sublime-syntax
```diff
@@ -6,6 +6,7 @@ scope: source.ruby.rails
 extends: Packages/Ruby/Ruby.sublime-syntax
 
 file_extensions:
+  - arb
```
deathaxe, line 9: The first entry is what is provided as default file extension when saving files. Hence I'd probably keep `ruby.rails` as safe default rather than a special extension.

ocarreterom: I'm agree, I thought in alphabetical order.

michaelblyons (approved):
Works for me.

deathaxe (approved):

Issue #3489, opened 2022-09-25
[C#] Record declaration with curly brackets inside class


nhftk4320:
### What happened?

VS Code:
![image](https://user-images.githubusercontent.com/41443295/192143345-22ce8451-ba0e-4a44-840c-1cfee3b8b157.png)

Sublime Text:
![image](https://user-images.githubusercontent.com/41443295/192139655-1c7e99dd-dc3d-47f3-ae9c-ab8604ba5fcd.png)
The above code is completely valid, and contains an equal number of opening and closing braces. Yet the last is highlighted red.

This issue only occurs under the following conditions:
- Class or Struct parent
- Inner declaration is of record or record struct
- Record contains both parentheses and braces

PR #3493, opened 2022-09-29
[C#] Fix nested record definitions

deathaxe:
Fixes https://github.com/sublimehq/Packages/issues/3489

This commit ...

1. fixes `data_type_body` (now `data_type_expect_block`) to be popped off stack too early.
2. fixes meta scopes of record parameter lists, which used to be `meta.method.parameters`. Now using `meta.class.record.parameters`.
3. improves overall meta scope boundaries for record definitions.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -256,7 +256,7 @@ contexts:
       scope: meta.generic.cs punctuation.definition.generic.begin.cs
       push: type_argument
     - match: ';'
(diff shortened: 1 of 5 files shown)


keith-hall (approved):

keith-hall (approved):

michaelblyons (approved):
I've been using this PR for two weeks with no adverse effects. I don't rely on the feature added, but I don't see anything to worry about.

PR #3496, opened 2022-10-04
[ShellScript] Apply to .env

scop:
This commit makes us apply shell script syntax to `*.env` and `.env` files. These files are commonly used to store environment variables to be sourced or loaded otherwise, in shell syntax, often a simple subset of it.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -14,6 +14,7 @@ file_extensions:
   - bashrc    # e.g.: /etc/bash.bashrc
   - ash
   - zsh
+  - env
 
 hidden_file_extensions:
   - .bash_aliases
@@ -24,6 +25,7 @@ hidden_file_extensions:
   - .bash_profile
   - .bash_variables
   - .bashrc
+  - .env
   - .profile
   - .textmate_init
   - .zlogin


jfcherng (commented):
Make sense to my experience.

ShellScript/Bash.sublime-syntax
```diff
@@ -24,6 +25,7 @@ hidden_file_extensions:
   - .bash_profile
   - .bash_variables
   - .bashrc
+  - .env
```
keith-hall, line 28: I've also seen `.env.local` for environment variables which are only sourced for the local development environment. I'm not sure how common it is and whether it would be worth adding this as well?

scop: No real opinion here, but if we go that route, other similar candidates for consideration would include `.env.test`, `.env.ci`, `.env.dev`, `.env.development`, `.env.stg`, `.env.staging`, `.env.prod`, and `.env.production`.

I suppose something like `.env.*` does not work here?

jfcherng: > I suppose something like `.env.*` does not work here?

No.

scop:
> It's not "real" ShellScript, though:

True. The linked examples are just some implementations, there are more, and there is not a single source of truth for these files. In absence of a dedicated syntax, in my opinion and experience they're in practice either actual shell syntax, or close enough that this association makes sense.

Even if there was a dedicated syntax, I'm not sure it would do better than ShellScript.

jfcherng:
> Even if there was a dedicated syntax, I'm not sure it would do better than ShellScript.

https://packagecontrol.io/packages/DotENV

frou:
If adding more than 1 extension, also consider `.envrc` for [direnv](https://direnv.net/), which is a fairly popular tool.

BTW that's an example of where the files truly are Bash scripts.

deathaxe:
I am uncertain whether we should add too common file extensions to core packages as it is easy to add custom extension bindings via "Always open with..." while its harder to get rid of those which are already pre-defined.

scop:
To be fair, my use case is to get env files colorised by [bat](https://github.com/sharkdp/bat), and I just submitted the change I planned here for consistency. Turns out I hadn't done my homework and they were actually already partially done in bat (using DotENV), so I submitted https://github.com/sharkdp/bat/pull/2356 (and https://github.com/zaynali53/DotENV/pull/17) instead.

Given that DotENV exists and is apparently quite popular, it might be prudent to skip doing this here. Leaving open in case there are other opinions, but I won't pursue this further.

PR #3501, opened 2022-10-06
[Bash] Add .bash_history extension

jfcherng:
Every line in `.bash_history` is a bash command which was executed. It basically constains commands which shows when you press `↑` in a terminal.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -19,6 +19,7 @@ hidden_file_extensions:
   - .bash_aliases
   - .bash_completions
   - .bash_functions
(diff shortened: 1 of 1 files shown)


ShellScript/Bash.sublime-syntax
```diff
@@ -19,6 +19,7 @@ hidden_file_extensions:
   - .bash_aliases
   - .bash_completions
   - .bash_functions
+  - .bash_hisotry
```
keith-hall, line 22:
```suggestion
  - .bash_history
```

keith-hall (changes requested):

deathaxe (approved):

PR #3502, opened 2022-10-07
[PHP] Improve type hints

deathaxe:
This PR mainly optimizes parsing performance of type hints and makes them somewhat more robust with regards to incomplete expressions.

#### Type hints in function parameter lists

10k lines of:

```php
<?
function unionTypeFunction(int | string $param0, Foo | \Foo\Bar | ?int $param1, Foo|\Foo\Bar|?int $param2, (?A|B)&(C|?D)&?E $param2, string $param3, $param4): Foo|\Foo\Bar|?int|static {}
```

| ST4137 | this PR | delta
|--------|---------|--------
| 207ms  | 178ms   | -14%

#### Type hints in typed properties

10k lines of:

```php
<?
class Foo {
    public $var = 0; public int | string $param0; public Foo | \Foo\Bar | ?int $param1; public Foo|\Foo\Bar|?int $param2; public (?A|B)&(C|?D)&?E $param2; public string $param3; public $param4;
}
```

| ST4137 | this PR | delta
|--------|---------|--------
| 304ms  | 277ms   | -9%

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -987,9 +987,12 @@ contexts:
 
   arrow-function-parameters:
     - meta_include_prototype: false
-    - match: \(
-      scope: punctuation.section.group.begin.php
-      set: arrow-function-parameters-body
+    - match: (?=\()
+      set:
+        - arrow-function-parameters-body
+        - function-parameter-variable
+        - expect-type-hint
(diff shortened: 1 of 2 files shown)


jfcherng (approved):
Since it passes all old tests, it's most likely fine.

michaelblyons (approved):

PR #3503, opened 2022-10-07
[Git Formats] Highlight more trailer fields

### LINKED ISSUE #2140: [Git Formats] Should we highlight other trailer fields?
Git commit messages allow for something called a "trailer", which the "Signed-off-by" feature of `git commit -s` utilizes and [`git interpret-trailers`](https://git-scm.com/docs/git-interpret-trailers) formalizes. Its use can be differen from project to project and the interpretation rules can be configured. I didn't find a mention about other trailer formats that git recommends besides "Signed-off-by", but I guess "Co-authored-by" is also quite popular, seeing how even [Github has special handling for it](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors). I've also seen "Reviewed-by".

My question is whether we should match such additional trailer fields and if yes, only specific ones (and which) or try to generalize, e.g. by matching `[\w-]+?-by: `.

deathaxe:
Closes #2140

This commit adds support for more trailer fields beyond `Signed-off-by` and `Change-Id` using a fairly generic approach to match all fields terminated with `-by`.

Note:

Basically any key-value pair at the end of a commit message is treated as trailer field. This rule could be implemented via branches, but let's keep things simple for now.


michaelblyons (approved):
In #2140, I had suggested requiring `ed-by:` for the trailer, but I have now spent some more time thinking about English speech patterns, and realized that `{verb}ed-{preposition}-by:` is also valid English. I don't know if any trailer fields that fit that pattern (`Mucked-about-by:`?), but it's fine to leave that open.

deathaxe:
It was actually `Signed-off-by` which I wanted to match with that generic pattern as well.

michaelblyons:
> It was actually `Signed-off-by` which I wanted to match with that generic pattern as well.

Awesome. That makes a lot of sense.

FichteFoll (approved):

FichteFoll:
I consider this PR to close the mentioned issue, so I edited OP to reflect that and auto-close it once merged.

Issue #3486, opened 2022-09-20
[Python] Type annotations with PEP-604 union types causes 'illegal' highlighting


FichteFoll:
### What happened?

```py
a: str | None = None
b: str | None = 'b'
```

![2022-09-20_17-53-11](https://user-images.githubusercontent.com/931051/191306408-8844b5d3-4202-4a60-8bf5-3f6adf1ed2e5.png)

https://peps.python.org/pep-0604/

This might be a regression because I think it wasn't like that before.

jfcherng:
Maybe similar to https://github.com/sublimehq/Packages/issues/3290

kaste:
I'm really skeptical when the syntaxes want to be fast linters.  `None = None` produces a syntax error in flake8 and mypy.  

IMO `invalid.` should be used (like a raise/panic) when the syntax cannot solve the puzzle anymore.  It cannot produce valid or sensible colors here and maybe on the following line(s) anymore so it marks red.  That also means as the syntaxes improve here over time we get _less and less_ invalid reds/states.  (I remember modern Javascript and ST3 produced a lot red states.)  If we do the linters work here, we do get more and more red states.  We should not do this.  We also see that the invalid states are rendered usually unreadable.  That should guide us.

PR #3506, opened 2022-10-20
[Python] Mark variable annotation with a `meta.` block

kaste:
Use ~`meta.variable.parameters.annotation.python`~ `meta.variable.annotation.python` to mark annotation blocks when assigning.

This closely follows the `meta.function.parameters.annotation.python` we have in the function signatures.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1018,15 +1018,17 @@ contexts:
     - include: illegal-assignment-expression
     - match: ':'
       scope: punctuation.separator.annotation.variable.python
-      push:
-        - meta_scope: meta.variable.annotation.python
-        - match: (?=$|=|#|;)
-          pop: 1
-        - include: expression-in-a-group
+      push: variable-annotation
     - match: \+=|-=|\*=|/=|//=|%=|@=|&=|\|=|\^=|>>=|<<=|\*\*=
       scope: keyword.operator.assignment.augmented.python
     - match: '=(?!=)'
       scope: keyword.operator.assignment.python
+
+  variable-annotation:
+    - meta_scope: meta.variable.annotation.python
+    - match: (?=$|=|#|;)
+      pop: 1
+    - include: expression-in-a-group
 
   operators:
     - match: <>
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1019,7 +1019,7 @@ contexts:
     - match: ':'
       scope: punctuation.separator.annotation.variable.python
       push:
-        - meta_scope: meta.variable.parameters.annotation.python
+        - meta_scope: meta.variable.annotation.python
         - match: '(?==|$|#)'
           pop: 1
         - include: expression-in-a-group
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1020,7 +1020,7 @@ contexts:
       scope: punctuation.separator.annotation.variable.python
       push:
         - meta_scope: meta.variable.annotation.python
-        - match: '(?==|$|#)'
+        - match: (?=$|=|#|;)
           pop: 1
         - include: expression-in-a-group
     - match: \+=|-=|\*=|/=|//=|%=|@=|&=|\|=|\^=|>>=|<<=|\*\*=
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1016,8 +1016,15 @@ contexts:
 
   assignments:
     - include: illegal-assignment-expression
-    - match: ':'
-      scope: punctuation.separator.annotation.variable.python
+    - match: '(?=:)'
+      set:
+        - match: ':'
+          scope: punctuation.separator.annotation.variable.python
+          set:
+            - meta_scope: meta.variable.parameters.annotation.python
+            - match: '(?==|$|#)'
+              pop: 1
+            - include: expression-in-a-group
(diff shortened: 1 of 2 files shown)


kaste:
E.g. the function signature could (always) be styled as  

image

After this patch, simple assignements can follow 

image

This is just a clear-as-possible example.  Just setting a background will keep the normal syntax highlighting of the tokens in-tact. 

Note.  I wanted to add tests for 

```
foo: \
    bar = zoo

zoo: \
    loo \
    = too
```
 which works

image

but as soon as I start the test line with an `#` the pattern pops (because it must pop for python comments; I added that.)

In other words it makes this

image

and `bar = zoo` are not marked anymore.

Python/Python.sublime-syntax
```diff
@@ -1016,8 +1016,17 @@ contexts:
 
   assignments:
     - include: illegal-assignment-expression
-    - match: ':'
-      scope: punctuation.separator.annotation.variable.python
+    - match: '(?=:)'
+      set:
```
deathaxe, line 1020: 1. This would need to be a `push` as `assignments` is included in `statements` which is part of `main`.

2. Why adding the lookahead?

Python/Python.sublime-syntax
```diff
@@ -1016,8 +1016,17 @@ contexts:
...
+        - match: ':'
+          scope: punctuation.separator.annotation.variable.python
+          set:
+            - meta_scope: meta.variable.parameters.annotation.python
```
deathaxe, line 1024: If we expect annotations to be primarily used for type-hints existing `meta.function.parameters.annotation` is not very suiteable anymore. The concept of `parameters` in assignments feels even less appropriate, IMHO.

`meta.annotation` would generally be a good scope, but it is unfortunatelly used/reserved for member attributes/modifiers in various syntaxes

example:

```py
@annotation
def func():
    pass
```

I have not final proposal for a suitable scope, but `meta.type.php` was introduced for type annotations in PHP recently. Maybe we want to use something like that for all kinds of type hints - also in function parameter lists and return types.

Python/Python.sublime-syntax
```diff
@@ -1016,8 +1016,17 @@ contexts:
...
+            - match: '(?==|$|#)'
+              pop: 1
+            - include: expression-in-a-group
```
deathaxe, line 1027: This is a good starting point, but I guess we want to use the same context used by function parameters to also fix #3486. I've starded work for such a fix at https://github.com/deathaxe/sublime-packages/tree/pr/python/add-type-hints some weeks ago, but got stuck due to work on Package Control and real life.

deathaxe:
In case of testing line continuation - it requires several tests to test such constructs. First test testing line 1, 2nd test testing line 2 (with content of 1 included) and so forth

```
vari = expr \
#           ^ punctuation


vari = expr \
  expr1 \
#       ^ punctuation

vari = expr \
  expr1 \
  expr2 \
#       ^ punctuation

...
```

Python/Python.sublime-syntax
kaste: Both because it is copied from `function-parameters: `.  I push a change so we can compare.  Don't know if I get this to work actually.

kaste: Ok, I pushed a commit and it checks green.

Python/Python.sublime-syntax
kaste: It was meant as a babystep anyway.  I do not like the "invalid" markers actually; they tend to produce problems in the distant future; and that is a case here probably.

Python/Python.sublime-syntax
kaste: Yeah `meta.variables.parameters` is non-sense.  Simply `meta.variable.annotation.python` is a better fit here.

We should not restrict to type-hints though.  These still can be used for meta-programming.

deathaxe (approved):

Python/Python.sublime-syntax
```diff
@@ -1018,6 +1018,11 @@ contexts:
     - include: illegal-assignment-expression
     - match: ':'
       scope: punctuation.separator.annotation.variable.python
+      push:
+        - meta_scope: meta.variable.annotation.python
+        - match: '(?==|$|#)'
```
deathaxe, line 1023: We should probably include `,` and `;` to make sure to pop annotations in lists or multiple inline expressions

```suggestion
        - match: $|(?=[,;=#])
```

```
var1: Type1, var2: Type2;print("foo")
```

kaste: The `,` is a syntax error here.  

image

The `;` is allowed though and should pop of course.

kaste: Done, still green

Python/Python.sublime-syntax
```diff
@@ -1018,6 +1018,11 @@ contexts:
     - include: illegal-assignment-expression
     - match: ':'
       scope: punctuation.separator.annotation.variable.python
+      push:
+        - meta_scope: meta.variable.annotation.python
+        - match: (?=$|=|#|;)
+          pop: 1
+        - include: expression-in-a-group
```
FichteFoll, line 1025: Could you introduce a named context for this? We generally try to avoid inline contexts since syntax definitions can inherit/extend others since it allows more finely tuning the exact contexts you need.

FichteFoll (changes requested):

Python/Python.sublime-syntax
kaste: Done

FichteFoll (approved):

deathaxe (approved):

PR #3508, opened 2022-10-22
[LaTeX] Fixes for newcommand and DeclareMathOperator

ngc92:
For the brace-less version, i.e. `\newcommand\foo{\bar}`, the old syntax did not pick up more than a single character of the control sequence, and the assigned style differed from the brace-version (i.e. captures 4 and 6 were different)

LaTeX is much more lenient regarding the placement of whitespace than the current syntax highlighting allows. Therefore, I've added arbitrary-amount white-space matching between elements in the regex. 
Note that for `\DeclareMathOperator`, I've tried `\DeclareMathOperator{ \name}{}`, but this causes an error. I'm not sure if that is intended behaviour, or just an oversight, but I've refrained from allowing spaces there -- I'm not sure, though, if the syntax-highlighting should pick up on this specific quirk. As in, for another programming language, if we knew that calling a function with a particular argument always causes the program to crash, we would still color this as a valid function call.

Finally, I've also removed the [] optional argument for `\DeclareMathOperator`, since this function does not take allow for an optional.


LaTeX/LaTeX.sublime-syntax
```diff
@@ -112,23 +112,29 @@ contexts:
...
         (?:
-          (\{)(\\[A-Za-z@]+)(\})
-          | (\\[A-Za-z@])+
+          (\{)\s*(\\[A-Za-z@]+)\s*(\})
+          | (\\[A-Za-z@]+)
         )
-        (?:(\[)(?:[^\]]*)(\]))*
+        \s*
+        # optional argument specification in square brackets
+        (?:(\[)(?:[^\]]*)(\])\s*)*
```
deathaxe, line 128: The final `*` operator says there are possibly multiple bracket pairs. ST will only capture and scope one of them.

grafik

If the max amount of bracket pairs is known it may be enough to add some capture groups for each. 

A more modern approach however would be to refactor the pattern to push contexts for each part on stack. It would also improve UX by handling incomplete expressions better.

deathaxe:
> Note that for `\DeclareMathOperator`, ... I've refrained from allowing spaces there -- I'm not sure, though

Basically a syntax definition's task is not to do linting, even though some syntaxes of this repo try to do so. Maybe because they were developed at a time LSP or linters were not widely available in ST. In some cases highlighting illegal syntaxes also helps new users to find trivial bugs in their code.

In general however we need to keep in mind - an editor's task is to provide good UX while writing code. It's the nature of incomplete code to be invalid. Therefore an overeager syntax would produce many illegal (red?) code while writing, which feels bad.

A syntax definition's priority should therefore be to be as permissive as possible with regards to incomplete or intermediatly wrong expressions - to create the impression ST would already know what come's next and always highlight tokens well.

Therefore I'd say it is ok to capture whitespace between bracktes and content everywhere. If in doupt it could be captured and highlight illegal. The concequences of not highlighing anything just because of an extra space are worse.

LaTeX/LaTeX.sublime-syntax
ngc92: To keep this changelist small, I'd go with the simple approach for now. LaTeX allows at most two optional arguments, 
the first specifying how many arguments the new command should take, and the second giving a default value for the first argument. 
What should be the scopes for these? 
`meta.function.parameters.default-value.python` for the default value, I guess?

ngc92:
I've changed it so that it allows the spaces within `\DeclareMathOperator` now.

LaTeX/LaTeX.sublime-syntax
ngc92: In the long run, I think it really would be necessary to split this into several contexts. The current approach, for example, also fails in situations like
```LaTeX
\newcommand\baz[1][dificu{]}t] {\textbf{#1}}
```
where the default value contains a closing bracket, escaped within a brace group.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -136,8 +140,12 @@ contexts:
         5: punctuation.definition.group.brace.end.latex
         6: support.function.latex entity.name.newcommand.latex
         7: punctuation.definition.group.bracket.begin.latex
-        8: punctuation.definition.group.bracket.end.latex
-        9: punctuation.definition.group.brace.begin.latex
+        8: constant.numeric.value.latex
```
deathaxe, line 143: Maybe not high prio in LaTeX, but numbers adding a `meta.number` scope to comply with other syntaxes of this repo would be appreciated.

```suggestion
        8: meta.number.integer.decimal.latex constant.numeric.value.latex
```

ngc92:
OK, so I've looked a bit more into this, if you like I can try to update this to a somewhat larger PR that splits the newcommand parsing into several contexts.

It seems to me that the scoping here is also a bit different than what other languages do, as the `function` scope is applied also to the "function body". That seems to be different in python, for example. 

Making this parsing work for all valid LaTeX `\newcommand` constructions seems almost impossible, as LaTeX allows comments almost everywhere. I think, at least for now, allowing comments between elements, but not within elements, should be fine. That is, I would enable something like 
```LaTeX
\newcommand % split long
\name % definitions over
[1]       % multiple lines
{text # 1}
```
but not 
```LaTeX
\newcommand{\foo}%
[%
3%
]%
{test #1 #2 #3}
```

deathaxe:
LaTeX hasn't been updated significantly for a long time. Hence it's not surprising some scopes are not up to date.

Feel free to put some more work into the PR.

ngc92:
I've updated the parsing to use several context. This means that even pretty gnarly versions of newcommand will be parsed correctly, such as this test case:
```LaTeX
\newcommand\baz % 
%  ^ meta.function.latex
%          ^^^^ entity.name.newcommand.latex
  [1] 
%  ^ constant.numeric.value.latex
 [difi % other
%^^^^^ meta.function.parameters.default-value.latex
%      ^^^^^^^^ meta.function.parameters.default-value.latex comment.line.percentage.tex
  cu{]}t] {\textbf{#1}}
% ^^^^^^^ meta.function.parameters.default-value.latex
%         ^^^^^^^^^^^^^ meta.function.definition.latex
```

That is more robust than I had originally hoped for :)

I'd like to improve the checking in the test cases a bit still, but before that, could I get some feedback on the scopes used here?
I'm not sure if `storage.modifier.newcommand.latex` makes sense, semantically, but it is in use currently, and has an effect on syntax highlighting. Personally, I like that commands that create other commands are highlighted differently from regular latex commands, so I'd like to keep that. 

The old scopes also made it possible to distinguish the different command-generating commands, i.e. whether we are now defining a new name as newcommand or DeclareMathOperator. This is no longer possible with the reusable scopes. I'm not sure if that was a useful feature to begin with, though.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,79 +110,89 @@ contexts:
...
-        \s*
-        # optionally, the number of arguments. LaTeX allows only up to 9, so a single digit
-        (?:(\[)\s*(\d)\s*(\]))?
-        \s*
-        # optionally, a default argument value
-        (?:(\[)([^\]]*)(\]))?
-        \s*
-        # start of actual command definition
-        (\{)
+    - match: '(\\)(?:(?:new|renew|provide)command\*?)'
```
michaelblyons, line 113: No need for the double wrapping of no-capture group.

michaelblyons: I'm ambivalent about the trailing storage modifier. I know it was in the old one.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,79 +110,89 @@ contexts:
...
-          (\{)\s*(\\[A-Za-z@]+)\s*(\})
-          | (\\[A-Za-z@]+)
-        )
-        \s*
-        (\{)
+
+    # if we enounter anything that is not an opening bracket, whitespace, or
+    # a comment, then there are no argument specifications and we can continue
+    # with the definition
+    - match: '(?=\S)'
```
michaelblyons, line 153: I suggest pulling this out into a utility include, if there isn't one already. I think this one has a particular name in other syntax files, maybe `immediately-pop`? Then include where relevant.

michaelblyons (changes requested):
Thanks for giving LaTeX some attention!

ngc92:
I've found two more corner cases where the current approach breaks:
First, if the argument number is specified indirectly, such as in this test case:
```
\def\one{1}
\newcommand{\weirder}[\one] [default] {this is #1 arg}
```
second, newcommand also allows the new definition to be a single token, that then need not be surrounded by braces, e.g.
```
\newcommand\foo\bar
```

I've added support for the former (though the direct literal, which will be the vast majority of cases, receives more specific scoping)
and would like to support the latter. However, that can be a bit annoying, because when you are typing, and the next element in the text is another command, then the highlighting might change.

To prevent this, and also to better model the syntax of LaTeX, I would like to introduce a bail-out option that pops the scope if there is a blank line -- which corresponds to a new paragraph in LaTeX. I cannot get this to work, though, `^\s*$` doesn't seem to do the job.

michaelblyons:
> I cannot get this to work, though, `^\s*$` doesn't seem to do the job.

I did ` - match: ^(?=\r?\n)` in one of mine. Maybe adding a `\s*` in there will work?

ngc92:
> > I cannot get this to work, though, `^\s*$` doesn't seem to do the job.
> 
> I did ` - match: ^(?=\r?\n)` in one of mine. Maybe adding a `\s*` in there will work?

thanks, that did it, now the contexts are correctly exited at newlines.

I don't think this is a regression I've introduced in this PR, but I've tried the `Symbol List`, and it appears that it does not work. Here is a minimal example:
```LaTeX
\documentclass{article}
\newcommand{\test}{test}
\begin{document}
  \test
\end{document}
```
Interestingly, if I change the global scope from `text.tex.latex` to `text.latex`, it works. Any ideas?

LaTeX/LaTeX.sublime-syntax
```diff
@@ -83,17 +83,25 @@ contexts:
...
     - include: main
 
+  # matches any nospace and pops the context
+  immediate-pop:
+    - match: '(?=\S)'
+      pop: true
+
+  # pops out of the current context if there is a new paragraph
+  paragraph-pop:
+    - match: ^(?=\s*\r?\n)
```
deathaxe, line 93: We don't need to care about \r in context of syntax definitions as line feeds are normalized by ST. Using `$` instead of `\n` may be needed in case paragraph end needs to be matched at the end of a file as well.

```suggestion
    - match: ^(?=\s*$)
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -83,17 +83,25 @@ contexts:
       pop: true
     - include: main
 
+  # matches any nospace and pops the context
+  immediate-pop:
```
deathaxe, line 87: We started using such helper contexts in various syntaxes. I'd like to keep their naming and behavior somewhat equal to avoid bad habit bugs.

This one would be

```suggestion
  else-pop:
```

... with regards to its content.

An `immediately-pop` would look like

```yaml
  immediately-pop:
    - match: ''
      pop: true
```

deathaxe (commented):
Two tiny suggestions. May have a closer look on it later.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,67 +118,120 @@ contexts:
...
+
+  newcommand-argspec:
+    - include: paragraph-pop
+    # direct specification of the argument count: This is what we usually expect,
+    # and what we want to support well
+    - match: '(\[)\s*(\d)\s*(\])'
+      scope: meta.function.parameters.latex
+      captures:
+        1: punctuation.definition.group.bracket.begin.latex
+        2: constant.numeric.value.latex meta.number.integer.decimal.latex
```
deathaxe, line 146: We always try to first apply all meta scopes, followed by normal ones.

```suggestion
        2: meta.number.integer.decimal.latex constant.numeric.value.latex
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,67 +118,120 @@ contexts:
...
+    # however, we might also get weirder argument specifications. Most of them
+    # will be wrong, but determining that would be very difficult, so we're just
+    # very lenient here and accept anything enclosed in brackets.
+    - match: '\['
+      scope: punctuation.definition.group.bracket.begin.latex
+      push:
+        - meta_scope: meta.function.parameters.latex
+        - match: '\]'
+          scope: punctuation.definition.group.bracket.end.latex
+          pop: 2
```
deathaxe, line 159: Adding a pattern to match a number within brackets could help to avoid duplicated patterns for brackets (by being able to remove the above one L#142-148).

```suggestion
        - match: '\]'
          scope: punctuation.definition.group.bracket.end.latex
          pop: 2
        - match: \d+\b
          scope: meta.number.integer.decimal.latex constant.numeric.value.latex
```

Just considder a construct like ...

```latex
\newcommand{\foo}[
  1
]
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,67 +118,120 @@ contexts:
...
+    - match: '(\\)DeclareMathOperator\*?'
+      scope: meta.function.latex support.function.declare-math-operator.latex storage.modifier.newcommand.latex
+      captures:
+        1: punctuation.definition.backslash.latex
+      push:
+        - newcommand-definition
+        - newcommand-commandname
+
+  newcommand-argspec:
+    - include: paragraph-pop
```
deathaxe, line 139: As `paragraph-pop` doesn't contain overlapping patterns, we could easily move it to the end of the context - just before `- include: else-pop`. This way the parser tries to match a `[` which is the most likely expected token at this position.

_Note: It applies to all \newcommand related contexts._

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,67 +118,120 @@ contexts:
...
         - include: general-constants
         - include: general-commands
         - include: macro-braces
 
+    # but in principle, a single token is also valid,
+    # thus either a command sequence ...
+    - match: '(\\)[A-Za-z@]+'
+      captures:
+        1: punctuation.definition.backslash.latex
+      scope: meta.function.definition.latex support.function.general.latex
```
deathaxe, line 221: The `scope` should always be the first before `captures` or other keys follow. Just to have a common scheme of how rules are organized.

```suggestion
      scope: meta.function.definition.latex support.function.general.latex
      captures:
        1: punctuation.definition.backslash.latex
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,67 +118,120 @@ contexts:
...
+      pop: true
+
+    # or a single, potentially escaped character. no further effort at
+    # differentiation here
+    - match: '(\\)?\S\b'
+      captures:
+        1: punctuation.definition.backslash.latex
+      scope: meta.function.definition.latex
```
deathaxe, line 229: Same here...

```suggestion
      scope: meta.function.definition.latex
      captures:
        1: punctuation.definition.backslash.latex
```

LaTeX/syntax_test_latex.tex
```diff
@@ -28,31 +28,120 @@
...
+%            ^^^^ entity.name.newcommand.latex
+%                     ^ support.function.general.latex
+
+\newcommand* {\foo }{\bar}
+%   ^ meta.function.latex
+%             ^^^^ entity.name.newcommand.latex
+%                      ^ support.function.general.latex
+
+\newcommand \foo {\bar}
+%   ^ meta.function.latex
```
deathaxe, line 81: If spaces are allowed, we should ensure to maintain an uninterrupted `meta.function` scope covering the whole statement.

```suggestion
\newcommand \foo {\bar}
%^^^^^^^^^^^^^^^^^^^^^^ meta.function
```

This is mainly achieved by `meta_scope` to also cover intermediate comments.

LaTeX/LaTeX.sublime-syntax
ngc92: done

LaTeX/LaTeX.sublime-syntax
ngc92: I've applied this change, and also updated the test to have the [ and number on different lines to check this case.

LaTeX/syntax_test_latex.tex
deathaxe: The meta scope is probably the most tricky part - as in other syntaxes as well. You are right with stacking `meta.function` not being desirable even though some older syntaxes still do something like that.

The bracket contexts could use `clear_scope: 1` to remove a possibly global `meta.function.latex` meta scope. The `newcommand-definition` context may require special treatment for the parts not surrounded by brackets - in case `meta.function.definition` should keep as a special scope.

If it denotes the function body, it could also be just scoped `meta.function.latex`. This is what most other syntaxes do as well. They use `meta.function meta.block` most likely.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,67 +118,111 @@ contexts:
...
+        - newcommand-commandname
+    - match: '(\\)DeclareMathOperator\*?'
+      scope: meta.function.latex support.function.declare-math-operator.latex storage.modifier.newcommand.latex
+      captures:
+        1: punctuation.definition.backslash.latex
+      push:
+        - newcommand-definition
+        - newcommand-commandname
+
+  newcommand-argspec:
```
deathaxe, line 138: May the context be re-used for other function definitions? If so maybe a more common name may be useful?

LaTeX/syntax_test_latex.tex
ngc92: thanks, I wasn't really aware of `clear_scope`. That certainly makes things easier. Yes, I think the part in braces generally corresponds to a function body. Though you could also interpret it like e.g. the definition of a macro in the C family of languages. 

In any case, I'm not really sure how to implement this in the syntax file. Since we are pushing multiple contexts, I cannot set a meta_scope there, right?

LaTeX/LaTeX.sublime-syntax
ngc92: There is also the \NewDocumentCommand family of functions. These are from the xparse package, but have recently been included into the LaTeX kernel. They have a different way of specifying arguments, but the function definition and command-name parts may be the same. 
On the other hand, AFAIK, the Plain-TeX \def command actually requires that its command definition be enclosed in braces, so that is slightly different. At this point, I would keep the names as they are.

LaTeX/syntax_test_latex.tex
deathaxe: Basically there are 2 major ways of adding seperate meta scopes for each sub-context.

1. push the first context and than set all the others one after another using lookahead patterns. Each of them can then have its own meta scope without the previous one overlapping.
2. use a multi-push statement (as you did here) and add a common scope meta scope to the first one being pushed (the last remaining on stack; `newcommand-definition`) and clear it whenever needed in the other contexts.

The latter is the somewhat more "modern" approach as chances are highier to get re-usable contexts when designing multi-push patterns. Results are more or less the same performance-wise.

I'll try to work out a suggestion how to do that.

deathaxe: A suggestion is posted at https://github.com/ngc92/SublimeTextPackages/pull/1

ngc92: @deathaxe Thank you very much. One question, though: You have removed the '' around the match patterns.
Personally, I'm not really a fan of that, because it leads to inconsistencies, e.g. the `immediately-pop` context still uses `'`.
Is this a style convention that you have decided to use for all syntax files? If that is the case, then all the other `match` patterns within the `TeX` and `LaTeX` syntaxes should also be updated, otherwise we have an inconsistent style within the file itself.

deathaxe: Over time a habit/coding-style has evolved which includes removing as many quotation/punctuation as possible. It's not a strong language convention though.

While most patterns in LaTeX still use single quotes which may be a result of automatic convertion from tmLanguage format most of them are not really needed.

We could also quote scope names or includes, but we don't do that. The same may apply to patterns. Sure, some may need extra escaping or still require quotes, but it doesn't really justify to quote everything.

Just have a look at other syntaxes, like Java, PHP, ... which have been recently rewritten. Quotes are almost completely gone.

This change could also be handled in a separate PR, if it feels better to you.

ngc92: I think its fine to do it in this PR then, so long as we immediately afterwards have a second PR that strips the other quotes to make the files' style internally consistent again.

deathaxe: I would appriciate that.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,66 +122,128 @@ contexts:
...
+    - match: \[
+      scope: punctuation.definition.group.bracket.begin.latex
+      push:
+        - clear_scopes: 1
+        - meta_scope: meta.function.parameters.latex
```
deathaxe, line 147: What we probably should also try to do is to avoid anonymous contexts. 

The 2 major reasons are:

1. support inheritance: only named contexts can be addressed in case a 3rd-party syntax wants to extend one of the contexts.
2. diagnostics: ST can display the context which is involved in highlighting via `ctrl+shift+alt+p`. Anonymous contexts are hard to find there.


_Note: I haven't made that change in my probosal as I wanted to concentrate on meta the scope._

ngc92: With this change, the `newcommand-block-body` is essentially the same as the anonymous context pushed in `macros` from `TeX.sublime-syntax`, except for the scope names. Could we unify that in a sensible way?

ngc92:
Is there anything else to do here for this change? The discussion seems to have mostly moved to more general improvements to the (La)TeX syntaxes not necessarily related to this PR.

deathaxe (dismissed):
I think we can go with it.

deathaxe:
No, they aren't gated by v2 syntax as they don't brack backward compatibility.

We probably should try to replace all `pop: true` by `pop: 1` and try to move on to v2 syntax in near future. But one step after the other.

michaelblyons (approved):

ngc92:
When working on similar changes for TeX, I noticed that there was still one problem with the new definitions. We need to make sure that we actually pick up a command `\newcommand`, and not just any command that starts with `\newcommand` (however unlikely that might be in real code), so I've added a look-ahead here.

michaelblyons (approved):

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,66 +122,128 @@ contexts:
...
-          (\\)
-          (?:(?:new|renew|provide)command\*?)
-        )
-        (?:
-          (\{)(\\[A-Za-z@]+)(\})
-          | (\\[A-Za-z@])+
-        )
-        (?:(\[)(?:[^\]]*)(\]))*
-        (\{)
+    - match: (\\)(?:new|renew|provide)command\*?(?=[^a-zA-Z@])
```
deathaxe, line 125: Negative lookaheads have the charm to also enable highlighting a valid token at eof. In this special case where `*` may terminate a command I wonder if an alternation would be better suited.

```suggestion
    - match: (\\)(?:new|renew|provide)command(?:\*|(?![A-Za-z@]))
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -110,66 +122,128 @@ contexts:
...
       push:
-        - meta_scope: meta.function.newcommand.latex
-        - match: \}
-          scope: punctuation.definition.group.brace.end.latex
-          pop: true
+        - newcommand-definition
+        - newcommand-optarg
+        - newcommand-argspec
+        - newcommand-commandname
+    - match: (\\)DeclareMathOperator\*?(?=[^a-zA-Z@])
```
deathaxe, line 134: It's the same here.

deathaxe (approved):

michaelblyons (approved):

PR #3510, opened 2022-10-23
[Rust] Various tweaks and fixes to type matching

FichteFoll:
See commits for more details.

|Before|After|
|:-|:-|
|||

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -357,17 +357,6 @@ contexts:
     - match: '(?=\S)'
       pop: true
 
-  support-type:
-    - match: '(Vec|Option|Result|BTreeMap|HashMap|Box|Rc|Arc|AsRef|AsMut|Into|From)\s*(?=<)'
-      scope: support.type.rust
-      push: generic-angles
-    - match: '[?]?(?=\bSized\b)'
-      scope: keyword.operator.rust
-    - match: '[!]?(?=\bSync|Send\b)'
-      scope: keyword.operator.rust
-    - match: \b(Copy|Send|Sized|Sync|Drop|Fn|FnMut|FnOnce|Box|ToOwned|Clone|PartialEq|PartialOrd|Eq|Ord|AsRef|AsMut|Into|From|Default|Iterator|Extend|IntoIterator|DoubleEndedIterator|ExactSizeIterator|Option|Some|None|Result|Ok|Err|SliceConcatExt|String|ToString|Vec)\b
-      scope: support.type.rust
-
   return-type:
     - match: '\bimpl\b'
       scope: storage.type.impl.rust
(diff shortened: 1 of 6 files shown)


Rust/Rust.sublime-syntax
```diff
@@ -633,24 +612,67 @@ contexts:
           pop: true
         - match: ','
           scope: punctuation.separator.rust
-        - include: type-any-identifier
+        - include: types-any
     - match: '\+'
       scope: keyword.operator.rust
+    - match: '\b_\b'
```
keith-hall, line 618: What does this operator do? If it's the same as in other languages, I believe we scope it differently: https://github.com/sublimehq/Packages/commit/83a8f3e886b56c5c87957adb5c5f2ab61833f3b0

I realize you only moved this part, but would be nice to improve this at the same time if relevant :)

keith-hall (dismissed):

Rust/Rust.sublime-syntax
```diff
@@ -633,24 +612,67 @@ contexts:
...
+  type-any-identifier-path:
+    - meta_scope: meta.path.rust
+    - match: '::'
+      scope: punctuation.accessor.rust
+      pop: true
+    - include: type-identifiers
+
+  type-any-identifiers:
+    - match: '[!?]'
+      scope: storage.modifier.rust
```
jwortmann, line 641: Can those symbols get a unique subscope, so that they can be distinguished from actual modifier keywords (here `const`, `pub`, `mut`, `ref`, ...), which already use `storage.modifier.rust`? I would want to highlight the `!` and `?` different in my color scheme, just for the reason that they are symbols.

I don't use Rust and don't know about the meaning here, but for example the TypeScript syntax uses `storage.modifier.optional` and `storage.modifier.definite` when those symbols are used as a modifier suffixes. Or PHP uses `storage.type.nullable` for `?`. Perhaps one of these subscopes can be reused, or you can find good alternatives which fit here.

Rust/Rust.sublime-syntax
```diff
@@ -633,24 +612,67 @@ contexts:
...
         - match: ','
           scope: punctuation.separator.rust
-        - include: type-any-identifier
+        - include: types-any
     - match: '\+'
       scope: keyword.operator.rust
+    - match: '\b_\b'
+      scope: keyword.operator.rust
+    - match: '!(?!\w)'
+      scope: storage.type.rust
```
jwortmann, line 621: Would be great for this to get a unique subscope too which describes what this `!` type means, so that it's possible to highlight it different from builtin type names and other alphanumeric type names.

Rust/Rust.sublime-syntax
FichteFoll: It's used as a placeholder for type inference and thus has a special role in type signatures. I tried to formulate that into a proper `storage` scope, PTAL.

See also https://stackoverflow.com/questions/61950463/what-is-an-underscore-in-rust-type.

keith-hall (approved):

deathaxe (approved):

PR #3514, opened 2022-10-30
[JSON] Add .sublime_session extension

jfcherng:


JSON/JSON.sublime-syntax
```diff
@@ -18,6 +18,7 @@ file_extensions:
   - sublime-settings
   - sublime-theme
   - sublime-workspace
+  - sublime_session
```
FichteFoll, line 21: Imo this should be a hidden extension because you're not supposed to manually create such files (hidden extensions are not proposed in the save as dialog).

FichteFoll (approved):

deathaxe (approved):

PR #3526, opened 2022-11-07
[CSS] add typing info, fix typos

jrappen:


CSS/css_completions.py
```diff
@@ -1,44 +1,58 @@
-import re
+from __future__ import annotation
+
 import sublime
-import sublime_plugin
+import sublime_plugi
```
michaelblyons, line 4: ??

deathaxe:
I dislike the typing infos.

rchl:
IMO type hints for function arguments and return values should be useful for those who work on this code but annotating variables in most cases is completely redundant and unnecessary.

CSS/css_completions.py
```diff
@@ -62,7 +69,12 @@ def re_value(self):
         return re.compile(r"^(?:\s*(:)|([ \t]*))([^:]*)([;}])")
 
     @timing
-    def on_query_completions(self, view, prefix, locations):
+    def on_query_completions(
+        self,
+        view: sublime.View,
+        prefix: str,
+        locations: typing.List[sublime_types.Point]
+    ) -> typing.Union[None, typing.List[sublime_types.CompletionValue], typing.Tuple[typing.List[sublime_types.CompletionValue], sublime.AutoCompleteFlags], sublime.CompletionList]:
```
rchl, line 77: I believe for this code this is enough.
```suggestion
    ) -> typing.Union[None, sublime.CompletionList]:
```

CSS/css_completions.py
```diff
@@ -90,8 +102,8 @@ def on_query_completions(self, view, prefix, locations):
             return sublime.CompletionList(items, sublime.INHIBIT_WORD_COMPLETIONS)
         return None
 
-    def complete_property_name(self, view, prefix, pt):
-        text = view.substr(sublime.Region(pt, view.line(pt).end()))
+    def complete_property_name(self, view: sublime.View, prefix: str, pt: sublime_types.Point):
```
rchl, line 105: Return value should be annotated.

Also should change the return statement to returns a list instead of a generator as I don't think it was intentional to return generator here and it also makes types invalid for `sublime.CompletionList` constructor.

CSS/css_completions.py
```diff
@@ -90,8 +102,11 @@ def on_query_completions(self, view, prefix, locations):
             return sublime.CompletionList(items, sublime.INHIBIT_WORD_COMPLETIONS)
         return None
 
-    def complete_property_name(self, view, prefix, pt):
-        text = view.substr(sublime.Region(pt, view.line(pt).end()))
+    def complete_property_name(
+        self, view: sublime.View, prefix: str, pt: sublime_types.Point
+    ) -> typing.List[sublime.CompletionItem]:
+
+        text: str = view.substr(sublime.Region(pt, view.line(pt).end()))
```
rchl, line 109: unnecessary annotation

rchl (commented):
Should probably go all the way and also annotate functions in `CSS/completions/common.py`, `CSS/completions/function_args.py` and `CSS/completions/properties.py` (with ` -> typing.Dict[str, typing.List[typing.Union[List[str], str]]]` I guess).

CSS/css_completions.py
```diff
@@ -164,7 +182,13 @@ def complete_property_value(self, view, prefix, pt):
 
         return completions
 
-    def complete_function_argument(self, view: sublime.View, prefix, pt):
+    def complete_function_argument(
+        self,
+        view: sublime.View,
+        prefix: str,
+        pt: sublime_types.Point
+    ) -> typing.Optional[typing.List[sublime.CompletionItem]]:
```
rchl, line 190: I think you should switch to importing specific types from `typing` in the import statement because:
a) it makes code shorter, cleaner and easier to read
b) that seems to be the convention used in ST types.

And then there would also be even less reason to break the lines for each argument which takes a lot of vertical space.

CSS/css_completions.py
```diff
@@ -1,11 +1,16 @@
...
+
 import sublime
 import sublime_plugin
-import timeit
 
+from . import completions
 from functools import cached_property, wraps
+import re
+import timeit
+from typing import List, Optional, TYPE_CHECKING, Union
```
deathaxe, line 10: Global vs. local imports ?!

PR #3527, opened 2022-11-07
[Diff] add typing info

jrappen:


Diff/diff.py
```diff
@@ -45,120 +56,176 @@ def run(self, files):
...
+    def run(self, edit: sublime.Edit):
+
+        v: sublime.View = self.view
+        w: typing.Union[None, sublime.Window] = v.window()
+        if not w:
+            sublime.status_message('Unable to diff with missing window')
+            return
 
-        fname = self.view.file_name()
+        fname: typing.Union[None, str] = v.file_name()
```
rchl, line 99: `typing.Optional` is more appropriate in those cases. But also there is no point in annotating types in those cases since the type will be evaluated from the value already.

Diff/diff.py
```diff
@@ -1,14 +1,19 @@
...
+
 import difflib
-import os.path
+import os
 import time
 
 import sublime
 import sublime_plugin
 
+import typing
```
rchl, line 10: Are you following any logic in regards to grouping imports? Feels like `typing` should be grouped together with other standard libraries.

Diff/diff.py
```diff
@@ -1,14 +1,19 @@
...
+
 
-def splitlines_keep_ends(text):
-    lines = text.split('\n')
+def splitlines_keep_ends(
+    text: str
+) -> typing.List[str]:
```
rchl, line 14: Why adding line breaks?

IMO there should be either explicit line limit defined that would dictate whether line should be broken or, if there is no limit, then just don't break lines arbitrarily.

There is benefit to having as much code as possible to fit within the editor view.

Diff/diff.py
```diff
@@ -45,120 +52,176 @@ def run(self, files):
...
         return len(files) == 2
 
 
 class DiffChangesCommand(sublime_plugin.TextCommand):
 
-    def run(self, edit):
+    def run(self, edit: sublime.Edit):
+
+        v: sublime.View = self.view
+        w: typing.Optional[sublime.Window] = v.window()
```
rchl, line 89: I don't know conventions in this repo but single letter (or abbreviated) variables are often discouraged.

Diff/diff.py
```diff
@@ -170,26 +233,41 @@ def run(self, edit, **kwargs):
...
-        selected_views = list(get_selected_views(self.view.window()))
-        if len(selected_views) == 2:
-            return 'Diff Selected Tabs...'
+        if views is None:
+            return False
+        else:
+            return len(views) == 2
+
+    def description(self, **kwargs) -> str:
+        w: typing.Optional[sublime.Window] = self.view.window()
```
rchl, line 266: This is again a convention thing, but it shouldn't be necessary to annotate in cases where the type is easily inferred from the value.

Diff/diff.py
```diff
@@ -1,14 +1,16 @@
-import codecs
-import difflib
-import os.path
-import time
+from __future__ import annotations
 
 import sublime
 import sublime_plugin
 
+import difflib
```
deathaxe, line 6: Why sorting global python imports after local/private sublime_... imports?

Diff/diff.py
```diff
@@ -45,120 +47,162 @@ def run(self, files):
...
-        return [view for view in selected_views if view.id() != tab_context_view.id()] + [tab_context_view]
+def get_views_from_tab_context(
+    active_view: sublime.View,
+    **kwargs
+) -> typing.Union[None, typing.List[sublime.View]]:
+    window: typing.Optional[sublime.Window] = active_view.window()
+    if window is None:
+        return None
+    selected_views: typing.List[sublime.View] = get_selected_views(window)
+    if window is not None:
```
rchl, line 168: if we are returning on `window is None` then we don't need to check it a second time below.

Diff/diff.py
```diff
@@ -170,26 +214,41 @@ def run(self, edit, **kwargs):
...
-            return 'Diff Selected Tabs...'
+        if views is None:
+            return False
+        else:
+            return len(views) == 2
+
+    def description(self, **kwargs) -> str:
+        window: typing.Optional[sublime.Window] = self.view.window()
+        if window is not None:
+            selected_views = list(get_selected_views(window))
```
rchl, line 250: Maybe just return empty string on `window is None` to make the logic read better.

Diff/diff.py
```diff
@@ -45,120 +47,162 @@ def run(self, files):
...
-    selected_views = list(get_selected_views(active_view.window()))
-    if 'group' in kwargs and 'index' in kwargs:
-        tab_context_view = get_view_from_tab_context(active_view, **kwargs)
-        # if the tab which was right clicked on is selected, exclude it from the selected views and re-add it afterwards
-        # so that the order of the diff will be determined by which tab was right-clicked on
-        return [view for view in selected_views if view.id() != tab_context_view.id()] + [tab_context_view]
+def get_views_from_tab_context(
+    active_view: sublime.View,
+    **kwargs
+) -> typing.Union[None, typing.List[sublime.View]]:
```
rchl, line 163: It would simplify code to always return a list but empty, if needed. Unless there is specific need to differentiate those cases.

Diff/diff.py
```diff
@@ -1,14 +1,16 @@
...
+import typing
 
 import sublime
 import sublime_plugin
 
 
-def splitlines_keep_ends(text):
-    lines = text.split('\n')
+def splitlines_keep_ends(text: str) -> typing.List[str]:
+    lines: typing.List[str] = text.split('\n')
```
rchl, line 13: also not necessary

Diff/diff.py
```diff
@@ -63,92 +65,118 @@ def is_visible(self, files):
 
 class DiffChangesCommand(sublime_plugin.TextCommand):
 
-    def run(self, edit):
+    def run(self, edit: sublime.Edit):
+
+        view: sublime.View = self.view
+        window: typing.Optional[sublime.Window] = view.window()
```
rchl, line 71: unnecessary annotations

also in many places below. I won't be repeating myself

PR #3528, opened 2022-11-07
[TeX] improved support for \def and friends

ngc92:
Improvements:
* Added better scoping for argument specifications of `\def`
* Allowed `\def` also for backslash+nonletter type commands
* Accept `\def`, `\edef`, `\gdef`, and `\xdef`
* Started a `syntax-test-tex.tex` file

I've tried to implement this like the LaTeX `\newcommand` #3508, i.e. with multiple pushes, but could not get the scopes to work, so right now this is a chain of `set`s.

The `\edef` and `\xdef` variations are slightly different in terms of TeX, as they do not accept arguments, but based on the not-a-linter argument I've decided to go with the simpler implementation that treats all these as the same. For correct code, this does not make a difference.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -62,6 +62,8 @@ variables:
   # type names, and if they were highlighted wrongly the user could not fix this.
   letter: '[A-Za-z@]'
   nonletter: '[^A-Za-z@]'
+  # look-ahead for end of command sequence
+  endcs: '(?!{{letter}})'
 
 contexts:
   prototype:
@@ -114,11 +116,11 @@ contexts:
       scope: comment.line.percentage.tex
 
   controls:
-    - match: ((\\)(else|expandafter|fi|if(case|cat|dim|eof|false|hbox|hmode|inner|mmode|num|odd|true|undefined|vbox|vmode|void|x)?))\b
+    - match: ((\\)(else|expandafter|fi|if(case|cat|dim|eof|false|hbox|hmode|inner|mmode|num|odd|true|undefined|vbox|vmode|void|x)?)){{endcs}}
       captures:
         1: keyword.control.tex
         2: punctuation.definition.backslash.tex
-    - match: ((\\)(?:input))\b
+    - match: ((\\)(?:input)){{endcs}}
       scope: meta.function.input.tex
       captures:
         1: keyword.control.input.tex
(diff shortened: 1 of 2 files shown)


michaelblyons:
Is LaTeX more-or-less a superset of TeX? If so, should later work make LaTeX an extension of TeX? (Pardon the tangentially-related meta-note. 😉)

deathaxe:
Looks like that.

LaTeX/TeX.sublime-syntax
```diff
@@ -68,30 +78,43 @@ contexts:
...
+      captures:
+        1: punctuation.definition.backslash.tex
 
   comments:
     - match: '%.*$\n?'
       scope: comment.line.percentage.tex
 
   controls:
-    - match: ((\\)(else|expandafter|fi|if(case|cat|dim|eof|false|hbox|hmode|inner|mmode|num|odd|true|undefined|vbox|vmode|void|x)?))\b
+    - match: ((\\)(else|fi|ftrue|if(case|cat|dim|eof|false|hbox|hmode|inner|mmode|num|odd|undefined|vbox|vmode|void|x)?))\b
```
michaelblyons, line 117: Mismerge?

Also, you could extend the scope in capture 1 to `keyword.control.conditional.tex` or split the regex so that `\else` (`keyword.control.conditional.else.tex`) and `\fi` (`keyword.control.conditional.end.tex`) are alone and the others get `keyword.control.conditional.if.tex`.

michaelblyons (changes requested):

michaelblyons (approved):

LaTeX/TeX.sublime-syntax
michaelblyons: Sure. 🙂

LaTeX/TeX.sublime-syntax
```diff
@@ -148,21 +171,67 @@ contexts:
         - include: main
 
   macros:
-    - match: ((\\)def)\s*((\\)[A-Za-z@]+)\s*[^\{]*?\s*(\{)
+    # Note \edef and \xdef have slightly different syntax, in that they do not
+    # allow for arguments. However, every valid \edef could be a valid \def,
+    # so syntax highlighting will work properly with correct code.
+    # we need to make sure to match only full commands, so we look-ahead for
+    # any following non-letter
+    - match: (\\)[gex]?def(?={{nonletter}})
```
deathaxe, line 179: I'd suggest a variable like `{{break}}` which is defined as `(?!{{letter}})` to also use the strategy of negative lookahead as in LaTeX.

ngc92: this also gives us the option of ensuring that the other commands/keywords actually terminate correctly, instead of \b which misses that an `@` can continue a command name.

deathaxe (approved):

michaelblyons (approved):

PR #3530, opened 2022-11-07
[ShellScript] add typing info

jrappen:


rchl:
Can you explain the need for this? Technically you've only added annotations to variables that already would have proper types inferred correctly assuming that the `sublime` API is typed.

rchl:
I see that the `import Default.exec` would mess things up unless its location is made known to the type checker but that's the case with or without this change

rchl:
> You don't get the correct typing info without a cross-reference, though?

But that's because the type of `import Default.exec` is not known so i feel that this is fixing types at the wrong end.

Should maybe instead try to ensure that the type checker knows that it's a TextCommand and then it would not be necessary to manually add redundant types

rchl:
> how do I do that?

Ideally by adding stubs for the Default "module" but I'm not sure if that would be acceptable option given that ST devs probably have actual Default directory in their checkout.

PR #3532, opened 2022-11-09
[Python] Refactor string contexts to support inheritance

### LINKED ISSUE #3462: [Python] Enable meta_prepend / meta_append usage.
### What happened?

Currently my HTML highlighter plugin overrides the SQL highlighter :cry:  which is unfortunate. Discord says we need support for `meta_prepend` and `meta_append` for Python.

https://discord.com/channels/280102180189634562/280102180189634562/1005140944565321848

Ideally I would love to see HTML in Python added to core, but enabling `meta_prepend` and `meta_append` would be fine in the interim.

https://github.com/gnat/sublime-python-html

183111942-9d6e8dc7-6b82-47fb-9492-07d25ffc6523

deathaxe:
Fixes #3462

This PR contains structural changes to enable 3rd-party syntaxes to support more syntaxes within strings.

_Note: Python is still sublime-syntax V1, which may change in a future commit and will therefore require adjustments in each 3rd-party syntax which is going to extend it._

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -1940,29 +1940,33 @@ contexts:
         )__\b
       scope: support.variable.magic.python
 
+###[ STRINGS ]################################################################
+
   docstrings:
-    - match: ^\s*(?=(?i)(ur|ru|u|r)?("""|'''))
-      push:
-      - match: (?i)(u)?("""|''')
-        captures:
-          1: storage.type.string.python
-          2: punctuation.definition.comment.begin.python
-        set:
-          - meta_scope: comment.block.documentation.python
-          - include: escaped-unicode-char
-          - include: escaped-char
-          - match: '\2'
-            scope: punctuation.definition.comment.end.python
-            pop: true
-      - match: (?i)(u?ru?)("""|''')
-        captures:
-          1: storage.type.string.python
-          2: punctuation.definition.comment.begin.python
-        set:
-          - meta_scope: comment.block.documentation.python
-          - match: '\2'
-            scope: punctuation.definition.comment.end.python
-            pop: true
+    - match: ^\s*(?i)(u)?("""|''')
+      captures:
+        1: storage.type.string.python
+        2: punctuation.definition.comment.begin.python
+      push: inside-docstring
+    - match: ^\s*(?i)(ur|ru?)("""|''')
+      captures:
+        1: storage.type.string.python
+        2: punctuation.definition.comment.begin.python
+      push: inside-raw-docstring
+
+  inside-docstring:
+    - meta_scope: comment.block.documentation.python
+    - match: \2
+      scope: punctuation.definition.comment.end.python
+      pop: true
+    - include: escaped-unicode-char
+    - include: escaped-char
+
+  inside-raw-docstring:
+    - meta_scope: comment.block.documentation.python
+    - match: \2
+      scope: punctuation.definition.comment.end.python
+      pop: true
 
   escaped-char:
     - match: '(\\x\h{2})|(\\[0-7]{1,3})|(\\[\\"''abfnrtv])'
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -2226,12 +2226,14 @@ contexts:
     - match: '"'
       scope: punctuation.definition.string.end.python
       set: after-expression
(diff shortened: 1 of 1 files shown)


Python/Python.sublime-syntax
```diff
@@ -1940,897 +1940,1152 @@ contexts:
...
-  line-continuation-inside-string:
-    - match: (\\)$\n?
-      captures:
-        1: punctuation.separator.continuation.line.python
-    - match: \n
-      scope: invalid.illegal.unclosed-string.python
-      set: after-expression
+  inside-docstring:
+    - meta_scope: comment.block.documentation.python
+    - match: \2
```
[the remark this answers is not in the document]
deathaxe: This hasn't changed compared to original code. Changing that would mean to duplicate all those contexts.

deathaxe:
Which additions?

If I don't miss anything the only addition is adding `meta_include_prototype: false`. Everything else should just be structural changes. So if something isn't covered by test cases it wouldn't have been before. Didn't go through any test cases in detail, tbh. Just made sure not to break anything.

deathaxe:
I can't say something about the test coverage of Python. If there are certain things missing, we could of course add them.

gnat:
>Note: Python is still sublime-syntax V1, which may change in a future commit and will therefore require adjustments in each 3rd-party syntax which is going to extend it.

As a primary Python dev myself, most 3rd party syntax extensions for Python are frankly terrible.

The built in syntaxer, + SQL in strings, + HTML in strings, would be better than what we currently have.

deathaxe:
The note more or less is a warning for those who will extend Python based on this PR to add HTML highlighting within strings. A future change of Python to sublime-syntax v2 may requires some adjustmants of any extending syntax as well.

michaelblyons:
I forgot to test drive this, but I have it enabled now.

michaelblyons (dismissed):

Python/Python.sublime-syntax
```diff
@@ -1948,895 +1948,1150 @@ contexts:
...
+    - include: escaped-unicode-chars
+    - include: escaped-chars
+    - include: string-placeholders
+    - include: string-replacements
+
+###[ STRING CONTENTS ]########################################################
+
+  line-continuation-inside-block-string:
+    - match: \\$
+      scope: punctuation.separator.continuation.line.python
```
FichteFoll, line 2881:
```suggestion
  line-continuations-inside-block-string:
    - match: \\$
      scope: punctuation.separator.continuation.line.python
```

Shouldn't this be plural since it doesn't pop? The following non-block variant is slightly more difficult because it *can* pop in some situations. Perhaps something related to `line-continuation-or-pop`?

FichteFoll (approved):
Just one minor thing, otherwise LGTM.

I was almost gonna review the entire refactor branch until I noticed I was looking at the wrong branch in Merge because reviewing Packages PRs with proper syntax highlighting is a much better experience

Python/Python.sublime-syntax
deathaxe: Current context names are just copied from previous code, but your suggestion is very correct with the popping vs. non-popping nameing convention.

deathaxe: How about naming them `string-continuations` vs. `string-continuation-or-illegal-end`?

FichteFoll: I like `string-continuations`, but shouldn't it be `string-continuations-or-illegal-end` with an s for the otherone?

michaelblyons (approved):

FichteFoll (approved):

Python/Python.sublime-syntax
deathaxe: I guess we could debate on where in a context name plural is to be used and where to use singular endlessly. Already renamed such contexts often, internally because of not being satisfied with my own choices every here and then.

Always tried to keep plural at the end of a name and just use singular if any pattern is popping. Don't say that's a perfect choice so. I'd be fine with your suggestion as well. If we don't care about the capture groups, we could probably also split them to `string-continuations` and `illegal-string-end` completely.

deathaxe: TBH, like the idea of splitting it up, best for re-usability reasons of patterns. Pushed a commit with according changes.

michaelblyons (approved):

FichteFoll (approved):

PR #3533, opened 2022-11-09
[TeX] extended control flow matches and scopes

ngc92:
As requested by @michaelblyons , some more fine-grained scoping for control structures in TeX.

Q: How important is it that these scopes are balanced? In the examples in the test, 
the `keyword.control.conditional.if.tex` never has any corresponding `end` scope, because the loop-ending `repeat` closes both the `loop`, and the `if` as part of the loop condition.

 And the conditions `\if@director` would be using a custom-defined if condition. So we have `else` and `fi` without a corresponding start of that scope. I'm not sure if this can be fixed. Maybe we can assume that all commands that start with the sequence `\if` and are not followed by brace-enclosed arguments probably constitute the beginning of an `if`, but I'm not sure how many false positives (or false negatives) this would bring.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -116,43 +116,42 @@ contexts:
       scope: comment.line.percentage.tex
 
   controls:
-    - match: ((\\)ifcase){{endcs}}
+    - match: (\\)ifcase{{endcs}}
+      scope: keyword.control.conditional.switch.tex
       captures:
-        1: keyword.control.conditional.switch.tex
-        2: punctuation.definition.backslash.tex
-    - match: ((\\)or){{endcs}}
+        1: punctuation.definition.backslash.tex
+    - match: (\\)or{{endcs}}
+      scope: keyword.control.conditional.case.tex
       captures:
-        1: keyword.control.conditional.case.tex
-        2: punctuation.definition.backslash.tex
-    - match: ((\\)loop){{endcs}}
+        1: punctuation.definition.backslash.tex
+    - match: (\\)loop{{endcs}}
+      scope: keyword.control.loop.do-while.tex
       captures:
-        1: keyword.control.loop.do-while.tex
-        2: punctuation.definition.backslash.tex
-    - match: ((\\)repeat){{endcs}}
+        1: punctuation.definition.backslash.tex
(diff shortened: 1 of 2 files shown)


LaTeX/TeX.sublime-syntax
```diff
@@ -116,15 +116,42 @@ contexts:
...
+    - match: (\\)ifcase{{endcs}}
+      scope: keyword.control.conditional.switch.tex
       captures:
-        1: keyword.control.tex
-        2: punctuation.definition.backslash.tex
-    - match: ((\\)(?:input)){{endcs}}
-      scope: meta.function.input.tex
+        1: punctuation.definition.backslash.tex
+    - match: (\\)or{{endcs}}
+      scope: keyword.control.conditional.case.tex
```
deathaxe, line 124: The `\or` looks more like an operator to me. Is there something like `\and` or `\not` as well?

```suggestion
      scope: keyword.operator.logical.tex
```

ngc92: I don't think it should be a logical operator. Then you'd expect \or to combine two boolean-like entities, right? I guess some languages (lua) idiomatically also use `or`-like constructs to get non-bool results, so that is not a strict criterion.

The following code from the tests 
```TeX
\ifcase 2 a\or b\or c\or d\else e\fi
```
would correspond more-or-less (I'd have to look up if TeX uses 0 or 1 to start counting) to the following `C` code:
```c
switch(2) {
 case 0: a; break;
 case 1: b; break;
 case 2: c; break;
 case 3: d; break;
 default: e;
}
```

deathaxe: The scope is ok then.

deathaxe (approved):

michaelblyons (dismissed):
Unneeded no-capture group on `input`, and I have no basis for judging the scope on`expandafter` (but it sounds like it makes sense).

Thanks for doing this. 👍

ngc92:
> Unneeded no-capture group on `input`, and I have no basis for judging the scope on`expandafter` (but it sounds like it makes sense).

I've removed the input capture group. 
Btw, is there a way to avoid all the duplication for scoping the leading backslash in the commands?

michaelblyons (approved):

deathaxe (approved):

PR #3542, opened 2022-11-12
[CSS] Fix missing commas in property completions

deathaxe:
This PR is an excerpt from #3526.

The goal is to have a PR per topic. Typing and typos are unrelated.

### DIFF
--- a/CSS/completions/properties.py
+++ b/CSS/completions/properties.py
@@ -369,7 +369,8 @@ def get_properties():
         ],
         "flex": [
             "<calc>",
-            "none" "auto",
+            "none",
+            "auto",
         ],
         "flex-basis": [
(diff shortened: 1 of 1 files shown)


michaelblyons (approved):

jfcherng (approved):

deathaxe:
To track changes we should keep PRs focused on certain topics, if they are not "rewrites". It was just too easy to create the little PR instad of writing tons of request messages. The other fact is - yes - I wanted the typos to be merged asap so they may go into next fixup stable release - if there's one. I would have left the decision about typing to other maintainers, as I in general don't see much value in it for such little plugins. I am still too shocked of the massive typing PR LSP recently received.

PR #3545, opened 2022-11-12
[TeX] Initial support for registers

ngc92:
This PR adds highlighting for registers, as well as for register allocation functions. Registers created with `\new...` functions will be visible in the symbol list.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -145,26 +145,26 @@ contexts:
       push: def-registername
 
   def-registername:
+    - match: \{
+      scope: punctuation.definition.group.brace.begin.tex
+      push: inside-register-name-group
     - match: (\\){{cmdname}}
       scope: entity.name.constant.tex
       captures:
         1: punctuation.definition.backslash.tex
       pop: 1
-
-    - match: \{
-      scope: punctuation.definition.group.brace.begin.tex
-      push:
-        - meta_scope: meta.group.brace.latex
-        - match: \}
-          scope: punctuation.definition.group.brace.end.tex
-          pop: 2
-        - match: (\\){{cmdname}}
-          scope: entity.name.constant.tex
-          captures:
-            1: punctuation.definition.backslash.tex
-
     - include: paragraph-pop
     - include: else-pop
+
+  inside-register-name-group:
+    - meta_scope: meta.group.brace.latex
+    - match: \}
+      scope: punctuation.definition.group.brace.end.tex
+      pop: 2
+    - match: (\\){{cmdname}}
+      scope: entity.name.constant.tex
+      captures:
+        1: punctuation.definition.backslash.tex
 
   comments:
     - match: '%.*$\n?'
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -123,13 +123,12 @@ contexts:
       captures:
         1: punctuation.definition.backslash.tex
         2: storage.type.tex
-      push:
-        - meta_scope: variable.language.tex
-        - include: register-id
+      push: register-id
 
   register-id:
+    - meta_scope: meta.register.tex
     - match: \d+
-      scope: constant.numeric.tex
+      scope: meta.number.integer.decimal constant.numeric.value.tex
       pop: 1
     - match: (\\){{cmdname}}
       captures:
@@ -146,15 +145,24 @@ contexts:
       push: def-registername
 
   def-registername:
-    - match: (?=\\)
-      set:
+    - match: (\\){{cmdname}}
+      scope: entity.name.constant.tex
+      captures:
+        1: punctuation.definition.backslash.tex
+      pop: 1
+
+    - match: \{
+      scope: punctuation.definition.group.brace.begin.tex
+      push:
+        - meta_scope: meta.group.brace.latex
+        - match: \}
+          scope: punctuation.definition.group.brace.end.tex
+          pop: 2
         - match: (\\){{cmdname}}
           scope: entity.name.constant.tex
           captures:
             1: punctuation.definition.backslash.tex
-          pop: 1
-        - include: paragraph-pop
-        - include: else-pop
+
     - include: paragraph-pop
     - include: else-pop
 
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -67,6 +67,8 @@ variables:
   nonletter: '[^A-Za-z@]'
   # look-ahead for end of command sequence
   endcs: '(?!{{letter}})'
+  # a command name: either a sequence of letters, or a single non-letter
+  cmdname: '(?x: {{letter}}+ | {{nonletter}})'
 
 contexts:
   prototype:
@@ -81,6 +83,7 @@ contexts:
     - include: block-math
     - include: inline-math
     - include: registers
+    - include: register-allocations
     - include: general-constants
     - include: general-commands
 
@@ -116,12 +119,44 @@ contexts:
         1: punctuation.definition.backslash.tex
 
   registers:
-    - match: (\\)({{registers}})(\d*)
-      scope: variable.language.tex
+    - match: (\\)({{registers}}){{endcs}}
       captures:
         1: punctuation.definition.backslash.tex
         2: storage.type.tex
-        3: constant.numeric.tex
+      push:
+        - meta_scope: variable.language.tex
+        - include: register-id
+
+  register-id:
+    - match: \d+
+      scope: constant.numeric.tex
+      pop: 1
+    - match: (\\){{cmdname}}
+      captures:
+        1: punctuation.definition.backslash.tex
+      pop: 1
+    - include: paragraph-pop
+    - include: else-pop
+
+  register-allocations:
+    - match: (\\)new{{registers}}{{endcs}}
+      scope: keyword.declaration.register.tex storage.modifier.register.tex
+      captures:
+        1: punctuation.definition.backslash.tex
+      push: def-registername
+
+  def-registername:
+    - match: (?=\\)
+      set:
+        - match: (\\){{cmdname}}
+          scope: entity.name.constant.tex
+          captures:
+            1: punctuation.definition.backslash.tex
+          pop: 1
+        - include: paragraph-pop
+        - include: else-pop
+    - include: paragraph-pop
+    - include: else-pop
 
   comments:
     - match: '%.*$\n?'
@@ -229,12 +264,7 @@ contexts:
       set:
         - clear_scopes: 1
         - meta_scope: meta.function.identifier.tex
-        - match: (\\){{letter}}+
-          scope: entity.name.definition.tex
-          captures:
-            1: punctuation.definition.backslash.tex
-          set: def-argspec
-        - match: (\\){{nonletter}}
+        - match: (\\){{cmdname}}
           scope: entity.name.definition.tex
           captures:
             1: punctuation.definition.backslash.tex
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -130,10 +130,7 @@ contexts:
     - match: \d+
       scope: meta.number.integer.decimal constant.numeric.value.tex
       pop: 1
-    - match: (\\){{cmdname}}
-      captures:
-        1: punctuation.definition.backslash.tex
-      pop: 1
+    - include: general-commands
     - include: paragraph-pop
     - include: else-pop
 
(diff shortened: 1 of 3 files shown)


ngc92:
I think github takes its commit messages for the merge from the PR description, so I'm putting this here as a first comment.

As usual, I'm not really sure about scoping. For specifying a register as a literal, it makes sense to treat e.g. `\box128` as a single entity, with the `box` and `128` subscoped as specifying the type and a numeric constant. But if the number is the result of evaluating another macro, such as in this test `\toks \five = 8`, I'm not sure if it makes sense to scope the entire thing as a `variable`. 

This might me more of an academic question, though, because except for extremely low level code, such constructs should be rather rare, I think. 

Specifying a fixed register number is maybe comparable to hard-coding a pointer address in C, if you're doing some extremely close-to-hardware stuff, this might make sense, but generally you wouldn't do that. Similarly, I think most register interactions are the result of the `\newcount` (and similar) commands.

Finally, should I make specific scopes for the different kind of registers? For syntax highlighting, I'm not sure if it is really necessary, but e.g. if the Goto-symbol dialog would show the name and the type of the register, that could be useful.

deathaxe:
Maybe a meta scope would be more suitable than scoping the whole expression `variable`. Stacking `storage` type and `variable` onto each other is a bit uncommon. I guess something like `meta.register.tex` would be reasonable.

I think scoping all registers is enough.

LaTeX/TeX.sublime-syntax
```diff
@@ -111,6 +118,46 @@ contexts:
...
+  registers:
+    - match: (\\)({{registers}}){{endcs}}
+      captures:
+        1: punctuation.definition.backslash.tex
+        2: storage.type.tex
+      push:
+        - meta_scope: variable.language.tex
+        - include: register-id
+
+  register-id:
```
deathaxe, line 130: If `register-id` is not to be used anywhere else, you could directly push it onto stack to avoid creating anonymous contexts.

Also added suggested `meta.register.tex`. ... needs test cases to be adjusted.

```suggestion
      push: register-id

  register-id:
    - meta_scope: meta.register.tex
```

LaTeX/TeX.sublime-syntax
```diff
@@ -111,6 +118,46 @@ contexts:
...
+    - match: \d+
+      scope: constant.numeric.tex
```
deathaxe, line 132: I'd suggest to add `.value` to be specific enough in case a color scheme addresses all parts of numeric constants (`constant.numeric.[base|value|suffix]`).

```suggestion
      scope: constant.numeric.value.tex
```

LaTeX/TeX.sublime-syntax
```diff
@@ -111,6 +118,46 @@ contexts:
...
+    - match: (\\){{cmdname}}
+      captures:
+        1: punctuation.definition.backslash.tex
+      pop: 1
+    - include: paragraph-pop
+    - include: else-pop
+
+  register-allocations:
+    - match: (\\)new{{registers}}{{endcs}}
+      scope: keyword.declaration.register.tex storage.modifier.register.tex
```
deathaxe, line 143: We don't have a standard scope specified for `new` operators, but none of those is used anywhere else for it.

C# uses `keyword.operator.new`, other syntaxes use `keyword.other.new` or `keyword.other.storage.new`.

`keyword.declaration` is meant for common declaration keywords which are no real data types. Something like `def` in python or `function` in JS is such category of keyword.

`storage.modifier` is something like `public`, `private`, `readonly`, ...

LaTeX/TeX.sublime-syntax
```diff
@@ -111,6 +118,46 @@ contexts:
...
+  def-registername:
+    - match: (?=\\)
+      set:
+        - match: (\\){{cmdname}}
+          scope: entity.name.constant.tex
+          captures:
+            1: punctuation.definition.backslash.tex
+          pop: 1
+        - include: paragraph-pop
+        - include: else-pop
```
deathaxe, line 157: This context is of the same category as `register-id`. The lookahead `(?=\\)` is not required in this situation. The reason I used it in LaTeX's `newcommand-definition` context is the `{` which also needed to be handled.

```suggestion
    - match: (\\){{cmdname}}
      scope: entity.name.constant.tex
      captures:
        1: punctuation.definition.backslash.tex
      pop: 1
```

LaTeX/TeX.sublime-syntax
deathaxe: Maybe even add `meta.number.integer.decimal` to fully comply with the other syntaxes.

LaTeX/TeX.sublime-syntax
ngc92: Good observation about the braces! However, I've just checked, and it is also possible to put the braces there in this case. So I'll have to adapt this and add some tests :)

LaTeX/TeX.sublime-syntax
ngc92: This is modeled after the general macro def keyword, which we currently scope as `keyword.declaration.function.tex storage.modifier.definition.tex`. I think the storage modifier is "wrong", but before it was argued that it should stay there for compatibility with older color schemes?

ngc92:
> Maybe a meta scope would be more suitable than scoping the whole expression `variable`. Stacking `storage` type and `variable` onto each other is a bit uncommon. I guess something like `meta.register.tex` would be reasonable.
> 
> I think scoping all registers is enough.

I've changed the scoping to `meta.register.tex`, but I'm not sure if the end result is good. Now we have the first instance where a control sequence, i.e. some backslash-marked `\macro`, is not highlighted in any form with any of the default color schemes.

LaTeX/TeX.sublime-syntax
```diff
@@ -111,6 +118,54 @@ contexts:
...
+        2: storage.type.tex
+      push: register-id
+
+  register-id:
+    - meta_scope: meta.register.tex
+    - match: \d+
+      scope: meta.number.integer.decimal constant.numeric.value.tex
+      pop: 1
+    - match: (\\){{cmdname}}
+      captures:
```
deathaxe, line 134: Is this the unscoped `\macro` you are concerned about?

We could add a proper `scope` here, right?

ngc92: yes, I'm just not sure which scope. Maybe `variable.parameter.tex`?

LaTeX/TeX.sublime-syntax
```diff
@@ -111,6 +118,54 @@ contexts:
...
+          scope: punctuation.definition.group.brace.end.tex
+          pop: 2
+        - match: (\\){{cmdname}}
+          scope: entity.name.constant.tex
+          captures:
+            1: punctuation.definition.backslash.tex
+
+    - include: paragraph-pop
+    - include: else-pop
+
```
deathaxe, line 168: 1. I'd suggest to try to avoid anonymous contexts in new/rewritten code in order to improve support for inheritance. We should therefore always try to directly push a named context (see: `inside-register-name-group`). Syntaxes use different kinds of naming schemes one of them is to name contexts `inside-` if they describe a part inside of brackets. Another scheme is to just append `-body` (e.g.: `register-name-group-body`). We should decide for one and go with it.

2. Maybe a stupid habbit of no practical use, but maybe we should try to organice bracket contexts (and pushing ones) first, if possible. Hence I'd suggest to move `- match: \{` to the beginning. Just to have some kind of highlevel code structure strategy.

```suggestion
    - match: \{
      scope: punctuation.definition.group.brace.begin.tex
      push: inside-register-name-group
    - match: (\\){{cmdname}}
      scope: entity.name.constant.tex
      captures:
        1: punctuation.definition.backslash.tex
      pop: 1
    - include: paragraph-pop
    - include: else-pop

  inside-register-name-group:
    - meta_scope: meta.group.brace.latex
    - match: \}
      scope: punctuation.definition.group.brace.end.tex
      pop: 2
    - match: (\\){{cmdname}}
      scope: entity.name.constant.tex
      captures:
        1: punctuation.definition.backslash.tex

```

LaTeX/TeX.sublime-syntax
deathaxe: I see. We can also refine those scopes altogether in a future PR. I can recall concernes about backward compatibility when we introduced `keyword.declaration` to replace `storage.type` where appropriate. There's a so called forward-fill in ST's engine, which ensures both are highlighted the same in case no rule for `keyword.declaration` is present.

There's no such rule for `storage.modifier` however. Maybe you are right with it beeing more safe to keep it this way for now.

LaTeX/TeX.sublime-syntax
deathaxe: `variable.parameter` is normally used to scope argument names in function definitions.

I'd try to find a scope name which fits to `\def\five{0}`, so rules for goto definition vs. goto reference can be declared.

`\five` being scoped `meta.function entity.name.definition` seems a bit weird to me.

Maybe we should decide whether we want to scope macros `meta.function entity.name.function` or just `meta.macro entity.name.macro`

The latter one would probably be more generic in case macros are used for constants as well, which can't be distinguished by static syntax rules.

The proper counter part for 
- `entity.name.function` is `variable.function`. 
- `entity.name.macro` may be `constant.other.macro` (used in Erlang)

LaTeX/TeX.sublime-syntax
ngc92: I'm not sure if the inheritance argument makes much sense here, in that it seems very unlikely that there'd be a situation in which you'd change `inside-register-name-group` but not `def-registername` itself, but its probably best to be consistent with these things.

deathaxe: All syntaxes move towards the the goal to only use named contexts. Transition should be made as we move forwared with other things as well. Maybe create dedicated PRs for such restructurings as well, such as #3532.

We should not make any assumptions about possibilities or likelyness of contexts being inherit or not as a base syntax can't know that for sure.

For consistency reasons all contexts should be named at some point so all are available for inheritance if required.

deathaxe: Just as an example: Assume LaTeX would allow more complex expressions in braced register names.

It would need to rewrite the whole `def-registername` context to do with your code. Using named contexts for `inside-register-name-group` enables it to inject such patterns into braces without touching more than really needed.

ngc92: So would you also argue that there should not be more than one `match` pattern (except maybe else-pop) per context? Because one could have (almost equivalently) defined this context also as matching either
`{{cmdname}}` or `\{\s*{{cmdname}\s*\}`, in which case the discussion about anonymous contexts would not have come up, but the situation (a derived syntax might want to change the stuff in the braces) would be the same.

deathaxe: The smaller contexts are the more fine gained an inheriting syntax can override single aspects. I wouldn't however argue to reduce all contexts to a single pattern. We could do that, but at some point we run out of names 😆 - or they would get insane long, maybe.

Patterns like `\{\s*{{cmdname}\s*\}` are often a relic of old TextMate syntaxes. In case such expressions are well known to be used in single line statements only, we could probably keep them as is, but there are several compelling reasons not to do so:

1. if a syntax supports them to span multiple lines, I'd always push dedicated contexts. It would help not breaking highlighting in such scenarios - even if they appear temporarily while editing content.

2. pushing/popping contexts is cheaper than huge patterns with several capture groups, which increases overall performance of a syntax.

3. UX with regards to incomplete code is often better. Just imagine `{}` being typed without the context name. Braces wouldn't be highlighted with capture groups, but they are with pushing contexts.  
   Basically I'd compare braces to quotation marks in strings here. All strings push a context onto stack and so all those braces should do as well, IMHO.

4. readability, debugging may become easiear as named contexts are listed in the "show scope" popup. _(targets named contexts)_

5. supporting inheritance (or providing more flexibility to support it) is the most recent argument added to that list. _(targets named contexts)_

LaTeX/TeX.sublime-syntax
ngc92: we could also, at least for now, decide to not scope this directly, but go back to the main scope and let `general-commands` handle this. Given that I think such constructs will be extremely unlikely in actual user code, that would probably be good enough. Still, there remains the question of how to scope macros in general. At least for the language-provided ones, I think it would be nice to have separate scopes depending on whether they do take an argument or not, but you are right that we in general cannot determine this from the syntax alone.

deathaxe: Agree, would probably enough. If we would like to maintain `meta.register` `general-commands` could be included in `register-id` replacing the `- match: (\\){{cmdname}}` pattern.

I'd be fine with both.

ngc92: yes, let's do that for now.

LaTeX/TeX.sublime-syntax
deathaxe: Totally forgot about variable interpolation. If a syntax wants to use TeX/LaTeX as template, it might be useful to be able to inject variable interpolations.

deathaxe (approved):

michaelblyons (approved):

PR #3548, opened 2022-11-12
[LaTeX] replace "captures: 0: ..." in LaTeX Log

jrappen:


michaelblyons (approved):

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -38,13 +37,11 @@ contexts:
         - match: '[0-9]+\-\-[0-9]+'
           scope: variable.parameter.hyphenation.latex2
     - match: (<)
```
deathaxe, line 39: We could also remove the capture groups.
```suggestion
    - match: '<'
```

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -38,13 +37,11 @@ contexts:
...
-      captures:
-        0: punctuation.definition.string.begin.log.latex
+      scope: punctuation.definition.string.begin.log.latex
       push:
         - meta_scope: string.unquoted.other.filename.log.latex
         - match: (>)
```
deathaxe, line 43: Same here

deathaxe (approved):

LaTeX/LaTeX Log.sublime-syntax
deathaxe: There are more such obsolete capture groups. May be addressed in a separate PR.

Issue #3531, opened 2022-11-07
[LaTeX] LaTeX Syntax Development


ngc92:
I'm planning to do some more work on the (La)TeX syntax. I've opened this issue to make sure that general discussion does not get lost after a particular PR is merged.

First, some general info (disclaimer: I'm no expert; this may not be entirely accurate):
TeX is a macro programming language, and works by replacing tokens until only primitive things remain. This process is highly configurable from within the language itself, so that in principle the user can change the available syntax almost completely. Consequently, we cannot make a useful syntax highlighter for the actual language, but only for "idiomatic usage". 

Based on this, people have made predefined sets of macros (formats) for end-user "convenience" (necessary unless you're an expert, I would say). Popular ones are
* `LaTeX` document - Probably best considered an extension of (Plain)TeX. 
* `BiBTeX` - Looks quite different. Already supported by Sublime.
* `ConTeX` - An alternative to LaTeX, also an extension of plain TeX. Haven't used it myself. Currently no Sublime support.
* `Expl3` - The new, internal programming layer for LaTeX. May be embedded in regular LaTeX, though typically for package implementers. Also an extension of TeX, I think.
* `LaTeX` package - When implementing reusable packages, there are different structural commands, and it would probably make sense not to slow the syntax highlighting down with the huge amount of typesetting-related commands.
* `TikZ/PGF`- Maybe best considered an embedded DSL for specifying graphics within a (La)TeX document. 
* `dtx` - "Documented (La)TeX" - a rather weird format, in which most of the programming is actually written inside TeX comments, which makes the current syntax highlighting pretty useless there. Many LaTeX packages are distributed in this way.
* `ins` - Installer files for LaTeX packages. Also use TeX syntax, with a few extra commands available.

Then, of course, there are some highly popular packages, some of which are already supported by the current syntax.

I don't think I'm familiar enough with the details of sublime's syntaxes yet to start any big projects, so for the time being I'll try my hand at smaller improvements.

Things I'm planning to do for now:
* add `let`, `futurelet` to TeX (https://en.wikibooks.org/wiki/TeX/let)
* add `\NewDocumentCommand` and similar to LaTeX
* Extract common (parts of) match patterns into variables
* add more TeX primitives and LaTeX macros to be recognized. Without special syntax support, just mark them as "built-ins".

[...]

PR #3555, opened 2022-11-13
[TeX.DocStrip] Basic DocStrip support

ngc92:
This PR adds support for the `DocStrip` format of TeX, used by many LaTeX packages to specify their installation. 
https://www.texlive.info/CTAN/macros/latex/base/docstrip.pdf

I've inherited from the base `TeX` syntax, and just prepended the new commands to the `main` context. This is not quite accurate, in the sense that many `plainTeX` constructs would simply be invalid in a docstrip file -- as there is no typesetting to be done, typesetting and math commands don't really make sense. But I think it is OK to highlight these commands, and probably better to handle things like this, because if we manually select the subset that is supported, and later on add something new in `TeX` that also works in `DocStrip`, we might forget to change `DocStrips` `main` context.

In general, `DocStrip` files are usually rather short and simple, and typically look like the examples given in https://texdoc.org/serve/dtxtut.pdf/0 (appendix A).

Finally, since the goal is not to produce typeset output, but instead to describe how files are generated, I've used a `source` main scope.

### DIFF
--- a/LaTeX/DocStrip.sublime-syntax
+++ b/LaTeX/DocStrip.sublime-syntax
@@ -1,6 +1,6 @@
 %YAML 1.2
 ---
-# http://www.sublimetext.com/docs/3/syntax.html
+# https://www.sublimetext.com/docs/syntax.html
 name: TeX (DocStrip)
 file_extensions:
   - ins
--- a/LaTeX/DocStrip.sublime-syntax
+++ b/LaTeX/DocStrip.sublime-syntax
@@ -19,11 +19,11 @@ contexts:
 
   controls:
     - meta_append: true
-    - match: (\\)(?:endbatchfile){{endcs}}
+    - match: (\\)endbatchfile{{endcs}}
       scope: keyword.control.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
-    - match: (\\)(?:(?:batch)?input){{endcs}}
+    - match: (\\)(?:batch)?input{{endcs}}
       scope: meta.function.input.tex keyword.control.input.tex
       captures:
         1: punctuation.definition.backslash.tex
@@ -33,7 +33,7 @@ contexts:
         1: punctuation.definition.backslash.tex
 
   docstrip-preamble:
-    - match: (\\)((?:declare)?(?:pre|post)amble){{endcs}}
+    - match: (\\)(?:declare)?(?:pre|post)amble{{endcs}}
       scope: keyword.context.block.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
@@ -43,19 +43,19 @@ contexts:
   preamble-content:
     - meta_include_prototype: false
     - meta_content_scope: markup.raw.verbatim.tex
-    - match: (\\)(?:end(?:pre|post)amble){{endcs}}
+    - match: (\\)end(?:pre|post)amble{{endcs}}
       scope: keyword.context.block.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
       pop: 1
 
   docstrip-keywords:
-    - match: (\\)(:?file|from|generateFile){{endcs}}
+    - match: (\\)(?:file|from|generateFile){{endcs}}
       scope: keyword.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
       push: file-argument
-    - match: (\\)(:?generate|needed){{endcs}}
+    - match: (\\)(?:generate|needed){{endcs}}
       scope: keyword.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
@@ -73,13 +73,13 @@ contexts:
     - include: macro-brace-content
 
   docstrip-config:
-    - match: ((\\)(:?usedir|showdirectory|BaseDirectory|DeclareDir|UseTDS|maxfiles|maxoutfiles)){{endcs}}
+    - match: (\\)(?:usedir|showdirectory|BaseDirectory|DeclareDir|UseTDS|maxfiles|maxoutfiles){{endcs}}
       scope: support.function.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
 
   docstrip-user-io:
-     - match: (\\)(:?Msg|Ask){{endcs}}
+     - match: (\\)(?:Msg|Ask){{endcs}}
        scope: support.function.tex.docstrip
        captures:
          1: punctuation.definition.backslash.tex
@@ -90,7 +90,7 @@ contexts:
       captures:
         1: punctuation.definition.backslash.tex
 
-    - match: (\\)(?:(?:Double)perCent|MetaPrefix|empty){{endcs}}
+    - match: (\\)(?:(?:Double)?perCent|MetaPrefix|empty){{endcs}}
       scope: constant.language.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
--- a/LaTeX/DocStrip.sublime-syntax
+++ b/LaTeX/DocStrip.sublime-syntax
@@ -39,10 +39,9 @@ contexts:
       scope: keyword.context.block.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
-      push: preamble-content
+      push: docstrip-preamble-content
 
-
-  preamble-content:
+  docstrip-preamble-content:
     - meta_include_prototype: false
     - meta_content_scope: markup.raw.verbatim.tex
     - match: (\\)end(?:pre|post)amble{{endcs}}
@@ -56,21 +55,20 @@ contexts:
       scope: keyword.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
-      push: file-argument
+      push: docstrip-file-argument
     - match: (\\)(?:generate|needed){{endcs}}
       scope: keyword.tex.docstrip
       captures:
         1: punctuation.definition.backslash.tex
 
-
-  file-argument:
+  docstrip-file-argument:
     - match: \{
       scope: punctuation.definition.group.brace.begin.tex
-      set: file-argument-path
+      set: docstrip-file-argument-path
     - include: else-pop
     - include: paragraph-pop
 
-  file-argument-path:
+  docstrip-file-argument-path:
     - meta_content_scope: meta.path.tex.docstrip
     - include: macro-brace-content
 
--- a/LaTeX/DocStrip.sublime-syntax
+++ b/LaTeX/DocStrip.sublime-syntax
@@ -70,7 +70,7 @@ contexts:
 
   docstrip-file-argument-path:
     - meta_content_scope: meta.path.tex.docstrip
-    - include: macro-brace-content
+    - include: macro-braces-body
 
   docstrip-config:
     - match: (\\)(?:usedir|showdirectory|BaseDirectory|DeclareDir|UseTDS|maxfiles|maxoutfiles){{endcs}}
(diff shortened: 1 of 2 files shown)


LaTeX/DocStrip.sublime-syntax
```diff
@@ -0,0 +1,97 @@
...
+    - include: docstrip-user-io
+    - include: docstrip-keywords
+    - include: docstrip-constants
+
+  controls:
+    - meta_append: true
+    - match: ((\\)(?:endbatchfile)){{endcs}}
+      captures:
+        1: keyword.control.tex.docstrip
+        2: punctuation.definition.backslash.tex
```
deathaxe, line 25: Maybe cleanup some unused capture groups such as ...

```suggestion
    - match: (\\)endbatchfile{{endcs}}
      scope: keyword.control.tex.docstrip
      captures:
        1: punctuation.definition.backslash.tex
```

Applies to various other patterns in this syntax as well.

LaTeX/DocStrip.sublime-syntax
```diff
@@ -56,10 +55,10 @@ contexts:
       captures:
         1: punctuation.definition.backslash.tex
       push: file-argument
-    - match: ((\\)(:?generate|needed)){{endcs}}
+    - match: (\\)(:?generate|needed){{endcs}}
```
deathaxe, line 58: Typo?

```suggestion
    - match: (\\)(?:generate|needed){{endcs}}
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -75,23 +74,23 @@ contexts:
 
   docstrip-config:
     - match: ((\\)(:?usedir|showdirectory|BaseDirectory|DeclareDir|UseTDS|maxfiles|maxoutfiles)){{endcs}}
```
deathaxe, line 76:
```suggestion
    - match: (\\)(?:usedir|showdirectory|BaseDirectory|DeclareDir|UseTDS|maxfiles|maxoutfiles){{endcs}}
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -75,23 +74,23 @@ contexts:
...
+      scope: support.function.tex.docstrip
       captures:
-        1: support.function.tex.docstrip
-        2: punctuation.definition.backslash.tex
+        1: punctuation.definition.backslash.tex
 
   docstrip-user-io:
-     - match: ((\\)(:?Msg|Ask)){{endcs}}
+     - match: (\\)(:?Msg|Ask){{endcs}}
```
deathaxe, line 82:
```suggestion
     - match: (\\)(?:Msg|Ask){{endcs}}
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -19,35 +19,34 @@ contexts:
...
-        2: punctuation.definition.backslash.tex
+        1: punctuation.definition.backslash.tex
       push: preamble-content
 
 
   preamble-content:
     - meta_include_prototype: false
     - meta_content_scope: markup.raw.verbatim.tex
-    - match: ((\\)(?:end(?:pre|post)amble)){{endcs}}
+    - match: (\\)(?:end(?:pre|post)amble){{endcs}}
```
deathaxe, line 46: Maybe a question of taste, but the following would work as well

```suggestion
    - match: (\\)end(?:pre|post)amble{{endcs}}
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -19,35 +19,34 @@ contexts:
...
+    - match: (\\)ifToplevel{{endcs}}
+      scope: keyword.control.conditional.tex.docstrip
       captures:
-        1: keyword.control.conditional.tex.docstrip
-        2: punctuation.definition.backslash.tex
+        1: punctuation.definition.backslash.tex
 
   docstrip-preamble:
-    - match: ((\\)((?:declare)?(?:pre|post)amble)){{endcs}}
+    - match: (\\)((?:declare)?(?:pre|post)amble){{endcs}}
```
deathaxe, line 36: I'd suggest to either remove capture group

```suggestion
    - match: (\\)(?:declare)?(?:pre|post)amble{{endcs}}
```

or convert capture group in normal group

```suggestion
    - match: (\\)(?:(?:declare)?(?:pre|post)amble){{endcs}}
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -19,35 +19,34 @@ contexts:
 
   controls:
     - meta_append: true
-    - match: ((\\)(?:endbatchfile)){{endcs}}
+    - match: (\\)(?:endbatchfile){{endcs}}
```
deathaxe, line 22: The following would be enough

```suggestion
    - match: (\\)endbatchfile{{endcs}}
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -19,35 +19,34 @@ contexts:
...
+      scope: keyword.control.tex.docstrip
       captures:
-        1: keyword.control.tex.docstrip
-        2: punctuation.definition.backslash.tex
-    - match: ((\\)(?:(?:batch)?input)){{endcs}}
-      scope: meta.function.input.tex
+        1: punctuation.definition.backslash.tex
+    - match: (\\)(?:(?:batch)?input){{endcs}}
```
deathaxe, line 26: Same here

```suggestion
    - match: (\\)(?:batch)?input{{endcs}}
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -0,0 +1,96 @@
+%YAML 1.2
+---
+# https://www.sublimetext.com/docs/syntax.html
+name: TeX (DocStrip)
```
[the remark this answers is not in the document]
ngc92: I can do that. If this is the preferred order, though, I'd suggest to update https://www.sublimetext.com/docs/syntax.html

LaTeX/DocStrip.sublime-syntax
ngc92: seems like, and then copied all over the file :(

LaTeX/DocStrip.sublime-syntax
```diff
@@ -0,0 +1,98 @@
...
+
+  docstrip-preamble:
+    - match: (\\)(?:declare)?(?:pre|post)amble{{endcs}}
+      scope: keyword.context.block.tex.docstrip
+      captures:
+        1: punctuation.definition.backslash.tex
+      push: preamble-content
+
+
+  preamble-content:
```
deathaxe, line 45: Just noticed ...

1. Are the 2 empty lines of intent?
2. I'd suggest to prepend `docstrip-` to indicate it belongs to `docstring-preamble` context and being special docstrip context. Or is `preamble-content` something which may appear in other TeX dialects and may be moved to TeX at some point?

```suggestion
      push: docstrip-preamble-content

  docstrip-preamble-content:
```

LaTeX/DocStrip.sublime-syntax
```diff
@@ -0,0 +1,98 @@
...
+      captures:
+        1: punctuation.definition.backslash.tex
+      push: file-argument
+    - match: (\\)(?:generate|needed){{endcs}}
+      scope: keyword.tex.docstrip
+      captures:
+        1: punctuation.definition.backslash.tex
+
+
+  file-argument:
```
deathaxe, line 66: Same here.

LaTeX/DocStrip.sublime-syntax
ngc92: well, LaTeX has the concept of a "preamble", but it means something entirely different. So prefixing with docstrip definitely makes sense

deathaxe (approved):

jwortmann:
> I've used a `source` main scope

This means that color schemes which have defined special highlighting rules for TeX/LaTeX (which seems quite possible because LaTeX isn't really comparable to the other programming or markup languages and for historic reasons the syntax doesn't pay much attention to the scope naming guidelines), by using the `text.tex` base scope, would need to define the same rules again for the `source` scope in order to make them work with such files too. It isn't really a backward compatibility break, because the .ins file extension wasn't recognized by the LaTeX syntaxes before, but it seems that those files would be highlighted relatively well with the regular LaTeX syntax and I wonder whether changing the base scope would have do anything useful. And in general to change the base scope from `text` to `source` while extending a syntax seems quite unprecedented afaik.

deathaxe:
https://github.com/sublimehq/Packages/issues/3531 contains some discussion about `source` vs. `text` and why the one or the other may (or may not?) be more suitable for some of the file types - mainly due to purpose though.

deathaxe:
Maybe we should keep going with `text.tex. ...` main scope for compatibility reasons with existing color schemes.

ngc92:
Is this really an issue? First, the `ins` files would not even have been recognized by default before, and second, almost any command that appears in a typical `ins` file will have a different scope that if you were to highlight as `TeX`, as all the commands would be scoped as unspecific `support.function` with standard `TeX` syntax.

Conceptually, I guess the `.ins` file is somewhat comparable to a `Makefile`, and these are also scoped `source`.

deathaxe:
TBH, I am a bit uncertain about it. While being fine with scoping source code `source`, arguments about backward compatibility with existing maybe no longer maintained color schemes is to be considdered.

I don't have any numbers about possibly effected color schemes.

Default color schemes wouldn't be effected however.

We've actually haven't switched from `text.` to `source.` by inheritance in any other syntax yet. TeX may be special or original `text.` main scope was just not so well choosen with regards to the nature of TeX.

_The `.ins` extension hasn't been assigned to a syntax by intent, I guess. It's quite generic and may be used by various syntaxes._

So we should not add that at this point.

deathaxe:
Do we really want to bind `*.ins` file extension to that syntax?

Probably high risk of false positives, no?

LaTeX/DocStrip.sublime-syntax
```diff
@@ -0,0 +1,96 @@
...
+---
+# https://www.sublimetext.com/docs/syntax.html
+# https://www.texlive.info/CTAN/macros/latex/base/docstrip.pdf
+name: TeX (DocStrip)
+scope: source.tex.docstrip
+version: 1
+extends: Packages/LaTeX/TeX.sublime-syntax
+
+file_extensions:
+  - ins
```
michaelblyons, line 11:
```suggestion
file_extensions: []
```
https://fileinfo.com/extension/ins

michaelblyons (changes requested):
@deathaxe has a good point.

LaTeX/DocStrip.sublime-syntax
FichteFoll: I'm not really getting the feeling that any of the other usages of the "ins" extensions would be more likely than this one tbh.

> [9 file types](https://fileinfo.com/extension/ins) use the .ins file extension.
> 
> 1. [LaTeX Installer Script](https://fileinfo.com/extension/ins#latex_installer_script)
> 2. [Ensoniq Instrument](https://fileinfo.com/extension/ins#ensoniq_instrument)
> 3. [Internet Naming Service File](https://fileinfo.com/extension/ins#internet_naming_service_file)
> 4. [Adlib Tracker Instrument File](https://fileinfo.com/extension/ins#adlib_tracker_instrument_file)
> 5. [Internet Settings File](https://fileinfo.com/extension/ins#internet_settings_file)
> 6. [Sample Cell II Instrument Definition File](https://fileinfo.com/extension/ins#sample_cell_ii_instrument_definition_file)
> 7. [Inspiration File](https://fileinfo.com/extension/ins#inspiration_file)
> 8. [AGT Game Instruction File](https://fileinfo.com/extension/ins#agt_game_instruction_file)
> 9. [Internet Explorer Installation Settings](https://fileinfo.com/extension/ins#internet_explorer_installation_settings)

deathaxe: Ok, than leave it as is.

PR #3557, opened 2022-11-13
[TeX] refactor LaTeX Log syntax

jrappen:
This commit adds named contexts and a missing EOL pop for strings.

### DIFF
--- a/LaTeX/LaTeX Log.sublime-syntax
+++ b/LaTeX/LaTeX Log.sublime-syntax
@@ -78,7 +78,7 @@ contexts:
       scope: punctuation.definition.string.end.log.latex
       pop: 1
 
-    - match: (?:.*/.*\.pdf)
+    - match: '.*/.*\.pdf'
       scope: >-
         support.function.with-arg.latex
         entity.name.function.filename.latex
--- a/LaTeX/LaTeX Log.sublime-syntax
+++ b/LaTeX/LaTeX Log.sublime-syntax
@@ -25,7 +25,7 @@ contexts:
       push:
         - meta_scope: meta.log.latex.hyphenation
 
-        - match: '(?:\[\]\n)'
+        - match: \[\]\n
           scope: keyword.control.hyphenation.latex
           pop: 1
 
--- a/LaTeX/LaTeX Log.sublime-syntax
+++ b/LaTeX/LaTeX Log.sublime-syntax
@@ -5,21 +5,21 @@ scope: text.log.latex
 
 first_line_match: "This is (pdf|pdfe)?TeXk?, Version "
 
+################################################################################
+
 contexts:
-  eol-pop:
-    - match: '$\n?'
-      pop: 1
-
   main:
-    - include: invalids
-    - include: functions
-    - include: types
-    - include: tags
-    - include: comments
+    - include: log-entries
+    - include: file-paths
+    - include: doc-comments
+    - include: under-over-full
+    - include: strings
 
+  doc-comments:
     - match: ".*FiXme:"
       scope: meta.log.latex.fixme
 
+  under-over-full:
     - match: (?:Overfull|Underfull)
       scope: keyword.control.hyphenation.latex
       push:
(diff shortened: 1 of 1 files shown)


LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -63,14 +59,21 @@ contexts:
...
+
+  angle-bracket-string-content:
     - meta_scope: string.unquoted.other.filename.log.latex
 
     - match: '>'
       scope: punctuation.definition.string.end.log.latex
       pop: 1
 
-    - match: (.*/.*\.pdf)
+    - match: (?:.*/.*\.pdf)
```
deathaxe, line 81: I'd probably still prefer quotes over groups.

```suggestion
    - match: '.*/.*\.pdf'
```

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -6,44 +6,81 @@ scope: text.log.latex
...
+
+  angle-bracket-string-content:
+    - meta_scope: string.unquoted.other.filename.log.latex
+
+    - match: '>'
+      scope: punctuation.definition.string.end.log.latex
+      pop: 1
+
+    - match: '.*/.*\.pdf'
+      scope: >-
```
deathaxe, line 82: AFAIK it should also work without `>-`

```suggestion
      scope:
```

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -6,44 +6,81 @@ scope: text.log.latex
...
+  types:
     - match: .*\.cls
       scope: entity.name.type.class
+
+  tags:
     - match: .*\.cfg
       scope: entity.name.tag.configuration
+
     - match: .*\.def
       scope: entity.name.tag.definition
```
deathaxe, line 60: All of those just seem to address different kinds of file paths. Putting them into a single context would probably be enough.

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -6,44 +6,81 @@ scope: text.log.latex
...
+
+  invalids:
     - match: ".*Warning:"
       scope: invalid.deprecated
+
     - match: '[^:]*:\d*:.*'
       scope: invalid.deprecated
+
     - match: .*Error|^!.*
       scope: invalid.illegal
```
deathaxe, line 45: These are just log entry categories. `.*Info.*` from line 63 belongs to them as well as the `.*Fixme:` pattern does.

grafik

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -6,44 +6,81 @@ scope: text.log.latex
 first_line_match: "This is (pdf|pdfe)?TeXk?, Version "
 
 contexts:
+  eol-pop:
+    - match: '$\n?'
+      pop: 1
+
```
deathaxe, line 12: The `main` context should always be the first one. Most generic/common prototype like contexts should be located at the end of the document.

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -6,44 +6,81 @@ scope: text.log.latex
...
+      scope: keyword.control.hyphenation.latex
+      push:
+        - meta_scope: meta.log.latex.hyphenation
+
+        - match: \[\]\n
+          scope: keyword.control.hyphenation.latex
+          pop: 1
+
+        - match: '[0-9]+\-\-[0-9]+'
+          scope: variable.parameter.hyphenation.latex2
```
deathaxe, line 33: If `main` is turned into an organizational context, it should only contain `- include: ...` patterns.

deathaxe (changes requested):
As a general note: I don't think we need empty lines between each patterns. I'd use those only to group patterns in a very complex/big context. But before that, I'd split that up into several smaller ones.

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -5,21 +5,21 @@ scope: text.log.latex
...
-    - include: comments
+    - include: log-entries
+    - include: file-paths
+    - include: doc-comments
+    - include: under-over-full
+    - include: strings
 
+  doc-comments:
     - match: ".*FiXme:"
       scope: meta.log.latex.fixme
```
deathaxe, line 20: Despite the strange scopes being choosen by the original author, I'd say "FiXme" is another log-entry category - just like a warning. I may be wrong though.

LaTeX/LaTeX Log.sublime-syntax
```diff
@@ -5,45 +5,101 @@ scope: text.log.latex
...
+      scope: invalid.deprecated
+    - match: '[^:]*:\d*:.*'
+      scope: invalid.deprecated
+
+  info-log-entries:
+    - match: '.*Info.*'
       scope: comment.block.documentation
+
+  fixme-log-entries:
     - match: ".*FiXme:"
```
deathaxe, line 74: Let's add this single change for consistency with the other log patterns. Thre rest looks good to me.

```suggestion
    - match: '.*FiXme:'
```

deathaxe (approved):

PR #3559, opened 2022-11-14
[C#] Add folding rule for #region directive

jwortmann:
This adds the ability to fold `#region` preprocessor directives in C\#.

I've noticed one small occasion where it doesn't work correctly:
If there are nested regions with the same (meta-)scope, and the inner region is folded, then the folding range of the outer region erroneously stops at the end of the folded inner region too. However, if the inner region is not folded and you only fold the outer region, then everything works as expected.

For example, this works fine for any folding order (the inner regions are inside the meta-scope from the class):
```cs
#region Outer
public class MyClass {
    #region Inner 1
    // Hello
    #endregion

    #region Inner 2
    // World
    #endregion
}
#endregion
```

But the following example doesn't work correctly for the outer region if one of the inner regions was folded first:
```cs
#region Outer
// public class MyClass {
    #region Inner 1
    // Hello
    #endregion

    #region Inner 2
    // World
    #endregion
// }
#endregion
```

This issue could perhaps be mitigated if the `#region ... #endregion` block would push its own meta-scope, but I don't think that's even possible when these preprocessor lines are allowed to occur on any line, i.e. on any level of the current context stack.

But ultimately I think that it is a Sublime Text core issue, because the behavior whether the folding of the outer region works correctly, depends on the current folding state of the inner regions. I would expect that it must only be affected by the syntax scopes and nothing else.

---

By the way, some of the syntax tests on the master branch use a wrong comment marker and would not pass otherwise :D
https://github.com/sublimehq/Packages/blob/d04bd5e1c492ad0a2b1c552b42dc80f3e1828e9a/C%23/tests/syntax_test_PreprocessorDirectives.cs#L69-L70
I fixed it in an additional commit.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -178,20 +178,24 @@ contexts:
       pop: true
 
   preprocessor_region:
-    - match: '^\s*(#)\s*(region)\b[^\S\n]*(.*)($\n?)'
+    - match: '^\s*(#)\s*(region)\b\s*(\S.*)'
       scope: meta.preprocessor.region.cs
       captures:
         1: keyword.other.preprocessor.cs
         2: keyword.other.preprocessor.cs
-        3: entity.name.section.cs
-        4: meta.whitespace.newline.cs
+        3: entity.name.section.cs meta.fold.begin.cs
+    - match: '^\s*(#)\s*(region)\b\s*'
+      scope: meta.preprocessor.region.cs
+      captures:
+        1: keyword.other.preprocessor.cs
+        2: keyword.other.preprocessor.cs meta.fold.begin.cs
     - match: '^\s*(#)\s*(endregion)\b[^\S\n]*(.*)($\n?)'
-      scope: meta.preprocessor.endregion.cs
+      scope: meta.preprocessor.region.cs
       captures:
         1: keyword.other.preprocessor.cs
         2: keyword.other.preprocessor.cs
         3: variable.other.section.cs
-        4: meta.whitespace.newline.cs
+        4: meta.fold.end.cs
 
   # Pops out at the end of the line and handles comments.
   # Marks the rest of the line as invalid.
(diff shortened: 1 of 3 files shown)


jwortmann:
I found a workaround for the inconsistent folding behavior described above. It turned out that it only happens if the "begin" scope for the folding region represents exactly the linebreak character. Now I applied the folding scope to the relevant token preceding the linebreak (either `#region` keyword or the region label), and with this it seems to work well in all cases.

I think the unreliable behavior when using the linebreak character as folding begin scope is still a bug in Sublime Text though, which should be fixed eventually.

keith-hall (approved):

deathaxe:
I'd hesitate to add further folding rules especially if obvious edge cases are known.

jwortmann:
So is this on hold until the syntax based folding will become more stable in a future Sublime Text release?

During a bit of testing I didn't notice any more bugs for the #region folding, but of course I can't guarantee that no possible edge cases which I didn't consider might exist (but the same could be true for all of the existing folding rules in syntaxes).

And compared to the syntax based folding for parentheses and braces (which provides limited value for approximately 99.85% of all code bases - assuming that most people use indentation for such blocks anyway, and therefore it was already possible to fold them via indentation based folding), the new functionality to fold #region blocks from this PR seems to be much more useful to me and exactly one of the use case why syntax based folding was implemented.

deathaxe:
The value of this PR is for sure and I can follow all your arguments. Brackets are probably the most likely cause to break syntax based folding if meta scopes are not applied perfectly well.

Just saw your description about situations folding wouldn't work and had those frustrated feedback comments in mind after ST4142 release.

Adding features with known edge cases which may further increase possible folding inconsistencies may not be an option in general. I also dropped Markdown heading folding for exactly this reason.

I can't however reproduce the "not-working" edge case. Calclulating fold regions fails if multiple fold-begin/end tokens are connected (no gap inbetween) and share the same scope. ST can't distinguish them then. But this can't happen with your implementation. All regions can be folded/unfolded via fold buttons in any order. Fold by level not working is a core bug (https://github.com/sublimehq/sublime_text/issues/5706) if this was what you tried.

If so, I guess it is safe to merge after all.

deathaxe (approved):

PR #3565, opened 2022-11-16
Add support for zsh completions

ratijas:
All zsh completions found in /usr/share/zsh/ start with a magic word `#compdef` or `#autoload`, the exact format of which is documented in ZSHCOMPSYS(1) man page:

> The # is part of the tag name and no white space is allowed after it.
> The #compdef tags use the compdef function described below; the main
> difference is that the name of the function is supplied implicitly.


deathaxe:
Is anyone aware of possible ambiguities with C family pre-processor directives, which may cause extensionless sources to be assigned the wrong syntax due to this PR?

FichteFoll:
I would expect most C files to have an extension whereas these files (at least for me in the `site-functions` sub-directory) don't have an extension at all. That said, I'm not a C developer.

FichteFoll (approved):

deathaxe (approved):

deathaxe:
Standard library comes along with some extensionless files, but there are no conflicts found. So let's give it a try.

PR #3569, opened 2022-11-18
[C#] Drop "build" file extension assignment

### LINKED ISSUE #1862: File named "build" always opens as "NAnt Build FIle" type

deathaxe:
Fixes #1862

This commit unassignes `build` file extension from NAnt Build File syntax as it is a too generic extension used by many syntaxes.

With this change `*.build` files need to be assigned to NAnt Build File manually if desired.

### DIFF
--- a/C#/Build.sublime-syntax
+++ b/C#/Build.sublime-syntax
@@ -1,10 +1,9 @@
 %YAML 1.2
 ---
-# http://www.sublimetext.com/docs/3/syntax.html
+# http://www.sublimetext.com/docs/syntax.html
 name: NAnt Build File
-file_extensions:
-  - build
 scope: source.nant-build
+
 contexts:
   main:
     - match: <!--


michaelblyons (approved):

jfcherng (approved):

FichteFoll:
I assume there isn't a first_line_match rule we could add here instead to attempt to resolve ambiguities with this extension?

michaelblyons:
Looks like build files are XML, so the first line says "Hey I'm XML" rather than something helpful. See [sample from official doc](https://github.com/nant/nant/blob/master/doc/help/fundamentals/buildfiles.html#L31-L48), unescaped below.

```xml
<?xml version="1.0"?>
    <project name="Hello World" default="build" basedir=".">
        <description>The Hello World of build files.</description>
        <property name="debug" value="true" overwrite="false" />
        <target name="clean" description="remove all generated files">
            <delete file="HelloWorld.exe" failonerror="false" />
            <delete file="HelloWorld.pdb" failonerror="false" />
        </target>
        <target name="build" description="compiles the source code">
            <csc target="exe" output="HelloWorld.exe" debug="${debug}">
                <sources>
                    <includes name="HelloWorld.cs" />
                </sources>
            </csc>
        </target>
    </project>
```

FichteFoll:
Well, that's unfortuantely not enough to distinguish a "build" file from any other xml file.

deathaxe:
TBH, I don't know why a NAnt syntax exists at all. It is just a very naive and basic XML syntax which doesn't add any value compared to using XML.sublime-syntax.

michaelblyons:
> TBH, I don't know why a NAnt syntax exists at all. It is just a very naive and basic XML syntax which doesn't add any value compared to using XML.sublime-syntax.

Took a look. He's not wrong. IMO delete it.

If someone _really wants to,_ they can extend XML and highlight specific properties.

PR #3575, opened 2022-11-19
[LaTeX] Improve LaTeX

deathaxe:
This PR is introduced to track changes on develop/latex branch, which is intented to be used to collect and review various improvements implemented by @ngc92 without creating too much noise on `master` branch.

The idea is to base all work/refactoring on this branch and merge them into this one using PRs.

Once a milestone with reasonable changes is hit, we can merge it to master branch as "Rewrite LaTeX" or whatever name/category is suitable.

The starting point is #3546 but with the opportunity to use this repo as working area instead of my "private" one.

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -363,16 +363,14 @@ contexts:
         3: punctuation.definition.group.brace.begin.tex
       push:
         - meta_scope: meta.function.box.latex
-        - match: \}
-          scope: punctuation.definition.group.brace.end.tex
-          pop: 1
+        - include: brace-group-end
         - include: main
     - match: ((\\)(?:framebox|makebox))\b
       captures:
         1: support.function.box.latex
         2: punctuation.definition.backslash.latex
       push:
-        - [{meta_scope: meta.function.box.latex}, {match: '', pop: 1}]
+        - box-meta
         - argument
         - optional-arguments
     - match: ((\\)parbox)\b
@@ -380,7 +378,7 @@ contexts:
         1: support.function.box.latex
         2: punctuation.definition.backslash.latex
       push:
-        - [{meta_scope: meta.function.box.latex}, {match: '', pop: 1}]
+        - box-meta
         - argument
         - argument
         - optional-arguments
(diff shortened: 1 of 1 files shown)


deathaxe:
This PR includes countless improvements for TeX and LaTeX, which might be worth releasing anytime soon.

There are propably more things to do, but those could be done in separate PR or dev branch.

jwortmann:
Just a few comments from looking at the syntax tests and comparing with the previous highlighting after opening a few files:

* #3608 is not included yet, which makes `^` and `_` look quite strange for one being highlighted as `keyword.operator` and the other as `punctuation.separator`. I would suggest to include the other PR first, before merging this one.
* Some plain TeX macros like `\centering` or `\thinspace` are now `constant.language`, what makes them appear quite different from LaTeX macros, e.g. `\newline`, in most color schemes, so that they stand out very much in a document. There is a big wall of reasoning text in the commit message 700ed591874f2338a5eab37cbb732f15570ece4b, but I wonder whether this scope change is really useful, and I also don't really get the classification as "constant-like objects" (i.e. the `constant` scope seems pretty arbitrary to me, why not e.g. `keyword`, or `variable.language`, ...). I wouldn't consider this as a blocking issue, but if you would ask me, I'd keep using the current `support.function` scope and I don't think that the differentiation between plain TeX and LaTeX macros is important to the user.
* The content of `\textsc{}` is now `markup.raw`, which seems also a bit surprising and arbitrary to me, especially because this scope gets a different background color in some color schemes. There is no particular scope name in the guidelines for smallcaps, but it mentions `markup.other` for "other" markup. So maybe `markup.other.smallcaps` would be a better choice here.

deathaxe:
> About #3608

I hasitated to merge this in due to comments about changed string scopes probably causing issues with LaTeXTools.

In general I can moderate only and/or give some technical advices about syntax/context structure as I am not involved with LaTeX in any way. I know what it is, understand some very basics, but probably won't use or spend time learning its details anytime soon. It's just no priority at the moment.

I am however happy to discuss or merge in proposals for further improvements.

ngc92:
I may have some time end of this/beginning of next month to look into this again.

michaelblyons (approved):

michaelblyons (dismissed):
@jwortmann 's comments are valid, but I'm in favor of "Merge and then iterate." Or someone can open a PR to this PR.

deathaxe:
I'd at least vote for updateing to sublime-syntax version 2 before.

FichteFoll:
I'll test out the branch on some of my old LaTeX files later when I find the time, though I expect no need for further changes in order to get this merged. Holding off my approval until then.

michaelblyons (approved):

FichteFoll (approved):
Well, I obviously forgot about this PR, but I just loaded it to look at a couple of my LaTeX files and they all looked normal, so here goes.

PR #3576, opened 2022-11-19
[TeX] Improved highlighting in macro definition bodies

ngc92:
This PR performs a slight reorganization of the macro-braces context, and increases the amount of constructs that are highlighted within. This also unifies (to the extend possible) the handling of "function bodies" in TeX and LaTeX.
Most importantly, now we have highlighting for the replacement-placeholders.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -283,6 +283,10 @@ contexts:
     - meta_scope: meta.function.body.tex
     - include: immediately-pop
 
+  macro-braces:
+    - match: (?=\{)
+      push: macro-braces-begin
+
   macro-braces-begin:
     - match: \{
       scope: punctuation.definition.group.brace.begin.tex
@@ -298,12 +302,12 @@ contexts:
   # only partially matched.
   macro-braces-content:
     - meta_scope: meta.group.brace.tex
-    - match: (?=\{)
-      push: macro-braces-begin
+    - include: macro-braces
     - include: macro-braces-end
     - include: macro-placeholders
     - include: controls
     - include: registers
+    - include: math-builtin
     - include: general-constants
     - include: general-commands
 
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -248,7 +248,6 @@ contexts:
     - include: else-pop
 
   def-commandname:
-    - clear_scopes: 1
     - meta_scope: meta.function.identifier.tex
     - match: (\\){{cmdname}}
       scope: entity.name.definition.tex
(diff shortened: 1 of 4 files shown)


LaTeX/TeX.sublime-syntax
```diff
@@ -277,30 +272,51 @@ contexts:
     - include: paragraph-pop
 
   def-definition:
-    - match: \{
-      scope: punctuation.definition.group.brace.begin.tex
-      set: def-definition-body
+    - match: (?=\{)
+      set:
+        - def-scope-as-body
+        - macro-braces-begin
```
deathaxe, line 278: This pattern can directly replace line 265-267 to avoid duplicate lookahead evaluation.

The whole `def-definition` context becomes somewhat obsolete

LaTeX/TeX.sublime-syntax
```diff
@@ -277,30 +272,51 @@ contexts:
...
+  def-scope-as-body:
+    - meta_scope: meta.function.body.tex
+    - include: immediately-pop
 
-  # within macros, it is possible that only part of some nested struture
-  # is present. To prevent this from causing problems, here we only match
-  # elements that are simple, i.e. that do not push to the stack.
   macro-braces:
+    - match: (?=\{)
+      push: macro-braces-begin
```
deathaxe, line 288: It's finally probably even better to directly consume the `{` and push into content to avoid the additional context switch.

```suggestion
    - match: \{
      scope: punctuation.definition.group.brace.begin.tex
      push: macro-braces-body
```

LaTeX/TeX.sublime-syntax
```diff
@@ -277,30 +272,51 @@ contexts:
...
+
+  macro-braces-begin:
     - match: \{
       scope: punctuation.definition.group.brace.begin.tex
-      push: macro-brace-group-body
+      set: macro-braces-content
```
deathaxe, line 293: I often found it useful in inheritance scenarios to split contexts into `-body` and `-content` to be able to prepend/append contexts to content as needed but keeping the `-end` context at the very beginning to ensure it is always matched first.

```suggestion
      set: macro-braces-body
      
  macro-braces-body:
    - meta_scope: meta.group.brace.tex
    - include: macro-braces-end
    - include: macro-braces-content      
```

LaTeX/TeX.sublime-syntax
```diff
@@ -277,30 +272,51 @@ contexts:
...
 
-  macro-brace-group-body:
+  # within macros, it is possible that only part of some nested struture
+  # is present. Don't include any context's here that break if they are
+  # only partially matched.
+  macro-braces-content:
     - meta_scope: meta.group.brace.tex
-    - include: brace-group-end
+    - include: macro-braces
+    - include: macro-braces-end
```
deathaxe, line 306: With the former suggestion `macro-braces-end` and `meta_scope` become obsolete in this context.

```suggestion
```

deathaxe (commented):
As a general note `clear_scope` in `def-argspec` and `def-commandname` should be removed because all contexts are just set onto stack after each other. Currently the main scope is cleared by them, which is not so ideal.

`clear_scope` is used in `\newcommand` macro as we multi-push multiple contexts with the latest one assigning a `meta.function` scope to everything, which needs to be cleared by the argspec, ... contexts.

LaTeX/TeX.sublime-syntax
```diff
@@ -277,30 +272,51 @@ contexts:
...
+        - macro-braces-begin
     - include: paragraph-pop
     - include: else-pop
 
-  def-definition-body:
-    - meta_scope: meta.function.body.tex meta.group.brace.tex
-    - include: macro-brace-group-body
+  def-scope-as-body:
+    - meta_scope: meta.function.body.tex
+    - include: immediately-pop
```
deathaxe, line 284: Contexts which are meant to apply meta_scopes only should exclude prototypes to prevent inheriting syntaxes to inject code via `prototype` context.

```suggestion
  def-scope-as-body:
    - meta_include_prototype: false
    - meta_scope: meta.function.body.tex
    - include: immediately-pop
```

Not sure about the context name. Such meta_scope only contexts are often named `...-meta`. Maybe something like `def-body-meta` would be suitable?

PR #3578, opened 2022-11-19
Update Python.sublime-build

ora:
Windows Python exe is incorrect, was py for some reason.  Builds fine for me now after change.


deathaxe:
`py` is the python launcher installed with Python for Windows.

see: https://github.com/sublimehq/Packages/pull/2379

PR #3581, opened 2022-11-19
[TeX] added scoping of "constants"

ngc92:
This adds scoping/highlighting for the built-in constant-like objects in TeX.

The decision which control sequences to include here is based more on expected usage, than on how this is defined in the underlying language. I think for a macro-processing system like TeX, this is probably the best approach. 

Thus I've considered as constants in this sense:
 * The underlying parameters of the TeX typesetting algorithm. 
 * Registers with pre-defined meaning
 * simple macros (i.e. without parameters) that insert a fixed element into the typesetting
 
In particular, a parameterless macro that a) assigns to some internal value, or b) changes how subsequent input tokens will be interpreted are not considered constants here.

The typesetting parameters and registers can be assigned-to, though for most of these this would typically only happen once in the beginning of the document (usually hidden from the user, as part of the TeX format that is used). However, it is possible to change these values in the middle of the document (though some **really** should not be changed), so in that sense these are not all technically constants. 

For the assignment, I've opted to not try to distinguish constants that are dimensions from constants that are pure numbers, so both get the same scoping. In extremely weird corner cases, this might lead to erroneous scoping, e.g. if you write 
```
\hyphenpenalty=5pt
```
then the current scoping would not know that the penalty is a number, so the `pt` are actually letters to typeset, and not part of the unit. Such cases should be very rare, though.

Similarly, since macros can hide part of the syntax, there are situations where we have no chance of knowing which scoping is correct. For example, in
```
\abovedisplayskip\value plus 5pt
```
if `\value` evaluates to some measurement, then the `plus 5pt` would be part of the measurement, otherwise, it is normal text.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -205,15 +205,13 @@ contexts:
     - include: else-pop
     - include: paragraph-pop
 
-  decimal-float-meta:
-    - meta_include_prototype: false
+  decimal-float:
     - meta_scope: meta.number.float.decimal.tex
-    - include: immediately-pop
+    - include: tex-dimension-suffix
 
-  decimal-integer-meta:
-    - meta_include_prototype: false
+  decimal-integer:
     - meta_scope: meta.number.integer.decimal.tex
-    - include: immediately-pop
+    - include: tex-dimension-suffix
 
   tex-dimension-suffix:
     - match: '{{dimunits}}'
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -175,10 +175,19 @@ contexts:
     - include: paragraph-pop
 
   tex-constant-value:
-    - match: -?\d+\.\d*
-      scope: meta.number.float.decimal.tex constant.numeric.value.tex
-    - match: -?\d+
-      scope: meta.number.integer.decimal.tex constant.numeric.value.tex
+    - match: ((-?)\d+(\.)\d*)({{dimunits}}?)
+      scope: meta.number.float.decimal.tex
+      captures:
+        1: constant.numeric.value.tex
+        2: keyword.operator.arithmetic.tex
+        3: punctuation.separator.decimal.tex
+        4: constant.numeric.suffix.tex
+    - match: ((-?)\d+)({{dimunits}}?)
+      scope: meta.number.integer.decimal.tex
+      captures:
+        1: constant.numeric.value.tex
+        2: keyword.operator.arithmetic.tex
+        3: constant.numeric.suffix.tex
     - match: (`)(\\)?({{charbycode}}|.)
       scope: meta.number.integer.tex
       captures:
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -56,6 +56,52 @@ variables:
   registers: |-
     (?x: box | count | dimen | muskip | skip | toks)
 
+  # here, we define variables for the names of the constants and parameters
+  # of the TeX algorithm. Some of these are defined in terms of pre-reserved
+  # registers instead of built-ins, but in terms of syntax highlighting this
+  # should not make a difference
+  penalties: |-
+    (?x: line | hyphen | exhyphen | binop | rel | club | widow
+    | displaywidow | broken | predisplay | postdisplay | interline
+    | floating | output | inter(?:display|footnote)line)
+
+  tracings: |-
+    (?x: online | macros | stats | paragraphs | pages | output | lostchars
+    | commands | restores)
+
+  skips: |-
+    (?x: baseline | line | par | (?:above|below)display(?:short)?
+    | left | right | top | splittop | tab | space | xspace | parfill
+    | (?:thin|med|thick)mu | normalbaseline | normalline | hide)
+
+  constants: |-
+    (?x: (?:pre)?tolerance | [hv]badness
+    | (?:doublehyphen|finalhyphen|adj)demerits
+    | looseness | pausing | holdinginserts | language | uchyph | lefthyphenmin
+    | righthyphenmin | globaldefs | maxdeadcycles | hangafter | fam | mag
+    | (?:escape|defaulthyphen|defaultskew|endline|newline)char
+    | delimiterfactor | time | day | month | year
+    | showboxbreadth | showboxdepth | errorcontextlines
+    | [hv]fuzz | overfullrule | [hv]size | maxdepth | splitmaxdepth
+    | boxmaxdepth | lineskiplimit | delimitershortfall | nulldelimiterspace
+    | scriptspace | mathsurround | predisplaysize | displaywidth | displayindent
+    | parindent | hangindent | [hv]offset | (?:small|med|big)skipamount
+    | normallineskiplimit | jot | {{skips}}skip | tracing{{tracings}}
+    | {{penalties}}penalty | maxdimen | centering | p@ | z@ | z@skip | voidb@x
+    | magstephalf | magstep\d | @ne | tw@ | thr@@ | sixt@@n | @cclv | @cclvi
+    | @m | @M | @MM | m@ne | count@ | dimen@ | dimen@i | dimen@ii | skip@
+    | toks@ | insc@unt | allocationnumber)
+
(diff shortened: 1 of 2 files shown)


deathaxe:
Made a little tweak to numbers in order to ensure `meta.number` scope covers everything including sign/value/prefix as this is how we want to scope numeric constants everywhere.

LaTeX/TeX.sublime-syntax
```diff
@@ -108,6 +154,55 @@ contexts:
...
+        2: keyword.operator.arithmetic.tex
+        3: constant.numeric.suffix.tex
+    - match: (`)(\\)?({{charbycode}}|.)
+      scope: meta.number.integer.tex
+      captures:
+        1: keyword.operator.tex
+        2: punctuation.definition.backslash.tex
+        3: constant.character.tex
+    - match: '{{dimunits}}'
+      scope: constant.numeric.suffix.tex
```
deathaxe, line 198: With modified numbers, do we need/want to keep it as fallback in case leading number is missing?

LaTeX/TeX.sublime-syntax
```diff
@@ -108,6 +154,55 @@ contexts:
...
+
+    - match: plus|minus
```
deathaxe, line 200: Are word breaks required to ensure not to highlight `plus` within `someplustext` or is something like `plus5` valid?

LaTeX/TeX.sublime-syntax
```diff
@@ -108,6 +154,55 @@ contexts:
...
+      scope: keyword.operator.tex
```
deathaxe, line 201:
```suggestion
      scope: keyword.operator.arithmetic.tex
```

ngc92:
This won't work if there are spaces between the number and the unit, I think.
While I think for readability one generally doesn't want to have a space there, it is allowed by TeX.

LaTeX/TeX.sublime-syntax
ngc92: no, I don't think the number is allowed to be missing

LaTeX/TeX.sublime-syntax
```diff
@@ -108,6 +154,55 @@ contexts:
...
+      captures:
+        1: constant.numeric.value.tex
+        2: keyword.operator.arithmetic.tex
+        3: constant.numeric.suffix.tex
+    - match: (`)(\\)?({{charbycode}}|.)
+      scope: meta.number.integer.tex
+      captures:
+        1: keyword.operator.tex
+        2: punctuation.definition.backslash.tex
+        3: constant.character.tex
```
deathaxe, line 196: I wonder if those char codes refer to escape sequences in other syntaxes. If so, we'd probably want to scope them `constant.character.escape` and drop `meta.number` as this has been used for number literals only.

```suggestion
    - match: (`)((\\)?(?:{{charbycode}}|.))
      captures:
        1: keyword.operator.tex
        2: constant.character.escape.tex
        3: punctuation.definition.backslash.tex
```

ngc92: The main purposes of the ^^ sequences, as I understand it, is to insert characters that you don't have on your keyboard, or are otherwise un-printable. So `^^Z`, for example, is not an escape sequence for the letter `Z`, but instead refers to, I think, the character in ascii code that you get if you subtract 65 from the character code of `Z`.

The `meta.number` is here, because the backtick is the operation "get the number associated with the character", so that the entire expression is just a convoluted way of getting a number.

deathaxe: Sounds like something such as unicode escapes (e.g.:`\u2041`) in python to me.

https://github.com/sublimehq/Packages/blob/4c02b6a139c6944a7e1a7cbe585f912de99bdcf1/Python/Python.sublime-syntax#L1976-L1981

LaTeX/TeX.sublime-syntax
```diff
@@ -175,34 +175,54 @@ contexts:
...
-        2: keyword.operator.arithmetic.tex
-        3: constant.numeric.suffix.tex
+    - match: (-?)\d+(\.)\d*
+      scope: constant.numeric.value.tex
+      captures:
+        1: keyword.operator.arithmetic.tex
+        2: punctuation.separator.decimal.tex
+      push:
+        - decimal-float-meta
+        - tex-dimension-suffix
```
deathaxe, line 185: Can units appear on the next line?

```
\foo=12
   pt
```

If not I'd probably keep going with capture groups

LaTeX/TeX.sublime-syntax
```diff
@@ -175,34 +175,54 @@ contexts:
...
+    - match: ((-?)\d+)
+      scope: meta.number.integer.decimal.tex constant.numeric.value.tex
```
deathaxe, line 187: `meta.number` is duplicate here and the whole number seems scoped `keyword.operator`.

grafik

LaTeX/TeX.sublime-syntax
```diff
@@ -175,34 +175,54 @@ contexts:
...
+
+  decimal-integer-meta:
+    - meta_include_prototype: false
+    - meta_scope: meta.number.integer.decimal.tex
+    - include: immediately-pop
+
+  tex-dimension-suffix:
+    - match: '{{dimunits}}'
+      scope: constant.numeric.suffix.tex
+      pop: 1
```
deathaxe, line 221: I'd probably avoid multi-push here and do the following:

```suggestion
  decimal-float:
    - meta_scope: meta.number.float.decimal.tex
    - include: tex-dimension-suffix

  decimal-integer:
    - meta_scope: meta.number.integer.decimal.tex
    - include: tex-dimension-suffix

  tex-dimension-suffix:
    - match: '{{dimunits}}'
      scope: constant.numeric.suffix.tex
      pop: 1
```

Suffix context is simple enough to not bloat compiled syntax significantly, while saving a context switch for such basic language term (which may occur often in general) may help keeping performance up. Furthermore base syntaxes should probably stay as simple as possible with regards to amount of contexts and nesting etc.

LaTeX/TeX.sublime-syntax
```diff
@@ -175,34 +175,54 @@ contexts:
...
+        - decimal-float-meta
+        - tex-dimension-suffix
+    - match: ((-?)\d+)
+      scope: meta.number.integer.decimal.tex constant.numeric.value.tex
+      captures:
+        1: keyword.operator.arithmetic.tex
+        2: constant.numeric.suffix.tex
+      push:
+        - decimal-integer-meta
+        - tex-dimension-suffix
```
deathaxe, line 193: Corresponding change for removing multi-push would be

```suggestion
      push: decimal-integer
```

LaTeX/TeX.sublime-syntax
deathaxe, line 185: ... and

```suggestion
      push: decimal-float
```

LaTeX/TeX.sublime-syntax
ngc92: in that case, yes, it is somewhat comparable. However, with the backtick, the entire expression still is treated like a number, e.g. 
```
\abovedisplayskip=`\^^Zmm
```
is a valid quantity meaning 25 millimeters.

LaTeX/TeX.sublime-syntax
ngc92: from the point of TeX, a single newline is (mostly) the same as a regular space, so yes.

LaTeX/TeX.sublime-syntax
ngc92: fixed.  Btw, is there a good way to check for something like this in the automatic tests?

deathaxe:
If somehow possible I'd always try to avoid skipping multi-push contexts via `pop: 2` etc. It works and we already did something like that, but that should always be the last option, IMHO. It's much harder to follow state changes then.

LaTeX/TeX.sublime-syntax
deathaxe: OK, than let's keep it as is.

ngc92:
yes, that's why I though it necessary to put a comment there :)
With the new changes, do we want to keep the `-value` and `value-content` scopes both, or combine them?

deathaxe:
I guess we could merge them. I'd probably even try to keep a single pattern for floats. That's what I did in other syntaxes at least.

PR #3583, opened 2022-11-19
[Python] enable auto match in console input

jrappen:


deathaxe (approved):
I'd personally up-vote that with syntax also being enabled.

I may be biased though as I enabled that feature for all widgets in _User/Widget.sublime-settings_.

jfcherng (approved):

keith-hall:
Does this mean that if we have `auto_match_enabled` set to false in our User preferences file, then this would take precedence? :(

jfcherng:
Then I guess this turns into "the console input widget should respect syntax-specific settings and ST settings"?

deathaxe:
That's the downside, yes. That's one reason why I kept that local so far.

deathaxe:
What I've learned from MarkdownEditing is it's probably easier to add a specific setting locally than to remove it again. I have no strong opinion about what's the better default here.

PR #3584, opened 2022-11-20
[Go] fix some operator scopes

jrappen:


Go/Go.sublime-syntax
```diff
@@ -374,28 +374,49 @@ contexts:
...
+    - match: '(==)|(!=)'
+      captures:
+        1: keyword.operator.comparison.equal.value.go
+        2: keyword.operator.comparison.not-equal.value.go
+    - match: '(<=)|(>=)|(<)|(>)'
+      captures:
+        1: keyword.operator.relational.less-equal.go
+        2: keyword.operator.relational.greater-equal.go
+        3: keyword.operator.relational.less.go
+        4: keyword.operator.relational.greater.go
```
deathaxe, line 400: I don't think we need too detailed scopes.

deathaxe: It would effect all syntaxes if we'd want to add that and the little value it provides is not worth the effort and possible performance implications.

mitranim:
> @mitranim Can you check your `match-star` vs. `match-operators` logic? It seems messed up sometimes.
> 
> image
> 
> Like here it probably should continue with the stars, but breaks because an operator is ahead.

Hmm don't see any problems in this example. There are only stars, which are correctly scoped, and no other operators.

(I agree with Deathaxe about not wanting too-detailed scopes for operators: it increases maintenance with no clear benefit.)

PR #3585, opened 2022-11-20
[LaTeX] xparse command definition support

ngc92:
This PR adds support for the `NewDocumentCommand` family of functions, available as the `xparse` package and part of the `LaTeX` kernel since 2020. 

To keep things a bit better organized, I've put all the xparse-related tests in a separate file.


ngc92:
in any case, should these scopes end in `.latex`, or in `.tex.latex`?

deathaxe:
Whether base scopes `.tex` or `.latex` are tested or not depends on desired accuracy. I wouldn't recommend to drop them in general. Most older syntaxes did not include them in syntax tests as it was very common for all of them being the same. Things have changed since we started to embed syntaxes into each other or even inherit them.

Therefore I'd rather recomment to be as specific as possible in general - especially if a test file targets its own syntax. We may be a bit more lazy when testing forein (embedded syntaxes) to avoid too many conflicts in case those change.

As all latex related scopes end with just `.latex` I don't think we need to modify them to `.tex.latex`. At least we've choosen the former solution in other syntaxes so far.

In SQL (#3046) we use `.sql` for basic statements and `.tsql` or `.mysql` for dialect specific stuff.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -102,6 +102,15 @@ contexts:
         - newcommand-definition
         - newcommand-commandname
 
+    - match: '(\\)(?:(?:New|Renew|Provide|Declare)(:?Expandable)?DocumentCommand)'
```
deathaxe, line 105: Maybe

1. keep unquoted for consistency reasons
2. append word break pattern

```suggestion
    - match: (\\)(?:(?:New|Renew|Provide|Declare)(:?Expandable)?DocumentCommand){{endcs}}
```

or

```suggestion
    - match: (\\)(?:(?:New|Renew|Provide|Declare)(:?Expandable)?DocumentCommand)(?:\*|(?![A-Za-z@]))
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +207,123 @@ contexts:
...
+          scope: punctuation.definition.group.brace.begin.latex
+          set: xparse-argspec-content
+
+    - include: paragraph-pop
+    - include: else-pop
+
+  xparse-argspec-content:
+    - clear_scopes: 1
+    - meta_scope: meta.function.parameters.latex
+    - include: brace-group-end
```
deathaxe, line 227: I wouldn't change stacking strategy of a single contexts for very good reasons. Assuming `pragraph-pop` is not required within braced groups _(as it doesn't appear in others)_ we should keep pushing here.

```suggestion
  xparse-argspec:
    - match: \{
      scope: punctuation.definition.group.brace.begin.latex
      push: xparse-argspec-content
    - include: paragraph-pop
    - include: else-pop

  xparse-argspec-content:
    - clear_scopes: 1
    - meta_scope: meta.function.parameters.latex
    - match: \}
      scope: punctuation.definition.group.brace.end.latex
      pop: 2
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +207,123 @@ contexts:
...
+    - include: macro-braces
+    - include: xparse-single-argspec
+    # not actually allowed, but included here to future-proof highlighting
+    - include: general-constants
+    - include: general-commands
+    - include: paragraph-pop
```
deathaxe, line 233: `paragraph-pop` doesn't appear in other comparable contexts. Is it really needed?

```suggestion
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +207,123 @@ contexts:
...
+
+  xparse-single-argspec:
+    # modifiers and simple specs
+    - match: '[+!movsb]'
+      scope: string.unquoted.latex meta.annotation.parameters.latex
+
+    - match: t
```
deathaxe, line 240: Do we probably need to check for word boundaries to prevent matching letters within unmatched tokens?

```suggestion
    - match: t\b
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +207,123 @@ contexts:
...
+      push:
+        - meta_scope: meta.function.parameters.default-value.latex
+        - match: '\}'
+          scope: punctuation.definition.group.brace.end.latex
+          pop: 2
+        - include: general-constants
+        - include: general-commands
+        - include: macro-braces
+    - match: (?=\})
+      pop: 1
```
deathaxe, line 325: Maybe

1. use a named context
2. use a more generic bailout (`else-pop`) instead of `(?=\})`?
3. maybe add a `clear_scope: 1` to avoid stacking multiple `meta.function.parameter` scopes? _(not part of my suggestion)_
```suggestion
    - match: \{
      scope: punctuation.definition.group.brace.begin.latex
      push: xparse-optarg-value-body
    - include: else-pop

  xparse-optarg-value-body:
    - meta_scope: meta.function.parameters.default-value.latex
    - match: \}
      scope: punctuation.definition.group.brace.end.latex
      pop: 2
    - include: general-constants
    - include: general-commands
    - include: macro-braces
```

LaTeX/syntax_test_xparse.tex
```diff
@@ -0,0 +1,230 @@
...
+%                  ^ punctuation.definition.group.brace.begin.latex
+%                   ^^^^^^^^^ entity.name.newcommand.latex
+%                   ^ punctuation.definition.backslash.latex
+%                            ^ punctuation.definition.group.brace.end.latex
+%                             ^^ meta.function.parameters.latex
+%                             ^ punctuation.definition.group.brace.begin.latex
+%                              ^ punctuation.definition.group.brace.end.latex
+%                               ^^^^^^^^^ meta.function.body.tex meta.group.brace.tex
+%                               ^ punctuation.definition.group.brace.begin.tex
+%                                       ^ punctuation.definition.group.brace.end.tex
```
deathaxe, line 17: Not a big deal, but it is sometimes useful to group meta scope tests and "normal" scope tests especially if checking boundaries etc. is of interest.

```suggestion
%^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - meta.function meta.function
%^^^^^^^^^^^^^^^^^^ meta.function.latex keyword.declaration.function.latex storage.modifier.newcommand.latex
%                  ^^^^^^^^^^^ meta.function.identifier.latex
%                             ^^ meta.function.parameters.latex
%                               ^^^^^^^^^ meta.function.body.tex meta.group.brace.tex
%                                        ^ - meta.function
%                  ^ punctuation.definition.group.brace.begin.latex
%                   ^^^^^^^^^ entity.name.newcommand.latex
%                   ^ punctuation.definition.backslash.latex
%                            ^ punctuation.definition.group.brace.end.latex
%                             ^ punctuation.definition.group.brace.begin.latex
%                              ^ punctuation.definition.group.brace.end.latex
%                               ^ punctuation.definition.group.brace.begin.tex
%                                       ^ punctuation.definition.group.brace.end.tex
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +207,123 @@ contexts:
...
+      captures:
+        1: punctuation.definition.backslash.latex
+      pop: 1
+    # or the token wrapped in a group
+    - match: \{
+      scope: punctuation.definition.group.brace.begin.tex
+      push: xparse-args-token
+    - match: \}
+      scope: punctuation.definition.group.brace.end.tex
+      pop: 1
```
deathaxe, line 305: With a look how this context is used, I have a feeling this could cause unbalanced closing braces to be consumed by accident.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +207,123 @@ contexts:
...
+    - include: general-commands
+    - include: paragraph-pop
+
+  xparse-single-argspec:
+    # modifiers and simple specs
+    - match: '[+!movsb]'
+      scope: string.unquoted.latex meta.annotation.parameters.latex
+
+    - match: t
+      scope: string.unquoted.latex meta.annotation.parameters.latex
```
deathaxe, line 241: I guess those are flags, just like shellscript flags (e.g.: `cmd -f`)?

In that case we could/should probably use the same scope as Bash...

```suggestion
      scope: meta.parameter.option.latex variable.parameter.option.latex
```

LaTeX/LaTeX.sublime-syntax
ngc92: This is a valid declaration of a new command:
```
\NewDocumentCommand{\MyMacro}{tm}{\IfBooleanTF{#1}{A}{B}}
```
This will result in `A` is used as `\MyMacro m` and in `B` is used as `\MyMacro`, i.e. the letter `m` is the token that the t-type argument will look for here. So `t` should be scoped as an argument specifier, whereas the `m` is the literal thing that is looked for.

deathaxe: Ok, so it basically follows the way how shellscript can chain flags. Thank's for clarification.

LaTeX/LaTeX.sublime-syntax
ngc92: it is needed in the sense that it will improve the highlighting while you are typing, if the closing brace is not there yet.

LaTeX/LaTeX.sublime-syntax
ngc92: I think I've fixed this up now, with better scoping for the case if there are no {} and avoiding double meta-scopes

LaTeX/LaTeX.sublime-syntax
ngc92: I'm not sure if that is an adequate comparison. (Almost) each letter represents one of the arguments that can be passed to the function that is being defined, e.g. `m` indicates that there is a mandatory argument, `o` indicates that there is an optional argument that needs to be wrapped in `[]`, `t` indicates a boolean argument that checks for the occurrence of a specific token. 
The exceptions are `+!`, which are modifiers that apply to the next letter that specifies an argument.

deathaxe: So those are more like data types or better modifiers (`storage.type`, `storage.modifier`)?

LaTeX/LaTeX.sublime-syntax
deathaxe: Pushed a proposal to handle it via `set`

LaTeX/LaTeX.sublime-syntax
deathaxe: Those are actually types (see: https://mirror.informatik.hs-fulda.de/tex-archive/macros/latex/contrib/l3packages/xparse.pdf page 2).

ngc92: well, that really depends on how you understand the word "type". In python, would you consider the two arguments with name `x` to be of different type? 
```
f(x)
g(*, x)
```
it is a specifier of how arguments to this function are picked up.

ngc92:
I just noticed that I made a few changes based on your suggestions, but did not actually push. Merge seems to work without conflict, though.

ngc92:
concerning meta-scoping, should we put the things between `{}` into a `meta.group.brace` scope, like we do in other places?

LaTeX/LaTeX.sublime-syntax
deathaxe: There's no perfect matching scope for it as this kind of argument list specification seems somewhat unique to LaTeX. We could invent one or interpret existing ones in ways which match the language scheme. Maybe using `storage.modifier` for all of them would be fine, too, or `keyword.declaration` or `keyword.other` if in doubt.

I am just not convinced of `string` scope. `meta.annotation` is reserved for decorators and attributes (e.g.: C#).

LaTeX/LaTeX.sublime-syntax
deathaxe: These lines result in function parameters not beeing terminated in case a stray `}` appears after a type specifier. 

While it would be ok in:

```latex
\NewDocumentCommand {\conjugate} { t } O{#1ed{}} O{#2} } {(#1,#2,#3)}
%                                ^^^^^^^^^^^^^^^^^^^^^^^ meta.function.parameters
```

... function body would be turned into parameters in scenarios, like:

```latex
\NewDocumentCommand {\conjugate} { t } {(#1,#2,#3)}
%                                ^^^^^^^^^^^^^^^^^^ meta.function.parameters
```

None of them is ideal, but I tend to recommend to prefer correctly stacked braced contexts. It has proven robust in other syntaxes of complex code architectures at least.

```latex
\NewDocumentCommand {\conjugate} { t } {(#1,#2,#3)}
%                                ^^^^^ meta.function.parameters
%                                     ^^^^^^^^^^^^^ - meta.function.parameters
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +208,134 @@ contexts:
...
+  xparse-optarg-brace-body:
+    - match: \}
+      scope: punctuation.definition.group.brace.end.latex
+      pop: 2
+    - include: general-constants
+    - include: general-commands
+    - include: macro-braces
+    # The optional value is a long argument (can span paragraphs), so no
+    # paragraph-pop here!
+
```
deathaxe, line 338: I'd try to clearly distinguish braced and unbraced value types by using two different contexts to implement them. It may be more robust, avoids cross references and is therefore more flexible with regards two which constructs are supported in each.

```suggestion
  xparse-optarg-value:
    - meta_include_prototype: false
    - match: (?={)
      set:
        - xparse-optarg-group-body
        - brace-group-begin
    - match: (?=\S)
      set: xparse-optarg-value-body
    - include: paragraph-pop

  xparse-optarg-group-body:
    - clear_scopes: 1
    - meta_scope: meta.function.parameters.default-value.latex meta.group.brace.latex
    - include: brace-group-end
    - include: general-constants
    - include: general-commands
    - include: macro-braces
    # The optional value is a long argument (can span paragraphs), so no
    # paragraph-pop here!

  xparse-optarg-value-body:
    - clear_scopes: 1
    - meta_scope: meta.function.parameters.default-value.latex
    - match: (\\){{cmdname}}{{endcs}}
      scope: support.function.latex
      captures:
        1: punctuation.definition.backslash.latex
    - match: (?=\})
      pop: 1
    - match: \S
      pop: 1
    - include: paragraph-pop

```

I am a bit uncertain about unbraced default values as I can't find any statements about what's supported here. This context looks a bit suspicious to me, because `support.function` pattern doesn't pop, while a single non-whitespace `\S` does after being consumed.

How about `macro-placeholders`? They seem to be supported within default values.

```latex
\NewDocumentCommand {\margins} { O{#3} m O{#1} m } {(#1,#2,#3,#4)}
```

LaTeX/LaTeX.sublime-syntax
ngc92: you're right. should we mark the offending brace as `invalid`, or just skip out of the context?

LaTeX/LaTeX.sublime-syntax
ngc92: yes, macro placeholders are supported. I think the single command should also pop. If there are no braces, then TeX reads the next token, which is either the next non-whitespace character or the command sequence.

deathaxe:
Defacto realtime collaboration :-)

LaTeX/LaTeX.sublime-syntax
deathaxe: Fail early and skip the rest is often the best and most general approach. It's very likely to have incomplete code while editing. Too many invalid scopings may quickly become annoying.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -198,6 +208,153 @@ contexts:
...
+      set: xparse-optarg-bare-value
+    - include: paragraph-pop
+
+  xparse-optarg-wrapped-value:
+    - clear_scopes: 1
+    - meta_scope: meta.function.parameters.default-value.latex meta.group.brace.latex
+    - include: brace-group-end
+    - include: general-constants
+    - include: general-commands
+    - include: macro-braces
```
deathaxe, line 342: I've seen `#1` being supported in default values..

```suggestion
    - include: macro-braces
    - include: macro-placeholders
```

deathaxe: Maybe more things like `tex-xonstants` are valid here?

ngc92: yes

ngc92: I'll change this to `macro-braces-content`, and add some tests

LaTeX/syntax_test_latex.xparse.tex
```diff
@@ -259,6 +269,17 @@
 %                   ^^^^^^^^^^ entity.name.newcommand.latex
 %                                 ^^ constant.character.latex constant.other.token.latex
 
+\NewDocumentCommand{\MyMacro}{e{\up \down}}{up: #1 down: #2}
+%                             ^ storage.type.latex
+%                              ^ punctuation.definition.group.brace.begin.latex
+%                               ^^^
```
deathaxe, line 275: ?

LaTeX/LaTeX.sublime-syntax
```diff
@@ -723,6 +706,14 @@ contexts:
...
           pop: true
 
+  xparse-args-bare-tokens:
+    - match: (\\){{cmdname}}{{endcs}}
+      scope: constant.other.token.latex
+      captures:
+        1: punctuation.definition.backslash.latex
+    # or a single character
+    - match: '\S'
+      scope: constant.character.latex constant.other.token.latex
```
deathaxe, line 717: What is this unused context meant for?

ngc92: oh, I added this when I wanted to unify the embellishment and token parts. In principle, for a token argument, only one token is valid, whereas within braces you can specify multiple tokens for an embellishment. But as we want to balance braces anyway, it turns out that the braced token context handles multiple tokens in any case, so the two constructs actually can share the same context, so this turned out to be unnecessary.

deathaxe:
What do you think about naming contexts like

- latex-newcommand
- xparse-newcommand

Basically this PR adds another set of New...Command macros which can replace latex's default newcommand macro. So expressing that in context name may be useful in case more macros are added in future?

ngc92:
> What do you think about naming contexts like
> 
>     * latex-newcommand-...
> 
>     * xparse-newcommand-...
> 
> 
> Basically this PR adds another set of New...Command macros which can replace latex's default newcommand macro. So expressing that in context name may be useful in case more macros are added in future?

I guess that makes sense. We probably should make a pass over all the macro-definition scopes (including the TeX ones) to make these consistent.

deathaxe:
Yes, we should. Already found a couple mixes of `.tex` vs. `.latex` in group punctuation `{` vs. `}`.

I wonder if we should in general scope all groups `.tex`.

ngc92:
> Yes, we should. Already found a couple mixes of `.tex` vs. `.latex` in group punctuation `{` vs. `}`.
> 
> I wonder if we should in general scope all groups `.tex`.

Probably. Grouping is a language feature that LaTeX inherits from TeX, so even if grouping is part of a "call" of a LaTeX macro, the underlying structure is the TeX one.

This actually leads to a few cases where the new xparse context still don't pick up "correct" (i.e., LaTeX will accept it) code, because when reading the next token, TeX strips away one optional pair of braces to consume an entire group. Which means that you could wrap the character constants for the argument specifiers (the `m`, `t` etc.) in braces and things would still work. However, since the braces really just group the next token, any argument specifier that takes additional parameters need to have them outside the braces, so e.g. `{O}{default}` would be correct, but `{O{default}}` would complain that the argument type `O{default}` does not exist. 
I don't think we need to support these weird corner cases, though.

deathaxe:
We can still add it if someone really complains. To keep things simple, I'd suggest to create type patterns like ...

```yaml
    - match: ({)[rd](})
      scope: meta.group.brace.tex
      captures:
        1: punctuation.definition.group.brace.begin.tex
        2: storage.type.latex
        3: punctuation.definition.group.brace.end.tex
      push:
        - xparse-newcommand-args-token
        - xparse-newcommand-args-token
```

ngc92:
This improves the situation, but you can still construct corner cases where it breaks. For example, I can place a comment
to get the following construction
```LaTeX
\NewDocumentCommand{\MyMacro}{{% comment
     m% and here
   }}{test #1}
```

deathaxe:
We'll probably always find such edge cases. The question is just how likeley they are and whether they are worth the required extra complexity to support it.

ngc92:
exactly. My point is that I though the `{m}` cases unlikely enough that they are not worth the additional effort.
If you want to include that, though, I wouldn't be much opposed. 
In any case, it would probably be good to have a test case, to check that even if we don't highlight that particular part correctly, the overall highlighting doesn't break.

PR #3590, opened 2022-11-23
[Rust] More small tweaks to type matching

FichteFoll:
- Fixes highlighting of unknown generic types
- Adds stacking of `storage.type` under `support.type`

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -525,7 +525,6 @@ contexts:
       scope: keyword.operator.arithmetic.rust
     - include: integers
     - include: block
-    - match: '{{identifier}}'
     - match: ':|,'
       scope: punctuation.separator.rust
     - match: '\+|='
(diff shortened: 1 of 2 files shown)


deathaxe:
Why stacking two type scopes?

With a closer and careful look at original TextMate scope naming guidelines `storage.type` is meant for any basic builtin data type, such as `int`, `class`, ..., while `support.type` is meant for any kind of library, 3rd-party or user defined data type.

We already abused that a bit by introducing `keyword.declaration` and limiting `support` scope to standard library like types.

I wouldn't start stacking those different kinds of data type scopes.

keith-hall:
I was wondering the same thing myself :)

FichteFoll:
Stacking was primarily inspired from the discussion at https://github.com/sublimehq/Packages/issues/1842#issuecomment-477158241. The discussion never reached a conclusion though and I also see that this sort of stacking is not used elsewhere, so I will drop the commit.

deathaxe (approved):

keith-hall (approved):

PR #3593, opened 2022-11-26
[TeX] Modifiers for \def's

ngc92:
This adds scoping for the `outer`, `global`, and `long` modifiers available to alter the behaviour of TeXs `\def` command.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -409,6 +409,11 @@ contexts:
         1: punctuation.definition.backslash.tex
       push: def-function-expect-identifier
 
+    - match: (\\)(:?outer|long|global){{endcs}}
+      scope: storage.modifier.definition.tex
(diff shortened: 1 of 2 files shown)


ngc92:
I'm not sure if the scope should be `storage.modifier.tex` or `storage.modifier.definition.tex`. We've used the former for the `+` modifier in `xparse`, which has a similar meaning to `\long` (as in, the arguments can be paragraphs). On the other hand, these are modifiers that affect the following definition (opposed to just the following parameter), so maybe the current approach is better?

deathaxe:
With `storage.modifier` in xparse a very basic scope is added without further details about its meaning.

We don't have further 3rd level scope conventions. So we could just ommit it if all modifiers have more or less the same meaning or we add further detailed scopes if we want to enable color schemes to highlight different types differently. In general, I'd be ok with keeping things as simple as possible in the first place.

PR #3596, opened 2022-11-30
[JSON] Bind jsonc extension by default

deathaxe:
This commit binds jsonc file extension to JSON.sublime-syntax.

see: https://github.com/SublimeText/AFileIcon/issues/63

### DIFF
--- a/JSON/JSON.sublime-syntax
+++ b/JSON/JSON.sublime-syntax
@@ -6,6 +6,7 @@ version: 2
 
 file_extensions:
   - json
+  - jsonc
   - sublime-build
   - sublime-color-scheme
   - sublime-commands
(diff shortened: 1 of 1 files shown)


jfcherng (approved):

michaelblyons:
`eslintrc.json` is in there twice. 😉

michaelblyons (approved):

michaelblyons:
Screen Shot 2022-11-30 at 4 22 20 PM

deathaxe:
`.eslintrc.json` should already be coverd by normal `json` extension.

Extensions like `.schema` are probably too generic to be added to a default package as those might be used by various tools of different syntax.

For most of the other extentsions A File Icon provides alias syntaxes to provide special icons.

With JSON beeing a very generic syntax used for many things these days a list of possible extensions to support may probably be endless. Not saying we shouldn't add most commonly used ones. The responsible PR exists - just need to revice some syntaxes to be not too restrictive.

PR #3599, opened 2022-12-03
[TypeScript] Fix deadlock caused by type expressions

### LINKED ISSUE #3598: Hang while pressing `(` with `auto_match_enabled: false` in TypeScript files
### Description of the bug

The process hangs, all windows becoming unresponsive, killing Sublime Text is required, when doing the following:

### Steps to reproduce

1. Start ST in Safe Mode
2. Open Preferences and set it to:
```json
{
  "auto_match_enabled": false,
}
```
3. Create a new file, set syntax to TypeScript, enter the following:
```typescript
type Foo = {
  // With the cursor at the end of the next line, press open parentheses
  bar: (cb:
};
```
4. Press `(` with the cursor after `cb:`

### Expected behavior

Doesn't hang =)

### Actual behavior

Sublime Text hangs

### Sublime Text build number

4143

Windows 11 22H2 22621.819

### (Linux) Desktop environment and/or window manager

_No response_

### Additional information

_No response_

### OpenGL context information

_No response_

deathaxe:
Fixes #3598

This commit refactors anonymous type functions to...

1. ensure one of type-function or type-group to be applied. Old implementation may have caused both branches to fail, which can cause an infinite loop.
2. ensures not to invoke `fail` command in generic type functions, which are not part of a branch point.
3. add named contexts to differentiate generic from normal type functions and all their different parts.

_Note: The syntax needs to be applied to Sublime Merge in order to use it for review. Otherwise it will end up in the endless loop, caused by the test case which verifies the bug is fixed._

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -888,20 +888,8 @@ contexts:
     - include: ts-type-special
     - include: ts-type-primitive
     - include: ts-type-basic
-
-    - match: (?=\<)
-      set:
-        - ts-type-function
-        - ts-generic-function-type-check
-        - ts-type-parameter-list
-
-    - match: (?=\()
-      pop: 1
-      branch_point: ts-function-type
-      branch:
-        - ts-type-function
-        - ts-type-parenthesized
-
+    - include: ts-type-generic-function
+    - include: ts-type-function-or-group
     - include: literal-string
     - include: literal-number
     - include: ts-type-template-string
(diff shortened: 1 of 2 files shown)


Thom1729 (approved):

Thom1729:
Thanks for the fix.

One alternative approach might be to reverse the branch order for `(` and guess a parenthesized type before a function type. This would match the behavior for expressions. Not sure why I did it the other way; I vaguely recall that I might have forgotten about parenthesized types and added them in as an afterthought.

deathaxe:
While it would help to reuse function-parameter-list context in generic type functions I am not sure what would be needed to reverse branch order. It seems to cause various highlighting issues which are probably more hard to tackle.

keith-hall (approved):

PR #3606, opened 2022-12-05
[YAML] Fix block scalar headers

### LINKED ISSUE #3605: [YAML] Valid block scalar header marked as invalid
### What happened?

Indented blocks in YAML are interpreted with the help of a [block scalar header](https://yaml.org/spec/1.2.2/#811-block-scalar-headers). The header includes two pieces of information:

- The [indentation indicator](https://yaml.org/spec/1.2.2/#rule-c-indentation-indicator), a single digit `1`–`9`, of how much leading whitespace to remove from content lines.
- The [chomping indicator](https://yaml.org/spec/1.2.2/#rule-c-chomping-indicator), `-` or `+`, which controls how trailing whitespace is handled.

The YAML spec says that you can provide these indicators [in either order](https://yaml.org/spec/1.2.2/#811-block-scalar-headers). So these are both valid YAML:

```yaml
text: |1-
 Some text

text2: |-1
 Some more text
```

However, Sublime renders the second as invalid:

image

The issue is on [line 698](https://github.com/sublimehq/Packages/blob/6b057cb74482d88518425f0b5c5a27f409c268da/YAML/YAML.sublime-syntax#L698) of the `YAML.sublime-syntax` file:

```
  block-scalar:
    # http://www.yaml.org/spec/1.2/spec.html#style/block/scalar
    # c-l+literal(n) | c-l+folded(n)
    - match: (?:(\|)|(>))([1-9])?([-+])?  # c-b-block-header(m,t)
      captures:
        1: keyword.control.flow.block-scalar.literal.yaml
        2: keyword.control.flow.block-scalar.folded.yaml
        3: constant.numeric.indentation-indicator.yaml
        4: storage.modifier.chomping-indicator.yaml
      push: block-scalar-begin
```

It only matches the number-first orientation.

deathaxe:
Fixes #3605

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -695,12 +695,14 @@ contexts:
   block-scalar:
     # http://www.yaml.org/spec/1.2/spec.html#style/block/scalar
     # c-l+literal(n) | c-l+folded(n)
-    - match: (?:(\|)|(>))([1-9])?([-+])?  # c-b-block-header(m,t)
+    - match: (?:(\|)|(>))(?:([1-9])([-+])|([-+])?([1-9])?)  # c-b-block-header(m,t)
       captures:
         1: keyword.control.flow.block-scalar.literal.yaml
         2: keyword.control.flow.block-scalar.folded.yaml
         3: constant.numeric.indentation-indicator.yaml
         4: storage.modifier.chomping-indicator.yaml
+        5: storage.modifier.chomping-indicator.yaml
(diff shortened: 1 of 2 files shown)


michaelblyons:
Is it necessarily a 1-digit number? Also is 0 allowed? `(0|[1-9]\d*)` or `([1-9]\d*)`?

deathaxe:
That's correct (see https://yaml.org/spec/1.2.2/#rule-c-indentation-indicator)

PR #3601, opened 2022-12-03
[CSS] Fix html style tag folding not working

deathaxe:
This commit re-enables indentation based folding in CSS to fix an issue which causes `<style>` tags not being foldable in HTML syntax anymore.

As `source.css` begins immediately after the style tag, CSS's folding settings start ot apply directly after a style tag. ST therefore disables indentation based folding of `<style>` tags.

reported at:
- https://forum.sublimetext.com/t/can-no-longer-fold-html-style-tags/66219
- https://forum.sublimetext.com/t/code-folding-doesnt-work-js-go-html-maybe-more/66250/7
- https://forum.sublimetext.com/t/css-code-folding/66404

Other users reported folding issues in scenarious they use indentation to organize complex CSS sources. Hence it's probably a bad idea in general to disable indentation based folding.

PR #3617, opened 2022-12-08
[All] Re-enable indentation based folding

deathaxe:
This commit re-enables indentation based folding for all syntaxes which have not yet been covered by other PRs.

Syntax based folding doesn't cover all possible use cases and therefore shouldn't be the only one.

Related with #3568, #3601

### DIFF
--- a/C#/Fold.tmPreferences
+++ b/C#/Fold.tmPreferences
@@ -5,8 +5,6 @@
     <string>source.cs</string>
     <key>settings</key>
     <dict>
-        <key>indentationFoldingEnabled</key>
-        <false/>
         <key>foldScopes</key>
         <array>
             <dict>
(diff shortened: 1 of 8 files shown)


deathaxe:
Maybe Lisp is simple enough to restrict it to syntax based folding, but several other syntaxes 

1. don't provide consistent `meta.[group|block|...]` scopes under all circumstances to distinguish fold regions reliably (e.g. expressions like `( code ( code ))` ). 

2. don't provide rules to fold statements like `switch case`

In D the `case` blocks are currently not foldable as those rely on indentation folding.

```
switch (a) {

   case 0:
      // not foldable code
      break;

   case 1:
      // not foldablecode
      break;
}
```

IMHO, both indentation strategies live good enough next to each other to be enabled at the same time by default.

Issue #3607, opened 2022-12-06
[Meta] 


ngc92:
### What happened?

In order to look more into how scopes are named in different languages, I wrote a script that extracts all scopes from all the syntaxes here, removing language-specific suffixes. While looking over the results (not in detail -- there are too many scopes to do this quickly), I noticed the following things:

Typos:
```
entity.name.function.prepocessor   {'.ruby', '.perl'}
```

Potential inconsistencies in naming:
```
invalid.illegal.expect-separator   {'.erlang'}
invalid.illegal.expected-mapping-separator   {'.json'}
invalid.illegal.separator-expected 
invalid.illegal.hash-expected   {'.git.rebase'}

invalid.illegal.identifier   {'.css', '.perl', '.groovy', '.java', '.js'}
invalid.illegal.illegal-identifier   {'.js'}

invalid.illegal.stray-brace   {'.python'}
invalid.illegal.stray.brace.curly   {'.python'}

keyword.control.conditional.else   {'.matlab', '.dosbatch', '.haskell', '.perl', '.tex', '.python', '.shell.bash', '.r', '.cs', '.php', '.java', '.js'}
keyword.control.else   {'.applescript'}

keyword.control.assert   {'.groovy'}
keyword.control.flow.assert   {'.python', '.java'}
keyword.other.assert   {'.d'}
meta.declaration.assertion   {'.groovy'}

keyword.control.loop.for   {'.matlab', '.dosbatch', '.perl', '.python', '.shell.bash', '.r', '.cs', '.php', '.java', '.ruby', '.js'}
keyword.control.for-loop   {'.ocaml'}
meta.do-while   {'.js'}

keyword.control.loop.repeat   {'.r'}
keyword.control.repeat   {'.applescript'}

keyword.control.return   {'.lua'}
keyword.control.flow.return   {'.matlab', '.dosbatch', '.perl', '.c++', '.python', '.shell.bash', '.r', '.cs', '.php', '.c', '.java', '.ruby', '.js'}

keyword.declaration.import   {'.java', '.scala', '.haskell'}
keyword.control.import   {'.objc', '.c++', '.objc++', '.python', '.git.config', '.ocaml', '.cs', '.makefile', '.c', '.ruby', '.d'}
keyword.import   {'.js'}
keyword.other.import   {'.go', '.groovy'}

meta.enum   {'.objc', '.c++', '.objc++', '.cs', '.php', '.c', '.java', '.ts', '.d', '.rust'}
[...]

PR #3618, opened 2022-12-09
[LaTeX] Text formatting

ngc92:
The PR reworks the handling of text formatting commands in LaTeX. In particular, in cases where there are commands within formatted text, e.g. `\textbf{\TeX in bold}`, we no longer format the command names. 

One challenge is that there are several font properties that can be changed independently: Font family (roman, sans serif, typewriter), font weight (bold, medium, (possibly) light), and font shape (upright, italic, slanted, small-caps). Most of these do not have any corresponding `markup` scopes and highlighting rules (at least with default color schemes). At this point, only a subset of nesting of these properties is actually handled: Bold/italic combination, and nested \emph emphasize, as well as ensuring that all the commands that do not change one of these properties still typeset their arguments according to the outer command.

No provision has been made for any form of triple-nesting, i.e. `\textbf{\textit{\textup{a}}}` would not result in bold `a` right now. Not sure if supporting this would be worth it.

Finally, there is also an \underline command that could be combined with bold/italic. So far, I've not included this, because in LaTeX underline comes with some caveats, so that package commands like `\uline` or `\ul` are often recommended.

A more general question, is there a way to generate variations of the same context without excessive copy-and-paste? I've seen that there is a json-macro plugin, but I'm not sure if this is officially supported.

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -709,6 +709,10 @@ contexts:
 
   # https://en.wikibooks.org/wiki/LaTeX/Fonts#Font_styles
   text-decorators:
+    - include: text-normal-select-font-family
+    - include: text-normal-select-font-shape
+    - include: text-normal-select-font-weight
+
     - match: (\\)textnormal{{endcs}}
       scope: support.function.textnormal.latex
       captures:
@@ -717,10 +721,6 @@ contexts:
         - textnormal-meta
         - text-normal-body
         - expect-text-markup-open-brace
-
-    - include: text-normal-select-font-family
-    - include: text-normal-select-font-shape
-    - include: text-normal-select-font-weight
 
     - match: ((\\)text)(\{)
       captures:
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -702,57 +702,23 @@ contexts:
           scope: punctuation.definition.verb.latex
           pop: true
 
+###[ FONTS ]####################################################################
+
+  # https://en.wikibooks.org/wiki/LaTeX/Fonts#Font_styles
   text-decorators:
-    - match: ((\\)emph)(\{)
+    - match: (\\)textnormal{{endcs}}
+      scope: support.function.textnormal.latex
       captures:
-        1: support.function.emph.latex
-        2: punctuation.definition.backslash.latex
-        3: punctuation.definition.group.brace.begin.tex
+        1: punctuation.definition.backslash.latex
       push:
-        - meta_scope: meta.function.emph.latex
-        - meta_content_scope: markup.italic.emph.latex
-        - include: brace-group-end
-        - include: main
-    - match: ((\\)textit)(\{)
-      captures:
-        1: support.function.textit.latex
-        2: punctuation.definition.backslash.latex
-        3: punctuation.definition.group.brace.begin.tex
-      push:
-        - meta_scope: meta.function.textit.latex
-        - meta_content_scope: markup.italic.textit.latex
-        - include: brace-group-end
-        - include: main
-    - match: ((\\)textbf)(\{)
-      captures:
-        1: support.function.textbf.latex
-        2: punctuation.definition.backslash.latex
-        3: punctuation.definition.group.brace.begin.tex
-      push:
-        - meta_scope: meta.function.textbf.latex
-        - meta_content_scope: markup.bold.textbf.latex
-        - include: brace-group-end
-        - include: main
-    - match: ((\\)texttt)(\{)
-      captures:
-        1: support.function.texttt.latex
-        2: punctuation.definition.backslash.latex
-        3: punctuation.definition.group.brace.begin.tex
-      push:
-        - meta_scope: meta.function.texttt.latex
-        - meta_content_scope: markup.raw.texttt.latex
-        - include: brace-group-end
-        - include: main
-    - match: ((\\)textsl)(\{)
-      captures:
-        1: support.function.textsl.latex
-        2: punctuation.definition.backslash.latex
-        3: punctuation.definition.group.brace.begin.tex
-      push:
-        - meta_scope: meta.function.textsl.latex
-        - meta_content_scope: markup.italic.textsl.latex
-        - include: brace-group-end
-        - include: main
+        - textnormal-meta
+        - text-normal-body
+        - expect-text-markup-open-brace
+
+    - include: text-normal-select-font-family
+    - include: text-normal-select-font-shape
+    - include: text-normal-select-font-weight
+
     - match: ((\\)text)(\{)
       captures:
         1: support.function.text.latex
@@ -761,6 +727,7 @@ contexts:
       push:
         - include: brace-group-end
         - include: main
+
     - match: ((\\)underline)(\{)
       captures:
         1: support.function.text.latex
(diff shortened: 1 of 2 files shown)


LaTeX/LaTeX.sublime-syntax
```diff
@@ -702,75 +702,382 @@ contexts:
...
+
+  text-bold-body:
+    - include: brace-group-end
+    - include: text-bold-select-font-family
+    - include: text-bold-select-font-shape
+    - match: \{
+      scope: punctuation.definition.group.brace.begin.tex
+      push: text-bold-brace-group-body
+    - include: main
+    - match: .
```
deathaxe, line 917: This pattern causes ST to create a token for each single character.

See how each char of "bold" and "still bold" is highlighted differently.

grafik

A possible solution/improvement is to use a pattern which matches all chars, which the other patterns are not expected to start with.

The following pattern:

```suggestion
    - match: '[^\\{}]+'
```

produces more sane results:

grafik

_Note: Effects the other two occurances in `italic` and `bold-italic` as well._

deathaxe:
The rest looks good to me. Nice modular context structure.

LaTeX/LaTeX.sublime-syntax
ngc92: its a bit more complicated than that, e.g. you suggestion would break `\textbf{bold $normal math$}` (which is currently missing from the tests)

LaTeX/LaTeX.sublime-syntax
```diff
@@ -702,75 +702,382 @@ contexts:
...
       captures:
-        1: support.function.textbf.latex
+        1: support.function.text.latex
         2: punctuation.definition.backslash.latex
         3: punctuation.definition.group.brace.begin.tex
       push:
-        - meta_scope: meta.function.textbf.latex
-        - meta_content_scope: markup.bold.textbf.latex
+        - meta_scope: meta.function.underline.latex
+        - meta_content_scope: markup.underline.underline.latex
```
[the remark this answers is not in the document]
ngc92: possibly. Maybe I wanted to differentiate it from `markup.underline.link.latex`

LaTeX/LaTeX.sublime-syntax
deathaxe: I'd vote to add tests for some of the more exotic cases as well in general. As those are often root cause of breaking highlighting _(lessons learned)_.

I already was concerned about such cases.

If in doubt we can go with your solution - it's not wrong - just not optimal from performance/memory concumption point of view.

michaelblyons: `\w+|.`?

deathaxe: If all special chars were known those could be added to the char class `[^\\{}%$]+`. Otherwise we could refine your suggestion to something like `.[\s\w]*` maybe.

deathaxe: If it was acceptable to also scope keywords `markup.bold`, ... a `meta_content_scope` may be used as well to scope everything within braces accordingly.

ngc92: I think I prefer the more refined match over making everything `markup.bold`. Explicitly opting out only for a few characters might work, because
a) even if they have special meaning in latex, it really hurts only for those for which we do specific processing
b) any sort of fancy latex code (directly) inside text markup is very unlikely. E.g., the `#`, `_`, `^`, `&` characters would be problematic, but I think they aren't actually possible inside well-formed latex, so even if we do get these wrong, nothing too bad would be happening.

ngc92:
> Not sure what that `./.meta/blacklist.txt` is for. Seems unrelated to this PR anyways.

Sorry, should not have been added. That was just me taking notes as part of my script for #3607

LaTeX/LaTeX.sublime-syntax
```diff
@@ -15,6 +15,9 @@ first_line_match: |-
     ^ \s* \%+ .*? -\*- .*? \blatex\b .*? -\*-  # editorconfig
   )
 
+variables:
+  simpletext: '.[^\\{}%$~&#_^]+'
```
deathaxe, line 19:
```suggestion
  simpletext: '.[^\\{}%$~&#_^]*'
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -702,75 +705,382 @@ contexts:
...
+
+  text-bold-body:
+    - include: brace-group-end
+    - include: text-bold-select-font-family
+    - include: text-bold-select-font-shape
+    - match: \{
+      scope: punctuation.definition.group.brace.begin.tex
+      push: text-bold-brace-group-body
+    - include: main
+    - match: .{{simpletext}}
```
deathaxe, line 920:
```suggestion
    - match: '{{simpletext}}'
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -702,75 +705,382 @@ contexts:
...
+    - match: (\\)emph{{endcs}}
+      scope: support.function.emph.latex
+      captures:
+        1: punctuation.definition.backslash.latex
+      push:
+        - emph-meta
+        - text-normal-body
+        - brace-group-begin
+    - include: main
+    - match: .{{simpletext}}
```
deathaxe, line 1013:
```suggestion
    - match: '{{simpletext}}'
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -702,75 +705,382 @@ contexts:
...
+
+  text-italic-brace-group-body:
+    - meta_scope: meta.group.brace.tex
+    - include: brace-group-end
+    - include: text-italic-body
+
+  text-it-and-bf-body:
+    - include: brace-group-end
+    - include: main
+    - match: .{{simpletext}}
```
deathaxe, line 1076:
```suggestion
    - match: '{{simpletext}}'
```

LaTeX/LaTeX.sublime-syntax
```diff
@@ -702,75 +705,382 @@ contexts:
...
+      scope: markup.bold.latex markup.italic.latex
+
+  texttt-body:
+    - include: brace-group-end
+    - include: main
+    - match: .{{simpletext}}
```
deathaxe, line 1082:
```suggestion
    - match: '{{simpletext}}'
```

deathaxe (commented):
1. `simpletext` should match on single chars as well, hence changing `+` to `*`.
2. as `simpletext` variable already starts with `.` it is redundant when using it, hence changing patterns.

ngc92:
I've fixed the matching, though in the other way around. I think that makes the name "simpletext" more fitting for the variable, as now it only matches the "simple" part, instead of matching an arbitrary character followed by simple things.

I've also added a test case that would have broken with the old pattern.

LaTeX/LaTeX.sublime-syntax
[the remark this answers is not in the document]
deathaxe: `markup.underline.underline.latex` has been used in LaTeX before.

If we want to distinguish links and normal underlines via 3rd-level scope, I'd suggest `markup.underline.other.latex` as `other` is a rather general pattern used in various situations for common stuff.

deathaxe:
Reorganized contexts a bit and changed `underline.underline` to `underline.other`. Would you please double check my changes to make sure I haven't introduced some regressions?

If you are fine with it, I guess we can merge it.

ngc92:
One question is if we want to treat underline like the other text commands. With the current setup, it provides nested underline/bold etc nicely, but it also underlines commands, comments etc. within its scope. This is especially noticeable with incomplete text, e.g. if you have `\underline{typing some text in the middle of the document`, and don' t have a closing brace yet, then the entire rest of the document ends up being underline.

deathaxe:
I was thinking about the same during review. But wouldn't applying the strategy of the other contexts mean to add a lot of more contexts to make sure all nested formattings are scoped as "underlined" properly? I don't think it's worth the complexity. Incomplete statements may always break highlighting in some way.

PR #3622, opened 2022-12-13
[ShellScript] Fix build system target

### LINKED ISSUE #3621: Cancel a shell script build
### Problem description

Currently the "Cancel Build" option is not available when building a shell script (.sh). This option is available with Python

### Preferred solution

Enable the cancel option

### Alternatives

not sure, however I can create a custom build system but that has other issues

### Additional Information

_No response_

deathaxe:
Fixes #3621

This commit ...

1. enables "Build: Cancel" to terminate running shell scripts by a) forwarding `kill` and `kill_previous` as well as various other possibly relevant variables from sublime-build file to default build executor. b) adding `cancel` key to ShellScript.sublime-build

2. enables running shell scripts using MSYS2's `sh` shell on Windows. _(Using shebang is not useful on Win.)_

3. removes obsolete special treatment of `working_dir` and `env`.  Merging `env` into OS's environment variables is already handled well by default build executor.


michaelblyons (approved):

ShellScript/ShellScript.py
```diff
@@ -2,12 +2,24 @@
...
-    def run(self, **kwargs):
+    def run(
+        self,
+        working_dir="",
+        encoding="utf-8",
+        env={},
+        kill=False,
+        kill_previous=False,
+        word_wrap=False,
+        syntax="Packages/Text/Plain text.tmLanguage",
```
FichteFoll, line 19: All the parameters that are equivalent to the default and are not modified or read in the override can be omitted, which afaics should be everything except `working_dir`.

deathaxe: I don't agree.

deathaxe: The command can theoretically be bound to a key in a sublime-keaymap file to call the command. The explicit presence of those arguments is required for PackageDev to provide proper argument completions.

```
{
	"keys": [""],
	"command": "run_shell_script",
	"args": {
		"working_dir": "",
		"encoding": "utf-8",
		"env": {},
		"kill": false,
		"kill_previous": false,
		"word_wrap": false,
		"syntax": "Packages/Text/Plain text.tmLanguage"
	},
},
```

deathaxe: Furthermore not passing those arguments (especially `kill`) to default `ExecCommand` was a primary reason for #3621 as well as `{kill: true}` being missing from sublime-build file.

FichteFoll: I guess I didn't make it clear that `**kwargs` should still be forwarded. 

PackageDev completions are an argument I understand, but that sounds much more like a PD implementation detail than a reason to make a default package plugin repeat the same arguments with defaults as the `exec` command that it eventually forwards everything to.

FichteFoll (approved):
Less code is always an improvement in my eyes.

ShellScript/ShellScript.py
deathaxe: Somehow didn't have `**kwargs` forwarding being a solution in mind at all.

michaelblyons (approved):

rwols:
I realize I’m severely late to the party, but this change removes the ability to run a bash script from a temporary buffer not saved on-disk, which was one of the goals of the ShellScript.py python script. I’m guessing it’s not that big of a deal since I don’t see many people complaining on the forums.
Although I haven’t dutifully checked whether there was a new release of ST with this change included.

rwols:
Regardless of my concerns I do believe I can let @deathaxe just be in charge of anything

deathaxe:
This change is not yet part of any release.

Directly running scripts from unsaved buffers being a feature was indeed not noticed as the old buildsystem didn't run on Windows at all.

A question arising is why default build systems should behave differenty. If running scripts from unsaved buffers is of interest, it should work for any script interpreter (e.g. Python) the same way.

Once a file is created for a buffer and "Save all before Build" is enabled, running scripts works as before. Shell scripts from unsaved scratch buffers will stop working. Well, how likely are sophisticated scripts being managed this way? Wouldn't saving them to disk not be a more save alternative anyway?

PR #3625, opened 2022-12-14
[CSS] Add new syntax features of 2022

deathaxe:
This PR proposes to add CSS some syntax features, which have been added to specs in 2022 by MDN.

Primary source of information are:

1. https://developer.mozilla.org/en-US/docs/Web/CSS
2. https://github.com/mdn/data

also see: https://web.dev/state-of-css-2022


jfcherng (approved):

deathaxe:
1. It makes sense especially if many and/or long scopes are assigned to limit line lengths. I am uncertain about it to be used in general. Nevertheless this is a more general question not directly related with this PR
2. Thank's for pointing that out. All at-rules use this style, not only changes made by this PR. Change it for consistency reasons.

michaelblyons (approved):

deathaxe:
Anything I can do, to get this merged?

jfcherng (approved):

PR #3628, opened 2022-12-17
[Graphviz] Fix embedded html termination

### LINKED ISSUE #3627: [Graphviz] Syntax highlighting not properly terminated when multiple html tags used
### What happened?

Syntax highlighting for the use of html in labels only works properly if all tags apply to the whole label. For example, this label breaks the syntax highlighting for all subsequent lines because sublime thinks that the html input has not been terminated, even though it is a valid input.

```graphviz
label = <<u>this is underscored</u><br/>this is not>
```

deathaxe:
Fixes #3627

This commit ...

1. includes HTML source into Graphviz to be able to correctly terminate it if an otherwise stray `>` is matched.
2. fixes embedded scope to `meta.embedded.dot text.html.embedded.dot`

### DIFF
--- a/Graphviz/DOT.sublime-syntax
+++ b/Graphviz/DOT.sublime-syntax
@@ -60,17 +60,16 @@ contexts:
   embedded-html:
     - match: '<'
       scope: punctuation.section.embedded.begin.dot
-      push:
-        - meta_content_scope: source.dot.embedded.html
-        - match: '((>)?\s*)(>)'
-          captures:
-            1: source.dot.embedded.html
-            2: meta.tag punctuation.definition.tag.end.html
-            3: punctuation.section.embedded.end.dot
-          pop: true
-        - match: ''
-          embed: scope:text.html.basic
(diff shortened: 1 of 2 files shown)


Graphviz/DOT.sublime-syntax
```diff
@@ -60,17 +60,16 @@ contexts:
...
-          escape: (?=>\s*>)
+      push: embedded-html-body
+
+  embedded-html-body:
+    - meta_scope: meta.embedded.html.dot
+    - meta_content_scope: text.html.embedded.dot
+    - match: '>'
+      scope: punctuation.section.embedded.end.dot
+      pop: true
+    - include: scope:text.html.basic#html
```
michaelblyons, line 71: While you're in here, maybe change to Plain?

deathaxe: If that is enough. No problem.

michaelblyons (approved):

keith-hall (approved):

PR #3631, opened 2022-12-20
[PHP] Adjust snippet selectors

deathaxe:
This commit adds `text.html.php` to those snippets which are currently triggered by `embedding.php - source.php` only.

The goal is to enable those snippets in inherit syntax such as Laravel Blade, which don't use `embedding.php` but will probably be scoped `text.html.php.blade`.

> **Note**
> 
> Basically `embedding.php` is maintained only in order to enable those snippets in an empty document, as `text.html.php` is added to stack not before the first character is typed.


jfcherng (approved):

michaelblyons (approved):

deathaxe:
I've decided to stay with `text.html.blade` in https://github.com/Medalink/laravel-blade/pull/195 as it was the scope being used before.

Will need to think about relevance of this PR and possible other implications about snippet selectors in templates, I've learned about during the last few days.

PR #3634, opened 2022-12-22
[PHP] Add strftime() placeholders

deathaxe:
This commit scopes all strftime fomrat placeholders. 

see: https://www.php.net/manual/en/function.strftime

inspired by: https://forum.sublimetext.com/t/color-schemes-doesnt-work-properly/66479

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -2688,6 +2688,7 @@ contexts:
 
   string-placeholders:
     # https://www.php.net/manual/en/function.printf.php
+    # https://www.php.net/manual/en/function.strftime
     - match: '%%'
       scope: constant.character.escape.php
     # %[argnum$][flags][width][.precision]specifier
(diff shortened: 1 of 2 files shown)


PHP/PHP Source.sublime-syntax
```diff
@@ -2716,7 +2718,7 @@ contexts:
         [-+]? (?: [ 0] | \\'. )?    # flags
         \d*                         # width
         (?: (\.) \d* )?             # precision
-        [bcdeEfFgGhHosuxX]          # specifier
+        [aAbBCcdDeEFfgGhHIjklmMnopPrRSsTtuUVwWXxyYzZ] # specifier
```
michaelblyons, line 2721: `C`, `F`, `S`, `T`, and `X` do not follow the lower-case-first pattern. The whole set can be simplified to `[A-IMPS-Za-hj-pr-z]`, but that might be hard to read. I don't know.

Alternatively, make variables and put them into both placeholder regexes.

```yaml
  format_printf: '[bcdeEfFgGhHosuxX]'
  format_strftime: '[somethingNEW]'
  format_specifier: (?:{{format_printf}}|{{format_strftime}})
```

deathaxe: Variables would cause various duplicates. Not sure if ST would optimize them away. Not sure if changes are to be expected so adding vars is justified.

deathaxe: I stayed with explicit list of chars to be able to compare them one-by-one with specs in case they change in future.

deathaxe: ... too lazy to sort test cases.

michaelblyons (approved):

jfcherng (approved):

PR #3638, opened 2022-12-30
[ShellScript] scope colon as a path separator inside quoted strings

keith-hall:
This makes it much easier to visualize e.g. setting the `PATH` env var in scripts because the `:` can be highlighted as a separate color to the rest of the string.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -1332,7 +1332,6 @@ contexts:
     # characters ‘\’, ‘$’, and ‘`’.
     - include: string-escapes
     - include: string-interpolations
-    - include: string-path-separators
 
   heredocs-body-no-expansion:
     - meta_scope: meta.string.heredoc.shell
@@ -1996,7 +1995,7 @@ contexts:
       scope: constant.character.escape.shell
 
   string-path-separators:
-    - match: ':(?!$|\s|//)'
+    - match: ':(?=\$|/(?!/))'
       scope: punctuation.separator.sequence.shell
 
   string-escapes-ansi-c:
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -1925,6 +1925,7 @@ contexts:
         - meta_include_prototype: false
         - meta_scope: string.quoted.double.locale.shell
         - include: string-quoted-double-body
+    - include: string-commons
     - match: (?!{{cmd_literal_char}})
       pop: 1
 
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -1332,7 +1332,7 @@ contexts:
     # characters ‘\’, ‘$’, and ‘`’.
     - include: string-escapes
     - include: string-interpolations
-    - include: string-commons
+    - include: string-path-separators
 
   heredocs-body-no-expansion:
     - meta_scope: meta.string.heredoc.shell
@@ -1925,7 +1925,7 @@ contexts:
         - meta_include_prototype: false
         - meta_scope: string.quoted.double.locale.shell
         - include: string-quoted-double-body
-    - include: string-commons
+    - include: string-path-separators
     - match: (?!{{cmd_literal_char}})
       pop: 1
 
@@ -1968,7 +1968,7 @@ contexts:
     - include: line-continuations
     - include: string-escapes
     - include: string-interpolations
-    - include: string-commons
+    - include: string-path-separators
 
   string-ansi-c-body:
     - match: \'
@@ -1983,7 +1983,7 @@ contexts:
       scope: punctuation.definition.string.end.shell
       pop: 1
     - include: string-prototype
-    - include: string-commons
+    - include: string-path-separators
 
   any-escapes:
     - match: \\.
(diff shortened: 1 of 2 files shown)


jfcherng:
```sh
export SOMETHING=/etc/test:/var/test
#                         ^ punctuation.separator.sequence
```
should pass

deathaxe:
We've used `punctuation.separator.path` for path segment separaturs (slashes), only so far - not as separator for a list of paths. Hence I'd also vote for `punctuation.separator.sequence`.

As colons might appear everywhere in arbritary strings with any kind of meaning, I am however not sure whether we should make any assumption about the meaning of certain tokens/chars in a string, beyound escapes and placeholders.

keith-hall:
I believe I have addressed the suggestions.

> As colons might appear everywhere in arbritary strings with any kind of meaning, I am however not sure whether we should make any assumption about the meaning of certain tokens/chars in a string, beyound escapes and placeholders.

I generally agree, but in this case I believe the usefulness of scoping common patterns outweighs the "pure syntax" perspective. But if we vote to not have this, I guess I will need to maintain a local override or extended syntax which I was hoping to avoid as I thought this could benefit everyone...

ShellScript/Bash.sublime-syntax
```diff
@@ -1991,6 +1995,10 @@ contexts:
     - match: \\[$`"\\]
       scope: constant.character.escape.shell
 
+  string-commons:
+    - match: ':(?!$|\s|//)'
+      scope: punctuation.separator.sequence.shell
```
deathaxe, line 2000: Having a `string-commons` next to `string-prototype` feels a bit duplicate.

I'd therefore suggest to 

1. either rename it to something which expresses the meaning of the contained pattern more precisly 
2. or to move this pattern to `string-prototype`. It was added as empty map as there was no content yet. It needs to be extended via `- meta_prepend` by extending syntaxes anyway.

2nd solution would also add highlighting of path separators in ansi-c strings but exclude it from heredocs. Not sure about ansi-c strings, but I'd probably would not highlight colons in heredocs as those may contain any kind of larger text blocks with higher risk of false positives, woudln't they?

keith-hall: agreed with all your points, so I went for the 1st solution of renaming the new context :)

ShellScript/Bash.sublime-syntax
```diff
@@ -1991,6 +1995,10 @@ contexts:
     - match: \\[$`"\\]
       scope: constant.character.escape.shell
 
+  string-path-separators:
+    - match: ':(?!$|\s|//)'
```
FichteFoll, line 1999:
```suggestion
    - match: ':(?=\$|/)'
```
Perhaps this would be a decent approximation for a list of allowed characters after a colon when combining multiple paths together? It should be very rare to combine two relative paths.

deathaxe: Theoretical should also handle `Class::method()`. There may be various possible false positives anyway.

ShellScript/Bash.sublime-syntax
```diff
@@ -1332,7 +1332,7 @@ contexts:
     # characters ‘\’, ‘$’, and ‘`’.
     - include: string-escapes
     - include: string-interpolations
-    - include: string-commons
```
deathaxe: Do we really want to highlight it in heredocs?

If yes, should probably also include the other kinds below.

ShellScript/Bash.sublime-syntax
michaelblyons: Found another accidental match in `'([^.[:space:]]+)'`

ShellScript/Bash.sublime-syntax
keith-hall: probably we don't, I have now removed it from here :)

ShellScript/Bash.sublime-syntax
keith-hall: applied FichteFoll's suggestion, and added syntax tests to prove those examples are not scoped/treated as path separators

deathaxe: May something like `/path/:~/dir` be something to considder? If yes, ... `:(?=\$|~?/(?!/))`

FichteFoll: And the current (or a relative) directory `/path:.:../sibling`.

Can't think of anything else right now.

keith-hall: done

deathaxe (approved):

ShellScript/Bash.sublime-syntax
FichteFoll: Not sure why you require `.` and `~` to be followed by a slash, which is not a technical requirement, but it should be unlikely enough to not be relevant in practice, at least for `~`. `.` I could see but is probably also equivalent to an empty segment (which I remember seeing sometimes but probably way back on Windows).

Current changes are fine imo, but feel free to revise.

FichteFoll (approved):

ShellScript/Bash.sublime-syntax
deathaxe: You mean something like ...

```
- match: ':(?=\$|~|\.\.?|/(?!/))'
```

I haden't `bla:~:bla` in mind when writing my proposal.

keith-hall: The current regex suits my personal use case, and I'm worried that making it more lenient could cause more false positives. My thinking would be to merge as is, and tweak it based on feedback later. But if we want, we can change it now also, I don't mind, just need someone to make the decision :)

deathaxe: I made it for you ;-) as somehow sharing concerns about false positives.

PR #3639, opened 2022-12-30
[meta] Refactor syntax test CI workflows

deathaxe:
This commit...

1. avoids moving files around by checking out syntax repository into a directory the test runner is directly extracted to afterwards.
2. uses new "latest" url to reduce maintenance effort after each dev build being released.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ on:
 
 jobs:
   run_syntax_tests:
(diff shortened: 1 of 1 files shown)


deathaxe:
If readability would really be of interest we should probably go the route `syntax-test-action` has choosen by evaluating output and printing proper `::error::...` messages.

FichteFoll (approved):

jfcherng (approved):

PR #3644, opened 2023-01-21
[D] Fix !in operator highlighting

### LINKED ISSUE #3643: [D] `!in` operator not recognized
### What happened?

`!in` operator is not parsed correctly:
```
if (foo !in bar) {
    // do stuff
}
```
image

deathaxe:
Fixes #3643

### DIFF
--- a/D/tests/syntax_test_d.d
+++ b/D/tests/syntax_test_d.d
@@ -1898,6 +1898,28 @@ extern(1)
 //               ^ punctuation.section.block.begin.d
 //                ^ punctuation.section.block.end.d
 
+  if (a in b) {}
+//^^ keyword.control.conditional.d
+//   ^ punctuation.section.parens.begin.d
+//    ^ variable.other.d
+//      ^^ keyword.operator.comparison.d
+//         ^ variable.other.d
+//          ^ punctuation.section.parens.end.d
+//            ^^ meta.block.d
+//            ^ punctuation.section.block.begin.d
+//             ^ punctuation.section.block.end.d
+
+  if (a !in b) {}
+//^^ keyword.control.conditional.d
+//   ^ punctuation.section.parens.begin.d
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):

D/tests/syntax_test_d.d
```diff
@@ -1898,6 +1898,28 @@ extern(1)
...
+//   ^ punctuation.section.parens.begin.d
+//    ^ variable.other.d
+//      ^^ keyword.operator.comparison.d
+//         ^ variable.other.d
+//          ^ punctuation.section.parens.end.d
+//            ^^ meta.block.d
+//            ^ punctuation.section.block.begin.d
+//             ^ punctuation.section.block.end.d
+
+  if (a !in b) {}
```
FichteFoll, line 1912: I'm missing a test on `!` here. Should not not be an operator as well?

deathaxe: Not sure I can follow. Are you missing tests like `if (!a)`? 

A `!` following an identifier (`if(a ! b)`) seems to denote `a` to be a function call.

Anyway, the scope of this PR is to fix `!in`.

deathaxe: I'll therefore merge it. Hard enough to get approvals for such basic fixes. We can add more tests at any time in separate PRs.

FichteFoll: According to the ticket this PR addressed, which speaks of an "`!in` operator", I expected the exclamation mark to be part of the `in` operator much like `not in` in Python, but I don't know anything about D.

deathaxe: I guess it has exactly this meaning, but as the basic implementation already has existed before, but was only not matched due to some lookaheads in "parent-contexts", there was no reason to change scoping.

Issue #3648, opened 2023-01-31
[RFC] keywords likely to appear at the end of a line should `cancelCompletion`


deathaxe:
### Relational

Discussions about variables or keywords being expanded to something invalid by hitting <kbd>enter</kbd> key at the end of a line.

The reason is ST removing the perfect matching completion from the auto completion popup, which causes the 2nd best match to be committed instead of adding a newline.

> **Example**
>
> If the word `thena` has been typed somewhere in the file and <kbd>enter</kbd> is hit directly after fully typed keyword `then` at the end of a line, it is expanded to `thena`, while adding a newline is expected.

### Proposal

> it's an issue with the default packages that ship with Sublime Text: all keywords that commonly appear in a trailing position on a line should be mentioned in a cancelCompletion preference, ideally

source: https://discord.com/channels/280102180189634562/280157083356233728/1069813308074491984

The following pattern causes completions to be canceled, if a line is finalized by the given list of keywords.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
	<key>scope</key>
	<string>source.lua</string>
	<key>settings</key>
	<dict>
		<key>cancelCompletion</key>
		<string>^.*\b(?:break|do|end|else|repeat|return|then)$</string>
	</dict>
</dict>
</plist>
```

with `^.*(?:<list of keywords>)$` matching only if the keywords really finalize the line, basically every known keyword could be added here without preventing further completions. As soon as any character other than newline is added after given keyword, the pattern doesn't match and completions are re-enabled again.

#### Extending existing statements

It doesn't however prevent known keywords from being expanded, if an existing statement is to be splitted into multiple lines.

**before <kbd>enter</kbd>**

```lua
if foo then| print("bar") end
```

**after hitting <kbd>enter</kbd>**

```lua
if foo thena
    | print("bar") end
```

#### Arbritary identifiers

It doesn't work if the final word in the line is an arbritary identifier, such as a well-known variable.

```lua
var foo;
var foobar;

foo|
```

expands to

```
foobar
|
```

PR #3654, opened 2023-02-03
[Matlab] Cancel completions after keywords at EOL

jwortmann:
This allows to press the <kbd>Enter</kbd> key in order to insert a linebreak after a keyword at the end of the line, without accidentally inserting the 2nd best completion match (see #3648). It should improve the behavior while typing, if the default value for the `"auto_complete_commit_on_tab"` setting is used.

If I didn't miss anything, the keywords which could in practice appear at EOL are `arguments`, `break`, `catch`, `continue`, `else`, `end`, `enumeration`, `events`, `methods`, `otherwise`, `properties`, `return`, `spmd`, `try`.
But I think it doesn't hurt to also close the completion popup after other keywords.

### DIFF
--- a/Matlab/Completion Rules.tmPreferences
+++ b/Matlab/Completion Rules.tmPreferences
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+    <key>scope</key>
+    <string>source.matlab</string>
+    <key>settings</key>
+    <dict>
+        <key>cancelCompletion</key>
+        <string>^.*\b(?:arguments|break|case|catch|classdef|continue|else|elseif|end|enumeration|events|for|function|global|if|methods|otherwise|parfor|persistent|properties|return|spmd|switch|try|while)$</string>
+    </dict>
+</dict>
+</plist>


Matlab/Completion Rules.tmPreferences
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+    <key>scope</key>
+    <string>source.matlab</string>
+    <key>settings</key>
+    <dict>
+        <key>cancelCompletion</key>
+        <string>^.*\b(?:arguments|break|case|catch|classdef|continue|else|elseif|end|enumeration|events|for|function|global|if|methods|otherwise|parfor|persistent|properties|return|spmd|switch|try|while)$</string>
```
deathaxe, line 9: I just had a look at #2620 as Python was mentioned in the discussion on discord and found it omitting `^.*` compared to the suggested pattern in #3648. This may (or may not?) have a positive effect on performance, maybe. Thoughts?

```suggestion
        <string>\b(?:arguments|break|case|catch|classdef|continue|else|elseif|end|enumeration|events|for|function|global|if|methods|otherwise|parfor|persistent|properties|return|spmd|switch|try|while)$</string>
```

jwortmann: I just tested it, and if I remove the leading `^.*`, the completion cancelling doesn't work anymore for long words, for example when I type `otherwise` (also tested and confirmed with `enumeration`). But it still works for short words, e.g. `else`.
If I keep the `^.*` in the pattern, it seems to work as expected for all words in the pattern.

[matlab-completions.webm](https://user-images.githubusercontent.com/6579999/216699690-23499ba4-7c9b-476a-8734-e6580e543f0d.webm)

deathaxe: That's an interesting detail.

deathaxe (approved):

PR #3660, opened 2023-02-04
[JavaScript] fix indentation for case/default blocks

bakkot:
In JavaScript, it is a universal convention that `case` labels are indented one level beyond the containing switch - see e.g. the behavior of [Prettier](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBnA7gSxmAFgAgAooBXAGzIEp9gAdKMAQ1TnwEYl798AjAJziMA1gG56TFvgBMnBs1YBmJDS68BwsVAC+9HVBAAaEBAAOMTNFTJQjPnwjoACrYRWUjMukYBPK0f6MYEJwMADKjAC2cAAymFBwyABmHiz+fIHBYSaBcQDmyDB8JHBGcBE8cAAmlVXRjFC5JIy5cABiEHwRjDDmDcggjCQwEIYguDARZADquNhwqNlgcKGu2JgAbtje-WCofiBxLHwwjum5XUkpJSAAVqgAHqF5ZHAAiiQQ8JdkqSDZfEd+jxGBUyKMTHw4jApphKjBcMgABwABiMEIgLCm6RM-Qh8zgfHWCSMAEcPvBTqY3ANUABaeJVKqjARkzACU7NC5IZI-a4sCKYApFPnPOAAQR6kJ4QzgjgJsXi31+qFF70+CW5VyMMBBMLhCKQUm16UwZDyAGEIBEuSB5gBWUYkFgAFRBbh5v3WxQAklAarBQmBIWYxX7QjBvC8lXAtFogA), the de-facto standard formatter, as well as other formatters like [deno fmt](https://www.peterbartha.com/deno-playground#code=M4dwlgLgxgFgBACgHYFcA2aCUcDeAoOOKAQ2AFM4BGALgMLgCMAnM4gawG46Ty4AmWoR4UAzNVx1CzVpzoBfPAqA). This PR changes the formatting to match that convention (partially reverting https://github.com/sublimehq/Packages/pull/3213).

In an ideal world, the rule would be written to say that the line following a `case:`/`default:` should be indented only if it's not itself another `case:` or `default:`. But I don't think that's feasible given how these rules appear to work, so instead this always indents. That works correctly at least for the (considerable) subset of users who always put the contents of `case`/`default` blocks within braces.

### DIFF
--- a/JavaScript/Indentation Rules.tmPreferences
+++ b/JavaScript/Indentation Rules.tmPreferences
@@ -12,10 +12,6 @@
 			(?:
 			# dedent closing braces
 			  \}
-			# detent `case ... :`
-			| case\b
-			# detent `default:`
-			| default\b
 			)
 		]]></string>
 
@@ -27,10 +23,6 @@
 			# indent after opening braces (may be followed by whitespace or comments)
 			# but exclude lines such as `extern "C" {`
 			  .* \{ (?: \s* /\*.*\*/ )* \s* (?: //.* )? $
-			# indent after `case ... :`
-			| case\b
-			# indent after `default:`
-			| default\b
 			)
 		]]></string>
 
--- a/JavaScript/tests/syntax_test_js_indent_common.js
+++ b/JavaScript/tests/syntax_test_js_indent_common.js
@@ -548,52 +548,52 @@ function testIfElseIndentationWithBracesAndComment(v) {
 
 function testSwitchCaseIndentation(v) {
     switch (s) {
-    case
-    case:
-    case break
-    case: break
-    case "(": break
-    case ")": break;
-    case ":": break;
-    case ";": break;
-    case
+        case
+        case:
+        case break
+        case: break
+        case "(": break
+        case ")": break;
+        case ":": break;
+        case ";": break;
+        case
         break;
-    case:
+        case:
         break;
-    case ":"
+        case ":"
         break;
-    case ':':
+        case ':':
         break;
-    case NestedIfStatement:
+        case NestedIfStatement:
         if (s.endsWith() = '(')
             return;
         break;
-    case NestedSwitchCase:
+        case NestedSwitchCase:
         switch (v) {
-        case 0: break;
-        case 2:
+            case 0: break;
+            case 2:
             break;
         }
         break;
-    case NestedSwitchCaseBlock:
+        case NestedSwitchCaseBlock:
         {
             switch (v) {
-            case 0: break;
-            case 2:
+                case 0: break;
+                case 2:
                 break;
             }
             break;
         }
-    case NestedSwitchCaseBlock2:
+        case NestedSwitchCaseBlock2:
         {
             switch (v) {
-            case 0: break;
-            case 2:
+                case 0: break;
+                case 2:
                 break;
             }
         }
         break;
-    default:
+        default:
         break;
     }
 }


rchl:
Why is the code within the case "blocks" not indented? That doesn't seem right, even according to your prettier and deno links.

bakkot:
I don't think that's feasible given how these rules appear to work.

If you can figure out how to make that work without making the `case` labels themselves not-indented I'd be very happy to make that change, though!

rchl:
If we have to choose between those two options then I think that the current behavior makes more sense in general. But feel free to wait for others to chime in

bakkot:
The current behavior is wrong for approximately every JavaScript programmer, since everyone indents case labels. The behavior I'm suggesting is wrong only for people who don't wrap the contents of their `case` blocks on braces, which is a much smaller set of people. (Also, more minorly, the fix for such people is to hit tab after the `case:\n`, which is easier than hitting shift-tab after typing the word `case`, which is what's currently required.)

jwortmann:
I have only taken a quick look, but the current behavior seems to decrease the indentation of the `case` statements by 1 level, and increase the indentation for the following line by 1 level.

I assume the goal is to prevent the unindent of the very first `case` statement (which would then increase indentation of the whole block by 1 level, compared to current behavior).

Afaik there are two syntaxes (Python & Matlab) which currently have that ability, and the corresponding PRs are #3464 and #3466. You could try to adopt that same strategy for JavaScript. This likely needs some small adjustments to the syntax with special treatment for the scope of only the very first `case` statement`.

bakkot:
Ah, that looks like the right thing! Unfortunately, looking at those PRs, I don't think I'm going to be able to figure out how to port that logic.

deathaxe (changes requested):
> The behavior I'm suggesting is wrong only for people who don't wrap the contents of their case blocks in braces, which is a much smaller set of people.

I kindly disagree with this assumption.

This kind of behavior was the exact reason for https://github.com/sublimehq/Packages/pull/3464 and thus we shouldn't introduce such a regression here.

bakkot:
I am confident in that assumption when it comes to JavaScript. (I have seen a _lot_ of JS code.) So I do think this PR would be an improvement. To be more precise, I think https://github.com/sublimehq/Packages/pull/3213 was a regression with regards to the `case` behavior (a regression I only noticed when updating Sublime recently), and this PR fixes that regression.

But it sounds like the technique in #3464 would be better still. I don't understand the internals here well enough to port that to the JS syntax rules, unfortunately.

rchl:
> So I do think this PR would be an improvement.

I'm absolutely convinced that you are not correct and this is not an improvement at all. :)

With your approach here, the user has to constantly fight with indentation when adding new cases or entering case blocks. It will neither indent code inside case blocks nor it will de-dent `case` statements itself. I have to both manually indent when starting to write the code inside a block and then manually de-dent when adding new case and then manually indent the block again.

https://user-images.githubusercontent.com/153197/217045307-75cf2173-565f-4ea1-a0ec-00ca209cdb30.mov

With current code the user only needs to correct indentation once on writing the first case and then it works as expected:

https://user-images.githubusercontent.com/153197/217045788-59b8ead6-cf96-4e80-a839-c7a3aadbdee2.mov

bakkot:
@rchl Your description is only accurate if you don't put braces around your case blocks. If you do put braces, it all works exactly correctly, all of the time, instead of needing to be fixed every time you write a `switch`. That's how it worked in older versions, and it's the behavior I expected when upgrading; this was a regression in the experience.

https://user-images.githubusercontent.com/1653598/217050666-0b9ac321-16bb-4557-8c9e-7607b17c2fd8.mov


And as I've said, I think a very large majority of JS programmers put braces around all their case blocks, so I think this behavior was a regression for most JS programmers. Braces are actually encouraged by the language - the following is a syntax error.

```js
switch (expr) {
  case 0:
    let x = something();
    // etc
    break;
  case 1:
    let x = somethingElse();
    // etc
    break;
}
```
This is because the `case` blocks share a scope, so the above code is a redeclaration of `x`, which is an error.

Anyway, if someone is up for implementing #3464, it sounds like that would make everyone happy, so we don't need to settle this particular question.

deathaxe:
We can shorten that discussion by the statement, which is true for all syntaxes:

Syntaxes often support several styles of indentation and we can't implement them all at the same time, nor is ther a possibility to quickly switch between them. Furthermore ST's indentation engine comes along with various limitations, which make it hard or even impossible to implement them all accurately.

The one way or the other, each solution has its edge cases.

I've created a PoC for special treatment of first case statement. It comes along with the edge case of failing to re-indent a whole document as we'd need to dedent by two to get the finally closing brace to correct level.

```
switch (a) {
  case 1:
    break;
}  <-- needs to dedent by 2
```


We finally end up in:


```
switch (a) {
  case 1:
    break;
  }  <-- needs to dedent by 2
  switch (a) {
    case 1:
      break;
    }  <-- needs to dedent by 2
    ...
```

... without a technical solution to fix it.

rchl:
> And as I've said, I think a very large majority of JS programmers put braces around all their case blocks

That's subjective and I don't agree with that. Without any data to back this up, I can make a statement that most don't use braces unless it's necessary (due to having to limit the scope as you've shown) and it's most of the time not necessary.

deathaxe:
We can come back to this topic in future, once ST might have more sophisticated indentation engine.

PR #3665, opened 2023-02-12
[TypeScript] Implement satisfies operator

Thom1729:
Implement the `satisfies` operator from TypeScript 4.9.

This implementation does allow `x satisfies const`. I don't think that's valid, but there's no spec, so who's to say? (My usual answer is “AST Explorer”, but it doesn't seem to have updated to 4.9 yet.) Either way, it shouldn't do any harm.


rchl:
You can use `LSP-typescript`:

Screenshot 2023-02-12 at 14 43 46

deathaxe (approved):

FichteFoll (approved):

PR #3670, opened 2023-03-02
Syntax higlight for .CJS files as JAVASCRIPT

bf:
.cjs files are CommonJS files and should be syntax highlighted as Javascript. The .cjs extension is being used to discern between .mjs (ES6 modules) and .cjs (commonJS) files. The traditional .js doesn't give this information.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -8,6 +8,7 @@ version: 2
 file_extensions:
   - js
   - mjs
+  - cjs
   - htc
 
 first_line_match: |-


bf:
Further info on .CJS extension:
- https://nodejs.org/docs/latest/api/modules.html#enabling
- https://www.typescriptlang.org/docs/handbook/esm-node.html
- https://fileinfo.com/extension/cjs

jfcherng (approved):

michaelblyons:
Does not appear to conflict with Clojure (`.clj`, `.cljc`) or ClojureScript (`.cljs`), which I had heard of but do not use. 👍

deathaxe (approved):

PR #3672, opened 2023-03-04
[Go] improve control flow scopes

### LINKED ISSUE #3671: [Go] Type switch .(type) is no longer tagged as a keyword
### What happened?

The `type` keyword in a Go type switch is no longer tagged more finely than `meta.block.go` (it was previously tagged as `keyword.declaration.type.go` which is also probably wrong).

````
switch v := v.(type) {
case ...
```
In this case the `type` keyword is tagged as `source.go` `meta.block.go`, but should include an indication that it is a keyword.

It's not entirely clear to me where it would fit in the scope taxonomy, but something like `keyword.operator.type.go` or `keyword.assertion.type.go` would probably work.

keith-hall:
Fixes #3671

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -783,7 +783,7 @@ contexts:
 
   match-keyword-continue:
     - match: \bcontinue\b
-      scope: keyword.control.flow.break.go
+      scope: keyword.control.flow.continue.go
 
   match-keyword-default:
     - match: \b(default)\b(:)?
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -719,11 +719,20 @@ contexts:
 
   match-keyword-break:
     - match: \bbreak\b
-      scope: keyword.control.go
+      scope: keyword.control.flow.break.go
 
   match-keyword-case:
     - match: \bcase\b
-      scope: keyword.control.go
+      scope: keyword.control.switch.case.go
+      push: match-case-value
+
+  match-case-value:
+    - match: '{{predeclared_type}}'
+      scope: support.type.go
+    - match: ':'
+      scope: punctuation.separator.case-statement.go
+      pop: true
+    - include: match-any
 
   match-keyword-chan:
     - match: (?=\bchan\b)
@@ -774,11 +783,13 @@ contexts:
 
   match-keyword-continue:
     - match: \bcontinue\b
-      scope: keyword.control.go
+      scope: keyword.control.flow.break.go
 
   match-keyword-default:
-    - match: \bdefault\b
-      scope: keyword.control.go
+    - match: \b(default)\b(:)?
+      captures:
+        1: keyword.control.switch.case.go
+        2: punctuation.separator.case-statement.go
 
   match-keyword-defer:
     - match: \bdefer\b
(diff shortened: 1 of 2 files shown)


Go/Go.sublime-syntax
```diff
@@ -719,11 +719,20 @@ contexts:
 
   match-keyword-break:
     - match: \bbreak\b
-      scope: keyword.control.go
+      scope: keyword.control.flow.break.go
 
   match-keyword-case:
     - match: \bcase\b
-      scope: keyword.control.go
+      scope: keyword.control.switch.case.go
```
deathaxe, line 726:
```suggestion
      scope: keyword.control.conditional.case.go
```

According to other syntaxes in this repo `switch..case..default` scopes are organized within `keyword.control.conditional`.

Go/Go.sublime-syntax
```diff
@@ -774,11 +783,13 @@ contexts:
...
     - match: \bcontinue\b
-      scope: keyword.control.go
+      scope: keyword.control.flow.break.go
 
   match-keyword-default:
-    - match: \bdefault\b
-      scope: keyword.control.go
+    - match: \b(default)\b(:)?
+      captures:
+        1: keyword.control.switch.case.go
```
deathaxe, line 791:
```suggestion
        1: keyword.control.conditional.default.go
```

I guess we could argue about using 

1. `keyword.control.conditional.case` 
2. `keyword.control.conditional.switch.case`
3. `keyword.control.conditional.switch.default`

as alternative scopes, but other syntaxes just use the plain `default`. IIRC that was because of different possible statement structures, some of them not using `switch` keyword.

Go/Go.sublime-syntax
```diff
@@ -866,7 +877,7 @@ contexts:
 
   match-keyword-switch:
     - match: \bswitch\b
-      scope: keyword.control.go
+      scope: keyword.control.flow.switch.go
```
deathaxe, line 880:
```suggestion
      scope: keyword.control.conditional.switch.go
```

Go/Go.sublime-syntax
```diff
@@ -719,11 +719,20 @@ contexts:
...
+      scope: keyword.control.switch.case.go
+      push: match-case-value
+
+  match-case-value:
+    - match: '{{predeclared_type}}'
+      scope: support.type.go
+    - match: ':'
+      scope: punctuation.separator.case-statement.go
+      pop: true
+    - include: match-any
```
deathaxe, line 735:
```suggestion
    - match: ':'
      scope: punctuation.separator.case-statement.go
      pop: true
    - match: '{{predeclared_type}}'
      scope: support.type.go
    - include: match-any
```

Maybe just a personal habit, but I tend to organize popping patterns at the top of a context.

deathaxe (commented):
Would suggest some scope changes to align with other syntaxes such as JS, Java or Perl.

Go/Go.sublime-syntax
FichteFoll: I just went through https://github.com/sublimehq/Packages/issues/1228 again and the state this (and other RFC) discussions are in is just a little said with no definitive conclusion available and the information spread throughout the discussion. I wish I had more time at hand to get back to SNG and properly compile the results of these discussions into an encyclopedia you can look up …

Go/Go.sublime-syntax
```diff
@@ -774,11 +783,13 @@ contexts:
 
   match-keyword-continue:
     - match: \bcontinue\b
-      scope: keyword.control.go
+      scope: keyword.control.flow.break.go
```
FichteFoll, line 786:
```suggestion
      scope: keyword.control.flow.continue.go
```

FichteFoll (changes requested):

FichteFoll (approved):

deathaxe (approved):

PR #3686, opened 2023-03-25
[JSON] Tweak folding rules

deathaxe:
This commit makes closing brackets to directly follow a folded region instead of keeping them at the beginning of the next line.

The explicit rule is required as default behavior of ST has recently changed.


keith-hall (approved):

michaelblyons (approved):

deathaxe:
Thanks!

PR #3697, opened 2023-04-15
[CSS] Exclude trailing newline from folding regions

deathaxe:
This commit makes closing brackets to directly follow a folded region instead of keeping them at the beginning of the next line.

The explicit rule is required as default behavior of ST has recently changed.

see also: #3686

### DIFF
--- a/CSS/Fold.tmPreferences
+++ b/CSS/Fold.tmPreferences
@@ -12,6 +12,8 @@
                 <string>punctuation.section.block.begin</string>
                 <key>end</key>
                 <string>punctuation.section.block.end</string>
+                <key>excludeTrailingNewlines</key>
+                <false/>
             </dict>
         </array>
     </dict>


keith-hall (approved):

michaelblyons:
Sorry for my ignorance with folding behavior. Does your commit change this (when folded):

```css
.foo {•••
}
```
to this:
```css
.foo {•••}
```

michaelblyons (approved):
Cool. Sounds like a thing people would want

deathaxe:
Yep, it feels better with all expressions which do not need alternations such as `if ... { ... } else { ... }`.

keith-hall:
This is setting exclude trailing newline to false, so shouldn't the PR title have been "include trailing newline in folding regions"?

deathaxe:
Well, you are right. I had only the setting name in mind when writing the title.

PR #3700, opened 2023-04-16
[CSS] fix color functions

deathaxe:
This PR...

1. drops `gray()` and `device-cmyk()` functions
2. fixes highlighting angle values in `lch()` functions
3. fixes angle values causing scalar floating values beginning with `0` from not being scoped correctly.

Note: Some of the color-adjuster-functions are also not supported by browsers, but kept in place as some are used by PackageDev.


michaelblyons (approved):

jfcherng (approved):

FichteFoll:
> Note: Some of the color-adjuster-functions are also not supported by browsers, but kept in place as some are used by PackageDev.

PD should not be an argument for keeping old and never used functions in the default CSS syntax. PD can instead extend the CSS syntax with the functions ST uses internally.

deathaxe:
Sure. It should however do so, first. I am currently working on refactoring CSS to make it a good base for being extended by PD, LESS, SCSS, ... .

PR #3713, opened 2023-04-28
[C++] Add std::begin($1), std::end($1) snippet

nabijaczleweli:
Ideally I'd also love to kill the weird current `$1.begin(), $1.end()` which definitionally doesn't really work, but I think I can just do that locally maybe.


deathaxe:
Snippets in general seem to be a rather personal topic. Ideally we could migrate all of them to external packages so they could be installed on demand. On the other hand various users may have gotten used to them being present by default. Hence removing anything is rather problematic, maybe.

As this snippet is already provided by [C++ Snippets](https://packagecontrol.io/packages/C%2B%2B%20Snippets) (see: https://github.com/Rapptz/cpp-sublime-snippet/blob/master/utilities/beginend.sublime-snippet) we probably don't need to add it here.

nabijaczleweli:
Then probably also axe the `$1.begin(), $1.end()` one which is just like this one but is fundamentally broken IMO

deathaxe:
If there are snippets which don't make sense at all anymore, yes. I haven't looked into any snippet for ages.

PR #3717, opened 2023-04-29
[CSS] Make CSS a good base syntax for extensions

deathaxe:
This PR proposes structural changes to CSS sublime syntax inspired by HTML, to make it a good base for 3rd-party syntaxes such as LESS and SCSS.

**Changes:**

1. Convert anonymous into named contexts
2. Strategic reorganization of contexts and includes
3. rename contexts inspired by HTML/Python 

   ```
   - <context>
     - <context>-body
     - <context>-end
     - <context>-content
   ```

4. adds some syntax test examples to illustrate how these changes improve interpolation support in ASP/JSP/PHP.

**Notes**

- It intents to change CSS highlighting behevior as little as possible at this point.

- The overall parsing performance is increased by 20-25%.

- Here's a [LESS](https://github.com/deathaxe/LESS/tree/st4) prototype, which makes use of this PR.

### DIFF
--- a/ASP/Embeddings/CSS (for ASP).sublime-syntax
+++ b/ASP/Embeddings/CSS (for ASP).sublime-syntax
@@ -21,6 +21,10 @@ hidden: true
 
 extends: Packages/CSS/CSS.sublime-syntax
 
+variables:
+
+    ident_start: (?:{{nmstart}}|<%)
+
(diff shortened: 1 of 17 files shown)


michaelblyons (approved):
Nice!!

PR #3718, opened 2023-05-01
[Git Formats] Support `.git-blame-ignore-revs` files

mataha:
Git Blame Ignore Revisions file consists of a list of object names (i.e. one unabbreviated SHA-1 per line) that should be excluded from the blame view using git's `git blame --ignore-revs-file <file>` command or, on GitHub, by committing a `.git-blame-ignore-revs` file to the root directory of a repository. Versions of git 2.20 and higher permit comments (`#`), empty lines and trailing/leading whitespace; older versions error on everything but a SHA-1 per line.

I believe the syntax of these files is simple enough to warrant including support for them in the default package without placing much (to be honest, any) potential maintenance burden.

Included syntax tests take an average of 0.3ms over 10 runs here.

### DIFF
--- a/Git Formats/Git Blame Ignore Revisions.sublime-syntax
+++ b/Git Formats/Git Blame Ignore Revisions.sublime-syntax
@@ -16,7 +16,7 @@ hidden_file_extensions:
   - .git-blame-ignore-revs    # ~/.git-blame-ignore-revs
 
 variables:
-  comment_char: '[#]'         # Can't use Git Common here - ';' is not allowed
+  comment_char: \#            # Can't use Git Common here - ';' is not allowed
   object_name: \b\h{40}\b
 
 contexts:
--- a/Git Formats/Git Blame Ignore Revisions.sublime-syntax
+++ b/Git Formats/Git Blame Ignore Revisions.sublime-syntax
@@ -29,6 +29,7 @@ contexts:
       push: comment-body
 
   comment-body:
+    - meta_include_prototype: false
     - meta_scope: comment.line.git.blame.ignorerevs
     - match: \n
       pop: 1
--- a/Git Formats/Git Blame Ignore Revisions.sublime-syntax
+++ b/Git Formats/Git Blame Ignore Revisions.sublime-syntax
@@ -22,7 +22,7 @@ contexts:
 
   main:
     - include: comments
-    - include: revision
+    - include: revisions
     - include: illegal
 
   comments:
@@ -36,12 +36,12 @@ contexts:
     - match: \n
       pop: 1
 
-  revision:
+  revisions:
     - match: '{{object_name}}'
       scope: constant.other.hash.git.blame.ignorerevs
-      push: rest
+      push: expect-eol
 
-  rest:
+  expect-eol:
     - include: comments
     - include: illegal
     - include: eol-pop
--- a/Git Formats/Git Blame Ignore Revisions.sublime-syntax
+++ b/Git Formats/Git Blame Ignore Revisions.sublime-syntax
@@ -21,17 +21,21 @@ variables:
 
 contexts:
   prototype:
-    - include: comments
+    - include: comment-start
+
+  comment-start:
+    - match: '[{{comment_char}}]'
+      scope: punctuation.definition.comment.git.blame.ignorerevs
+      push: comment-content
+
+  comment-content:
+    - meta_scope: comment.line.git.blame.ignorerevs
+    - match: \n
+      pop: 1
 
   main:
     - include: revision
     - include: illegal
-
-  comments:
-    - match: ({{comment_char}}).*\n?
-      scope: comment.line.git.blame.ignorerevs
-      captures:
-        1: punctuation.definition.comment.git.blame.ignorerevs
 
   revision:
     - match: '{{object_name}}'
--- a/Git Formats/Git Blame Ignore Revisions.sublime-syntax
+++ b/Git Formats/Git Blame Ignore Revisions.sublime-syntax
@@ -16,15 +16,17 @@ hidden_file_extensions:
   - .git-blame-ignore-revs    # ~/.git-blame-ignore-revs
 
 variables:
-  comment_char: \#            # Can't use Git Common here - ';' is not allowed
   object_name: \b\h{40}\b
 
 contexts:
-  prototype:
+
+  main:
     - include: comments
+    - include: revision
+    - include: illegal
 
   comments:
-    - match: '[{{comment_char}}]'
+    - match: \#+
       scope: punctuation.definition.comment.git.blame.ignorerevs
       push: comment-body
 
@@ -34,21 +36,18 @@ contexts:
     - match: \n
       pop: 1
 
-  main:
-    - include: revision
-    - include: illegal
-
   revision:
     - match: '{{object_name}}'
       scope: constant.other.hash.git.blame.ignorerevs
       push: rest
 
   rest:
+    - include: comments
     - include: illegal
     - include: eol-pop
 
   illegal:
-    - match: '[^\s{{comment_char}}]+'
+    - match: '[^\s#]+'
       scope: invalid.illegal.stray-character.git.blame.ignorerevs
 
   eol-pop:
--- a/Git Formats/Git Blame Ignore Revisions.sublime-syntax
+++ b/Git Formats/Git Blame Ignore Revisions.sublime-syntax
@@ -32,7 +32,6 @@ contexts:
       scope: comment.line.git.blame.ignorerevs
       captures:
         1: punctuation.definition.comment.git.blame.ignorerevs
-    - include: eol-pop
 
   revision:
     - match: '{{object_name}}'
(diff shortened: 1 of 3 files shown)


deathaxe (changes requested):
Thanks for contribution. I think we can add that syntax to this repo. Would suggest some changes however.

The comments below would result in the following context structure.

```yaml
contexts:
  prototype:
    - include: Git Common.sublime-syntax#comments

  main:
    - include: hash
    - include: illegal

  hash:
    - match: '{{revision}}'
      scope: constant.other.hash.git.blame.ignorerevs
      push: rest

  rest:
    - include: illegal
    - include: eol-pop

  illegal:
    - match: '[^\s#]+'
      scope: invalid.illegal.stray-character.git.blame.ignorerevs

  eol-pop:
    - match: $
      pop: 1
```

The overall performance would be 2x faster than the current approach due to changed `illegal` pattern.

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -0,0 +1,41 @@
...
+      scope: constant.other.hash.git.blame.ignorerevs
+      push: rest
+
+  rest:
+    - match: \S
+      scope: invalid.illegal.stray-character.git.blame.ignorerevs
+    - include: end
+
+  end:
+    - match: $\n?
```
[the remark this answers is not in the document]
mataha: Matching a single `$` now.

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -0,0 +1,41 @@
...
+variables:
+  revision: \b\h{40}\b
+
+contexts:
+  prototype:
+    - include: Git Common.sublime-syntax#comments
+
+  main:
+    - include: hash
+    - include: rest
```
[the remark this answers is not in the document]
mataha: Done.

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -0,0 +1,52 @@
...
+  main:
+    - include: revision
+    - include: illegal
+
+  comments:
+    - match: ({{comment_char}}).*
+      scope: comment.line.git.blame.ignorerevs
+      captures:
+        1: punctuation.definition.comment.git.blame.ignorerevs
+    - include: eol-pop
```
deathaxe, line 35:
```suggestion
```

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -0,0 +1,51 @@
...
+contexts:
+  prototype:
+    - include: comments
+
+  main:
+    - include: revision
+    - include: illegal
+
+  comments:
+    - match: ({{comment_char}}).*
```
deathaxe, line 31: We want to scope the `\n` in comments, if present as some plugins rely on it. Maybe uncommon for this file, but generally applied to all syntaxes.

```suggestion
    - match: ({{comment_char}}).*\n?
```

keith-hall: Generally it's better to match the comment start token, then push into a named context for the comment contents, which will pop after consuming the `\n` char. Why?
Because it helps with both embedding and extending.

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -0,0 +1,41 @@
...
+    - include: hash
+    - include: rest
+
+  hash:
+    - match: '{{revision}}'
+      scope: constant.other.hash.git.blame.ignorerevs
+      push: rest
+
+  rest:
+    - match: \S
```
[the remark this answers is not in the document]
mataha: No. `git config` [always uses `#` or `;`](https://git-scm.com/docs/git-config#EXAMPLES), while `fsck.skipList` [permits `#` only](https://git-scm.com/docs/git-fsck#Documentation/git-fsck.txt-fsckskipList).

See [the source](https://github.com/git/git/blob/master/oidset.c#L60-L70):
```c
/*
 * Allow trailing comments, leading whitespace
 * (including before commits), and empty or whitespace
 * only lines.
*/
name = strchr(sb.buf, '#');
if (name)
    strbuf_setlen(&sb, name - sb.buf);
strbuf_trim(&sb);
if (!sb.len)
    continue;
```

Git Formats/Git Blame Ignore Revisions.sublime-syntax
deathaxe: Agree. Wouldn't have argued for it though as this pattern was copied from _Git Common.sublime-syntax_, most patterns in _Git Commit_ etc. also use blump capture groups only and - am I wrong? - those files seem rather uncommon to be candidates for extending them with interpolation.

Changing that would be welcome, however.

mataha: Something like this?

```yaml
  comment-start:
    - match: '[{{comment_char}}]'
      scope: punctuation.definition.comment.git.blame.ignorerevs
      push: comment-content

  comment-content:
    - meta_scope: comment.line.git.blame.ignorerevs
    - match: .*\n?
      pop: 1
```

michaelblyons:
```yaml
comment-start:
  - match: '{{comment_char}}'
    scope: punctuation....
    push: ...

comment-content:
  - meta_scope: comment...
  - match: $\n?
    pop: 1
```

Pardon any typos or shortening. I'm on a phone.

michaelblyons: The point is explicitly to avoid the `.*` in the match.

michaelblyons: IIRC the `$` is relevant only on EOF, but at that point the pop is not really important. 😉

mataha: I have made changes. Question: wouldn't set be better than push here for overwriting the `rest` context if applicable?

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -21,18 +21,22 @@ variables:
 
 contexts:
   prototype:
-    - include: comments
+    - include: comment-start
+
+  comment-start:
+    - match: '[{{comment_char}}]'
```
michaelblyons, line 27: Injection here becomes `[[#]]`, and I don't think you want that.

michaelblyons (changes requested):

Git Formats/Git Blame Ignore Revisions.sublime-syntax
deathaxe:
```suggestion
    - match: \#
```

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -21,18 +21,22 @@ variables:
 
 contexts:
   prototype:
-    - include: comments
+    - include: comment-start
+
+  comment-start:
```
deathaxe, line 26:
```suggestion
    - include: comments

  comments:
```

Git Formats/Git Blame Ignore Revisions.sublime-syntax
deathaxe: No, because `set` would replace `main` with `rest` on stack.

mataha: Nitpicking, but ST has guards that don't allow `main` to be replaced and instead push a new context on the stack, right? (That wouldn't apply if this syntax were to be extended, of course.)

Git Formats/Git Blame Ignore Revisions.sublime-syntax
mataha: Should I just copy Erlang at this point? It captures all comment characters though:

```yaml
  comments:
    - match: \%+
      scope: punctuation.definition.comment.erlang
      push: comment-body

  comment-body:
    - meta_scope: comment.line.percentage.erlang
    - match: \n
      pop: 1
```

Git Formats/Git Blame Ignore Revisions.sublime-syntax
deathaxe: It has some safe guards, but still setting `rest` onto stack is wrong - nothing else is of interest. It would cause trouble as soon as this syntax is included in another one via `- include: scope:text.git...`

Git Formats/Git Blame Ignore Revisions.sublime-syntax
deathaxe: I've pushed relevant changes to maintain a somewhat useful balance between code changes and effort for descriptions and discussions.

deathaxe (approved):

Git Formats/Git Blame Ignore Revisions.sublime-syntax
```diff
@@ -0,0 +1,55 @@
...
+
+contexts:
+
+  main:
+    - include: comments
+    - include: revisions
+    - include: illegal
+
+  comments:
+    - match: \#+
```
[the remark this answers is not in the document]
deathaxe: It looks better if punctuation is highlighted differently.

grafik

michaelblyons (approved):

PR #3719, opened 2023-05-01
[HTML] Sync known html tag names with CSS

deathaxe:
This PR proposes to apply tag name variables from CSS.sublime-syntax to HTML.sublime-syntax to get both syntaxes inline. Some html5 tags such as `del`, `ins`, `slot` etc. have been missing.

Variables are grouped into `html_block_tags` and `html_inline_tags` to maintain current scoping (`meta.tag.[structure|block|inline]`) of HTML.

> The main purpose of those variables is to better keep track of possible changes on MDN.

A dedicated named context is created for each kind of tag content in case an inheriting syntax wants to replace or extend one of them.

Dedicated pattern for `<a...>` tags is removed as it doesn't push into any special context nor is the scope used in completions or snippets.

#### CSS

Variable names and formatting is adjusted in CSS. Word boundary checks are removed as tag names appear in selectors only, which already maintain word boundaries of "unknown" tags/classes/ids.

#### Notes

1. Beyond `a` tag and some new html5 tags, syntax highlighting is not effected.
2. Parsing performance is unchanged.

### DIFF
--- a/HTML/HTML.sublime-syntax
+++ b/HTML/HTML.sublime-syntax
@@ -19,20 +19,88 @@ first_line_match: (?i)<(!DOCTYPE\s*)?html
 ###############################################################################
 
 variables:
-  block_tag_name: |-
-    (?ix:
-      address|applet|article|aside|blockquote|center|dd|dir|div|dl|dt|figcaption|figure|footer|frame|frameset|h1|h2|h3|h4|h5|h6|header|iframe|menu|nav|noframes|object|ol|p|pre|section|ul
-    ){{tag_name_break}}
+  # HTML tags
+  # https://developer.mozilla.org/en-US/docs/Web/HTML/Element
+  # Note: Sections are sorted by expected frequency.
 
-  inline_tag_name: |-
-    (?ix:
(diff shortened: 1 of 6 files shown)


HTML/HTML.sublime-syntax
```diff
@@ -19,20 +19,88 @@ first_line_match: (?i)<(!DOCTYPE\s*)?html
...
+    | sup | u | var | wbr ){{tag_name_break}}
+
+  html_media_tags: |-
+    (?xi: area | audio | img | map | track | video ){{tag_name_break}}
+
+  html_embedded_tags: |-
+    (?xi: embed | iframe | object | param | picture | source ){{tag_name_break}}
+
+  html_script_tags: |-
+    (?xi: canvas | noscript (?# | script ) ){{tag_name_break}}
```
keith-hall, line 77: it seems odd to see `script` tag commented out in a context called `html_script_tags`. Can you explain the reasoning, please?
Follow up question, why comment them out instead of just delete them if they are not needed?

deathaxe: All variables mainly exist to keep track of MDN changes more easily. The `script` tag is handled by a dedicated context `script-tag` and is listed twice on MDN - in header tags and script tags section. By commenting out, I wanted to make clear it was not forgotten/missed but disabled by intent.

keith-hall (approved):

michaelblyons (approved):

PR #3721, opened 2023-05-02
[Go] support language hints for backtick strings

keith-hall:
It's very common in Go code to see ["language injection comments" as popularized by JetBrains' GoLand](https://www.jetbrains.com/help/go/using-language-injections.html#use-language-injection-comments).

The idea is that a comment will give a language hint as to what type of code is embedded in the following backtick string, for syntax highlighting and completions purposes. i.e. `// language=sql`

This PR adds such support to Sublime, for embedded SQL, JSON and regular expressions. These are the main embedded languages I've come across.

### DIFF
--- a/Go/Go (Embedded Backtick String).sublime-syntax
+++ b/Go/Go (Embedded Backtick String).sublime-syntax
@@ -1,0 +1,63 @@
+%YAML 1.2
+---
+name: Go (Embedded Backtick String)
+scope: source.go.embedded-backtick-string
+version: 2
+extends: Packages/Go/Go.sublime-syntax
+hidden: true
+
+contexts:
+  # https://www.jetbrains.com/help/go/using-language-injections.html#use-language-injection-comments
+  inside-comment-magic-language-injection-language-after-language-identifier:
+    - meta_content_scope: comment.line.double-slash.go
+    # TODO: support prefix and suffix
+    - match: $\n?
+      set:
+        - main
+        - consume-comments-at-beginning-of-line
+
+  pop-at-bol:
+    - match: ^
+      pop: true
+
+  main:
+    - meta_prepend: true
+    - match: $\n?
+      pop: true
+
+  consume-comments-at-beginning-of-line:
+    - include: match-comments
+    - match: ^\s*(?=/[/*])
+    - include: pop-at-bol
+
+  match-raw-string:
+    - match: '`'
+      scope: punctuation.definition.string.begin.go
+      embed: embedded-language
+      escape: '`'
+      escape_captures:
+        0: punctuation.definition.string.end.go
+
+  embedded-language:
+    - meta_include_prototype: false
+
+  match-parens:
+    - meta_prepend: true
+    - match: (?=\))
+      pop: true
+
+  match-brackets:
+    - meta_prepend: true
+    - match: (?=\])
+      pop: true
+
+  match-braces:
+    - meta_prepend: true
+    - match: (?=\})
+      pop: true
+
+  match-backtick-string-formatting:
+    - include: match-template-string
+    - match: \%%
+      scope: constant.character.escape.go
+    - include: match-fmt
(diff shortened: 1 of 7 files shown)


Go/Go (Embedded Backtick String).sublime-syntax
```diff
@@ -0,0 +1,63 @@
...
+scope: source.go.embedded-backtick-string
+version: 2
+extends: Packages/Go/Go.sublime-syntax
+hidden: true
+
+contexts:
+  # https://www.jetbrains.com/help/go/using-language-injections.html#use-language-injection-comments
+  inside-comment-magic-language-injection-language-after-language-identifier:
+    - meta_content_scope: comment.line.double-slash.go
+    # TODO: support prefix and suffix
```
keith-hall, line 13: I've never actually seen these used, and it would likely be not trivial to implement, so I plan to leave this TODO comment here for now.

Go/SQL (Embedded Go Backtick String Formatting).sublime-syntax
keith-hall, line 1: I didn't apply this technique to the JSON and regex embeds (yet) because I thought it is perhaps not necessary. I'm not sure how many people would ever use string formatting when building a regular expression pattern in Go code, for example. It feels like we should apply YAGNI and only work on it if someone requests it as time is a finite resource :)

deathaxe:
Maybe we can organize all hidden syntaxes which are used for embedding purposes in `Embeddings` folder following the scheme of packages such as PHP?

deathaxe:
Proposed some changes at https://github.com/forkeith/Packages/pull/24

michaelblyons:
I think one of the regulars uses Go. Is it Terminal?

keith-hall:
> I think one of the regulars uses Go. Is it Terminal?

Yep: https://github.com/sublimehq/Packages/pull/3651#issuecomment-1436158151

@TerminalFi want to give this PR a try? :)

Go/Go (Embedded Backtick String).sublime-syntax
deathaxe: Reasonable.

TerminalFi:
> > I think one of the regulars uses Go. Is it Terminal?
> 
> Yep: [#3651 (comment)](https://github.com/sublimehq/Packages/pull/3651#issuecomment-1436158151)
> 
> @TerminalFi want to give this PR a try? :)

Yup! I’ll load it up today.

TerminalFi:
I do find it interesting that an unlimited amount of white spaces can prefix these 

```
// language=sql
// language = sql
// language      =          sql
```

are all valid. Intentional? Should a more strict latter be required ?

keith-hall:
> I do find it interesting that an unlimited amount of white spaces can prefix these
> 
> ```
> // language=sql
> // language = sql
> // language      =          sql
> ```
> 
> are all valid. Intentional? Should a more strict latter be required ?

If GoLand is strict about no whitespace then we should be too. I just didn't know because I have never used GoLand ;) but yeah, come to think of it, I don't think I have ever seen spaces around the equals sign in the codebase I have worked on.

TerminalFi:
> > I do find it interesting that an unlimited amount of white spaces can prefix these
> > ```
> > // language=sql
> > // language = sql
> > // language      =          sql
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > are all valid. Intentional? Should a more strict latter be required ?
> 
> If GoLand is strict about no whitespace then we should be too. I just didn't know because I have never used GoLand ;) but yeah, come to think of it, I don't think I have ever seen spaces around the equals sign in the codebase I have worked on.

I’ve never used golang either. I’ll test it out today though and see whether or not it’s stricr

TerminalFi:
@keith-hall Testing in Goland, for automatic detection it requires no spaces on either side. So `// language=html` is the expected value. Otherwise it requires you to enable it via a popup menu and you force the syntax. Unrelated to the comment

deathaxe (approved):

michaelblyons (approved):

PR #3722, opened 2023-05-04
[Batch File] Improve batch file matching

mataha:
This PR aims to improve batch file (i.e. `cmd.exe` script) matching by:

  - adding `.bat` and `.cmd` to the list of valid file extensions (dotfile scripts e.g. `%USERPROFILE%\.{bat,cmd}` invoked via `{HKLM,HKCU}\Software\Microsoft\Command Processor\AutoRun`);
  - enhancing `first_line_match` in order for it to match most of the common first line patterns in batch files, specifically "ifdebugecho" (`@if "%DEBUG%"=="" @echo off`) and its variants.

Catches all of the following:

  - `@echo off`
  - `@if "%DEBUG%"=="" echo off`
  - `@if "%DEBUG_BATCH_ECHO:"=%"=="on" (echo on)`
  - `if not defined DEBUG @echo off`
  - `if 0==1: echo =r"""...`
  - `if "%~n0"=="" (@echo off) else (@goto :EOF)`

There are more, but the goal is not to be perfect.

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -9,10 +9,14 @@ file_extensions:
   - bat
   - cmd
 
+hidden_file_extensions:
+  - .bat
+  - .cmd
+
 first_line_match: |-
   (?xi:
-    ^ \s* (?:\:|rem) .*? -\*- .*? \bbat\b .*? -\*-  # editorconfig
-  | ^ \s* @echo\s+(?:on|off)\b
+    ^ [@\s]* (?:\:|rem) .*? -\*- .*? \b(?:bat|cmd)\b .*? -\*-  # editorconfig
+  | ^ [@\s]* (?:if \s+ \S .*? =?=? .*? \S \s+)? [@(\s]* echo \s+ (?:on|off|=) \b
   )
 
 ###############################################################################


FichteFoll (dismissed):
Sounds reasonable to me.

Batch File/Batch File.sublime-syntax
```diff
@@ -9,10 +9,14 @@ file_extensions:
...
+hidden_file_extensions:
+  - .bat
+  - .cmd
+
 first_line_match: |-
   (?xi:
-    ^ \s* (?:\:|rem) .*? -\*- .*? \bbat\b .*? -\*-  # editorconfig
-  | ^ \s* @echo\s+(?:on|off)\b
+    ^ [@\s]* (?:\:|rem) .*? -\*- .*? \b(?:bat|cmd)\b .*? -\*-  # editorconfig
+  | ^ [@\s]* (?:if \s+ \S .*? =?=? .*? \S \s+)? [@(\s]* echo \s+ (?:on|off|=) \b
```
deathaxe, line 19: Not only `if a == b` but also `if 1 NEQ 0` are valid conditional expressions.

Your current pattern does match both, but that's not obvious. The `if ...` part may also suffer from catastrophic backtracking.

I'd therefore suggest the following simplification.

```suggestion
  | ^ [@\s]* (?:if \s .*? [@(\s] )? echo \s+ (?:on|off|=) \b
```

If the line starts with an `if` keyword, consume anything until `@echo`, `(echo`  or ` echo` are found. This should allow any kind of conditional expression.

Without `if` a line must start with `@echo` or ` echo` or ` @ @ @ echo`.

deathaxe (commented):
Thanks for this PR.

It improves first line detection, even though I'd like to suggest a tiny change/optimization.

It also reveals a syntax highlighting short comming of current syntax definition with regards to `@` operator, which we might want to look into in another PR.

Batch File/Batch File.sublime-syntax
mataha: Sounds good (and then `.*?` takes care of complex expressions anyway).

mataha: Done.

deathaxe (approved):

michaelblyons (approved):

PR #3740, opened 2023-05-12
[Markdown] Add language separator for fenced code blocks

deathaxe:
This commit adds `:` as valid language name terminator.

The following code block

~~~md
```jsx:file.jsx
import Router from 'next/router'

function MyComponent() {
	const [show, setShow] = useState(false)

	useEffect(() => {
    console.log(2)
	}, [])

	return <>...</>
}
~~~

renders as:

```jsx:file.jsx
import Router from 'next/router'

function MyComponent() {
	const [show, setShow] = useState(false)

	useEffect(() => {
    console.log(2)
	}, [])

	return <>...</>
}
```

see: https://github.com/withastro/astro/blob/main/examples/with-markdown-plugins/src/pages/index.md

Note: CommonMark writes the "first word" is used as language identifier, so there may be more valid separator chars.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -141,7 +141,7 @@ variables:
     (?x:             # first word of an infostring is used as language specifier
       (
         [[:alpha:]]  # starts with a letter to make sure not to hit any attribute annotation
-        [^`\s]*      # optionally followed by any nonwhitespace character (except backticks)
+        [^\s:;`]*    # optionally followed by any nonwhitespace character (except backticks)
       )
     )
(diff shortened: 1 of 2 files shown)


keith-hall (dismissed):

michaelblyons (approved):

jwortmann:
Could you add `;` too?
For example: https://github.com/JuliaLang/julia/blob/c6fc12c9eee6aa4fb1f584840e5a0f988009d82f/doc/src/manual/control-flow.md?plain=1#L127

deathaxe:
Done! Actually revealed another edge case with unknown language ids, which is also fixed.

PR #3741, opened 2023-05-13
[Java] Rename class scopes

deathaxe:
This PR proposes to change scope of classes to `support.class`.

Java decided to use `storage.type.class` as `support` scopes are said to be used for types provided by (standard) libraries only. Thus `storage.type` would be the scope for all user defined data types.

The truth according to TextMate's is exactly vice versa.

`storage.type` is meant for basic builtin data types of a language, such as `int`, `char`, ...

`support.class`, `support.type`, ... however are meant for all kinds of custom or user defined data types - not only those from standard libraries.

As such it makes very much sense most (all) other syntaxes use `support.class` for `ClassTypes`.

This commit intents to make Java follow those rules, too, again.

### DIFF
--- a/Java/Java.sublime-syntax
+++ b/Java/Java.sublime-syntax
@@ -1966,7 +1966,7 @@ contexts:
 
   receiver-unqualified-type:
     - match: '{{identifier}}'
-      scope: storage.type.class.java
+      scope: support.class.java
       push:
         - receiver-unqualified-type-end
         - maybe-array-modifiers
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

FichteFoll (approved):
I did find it a little weird that you went this road with Java but I was find trying it out for a while. In the end, both are fine for me, but since Java is in the overwhelming minority, it makes sense to align Java with the others for now until we maybe find another solution do this (see #1842).

PR #3746, opened 2023-05-15
[CSS] Improve Completions

deathaxe:
- Fix `url()` reference in property values
- Add `backdrop-filter` values


keith-hall (approved):
Looks good to me. For those wondering about the `<uri>` -> `<url>` change, it is looked up and `url` has completions at https://github.com/sublimehq/Packages/blob/719209b7a9405e71c77dc84748242924b07c66cb/CSS/completions/common.py#L495

michaelblyons (approved):

PR #3750, opened 2023-05-15
[Batch File] Fix interpretation of numeric handles

mataha:
An optional numeric handle consists of **exactly one** number between 0 and 9.

In action:

```batchfile
~\Packages$ rem `set 12 ^>nul`, not `set 1 2^>nul`
~\Packages$ set 12>nul
Environment variable 12 not defined
```

Unofficial source for reference: https://ss64.com/nt/syntax-redirection.html

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -1540,12 +1540,12 @@ contexts:
 ###[ PATH PATTERNS ]##########################################################
 
   redirection-interpolations:
-    - match: \s*(?=\d*(?:<&?|>[&>]?))
+    - match: \s*(?=(?:\d)?(?:<&?|>[&>]?))
       push: redirection-interpolations-body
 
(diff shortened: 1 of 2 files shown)


Batch File/Batch File.sublime-syntax
```diff
@@ -1540,12 +1540,12 @@ contexts:
 ###[ PATH PATTERNS ]##########################################################
 
   redirection-interpolations:
-    - match: \s*(?=\d*(?:<&?|>[&>]?))
+    - match: \s*(?=(?:\d)?(?:<&?|>[&>]?))
```
deathaxe, line 1543: The group could be ommited.

```suggestion
    - match: \s*(?=\d?(?:<&?|>[&>]?))
```

deathaxe (approved):

FichteFoll (approved):

PR #3753, opened 2023-05-16
[Lua] Fix block comment end

Sainan:
This is how comments in Lua work:
```Lua
-- This is a comment

--[[This is block comment
spanning over multiple lines]]
```
I've personally never seen `--]]` used to end block comments.

### DIFF
--- a/Lua/Comments.tmPreferences
+++ b/Lua/Comments.tmPreferences
@@ -25,6 +25,18 @@
 				<key>value</key>
 				<string>]]</string>
 			</dict>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_START_3</string>
+				<key>value</key>
+				<string>--[[</string>
+			</dict>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_END_3</string>
+				<key>value</key>
+				<string>--]]</string>
+			</dict>
 		</array>
 	</dict>
 </dict>
--- a/Lua/Comments.tmPreferences
+++ b/Lua/Comments.tmPreferences
@@ -23,7 +23,7 @@
 				<key>name</key>
 				<string>TM_COMMENT_END_2</string>
 				<key>value</key>
-				<string>--]]</string>
+				<string>]]</string>
 			</dict>
 		</array>
 	</dict>


michaelblyons:
See discussion on #948 for context.

Sainan:
I do understand that's a neat trick, but the Comments.tmPreferences is specifically for the "Toggle (Block) Comment" feature in Sublime Text. I figure the amount of people using this feature specifically to produce block comments that they can later manually toggle instead of using the toggle feature is rather lower.

It would be more likely used to comment/uncomment code, for example:
```Lua
--[[if a == 3 then
    print("a is 3")
end]]
```

In this case, using `--]]` instead of `]]` would look unnatural. It's of course a small thing either way.

keith-hall:
Perhaps a compromise could be to add the minus-less version as `TM_COMMENT_START_3`/`TM_COMMENT_END_3` and leave number 2 the way it currently is. I believe then you can simply update your toggle comment keybinding for Lua files to tell it to prefer variant 3 :)

michaelblyons:
Would it improve behaviour at all to have `(?:--)?\]\]` as the match for the `punctuation.definition.comment` comment terminator?

Sainan:
@keith-hall Adding `--[[` and `]]` for number 3 does indeed solve the issue with "Toggle Block Comment" being unable to remove block comments, so that would be another way to resolve this issue if we're dead-set on keeping `--]]` as terminator.

@michaelblyons I don't think the grammar has anything to do with the Comments.tmPreferences.

michaelblyons:
Here's my proposal:

1. Default block comment style follows this PR's request. Since it's what the language actually uses.
2. A variant is added to do the previous `--]]` closing, so people can pick that if they liked it.

Once again, bear in mind that I do not write Lua and therefore have no "skin in the game." 

---

> I don't think the grammar has anything to do with the Comments.tmPreferences.

I had weird stuff with a block comment tweak I was running into that I think was solved by grammar edits, but it could have been something else that fixed it. 🤷‍♂️

Sainan:
I did implement the change as suggested by @michaelblyons in case it was not noticed.

michaelblyons (approved):

FichteFoll (approved):

PR #3755, opened 2023-05-16
[Batch File] Fix escape sequences, word boundaries and redirections

deathaxe:
This PR modifies behavior of Batch File.sublime-syntax to treat redirections as normal command arguments. That's what Bash also does.

Up to this commit, redirections have been expected, with some exceptions, to appear at the end of a command only. Thus they terminated statements.

This commit therefore applies following changes:

1. `<` and `>` are removed from command terminator patterns

2. a new `redirection_or_eoc` variable is introduced to terminate tokens, which are known not to contain redirections.

3. the `metachar` variable and some of its children are tweaked, mostly to reduce duplicated characters and to clarify dependencies.

4. `redirections` context is included into all relevant contexts

5. various remaining issues related to unquoted vs. quoted escape sequences and interpolations are fixed by including clearly named contexts:

   - quoted-escapes-and-interpolations
   - unquoted-escapes-and-interpolations
   
   into related quoted and unquoted token contexts.

   That's a follow-up of #3743.

6. word boundaries and line continuation treatment is improved in various commands and contexts

#### Benchmarks

Parsing performance is not changed by this PR.

### DIFF
--- a/Batch File/Batch File (Compound).sublime-syntax
+++ b/Batch File/Batch File (Compound).sublime-syntax
@@ -21,4 +21,4 @@ hidden: true
 extends: Packages/Batch File/Batch File.sublime-syntax
 
 variables:
-  command_terminator_chars: '[|&<>)]'
+  eoc_char: '[\n|&)]'
--- a/Batch File/tests/syntax_test_batch_file.bat
+++ b/Batch File/tests/syntax_test_batch_file.bat
@@ -471,14 +471,16 @@ ECHO : Not a comment ^
 :::: [ Conditionals ] :::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 
    IF foo EQU bar echo "equal"
-:: ^^  keyword.control.conditional.if.dosbatch
+:: ^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.statement.conditional.dosbatch
+:: ^^ keyword.control.conditional.if.dosbatch
 ::    ^^^ string.unquoted.dosbatch
 ::        ^^^ keyword.operator.comparison.dosbatch
 ::            ^^^ string.unquoted.dosbatch
 ::                ^^^^ support.function.builtin.dosbatch
 
    IF NOT foo EQU bar echo "equal"
-:: ^^  keyword.control.conditional.if.dosbatch
+:: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.statement.conditional.dosbatch
+:: ^^ keyword.control.conditional.if.dosbatch
 ::    ^^^ keyword.operator.logical.dosbatch
 ::        ^^^ string.unquoted.dosbatch
 ::            ^^^ keyword.operator.comparison.dosbatch
@@ -486,6 +488,7 @@ ECHO : Not a comment ^
 ::                    ^^^^ support.function.builtin.dosbatch
 
    IF NEQ == NEQ echo "equal"
+:: ^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.statement.conditional.dosbatch
 :: ^^ keyword.control.conditional.if.dosbatch
 ::    ^^^ string.unquoted.dosbatch
 ::        ^^ keyword.operator.comparison.dosbatch
@@ -493,6 +496,7 @@ ECHO : Not a comment ^
 ::               ^^^^ support.function.builtin.dosbatch
 
    IF NEQ NEQ NEQ echo "equal"
+:: ^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.statement.conditional.dosbatch
 :: ^^ keyword.control.conditional.if.dosbatch
 ::    ^^^ string.unquoted.dosbatch
 ::        ^^^ keyword.operator.comparison.dosbatch
@@ -500,6 +504,7 @@ ECHO : Not a comment ^
 ::                ^^^^ support.function.builtin.dosbatch
 
    IF foo == bar echo "equal"
+:: ^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.statement.conditional.dosbatch
 :: ^^ keyword.control.conditional.if.dosbatch
 ::    ^^^ string.unquoted.dosbatch
 ::        ^^ keyword.operator.comparison.dosbatch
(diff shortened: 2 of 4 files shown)


Batch File/tests/syntax_test_batch_file.bat
```diff
@@ -1408,36 +1434,49 @@ put arg1 arg2
...
 ::         ^^^ constant.language.null.dosbatch
 
-   dir > f.txt 2>&1
+   ECHO <contet.txt and others
+:: ^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.command.echo.dosbatch
+::      ^^^^^^^^^^^ meta.string.dosbatch meta.redirection.dosbatch - string string
+
+   ECHO "<contet.txt and others"
+:: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.command.echo.dosbatch
+::      ^^^^^^^^^^^^^^^^^^^^^^^^ meta.string.dosbatch string.unquoted.dosbatch - meta.interpolation - meta.redirection
```
michaelblyons, line 1450: Not `string.quoted.double`?

Batch File/Batch File (Compound).sublime-syntax
```diff
@@ -21,4 +21,4 @@ hidden: true
 extends: Packages/Batch File/Batch File.sublime-syntax
 
 variables:
-  command_terminator_chars: '[|&<>)]'
+  eoc_char: '[\n|&)]'
```
mataha, line 24: If I correctly understand the ramifications of adding a newline here, this:
```batchfile
for /f "usebackq delims=" %%i in (`echo a
echo b`) echo %%~i
```
Would make you think that the output will be `a` followed by `b`. But it actually is:
```batchfile
for /f "usebackq delims=" %%i in (`echo a echo b`) echo %%~i
```
And echoes `a echo b`.

Newlines don't terminate embedded commands, but are folded to a single space (like YAML's folded scalars).

Batch File/tests/syntax_test_batch_file.bat
deathaxe: No, because the output is `"<contet.txt and others"`. Quotes are treated literal in this situation. I know, it's weird. I tried to scope those quotes `punctuation` which are not rendered in output, to make it visible.

The main issue with those old shell languages (also Bash), is that every token / identifier may consist of quoted and unquoted parts, which are chained together. There is basically no such concept of pure quoted or unquoted strings.

Batch File/Batch File (Compound).sublime-syntax
deathaxe: The original intent was to simplify 

https://github.com/sublimehq/Packages/blob/34945b38165ee884ae5a0047e2e279692f0050f6/Batch%20File/Batch%20File.sublime-syntax#L46

by moving `$` to `eoc_char`, but well, this may have unwanted side effects.

deathaxe:
It appears I need to have a look at word boundaries in general, as those new tests from @mataha  revealed. I faked it by adding `\b\d` to numeric file handlers in redirections, but that's not enough.

Batch File/Batch File (Compound).sublime-syntax
deathaxe: Each command ends at eol, even in compund commands. The correct example for why `\n` is there would be:

```dosbatch
for %%i in (a b c) do (
	echo a
	echo b
)
```

Without `\n` in `eoc_char` it would look like:

grafik


But the 2nd `echo` is an individual command:

grafik

mataha: My bad, I misunderstood what you mean by that. Correct, these are individual commands.

Batch File/tests/syntax_test_batch_file.bat
michaelblyons: Understood. _Yuck!_

michaelblyons:
I gave this a test run. It's definitely improved. I do not know whether it's worth handling these weird and possibly ambiguous cases:

Option highlighting.
```batch
apple -[banana='/'].cherry:"%dogwood%"

apple /b/c dogwood
```

Double variable punctuation. (Maybe just a bug in the code. I haven't seen this anywhere else.)
```batch
IF NOT "!%apple%!" == "!%banana%!"
```

mataha:
> Double variable punctuation. (Maybe just a bug in the code. I haven't seen this anywhere else.)
> 
> ```batchfile
> IF NOT "!%apple%!" == "!%banana%!"
> ```

This is a common idiom that enables passing variables by reference. Your eyes may bleed, but consider the following:

```batchfile
echo off & setlocal DisableDelayedExpansion
set object.value=2
call :double object
echo(Value: %object.value%
endlocal & exit /b

:double (*object)
    setlocal EnableDelayedExpansion
    set arg=%~1
    set /a value=!%arg%.value! * 2 >nul
    endlocal & set "%arg%.value=%value%" & exit /b
```

deathaxe:
> apple -[banana='/'].cherry:"%dogwood%"

Basically arguments can be of any syntax as it is up to the called process to interpret the words. I guess we can find hundrets of failing examples. The only thing the shell interpreter does is passing the words (started and terminated by `{{metachar}}`) to the called command.

It could theoretically also be some kind of script being passed, regexp or whatever.

The same is true for `/b/c`. It is passed as a single word, as `/` is not a `{{metachar}}`. I've just added `:` and `=` to option terminators to support common terms like `/var:val` or `--var=val`.

>  "!%apple%!" 

This wasn't easy or even unable to solve when Batch was written for ST3. Considder `%apple:...%` also may contain replacements as the outer variable can. This needs branching to be solved reliably. This is IMHO however a separate topic.

mataha:
Re: above - that's not true for builtins:
```
C:\Users\user>for /f/?/x
Runs a specified command for each file in a set of files.

FOR %variable IN (set) DO command [command-parameters]

  %variable  Specifies a single letter replaceable parameter.
  (set)      Specifies a set of one or more files.  Wildcards may be used.
  command    Specifies the command to carry out for each file.
  command-parameters
             Specifies parameters or switches for the specified command.

...
```

deathaxe:
Like I said: It's up to the called command to interpret arguments' content. As such behavior differs for each called command, which syntax can't handle globally.

Just write a batch command and pass `mycommand /f/b/c`. All of them are assigned to `%1`. Same is true for externally called programs. `/f/b/c` could be a set of options or a path. It depends on the called command to interpret.

We could add some more special casing contexts to treat the most likely things well - as Bash provides dedicated rules for most known commands - but there is no general correct solution.

Syntactically, the shell passes words `/f/b/c` as a whole. If a command splits them to `/f` `/b` and `/c` the shell and its syntax can't know.

michaelblyons (approved):

jfcherng (approved):

PR #3762, opened 2023-05-17
[Rails] Toggle ERB instead of HTML comments

### LINKED ISSUE #3756: [Rails] Wrong comments for html.erb templates
### What happened?

Currently the HTML (Rails) syntax is using HTML comments, resulting in an error.

Current behavior
`<!-- <h1 class="sidebar__name"><%= current_user.name %></h1> -->`

Expected behavior
`<%# <h1 class="sidebar__name"><%= current_user.name %1></h1> %>`

deathaxe:
Fixes #3756

This PR adds definition for ERB comments to be added/removed when toggling comments in ERB templates.

### DIFF
--- a/Rails/HTML (Rails) - Comments.tmPreferences
+++ b/Rails/HTML (Rails) - Comments.tmPreferences
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+	<key>scope</key>
+	<string>
+		source.css.rails,
+		source.js.rails,
+		source.json.rails,
+		text.html.rails,
+		text.html.rails source.css.embedded,
+		text.html.rails source.js.embedded,
+		text.html.rails source.json.embedded
+	</string>
+	<key>settings</key>
+	<dict>
+		<key>shellVariables</key>
+		<array>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_START</string>
+				<key>value</key>
+				<string><![CDATA[<%# ]]></string>
+			</dict>
+			<dict>
+				<key>name</key>
+				<string>TM_COMMENT_END</string>
+				<key>value</key>
+				<string><![CDATA[ %>]]></string>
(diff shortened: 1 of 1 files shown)


deathaxe:
I wonder if template syntaxes should support toggling both, the primary template comment (e.g.: `<%# ... %>` and the underlying syntax's native one (e.g.: `<!-- ... -->` or `/**/`) via custom key bindings.

If so, 

1. which bindings would be appropriate?
2. would the native one need to be moved to variant=2 and bound to a new key combination?
3. would the template comment be added as variant 2 and bound to a new key combination?
4. Would both need to support `line` and `block`?

michaelblyons:
I used to do a lot of work in ST with [HTML (C#)][cshtml]. I was pretty happy with [only the HTML part][tmpref] doing "special" comments[^2] and leaving embedded JS, CSS, etc. to do their own kind of commenting. I'm open to hearing other people's workflows and whether they think it's preferable to use server-side comments for those, whether it's necessary to have _both_ server-side and client-side comments available with hotkeys, etc.

It is very helpful that HTML's native comments are already "block," so I didn't need to worry about line/block comment precedence.

[^2]: Ignore the `embedding.*`. I know I need to change it to `text.html.cs.webforms` or something.

[cshtml]: https://github.com/michaelblyons/SublimeSyntax-HTML-CSharp
[tmpref]: https://github.com/michaelblyons/SublimeSyntax-HTML-CSharp/blob/master/support/ASP.Net/Comments.tmPreferences

michaelblyons (approved):

deathaxe:
JS/CSS have been added as block comments accross script/style tags didn't work with `text.html.rails` being the only scope.

IMHO, ideally both kinds of comments could be toggled.

Maybe it makes most sense to primarily toggle server-side comments in templates and have an alternative binding to toggle client-side comments.

The primary issue is client-side comments need to be re-defined by the templating package (Rails, JSP) to be moved to variant 2. That's just a bit more work.

PR #3764, opened 2023-05-17
[Go] Add Template Snippets

deathaxe:
This PR proposes to add basic snippets for Go templates.

#### Examples:

    tif      => {{ if | }}
    tifend   => {{ if | }} ... {{ end }}
    telse    => {{ else }}

Triggers follow the scheme to prefix each valid template keyword with `t`.

#### Note:

We basically don't want to add new snippets, so an alternative would be to publish them as separate package, but I always found some helpers to add template tags very helpful.

Thoughts?


michaelblyons (approved):

keith-hall (approved):
I think these could be worth adding, especially as people can easily disable them from the preferences if they don't want them

Issue #3694, opened 2023-04-07
[RFC] Provide a generic way to highlight operator keywords


jwortmann:
### What happened?

This issue is supposed to track the status of the problem described in the rejected and now closed PR #2090, because finding and implementing a solution is still relevant to me. Probably most of the important aspects have already been discussed in the referenced PR, however, no final decision was taken.

tldr; it is currently impossible in Sublime Text color schemes to target operator keywords in a general way, and sometimes there doesn't even exist a workaround to highlight certain operator keywords, because they use the exact same scope as operator symbols.

There is also a rule for it in the [scope naming guidelines](https://www.sublimetext.com/docs/scope_naming.html#keyword), but it is not widely used in the default syntaxes, and the consensus seems to be that this rule should not be followed.

![ScopeNamingGuidelines](https://user-images.githubusercontent.com/6579999/230657852-7f0b155f-7eda-4d61-8f4b-81ce762ab16f.png)

---

Finally, here are screenshots from a few other editors, with two code examples from Rust & PHP. Mostly (but not always) those editors are able to highlight operator keywords, but no operator symbols, and this is even the default style in their built-in themes (pay attention to `in` keyword in Rust, and `and` & `or` keywords in PHP):

<details><summary>Visual Studio Code</summary>

(Fails to highlight `and` & `or` keywords in PHP)

![VisualStudioCode](https://user-images.githubusercontent.com/6579999/230657957-0bb9f8f9-cb40-4693-9cd3-4d8c34e869e7.png)

</details>

<details><summary>Notepad++</summary>

![Notepad++](https://user-images.githubusercontent.com/6579999/230657977-d64c72a6-8a4e-46c6-8b2e-4883e6a5d290.png)

</details>

<details><summary>JetBrains Fleet</summary>

![JetBrainsFleet](https://user-images.githubusercontent.com/6579999/230657998-4f34b2f3-4db6-4ce6-af98-c5e560905c24.png)

</details>

<details><summary>Syntax highlighting on GitHub</summary>

```rust
fn main() {

    for _ in 1..3 {}

    let n = 1 + 1;

    match n {
        1 => println!("One"),
        2 => println!("Two"),
    }
}
```

(Fails to highlight `and` & `or` keywords in PHP)
```php
<?php

$a = (false && foo());
$b = (true  || foo());
$c = (false and foo());
$d = (true  or  foo());

?>
```

</details>

frou:
I agree that this is a very conventional style and not some oddball request.

[...]

PR #3765, opened 2023-05-18
[TypeScript] Add missing sub-scopes to operator keywords

jwortmann:
Some operator keywords in the TypeScript syntax use quite specific sub-scopes, for example
https://github.com/sublimehq/Packages/blob/17188f5973c6762ce60793dc0f7711243853ca94/JavaScript/TypeScript.sublime-syntax#L779-L784
and
https://github.com/sublimehq/Packages/blob/17188f5973c6762ce60793dc0f7711243853ca94/JavaScript/TypeScript.sublime-syntax#L801-L802
while some others do not.

This PR adds the corresponding sub-scopes to the operator keywords where they were missing, so that color schemes are able to target them one-by-one (as a workaround for #3694).

I didn't add it to the operator symbols where the sub-scopes are missing, because I'm not a TypeScript user and don't know what they mean:
https://github.com/sublimehq/Packages/blob/17188f5973c6762ce60793dc0f7711243853ca94/JavaScript/TypeScript.sublime-syntax#L725-L726
https://github.com/sublimehq/Packages/blob/17188f5973c6762ce60793dc0f7711243853ca94/JavaScript/TypeScript.sublime-syntax#L804-L808

But I could add them as well, if so desired, and someone tells me the appropriate scope names.

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -397,7 +397,7 @@ contexts:
             scope: variable.other.readwrite.js
             push:
               - match: in{{identifier_break}}
-                scope: keyword.operator.type.js
+                scope: keyword.operator.type.in.js
                 set:
                   - ts-type-expression-end
                   - ts-type-expression-end-no-line-terminator
@@ -726,7 +726,7 @@ contexts:
       scope: keyword.operator.type.js
 
     - match: as{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.as.js
       push:
         - match: const{{identifier_break}}
           scope: storage.modifier.const.js
@@ -739,7 +739,7 @@ contexts:
             - ts-type-expression-begin
 
     - match: satisfies{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.satisfies.js
       push:
         - ts-type-meta
         - ts-type-expression-end
@@ -786,7 +786,7 @@ contexts:
       scope: keyword.operator.word.js
       push: [ts-type-expression-end-no-line-terminator, ts-type-expression-begin]
     - match: as{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.as.js
       push:
         - ts-type-expression-end
         - ts-type-expression-end-no-line-terminator
@@ -869,11 +869,11 @@ contexts:
 
   ts-type-expression-begin:
     - match: keyof{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.keyof.js
     - match: typeof{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.typeof.js
     - match: infer{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.infer.js
     - match: new{{identifier_break}}
       scope: keyword.operator.new.js
 
@@ -887,7 +887,7 @@ contexts:
       scope: storage.modifier.abstract.js
 
     - match: import{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.import.js
       set:
         - match: \(
           scope: punctuation.section.group.begin.js
--- a/JavaScript/tests/syntax_test_typescript.ts
+++ b/JavaScript/tests/syntax_test_typescript.ts
@@ -723,15 +723,15 @@ function f(this : any) {}
 /* Assertions */
 
 x as boolean;
-//^^ keyword.operator.type
+//^^ keyword.operator.type.as
 //   ^^^^^^^ meta.type support.type.primitive.boolean
 
 x satisfies boolean;
-//^^^^^^^^^ keyword.operator.type
+//^^^^^^^^^ keyword.operator.type.satisfies
 //          ^^^^^^^ meta.type support.type.primitive.boolean
 
 x as const;
-//^^ keyword.operator.type
+//^^ keyword.operator.type.as
 //   ^^^^^ storage.modifier.const
 
     foo!.bar;
@@ -900,11 +900,11 @@ let x: 0xabc;
 
 let x: typeof Foo;
 //     ^^^^^^^^^^ meta.type
-//     ^^^^^ keyword.operator.type
+//     ^^^^^ keyword.operator.type.typeof
 //            ^^^ support.class
 let x: keyof Foo;
 //     ^^^^^^^^^ meta.type
-//     ^^^^^ keyword.operator.type
+//     ^^^^^ keyword.operator.type.keyof
 //           ^^^ support.class
 
 let x: Foo.bar;
@@ -1038,8 +1038,8 @@ let x: {
 //  ^^^^^^^^^^^^^^^^ meta.brackets
 //  ^ punctuation.section.brackets.begin
 //    ^ variable.other.readwrite
-//      ^^ keyword.operator.type
-//         ^^^^^ keyword.operator.type
+//      ^^ keyword.operator.type.in
+//         ^^^^^ keyword.operator.type.keyof
 //               ^ meta.brackets support.class
 //                 ^ punctuation.section.brackets.end
 //                   ^ punctuation.separator.type
@@ -1054,10 +1054,10 @@ let x: {
 //  ^^^^^^^^^^^^^^^^^^^^^ meta.brackets
 //  ^ punctuation.section.brackets.begin
 //    ^ variable.other.readwrite
-//      ^^ keyword.operator.type
-//         ^^^^^ keyword.operator.type
+//      ^^ keyword.operator.type.in
+//         ^^^^^ keyword.operator.type.keyof
 //               ^ meta.brackets support.class
-//                 ^^ keyword.operator.type
+//                 ^^ keyword.operator.type.as
 //                    ^ meta.brackets support.class
 //                      ^ punctuation.section.brackets.end
 
@@ -1067,8 +1067,8 @@ let x: {
 //             ^^^^^^^^^^^^^^^^ meta.brackets
 //             ^ punctuation.section.brackets.begin
 //               ^ variable.other.readwrite
-//                 ^^ keyword.operator.type
-//                    ^^^^^ keyword.operator.type
+//                 ^^ keyword.operator.type.in
+//                    ^^^^^ keyword.operator.type.keyof
 //                          ^ meta.brackets support.class
 //                            ^ punctuation.section.brackets.end
 //                              ^ storage.modifier
@@ -1186,7 +1186,7 @@ let x: T extends infer U ? V : W;
 //     ^^^^^^^^^^^^^^^^^^^^^^^^^ meta.type
 //     ^ support.class
 //       ^^^^^^^ keyword.operator.type.extends
-//               ^^^^^ keyword.operator.type
+//               ^^^^^ keyword.operator.type.infer
 //                     ^ support.class
 //                       ^ keyword.operator.type
 //                         ^ support.class
@@ -1195,7 +1195,7 @@ let x: T extends infer U ? V : W;
 
 let x: import ( "foo" ) . Bar ;
 //     ^^^^^^^^^^^^^^^^^^^^^^^ meta.type
-//     ^^^^^^ keyword.operator.type
+//     ^^^^^^ keyword.operator.type.import
 //            ^^^^^^^^^ meta.group
 //            ^ punctuation.section.group.begin
 //              ^^^^^ meta.string string.quoted.double


jfcherng (approved):

JavaScript/TypeScript.sublime-syntax
```diff
@@ -726,7 +726,7 @@ contexts:
       scope: keyword.operator.type.js
 
     - match: as{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.as.js
```
deathaxe, line 729: The keyword `as` appears in many syntaxes with very similar meaning as well.

It is often used to assign the name or value from a left hand operator to the right.

As such it is just a `keyword.operator.assignment`. Maybe `keyword.operator.assignment.alias` as this is what it is often used for.

JavaScript/TypeScript.sublime-syntax
```diff
@@ -739,7 +739,7 @@ contexts:
             - ts-type-expression-begin
 
     - match: satisfies{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.satisfies.js
```
deathaxe, line 742: I am not very confident with regards to all those type evaluation related keywords such as `typeof`, `instanceof`, ... accross all syntaxes.

Java syntax specification calls related expressions `type comparison expression`.

As such, those keywords could all be grouped under `keyword.operator.type`. In general, all of them - like `in` - have just the meaning to check, whether an object is of certain type.

Thus `keyword.operator.logical` or - following Java's naming - `keyword.operator.comparison` would be a good base scope. Maybe `keyword.operator.comparison.type`, which would probably be better aligned with suggestions for scoping `in` and `as`.

Same applies for the keywords below.

JavaScript/TypeScript.sublime-syntax
```diff
@@ -887,7 +887,7 @@ contexts:
       scope: storage.modifier.abstract.js
 
     - match: import{{identifier_break}}
-      scope: keyword.operator.type.js
+      scope: keyword.operator.type.import.js
```
deathaxe, line 890: A import a `keyword.operator`? Isn't `import` not just a `keyword.control`? I wouldn't scope it different from normal/other import keywords.

deathaxe (commented):
Haven't had a closer look at all those scopes before in JS, but I am not very confident with scoping all keywords `keyword.operator.type`, just because they appear int type hints. Adding more specific scopes, doesn't improve that situation.

Some keywords such as `as` or `in` appear in various syntaxes and context with similar meanings. Thus using a common scope would probably be useful.

The following comments provide some suggestions from experience made while working on different syntaxes.

They are made despite the open word-like vs. symbolic operator scoping issue.

JavaScript/TypeScript.sublime-syntax
```diff
@@ -397,7 +397,7 @@ contexts:
             scope: variable.other.readwrite.js
             push:
               - match: in{{identifier_break}}
-                scope: keyword.operator.type.js
+                scope: keyword.operator.type.in.js
```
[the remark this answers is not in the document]
deathaxe: You are right with `keyword.operator.comparison` probably being a better choice. 

I can't really follow operator's being rather messy, though. They may probably not yet be perfect, but mainly because we've not yet touched all syntaxes in detail. Type specific operators and keywords haven't been standardized yet and there might probably always be syntax specific special cases not fitting into any existing scheme, but well.

The discussion about `in` was probably made with regards to Python, as this is what it does. It scopes them differently depending on statement. I don't really find that useful and some other syntaxes already follow the `keyword.operator` scheme.

deathaxe:
JavaScript is the only syntax making use of `keyword.operator.type` so far, except Go, which scopes a keyword `type` like that. Not sure if we want to open jet another route for further defragmentation of scopes.

What's the use case for more specific type scopes beyond probably targeting them individually to tint them like normal keywords or even plain text?


No progress with regards operator scopes has been seen, as it is rather breaking and huge, and other more relevant/critical changes seemed more important, with regards to the spare time left to work on open source projects. And most color schemes are fine with scoping all operators with the same color. There are only few, which want symbol-like operators to look like plain text.

JavaScript/TypeScript.sublime-syntax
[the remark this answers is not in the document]
Thom1729: This particular `in` isn't the regular JS `in` operator (which is unchanged in TS), but rather special syntax for mapped types:

```ts
type FlagSet<Keys extends string> = {
  [K in Keys]: boolean
}
```

Honestly, it's not really an operator in the conventional sense. But to the extent that we treat it like one, it *is* type-specific, and it's not a comparison.

It should probably have the same scope as the `in` or `of` in a for loop:

```js
for (const x in y) {}
for (const x of y) {}
```

Both of those are currently `keyword.operator.word`, which is also questionable. By contrast, the actual `in` operator is currently `keyword.operator`, with no `word`.

If I picked a brand-new scope today for `in`/`of` in for loops in mapped types, I might pick `keyword.declaration`. I don't have a strong opinion about changing it now or what to change it to, but I acknowledge that the current scopes are a bit silly.

JavaScript/TypeScript.sublime-syntax
Thom1729: `satisfies` is a weird TS-specific operator that's basically an inline type check that can also inform inference. I would scope it either identically to `as` or using the same approach as `as`. They both are only used in the type system and have no runtime effect.

JavaScript/TypeScript.sublime-syntax
Thom1729: This is a syntax for using imported types in a non-module context where you can't use `import` or `import type` declarations:

```ts
type Foo = import("./someModule").Foo;
```

I feel weird about the scoping. It's not really an operator (it's not operating on the `"./someModule"` type), but it's not really a control keyword either:

1. I'm actually a bit confused about `import` being a control keyword in the first place (although that's what the JS syntax currently does). I think that makes sense in some languages, where importing is essentially copy-pasting another file, or a higher-level language feature that loads and executes the other file in-line. In fairness, JS is such a language, but I wonder if it really generalizes?
2. TS has `import type from 'path'`, which definitely isn't a control keyword; the entire point of it is that it has no runtime effect, whereas even `import { type T } from 'path'` causes the module to be loaded. But it's basically the same syntax, so it probably should use the same scope even if it's not a perfect fit.
3. The `import("./someModule")` syntax definitely isn't a control keyword, and it also isn't just a variation on syntax that could be a control keyword. But it's not an operator either, and it probably should be at least similar to other import stuff.

I'm not sure how other syntaxes treat import keywords. I'm pretty sure that JS's `keyword.control.import-export` is not standard. If something like `keyword.import` were standard, it would make sense for all of the above.

jwortmann:
> What's the use case for more specific type scopes beyond probably targeting them individually to tint them like normal keywords or even plain text?

As already explained above, this is the only reason. This PR intentionally doesn't fix or "change" any scopes, but only makes them more specific by adding the sub-scope. And as a side effect, it makes this syntax a little bit more consistent in itself, because as shown in the opening comment, three of those sub-scopes already exist, while the scopes of those other keywords currently end in `.type.js`. With this PR, all of them (excluding the three symbols where I don't know what would be an appropriate sub-scope) would have 4 levels before the language suffix.

> more relevant/critical changes seemed more important, with regards to the spare time left to work on open source projects

This was not meant as a request or anything to yourself personally. It's rather an assessment that there is nobody from SublimeHQ anymore who would make such a decision, especially if it would be a breaking change, like for example wbond did when the number scoping scheme was changed. And the number scopes, by the way, are *still* not updated in the scope naming guideline docs on the website. This makes me assume that they are basically abandoned.

But in regard to the word vs symbol operator issue, solving it doesn't necessarily imply a breaking change. It could be done in a fully backwards compatible way, for example by just applying `meta.operator.word` to all word operators, and leaving the regular scopes as they are now. Frankly, it would feel like a quite hacky solution, and I would also prefer having one of your cleaner (but breaking) suggestions from #3694. But the problem is that there seems nobody to be here in this repo who is empowered or willing to make a decision for this.

> And most color schemes are fine with scoping all operators with the same color. There are only few, which want symbol-like operators to look like plain text.

I believe that this is rather a consequence, and not the cause of the current state of the scope names. As I have shown in some screenshots in #3694, there are many other editors where such a style is even the default style. Those are only a few examples, and I could probably go on with many more. For example also the Matlab IDE, and IIRC Netbeans IDE and also Visual Studio don't highlight the symbols. It is much more that Sublime Text is basically the *exception*, where it's sheer impossible or at least very hard (requiring very specific scopes to target only operator keywords/symbols in all languages) to use this highlighting style. And this is why most color schemes don't bother to try that, or just use a more generic selector like `keyword - keyword.operator, keyword.operator.word`.

JavaScript/TypeScript.sublime-syntax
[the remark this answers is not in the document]
deathaxe: I somehow had python's `if foo in [a,b,c]` statement in mind, when it came to `logical` vs. `comparison` scopes.

I wonder if it makes sense to introduce a `keyword.operator.iterator` scope for such keywords in loops. It might even be suitable in test statements such as the one above, if the forementioned scopes are not desired.

JavaScript/TypeScript.sublime-syntax
deathaxe: Loading something or just declaring an import, I second your concerns about `control` scope. `includes` and `imports` are more like general directives, which may be executed by a preprocessor, compiler or at runtime - depending on language.

Compiled languages like C or Erlang have various keyword being used in that non-control-flow context, which are "interpreted/evaluated" ahead of runtime (compile or import time). We should at least have a 2nd-level scope, which all of them could be grouped under. IIRC, there is an open RFC about that topic though. 

That said, I have a feeling we shouldn't push direction too much into semantic highlighting of keywords as we often can't reliably detect and scope them.

deathaxe:
> https://github.com/sublimehq/Packages/issues/3694, there are many other editors where such a style is even the default style.

I wouldn't call those examples "many" 50:50 at best.

> meta.operator.word to all word operators,

Meta scopes are primarily for larger parts of code. I am personylly not willing to add a meta scope for each char next to a normal scope. That's weird to have dozens for each tiny token.

jwortmann:
> 50:50 at best

Do you have any evidence to support this claim?

> Meta scopes are primarily for larger parts of code. I am personylly not willing to add a meta scope for each char next to a normal scope. That's weird to have dozens for each tiny token.

Yeah, as written above I also wouldn't like it very much. But I mentioned it, because it was a proposed solution by wbond and the advantage is that it would be a non-breaking change.

> in case those would be changed to something more general in near future

If you are referring to a solution to #3694, then this is the point where I don't see any solution on the radar **in the near future**, or even being implemented at all. This is why I made this PR, and did it in the most simple way by not changing any of the existing scope parts. If you refer to the `keyword.operator.type` scope being wrong in general for those keywords, and changing all of its instances to different scopes like `keyword.operator.logical`, `keyword.control.import` and so on, then I would simply keep out of this discussion, because I don't use TypeScript and don't have the knowledge about how they should be scoped in the specific situations. But I think that should rather be the matter for a different PR, or alternatively someone else could make the appropriate changes here. At least I didn't see any open PR about this.

> is of little value

This is your opinion, but I would say the value is definitely bigger than zero and it is at least a step in the direction to improve the status quo. The changes are fully backwards compatible, minimum effort to review, and they hurt nobody. It doesn't mean that the changed scopes are set in stone for all time. If you have any improvements or find good conventions later, it can just be changed again later in another PR. But rejecting this one, just for the reason that there is not yet a perfect solution for the underlying issue, would be pretty unkind, unless you have a convincing reason why the PR is bad.

jwortmann:
Why was this PR closed? It already had 2 approvals?

jfcherng:
> If the goal is only to special target them in a single color scheme, that's probably not worth it

Fwiw, I like this.

deathaxe:
Fedup of endless worthless discussions!

jwortmann:
But all those discussion were started by you. I pointed out several times that the `keyword.operator.type` scope is not even part of this pull request, and even if it were some other random scope which doesn't make sense at all, for example `keyword.operator.foobar`, then the PR would still leave it like that. In particular, the PR *does not* excess its usage, because all of the affected keywords already had it like that before. So I would consider all comments about `keyword.operator.type` as offtopic here in this PR.

You are free to create a RFC to discuss the `keyword.operator.type` scope, or to make another PR to erase that scope. But I won't participate in such a discussion, because I don't care about what this scope is. What I do care about, and what the single intention of this PR is, is to allow to distinguish operator keywords from operator symbols in TypeScript.

What I don't understand is, that by closing the PR, and continuously arguing that it were only me who wants to highlight keywords differently from operator symbols, you are actively trying to prevent making this highlighting style possible for users who want it. I am certain that I'm not the only one on this world you prefers this highlighting style, in the RFC issue someone else also agreed that it isn't just some oddball request, and I hope by providing a list of other editors, IDEs and websites (you can add GitLab to this list as well btw), which use this style even by default, it is clear that it is a very convenient style. So why make life harder for others, if it doesn't affect you? Nobody forces you to use such a highlighting style, and if you don't want it, with this PR everything stays the same as before for you. Why not just hit the merge button and everybody is happy?

Issue #3778, opened 2023-05-26
[Scala] "with" keywords highlighted as out of place when comment between them and extends


fhunstock-semron:
### What happened?

Given the following scala code,
this will highlight the `with` keywords as out of place:
```scala

class c()
    extends a()
    // some comment
    with foo with bar {

    }
```
while the following works fine:
```scala
class c()
    extends a()
    with foo with bar {

    }
```

Tested on build 4143 on the stable channel

djspiewak:
This may not actually be solvable.

The problem here is inferring when the class definition ends in general (especially when you don't have an open curly brace) is based on a complex series of rules, most notably including a double newline. This is the same set of rules that applies to other type definitions. The issue in your example is there *is* a double newline, it's just hidden by the comment. Scala reasonably considers the comment line to be "not blank", but it's harder for Sublime to make the same judgement.

I would definitely like to fix this, but I'm somewhat skeptical that it's in the cards.

PR #3779, opened 2023-05-26
[Scala] Fix `with` keyword after comments in class declarations

deathaxe:
I am not very confident, but would this PR be an appropriate fix for #3778?

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -520,6 +520,11 @@ contexts:
   # this is included when a single newline is found in a declaration
   # you should never push/set this context, only include
   decl-newline-double-check:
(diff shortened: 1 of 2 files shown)


djspiewak:
Possibly! It looks promising. Can you add a test showing that it still infers the double newline when one line is a comment?

deathaxe:
Can you give an example of what's missing in the added test cases?

FichteFoll (approved):
I don't know much about Scala, but this looks reasonable.

djspiewak:
Sorry for the long delay on this. I think #4124 includes a more robust fix for this problem and avoids the issue of disabling comments in very specific intermediate contexts.

Issue #3766, opened 2023-05-18
[RFC] Path wildcard scope


michaelblyons:
```bash
cat *.txt
```

For `*` and `?` as wildcards[^1], we have several schemes:

| Scope | Syntaxes | Comments |
|-|-|-|
| `constant.other.placeholder` | Batch | |
| `keyword.operator.path.asterisk` | Git ||
| `keyword.operator.quantifier` | Bash, _lots_ of Regex syntaxes | It's used twice in Bash: Once as regex quantifier and once as argument wildcard for paths. |
| `keyword.operator.wildcard` | File Patterns, SQL, Java | File patterns single level, SQL `LIKE` strings |
| `variable.language.wildcard` | File Patterns, SQL, Makefile | File patterns multilevel, SQL `SELECT *` |

I think Regex is its own thing and should stay as `keyword.operator.quantifier`, but for the others, should they be consolidated to one or two scopes?

I suggest that `**`-type multilevel ones use `variable.language.wildcard` and `*`-type short ones use `keyword.operator.wildcard`, but that's just what I did for File Patterns for the "Find in Files" paths. Other opinions welcome.

[^1]: In SQL, `%` and `_`.

deathaxe:
1. `constant.other.placeholder` seems most likely being used for placeholders in format strings (e.g.: `%s`). IIRC, this scope was also used for path wildcards in older syntaxes, which is not wrong, but hard for color schemes to distinguish them, in case both appear next to each other.

    As such a dedicated scope is appriciated.

2. As `?` and `*` don't denote placeholders or wildcards, but operators in regexp, the chosen scope is ok as is.

3. I have been struggling with `keyword.operator.wildcard` vs. `variable.language.wildcard` various times in the past.

   Which scope is more suitable, semantically?

   With regards to paths we might also take into account special signs like `~`, `.` or `..` and how they look compared to wildcards - using simple color schemes. Are they even of same category? 
   
   - `~` = `variable.language.environment` (Git Link)
   - `.`, `..` = `constant.language.path`
  
   Wildcards look like operators and "manipulate" a string in some way.
   
   They can also be interpreted as some kind of interpolation or expansion, thus `variable` is suitable, too.
   
   Why would you distinguish `*` vs. `**`? Both are just wildcards, aren't they?


What we might take into account with regards to highlighting using simple color scheme rules is:

1. wildcards may appear together with placehilders in format strings
[...]

PR #3781, opened 2023-05-31
[CSS] Change wildcard operator scopes

deathaxe:
This commit follows the suggestion of #3766 and scopes wildcards `?` and `*` `keyword.operator.wildcard`.


michaelblyons (approved):

deathaxe:
With a look at Mariana color scheme I wonder whether `variable.language.wildcard` would be the better choice in general, especially when combined with chaining (or other) operators.

PR #3782, opened 2023-06-01
[Markdown] Add folding rules

deathaxe:
Fixes https://github.com/sublimehq/sublime_text/issues/6005

This commit enables scope based folding of headings and fenced code blocks.

### DIFF
--- a/Markdown/Fold.tmPreferences
+++ b/Markdown/Fold.tmPreferences
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plist version="1.0">
+<dict>
+    <key>scope</key>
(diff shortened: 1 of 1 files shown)


jfcherng:
Does it make sense to let h1 fold until next h1, h2 fold to next h2 etc?

deathaxe:
Idially yes, but ST doesn't handle that well. Folding goes wild on certain headings.

FichteFoll (approved):
As mentioned, it would of course be nice if this was aware of the heading levels, but that's not technically possible.

Regardless, this is definitely an easy and quick improvement over the existing state (with no such folding whatsoever) and worked nicely on a few test files.

jfcherng (approved):

PR #3785, opened 2023-06-03
[CSS] Add nested rules support

### LINKED ISSUE #3688: [CSS] Support CSS Nesting
### What happened?

Both Chrome 112 and Safari Technical Preview 162 now implement [CSS Nesting](https://www.w3.org/TR/css-nesting-1/).

More information is available here:
https://developer.chrome.com/articles/css-nesting/
https://webkit.org/blog/13813/try-css-nesting-today-in-safari-technology-preview/

Sublime doesn't yet highlight nested selectors. In the following example, `color` lacks the `support.type.property-name.css` scope:

```
.foo { .bar { color: red; } }
```

I worked around this in my local `CSS.sublime-syntax` file by appending the following to `rule-list-body`:

```
    - match: (?=[&:.*#\[]|{{combinators}})
      push: main
```

deathaxe:
Resolves #3688

This commit implements support for nested selectors and rule lists according to current draft specification at https://drafts.csswg.org/css-nesting-1.

### DIFF
--- a/ASP/syntax_test_asp.asp
+++ b/ASP/syntax_test_asp.asp
@@ -208,7 +208,7 @@
     '       ^^^^^^^^ source.asp.embedded.html variable.other.asp
     '               ^^ punctuation.section.embedded.end.asp
     '                  ^ punctuation.section.block.begin.css
-    '                    ^^^^^^^^^ meta.embedded.asp
+    '                    ^^^^^^^^^ meta.property-name.css support.type.property-name.css meta.embedded.asp
     '                             ^ punctuation.separator.key-value.css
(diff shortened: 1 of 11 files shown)


deathaxe:
While changes to CSS are simple, the impact on all sorts of inherited syntaxes used in ASP, JSP, Rails, PHP, ... is rather significant.

Sometimes it's enough to change syntax test assertion punctuations.

But there are also cases, which need more work.

deathaxe:
Tailwind CSS has implemented this PRs changeset for a while without negative feedback.

Less and Liquid have been updated accordingly.

Other packages like Vue, Blade seem uneffected.

keith-hall (approved):

jfcherng (approved):

PR #3791, opened 2023-06-11
[Python] Fix target lists in generators

deathaxe:
This commit fixes unpacking operations in generator target lists and maybe other expressions by replacing limited `target-lists` by normal expressions.

### DIFF
--- a/Python/tests/syntax_test_python.py
+++ b/Python/tests/syntax_test_python.py
@@ -2383,9 +2383,9 @@ generator = ((k1, k2, v) for ((k1, k2), v) in xs)
 #            ^^^^^^^^^^^ meta.sequence.tuple.python
 #           ^^ punctuation.section.sequence.begin.python
 #                      ^ punctuation.section.sequence.end.python
-#                            ^^ punctuation.section.tuple.begin.python
-#                                    ^ punctuation.section.tuple.end.python
-#                                        ^ punctuation.section.tuple.end.python
+#                            ^^ punctuation.section.sequence.begin.python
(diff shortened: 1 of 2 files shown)


deathaxe:
FWIW, this PR fixes an issue of same category as https://github.com/sublimehq/Packages/issues/3794. 

The added pattern of the test case was found on a blog about using unpack operators in python.

It revealed assumed/applied simplification to for statements might be insufficient. Hence reverting to support full set of expressions.

michaelblyons (approved):

jfcherng (approved):

Python/tests/syntax_test_python.py
```diff
@@ -2406,11 +2406,54 @@ def f(): pass
...
 #                  ^ punctuation.section.sequence.end.python
-#                        ^^ punctuation.section.tuple.begin.python
-#                                ^ punctuation.section.tuple.end.python
-#                                    ^ punctuation.section.tuple.end.python
+#                        ^^ punctuation.section.sequence.begin.python
+#                                ^ punctuation.section.sequence.end.python
+#                                    ^ punctuation.section.sequence.end.python
 #                                           ^ punctuation.section.mapping.end.python
 
+unpack_ = [*(_ for [[][:], *_] in [((1, 2), "woops")])]
```
FichteFoll, line 2414: I am legitimately surprised that this is *not* a syntax error, specifically allowing to assign something to `[]`. :roll_eyes: 
But well, neither is `[] = []` so it's at least consistent.

PR #3797, opened 2023-07-01
[Python] Fix string replacement fields

### LINKED ISSUE #3649: [Python] Breaks on certain dict
### What happened?

```python
x = {
  "diagnostics": [
    {
      "message": "Type '{ _client: { _user?: { [x: string]: unknown; id: string; anon?: boolean | undefined; name?: string | undefined; role?: string | undefined; teams?: string[] | undefined; username?: string | undefined; ... 18 more ...; roles?: string[] | undefined; } | { ...; } | undefined; ... 202 more ...; commitMessage: (id: string...' is missing the following properties from type 'Channel<DefaultGenerics>': _client, type, id, data, and 74 more."
    },
  ]
}
```

Screenshot 2023-02-01 at 15 31 48

deathaxe:
Fixes #3649

This PR closely implements python's format string syntax specification.

see: https://docs.python.org/3.11/library/string.html#format-string-syntax

Only valid qualified identifiers are matched as replacement field names in order to reduce chance of false positives, even though python syntactically resolves all kinds of key names. _It appears to be an implementation detail instead of desired behaviour._

Replacement field termination is handled separately for each kind of quotation to ensure to correctly terminate replacement fields at the end of a string.

Format specs are handled more lazily following f-string implementation to better support time formats.

#### Benchmarks

Parsing performance is not effected noticeably.

Checked with 10k lines of differently complex templates:

```py
# 73ms -> 75ms
"From {} to {}"
```

```py
# 74ms -> 76ms
"From {0} to {0}"
```

```py
# 83ms -> 90ms
"From {0.prop[4]} to {0.prop[4]}"
```

```py
# 110ms -> 105ms
"From {0.prop[4]!r:{p}^9} to {0.prop[4]!r:{p}^9}"
```

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -2559,7 +2559,7 @@ contexts:
 
   triple-double-quoted-plain-raw-f-string-content:
     - include: string-prototype
-    - include: f-string-replacements
+    - include: triple-double-quoted-f-string-replacements
 
   triple-double-quoted-raw-f-strings:
     # Triple-quoted raw f-string, treated as regex
@@ -2581,7 +2581,7 @@ contexts:
 
   triple-double-quoted-raw-f-string-content:
     - include: string-prototype
-    - include: f-string-replacements-regexp
+    - include: triple-double-quoted-f-string-replacements-regexp
 
   triple-double-quoted-plain-raw-u-strings:
     # Triple-quoted capital R raw string, unicode or not, no syntax embedding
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):
It has been working for me. Confirmed more rigorous than the previous format.

I don't have too many files with nested format strings, but I tinkered with them, trying to fool the syntax and it was always correct. The test cases are more pathological than anything I tried.

jfcherng (approved):

PR #3806, opened 2023-07-12
[ShellScript] Fix test expressions

### LINKED ISSUE #3804: [ShellScript] Arguments to `[` and `test` incorrectly have `variable` scope
### What happened?

```sh
# Source xprofile
[ -f /etc/xprofile ] && . /etc/xprofile
[ -f ~/.xprofile ] && . ~/.xprofile

# source xinitrc.d files
if [ -d /etc/X11/xinit/xinitrc.d ] ; then
 for f in etc/X11/xinit/xinitrc.d/?*.sh ; do
  [ -x "$f" ] && . "$f"
 done
 unset f
fi
```

deathaxe:
Fixes #3804

This commit ...

1. scopes unquoted identifiers in test expressions `string.unquoted` instead of `variable.parameter`.
2. removes pattern matching (`=~ ...`) from `test ...` command as the command supports basic string comparisons only.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -723,10 +723,10 @@ contexts:
     - match: \(
       scope: punctuation.section.group.begin.shell
       push: cmd-test-args-group-body
-    - match: ([=!]=|=~)\s*
-      captures:
-        1: keyword.operator.comparison.shell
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):
Nice.

jfcherng (approved):

PR #3809, opened 2023-07-18
[PHP] PHP 8.3 new syntax

jfcherng:
PHP 8.3 is scheduled to be 

- first alpha on `Jun 08 2023`
- feature freeze on `Jul 18 2023`
- stable release on `Nov 23 2023`

**References**:

- PHP RFCs: https://wiki.php.net/rfc#php_83
- PHP milestone: https://github.com/php/php-src/milestone/5?closed=1
- PHP release schedule: https://wiki.php.net/todo/php83

### Need new parsing paths

- [x] [PHP RFC: Readonly amendments](https://wiki.php.net/rfc/readonly_amendments)
- [x] [PHP RFC: Typed class constants](https://wiki.php.net/rfc/typed_class_constants) (by @deathaxe)

- [X] [PHP RFC: Marking overridden methods (#[\Override])](https://wiki.php.net/rfc/marking_overriden_methods)
- [X] [PHP RFC: mb_str_pad](https://wiki.php.net/rfc/mb_str_pad)
- [X] [PHP RFC: json_validate](https://wiki.php.net/rfc/json_validate)

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -134,9 +134,9 @@ variables:
     | MongoGridFSFile | MongoGridfsFile | MongoId | MongoInt32 | MongoInt64 | MongoRegex
     | MongoTimestamp | MultipleIterator | NoRewindIterator | Normalizer | NumberFormatter
     | OCI-Collection | OCI-Lob | OutOfBoundsException | OutOfRangeException | OuterIterator
-    | OverflowException | PDO | PDOStatement | ParentIterator | ParseError | Phar | PharData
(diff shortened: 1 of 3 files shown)


jfcherng:
Not sure whether I missing something but these are what I found.

deathaxe:
Not sure if that's a an older issue, but https://wiki.php.net/rfc/typed_class_constants may need some attention.

The `const` modifier seems to break type highlighting.

```
class Test {
    use T;
 
    private const int A = 1;
    public  mixed B = 1;
    public const int C = 1;
    public const Foo|Stringable|null D = null;
 
    // This is illegal since the type cannot change when T::E is redefined
    public const array E = [];
}
```

grafik

deathaxe (approved):

michaelblyons (approved):

PR #3814, opened 2023-08-02
[Python] add more sql indicators

keith-hall:
`MERGE INTO` is ANSI SQL: https://en.wikipedia.org/wiki/Merge_(SQL)#Usage `BEGIN` works for MySQL and TSQL


deathaxe:
Could you add the change to _PHP Source_ as well?

keith-hall:
> Could you add the change to _PHP Source_ as well?

Done

deathaxe (approved):

jfcherng (approved):

michaelblyons (approved):

PR #3816, opened 2023-08-02
[Regular Expressions] Add support for quoted find path patterns

deathaxe:
This PR refactors _File Pattern.sublime-syntax_ to 

1. add support for quoted path patterns (e.g.: `"//my/path with,comma/*"`)
2. distinguish path patterns and file patterns using meta scopes, so color schemes can choose separate highlighting for wildcards depending on their function, despite using the same `constant.other.wildcard` scope in both.

   - path patterns, which contain separators such as `/` or `\`, are scoped `meta.path.file-pattern`
   - file patterns without path separators such as `*.py` are scoped `meta.file.file-pattern`

3. moves syntax tests to a sub directory


michaelblyons:
I have been misreading the documentation for years. 🤦‍♂️

jfcherng (approved):

michaelblyons (approved):

PR #3818, opened 2023-08-03
[C] Add typeof[_unqual]() handling

nabijaczleweli:
Test screenshot:

All three forms are in use in the wild. The __-prefixed versions two have been available for a long time. The plain `typeof()` and `typeof_unqual()` is [new in C23](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3096.pdf).

Derived from the C++ decltype code.

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -48,7 +48,7 @@ variables:
   type_qualifier: 'const|volatile'
   compiler_directive: 'inline|restrict|__restrict__|__restrict'
   modifiers: '{{storage_classes}}|{{type_qualifier}}|{{compiler_directive}}'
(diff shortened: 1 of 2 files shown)


deathaxe (approved):

nabijaczleweli:
Also added `|typeof_unqual`, which is also new in C23 – https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3096.pdf.

deathaxe:
Note that current C/C++ syntaxes don't support any of the more recent language features.

nabijaczleweli:
But even as a modern C and C++ user so far I haven't really run into issues, except for this.

deathaxe (approved):

michaelblyons (approved):

PR #3827, opened 2023-08-21
[ShellScript] Improve test expression and pattern matching highlighting

### LINKED ISSUE #3009: [ShellScript] Improve highlighting accuracy of test expressions
According to the bash docs...
1. the `test ...` command doesn't support Perl like pattern matching via `=~` operator.
2. supported features / allowed expressions differ between `[ ... ]`  and `[[ ... ]]` operators.
3. highlighting would probably need to be distinguished between string/pattern matching and arithmetic comparisons
   * In case of presence of `-eq` or equivalent operator the test expression needs to be treated as arithmetic expression like within `(( ... ))`
   * otherwise tests compare strings or do pattern matching
   * or perform unary file tests

All different kinds of test statements share the same quite naive implementation at the moment, which causes some inaccuracies in some edge cases.

### LINKED ISSUE #3825: [Bash] Wrong highlighting for nested group in regex
### What happened?

```bash
if ! [[ ${PY_FULL_VER} =~ ^[0-9]+\.[0-9]+\.[0-9]+(([ab]|rc)[0-9]+)?$ ]]; then
#                                                 ^ punctuation.definition.group.begin.regexp.shell
#                                                                ^ punctuation.definition.group.end.regexp.shell
    echo "[ERROR] Invalid Python version." && exit 1
fi
```

image

deathaxe:
Resolves #3009
Fixes #3825

This PR ...

1. refactors [Pattern Matching](https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html) and implements [Posix Extended Regular Expressions](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html) to improve highlighting accuracy.

2. refactors test expressions to distinguish different feature levels and behaviors of `test ...`, builtin test expressions `[ ... ]` and compound test command `[[ ... ]]`. Only the last one supports pattern matching after `==` or `!=` operators as well as extended regular expressions after `=~` operator.

It is a somewhat larger commit, but mainly due to various "duplications" needed to handle different sorts of bailout requirements, depending on the context pattern matching appears in.

Regular Expressions have been based on _RegExp (Basic)_. Due to significant simplifications and overlapping wiht Bash's pattern matching it finally appeared more suitable to copy relevant code instead of extending the original syntax. It should also help to maintain syntax highlighting in syntaxes such as Makefile, which inherit Bash syntax.

#### Benchmarks

Based on various real world shell scripts using pattern matching to some degree, no difference in parsing performance has been encountered.

#### Notes

Patterns are scoped `meta.string.regexp` instead of `meta.pattern`, but neither `source.regexp` nor `string.[unquoted|regexp]` is applied. Pattern expansions can appear everywhere, but not everything is or should be scoped string. So patterns would be highlighted different, which may not be desirable. Depending on whether patterns appear in strings or not, additional contexts would be needed to correctly clear string scopes in interpolations, which would add some extra complexity.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -700,10 +700,10 @@ contexts:
   cmd-test:
     - match: \[\[(?=\s)
       scope: support.function.test.begin.shell
-      push: compound-test-expression-body
+      push: compound-test-body
     - match: \[(?=\s)
       scope: support.function.test.begin.shell
-      push: builtin-test-expression-body
+      push: builtin-test-body
     - match: test{{cmd_break}}
       scope:
         meta.function-call.identifier.shell
@@ -720,7 +720,7 @@ contexts:
     - include: test-expression-common
     - include: eoc-pop
 
-  builtin-test-expression-body:
+  builtin-test-body:
     - meta_scope: meta.conditional.shell
     - match: '{{wspace}}(\])'
       captures:
@@ -732,7 +732,7 @@ contexts:
       scope: keyword.operator.comparison.shell
     - include: test-expression-common
 
-  compound-test-expression-body:
+  compound-test-body:
     - meta_scope: meta.conditional.shell
     - match: '{{wspace}}(\]\])'
       captures:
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

michaelblyons:
I _might_ have found a bug. I remembered writing a script to parse Git status lines, (will edit in the regex or a simpler version) where I discovered that I had to define it as a variable and do an `[[ $text =~ $pattern ]]` to match it.

I ran some experiments with your code with the regex back in the test (`[[ $text =~ \#\# ... ]]`), which _does_ correctly end the regex highlighting immediately after the first space.

However, it doesn't end for spaces inside capture groups (`[[ $text =~ \#\#( )... ]]`), and I suspect that it should. I didn't get it to match the expected expression, but maybe I didn't tinker enough.

deathaxe:
From my experiments, whitspace doesn't terminate patterns within quotes and capture groups. That's actually one reason for all the variants of pattern contexts of this PR.

```sh
$ if [[ " " =~ ^( )$ ]]; then echo yes; fi
yes
```

michaelblyons:
You're right. I don't know what I was doing differently before.

Issue #2575, opened 2020-11-06
[RFC] Is it useful to mark “the full path” of a function call as meta.function-call?


Thom1729:
From the [Scope Naming Guidelines](https://www.sublimetext.com/docs/scope_naming.html#meta):

> Function names, including the full path, and all parameters should receive the following scope. The name of the function or method should be `variable.function`, unless the function is scoped with `support.function`.
> - `meta.function-call`

According to the guidelines, the entirety of the following line of Python should be marked `meta.function-call`:

```py
Foo.bar.baz(x, y, z=42);
```

There are three things I don't like about this.

### 1. It's not useful.

In the above example, a high-quality syntax will apply plenty of useful scopes:

- `Foo` might be scoped as a variable, a constant, or a class, depending on the language.
- `bar` will be scoped as a property access.
- `baz` will be scoped as a function.
- The entire argument list will probably be scoped in some way.
- The commas in the argument list will be scoped as punctuation.
- `z` will be scoped `variable.parameter`.

All of these scopes may be useful for syntax highlighting and even for basic code intelligence features. But it's not clear that scoping the entire line `meta.function-call` adds anything of value. It doesn't make sense to highlight the entire line as a whole, but it's not necessary for highlighting the individual pieces (because more detailed scopes exist). And it's probably useless for code intelligence because of (2) and (3).

### 2. It's ambiguous.

The issue here is that “the full path” is supposed to be marked `meta.function-call`. It's clear how this applies to simple cases like the Python example above. It's not clear how it applies in more complex cases. In the following examples, what constitutes “the full path”?

- `foo().bar()`
- `(arbitrarily + complex - expression).bar()`
- `foo[something].bar()`
- `foo.bar()()`

These questions surely have answers — perhaps many answers. It's hard to come up with an obvious rule that would apply sensibly across a variety of languages.

### 3. It's really hard to implement.

Detecting whether a mere identifier is the name of a function being called can be tricky. In most languages, there may be a newline between the identifier and the open paren. In many languages, there may be type parameters between them as well. Nearly all syntaxes I've seen cope with this on a best-effort basis using a lookahead before the identifier. In TypeScript, it has to be exact or highlighting can break, and so that definition uses a fairly complex implementation using branching that's still not entirely satisfactory.

[...]

PR #3830, opened 2023-08-24
[JavaScript] Don't try to scope entire call paths

### LINKED ISSUE #3790: [TypeScript] Method calls aren't scoped when they have type arguments
In the following...

Screenshot 2023-06-11 at 12 05 39

* `foo` is scoped as
  * source.ts
  * meta.function-call.js
  * variable.function.js
* `bar` is scoped as
  * source.ts
  * variable.function.js
* `baz` is scoped as
  * source.ts
  * meta.function-call.method.js
  * variable.function.js
* `qux` is scoped as
  * source.ts
  * meta.property.object.js

The issue is that the `qux` identifier should have a scope indicating that it is a method/function being called, like the others do.

(ST 4143)

Thom1729:
Fix #3790.

The scope naming guidelines recommend giving e.g. all of `foo.bar.baz()` the scope `meta.function-call`. For reasons discussed in #3790 and in #2575, this is quite difficult to implement in a reliable manner. Currently, the JavaScript syntax definition tries to do this using lookaheads, but it is far from reliable, leading to observable highlighting bugs such as #3790.

This PR rips out the existing code scoping function call paths and instead just scopes the function/method identifier and arguments. This should be sufficient for any of the actual use cases discussed in the linked issues while also being much more reliable. Notes:

- It's still not *perfectly* reliable, because we're still detecting the function parens (and in TS, type arguments) via a lookahead, but at least it's a simpler lookahead. It should work well in most cases and fail gracefully in other cases.
- Originally, the JavaScript syntax scoped method calls as `meta.function-call.method`. This is not in the guidelines and no other core syntax did this, so I removed it for simplicity.
- Also, the JavaScript syntax did *not* scope the arguments list as the more specific `meta.function-call.arguments`, which is widely done, so I added that in.
- This does result in over-scoping when referring to, but not calling, a builtin method. The solution there, if we care, is probably to slightly restructure the way those support scopes work.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1086,8 +1086,6 @@ contexts:
         - async-arrow-function
         - literal-variable
 
-    - include: literal-call
-
     - match: (?={{possible_arrow_function_begin}})
       pop: 1
       branch_point: arrow-function
@@ -1955,7 +1953,7 @@ contexts:
         1: punctuation.accessor.js
         2: punctuation.section.group.begin.js
       set:
-        - meta_scope: meta.group.js
+        - meta_scope: meta.function-call.arguments.js meta.group.js
         - match: \)
           scope: punctuation.section.group.end.js
           pop: 1
@@ -1986,14 +1984,7 @@ contexts:
 
     - match: '{{dot_accessor}}'
       scope: punctuation.accessor.js
-      push:
-        - match: '(?={{identifier_name}}\s*(?:{{dot_accessor}})?\()'
-          set:
-            - call-method-meta
-            - function-call-arguments
-            - call-path
-            - object-property
-        - include: object-property
+      push: object-property
 
   literal-number:
     # floats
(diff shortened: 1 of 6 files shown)


Thom1729:
Fixed the related failure in the PHP syntax test.

deathaxe:
The current way of applying `meta.function-call` would cause a gap in scope if argument list does not directly follow a function name.

```js
    bar (baz),
//  ^^^ meta.function-call
//     ^ - meta.function-call
//      ^^^^^ meta.function-call.arguments
```

JavaScript/JavaScript.sublime-syntax
```diff
@@ -2066,52 +2057,22 @@ contexts:
...
 
   call-method-name:
     - include: support-property
     - match: '{{identifier_name}}'
-      scope: variable.function.js
+      scope: meta.function-call.js variable.function.js
       pop: 1
     - match: '(#){{identifier_name}}'
-      scope: variable.function.js
+      scope: meta.function-call.js variable.function.js
```
deathaxe, line 2075: Would it make sense to merge the two patterns

```yml
    - match: '{{identifier_name}}'
      scope: meta.function-call.js variable.function.js
      pop: 1
    - match: '(#){{identifier_name}}'
      scope: meta.function-call.js variable.function.js
      pop: 1
```

to

```yml
    - match: '(#)?{{identifier_name}}'
      scope: meta.function-call.js variable.function.js
      captures:
        1: punctuation.definition.js
      pop: 1
```

?

Thom1729: Agreed, done.

Thom1729:
> The current way of applying `meta.function-call` would cause a gap in scope if argument list does not directly follow a function name.

Yeah, I was wondering if anyone would notice or care. I'll count that as a “yes”. 😉

I also found a really minor bug. In `foo?.()`, the `?.` gets the `.arguments` scope, which is not correct. I'll tweak it a bit, probably cut a couple of contexts that aren't needed anymore.

Thom1729:
Should be all good now. The part that gets the whitespace after a variable name is a bit of a kludge, but it was either that or do something with the arrow function detection code, and nah.

deathaxe (approved):

michaelblyons (approved):

ratijas:
This required me to update my downstream QML syntax once again. Which is fine in itself. But I do wonder how to keep CI green when it's running against multiple versions of Sublime Text, so obviously packages are going to be mix-versioned? For now I removed JS-specific scopes from the tests, but ideally it would be good to have them in-place as well.

deathaxe:
In case core syntax changes effect inheriting 3rd-party packages those need to ship ST build specific releases by either leave older package versions alone or maintain multiple branches in parallel.

[Less](https://github.com/SublimeText/Less/releases/tag/3.0.3), Liquid and other packages are doing so, already. They maintain tags such as `<st_build>-<semver>` (e.g.: `4107-1.1.0`, `4134-1.1.0`) to ship dedicated releases with the prefix denoting least required ST build.

PR #3831, opened 2023-08-26
[CSS] Fix nested at-rules highlighting

### LINKED ISSUE #3820: [CSS] Support nesting properties directly under at-rules
### What happened?

When nesting at-rules inside common style rules, syntax highlighting breaks inside the at-rule block unless you start with a new style rule. This is not according to spec since all property declarations applies to the closest ancestor selector, regardless if nested inside an at-rule or not.

Nesting properties like that of course only applies to at-rules that allows it normally, like `@layer`, `@media`, `@container` etc…

See specification: https://www.w3.org/TR/css-nesting-1/#conditionals

```CSS
button {
  @layer defaults {
    /* Syntax highlighting currently doesn’t work for rules nested directly under an at-rule. */
    all: unset;
    offset: revert;
    cursor: pointer;
    padding: 1ex;
    font: inherit;
    line-height: 1cap;
    text-wrap: nowrap;
  }
  @layer components {
    &.report-btn {
      /* Syntax highlighting works as expected when starting a new style rule, even when nested inside an at-rule. */
      color: white;
      background-color: color(display-p3 0.7 0 0);
      border: 2px solid;
      border-radius: 4vmin;
    }
  }
}
```

deathaxe:
Fixes #3820

This PR...

1. adds separate contexts for nested at-rules.

   Global at-rules push [&lt;style-blocks&gt;](https://www.w3.org/TR/css-syntax-3/#typedef-style-block) on stack while nested at-rules push [&lt;rule-lists&gt;](https://www.w3.org/TR/css-syntax-3/#typedef-rule-list) represented by `property-lists` context.

2. refactors at-rules to make use of multi-push statements so more general purpose contexts can be created and re-used more often, which reduces syntax cache size and should limit amount of special purpose contexts, which might need to be addressed by inheriting 3rd-party syntaxes.

3. replaces `at-rule-end` by `else-pop` to follow a fail as early as possible approach, which may help to bail-out from statements more reliably.

   e.g. at-rules with missing semi-colon suffered leaking into following at-rules or selectors, before.

4. applies lazier parsing of various at-rules' property-lists in order to reduce complexity/special casing and increase re-using existing contexts.

   While it doesn't improve performance, syntax cache size is reduced and re-using existing contexts is increased.

#### Notes:

Overall parsing performance keeps unchanged.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -275,15 +275,6 @@ variables:
   counter_system_constants: |-
     \b(?xi: cyclic | numeric | alphabetic | symbolic | additive | fixed ){{break}}
 
-  # @font-face Property Names
-  # https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face
-  font_face_property_names: |-
-    \b(?xi:
-      ascent-override | descent-override | font-display | (?# font-family | ) src
-    | font-feature-settings | font-variation-settings | font-stretch | font-style
-    | font-weight | font-variant | line-gap-override | size-adjust | unicode-range
-    ){{break}}
-
   # @page Margin Names
   # https://developer.mozilla.org/en-US/docs/Web/CSS/@page
   page_margin_names: |-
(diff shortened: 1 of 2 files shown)


gnat:
Just encountered this too. Thanks for getting right on it @deathaxe Looking forward to the merge!

keith-hall (approved):

PR #3836, opened 2023-09-02
[JavaScript] Support explicit resource management proposal (and rewrite for loops in the process)

Thom1729:
[This feature](https://github.com/tc39/proposal-explicit-resource-management) is currently in Stage 3 of the TC39 process. In addition, it is part of [TypeScript 5.2](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management), which was released last week.

The lookaheads are exact (unless I made a mistake). Since `using` is not a reserved word, the proposal spec has to restrict line endings between `using` and the variable name to ensure backward-compatibility.

As part of this PR, I rewrote `for` loop handling. JavaScript basically has two different kinds of for loops: C-style (e.g. `for (let i = 0; i < len; i++) {}`) and `in`/`of` (e.g. `for (const item of list) {}`). The existing implementation predates branching, so it tries to handle both cases simultaneously. This works surprisingly well in practice, but `using` would have made a mess of it. This PR uses branching to try to parse the loop condition as in/of, and falls back to C-style if that fails. It's more verbose, but it should be less confusing, and as a bonus it correctly handles the case where the binding of an `in`/`of` loop is an arbitrary left expression (why!?).

The only known bug is that `for (using of in obj) {}` is highlighted slightly wrong. Basically the only relevant syntactic difference between `in` and `of` loops is a dumb special case related to this. I could fix it, but that would likely require splitting the handling of `in` and `of` loops, which I think is not worth the tiny benefit.

The PR also incidentally removes a couple of unwanted `.js` suffixes in the tests.

### DIFF
--- a/JavaScript/tests/syntax_test_js_control.js
+++ b/JavaScript/tests/syntax_test_js_control.js
@@ -163,6 +163,21 @@
 //      ^^^^^^^^^^^ meta.group
 //         ^^ keyword.operator.word
 
+    for (x.y.z of list) {}
+//  ^^^^^^^^^^^^^^^^^^^^^^ meta.for
+//  ^^^ keyword.control.loop.for
+//      ^^^^^^^^^^^^^^^ meta.group
+//      ^ punctuation.section.group.begin
+//       ^ variable.other.readwrite
+//        ^ punctuation.accessor
+//         ^ meta.property.object
+//          ^ punctuation.accessor
+//           ^ meta.property.object
+//             ^^ keyword.operator.word
+//                ^^^^ variable.other.readwrite
+//                    ^ punctuation.section.group.end
+//                      ^^ meta.block
+
     for await (const x of list) {}
 //  ^^^ keyword.control.loop.for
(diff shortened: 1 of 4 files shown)


JavaScript/tests/syntax_test_js_control.js
```diff
@@ -163,10 +163,119 @@
...
+//              ^^^^^ keyword.declaration
+//        ^^^^^ keyword.declaration
+//              ^^^^^ keyword.declaration
+//                    ^ meta.binding.name variable.other.readwrite
+//                      ^^ keyword.operator.word
+//                         ^^^^ variable.other.readwrite
+//                              ^ punctuation.section.group.end
+//                                ^^ meta.block
+
+    for ( await using of of list ) {}
```
michaelblyons, line 212: Yuck. Truly insane.

michaelblyons (approved):

deathaxe (approved):

PR #3837, opened 2023-09-03
[TypeScript] Implement labeled tuple members

Thom1729:
E.g. `type Rational = [numerator: bigint, denominator: bigint];`

Introduced in [TypeScript 4.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements) more than three years ago, but documented nowhere outside those release notes (plus the recent [5.2 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#named-and-anonymous-tuple-elements)).

I used the scope `meta.mapping.key` for the labels, which doesn't seem quite right but I don't know what would be more right.

The implementation uses branching because a lookahead would have to span multiple lines. I initially intended to use a cover grammar approach, but that would be a mess if the label were the name of a type operator (which you would think wouldn't be allowed, but here we are).

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -925,14 +925,42 @@ contexts:
           scope: punctuation.section.sequence.end.js
           pop: 1
         - include: comma-separator
-        - match: \.\.\.
-          scope: keyword.operator.spread.js
         - match: (?=\S)
           push:
             - ts-type-annotation-optional
             - ts-type-expression-end
             - ts-type-expression-end-no-line-terminator
             - ts-type-expression-begin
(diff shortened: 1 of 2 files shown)


keith-hall:
C# and Python support named tuples, it may make sense to check what scopes they use

deathaxe:
Sorry, for directly pushing. I wanted to create a PR to this PR, but Sublime Merge decided to not give me an option to push it to my local repo, but decided to directly push it here.

Thom1729:
LGTM.

deathaxe (approved):

michaelblyons (approved):

PR #3838, opened 2023-09-03
[TypeScript] Standardize on .js suffix

Thom1729:
The TypeScript syntax uses the `.js` suffix, but there were a couple of places where a `.ts` slipped through.

Also strip the `.js` suffix from the tests, as is convention.

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -180,12 +180,12 @@ contexts:
   import-brace:
     - meta_prepend: true
     - match: type{{identifier_break}}
-      scope: keyword.control.import-export.ts
+      scope: keyword.control.import-export.js
 
   export-brace:
     - meta_prepend: true
     - match: type{{identifier_break}}
-      scope: keyword.control.import-export.ts
+      scope: keyword.control.import-export.js
 
   property-access:
     - meta_prepend: true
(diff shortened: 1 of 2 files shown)


keith-hall:
> The TypeScript syntax uses the `.js` suffix,

Why? JS != TS... Normally it makes sense when inheriting syntaxes use a separate suffix which would match their base scope suffix, then it's clear whether a token is from the original syntax or the extended one etc.

deathaxe:
Actually TypeScript, JSX and TSX have followed the route of using `.js` suffix from beginning on. Hence this PR is just consequently maintaining scope naming consistency.

It doesn't probably make sense to enforce unique suffixes as it may quickly end up looking like wild mess, especially if brackets or other enclosing punctuations/tokens are re-used from between syntaxes. In such cases it's just better to decide for a generic suffix and re-use it everywhere.

_#3575 may be a good example for wildly mixed brace scopes, sometimes even not balanced._

Hence distinct suffixes may make sense only, if completely unique features (statements or expressions) are to be emphasized. They should not be used just to find the source of a pattern.

keith-hall (approved):

Thom1729:
↑ What @deathaxe said.

In principle, I think that we could identify TS-specific tokens and give them the `.ts` suffix. In practice, I think that this would be annoying to maintain and not very useful.

deathaxe (approved):

PR #3840, opened 2023-09-16
[Go] Assign .tmpl extension to Go templates

deathaxe:
Go templates use `*.tmpl` file extension by default according to https://docs.gomplate.ca/usage/.

### DIFF
--- a/Go/HTML (Go).sublime-syntax
+++ b/Go/HTML (Go).sublime-syntax
@@ -10,6 +10,7 @@ extends: Packages/HTML/HTML.sublime-syntax
 file_extensions:
   - gohtml
   - go.html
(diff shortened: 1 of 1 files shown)


michaelblyons (approved):
There is less non-Go use of this extension than I expected.

FichteFoll (approved):
It looks like it is used in [some projects](https://github.com/search?q=path%3A*.tmpl&type=code) outside of Go, most notably ruby, but most of the results are just forks of the ruby repo. Another popular repo whose forks frequently occur is tensorflow.

Either way, it's not like there's anything else one could reasonably suggest for such a generic extension, it seems, and Go does seem to have the majority here.

PR #3847, opened 2023-09-23
[TypeScript] Better highlighting while typing declarations

Thom1729:
Highlighting gets wonky while typing e.g.:

```ts
interface // cursor position
class Foo {}
```

This PR fixes that by adding a couple of else-pops, and by ensuring that only valid (i.e. non-reserved) identifiers are matched as the names of interfaces, type aliases, and namespaces.


michaelblyons (approved):

deathaxe (approved):
To ensure reserved words are not matched as identifiers is always a win.

PR #3849, opened 2023-09-29
[JavaScript] [TypeScript] Clean up export code and support more cases

Thom1729:
The existing code handling export declarations was weirdly overbroad, explicitly handling cases that aren't proper syntax. (Probably this is the legacy of copy-pasting from the import declaration code.) On the other hand, it failed to support at least one case involving strings in place of identifiers. In addition, `export type {}` wan't properly supported in TypeScript, which is how I started on this.

This PR fixes the known export bugs/missing features and generally cleans up the code. It also removes a test that's invalid syntax, and fixes the semicolon scoping on a test that wasn't correct.

I should probably give the import declaration code a good cleaning as well, but I'll put that off until after this PR.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -257,13 +257,6 @@ contexts:
       fail: import-statement
     - include: else-pop
 
-  export-statement:
-    - match: export{{identifier_break}}
-      scope: keyword.control.import-export.js
-      set:
-        - export-meta
-        - export-extended
-
   import-meta:
     - meta_include_prototype: false
     - meta_scope: meta.import.js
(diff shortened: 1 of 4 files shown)


JavaScript/JavaScript.sublime-syntax
```diff
@@ -359,28 +361,15 @@ contexts:
...
 
   export-item:
     - match: '\{'
       scope: punctuation.section.block.begin.js
       set: export-brace
-    - match: '{{non_reserved_identifier}}'
-      scope: variable.other.readwrite.js
-      pop: 1
     - match: '\*'
       scope: constant.other.js
```
deathaxe, line 371: Does it make sense to apply recently introduced common wildcard scope to all those asterisks?

```suggestion
      scope: constant.other.wildcard.asterisk.js
```

Thom1729: I'm sure it would — I just now read that RFC. Is it meant to apply to wildcard imports? That seems reasonable to me.

It would also be an easy change to apply in a standalone PR, or in an import-focused followup. I'm thinking in that PR I might re-consolidate some of the contexts that were split in the past.

deathaxe: Wildcard scopes are meant for universal usage without any restrictions to certain kinds of expressions. Java also scopes wildcard imports this way.

I am fine with addressing that in a separate PR.

deathaxe (approved):

michaelblyons (approved):

PR #3855, opened 2023-10-19
[C/C++/ObjC/ObjC++] Fix javadoc comment word wrap

BenjaminSchaaf:
This was broken by e4e5273d8ca764ebcd79b2ecfbed592036944265

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -106,6 +106,9 @@ contexts:
     - match: \*/
       scope: punctuation.definition.comment.end.c
       pop: true
+    - match: ^\s*(\*)(?!\**/)
+      captures:
+        1: punctuation.definition.comment.c
 
   line-comments:
     - match: ^(//) =\s*(.*?)\s*=\s*$\n?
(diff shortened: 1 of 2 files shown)


deathaxe:
I can remember discussions about when leading asterisks are to be treated as punctuation. Especially Mitranim had some objections about it, IIRC.

The arguments were about JavaDoc always requiring `/**` and stripping leading asterisks, while normal block comments beginning with `/*` wouldn't treat leading asterisks special.

Go therefore even has an indentation based heuristic to decide when to treat them punctuation. 

https://github.com/sublimehq/Packages/blob/92b4c0bdafaa859621539602a0952b8edcc591c0/Go/Go.sublime-syntax#L283-L288

JavaScript also added this kind of separation in f850f0a8787e50ad21b9f134c469e673f1bd4eb7. It didn't scope those at all before.

At least it not being present in normal block comments is not broken but by intent.

I have no objections about it. Actually scope them `punctuation` in normal block comments of Java as well.

deathaxe (approved):

michaelblyons (approved):

FichteFoll (approved):

PR #3865, opened 2023-11-05
[Batch File] Tweak comment collapsing

mataha:
This commit introduces two changes:

 - `@rem` (silenced `rem`) and `:::` (safe equivalent of `::`) have been added
 - `rem` and its silent equivalent `@rem` now use `TM_COMMENT_CASE_INSENSITIVE` introduced in Sublime Text 4153, which allows collapsing comments casefold

### DIFF
--- a/Batch File/Comments.tmPreferences
+++ b/Batch File/Comments.tmPreferences
@@ -15,9 +15,33 @@
 			</dict>
 			<dict>
 				<key>name</key>
+				<string>TM_COMMENT_CASE_INSENSITIVE</string>
+				<key>value</key>
+				<string>yes</string>
+			</dict>
+			<dict>
+				<key>name</key>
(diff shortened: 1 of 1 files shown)


FichteFoll:
> these don't seem to have been added there nor to the tests

Because they're not needed. Now, one could argue that all colons in `:::` should get the `punctuation.definition.comment` scope (currently only the first 2), but that's orthogonal to this PR which is mostly concerned about the `toggle_comment` command.

mataha:
> > these don't seem to have been added there nor to the tests
> 
> Because they're not needed.

Exactly.

> Now, one could argue that all colons in `:::` should get the `punctuation.definition.comment` scope (currently only the first 2), but that's orthogonal to this PR which is mostly concerned about the `toggle_comment` command.

I've addressed this in #3866.

deathaxe (approved):

PR #3868, opened 2023-11-16
[Go] fix backtick strings inside function calls

keith-hall:
Previously the closing paren would be erroneously scoped as illegal

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -916,6 +916,8 @@ contexts:
       pop: true
     - match: '{{predeclared_type}}'
       scope: support.type.go
+    - match: (?=;)
+      pop: 1
     - include: match-any
 
   match-keyword-chan:
--- a/Go/tests/syntax_test_go.go
+++ b/Go/tests/syntax_test_go.go
@@ -4961,7 +4961,7 @@ by accident, but if necessary, such support could be sacrificed.
 
     break
 //  ^^^^^ keyword.control.flow.break.go
-    case
+    case ;
 //  ^^^^ keyword.control.conditional.case.go
     continue
 //  ^^^^^^^^ keyword.control.flow.continue.go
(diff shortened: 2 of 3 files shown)


michaelblyons:
Is there a negative scope test that would check for the unpopped context without the SQL demonstration?

keith-hall:
not sure if I entirely understood what you were asking but I pushed a commit to add a test for incomplete case statements, to ensure they pop off the stack when a semi-colon is reached.

Go/tests/syntax_test_go.go
```diff
@@ -4961,8 +4961,12 @@ by accident, but if necessary, such support could be sacrificed.
 
     break
 //  ^^^^^ keyword.control.flow.break.go
-    case
+    case no_colon_here_while_user_is_typing
 //  ^^^^ keyword.control.conditional.case.go
+        ; a := b
+//      ^ punctuation.terminator
+//          ^ - punctuation.separator.case-statement
+//          ^^ keyword.operator.assignment
```
deathaxe, line 4969: I'd leave those lines dedicated to keyword tests.

Line 5629ff. seems more appropriate to add higher level (incomplete) expression tests.

Go/Go.sublime-syntax
```diff
@@ -916,6 +916,8 @@ contexts:
       pop: true
     - match: '{{predeclared_type}}'
       scope: support.type.go
+    - match: (?=;)
+      pop: 1
```
deathaxe, line 920: If improving support for incomplete statements is the goal, it may also make sense to pop on stray closing brackets and keywords, which are not valid in expressions, such as `case`, `if`, ...

Go/tests/syntax_test_go.go
keith-hall: moved the test :)

Go/Go.sublime-syntax
keith-hall: maybe let's leave that for a future PR then, it sounds quite a bit more involved :) if needed, I can drop the commit which fixed the `case` statement pop at semi-colon from this PR.

deathaxe: It is not an error to have it there, even if more cases may be relevant. It's fine to address them later.

deathaxe (approved):

michaelblyons (approved):

PR #3869, opened 2023-11-17
[CSS] Add support for at-scope rule

### LINKED ISSUE #3867: [CSS] Add highlighting for `@scope` at-rule
### What happened?

The current CSS.sublime-syntax breaks `@scope` at-rule highlighting

```css
@scope (.light-scheme) {
  a {
    color: red;
  }
}
```
Screenshot 2023-11-12 at 4 29 08 PM

my current color-scheme & theme are `Bass Blackcomb` & `Treble Adaptive` respectively

deathaxe:
Resolves #3867

This commit implements support for @scope rules.

see: https://developer.mozilla.org/en-US/docs/Web/CSS/@scope

Notes: At the time of writing it is an experimental feature not supported by all major browsers.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -387,6 +387,7 @@ contexts:
     - include: at-page
     - include: at-page-margin
     - include: at-property
+    - include: at-scope
     - include: at-scroll-timeline
     - include: at-other
 
@@ -406,6 +407,7 @@ contexts:
     - include: at-page
     - include: at-page-margin
     - include: at-property
+    - include: at-scope
     - include: at-scroll-timeline
     - include: at-other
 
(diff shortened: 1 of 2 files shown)


thecodechef:
This is Screenshot of your syntax tests when i see it from my end
Screenshot 2023-11-17 at 11 16 00 AM
> the `.limit > *` part is highlighted but the `.class` is not thats what i mean by that

deathaxe:
How is Github's syntax not supporting this new experimental feature related with the fix for ST's syntax definition?

grafik

thecodechef:
Because when I also use it in the latest stable build I get the same behavior

deathaxe:
CSS has significanty changed since ST4152. Hence, you'd need to replace the whole CSS.sublime-syntax file. It may however have effects on 3rd-party syntaxes such as Less as those may need some tweaks to follow these changes.

thecodechef:
Which is what I did I extracted Packages/CSS using PackageResourcwViewer then copied your version and replaced the other

deathaxe:
As all CI tests (against ST4143 and ST4165) are passing it must be an issue on your end.

thecodechef:
That may be true, but tests can always send false positives. I have been using Sublime Text ever since Sublime Text 2 I'm not new. & furthermore it was when I changed your code thats it started working like it should

CSS/CSS.sublime-syntax
```diff
@@ -892,6 +894,32 @@ contexts:
     - include: comments
     - include: else-pop
 
+###[ SCROPE AT-RULE ]##########################################################
```
michaelblyons, line 897: Typo (but no practical effect)

deathaxe:
As test results are equal with what I see, when symlinking this branch into a vanilla ST4152 portable setup's Packages folder, I tend to trust them.

thecodechef:
'''
push:
  - include: at-scope-content-keywords
  - include: at-scope-selector
'''
Is what I meant in that comment

deathaxe:
Note: [@scope](https://developer.mozilla.org/en-US/docs/Web/CSS/@scope#browser_compatibility) is supported by all major browsers and no longer marked experimental.

michaelblyons (approved):

nilaallj:
Any update on this?

deathaxe:
This PR is waiting a 2nd approval by a maintainer to enable merging.

gnat:
Could the maintainers of the CSS syntax just stop highlighting at rules (ex: `@media`, `@scope`) after the first whitespace (` `)  is encountered?

The current behavior of continuing to highlight feels broken- and it'll continue to break as new rules are added to CSS.

PR #3876, opened 2023-11-25
[JSON] Add support for inheritance

deathaxe:
This PR...

1. renames contexts to express popping/non-popping behavior via singular/plural.
2. adds named contexts for nearly everything, so it can be addressed by inherit syntaxes.

... while keeping behavior and highlighting as simple and fast as it is without bloating the package.

required for https://github.com/SublimeText/Liquid/issues/10

The new syntax is inherit in ASP, JSP, PHP, ERB and HAML to support JSON highlighting in templates.

Without this PR syntax highlighting errors occur, within `<script type="application/json">` tags due to lack of interpolation support of JSON syntax, which is already embedded in HTML.sublime-syntax.

### DIFF
--- a/ASP/Embeddings/JSON (for ASP).sublime-syntax
+++ b/ASP/Embeddings/JSON (for ASP).sublime-syntax
@@ -5,25 +5,21 @@
 #
 # It powers VBScript highlighting within ASP tags in embedded CSS.
 #
-#   <style>
+#   <script type="application/json">
 #     <%! ... %>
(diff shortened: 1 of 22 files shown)


deathaxe:
Note, syntax behavior itself hasn't changed beyound switching to `version: 2` and adding named contexts (on purpose).

jfcherng (approved):

michaelblyons (approved):

PR #3882, opened 2023-12-02
[Rust] Add C-string literals

ehuss:
Rust 1.76 has added C-string literal syntax.

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -1269,6 +1269,8 @@ contexts:
     - include: raw-byte-string
     - include: string
     - include: raw-string
+    - include: c-string
+    - include: raw-c-string
 
   chars:
     - include: char
@@ -1383,6 +1385,35 @@ contexts:
 
   raw-string:
     - match: (r)((#*)")
+      captures:
+        1: storage.type.string.rust
+        2: punctuation.definition.string.begin.rust
+      push:
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.double.raw.rust
+        - match: '"\3'
+          scope: punctuation.definition.string.end.rust
+          pop: true
+
+  c-string:
+    - match: '(c)(")'
+      captures:
+        1: storage.type.string.rust
+        2: punctuation.definition.string.begin.rust
+      push:
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.double.rust
+        - match: '"'
+          scope: punctuation.definition.string.end.rust
+          pop: true
+        - match: '{{escaped_byte}}'
+          scope: constant.character.escape.rust
+        - match: '(\\)$'
+          scope: punctuation.separator.continuation.line.rust
+        - include: escaped-char
+
+  raw-c-string:
+    - match: (cr)((#*)")
       captures:
         1: storage.type.string.rust
         2: punctuation.definition.string.begin.rust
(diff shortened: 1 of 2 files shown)


Rust/Rust.sublime-syntax
```diff
@@ -1393,6 +1395,35 @@ contexts:
...
+        2: punctuation.definition.string.begin.rust
+      push:
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.double.rust
+        - match: '"'
+          scope: punctuation.definition.string.end.rust
+          pop: true
+        - match: '{{escaped_byte}}'
+          scope: constant.character.escape.rust
+        - match: '(\\)$'
```
michaelblyons, line 1411: Why capture group if you're not going to use it?

keith-hall: Not sure I'd call a tiny oversight a "tradition". And now is the perfect time to improve things like that - some engines which use `sublime-syntax` files can get a small performance boost from not using unnecessary capture groups.

keith-hall: If some matches end up being duplicated, it might make sense to factor them out into a new context which can be shared/included in the relevant places. It might be the right time to do this for the line continuation and escaped byte matches.

keith-hall (approved):

Rust/Rust.sublime-syntax
```diff
@@ -1393,6 +1398,33 @@ contexts:
...
+        2: punctuation.definition.string.begin.rust
+      push:
+        - meta_include_prototype: false
+        - meta_scope: string.quoted.double.rust
+        - match: '"'
+          scope: punctuation.definition.string.end.rust
+          pop: true
+        - include: escaped-byte
+        - include: line-continuation
+        - include: escaped-char
```
michaelblyons, line 1414: Sorry for maybe dumb question. I don't know Rust. Can a double quote be escaped? If so, this should go above the string end (and add regression tests). If not, please disregard.

ehuss: Double quotes can be escaped via `\"` like other strings.

I added some more tests around escaping.

I'm reluctant to rearrange the rules here. I'm uncertain if there was an explicit reason for the order the rules are specified, like if the order would impact performance (I'm not sure how smart the engines are for combining the rules). Logically it shouldn't matter since there isn't any overlap between these particular rules.

deathaxe: We always have popping end patterns at the beginning, followed by content such as escapes in strings.

As escape pattern starts with `\` there's no chance of escapes being consumed by end pattern by accident. So order doesn't matter.

Rust/Rust.sublime-syntax
```diff
@@ -1393,6 +1398,33 @@ contexts:
           scope: punctuation.definition.string.end.rust
           pop: true
 
+  c-string:
+    - match: '(c)(")'
+      captures:
+        1: storage.type.string.rust
```
deathaxe, line 1404: Well, in Python we don't scope leading modifiers `string.quoted`, which is applied by `meta_scope` here.

deathaxe: Python however seems to be unique with that regard. All others (C, C++, D, R) behave like this PR.

michaelblyons (approved):

PR #3884, opened 2023-12-09
[D] Add short function definition syntax

s-ludwig:
The short function definition syntax was introduced with DMD 2.101.0 and replaces an ordinary block statement function body with a lambda-like construct (`int foo(int x) => x + 1;`).

Due to the structure of the syntax definition file, I'm unsure whether this is complete, but it appears to work fine for the cases that I've seen so far.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -916,6 +916,8 @@ contexts:
   function-definition-after-arguments:
     - meta_scope: meta.function.d
     - include: function-attribute-in
+    - match: '=>'
+      set: [expect-end-of-line, value]
     - match: '='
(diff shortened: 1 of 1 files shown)


s-ludwig:
Corresponding part of the D spec: https://dlang.org/spec/function.html#function-bodies

s-ludwig:
Honestly, I'm just making educated guesses with the changes here, since the structure and some of the syntax of the syntax and test files is not really clear to me. It feels like this will take more time than I can currently spare for this. The two test cases this should cover would be these, though:
```
    int[] map(int[] array) => foo();
    int[] map(T)(int[] array) => foo();
```

I've opened an issue instead for now #3894

PR #3888, opened 2023-12-13
[Python] Add SQL syntax highlighting in f-strings

### LINKED ISSUE #3885: [NOT IMPORTANT] Python f-strings do not highlight as SQL syntax
### Problem description

Снимок экрана 2023-12-13 в 10 56 54 AM

### Preferred solution

f-strings should be highlighted with SQL syntax.

### Alternatives

N/A

### Additional Information

_No response_

deathaxe:
Resolves #3885

This PR proposes to add support for SQL syntax highlghting within f-strings.

I have no strong opinion about adding this. 
The change was easy, so proposing it for thosw who want it.

Note: With #3046 Python's syntax cache grows from about 3MB to 4MB with this PR merged.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -2829,7 +2829,31 @@ contexts:
     - meta_include_prototype: false
     - meta_content_scope: meta.string.python string.quoted.double.block.python
     - include: triple-double-quoted-string-end
+    - include: triple-double-quoted-f-string-syntax
+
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

FichteFoll:
I'd like to see an actual use case of using SQL with an f-string before adding this, considering that it significantly increases the cache size along with #3046 (by 33%!) for what seems like rather little gain.

PR #3892, opened 2023-12-15
[Markdown] Add GitHub Alerts syntax

### LINKED ISSUE #3891: [Markdown] Add support for GFM alert blocks
### What happened?

GitHub recently introduced a new feature for alert blocks in Markdown and the syntax is finalized now:

* [Blogpost](https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/)
* [Docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)

TLDR; the syntax looks like this:
```markdown
> [!NOTE]
> Useful information that users should know, even when skimming content.

> [!TIP]
> Helpful advice for doing things better or more easily.

> [!IMPORTANT]
> Key information users need to know to achieve their goal.

> [!WARNING]
> Urgent info that needs immediate user attention to avoid problems.

> [!CAUTION]
> Advises about risks or negative outcomes of certain actions.
```

Since GFM extensions are included in Sublime's Markdown syntax, it would be nice to have support for the alert blocks.

Some ideas/suggestions for the scope names:

* Note: `markup.quote.markdown markup.info.note.markdown`
* Tip: `markup.quote.markdown markup.info.tip.markdown`
* Important: `markup.quote.markdown markup.info.important.markdown`
* Warning: `markup.quote.markdown markup.warning.markdown`
* Caution: `markup.quote.markdown markup.error.markdown`

From the CommonMark point of view those alert blocks are still regular blockquotes, so I think the new scope parts should be stacked on top of `markup.quote`. This also ensures that the blocks receive the color scheme's standard highlighting for blockquotes if there are no special rules for `markup.info` etc. in the color scheme.

A question would be whether to apply the scopes only to the `[!NOTE]`, etc. part, or to the entire block. I don't have a strong preference about that, but in case it's applied to the entire block, then the `[!NOTE]` part should probably get another dedicated scope like for example `keyword.declaration.note` on top. And the example scopes from above are meant to include the existing `markup.paragraph` in that case, i.e. the final scope could be `markup.quote.markdown markup.info.note.markdown markup.paragraph.markdown` for the text. But I would tend to apply the new scope parts only to the `[!NOTE]` part of the block; then it would resemble how it's rendered on GitHub. On the other hand, the text is part of the alert box, so I guess semantically it makes more sense to apply the scope to the whole blockquote. And this would allow the color scheme for example to apply a background color to the alert block.

deathaxe:
Resolves #3891

This PR proposes to add support for syntax highlighting of GitHub alerts as proposed by #3891.

This PR uses dedicated scopes `markup.heading.alert` and `markup.quote.alert` to avoid unwanted negative impacts on highlighting in conjunction with legacy color schemes. Dedicated highlighting requires to opt-in via new color scheme rules. Otherwise those blocks keep highlighting as normal block quotes except `[!...]` marker being highlighted as normal heading.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -354,6 +354,56 @@ contexts:
 
   block-quotes:
     # https://spec.commonmark.org/0.30/#block-quotes
+    - match: '[ \t]{,3}(>)[ ]?((\[)!CAUTION(\]))'
+      captures:
+        1: punctuation.definition.blockquote.markdown
+        2: markup.heading.alert.caution.markdown
+        3: punctuation.definition.heading.begin.markdown
+        4: punctuation.definition.heading.end.markdown
+      push:
+        - block-quote-caution-meta
+        - block-quote-body
+        - block-quote-punctuation-body
+    - match: '[ \t]{,3}(>)[ ]?((\[)!WARNING(\]))'
+      captures:
+        1: punctuation.definition.blockquote.markdown
+        2: markup.heading.alert.warning.markdown
+        3: punctuation.definition.heading.begin.markdown
+        4: punctuation.definition.heading.end.markdown
+      push:
+        - block-quote-warning-meta
+        - block-quote-body
+        - block-quote-punctuation-body
+    - match: '[ \t]{,3}(>)[ ]?((\[)!IMPORTANT(\]))'
+      captures:
+        1: punctuation.definition.blockquote.markdown
+        2: markup.heading.alert.important.markdown
+        3: punctuation.definition.heading.begin.markdown
+        4: punctuation.definition.heading.end.markdown
+      push:
+        - block-quote-important-meta
+        - block-quote-body
+        - block-quote-punctuation-body
+    - match: '[ \t]{,3}(>)[ ]?((\[)!NOTE(\]))'
+      captures:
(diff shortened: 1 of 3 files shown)


jwortmann:
`markup.quote.alert` looks fine to me, but if you want to use `markup.heading.alert` for the alert type, then this scope should at least be excluded from the symbol list.

deathaxe:
Maybe `.markdown-gfm`, which all the other _Github Favoured Markdown_ constructs use. But actually I see barely use for that and would even rather drop all of those distinctions, especially as most (not all) of them have been adopted by CommonMark as well.

IIRC, only remaining GFM specific expressions are tables and now github alerts.

Any other admonation syntax could use exactly the same scopes. Not sure if anyone would want to highlight them different.

FichteFoll (approved):

michaelblyons (approved):

jwortmann:
Just a small discrepancy that I noticed:

If there is an empty line within an alert block, the alert scope is not applied anymore. But on GitHub, an empty line doesn't close the alert block:

alert

Not sure if this was intentionally made different in the syntax to keep the implementation simple, or if it was just missed.

PR #3896, opened 2023-12-27
[Lisp] Rewrite Syntax

### LINKED ISSUE #1968: [Lisp] Highlighting seems very gone

deathaxe:
Resolves #1968 

Supersedes #2387
Supersedes #2312

Inspired by #2387 

This PR actually started with #2387 but ended up being a complete rewrite. Hence opening a new PR seems more reasonable.

It uses rules from https://www.lispworks.com/documentation/common-lisp.html

EDIT: ~~It does currently not add dedicated scopes for globals (e.g.: `defvar`, `devparameter` or `defconstant`) (see #2319) as most other syntaxes in this repo do not treat variable definitions special or add them to index as well.~~


michaelblyons:
@s-clerc Test drive?

deathaxe:
It is just annoying to always need to either fold them or scroll down half the document just to get to the contexts. Variables are less likely navigation targets. With them being out of the way, it even less hurts to format lists to 1 item per line, which may help keeping diffs clean if keywords are added.

That's probably something to generally considder for all syntaxes.

michaelblyons:
> it even less hurts to format lists to 1 item per line, which may help keeping diffs clean if keywords are added.

Good point.

Lisp/tests/syntax_test_lisp.lisp
```diff
@@ -0,0 +1,2865 @@
...
+;                 ^ punctuation.section.parens.begin.lisp
+;                  ^^^^^^ meta.function-call.lisp support.function.lisp
+;                         ^^^^^ variable.other.lisp
+;                               ^^^^ variable.other.lisp
+;                                   ^^ punctuation.section.parens.end.lisp
+
+;;; with-clause
+
+  with var1 :float = 10.5 and var2 = true
+; ^^^^ keyword.declaration.variable.lisp
```
michaelblyons, line 2439: Someone better at Lisp than me should input whether this is right or if it's a context declaration. I would believe either.

Lisp/tests/syntax_test_lisp.lisp
```diff
@@ -0,0 +1,2865 @@
...
+;                             ^^^^ variable.other.lisp
+;                                  ^ keyword.operator.lisp
+;                                    ^^^^ constant.language.boolean.lisp
+
+;;; for-as-claus
+
+  for x-list = '(a b c) then (cdr x-list)
+; ^^^ keyword.control.loop.lisp
+;     ^^^^^^ variable.other.lisp
+;            ^ keyword.operator.lisp
```
michaelblyons, line 2455: Is this strict enough to know if it's an assignment, a comparison, or not?

michaelblyons (commented):
I have been all the way through the test file. It's very rigorous.

I have also discovered that I don't know Lisp loop structures at all.

Lisp/tests/syntax_test_lisp.lisp
deathaxe: Maybe someone else wants to implement https://www.lispworks.com/documentation/lw70/CLHS/Body/m_loop.htm#loop in a more detailed way, but the rather simple `loob-body` context can't ensure a more detailed meaning, I guess.

Lisp/tests/syntax_test_lisp.lisp
deathaxe: According to https://www.lispworks.com/documentation/lw70/CLHS/Body/m_loop.htm#loop, this is part of the `variable-clause` and looks rather like a variable declaration statement, followed by main-clauses, which describe looped statements.

michaelblyons (approved):
It's better than my attempt and better than the current one. It'd be nice to get Lisp-user eyes, but since none are available, I say ship it.

FichteFoll (approved):
I basically only checked it out locally and opened the syntax test file to check whether the tokens looked reasonable with my color scheme. This is most definitely not a proper review, but since we need 2 approvals and I don't see another contributor with some lisp experience appearing out of thin air, here's mine so that we can get this out there.

deathaxe:
Thank you guys!

PR #3898, opened 2023-12-29
[CSS] Relaxed Nesting Support

### LINKED ISSUE #3897: [CSS] Relaxed Nesting Support
### What happened?

According to https://developer.chrome.com/blog/css-nesting-relaxed-syntax-update?hl=en `&` is no longer needed for nested selectors to be valid.

Hence the following is now valid CSS in all major browsers.

```CSS
dl {
  dt {
    /* dl dt styles */
  }

  dd {
    /* dl dd styles */
  }
}
```

CSS on master still requires `&` to introduce nested selectors.

Any implementation must take into account various templating syntaxes being out in the wild, which require support for interpolation within everything.

deathaxe:
Resolves #3897

This commit implements relaxed syntax for nested style rules according to https://www.w3.org/TR/css-nesting-1/#nested-style-rule.

In general it means, nested selectors of nested style rules no longer need to start with `&` operator to denote html tags.

The implementation adopts `properties-or-selectors` context from Less syntax to achieve reliable distinction between declarations and selectors without loosing support for interpolation.

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -1547,6 +1547,7 @@ contexts:
         - selector-body
 
   maybe-property:
+    - meta_include_prototype: false
     - include: comments
     - include: property-end
     - include: property-identifiers
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -335,6 +335,7 @@ contexts:
     - include: stylesheet
 
   stylesheet:
+    # https://www.w3.org/TR/css-syntax-3/#typedef-stylesheet
     - include: comments
     - include: property-lists
     - include: selectors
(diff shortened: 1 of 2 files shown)


deathaxe:
It appears SCSS and Less are becoming obsolete.

deathaxe:
Note that lazy nesting is supported by all major browsers.

keith-hall (approved):

jfcherng (approved):

deathaxe:
Thanks for reviewing.

ryami333:
How is this fix distributed - does it come bundled in a mainstream Sublime build, or will this be published independently to the Sublime package registry sooner than that?

michaelblyons:
> does it come bundled in a mainstream Sublime build, 

Yes, eventually.

> will this be published independently to the Sublime package registry sooner than that?

No, but you can get it early with the README instructions, or I think jfcherng has a script to get the latest `master` from here.

PR #3906, opened 2024-01-23
[ShellScript] Build System works with spaces in the script filename

cernekj:


ShellScript/ShellScript.sublime-build
```diff
@@ -1,5 +1,5 @@
 {
-	"shell_cmd": "./$file_name",
+	"shell_cmd": "./\"$file_name\"",
```
deathaxe, line 2: Would it for formal reasons make sense to include `./` into quotes, as it is what's done in the "windows" part some lines after?

```suggestion
	"shell_cmd": "\"./$file_name\"",
```

FichteFoll: Should not be necessary since quotes function differently in UNIX/POSIX shells compared to Windows' cmd.exe (which is used for `shell_cmd`).
```
/bin Σ ./echo


/bin > ./"echo"


/bin ▶ "./echo"
```

FichteFoll (approved):

ShellScript/ShellScript.sublime-build
deathaxe: Sure, It was just a stylistic request to prevent mixed styles as there's already

https://github.com/sublimehq/Packages/blob/7b10f9b2208b3487ec8d357795e0f70bf90223ce/ShellScript/ShellScript.sublime-build#L10

deathaxe (approved):

PR #3911, opened 2024-01-28
[Rust] Use `keyword.declaration` scope

### LINKED ISSUE #3640: [Rust] Use `keyword.declaration` scope
### What happened?

The Rust syntax definition currently does not use the `keyword.declaration` scopes for:

- `static`
- `const`
- `let`
- `fn`
- `struct`
- `trait`
- `enum`

Undecided:

- `impl`

FichteFoll:
Following the example of Java, we're replacing the `storage.type` scope
with `keyword.declaration` without keeping the `storage.type` for
compatibility. Also, like in Java, the `use` keyword gets a
`keyword.declaration.import` scope to signal that it is introducing
names into the current namespace.

`type` aliases have slightly been refactored to properly match `where`
clauses. (I did a bit of work twice here because I forgot to `git pull`
before and included the remainder …)

Closes #3640.

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -200,7 +200,7 @@ contexts:
 
     - match: '\b(macro_rules!)\s+({{identifier}})\b'
       captures:
-        1: support.function.rust
+        1: keyword.declaration.macro.rust
         2: entity.name.macro.rust
       push: macro-block
 
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -176,20 +176,20 @@ contexts:
 
     - match: '\b(const)\s+({{identifier}})'
       captures:
-        1: keyword.declaration.rust
+        1: keyword.declaration.variable.constant.rust
         2: entity.name.constant.rust
       push: declaration-after-identifier
 
     - match: '\b(static)\s+(?:(mut)\s+)?({{identifier}})'
       captures:
-        1: keyword.declaration.rust
+        1: keyword.declaration.variable.static.rust
         2: storage.modifier.rust
         3: entity.name.constant.rust
       push: declaration-after-identifier
 
     - match: \b(let)\s+(?:(mut)\s+)?
       captures:
-        1: keyword.declaration.rust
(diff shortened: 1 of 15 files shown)


Rust/Rust.sublime-syntax
```diff
@@ -187,20 +176,20 @@ contexts:
...
-        1: storage.type.rust
+        1: keyword.declaration.rust
         2: storage.modifier.rust
         3: entity.name.constant.rust
       push: declaration-after-identifier
 
     - match: \b(let)\s+(?:(mut)\s+)?
       captures:
-        1: storage.type.rust
+        1: keyword.declaration.rust
```
keith-hall, line 192: Do we not have a subscope for `let`? I think Golang uses `keyword.declaration.var` for example

keith-hall (approved):

Rust/Rust.sublime-syntax
FichteFoll: I'm not aware of a language-agnostic subscope for variables, but I also didn't dig too deep into. For the sake of not overspecializing, I chose the defensive "no subscope" approach. I'll see if I can find some existing scope names or RFCs where this was discussed before. (I wish we had some properly formulated, agreed upon and indexed scope naming guidelines …)

deathaxe: Java scopes `var` `storage.type.variant` as this is what it semantically is, next to strict types such as `int` or `boolean`.

A possibly agnostic scope for such keywords would be `keyword.declaration.variable` as it is what those keywords most likely do.

Same like `keyword.declaration.function` being used for `def` in python for instance.

FichteFoll: `variable` was also my initial guess and if I'm not alone with that, that is enough of an argument for me to start using it. I agree that it doesn't apply to Java because the `var` there indicates type inference and not the declaration of a variable.

I also added the sub-scopes `static` and `constant` for the other two variable-esque declaration keywords `static` and `const`.

keith-hall (approved):

deathaxe (approved):

PR #3912, opened 2024-01-28
[Rust] Support macro repetitions inside structs

### LINKED ISSUE #3904: [Rust] Macro improperly marked as invalid
### What happened?

The following macro gives an erroneous `source.rust meta.macro.rust invalid.illegal.rust` scope:

```rust
macro_rules! test {
    ($($name:ident),+$(,)?) => {
        pub struct Foo {
            $(
                pub $name : Option<String>
            ),+
        }

        pub async fn thing() -> () {}
    }
}
```

image

From a cursory view in the repository and from testing it seems this is applied because of this rule https://github.com/sublimehq/Packages/blob/9ba6b939bb33625cd6b92fd64ac89b4b0ace2a8c/Rust/Rust.sublime-syntax#L1071-L1080

Adding the `;` after the `struct` definition *does* make the scope go away but ironically it leads to invalid code since you can't have `;` after a struct definition:

```
error: expected item, found `;`
  --> src/main.rs:15:2
   |
15 | };
   |  ^ help: remove this semicolon
   |
   = help: braced struct declarations are not followed by a semicolon
```

FichteFoll:
This only addresses a specific case of macro repetitions inside the struct body and similar changes to other statements with body blocks are likely also necessary, but I'm keeping the change to a limited in order to observe its effects first. Further reports of such issues can be handled analogously.

Closes #3904.

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -35,8 +35,11 @@ contexts:
     # however then we run into infinite recursion. It needs to be a prototype
     # since macro_rules! allows constructing new code, thus the metavariables
     # can be inserted in just about any position in the syntax.
-    - match: '\${{identifier}}'
+    # See also: https://doc.rust-lang.org/reference/macros-by-example.html
+    - match: \${{identifier}}
       scope: variable.other.rust
+    - match: \$\(
+      scope: punctuation.section.interpolation.begin.rust
 
   statements:
     # This isn't really just "statements", it also includes all types,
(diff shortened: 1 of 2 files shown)


FichteFoll:
The other reason is that I didn't have any (valid) examples for other block statements at hand to verify with and I'm not comfortable enough with macros to craft them ad-hoc.

FichteFoll:
This doesn't handle macros like `println!($($arg));` or `println!($($arg)?);` well and I don't really see how it could … Putting the PR on hold for now.

FichteFoll:
Not in the near future. I don't think this can be handled cleanly with the current structure and inside a `prototype` match. A better approach would likely be to either create a derived syntax with macro block support added in every relevant context that a macro pattern can be used, or directly add support for macro patterns into each syntax construct natively with the upside of only having one syntax to maintain but the downside of this single syntax becoming noticably more complex.

Haven't really thought about it more than that, however.

PR #3919, opened 2024-02-07
[XML] Add .drawio extension

jfcherng:
https://draw.io is a famous open-sourced flow chart maker. And its own file format is XML with `.drawio` as the filename extension.


keith-hall (approved):
For the curious, an official document stating that it is XML can be found at https://www.drawio.com/doc/faq/save-file-formats#:~:text=.drawio%20(XML%20file)%3A%20this,with%20our%20file%20extension%20name.

deathaxe (approved):

PR #3924, opened 2024-02-17
[YAML] Improve Performance

deathaxe:
This PR proposes some changes to improve parsing performance by about 18-20%

Syntax cache size drops from 58kB to 41kB (-30%).

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -225,15 +225,17 @@ contexts:
     # http://yaml.org/spec/1.2/spec.html#style/block/
     - include: block-scalar
     - include: block-collection
-    - include: flow-scalar-plain-out-11  # needs higher priority than flow-node-11, which includes flow-scalar-plain-in
-    - include: flow-node-11
+    - include: flow-alias
+    - include: flow-collection
+    - include: flow-scalar-out-11
 
   block-node-12:
     # http://yaml.org/spec/1.2/spec.html#style/block/
     - include: block-scalar
     - include: block-collection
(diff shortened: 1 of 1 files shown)


michaelblyons:
@keith-hall and @FichteFoll, this doesn't change anything you're extending, does it? Otherwise looks good to me.

deathaxe:
I've checked against 

1. PackageDev (Sublime Syntax Definition Syntax) 
   - uses `flow-scalar-plain-out-implicit-type-12`, `block-sequence`, `property`

2. YAML Pipelines
   - extends top-level `node` context

keith-hall (approved):

michaelblyons (approved):

YAML/YAML.sublime-syntax
```diff
@@ -94,20 +94,16 @@ variables:
   _flow_scalar_end_plain_in: |- # kind of the negation of nb-ns-plain-in-line(c) c=plain-in
     (?x:
       (?=
-          \s* $
-        | \s+ \#
-        | \s* : (\s|$)
-        | \s* : {{c_flow_indicator}}
-        | \s* {{c_flow_indicator}}
+          \s+ \#
+        | \s* (?: $ | {{c_flow_indicator}} | : (?:\s|$|{{c_flow_indicator}}) )
```
FichteFoll, line 98: Was the effect on performance here cause by not grouping or by using a catpuring group? Because I find the new grouping much harder to follow (though that could probably be solved by just splitting it over multiple lines).

deathaxe: The duplicate `\s*` had most impact.

PR #3927, opened 2024-02-18
[Diff] Scope trailing spaces

### LINKED ISSUE #1304: [Diff] Trailing Whitespace should be highlighted somehow
When I do a git diff in the terminal, lines with trailing whitespace are especially highlighted to indicate I'm adding needless characters. 

capture

It would be nice to see this in Sublime Text too.

deathaxe:
Resolves #1304

This PR...

1. ~~sets syntax specific setting `"draw_white_space": ["trailing_all"]` so trailing spaces are rendered by default.~~
2. scopes trailing spaces `meta.whitespace.trailing` in inserted/removed/modified lines, so color schemes can address them and possibly apply background color.

### Example:

grafik

#### Color Scheme Rules

```jsonc
{
	"scope": "markup.changed.diff meta.whitespace",
	"background": "color(var(orange) alpha(0.15))",
},
{
	"scope": "markup.deleted.diff meta.whitespace",
	"background": "color(var(red2) alpha(0.15))",
},
{
	"scope": "markup.inserted.diff meta.whitespace",
	"background": "color(var(green2) alpha(0.15))",
},
```

### DIFF
--- a/Diff/Diff.sublime-settings
+++ b/Diff/Diff.sublime-settings
@@ -1,5 +1,4 @@
 {
-    "draw_white_space": ["trailing_all"],
     "trim_trailing_white_space_on_save": "none",
     "translate_tabs_to_spaces": false,
 }
--- a/Diff/Diff.sublime-settings
+++ b/Diff/Diff.sublime-settings
@@ -1,4 +1,5 @@
 {
+    "draw_white_space": ["trailing_all"],
     "trim_trailing_white_space_on_save": "none",
     "translate_tabs_to_spaces": false,
 }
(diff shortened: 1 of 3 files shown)


keith-hall:
> 1. sets syntax specific setting `"draw_white_space": ["trailing_all"]` so trailing spaces are rendered by default.

Is this additive with the user preferences? If I have my draw white space setting set to all, having this syntax specific setting won't "undo"/hide/override my array, just add to it? Or will it only draw (perhaps selected and) trailing whitespace? (If I were at a computer I'd check it myself...)

deathaxe:
Well, syntax specific settings take precedence over user settings, so "trailing_all" would win in diffs.

keith-hall (approved):

michaelblyons (approved):

PR #3929, opened 2024-02-18
[C#] support primary constructors not just for records

### LINKED ISSUE #3925: [C#] Primary constructor
### What happened?

Hi,

I found an issue declaring a primary constructor in C# 12 on net8.0. C# seems to be evolving in a more rapid cadence and the syntax should be able to keep up.

```csharp
public class CompanyFeaturesService(
  ILogger<CompanyFeaturesService> logger) : ICompanyFeaturesService
{
}
```

image

The same rules apply to a `struct`.

```csharp
public readonly struct Distance(double dx, double dy)
{
    public readonly double Magnitude { get; } = Math.Sqrt(dx * dx + dy * dy);
    public readonly double Direction { get; } = Math.Atan2(dy, dx);
}
```

image

keith-hall:
fixes #3925

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -342,19 +342,19 @@ contexts:
       captures:
         1: keyword.declaration.class.cs
         2: entity.name.class.cs
-      push: [class_signature, data_type_signature]
+      push: [class_signature, data_type_constraint, maybe_constructor_parameters, data_type_signature]
     - match: \b(record)\s+(?:(struct)\s+)?({{name}})
       captures:
         1: keyword.declaration.class.record.cs
         2: keyword.declaration.struct.record.cs
         3: entity.name.class.cs
-      push: [record_signature, data_type_constraint, record_parameters, data_type_signature]
+      push: [record_signature, data_type_constraint, maybe_constructor_parameters, data_type_signature]
     - match: (?:\b(ref)\s+)?\b(struct)\s+({{name}})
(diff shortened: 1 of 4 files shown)


C#/C#.sublime-syntax
```diff
@@ -342,19 +342,19 @@ contexts:
       captures:
         1: keyword.declaration.class.cs
         2: entity.name.class.cs
-      push: [class_signature, data_type_signature]
+      push: [class_signature, data_type_constraint, maybe_constructor_parameters, data_type_signature]
```
deathaxe, line 345: With this common schema, should `data_type_signature` be renamed to `data_type_generic` with only consuming `<...>` generics? It seems rather unlikely, that constraints appear in front of primary constructor parameters.

Something like that, I have in mind.

```yml
  data_type_generic:
    - match: '<'
      scope: meta.generic.cs punctuation.definition.generic.begin.cs
      push: data_type_generic_body
    - match: (?=\S)
      pop: true

  data_type_generic_body:
    - meta_content_scope: meta.generic.cs
    - include: type_parameter
```

deathaxe: see also: https://github.com/forkeith/Packages/pull/30

keith-hall: Thanks, merged

deathaxe (approved):

michaelblyons (approved):

PR #3931, opened 2024-02-24
[Sublime] Add Syntax Test Results syntax

deathaxe:
This PR proposes to add a _Sublime_ package with _Sublime Syntax Test Results_ syntax definition for highlighting syntax test result outputs with regards to https://github.com/sublimehq/sublime_text/issues/3022.

Thoughts:

1. A package named "Sublime" does not yet exist on package control.
2. The package name is generic enough to potentially take more Sublime Text related syntax files.

Maybe more suitable alternative package names are appriciated.

Syntax definition is designed to highlight ST's default syntax test result output - despite the proposal in https://github.com/sublimehq/sublime_text/issues/3022, which addresses a reformatted output.

grafik

### DIFF
--- a/Sublime/Sublime Syntax Test Results.sublime-syntax
+++ b/Sublime/Sublime Syntax Test Results.sublime-syntax
@@ -66,7 +66,10 @@ contexts:
 
   expected-scope-body:
     - meta_scope: meta.scope.expected.sublime.syntax-test-results
-    - include: selector-body
+    - match: \]
+      scope: punctuation.section.scope.end.sublime.syntax-test-results
+      pop: 1
+    - include: scope-selectors
 
   found-scope:
     - match: \[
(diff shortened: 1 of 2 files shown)


FichteFoll:
I am definitely in favor of also reformatting the line display like in https://github.com/sublimehq/sublime_text/issues/3022 because it is *much* easier to compare the expected selector and actual scope that way.

michaelblyons:
> 2. The package name is generic enough to potentially take more Sublime Text related syntax files.

Are you thinking of moving the file patterns? That makes sense to me if this folder is added.

deathaxe:
I didn't have certain syntaxes in mind - just found adding a package named _Sublime Syntax Test Results_ with only one file/syntax inside rather too specific.

Not sure, what a good name for such ST related syntaxes would be.

Depending on how packages are finally built by sublimehq, I could also imagine creating a "Default" package in this repo. It however would require containing files to be added to "Default.sublime-package". Not sure if that's desirable.

deathaxe:
Closing as obsolete with regards to progress made in https://github.com/sublimehq/sublime_text/issues/3022.

Issue #1036, opened 2017-06-16
gutter markup scope names


deathaxe:
I recently found the following `markup.` scopes in the latest Monokai.tmTheme when looking for some possibly common scopes to mark warnings/errors for a package I am working on (fork of [EasySettings](https://github.com/deathaxe/easysettings))

- markup.deleted
- markup.inserted
- markup.changed _(missing in [ST Documentation](https://www.sublimetext.com/docs/3/scope_naming.html#markup))_
- message.error _(missing in [ST Documentation](https://www.sublimetext.com/docs/3/scope_naming.html#markup))_

SublimeLinter uses

- sublimelinter.mark.error
- sublimelinter.mark.warning

### Proposal

I suggest to extend the default `markup.` scope section in the documentation and actively make use of and propose them to be handled by color schemes.

**For use with diff and vcs**

- markup.deleted _(already in docs)_
- markup.inserted _(already in docs)_
- markup.changed
- markup.ignored
- markup.untracked

**For use with linter or debugger**

- markup.info
- markup.warning
- markup.error

### Motivation

The motivation of this proposal is to define some common markup scopes which can be used by all packages, which need to add some "message" like information to the gutter area or markup some pieces of source and make sure color schemes handle the most basic classes.

I would like to see the day, SublimeLinter does no longer hack all the color schemes to inject its scopes in there. Instead it should use standardized `markup.warning.linter` or `markup.error.linter` scopes which are handled by all color schemes.

rwols:
I wanted to write a proposal about `log.` scopes, but I think `markup.` also fits.

Essentially, something standard for output panels is missing. I thought of
* `log.fatal`,
* `log.error`,
* `log.warning`,
* `log.note`,
* `log.info`,
* `log.suggestion`.

Markup would also fit: 

* `markup.fatal`, --> unrecoverable error, stop.
* `markup.error`, --> error, but continue
* `markup.warning`, --> basic warning
* `markup.note`, --> extra info *around* errors and warnings, not required reading material.
* `markup.info`, --> general info about what is happening
* `markup.suggestion`. --> how to fix something

I'm fine with both approaches.

There are two approaches to these kinds of scopes: If an error occurs, do you scope the **whole line** that explains the error as `markup.error`, or just **some parts of the line** as `markup.error` like the word "error", and perhaps some text after it?

deathaxe:
@rwols: A good extension to my proposal.

[...]

PR #3932, opened 2024-02-25
[Python] Add todo highlighting

UngeheurenUngeziefer:
This PR suggests adding support for highlighting "todo" syntax inside comments. 

E.g. in one of color schemes:
todo_highlight

Currently, no packages have this functionality. Existing packages only allow for background color changes, not text color. It could be worthwhile to create a package with this feature. However, this would require modifications to the syntax file (or creation of alternative syntax file), which makes it more fitting to include in the default Python syntax.

I used string.todo to allow users to select alternative colors for this namespace. Using comment.todo might be a better option, but it will not highlight todos. This is because all existing themes do not use it, meaning it will remain as a regular comment.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -364,6 +364,12 @@ contexts:
 ###[ COMMENTS ]###############################################################
 
   comments:
+    - match: '(# ?)(TODO|ToDo|Todo|todo)\b(.*)'
+      captures:
+        1: comment
+        2: string.todo
+        3: string.todo
+
     # Special Sphinx comment syntax to denote documentation
     # https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute
     - match: '#:'
--- a/Python/tests/syntax_test_python_strings.py
+++ b/Python/tests/syntax_test_python_strings.py
@@ -1656,3 +1656,11 @@ f'''{
 }'''
 # <- meta.string.python meta.interpolation.python punctuation.section.interpolation.end.python
 #^^^ meta.string.python string.quoted.single.block.python punctuation.definition.string.end.python
+
+## Todo: some task
+#^ comment
+#  ^^^^^^^^^^^^^^^ string.todo
+
+##ToDo: some task
+#^ comment
+# ^^^^^^^^^^^^^^^ string.todo


jfcherng:
https://packagecontrol.io/packages/Colored%20Comments

deathaxe:
A proper TODO syntax targets a larger scope than just highlighting `TODO` keyword in comments. There's also `FIXME` and others to considder. 

Here's a comment about this topic from a former sublimehq developer: https://forum.sublimetext.com/t/todo-syntax-highlighting/27604/9

I still fully agree with his statements and fully support them.

Python/Python.sublime-syntax
```diff
@@ -364,6 +364,12 @@ contexts:
 ###[ COMMENTS ]###############################################################
 
   comments:
+    - match: '(# ?)(TODO|ToDo|Todo|todo)\b(.*)'
+      captures:
+        1: comment
```
keith-hall, line 369: not scoping the `#` as `punctuation.definition.comment` won't play well with features like joining/wrapping lines.

Python/Python.sublime-syntax
```diff
@@ -364,6 +364,12 @@ contexts:
 ###[ COMMENTS ]###############################################################
 
   comments:
+    - match: '(# ?)(TODO|ToDo|Todo|todo)\b(.*)'
```
keith-hall, line 367: we generally try to avoid matching `.*` - instead we prefer to push to a different context, so syntax definitions which inherit or re-use these rules can do extra processing on that text (it allows `prototype`s to apply etc.)

Python/Python.sublime-syntax
```diff
@@ -364,6 +364,12 @@ contexts:
...
+      captures:
+        1: comment
+        2: string.todo
+        3: string.todo
```
keith-hall, line 371: > I used `string.todo` to allow users to select alternative colors for this namespace. Using comment.todo might be a better option, but it will not highlight todos. This is because all existing themes do not use it, meaning it will remain as a regular comment.

This would give the wrong semantic meaning to the text. Sometimes I do text searches within specific scopes - like a word but only in strings, or only in comments. Scoping this wrongly would mess with this.

Instead of catering to existing color schemes like this, it would be better to stack scopes, like `comment.line` plus [`markup.info` or something](https://github.com/sublimehq/Packages/issues/1036)

keith-hall (commented):
Commenting purely from a technical perspective, and not joining the debate of whether we want to open the floodgates of having each syntax definition to scope `todo`s etc in comments.

Python/Python.sublime-syntax
FichteFoll: Secondly, just `comment` as the scope name is way too lazy. Please take a look at what scope name the other comment-related rules use.

FichteFoll:
Since there appears to be unanimous agreement on the mostly technical problems with this suggestion (myself included), I'm going ahead with closing this PR.

PR #3933, opened 2024-02-26
[PHP] Add build systems

Sainan:
I'm not sure if this is the right convention since I see no other default build systems differentiating between "Run" and "Lint" but I find this to be very useful, more so than trying to mess with some buggy Python plugins.

### DIFF
--- a/PHP/PHP.sublime-build
+++ b/PHP/PHP.sublime-build
@@ -1,7 +1,7 @@
 {
     "cmd": ["php", "$file"],
     "file_regex": "^(?:php:)?[\t ](...*?):([0-9]*):?([0-9]*)",
-    "selector": "source.php",
+    "selector": "embedding.php | source.php",
 
     "variants": [
         {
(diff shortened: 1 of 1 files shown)


PHP/Lint PHP.sublime-build
keith-hall, line 1: Typically a variant is used for linting instead of a separate build system. See https://github.com/sublimehq/Packages/blob/master/Python/Python.sublime-build#L12-L15

Sainan: Thanks for the note about variants, seems like a decent way of handling it. I've updated the commit.

jfcherng:
LGTM except that I prefer adding a trailing newline at EOF.

PHP/PHP.sublime-build
```diff
@@ -0,0 +1,12 @@
+{
+    "cmd": ["php", "$file"],
+    "file_regex": "^(?:php:)?[\t ](...*?):([0-9]*):?([0-9]*)",
+    "selector": "source.php",
```
jfcherng, line 4: `embedding.php | source.php` 

https://github.com/sublimehq/Packages/blob/472ba632a732366f2a304faed86d3f6b378aabe4/PHP/PHP.sublime-syntax#L4C8-L4C21

jfcherng (approved):

deathaxe (approved):

PR #3941, opened 2024-03-04
[Markdown] Simplify frontmatter punctuation scopes

deathaxe:
This commit renames scopes of `---` frontmatter markers, so they follow common scope naming schema:

```
   meta.frontmatter punctuation.section.frontmatter
```

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -311,25 +311,25 @@ contexts:
     - match: (---)\s*(?i:(json)\s*)\n
       captures:
         0: meta.frontmatter.markdown
-        1: punctuation.section.block.begin.frontmatter.markdown
+        1: punctuation.section.frontmatter.begin.markdown
         2: constant.other.language-name.markdown
       embed: scope:source.json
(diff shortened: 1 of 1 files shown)


FichteFoll (approved):

michaelblyons:
Apologies if dumb question: Why not `meta.frontmatter punctuation.section.block.begin`? Still targetable, fits more schemes' existing rules.

deathaxe:
The trigger was SCSS using this scheme and I found Markdown and some other syntaxes (e.g.: Liquid) not to. Those are planned for update as well to apply a common scopes.

1. Frontmatter is not scoped `meta.block` and probably shouldn't be. Thus using `punctuation.section.block` seems not a perfect match for scoping `---` markers.

2. `punctuation.section.block` scope is primarily used to denote `{` and `}` in C like syntaxes.

3. The change intents to apply scope naming guideline which uses `meta.<name> punctuation.section.<name>.[begin|end]` scheme.

4. Only few syntaxes extend punctuation scopes after `begin` or `end`.

michaelblyons (approved):

PR #3944, opened 2024-03-11
[CSS] Add :global() and :local() pseudo-classes

deathaxe:
This commit adds `:global(selector)` and `:local(selector)` pseudo-class support. The former is used by CSS Modules, Less and Astro framework. The ladder is used by CSS Modules only, but is added for completeness.

It is no default CSS feature, but appears to be specific enough to not cause any conflicts.

Adding it to core CSS would avoid the need to extend CSS for Astro, just for this single reason.

Vue.js also uses `:global` (see: https://vuejs.org/api/sfc-css-features.html#global-selectors).

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -90,6 +90,9 @@ variables:
 
   keyframe_selector_begin: (?=\b(?i:from|to){{break}}|\.?[\d,%])
 
+  pseudo_classes_with_selector_args: |-
+    (?xi: global | local | matches | is | where | not | has | host(?:-context)?)
+
   # Combinators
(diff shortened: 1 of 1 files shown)


FichteFoll (approved):

deathaxe:
Another request for supporting `:global()` was made for vue.js at [Sass](https://github.com/SublimeText/Sass/issues/110) package.

This request technically also hits CSS as also being supported/used by vue.js

Are there certain concerns against adding it here?

michaelblyons (approved):
Seems safe enough. If someone is really offended, we can revisit later.

keith-hall (approved):

jfcherng (approved):

PR #3947, opened 2024-03-14
[YAML] Add CITATION.cff file extension

jwortmann:
https://citation-file-format.github.io/

Also see https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md which states that the file must be named exactly `CITATION.cff` and that it is YAML.

### DIFF
--- a/YAML/YAML.sublime-syntax
+++ b/YAML/YAML.sublime-syntax
@@ -43,6 +43,8 @@ file_extensions:
   - yaml
   - yml
   - sublime-syntax
+
+hidden_file_extensions:
   - CITATION.cff
 
 first_line_match: ^%YAML( ?1.\d+)? # Technically the number is required, but we'll be a bit loose here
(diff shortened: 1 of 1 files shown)


jfcherng (dismissed):

jfcherng:
It's spec says CCF is yaml and the only reasonable usage seems to be `CITATION.cff`. 
https://citation-file-format.github.io/assets/pdf/cff-specifications-1.0.0.pdf

deathaxe (approved):

deathaxe:
Would it make sense to add it to `hidden_file_extensions` as it points to a whole file, which is unlikely to be used as normal file extension and thus should be hidden from "Save as" dialog?

jwortmann:
Yes, I think that would be better. Currently it shows up as `*.CITATION.cff` if the "Save as" dialog, which could be misleading. I'll change it, unless there are any objections.

deathaxe (approved):

jfcherng (approved):

michaelblyons (approved):

PR #3963, opened 2024-04-25
[D] Add highlighting for recognised __traits/pragmas

ichordev:
Changes:
- Valid `__traits` are highlighted with `constant.other.d`.
- Invalid `__traits` now trigger `invalid.illegal.d`, because they cause compile errors.
- Highlight known pragmas with `constant.other.d`. Invalid pragmas are not *meant* to cause compile errors in D (currently they do, but it's being fixed), so they do NOT trigger `invalid.illegal.d`.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -65,6 +65,9 @@ variables:
   language_variable: 'this|super'
   reserved: '{{keyword}}|{{basic_type}}|{{language_constant}}|{{language_variable}}'
 
+  traits: 'isAbstractClass|isArithmetic|isAssociativeArray|isFinalClass|isPOD|isNested|isFuture|isDeprecated|isFloating|isIntegral|isScalar|isStaticArray|isUnsigned|isDisabled|isVirtualFunction|isVirtualMethod|isAbstractFunction|isFinalFunction|isStaticFunction|isOverrideFunction|isTemplate|isRef|isOut|isLazy|isReturnOnStack|isCopyable|isZeroInit|isModule|isPackage|hasMember|hasCopyConstructor|hasPostblit|identifier|getAliasThis|getAttributes|getFunctionAttributes|getFunctionVariadicStyle|getLinkage|getLocation|getMember|getOverloads|getParameterStorageClasses|getPointerBitmap|getCppNamespaces|getVisibility|getProtection|getTargetInfo|getVirtualFunctions|getVirtualMethods|getUnitTests|parent|child|classInstanceSize|classInstanceAlignment|getVirtualIndex|allMembers|derivedMembers|isSame|compiles|toType|initSymbol|parameters'
+  pragmas: 'crt_constructor|crt_destructor|inline|lib|linkerDirective|mangle|msg|printf|scanf|startaddress|LDC_alloca|LDC_allow_inline|LDC_extern_weak|LDC_global_crt_ctor|LDC_global_crt_dtor|LDC_inline_ir|LDC_intrinsic|LDC_never_inline|LDC_no_moduleinfo|LDC_no_typeinfo|LDC_va_start|LDC_va_arg|LDC_va_end|LDC_va_copy'
+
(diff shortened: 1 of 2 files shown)


D/D.sublime-syntax
```diff
@@ -697,16 +700,20 @@ contexts:
       scope: punctuation.section.parens.end.d
       pop: true
   pragma-attribute:
+    - match: '\b({{pragmas}})\b'
+      scope: variable.function.d constant.other.d
```
deathaxe, line 704: Uncertain about `msg|printf|scanf|` but the others rather look like plain flags or something along those lines. Sure those still should be scoped `variable.function`?

Using `variable.function` in general looks suspicious, but I am probably not familiar enough with D to judge it.

D/D.sublime-syntax
```diff
@@ -1191,12 +1198,16 @@ contexts:
         - match: '\('
           scope: punctuation.section.parens.begin.d
           set:
-            - match: '\b{{name}}\b'
-              scope: variable.function.d
+            - match: '\b({{traits}})\b'
+              scope: variable.function.d constant.other.d
```
deathaxe, line 1202: Why stacking `variable.function` and `constant.other`?

deathaxe (commented):
A certain identifier not being supported by a compiler (causing compiler errors) may not directly be equal to invalid syntax.

In general being too over-eager may be annyoing to end users due to seeeing tons of red (illegal highlighted) code all the time while editing code. 

That's however too late to question for D anyway though.

D/D.sublime-syntax
ichordev: Look up the D specification. Also variable.function was already there

D/D.sublime-syntax
ichordev: So that the syntax highlighting falls back in case `constant.other` is not supported. If you prefer just `constant.other` then I can change it.

ichordev:
> A certain identifier not being supported by a compiler (causing compiler errors) may not directly be equal to invalid syntax.

Valid `__traits` are specified in D's grammar. The reference compiler (DMD) gives you an error, so LDC and GDC will both give you an error too because they use the same backend. As far as I know, this is intentional behaviour.

Like I noted, unrecognised `pragma`s are not supposed to cause errors because implementation-specific pragmas are provided for by the spec, and so they don't trigger invalid syntax highlighting here. Instead, unrecognised pragmas just don't get special highlighting so that you can see that something is wrong (i.e. you probably made a typo).

D/D.sublime-syntax
deathaxe: A am aware of `variable.function` already having been used. According to D specification, I'd judge said `msg`, `printf` and `scanf` pragmas possible functions, but the others just appear to be some sort of compiler flags. Thus `variable.function` doesn't seem appropriate - semantically.


In general current D syntax is somewhat lazy with regards to scope names - sometimes even doesn't strictly follow some of the official guidelines.

D/D.sublime-syntax
deathaxe: `constant.other` is a very basic scope required to be implemented by each color scheme. There's no fallback required. Actually this combination is just odd.

deathaxe:
A synatx definition should highlight tokens and ideally handle incomplete token/statements in sane ways to create the impression of it already knowing what's up next.

It is not necessarily required to do a linter's job. Even though the compiler isn't happy with a certain token, currently, support for new ones may be added at any time. So unsupported pragmas are not a syntax error. The compiler's parser would also read the full token, but just throw an error at a later point, as it doesn't know it.

ichordev:
> So unsupported pragmas are not a syntax error. The compiler's parser would also read the full token, but just throw an error at a later point, as it doesn't know it.

Yes, ***pragmas*** are [designed to accommodate compiler-specific identifiers](https://dlang.org/spec/pragma.html#vendor_specific_pragmas), so my PR already accounts for the fact that their identifiers are arbitrary by just ignoring unrecognised identifiers.

[Traits, on the other hand, are defined as an exhaustive list of keywords](https://dlang.org/spec/traits.html#grammar). A traits expression, by definition, must use one of these keywords. Yes, new keywords may be added, but the same holds for every other aspect of the language's syntax.

deathaxe (approved):

D/D.sublime-syntax
```diff
@@ -1191,12 +1198,16 @@ contexts:
...
+            - match: '\b({{traits}})\b'
+              scope: constant.other.d
               set:
                 - match: ','
                   scope: punctuation.separator.sequence.d
                   set: [value-parens-after, value-list]
+                - match: '\)'
+                  scope: punctuation.section.parens.end.d
+                  pop: true
+            - include: not-whitespace-illegal-pop
```
FichteFoll, line 1210: As @deathaxe mentioned, I think it would be preferrable if this didn't assign an illegal scope but rather a generic scope. That would still differ from the known-good values matched via `{{traits}}` but not mark anything else as illegal, which is especially annoying during typing.

ichordev: Feel free to reopen it then

FichteFoll: Right, I somehow though the branch was deleted, in which case a PR cannot be re-opened.

FichteFoll (approved):

PR #3965, opened 2024-04-25
[D] Separate built-in type-aliases from built-in types

ichordev:
Changes:
- Separated built-in type-aliases (e.g. `size_t`, `string`) from built-in types (e.g. `int`, `float`) so that their identifiers aren't treated as reserved (they aren't, e.g. you can make your own `string`), and so that they can be highlighted separately.


ichordev:
I swear it works this time.

keith-hall (approved):

D/D.sublime-syntax
```diff
@@ -2234,15 +2238,19 @@ contexts:
...
-          pop: true
-    - include: not-whitespace-illegal-pop
+      set: type-identifier-after
+    - include: not-whitespace-illegal-pop
+  type-identifier-after:
+      - match: \.
+        scope: meta.path.d punctuation.accessor.dot.d
+        set: type-identifier
+      - match: '(?=\S)'
+        pop: true
```
deathaxe, line 2253: Overindented

```suggestion
    - match: \.
      scope: meta.path.d punctuation.accessor.dot.d
      set: type-identifier
    - match: '(?=\S)'
      pop: true
```

D/D.sublime-syntax
```diff
@@ -2234,15 +2238,19 @@ contexts:
     - match: '\b{{name}}\b(?=\s*!)'
       scope: meta.function-call.template.d meta.path.d variable.function.d
       pop: true
+    - match: '\b{{builtin_type_alias}}\b'
+      scope: meta.path.d storage.type.d support.type.builtin-alias.d
```
deathaxe, line 2242: As above, just `support.type` would be enough.

```suggestion
      scope: meta.path.d support.type.builtin-alias.d
```

deathaxe (approved):

deathaxe:
Please fix the failing test. Otherwise LGTM!

deathaxe (approved):

keith-hall (approved):

PR #3966, opened 2024-04-25
[D] Add shortened function definition (#3894)

### LINKED ISSUE #3894: [D] Add support for shortened function syntax
### What happened?

D recently introduced a [shortened function syntax](https://dlang.org/spec/function.html#function-bodies) that works analogous to lambda expressions:

```
    int[] map(int[] array) => foo();
    int[] map(T)(int[] array) => foo();
```

This is currently parsed as a syntax error.

ichordev:
Fixes #3894.
Changes:
- Support shortened function definition syntax (e.g. `int main() => 0;`) in all its forms:
  - plain functions;
  - functions with one-line pre/postconditions;
  - template functions with template constraints.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -916,7 +916,7 @@ contexts:
   function-definition-after-arguments:
     - meta_scope: meta.function.d
     - include: function-attribute-in
-    - match: '='
+    - match: '=(?!>)'
       scope: keyword.operator.assignment.d
       set: [meta-function, expect-end-of-line, value]
     - match: '\bif\b'
@@ -939,6 +939,9 @@ contexts:
     - match: '\b(do|body)\b'
       scope: keyword.other.d
       set: block-statement
+    - match: '=>'
+      scope: keyword.declaration.function.d
+      set: [meta-function, expect-end-of-line, value]
     - match: '{{block_statement_loohahead}}'
       set: block-statement
     - include: expect-end-of-line
--- a/D/tests/syntax_test_d.d
+++ b/D/tests/syntax_test_d.d
@@ -1250,6 +1250,32 @@ extern(1)
 //                  ^^ meta.function.d meta.block.d
 //                  ^ punctuation.section.block.begin.d
 //                   ^ punctuation.section.block.end.d
+  int boo() => 5;
+//^^^ storage.type.d
+//    ^^^ meta.function.d entity.name.function.d
+//       ^^ meta.function.parameters.d
+//       ^ punctuation.section.group.begin.d
+//        ^ punctuation.section.group.end.d
+//          ^^ meta.function.d keyword.declaration.function.d
+//             ^ meta.number.integer.decimal.d
+//              ^ meta.function.d punctuation.terminator.d
+  int par(T)() if(true) => 7;
+//^^^ storage.type.d
+//    ^^^ meta.function.d entity.name.function.d
+//       ^^^^^ meta.function.parameters.d
+//       ^ punctuation.section.group.begin.d
+//        ^ variable.parameter.d
+//         ^ punctuation.section.group.end.d
+//          ^ punctuation.section.group.begin.d
+//           ^ punctuation.section.group.end.d
+//             ^^^^^^^^^^^^^ meta.function.d
+//             ^^ keyword.control.conditional.d
+//               ^ punctuation.section.parens.begin.d
+//                ^^^^ constant.language.d
+//                    ^ punctuation.section.parens.end.d
+//                      ^^ meta.function.d keyword.declaration.function.d
+//                         ^ meta.number.integer.decimal.d
+//                          ^ meta.function.d punctuation.terminator.d
     void bar();
   //^^^^ storage.type.d
   //     ^^^ meta.function.d entity.name.function.d


deathaxe:
When strictly following https://dlang.org/spec/function.html#function-bodies would shorthands with [FunctionContracts](https://dlang.org/spec/function.html#FunctionContracts) need to be supported as well?

```D
    int[] map(T)(int[] array) in(array) => 1;

    int[] map(T)(int[] array) out(array) => 2;

    int[] map(T)(int[] array) in(array) out(array) => 3;
```

grafik

Not sure if `out(array)` makes sense in that context.

ichordev:
> When strictly following https://dlang.org/spec/function.html#function-bodies would shorthands with [FunctionContracts](https://dlang.org/spec/function.html#FunctionContracts) need to be supported as well?

One-line pre/postconditions (AKA `InOutContractExpressions`) are already supported by my PR, as noted in the description I provided:
> * Support shortened function definition syntax __in all its forms__:
>   * functions with one-line __pre/postconditions__;

Multi-line pre/postconditions (AKA `InOutStatement`) are not allowed to be the last contract before the shortened function syntax. My PR currently ignores this rule for simplicity.

> Not sure if `out(array)` makes sense in that context.

You can't use `out(someThing)`, you will get a compiler error. It's `out(; someThing)` or `out(identifier; identifier == someOtherThing)`.

ichordev:
> The point or question is not about content of parentheses, but rather whether the `=>` being scoped illegal after the `out(...)` contract is wanted.

That is how the syntax already behaves with regular function declaration syntax. Modifying it is not even within the scope of this pull request.

> According to the syntax specs, I'd expect `=>` to be scoped the same in all three lines of provided screenshot/code example, but the last 2 differ.
> 
> Note: I am reading specs only, I don't have an compiler or sdk installed for testing.

If you read the grammar defined in the spec, you will see that the semicolon in `out` is required. It is also required by the pre-existing D syntax from this repository.
You are complaining that your second two examples are ‘scoped differently’, which is caused by you writing invalid code. If you don’t like the way the syntax parses syntactically invalid `out` contracts, then submit a PR to change it instead of complaining about it in this unrelated PR.

Furthermore, the pathetic cadence and bafflingly reasoned comments I have received from the reviewing of such trivial improvements has been stifling to say the least. There are a whole host of outdated aspects of  this syntax that should’ve been fixed years ago, and now I see why nobody has. If I had expected this much friction over a barebones 4 line addition to support a feature added [3 years ago](https://dlang.org/changelog/2.096.0.html#shortfunctions) then I would’ve left kept my changes upstream.

deathaxe:
According to the following rules extracted from https://dlang.org/spec/function.html#function-bodies 

```
FuncDeclaration:
    StorageClassesopt BasicType FuncDeclarator FunctionBody
    AutoFuncDeclaration

AutoFuncDeclaration:
    StorageClasses Identifier FuncDeclaratorSuffix FunctionBody

ShortenedFunctionBody:
    InOutContractExpressionsopt => AssignExpression ;

InOutContractExpressions:
    InOutContractExpression
    InOutContractExpression InOutContractExpressions

InOutContractExpression:
    InContractExpression
    OutContractExpression

OutContractExpression:
    out ( ; AssertArguments )
    out ( Identifier ; AssertArguments )
```

a shortend function definition such as

```
int[] map(T)(int[] array) in(array) out(array) => 3;
```

appears to be valid, while with this PR `=>` still is scoped invalid. The question was just, whether that's ok or whether a further change is required to scope `=>` as operator/keyword as well after an `OutContractExpression`.

ichordev:
> According to the following rules extracted from https://dlang.org/spec/function.html#function-bodies
> ```
> OutContractExpression:
>     out ( ; AssertArguments )
>     out ( Identifier ; AssertArguments )
> ```
> a [shortened] function definition such as 
> `int[] map(T)(int[] array) in(array) out(array) => 3;`
> appears to be valid

`out ( AssertArguments )` 'appears to be valid' does it?
```
OutContractExpression:
    out ( ; AssertArguments )
    out ( Identifier ; AssertArguments )
```
And yet `out ( AssertArguments )` is not on this list. Perhaps you forgot to read the list before sending it?

> with this PR `=>` still is scoped invalid. The question was just, whether that's ok or whether a further change is required to scope `=>` as operator/keyword as well after an `OutContractExpression`.

When you write an [`OutContractExpression`](https://dlang.org/spec/function.html#OutContractExpression), `=>` is still scoped as an operator:
```d
int[] map(T)(int[] array) out(; array) => 2;
int[] map(T)(int[] array) in(array) out(; array) => 3;
```
image
When you write `out(array)` it matches the start of an [`OutStatement`](https://dlang.org/spec/function.html#OutStatement), so it scopes `=>` as invalid because it is expecting a [`BlockStatement`](https://dlang.org/spec/statement.html#BlockStatement). This is a helpful feature, and can even save you from looking stupid by insisting that a D syntax error 'appears to be valid' according to a grammar that you didn't even bother to read.

michaelblyons:
> save you from looking stupid by insisting that a D syntax error 'appears to be valid' according to a grammar that you didn't even bother to read.

Keep it civil. Your previous post was likewise somewhat lacking in civility.

deathaxe:
Actually I wouldn't have expected a semicolon within the parentheses to make a difference, which is IMHO an over pedantic treatment - not caused by this PR though. A syntax should ideally be able to handle incomplete expressions in sane ways without breaking syntax highlighting - to some extend at least. I therefore haven't realized the difference between `OutStatement` and `OutContractExpression` as I followed the latter path only.

Maybe some additional test cases could help verifying those valid forms/variants of a function definition work as expected.

With the misanderstanding clarified this PR is correct and valid.

deathaxe:
> Keep it civil.

Well, true. Let's claim disappointment and unfamilarity with how open source projects try to ensure a least level of quality.

The truth is all maintainers here just being unpaid volunteers spending their spare time to help improving syntaxes. Most of us with limited time and scope with regards to familiar or used languages/syntaxes.

Despite those limitation we can't just merge anything reaching in, if parts of it seem unclear.

D is not one of the mainline syntaxes. It's also true, me not studying its specs to the deepest detail because I don't use or need it at all. Still trying to fix bugs, which seem simple to fix, despite limited knowledge about or motivation for it.

As such I can just appologize if the attempt to fill some gaps offended anyone.

PR #3969, opened 2024-04-26
[Markdown] Fix backticks and code-spans in table cells

### LINKED ISSUE #3968: [Markdown] Weirdness with whitespace-only inline code in tables
### What happened?

```
empty ` ` code

| `x` | `code` empty ` ` code in table |
|--:|-:|
| empty ` ` code in table | `code` empty ` ` code in table |
|body|

|  `code` empty ` ` code in table | nothing |
|--:|-:|
| empty ` ` code in table | `code` empty ` ` code in table |
|body|
```

Scopes are: 
- `text.html.markdown meta.table.header.markdown-gfm invalid.deprecated.unescaped-backticks.markdown`
- `text.html.markdown meta.table.markdown-gfm invalid.deprecated.unescaped-backticks.markdown`

Rendered:

---

empty ` ` code

| `x` | `code` empty ` ` code in table |
|--:|-:|
| empty ` ` code in table | `code` empty ` ` code in table |
|body|

|  `code` empty ` ` code in table | nothing |
|--:|-:|
| empty ` ` code in table | `code` empty ` ` code in table |
|body|

<details>
<summary>screenshot of rendering via the MarkdownPreview package using GFM flavor</summary>

screenshot

</details>

deathaxe:
Fixes #3968

This commit...

1. removes `invalid.deprecated` scope from single backticks in table cells
2. fixes code spans treatment in tables. Codespans require balanced backticks in a single line not containing a `|` cell separator are found.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -126,6 +126,12 @@ variables:
     | (?! \s*\-\s+ | \s+\|){{table_cell}}\|(?!\s+$)
     )
 
+  table_codespan_content: |-
+    (?x:
+      [^`|]             # first or only char must not be a backtick or pipe.
+      (?:[^|]*[^`|])?   # none must be a pipe, the last additionally must not be a backtick
+    )
+
   fenced_code_block_start: |-
     (?x:
       ([ \t]*)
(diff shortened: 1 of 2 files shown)


Markdown/Markdown.sublime-syntax
```diff
@@ -2366,14 +2366,22 @@ contexts:
...
     - include: images
-    - include: literals
+    - include: critics
+    - include: math-inline
+    - include: escapes
     - include: links
     - include: markups
 
+  table-cell-code-spans:
+    - match: (`+)[^`|]+(\1)
```
keith-hall, line 2379: Do we want to consider avoiding the use of Oniguruma by hard-coding the regex pattern for a single-backtick, double-backtick, triple-backtick etc. up to something reasonable like 5 or so?

deathaxe: Normal code-spans even use branching in combination with `\1` and are even way slower than this implementation for table cells.

A benchmark with `\1` vs. fixed patterns using following table with 10k lines reveals a defference of 62ms vs. 40ms, while using default branching approach would use 95ms on my box.

```
| column 1 | column 2
| ---      | ---
| ` ` | ` me ` | `` ` `` | ` `` ` | ``foo`bar``
```

But well, maybe 5-6 backticks are enough in real world use cases in favor of a bit better performance, even though I doubt so many table cells with code spans are common.

deathaxe:
The only downside of current implementation - theoretically - is template syntaxes such as Astro not being able to inject temblate tags into code spans in table cells.

```md
| column 1  | column 2
| ---       | ---
| `{{key}}` | `{{value}}`
```

Those would currently appear as normal raw text.

Markdown/Markdown.sublime-syntax
michaelblyons: Does `` `{1,5}`` not work?

keith-hall: The problem is that for balanced backticks, you'd still need to use a back reference for the closing backticks if you used ``(`{1,5})`` for the opening backticks

keith-hall:
> The only downside of current implementation - theoretically - is template syntaxes such as Astro not being able to inject temblate tags into code spans in table cells.
> 
> ```
> | column 1  | column 2
> | ---       | ---
> | `{{key}}` | `{{value}}`
> ```
> 
> Those would currently appear as normal raw text.

If that will really be a problem then I imagine it could be fixable by having the whole `table-cell-code-spans` use a lookahead and push into a context. If the lookahead handles the balanced backticks, the inner context can just pop when it sees the closing backticks and consume them all without worrying about how many there are

keith-hall (approved):

Markdown/Markdown.sublime-syntax
michaelblyons: Ah, okay. I misunderstood the difference as an unbounded length on the back red, rather than the use of one at all. Thanks!

michaelblyons (approved):

PR #3970, opened 2024-04-27
[D] Update keywords, attributes, and operator overloads

ichordev:
Changes:
- Removed `body` from `keywords` and as an alternative to `do` in function contracts because it is no longer a keyword, and no longer acts as an alternative to `do` in function contracts.
- Added the missing [`@gnuAbiTag`](https://dlang.org/phobos/core_attribute.html#.gnuAbiTag), [`@live`](https://dlang.org/spec/ob.html#live), [`@mustuse`](https://dlang.org/spec/attribute.html#mustuse-attribute), [`@optional`](https://dlang.org/spec/objc_interface.html#optional-attribute), [`@standalone`](https://dlang.org/changelog/2.107.0.html#dmd.standalone-attribute), and [`@weak`](https://dlang.org/phobos/core_attribute.html#.weak) attributes to `at_attributes`.
- Removed [D1 operator overloads](https://digitalmars.com/d/1.0/operatoroverloading.html) from `operator_overloads` because they have [not been supported](https://dlang.org/spec/operatoroverloading.html#old-style) for a very long time now.
- Added [missing operator overloads](https://dlang.org/spec/operatoroverloading.html) to `operator_overloads`.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -59,13 +59,13 @@ variables:
 
   construction_lookahead: '(?=~?\s*this\s*\()'
 
-  keyword: 'else|enum|export|extern|debug|default|delegate|delete|deprecated|do|body|break|case|cast|catch|class|const|continue|abstract|alias|align|asm|assert|auto|final|finally|for|foreach|foreach_reverse|function|goto|if|immutable|import|in|inout|interface|invariant|is|lazy|macro|mixin|module|new|nothrow|out|override|package|pragma|private|protected|public|pure|ref|return|scope|shared|static|struct|switch|synchronized|template|throw|try|typeid|typeof|union|unittest|version|while|with|__gshared|__traits|__vector|__parameters'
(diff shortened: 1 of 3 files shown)


deathaxe:
The `body` keyword seems to have been removed from specs with version 2.082 in 2018.

A syntax definition should be able to handle as many versions of a language as possible, to not break user experience with legacy code.

Same applies to other legacy keys, such as mensioned operator overloads. That's actually a good example for my arguments about the "usefullness" of such hard coded lists of keywords/constants and their implication with versions of a language.

They also should keep there as is to not break legacy code.

ichordev:
> Same applies to other legacy keys, such as [mentioned] operator overloads.
> [...]
> They also should keep there as is to not break legacy code.

The D1 operator overloads were marked as having ['no guarantee that these mechanisms will be supported in the future'](https://docarchives.dlang.io/v2.068.0/operatoroverloading.html#old-style) as of at least 9 years ago, which is more than half of D2's lifespan. I have never seen them in real code before, so please provide me some examples.

> A syntax definition should be able to handle as many versions of a language as possible, to not break user experience with legacy code.
>
> That's actually a good example for my arguments about the ["usefulness"] of such hard coded lists of keywords/constants and their implication with versions of a language.

Breaking changes in D have always been frequent; if you intend to support every version of D then your syntax highlighting will be unhelpful at best, and two-faced at worst.

On the other hand, it is incredibly useful when the syntax highlighting tells you whether or not you wrote the right thing, saving you looking it up.


> The `body` keyword seems to have been removed from specs with version 2.082 in 2018.

Actually, I just found out that it still works in place of `do`... but it's also not a keyword anymore, so this compiles:
```d
import std.stdio;

struct body{ //a struct named `body`
    int x;
}

void main()
in{
	assert(true);
}body{
    writeln("Hello world!");
}
```
So, I will add it back to the syntax for that.

PR #3976, opened 2024-05-01
[PHP] Fix file paths are recognized as regexes

jfcherng:
Reported on Discord: https://discord.com/channels/280102180189634562/280157083356233728/1235208897069973546

This PR mitigates the possibility that a path like `/foo/bar/` is recognized as a regex.

`/foo/bar/` is not a regex because there is unescaped `/` before it ends.

Yet still `/foo/` will be recognized as a regex anyway, even if you say `foo` is a directory under the root.


PHP/PHP Source.sublime-syntax
```diff
@@ -2576,12 +2576,14 @@ contexts:
...
       pop: 1
-    - match: \"
+    - match: /|"
       fail: branch-point-double-quoted
     - match: ''
       push: scope:source.regexp.php
       with_prototype:
-        - match: (?=/{{regex_modifier}}"|")
+        # for syntax test in regex
+        - match: // *\^.*$
```
jfcherng, line 2585: This is to deal with https://github.com/sublimehq/Packages/blob/41b7232d08850f9387c86b622dcdc949c188498d/PHP/tests/syntax_test_php.php#L5127-L5135 so that the test can continue to work without being failed due to unescaped `/` in a regex.

PHP/PHP Source.sublime-syntax
```diff
@@ -2576,12 +2576,14 @@ contexts:
...
       pop: 1
-    - match: \"
+    - match: /|"
       fail: branch-point-double-quoted
     - match: ''
       push: scope:source.regexp.php
       with_prototype:
-        - match: (?=/{{regex_modifier}}"|")
+        # for syntax test in regex
+        - match: ^// *\^.*$
```
deathaxe, line 2585: We should probably better refactor affected test cases instead of injecting non-regular patterns just to support existing ones.

Maybe ...

```php
<?php

$regex = '/
    a{,6}
/ux';
// <- meta.string.php string.quoted.single.php punctuation.definition.string.regex-delimiter.end.php
 // <- meta.string.php string.quoted.single.php meta.regex.modifier.php
//^ meta.string.php string.quoted.single.php meta.regex.modifier.php
// ^ meta.string.php string.quoted.single.php punctuation.definition.string.end.php

$regex = '/
    a{,6}
    b{3,} # this is comment
    c{3,6}/ux';
//   ^^^^^ keyword.operator.quantifier.regexp
```

jfcherng: I think I found another bug. 

```php
$regex = '/a# this is comment/ux'; // matches 'a'
$foo = preg_replace($regex, 'b', 'aaaaaa');
echo $foo; // bbbbbb, so "# this is comment" is indeed a comment
```

```php
$regex = '/
    c{3,6} # this is comment/ux';
//   ^^^^^ keyword.operator.quantifier.regexp
//         ^^^^^^^^^^^^^^^^^ comment (at this moment this failed)
```

jfcherng: > I think I found another bug.
> 
> ```
> $regex = '/a# this is comment/ux'; // matches 'a'
> $foo = preg_replace($regex, 'b', 'aaaaaa');
> echo $foo; // bbbbbb, so "# this is comment" is indeed a comment
> ```
> 
> ```
> $regex = '/
>     c{3,6}# this is comment/ux';
> //   ^^^^^ keyword.operator.quantifier.regexp
> //        ^^^^^^^^^^^^^^^^^ comment (at this moment this failed)
> ```

Resolved in 35bebe99c579d2a7ecedef041af5f8321051db20

PHP/PHP Source.sublime-syntax
```diff
@@ -2578,14 +2578,14 @@ contexts:
...
       push: scope:source.regexp.php
       with_prototype:
-        - match: (?=/{{regex_modifier}}"|")
+        - match: (?=/|")
           pop: 1
-        - match: \s((#)[^"]*$\n?)
+        - match: ((#)[^"]*(?:$\n?|(?=/)))
           captures:
             1: comment.regexp.php
             2: punctuation.definition.comment.regexp.php
```
deathaxe, line 2591: If comments don't need to be separated by space, we can probably drop one capture group.

Consuming up to `/` can also be simplified somewhat by adding it to the negative charset.

```suggestion
        - match: (#)[^/"]*$\n?
          scope: comment.regexp.php
          captures:
            1: punctuation.definition.comment.regexp.php
```

Actually it however might on the other hand be somewhat surprising to loose regexp highlighting just due to a slash in a comment.

grafik

```php
<?php

$pattern = '/
(?x)
\d+# comm/ent
/x';
```

jfcherng: Hmm... interesting. The comment can't have unescaped `/` either.

deathaxe: Pattern simplification also applies to single quoted patterns.

jfcherng: It turns out the suggested change doesn't pass the following test.

```php
$regex = '/
    c{3,6}#this is comment/ux';
//        ^^^^^^^^^^^^^^^^ comment.regexp
```

PHP/PHP Source.sublime-syntax
```diff
@@ -2578,17 +2578,17 @@ contexts:
...
+    - match: /|"
       fail: branch-point-double-quoted
     - match: ''
       push: scope:source.regexp.php
       with_prototype:
-        - match: (?=/{{regex_modifier}}"|")
+        - match: (?=/|")
           pop: 1
-        - match: \s((#)[^"]*$\n?)
+        - match: (#)[^/"]*($\n?|(?=/))
```
deathaxe, line 2588: It appears the negative charset already includes a possible `\n` at eol, but excludes trailing `/` or `"`. Thus pattern can be simplified.

```suggestion
        - match: (#)[^/"]*
```

PHP/PHP Source.sublime-syntax
```diff
@@ -2659,17 +2659,17 @@ contexts:
...
+    - match: /|'
       fail: branch-point-single-quoted
     - match: ''
       push: scope:source.regexp.php
       with_prototype:
-        - match: (?=/{{regex_modifier}}'|')
+        - match: (?=/|')
           pop: 1
-        - match: \s((#)[^']*$\n?)
+        - match: (#)[^/']*($\n?|(?=/))
```
deathaxe, line 2669: Simplification also possible here:

```suggestion
        - match: (#)[^/']*
```

deathaxe (commented):
Two more ideas. Otherwise LGTM.

deathaxe (approved):

michaelblyons (approved):

PR #3980, opened 2024-05-05
[HTML] Add variable for custom tag start character

deathaxe:
This commit adds a `tag_name_start` variable to represent a custom tag's first character across syntax definitions instead of hard-coding `[A-Za-z]`.

It enables template syntaxes to override such tags' names easier, if they begin with a template interpolation.

Example:

    <{{tag_name}} ...>

To scope `{{tag_name}}` entity.name.tag, a template syntax has needed to extend the `tag-other` context before this commit with something like:

    tag-other:
      - meta_prepend: true
      - match: </?(?={{) 
        scope: punctuation.definition.tag.begin.html 
        push:
          - tag-other-content
          - tag-other-name

even though `{{ ... }}` tags are already part of the `prototype` context.

With this commit it is enough to override the `tag_name_start` to add the template begin punctuation in order to achieve the same outcome.

    variables:
      tag_name_start: (?:[A-Za-z]|{{)

> [!NOTE]
>
> The new variable is added to all template syntaxes such as ASP, ERB, JSP and PHP in order to enable support for tag names, beginning with template interpolation.

### DIFF
--- a/HTML/HTML (Plain).sublime-syntax
+++ b/HTML/HTML (Plain).sublime-syntax
@@ -60,6 +60,24 @@
 #          pop: 1
 #        - ...
 #
+#  2 TAG NAME INTERPOLATION
+#
+#    Default HTML tags start with a case insensitive latter a..z.
+#
+#    If a template language supports interpolation within HTML tag names,
+#    it must also ensure a tag being identified as such, if its name starts
+#    with a template tag by overriding the `tag_name_start` variable.
+#
+#    Example:
+#
+#      variables:
+#        # make `{{` a valid tag name start char
+#        tag_name_start: (?:[A-Za-z]|{{)
+#
+#    Enables correct scoping of:
+#
+#        <{{tag_name}} ... >
+#
 ###############################################################################
 # https://www.sublimetext.com/docs/syntax.html
 # https://html.spec.whatwg.org/multipage/syntax.html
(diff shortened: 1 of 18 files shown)


michaelblyons (approved):
I like this a lot.

HTML/HTML (Plain).sublime-syntax
```diff
@@ -88,7 +88,8 @@ variables:
   tag_name_break_char: '[{{ascii_space}}/<>]'
   tag_name_break: (?={{tag_name_break_char}})
   tag_name_char: '[^{{tag_name_break_char}}]'
-  tag_name: '[A-Za-z]{{tag_name_char}}*'
+  tag_name_start: '[A-Za-z]'
```
FichteFoll, line 91: How about adding a comment here that explains the intent of this variable, i.e. being used for extensions? Will make it more obvious in the future that this is something external syntaxes (or internal even) actually depend on and reduces the likelyhood of changing/removing it by accident.

FichteFoll (dismissed):

HTML/HTML (Plain).sublime-syntax
deathaxe: Added a note and some docs at the beginning of the HTML (Plain) syntax.

michaelblyons (approved):

HTML/HTML (Plain).sublime-syntax
```diff
@@ -60,6 +60,24 @@
 #          pop: 1
 #        - ...
 #
+#  2 TAG NAME INTERPOLATION
+#
+#    Default HTML tags start with a case insensitive latter a..z.
```
FichteFoll, line 65: `case-insensitive letter`

HTML/HTML (Plain).sublime-syntax
```diff
@@ -60,6 +60,24 @@
...
+#
+#    If a template language supports interpolation within HTML tag names,
+#    it must also ensure a tag being identified as such, if its name starts
+#    with a template tag by overriding the `tag_name_start` variable.
+#
+#    Example:
+#
+#      variables:
+#        # make `{{` a valid tag name start char
```
FichteFoll, line 74: `start sequence`

FichteFoll (approved):

michaelblyons (approved):

PR #3982, opened 2024-05-06
[Rust] Fix identifiers in let-based pattern matching

### LINKED ISSUE #3981: [Rust] Pattern matching struct or enum highlighted as regular identifier
### What happened?

Examples:

```
let Some(a) = b;
if let Some(a) = b { ... }
```

The `Some` here is not highlighted as a builtin-enum. The same is true for other types as well.

BenjaminSchaaf:
Fixes #3981


keith-hall (approved):

FichteFoll (approved):
There's probably a better (more accurate but also elaborate) way to do this, but this gets the job done.

PR #3983, opened 2024-05-06
[CSS] Remove completion restrictions

deathaxe:
Suggest properties also within `meta.selector` as this is the most likely scope in front of nested selectors, if a property is not yet terminated by semi-colon.

As completions are much more likely, do no longer suppress word completions.

Without this commit auto-completion no longer provides property-name completions in front of nested selectors.

Example:

```css
div {
  color: red;
  bo|  <- no more completions here

  h2 {
    text-decoration: none;
  }
}
```

reported at https://github.com/SublimeText/Sass/issues/107

### DIFF
--- a/CSS/CSS.sublime-settings
+++ b/CSS/CSS.sublime-settings
@@ -4,7 +4,7 @@
 
     // Controls what scopes default completions will be provided in.
     // Can be a list of strings which are joined before matching.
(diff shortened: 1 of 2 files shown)


deathaxe:
Anything I can do for a step forward?

keith-hall (approved):

deathaxe:
Thanks!

PR #3992, opened 2024-06-17
[Batch File] Fix comment scope around newlines

mataha:
Multiple colons immediately followed by a non-empty line cause said line
to be scoped as a comment:

```batchfile
::: Set some variables
:::
set WSL_UTF8=1
```

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -161,7 +161,7 @@ contexts:
   comment-body:
     - meta_scope: comment.line.colon.dosbatch
     - include: line-continuations
-    - match: \n
+    - match: $\n?
       pop: 1
 
   ignored-tail-inner:
@@ -843,7 +843,7 @@ contexts:
     # meta_content_scope is used since rem should not be
     # highlighted as a comment, but a command
     - meta_content_scope: comment.line.rem.dosbatch
-    - match: \n
+    - match: $\n?
       scope: comment.line.rem.dosbatch
       pop: 1
 
--- a/Batch File/tests/syntax_test_batch_file.bat
+++ b/Batch File/tests/syntax_test_batch_file.bat
@@ -38,13 +38,18 @@
 :: ^^^^^^^^^^^^^^^^^ comment.line.rem.dosbatch
 
    REM ^
+
+   I'm a (com|ment)
+:: ^^^^^^^^^^^^^^^^^ comment.line.rem.dosbatch
+
+   REM ^
    I'm a (com|ment) ^
    not a comment
-:: ^^^^^^^^^^^^^ - comment
+:: ^^^^^^^^^^^^^^ - comment
 
 REM
    not a comment
-:: ^^^^^^^^^^^^^ - comment
+:: ^^^^^^^^^^^^^^ - comment
 
 REM This follows a REM command
 :: <- keyword.declaration.rem.dosbatch - comment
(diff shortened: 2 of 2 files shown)


Batch File/Batch File.sublime-syntax
```diff
@@ -161,7 +161,7 @@ contexts:
   comment-body:
     - meta_scope: comment.line.colon.dosbatch
     - include: line-continuations
-    - match: \n
+    - match: $
```
keith-hall, line 164: Generally the trailing new line char in line comments is deliberately scoped as a comment so that auto-complete  and snippets etc wouldn't trigger.

What problem are you actually solving here? Do.many of the new test assertions fail without this change?

Did you consider:
```suggestion
    - match: $\n?
```

FichteFoll: `\n` or `$\n?` shouldn't make any different practically.

That said, `\n` being part of the `comment` scope is indeed on purpose.

mataha: > What problem are you actually solving here? Do.many of the new test assertions fail without this change?

There are two issues:
 - multiple colons followed immediately by a non-empty line cause said line to be scoped as a comment
 - `\n` on an empty line can itself be escaped, making the following lines incorrectly scoped

> Did you consider:

Thanks for the tip, it slipped my mind. That solves the first issue.

deathaxe: > multiple colons followed immediately by a non-empty line cause said line to be scoped as a comment

... by intention as the linked ss64 site defines it as such.

see: https://github.com/sublimehq/Packages/blob/3b7c0d35953d66d452fa139fdcc0a3f3689dae15/Batch%20File/Batch%20File.sublime-syntax#L156

deathaxe: With regards to `\n` vs. `$\n?`, the former one is enough when using batch scripts on its own, the latter one may be more safe/reliable in cases of batch being extended in templating syntaxes, which may inject interpolation contexts into comments, which may already consume `\n`.

Batch File/Batch File.sublime-syntax
```diff
@@ -1967,7 +1967,15 @@ contexts:
 ###[ PROTOTYPES ]#############################################################
 
   bol-pop:
-    # Note: empty lines are ignored
+    - match: \n
+      scope: constant.character.escape.linefeed.dosbatch
```
deathaxe, line 1971: A normal `\n` is not an escape character.

As in Bash's line continuation, only the next line after `^` belongs to current line. Arbritary empty lines may not be ignored.

michaelblyons: > at most a single empty line in between.

What a horrible spec. 🤦

Batch File/Batch File.sublime-syntax
mataha: > > multiple colons followed immediately by a non-empty line cause said line to be scoped as a comment
> 
> ... by intention as the linked ss64 site defines it as such.
> 
> see:
> 
> https://github.com/sublimehq/Packages/blob/3b7c0d35953d66d452fa139fdcc0a3f3689dae15/Batch%20File/Batch%20File.sublime-syntax#L156

I've read that multiple times and don't know what you're alluding to. It's plain wrong anyway:

```batchfile
@echo off
::: comment.bat
:::
echo Not a comment
```

```console
> comment.bat
Not a comment
```

```batchfile
@echo off
rem comment2.bat
rem
echo Not a comment
```

```console
> comment2.bat
Not a comment
```

Batch File/Batch File.sublime-syntax
mataha: > A normal `\n` is not an escape character.

Except in special circumstances, where it's used to capture a linefeed:

```batchfile
::: newline.bat
@setlocal EnableDelayedExpansion & echo off
(set ^"\=^

)
echo Hi.!\!Hello.
```

```console
> newline.bat
Hi.
Hello.
```

Or escape it and use it to construct macros:

```batchfile
::: newline2.bat
@setlocal DisableDelayedExpansion & echo off
(set ^"\n=^^^

)
set ^"hi=(%\n%
    echo Hello%\n%
    echo World!%\n%
)"
%hi%
```

```console
> newline2.bat
Hello
World!
```

> As in Bash's line continuation, only the next line after `^` belongs to current line with at most a single empty line in between. Arbritary empty lines may not be ignored.
> 
> ```
> echo 12
> :: comment^
> 
> echo 13
> echo 14
> ```
> 
> prints ...
> 
> 12 14
> 
> while
> 
> ```
> echo 12
> :: comment^
> 
> 
> echo 13
> echo 14
> ```
> 
> prints...
> 
> 12 13 14

That's not how this mechanism works. When `^` is present and not within quotes, the next token loses its special meaning (if any) and `\n` is no exception (see above).

Batch File/Batch File.sublime-syntax
deathaxe: Snippet 1 is not handled correctly, currently. `echo` must not be scoped comment. `:: ` is ignored as not being a valid label or command. These may not be comments in the first place, but seem to be one of those common constructs for such. I actually only use these kinds rather than horrible "rem".

Snippet 2 and 4 show valid redirections, which may not have an effect due to nothing being in front of them. Those are however no comments.

Snippet 4 is working as expected.

Batch File/Batch File.sublime-syntax
deathaxe: > What a horrible spec. 🤦

Actually it was created without a spec. It is - as child of the 80s - just a mess with lot's of ugly "workarounds" which interpret language constructs into the syntax, which actually don't exist. Comments is one of those as we learn here.

Actually I wonder why someone would still use it for anything else but automation of trivial tasks.

The information about `^` are interesting.

`%= =%` however is an invalid or non-existing variable - just another workaround to mimic comments as `::` are, which are built on the fact cmd ignoring sorts of invalid expressions.

deathaxe: > > A normal \n is not an escape character.
>
> Except in special circumstances, where it's used to capture a linefeed: ...

Well, not according to the linked stackoverflow comment.

First `\n` after `^` is stripped. If 2nd character after `^` is an `\n`, too, it is treated literal, which means, it does not terminate the line after initial `^\n`.

That's why 

```cmd
@echo Hello^

World!
```

prints to 

```
Hello
World!
```

instead of interpreting `World!` as command.

That's also the reason why "label like comments" can contain an empty 2nd line.

```cmd
:: comment^

2nd line of comment
echo No comment
```

Due to circumstances described in phase 1, the empty line can even contain spaces or tabs.


However `rem` behaves inconsistent with regards to that.

If a comment consists of only a single token, it works as expected

```
rem token^

2nd line
echo No comment
```

If however it consists of multiple tokens, it does not.

```
rem token1 token2^

echo No comment
```

Well, I am not sure if we really want to handle all the special cases of that syntax, especially as a "single phase" parser such as ST's syntax engine can barely handle all edge cases produced by "multi phase" parsers with sane efforts. That's also true for Markdown.

Batch File/Batch File.sublime-syntax
mataha: "Snippets" 2 and 4 are just console invocations of snippets 1 and 3.

I'm using them in favor of `rem` as well, though not everywhere - in some places `:::` is invalid (such as in parenthesized blocks).

Note that `:::` is slightly safer than `::` due to `subst`.

mataha:
This PR has gotten quite unwieldy, so I've limited it to newline continuations only.

Batch File/Batch File.sublime-syntax
mataha: > Actually I wonder why someone would still use it for anything else but automation of trivial tasks.

For the same reason people use POSIX shells - to write simple, ageless scripts that will work anywhere, and can serve as an entry point to more complicated programs.

> The information about `^` are interesting.
> 
> `%= =%` however is an invalid or non-existing variable - just another workaround to mimic comments as `::` are, which are built on the fact cmd ignoring sorts of invalid expressions.

Yes, and it's mainly used in cases like these, where a line must be empty at evaluation time.

> First `\n` after `^` is stripped. If 2nd character after `^` is an `\n`, too, it is treated literal, which means, it does not terminate the line after initial `^\n`.

Indeed, and that's why `\n` is special in this context - because the first one is stripped.

> Due to circumstances described in phase 1, the empty line can even contain spaces or tabs.

That's not true. How did you come to that conclusion?

> Well, I am not sure if we really want to handle all the special cases of that syntax, especially as a "single phase" parser such as ST's syntax engine can barely handle all edge cases produced by "multi phase" parsers with sane efforts. That's also true for Markdown.

I am of the same opinion - handling every case is impossible without something like semantic highlighting. That doesn't mean I can't at least try...

deathaxe (approved):

deathaxe:
> That's not true. How did you come to that conclusion?

Well, just by playing trick on my self with a test script and `"trim_trailing_white_space_on_save"` setting being true, which removed it without me noticing it 🤣 .

michaelblyons (approved):

PR #3995, opened 2024-06-20
[Batch File] Add snippets

mataha:
The most common patterns only, i.e. no metaprogramming or structural code.

Specifically I've avoided adding snippets for subroutines and macros, as conventions for these vary wildly.

### DIFF
--- a/Batch File/Snippets/call.sublime-snippet
+++ b/Batch File/Snippets/call.sublime-snippet
@@ -1,0 +1,6 @@
+<snippet>
+	<tabTrigger>ca</tabTrigger>
+	<description>call …</description>
+	<scope>source.dosbatch - comment - string</scope>
+	<content><![CDATA[call $0]]></content>
+</snippet>
--- a/Batch File/Snippets/echo-off.sublime-snippet
+++ b/Batch File/Snippets/echo-off.sublime-snippet
@@ -1,0 +1,6 @@
+<snippet>
+	<tabTrigger>eo</tabTrigger>
+	<description>echo off</description>
+	<scope>source.dosbatch - comment - string</scope>
+	<content><![CDATA[echo off]]></content>
+</snippet>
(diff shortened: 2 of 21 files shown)


deathaxe:
Snippets are often rather related with personal taste. It's probably better to provide them as 3rd party package for those who want them.

For a general purpose package it is probably rather risky to provide snippets which use single character triggers such as `c` or those which match keywords `ge`. It may cause false positives (and resulting complaints) and requires some ammount of muscle memory to get used to such heavy shortcuts.

mataha:
What about snippets already present in this repository? Are they grandfathered in?

Good call on false positives - I've removed those. I still think there's value in `s` (`set`) and `e` (`echo`) though (I was referencing Java print and variable snippets when I wrote them) as those are used the most by a large margin, even in codebases where batch scripts are simple few-liners and serve to supplement building/running a project.

deathaxe:
> What about snippets already present in this repository? Are they grandfathered in?

For various of them answer is probably "yes".

Those helping with saving key strokes to add ugly to type template tags (e.g.: `if` expanding to `<?php if ... ?>`) in e.g. _PHP_ or _Go Templates_ are rather crucial and maintained.

The question which kind of snippets to add is rather hard to answer. Users' requirements differ too much to give the ultimate answer. They should probably really add value, be crucial to a syntax, be no target of frequent changes, easy to discover and offer low risk of conflicts with normal keywords or completions.

_I am personally for instance not a fan of 1-3 char triggers, even though I see their intent value. But that's just me. Others may have different oppinions about it._

Some complain about them being present at all, as they rather prefer to rely on LSP these days. The range is huge.

The `echo off` snippet for instance uses `eo` trigger, which just expands to `echo` keyword when hitting <kbd>tab</kbd> with my preferences. So it doesn't help me much as I still need to <kbd>ctrl+space</kbd> and use arrow keys to reliably add it. That's not faster than typing those 8 chars - to give one example.

Therefore it is probably a better choice to provide a "Batch Snippets" package, which can be installed by those who want it. And it's probably easier and faster to update and can be made available for earlier ST releases.

mataha:
https://github.com/sublimehq/Packages/issues/1988#issuecomment-500285962 has convinced me that this PR might not be the best idea.

PR #3999, opened 2024-06-22
[Batch File] Fix line continuation

deathaxe:
This commit applies the following rule:

If `^` appears at the end of a line, the linefeed is stripped and the next character is escaped or uses its special meaning, if any. If the next character is also a linefeed, it is treated literal.

REM command seems to be working special by line continuation working only directly after the first token.

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -2004,7 +2004,10 @@ contexts:
     # Note: Must consume \n to work.
     - match: ^\n
       set: bol-pop
-    - include: bol-pop
+    # The first caret at a continued line is escaped by the continuation caret
+    # and thus printed literal.
+    - match: ^\^?
+      pop: 1
 
   eoc-pop:
     # Note: An end of command appears in unquoted regions only!
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -1981,6 +1981,7 @@ contexts:
 ###[ PROTOTYPES ]#############################################################
 
   bol-pop:
+    - meta_include_prototype: false
     - match: ^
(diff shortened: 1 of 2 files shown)


Batch File/Batch File.sublime-syntax
```diff
@@ -1983,7 +1995,14 @@ contexts:
   line-continuations:
     - match: \^\n
       scope: punctuation.separator.continuation.line.dosbatch
```
mataha, line 1997: Wouldn't marking the caret be visually clearer? Even in comments.

```suggestion
    - match: (\^)\n
      scope: punctuation.separator.continuation.line.dosbatch
      captures:
        1: constant.character.escape.dosbatch
```

Batch File/Batch File.sublime-syntax
```diff
@@ -1983,7 +1995,14 @@ contexts:
...
-      push: bol-pop
+      push: line-continuation-body
+
+  line-continuation-body:
+    # The first linefeed diretly following a continuation is treated literal.
+    # Note: Must consume \n to work.
+    - match: ^\n
```
mataha, line 2003: I believe this doesn't solve the "leading caret after a continuation" problem:

```batchfile
::: `@` is not escaped here, the second `^` is
:::^
^
@echo This shouldn't be marked as a comment.
```

```suggestion
    - match: ^\^
      pop: 1
    - match: ^\n
```

I'm not sure whether the caret should be scoped as an escape character (as it will be styled even within comments).

mataha (commented):
Cleaner than what I was attempting. I have a few comments though.

mataha (approved):
Looks good.

michaelblyons (approved):

FichteFoll (approved):

PR #4000, opened 2024-06-22
[Batch File] Fix empty comments exceeding to next line

deathaxe:
This commit fixes "label comment" related patterns, to only scope leading colons `punctuation.definition.comment`.

Label comments make use of invalid labels being ignored. A label is invalid, if the first character after leading colon is a "metachar".

A label starting with `::` is already an invalid label and thus can be followed by any (meta) char to still be a "valid" comment.

The wrong pattern was the real reason for the issue addressed by PR #3992.

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -48,7 +48,7 @@ variables:
   eoc: (?=\s*{{eoc_char}})
   eoc_char: '[\n|&]'
 
-  label_comment: ':+[^{{label_start}}]'
+  label_comment: ':(?::+|(?!{{label_start}}))'
   label_start: '[^{{metachar}}:+]'
 
   path_terminator_chars: '[\s,;"{{redir_or_eoc_char}}]'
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):

Batch File/Batch File.sublime-syntax
```diff
@@ -48,7 +48,7 @@ variables:
   eoc: (?=\s*{{eoc_char}})
   eoc_char: '[\n|&]'
 
-  label_comment: ':+[^{{label_start}}]'
+  label_comment: ':(?::+|(?!{{label_start}}))'
   label_start: '[^{{metachar}}:+]'
```
mataha, line 52: Can I suggest the following (would require a minor fix in tests):

```suggestion
  label_comment: ':(?:{{label_terminator_chars}}+|(?!{{label_start}}))'
  label_start: '[^{{metachar}}{{label_terminator_chars}}]'
  label_terminator_chars: '[:+]'
```

As `+` (and technically `\s` but probably not relevant here) terminate a label as well ([source](https://www.dostips.com/forum/viewtopic.php?f=3&t=3803)).

deathaxe: After a closer look the whole topic seems to be way more complicated than expected.

The following is a valid `label` for instance.

```
:== label ==
```

I guess we should merge #3999 first as this one requires some larger changes, which requiere some of the other changes.

PR #4003, opened 2024-06-29
[JavaScript] Fix ST freezing

deathaxe:
Fixes https://github.com/sublimehq/sublime_text/issues/6416

Regressed by #3836

This commit refactors for loop conditions to prevent ST's syntax engine to run into an infinite loop, if multiple nested for loops appear within branches, such as arrow functions.

The issue seems to be triggered by

```yml
  for-condition-contents:
    - match: ''
      branch_point: for-in-of
      branch:
        - for-in-of
        - for-oldstyle
      pop: 1
```

in conjunction with newlines.

Note: It somewhat reorganizes expressions contexts a little bit.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -842,31 +842,25 @@ contexts:
     - include: else-pop
 
   for-condition:
-    - match: '\('
-      scope: punctuation.section.group.begin.js
-      set:
-        - for-condition-end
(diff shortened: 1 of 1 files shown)


Nefsen402:
I was able to confirm that this patch fixes the syntax highlighting errors for me. Thanks!

keith-hall (approved):

jfcherng (approved):

PR #4005, opened 2024-07-09
[Batch File] Refactor labels and label-like comments

deathaxe:
This PR applies tokenization rules from https://stackoverflow.com/questions/4094699/how-does-the-windows-command-interpreter-cmd-exe-parse-scripts/4095133#4095133 to labels and common command arguments.

As a result invalid labels are scoped "ignored" even without dedicated comment contexts and less special-casing patterns. They are however kept to maintain `comment.line` scopes for now.

Delimiter characters such as `=` are ignored, which fixes scoping of somewhat uncommon but valid labels like `:=== label ===`.

Delimiter chars are scoped `punctuation.separator` if not having any other special meaning (e.g.: `=` `keyword.operator.assignment`) depending on context.

Accuracy of various tokens - including some odds of batch - is increased to match highlighting better to interpreter's behavior.

### DIFF
--- a/Batch File/Batch File (Compound).sublime-syntax
+++ b/Batch File/Batch File (Compound).sublime-syntax
@@ -21,4 +21,4 @@ hidden: true
 extends: Packages/Batch File/Batch File.sublime-syntax
 
 variables:
-  eoc_char: '[\n|&)]'
+  eoc_char: \)\n\|&
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -69,9 +69,6 @@ variables:
   # A label variable is therefore terminated by `+`, `:` or word break (`metachar`)
   label_variable_break: (?![^:+{{metachar}}])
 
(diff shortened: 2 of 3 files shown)


mataha (commented):
This doesn't handle the "first character in a line can be anything" rule. Is that intentional?

deathaxe:
Any example of what's not working as expected would be helpful. Labels are treated special in some manners. Maybe I missed something.

mataha:
```batchfile
@echo off

for %%# in (1 2 3) do (call :label%%~#)
exit /b

@:label1 is a valid label unless in label context

$,=,= ,==, =,;;:,;=label2 see above, also 2>nul to suppress errors

*       :          label3 see above, also 2>nul to suppress errors

(echo(%0)
goto :EOF
```

deathaxe:
Well, I don't see any value it supporting all sorts of obscure expressions and doubt it is even possible.

FichteFoll (approved):
Batch parsing is really weird and I definitely did not read the referenced SO post, but these look like good changes to me.

michaelblyons (approved):

PR #4007, opened 2024-07-12
[Markdown] Match code block syntax names with Linguist aliases

mataha:
Mainly because `batchfile` didn't work for batch files...

This commit adds most of Linguist (GH language detector - [source][1]) aliases for languages for which fenced code block contexts exist in Markdown grammar.

I've omitted a few aliases because these name separate languages with their own third-party packages (notably `cake`, `octave` and some `sh` variants).

I've left ActionScript alone as I wasn't sure whether the grammar covers as3.

[1]: https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml

### DIFF
--- a/Markdown/Code Block Syntaxes.sublime-completions
+++ b/Markdown/Code Block Syntaxes.sublime-completions
@@ -18,13 +18,19 @@
 
 		// AppleScript
 		{
-			"trigger": "AppleScript",
+			"trigger": "applescript",
 			"annotation": "AppleScript",
 			"kind": ["markup", "s", "Syntax"],
 			"details": "Specifies <code>AppleScript</code> code highlighting"
 		},
 		{
 			"trigger": "osascript",
+			"annotation": "AppleScript",
+			"kind": ["markup", "s", "Syntax"],
+			"details": "Specifies <code>AppleScript</code> code highlighting"
+		},
+		{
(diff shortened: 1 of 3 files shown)


Markdown/Code Block Syntaxes.sublime-completions
```diff
@@ -143,20 +155,50 @@
...
+		},
 		{
 			"trigger": "cmd",
 			"annotation": "DOS batch",
 			"kind": ["markup", "s", "Syntax"],
 			"details": "Specifies <code>DOS batch</code> code highlighting"
 		},
 		{
-			"trigger": "dos",
+			"trigger": "cmd.exe",
```
deathaxe, line 177: Maybe we should provide completions only for a sane subset of supported tags.

mataha: Is this better? This matches Batchfile entry in Linguist:

```yml
Batchfile:
  type: programming
  aliases:
  - bat
  - batch
  - dosbatch
  - winbatch
  extensions:
  - ".bat"
  - ".cmd"
```

deathaxe (approved):

keith-hall (approved):

deathaxe:
Thanks for improvements!

PR #4008, opened 2024-07-12
[Git Formats] Support inline comments in CODEOWNERS

mataha:
Live example can be found [here][1].

[1]: https://github.com/gradle/gradle/blob/master/.github/CODEOWNERS#L18

### DIFF
--- a/Git Formats/Git Code Owners.sublime-syntax
+++ b/Git Formats/Git Code Owners.sublime-syntax
@@ -11,10 +11,12 @@ version: 2
 file_extensions:
   - CODEOWNERS                # *.codeowners, ./CODEOWNERS
 
+variables:
+  comment_char: \#
+
 contexts:
 
   main:
-    - include: Git Common.sublime-syntax#comments
     # note:
     # - ignore whitespace at the beginning of a line
     # - patterns may not start with operators: - !
@@ -28,6 +30,7 @@ contexts:
   pattern-content:
     # note: character classes are not supported
     - meta_content_scope: meta.path.pattern.git.codeowners entity.name.pattern.git.codeowners
+    - include: comments
     - match: \s+(?=([^\@\s]+)?\@)
       set: owners
     - include: Git Common.sublime-syntax#fnmatch-unquoted-file-extensions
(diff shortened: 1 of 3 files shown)


michaelblyons:
I think typically we match the comment character, scope it `punctuation`, and push into a `meta_scope`d context that is `pop`d by NL/EOL. Are you doing differently to avoid scoping the whitespace before a comment as a codeowner username?

mataha:
Yes, except the whitespace before a comment gets scoped as `meta`.

Git Formats/Git Code Owners.sublime-syntax
```diff
@@ -46,6 +50,18 @@ contexts:
...
+      set: comment-body
+
+  comment-body:
+    - meta_include_prototype: false
+    - match: ({{comment_char}}).*\n?
+      scope: comment.line.git.codeowners
+      captures:
+        1: punctuation.definition.comment.git.codeowners
+      pop: 1
+
```
michaelblyons, line 64:
```suggestion
  comments:
    - match: \s*(?={{comment_char}})
      set: comment-start

  comment-start:
    - meta_include_prototype: false
    - match: {{comment_char}}
      scope: punctuation.definition.comment.git.codeowners
      push: comment-body

    comment-body
      - meta_scope: comment.line.git.codeowners
      - match: \n?$
        pop: 1
```

michaelblyons: How about this?

mataha: - Shouldn't the `push` be a `set`? `comment-start` will remain on top of the stack otherwise.
- Probably doesn't matter, but why `\n$?` instead of `$\n?`?

deathaxe: Solution may be way simpler, by just tweaking word boundary patterns

see: https://github.com/mataha/Packages/pull/1

michaelblyons (dismissed):

Git Formats/Git Code Owners.sublime-syntax
mataha: Would changing `#;` to `#` only in CODEOWNERS make sense?

deathaxe: This would mean to duplicate comment and email implementation. IMHO chances for an email or name to contain ; in its name is low enough to avoid those extra efforts.

deathaxe (approved):

michaelblyons (approved):

PR #4009, opened 2024-07-13
[Mixed] Use detailed scope for true/false/null

jfcherng:
This is good for those who want to use different colors for true/false.

---

This PR changes scopes of booleans in most languages.

Splits scope `XXX.boolean.LANG` into

- `XXX.boolean.true.LANG`
- `XXX.boolean.false.LANG`

Some also introduce new detailed scope for `null`/`nil`.


jfcherng:
I think this has been done.

keith-hall (approved):

deathaxe (approved):
Seems complete after the first glance. Even if not, we can always add remainers.

Thanks.

PR #4013, opened 2024-07-23
[C/C++] Fix broken tests

BenjaminSchaaf:
These tests weren't being run because the `/*` needs to be on the same line as the test.


keith-hall (approved):

deathaxe (approved):

michaelblyons:
Did your new test code find them?

PR #4014, opened 2024-07-24
[Go] Add support for YAML Go templates

deathaxe:
This PR...

1. fixes missing `source.json.go` scope in snippet selectors
2. adds _YAML (Go).sublime-syntax_ with YAML highlighted Go Templates

### DIFF
--- a/Go/Go.sublime-syntax
+++ b/Go/Go.sublime-syntax
@@ -747,8 +747,24 @@ contexts:
     - match: \%(?:\[\d+\])?[ .\d*#+-]*[A-Za-z]
       scope: constant.other.placeholder.go
 
+  match-block-string-templates:
+    # This context is used for interpolation within quoted or unquoted multiline strings.
+    # Doesn't check for closing `}}` as it might be located on a sub-sequent line.
+    # It clears the `string` scope of the owning context.
+    # Note: used by YAML (Go).sublime-syntax
+    - match: ({{)(-?)
+      captures:
+        1: punctuation.section.interpolation.begin.go
+        2: keyword.operator.template.trim.left.go
+      push:
+        - pop-now-clear-scope
+        - match-template-inner
+        - match-template-function
+        - match-template-keyword
+
   match-string-templates:
     # This context is used for interpolation within quoted or unquoted strings.
+    # Matchs only if closing `}}` is found on same line to avoid false positives.
     # It clears the `string` scope of the owning context.
     - match: ({{)(-?)(?=.*?}})
       captures:
--- a/Go/Snippets/Template/block-end.sublime-snippet
+++ b/Go/Snippets/Template/block-end.sublime-snippet
@@ -3,6 +3,6 @@
 $0
 ${TM_TEMPLATE_START}end${TM_TEMPLATE_END}]]></content>
 <tabTrigger>tblockend</tabTrigger>
-<scope>(text.html.go | text.html.markdown.go | source.css.go | source.js.go | source.go string) - source.go.template</scope>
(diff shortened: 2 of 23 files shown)


FichteFoll (approved):
Works well for me.

FichteFoll:
As for where this is used: I've come across YAML with Go templates in the context of Helm and Rancher.

FichteFoll:
Actually, multi-line comments in blocks are not recognized (not sure if this is a common syntax):
```
{{/* 
Spread a key-value map into the "env" list 
*/}}
```

Furthermore, I think the string after `define` should probably be an `entity.name` subscope, i.e.
```
{{/* Spread a key-value map into the "env" list */}}
{{- define "chart.env-vars" -}}
{{- range $key, $val := . }}
- name: {{ $key }}
  value: {{ quote $val }}
{{- end }}
{{- end -}}
```
Currently it's a string.

keith-hall (approved):

deathaxe:
> Actually, multi-line comments in blocks are not recognized

That's probably caused by `match-string-templates` being used, which is designed for interpolating single-line strings and is used here as it clears the `string` scope. It intents to avoid highlighting unclosed `{{` in strings interpolation.

Same snippet is highlighted fine in Markdown, which uses `match-text-templates` primarily, but doesn't clear any scopes.

Maybe YAML needs a special casing to also support block strings.

deathaxe:
Now using a string interpolation context without safe guards (no check for closing `}}`) in YAML. As we are explicitly using it in a template syntax, it should be safe enough.

deathaxe:
> ... the string after `define` should probably be an `entity.name` subscope

Well, Go template syntax doesn't treat certain commands' arguments special by any means, currently to keep it lean.

According to https://docs.gomplate.ca/syntax/#nested-templates `define` is used do define inline templates. The counter-command `template` also treats first argument (the template name) as plain string.

grafik

Haven't investigeted which else commands would be affected.

That's certainly a separate feature request, not related with supporting YAML.

FichteFoll:
> That's certainly a separate feature request, not related with supporting YAML.

Yes, certainly.

FichteFoll (dismissed):

FichteFoll:
The lines with a sequence of `: ` get scoped as `meta.mapping.key` by the YAML syntax even though the part where this occurs is part of the embedded section:
```
{{/*
Unroll a list of default environment variable settings
*/}}
{{- define "EnvironmentsHelper.defaults" -}}
{{- if .Values.environmentDefaults -}}
{{- range $key, $value := .Values.environmentDefaults }}
{{- printf "- name: %s" $key | nindent 20 -}}
{{- printf "  value: \"%v\"" $value | nindent 20 -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

image

This is probably non-trivial and might require a refactoring of the YAML syntax since the unquoted `printf "- name` *is* a valid plain scalar that can be interpreted as a key. I would be fine with not fixing this in this PR.

deathaxe:
Sure, because in plain YAML a single string beginning a line is scoped a key and interpolation always and only takes place within keys or values.

```yml
key
^^^  meta.mapping.key
```

That's how YAML syntax is designed.

Also there's little chance to distinguish global directives such as `{{- define ... }}` from fully interpolated key-value pairs such as `{{ $key }}: {{ $value }}`.

It may be possible using branching to only scope keys as such, if followed by a `:`, but that would require a major change to YAML. ... not even thinking about all the odd nested key-value pairs stuff which are treated as keys.

There's little to nothing, which can be done about that, at the moment.

deathaxe:
I've tweaked the key lookaheads so they ignore anything within `{{ ... }}`.

FichteFoll (approved):

keith-hall (approved):

PR #4015, opened 2024-07-26
[JavaScript] Code and Data specific indentation rules

deathaxe:
Fixes https://forum.sublimetext.com/t/sublime-text-on-the-word-default/72953

This commit applies dedicated (json) indentation rules to JavaScript mappings to avoid unexpected (un-)indenting of keys, which look like reserved control structure keywords (e.g.: case, default, if, else, ...).


michaelblyons (approved):

rchl:
Was curious if it changes the behavior I struggle with often. It looks like it does. It's not ideal either before or after but it does change the behavior. Is it intended?

## Before

Screenshot 2024-07-26 at 14 10 50

## After

Screenshot 2024-07-26 at 14 09 50

deathaxe:
This change is intended to get the following:

grafik

I now realize it is me seeing that due to a custom key binding of mine, to fix bracket behavior.

```
	// Move closing bracket down by 2 lines and place caret onto next line with indentation
	{ "keys": ["enter"], "command": "insert_snippet", "args": {"contents": "\n\t$0\n"}, "context":
		[
			{ "key": "setting.auto_indent", "operator": "equal", "operand": true },
			{ "key": "preceding_text", "operator": "regex_contains", "operand": "\\[$", "match_all": true },
			{ "key": "following_text", "operator": "regex_contains", "operand": "^\\]", "match_all": true }
		]
	},
```

deathaxe:
Added a commit to also exclude keyword indentation treatment from `meta.sequence` and applied the key binding from JSON.

Can you check, whether it improves your situation, @rchl ?

JavaScript/Default.sublime-keymap
```diff
@@ -51,4 +51,24 @@
             { "key": "selection_empty", "operator": "equal", "operand": false, "match_all": true }
         ]
     },
+
+    // Add indented line in square brackets
```
FichteFoll, line 55: I've been wondering if this is something we'd want to have universally much like the same binding for curlies. I *do* have it in my user key bindings, though I also believe that having the closing bracket at the end of the line instead of in a new line might be more common than for curlies.

Now that I think about it, I should probably add the same for normal parentheses. :thinking:

JavaScript/Indentation Rules - Mappings.tmPreferences
```diff
@@ -0,0 +1,73 @@
...
+	<key>scope</key>
+	<string>
+		source.js meta.mapping,
+		source.js meta.sequence,
+		source.jsx meta.mapping,
+		source.jsx meta.sequence,
+		source.ts meta.mapping,
+		source.ts meta.sequence,
+		source.tsx meta.mapping,
+		source.tsx meta.sequence
```
FichteFoll, line 13: You might be able to simplify this via `(source.js, source.jsx, source.ts, source.tsx) & (meta.mapping, meta.sequence)`, but it won't be equivalent since it does not enforce the same order …

keith-hall (approved):

JavaScript/Indentation Rules.tmPreferences
```diff
@@ -2,16 +2,23 @@
 <plist version="1.0">
 <dict>
 	<key>scope</key>
-	<string>source.js, source.ts, source.jsx, source.tsx</string>
+	<string>
+		source.js - source.js meta.mapping - source.js meta.sequence,
+		source.jsx - source.jsx meta.mapping - source.jsx meta.sequence,
+		source.ts - source.ts meta.mapping - source.ts meta.sequence,
+		source.tsx - source.tsx meta.mapping - source.tsx meta.sequence
```
FichteFoll, line 9: I don't think we can do this. It will cause the patterns to not be active inside any mapping and there are some very common constructs with function definitions in objects, where it is normal to write statements inside a mapping, e.g.
```js
export default {
    setup() {
        // normal code be here, e.g. a switch statement
    }
}
```

Ultimately, we might need cooperation from the syntax definition to do this properly.

FichteFoll (changes requested):

JavaScript/Indentation Rules - Mappings.tmPreferences
deathaxe: Order is what I want to enforce as we need to deal with all the sorts of syntaxes interpolating each other, properly.

JavaScript/Indentation Rules.tmPreferences
deathaxe: And who asked, why people prefer Treesitter?

rchl:
> Added a commit to also exclude keyword indentation treatment from `meta.sequence` and applied the key binding from JSON.
> 
> Can you check, whether it improves your situation, @rchl ?

It does work nicely now.

JavaScript/Default.sublime-keymap
deathaxe: I never prefered closing brackets at the same line, personally.

Well, may it be a useful global behavior? I don't know. Maybe rather unlikely to happen and hard to judge.

Once, we have this binding in each package, we can still make a move :)

deathaxe:
I guess this is the point, when I remember, why I in general gave up to work on indentation stuff due to ST's engine being unusable.

deathaxe:
Well, this PR now requires ST4175+

keith-hall (dismissed):

michaelblyons (approved):

JavaScript/Default.sublime-keymap
deathaxe: In the meanwhile it seems more safe to treat this syntax-specific. 

_These bindings have been copied from JSON._

deathaxe:
Resolved a merge commit in indentation rules.

keith-hall (approved):

michaelblyons (approved):

deathaxe:
Not sure if this PR becomes to big, but it seems useful to also fix a tagged template string related indentation issue, I've come accross while testing things out.

I found single line template strings to cause following lines to be increased in indentation by accident.

FichteFoll:
Looks better now. I'd like to try this out and see if my idea to support arbitrary nesting of mapping literals and function definitions can also be achieved, but I can't promise when I'll find the time. Hopefully by the end of the week, though.

FichteFoll:
Finally found the time to test this and it was a success. You can find the commit (and cherry-pick it) here: https://github.com/FichteForks/Packages/commit/dfad4400693b9dd4482bcf20a54395a93da946e6

Alternatively I can also push it to your branch, I believe, but I wanted you to review it first.

FichteFoll (approved):

michaelblyons (approved):

PR #4018, opened 2024-07-26
[CI] Disable tests against ST4169

deathaxe:
This change is required for #4015 to succeed.

We can re-enable stable tests, once the next stable release is made.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,9 +46,9 @@ jobs:
 
           # latest stable build
           # https://www.sublimetext.com/download
-          - sublime-channel: stable
-            sublime-build: 4169
-            optional: true
+          # - sublime-channel: stable
+          #   sublime-build: 4169
+          #   optional: true
 
           # latest dev build
           # https://www.sublimetext.com/dev


keith-hall:
Do we know why it doesn't work as expected on current the stable build?

deathaxe:
Yes, because indentation rule selectors make use of https://github.com/sublimehq/sublime_text/issues/2152 to recursively select the correct ruleset in deeply nested alternating json objects/lists and function constructs.

```js
var foo = {             
  case: () => {          // data: don't dedent
    switch () {
    case:                // code: dedent
      var bar = {
        case: () => {    // data: don't dedent
          ...
        }
      }        
   }
}

This is available as of ST4173.

In all builds before, selectors match the outermost contexts only and thus indentation starts to fail in on of the inner object/function blocks.


### Changelog

```
- ...
- Improved accuracy of scope selectors
- ...
```

keith-hall (approved):

FichteFoll (approved):
We might also be able to use `continue-on-error: true` to make the action not fail checks, but it's not exactly allowing the failure.

See also https://github.com/actions/runner/issues/2347.

I think in general it's not that problematic if syntax tests break on a stable build because we are, after all, targeting the next ST build which is going to be closer to the last dev build regardless.

PR #4019, opened 2024-07-28
[JavaScript] Basic support for syntax highlighting in tagged templates

### LINKED ISSUE #179: [JavaScript] HTML and CSS syntax highlighting in tagged template strings
Is it possible to support HTML/CSS syntax highlighting inside a tagged template string? I am looking to do the equivalent of [this atom editor change](https://github.com/atom/language-javascript/pull/282).

My first attempt at this did not work out so well. I just tried modifying the existing literal-quasi (but locally renamed it to literal-template-string everywhere in the file), just to see if I could get the HTML referencing correctly, before creating a new kind of match just for the "ending in html" tagged template function case:

``` yaml
  literal-template-string:
    - match: '([a-zA-Z$_][\w$_]*)?(`)'
      captures:
        1: entity.template-string.tag.name.js
        2: punctuation.definition.template-string.begin.js
      push:
        - meta_content_scope: source.html.embedded.js
        - include: 'scope:text.html.basic'
      with_prototype:
        - match: "`"
          captures:
            0: punctuation.definition.template-string.end.js
          pop: true
        - match: '\${'
          captures:
            0: punctuation.template-string.element.begin.js
          push:
            - meta_scope: entity.template-string.element.js
            - match: "}"
              captures:
                0: punctuation.template-string.element.end.js
              pop: true
            - include: expression
```

But I get an `Apparent recursion within a with_prototype action: 25000 context sanity limit hit` error. I am very new to these kinds of syntax files, so I expect I am making some very simple mistakes. I appreciate receiving any pointers on how to fix the issue.

My guess is that since JavaScript.sublime-syntax's `expression` has `include: literal-template-string` and `literal-template-string` is referencing `source:text.html.basic`, which inside of it references `scope:source.js`, maybe some pathway with all of that is causing the recursion.

Related context:
- This does not take into account the changes in #165, I expect to redo some of it once that changeset lands.
- Previous discussion in Benvie/JavaScriptNext.tmLanguage#134, but I would want to scope the use of HTML to tagged template functions ending in `(?i)html`, to allow alternate highlighting for other content, like CSS for functions ending in `(?i)css`, as the Atom changeset does.
- GitHub's source display now does this kind of higlighting: [example test.js](https://github.com/jrburke/tts-syntax-highlight/blob/master/test.js).
- That test file is part of a [tts-syntax-highlight repo](https://github.com/jrburke/tts-syntax-highlight) to track this sort of enhancement in editors now that ES2015 template strings are used more.

deathaxe:
Resolves #179

This commit adds support for syntax highlighting of CSS/JS/JSON/HTML within tagged template strings.

grafik

JavaScript embeds extended syntax definitions to enable `${...}` interpolation in highlighted template tags, very much like most other template syntaxes (e.g.: Go Templates, Jinja2, Liquid, ...).

As templates are embedded, impact on syntax cache size and complexity of JavaScript itself is negligible. They are loaded on demand and thus shouldn't effect other template languages, significantly.

Compatibility notes:

1. Template syntaxes such as Jinja or Liquid need to take action, if they want to support their own interpolations such as `{{ ... }}` within tagged template tags.
2. This PR might affect _JS Custom_, which also supports adding syntax highlighted tagged template tags.
3. This PR uses a naive way to add JavaScript interpolation in ``js`...` `` tags very much like other template syntaxes do. JavaScript however already highlights `${...}` as some sort of dollar variable in various situations. It's currently not clear, if that would be enough which would mean to remove `prototype` manipulations.

### DIFF
--- a/JavaScript/Embeddings/CSS (for JS template).sublime-syntax
+++ b/JavaScript/Embeddings/CSS (for JS template).sublime-syntax
@@ -1,0 +1,23 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/syntax.html
+# highlight tagged template strings
+scope: source.css.js-template
+version: 2
+hidden: true
+
+extends: Packages/CSS/CSS.sublime-syntax
+
+variables:
+
+    ident_start: (?:{{nmstart}}|\${)
+
+contexts:
+
+  prototype:
+    - meta_prepend: true
+    - include: scope:source.js#text-interpolations
+
+  strings-content:
+    - meta_prepend: true
+    - include: scope:source.js#string-interpolations
--- a/JavaScript/Embeddings/HTML (for JS template).sublime-syntax
+++ b/JavaScript/Embeddings/HTML (for JS template).sublime-syntax
@@ -1,0 +1,122 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/syntax.html
+# highlight tagged template strings
+scope: text.html.js-template
+version: 2
+hidden: true
+
+extends: Packages/HTML/HTML.sublime-syntax
+
+variables:
+
+  tag_name_start: (?:[A-Za-z]|\${)
+
+contexts:
+
+  prototype:
+    - meta_prepend: true
+    - include: scope:source.js#text-interpolations
+
+  cdata-content:
+    - meta_prepend: true
+    - include: scope:source.js#string-interpolations
+
+  script-javascript-content:
+    - meta_include_prototype: false
+    - match: '{{script_content_begin}}'
+      captures:
+        1: comment.block.html punctuation.definition.comment.begin.html
+      pop: 1  # make sure to match only once
+      embed: scope:source.js.js-template
+      embed_scope: source.js.embedded.html
+      escape: '{{script_content_end}}'
+      escape_captures:
+        1: source.js.embedded.html
+        2: comment.block.html punctuation.definition.comment.end.html
+        3: source.js.embedded.html
+        4: comment.block.html punctuation.definition.comment.end.html
+
+  script-json-content:
+    - meta_include_prototype: false
+    - match: '{{script_content_begin}}'
+      captures:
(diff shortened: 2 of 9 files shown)


keith-hall (approved):

keith-hall:
There's a couple of test failures with the latest commit, otherwise looks good to me

deathaxe:
They revealed dome shortcommings of my implementation. 

Those are fixed.

I am still a bit uncertain about tags' scopes. 

1. JS originally uses `variable.function.tagged-template`. 
2. Bash and Perl use `entity.name.tag` to scope such language specifiers in front of HEREDOCs 
3. Markdown just uses `constant.other.language-name` in fenced code blocks.

Does it make sense to specify a common scoped used everywhere?

deathaxe:
Applied some changes to auto-indent content of tagged templates, correctly. This commit will create a conflict with #4015 and might need further tweak after that has been merged to apply changes to the new indentation rule file as well.

keith-hall:
> I am still a bit uncertain about tags' scopes.
> 
> 1. JS originally uses `variable.function.tagged-template`.
> 2. Bash and Perl use `entity.name.tag` to scope such language specifiers in front of HEREDOCs
> 3. Markdown just uses `constant.other.language-name` in fenced code blocks.
> 
> Does it make sense to specify a common scoped used everywhere?

Pretty sure 1 is due to how js calls a function for the tagged template literals so seems correct for js. 2 is likely a relic of html tag scoping - it shouldn't IMO be an `entity` if it isn't defining something which can be referenced later. Yes, I know it uses the same token to end the heredoc, but I vote that it doesn't count. 3 seems like the modern way to me.

keith-hall (dismissed):

deathaxe:
Tweaked patterns so embedded source scopes are stripped only, when needed - not in single line template strings.

Related with #4020.

michaelblyons (approved):

keith-hall (approved):

PR #4020, opened 2024-07-28
[HTML] Tweak embedded script/style code boundaries and comment indentation

deathaxe:
This PR...

1. teaches HTML to consume

   - any trailing whitespace until end of line after opening script/style tags or optionally following `<!--` punctuation. 
   - any leading whitespace in front of closing script/style tags or optionally preceded `-->` punctuation.

   to not apply embedded syntax's scope after standalone opening or in front of standalone closing tags.

   This change makes sure to ...

   - always apply HTML specific indentation rules to enclosing script/style tags or comment punctuation.

   - apply custom background highlighting of embedded syntaxes only on full lines, but not after or in front of standalone tags.

2. modifies indentation rules for HTML comments to ignore indentation of comments' content.

   This is primarily required to correctly handle indentation of embedded script/style source code, if it is enclosed in HTML comments. In such cases, additional indentation level is probably undesirable.

### Highlighting changes 

Before:

| without comments   | with enclosing comments
|--- |---
| grafik | grafik

After: 

| without comments   | with enclosing comments
|--- |---
| grafik | grafik

_Background in front of closing `</script>` no longer has custom `source.js.embedded` background._

grafik 

_However everything between tags is scoped `source.<name>.embedded` if tags and code exist in same line(s)._

### DIFF
--- a/ASP/syntax_test_asp.asp
+++ b/ASP/syntax_test_asp.asp
@@ -161,9 +161,8 @@
             Dim var = 0
     '  ^^^^^^^^^^^^^^^^^ source.asp.embedded.html - meta.tag
         -->
-    ' <- source.asp.embedded.html
-    '^^^ source.asp.embedded.html
-    '   ^^^^ - meta.tag - source
+    ' <- - source
+    '^^^^^^^ - meta.tag - source
     '   ^^^ comment.block.html punctuation.definition.comment.end.html
     </script>
     ' ^^^^^^^ meta.tag - source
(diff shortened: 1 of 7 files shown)


keith-hall (dismissed):
I always liked having the background coloring of embedded code start immediately after the open tag and end right before the close tag, but agree that it makes sense for indentation to leave those whitespaces unscoped. And that adding meta scopes just to satisfy me likely isn't worth the maintenance burden. Approved

deathaxe:
You prefer it even if those tags are on a separate line? The idea behind this change besides indentation is better visual appearance of embedded contents' background, which is inspired by how fenced codeblocks are handled in Markdown.

keith-hall:
> You prefer it even if those tags are on a separate line?

I do, but I accept that maybe I'm the odd one out here. I find it draws my eye exactly to the tag boundaries where the embedding really starts and ends.

> Those dependencies between core HTML and 3rd-party syntaxec could significantly be reduced by [sublimehq/sublime_text#3787](https://github.com/sublimehq/sublime_text/issues/3787).

Upvoted 🙂

deathaxe:
Sorry, added and pushed the PHP fix in wrong branch.

michaelblyons (approved):
I actually kind of like this, even if it's not 100% accurate for the start and end scopes. 😅

keith-hall (approved):

PR #4022, opened 2024-07-28
[JavaScript] fix tagged template string key bindings regression

deathaxe:
This PR...

1. scopes tagged template strings `meta.string.template` to give all of them a common scope without regards of used syntax highlighting
2. uses that scope in key bindings' selectors to enable them in syntax highlighted tagged templates.
3. restrict ` to JavaScript strings by enforcing scope order in selectors in binding 1 and 2
4. merge "selector" keys in binding 3 and 4

It fixes no longer working key bindings in syntax highlighted tagged templates, regressed by #4019.

### DIFF
--- a/JavaScript/Default.sublime-keymap
+++ b/JavaScript/Default.sublime-keymap
@@ -33,6 +33,26 @@
         ]
     },
 
+    // Add indented line in backquotes
+    { "keys": ["enter"], "command": "insert_snippet", "args": {"contents": "\n\t$0\n"}, "context":
+        [
+            { "key": "setting.auto_indent" },
+            { "key": "selection_empty", "match_all": true },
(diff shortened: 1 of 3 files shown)


deathaxe:
It's currently a bit unclear to me, whether auto-pairing backquotes not being enabled in JSX/TSX is by intent or whether we want to use the opportunity to add it.

FichteFoll:
> It's currently a bit unclear to me, whether auto-pairing backquotes not being enabled in JSX/TSX is by intent or whether we want to use the opportunity to add it.

I suspect this was an oversight as I see no reason to treat them differently, personally.

FichteFoll (approved):

michaelblyons (approved):

PR #4024, opened 2024-08-05
[ShellScript] Add ZSH

### LINKED ISSUE #1358: Misc. bash/shell issues, some RFEs

### LINKED ISSUE #2635: [ShellScript] arithmetic evaluation context has no comments in ZSH
- OS Version: Arch Linux

```zsh
% i=$(( # abc ))
# zsh: bad math expression: operator expected at `abc '
% i=$(( #def ))
# assuming $def in undefined, this silently fails and sets $i to zero.
```

It is not something unique to Zsh. Bash also has very similar rules regarding Arithmetic Evaluation context as demonstrated below.

```bash
$ i=$(( # abc ))
# bash: bad substitution: no closing `)' in $(( # abc ))
$ i=$(( #def ))
# bash: bad substitution: no closing `)' in $(( #def ))
```

Screenshot_20201124_012616

## Summary

Arithmetic Evaluation context is not subject to comments.

## Description

Syntax highlighting engine treats `#` symbol inside an Arithmetic Evaluation context as the beginning of a comment. Which is incorrect, because `#` in that context could be one of several operators but not a comment.

According to zsh documentation at http://zsh.sourceforge.net/Doc/Release/Arithmetic-Evaluation.html#Arithmetic-Evaluation:

> An expression of the form ‘##x’ where x is any character sequence such as ‘a’, ‘^A’, or ‘\M-\C-x’ gives the value of this character and an expression of the form ‘#name’ gives the value of the first character of the contents of the parameter name.

Real-world line which causes trouble: https://github.com/ohmyzsh/ohmyzsh/blob/7a76afd61e5daab6fb33f955930efa7f4cf16021/lib/functions.zsh#L184
Screenshot_20201124_012053

### LINKED ISSUE #3826: Add zsh script syntax highlight support
### Problem description

There's no zsh syntax option in latest ST4. No related packages too.

If I use bash instead, it gets very strange, like this:
image

I hope there will be an update to this. I appreciate it.

### Preferred solution

Add zsh syntax support.

### Alternatives

A package is also helpful.

### Additional Information

_No response_

### LINKED ISSUE #4033: [ShellScript] nested parentheses treated as arithmetic
### What happened?

Currently `$((` is treated as arithmetic even when it isn't. Example Bash snippet which finds modified txt files but doesn't return an exit code if grep finds no matches:
```bash
TXT_FILES=$((git diff --name-only --diff-filter=ACMR main | grep \.txt) || true)
if [[ -z "$TXT_FILES" ]]; then
    echo 'no text files modified'
fi
```

Adding a space after `$(` and before `(` fixes the highlighting but shouldn't be necessary.

deathaxe:
Fixes #1358
Fixes #2635
Resolves #3826
Fixes #4033

This PR refactors Bash as foundation for adding Zsh.

Used syntax specifications:

- https://www.gnu.org/software/bash/manual/bash.html
- https://zsh.sourceforge.io/Doc/Release/

Syntax follows officiel specifications as close as possible
to reliably support most language features.

## Bash highlights

- tokenize shell words according to official specifications
- improve brace expansions (actually add support for them, everywhere)
- improve glob patterns
- improve parameter expansions
- improve tilde expansions
- add support for arithmetic highlighting shell words
- add support for command highlighting shell words
- drop generic built-in command specific contexts
- add syntax based folding
- improve indentation rules
- improve symbol definitions
- add completions for keywords, built-in commands and variables
- ...

Various scopes of keywords, punctuation, etc. have been modified
to follow latest scope naming best practices.

Those may cause syntax tests of 3rd-party packages to fail.
Syntax highlighting itself should however still work.

## Zsh highlights

- support for ZSH specific alternative control structures
- support for expansion flags
- support glob modifiers

## Benchmarks

Parsing performance is evaluated using syntax test files from current release
and various other real world scripts.

This PR does not affect parsing performance.


michaelblyons:
Oof. Can't `with_prototype` anymore. It hits the sanity limit. I'm gonna have to make some edits to some things.

deathaxe:
This may be caused by various `embed`'s being used to support shell highlighting in quoted strings, which are expanded to push...with_prototype, if ShellScript itself is used like this.

Well, Makefile, MArkdown and YamlPipelines successfully make use of extended ShellScript with little effort to inject own rules. 

push...with_prototype is very likely to create undesirable results with most of current ST syntaxes anyway, due to the many special purpose context switches, they require to implement a language.

ShellScript/Bash.sublime-syntax
```diff
@@ -925,537 +824,732 @@ contexts:
...
   cmd-basic:
-    # basic user defined or external commands
+    - match: \$?\"
+      scope: punctuation.definition.quoted.begin.shell
+      set:
+        - cmd-name-body
+        - path-pattern-double-quoted-body
+        - path-pattern-quoted-begin
+    - match: \$\'
+      scope: punctuation.definition.quoted.begin.shell
```
michaelblyons, line 846: Found this doing some tests on one of my files. Should be punctuation.definition.string.begin, no? Also elsewhere in the file. My poor phone is not a good place for reviewing huge changes.

michaelblyons:
Initial impression is good. I think I'm over the shock of so many things being "string" scoped now. There are a couple odd things I've found so far, but most of them are not really decidable. (Is for example `awk -F' c-? '` a long option value, or an option with a string value?)

Is it worth making a special match for `/dev/null`?
Is it possible to identify the variable being assigned to in `mapfile` so that it is a `variable`, not a string?

ShellScript/Bash.sublime-syntax
deathaxe: Basically a shell only does quotation handling or quote removal. A quotation mark can appear at any position in a word. Thus - ignoring some escaping and expansions - `./"foo"/'bar'` is equal to `"./foo/bar"` both resolved to `./foo/bar`. You can even write `./fo"o/ba"r`.

Thus it actually was odd to scope some parts of such a word `string.quoted` while leaving others blank.

The current syntax therefore distinguishes `literals` and `strings`. While former is used to handle variables (`variable.function`, `variable.other`), the latter one is used for common command arguments.

`punctuation.definition.quoted` is used for quotation within variables, while `punctuation.definition.string` is for string scoped arguments.

On line 854 we are quoting a command, which is scoped `variable.function` but not string.

deathaxe:
> (Is for example `awk -F' c-? '` a long option value, ...

Speaking for the shell, yes. Words are separated by space, parens and pipe operators, only. Thus `-f' c-? '` is a single word.

It doesn't say anything about how the command `awk` interprets the words passed to it as arguments.

Quoting is however important for the shell with regards to character escaping, pattern matching or expansions being applied or not.

ShellScript/Bash.sublime-syntax
deathaxe: I've choosen `string` scopes as it clearly distinguishes variables from normal arguments in most color schemes.

Technical a concept of quoted or unquoted strings known from other languages does not exist. Hence it would also be correct to just scope everything `meta.word.[quoted|unquoted]` and remove all string-like scopes.

michaelblyons:
In a `case` statement, if there is white space between the patter and the `)`, is that white space part of `string.unquoted` or should it be unscoped?

For the most part, I have been extremely pleased with the Bash syntax. I don't use ZSH, so someone else will have to look at that.

deathaxe:
Case patterns actually are pattern lists, rather than a single pattern with alternation. It means each word is a single pattern `pat1 | pat2 | pat3 )`. As such whitespace doesn't belong to them and shoudln't receive string scope. Also `|` is not part of those patterns. It's rather a sequence separator or logical operator.

michaelblyons:
@okdana @Antibioticss Test drive the ZSH, maybe?

michaelblyons (dismissed):

michaelblyons (approved):

michaelblyons:
Okay, so there's a test failure from the Groovy changes, but otherwise still 👌🏼 from me.

michaelblyons (approved):

MatthiasPortzel:
I have a function alias, `r`, defined in my `.zshrc`. With this syntax, the closing bracket is highlighted red, despite working fine. Technically I may be overwriting an obscure ZSH builtin.

image

deathaxe:
Thank's for your feedback. It turned out to be a long-living design/priority issue, which Bash on master suffers from as well. Fixed.

keith-hall (approved):

michaelblyons (approved):

deathaxe:
Great!

PR #4029, opened 2024-08-18
DAVI script syntax support

bobimicroweber:
https://github.com/davinci-script/davi


jfcherng:
I don't think this will be accepted considering it's not a popular language. It should be added as a 3rd-party package.

bobimicroweber:
> I don't think this will be accepted considering (at least) it's not a popular language. It should be added as a 3rd-party package.

If they don't approve the pull request, I will write my own ide editor for davi.
No problem, as they decide.

jfcherng:
Just a FYI, even Swift didn't make it.

bobimicroweber:
> Just a FYI, even Swift didn't make it.

That's what Electron is for.. Anyone who gets rejected can make their own IDE :D

deathaxe:
May I suggest to create a DAVI repository at https://github.com/SublimeText instead and adding you as maintainer?

bobimicroweber:
my idea is fast installation and davi script support. Please suggest how to make it the easiest and everyone is happy.

deathaxe:
This would be the default way for any 3rd-party package then - registering it to Package Control.

Any bitbucket/gitlab/github repository can be registered as package by adding it to proper repository.json at https://github.com/wbond/package_control_channel/.

It can then be easily installed via...

1. Opening Command Palette
2. Executing "Package Control: Install Package"
3. Entering package name and hitting enter.

bobimicroweber:
Ok that is cool solution. Thank you!

PR #4034, opened 2024-09-12
[CSharp] Add the C# 11 type modifier `file` feature support

lindexi:
See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/file

The `file` is the keyword of type modifier as the type visibility.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -32,7 +32,7 @@ variables:
   unicode_char: '(?:\\u\h{4}|\\U\h{8})'
   escaped_char: '(?:\\[abfnrtv"''\\]|{{unicode_char}}|\\x[0-9a-fA-F]{1,4}|\\[0-9]{1,3})'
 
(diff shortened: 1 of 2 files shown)


michaelblyons:
Do you want to write some unit tests, too?

lindexi:
Thank you, let me learn how to write unit tests in this project.

lindexi:
@michaelblyons I added the unit test code, could you review my code?

michaelblyons:
I have approved running the tests, but there are some test failures. Once those look good, I think it'll be clear to merge.

lindexi:
@michaelblyons Thank you and I try fix the unit test. Sorry, I'm not familiar with test syntax.

michaelblyons:
If you're having trouble, you can install the PackageDev package, click at end of the line you want tests for in the unit test file, and Cmd+Enter (or Ctrl+Enter) and it should write a test suite for that line.

michaelblyons (approved):

deathaxe (approved):

PR #4035, opened 2024-09-14
[JavaScript] Fix top-level properties in css tagged template strings

deathaxe:
Styled components may just contain property value pairs instead of full stylesheets.

This commit therefore adjusts extended CSS syntax to support both, selectors and properties in main top-level context.

This commit fixes highlighting of

```js
const style = css`
  font-size: 1.5em;
  text-align: center;
  color: #BF4F74;
`;
```

| before | after
| ---    | ---
| grafik | grafik

---

CSS' main context includes `stylesheet` by default.

https://github.com/sublimehq/Packages/blob/e2e71392109eb9440123cebb21907db7e0c9d00c/CSS/CSS.sublime-syntax#L352-L364

This PR modifies main by including a modified `stylesheet` with `- include: selectors` being replaced by `- include: properties-or-selectors`.

https://github.com/sublimehq/Packages/blob/ea059a725c8487197cd433d34cd9ed53c3b9ecf7/JavaScript/Embeddings/CSS%20(for%20JS%20template).sublime-syntax#L25-L33

Note:

This change was already merged (by accident) for TypeScript via #4036.

https://github.com/sublimehq/Packages/blob/e2e71392109eb9440123cebb21907db7e0c9d00c/JavaScript/Embeddings/CSS%20(for%20TS%20template).sublime-syntax#L25-L33


michaelblyons (approved):

deathaxe:
OK, this should be it, finally.

michaelblyons (approved):

deathaxe:
Any blockers for this PR?

The only alternative solution I see would be to apply this kind of lazy nesting matching to CSS directly. It would safe significant amount of lines, but I still hesitate as CSS itself does not support nesting under all circumstances.

keith-hall (approved):

deathaxe:
Thanks!

PR #4047, opened 2024-09-29
[Common] Add git conflict marker highlighting

deathaxe:
This PR ...

1. uses _Diff/Diff.sublime-syntax#conflict-markers_ as sort of comments in all syntax bundled definitions.

   Comments are included in most contexts in all syntax definitions, thus conflict markers should be matched properly with high chances as well.
   
3. includes merge conflict patterns into known multi-line strings or heredocs.

    grafik

4. adds conflicts to local symbol list to support quick navigation or finding them

    grafik

#### Goals

1. improve UX when using ST as git merge tool
3. reduce risk of badly breaking syntax highlighting due to conflict markers

#### Motivation

1. This PR was created as possible fix/workaround for https://github.com/sublimehq/sublime_text/issues/6498, which is caused by git merge conflict markers, but applies to all syntax definitions. It is therefore a possible alternative for #4046, assuming there are no other "invalid syntax" scenarios, which can break JSX/TSX in such bad ways.
2. Git is a rather common tool these days, and ST can be used as an editor to edit commit messages or resolve conflicts. As such, git conflict markers shouldn't break syntax highlighting or even cause crashes or deadlocks.
4. various linters/tools, such as JSX/TSX language servers also detect git conflict markers and highlight them, most likely to avoid same kind of issues we see with regards to confusing syntax engine.

#### Remarks

1. ~~With this PR all syntax definitions depend on _Diff.sublime-syntax_ and thus require _Diff_ package to be present and enabled.~~

   - Dedicated `merge-conflictmarkers` contexts are added to each syntax to avoid them depend on Diff syntax as most of them already also contain dedicated "shebang" context.

2. Initial state of this PR provides a "works in most cases" quality based on existing comments/strings contexts. There may be some situations left, in which conflict markers are not yet detected. Support can however easily be improved in future.

3. Some syntaxes such as Markdown use `===` to highlight headings, which still may cause syntax highlighting issues caused by unavoidable ambiguities. So this PR can provide "the best we can" quality.


michaelblyons:
I have some thoughts, not in any order, nor necessarily exhaustive:

1. If Git is disabled, I expect the matches just don't show up. It's not catastrophic, right?
2. I like the concept. I have a minimal "Conflict" syntax lying around somewhere with `markup.line.(inserted|deleted)`, which is probably mostly obsolete with this change. 
3. In case it's not obvious to everyone, this can result in mismatched context start/end, but that was already happening anyway.
4. If this is well-received, we should also support `|||||` for `diff3` markers.
5. This is a great example of how injections would be helpful as a core feature.

deathaxe:
> If Git is disabled, I expect the matches just don't show up. It's not catastrophic, right?

The only downside would be console messages about "no such context".

> In case it's not obvious to everyone, this can result in mismatched context start/end, but that was already happening anyway.




    If this is well-received, we should also support ||||| for diff3 markers.

> This is a great example of how injections would be helpful as a core feature.

deathaxe:
An alternative location for the Merge syntax would be within _Text_ package, which if disabled causes lots of trouble anyway and is thus less likely to be.

This would be justified by conflict markers not being limited to git, but rather being a common unix convention, with regards to `diff3`.

michaelblyons:
> by the way ours/theirs is interrupting normal source code.

Yes, that's what I mean.  Agreed that it can't be helped.

```
<<<<<
func foo (bar) {
=====
func foo (bar, baz) {
>>>>>
// Extra nesting
```

deathaxe:
Still UX is way better then with most syntaxes highlighting those markers as operators, tag punctuation or even illegal.

deathaxe:
1. Renamed scopes to original unix `diff3` scope as those markers are not tight to git.
2. moved it to _Text_ package to always be available.

Ideally _Diff_ package seems the more suitable target, but it is not included in Sublime Merge and could also be disabled like _Git Formats_.

deathaxe:
Moved Diff3.sublime-syntax into _Diff_ package as it finally feels misplaced in _Text_ package.

@dpjohnst : As Diff3 is included by all other syntaxes, _Diff_ package needs to be shipped with Sublime Merge, once this PR gets merged.

michaelblyons:
> Moved Diff3.sublime-syntax into _Diff_ package as it finally feels misplaced in _Text_ package.

Good. I thought about recommending this, and forgot to post.

If I were doing this, I'd make the syntax a functional one called `Diff/Conflict.sublime-syntax` and have markup add/remove sections in between the conflict markers. (Unsure what to use for the Diff3 common base. I personally use comment for those, but I could see markup modified as another possibility.) You can still grab just the marker contexts for your injection into the other syntaxes.

deathaxe:
Name "Diff3" is inspired by the original unix tool producing output with such tags, like unix tool "diff" produces output with syntax "Diff".

Renaming it to _Conflict.sublime-syntax_ would mean to remove reference to diff3, which I don't find ideal.

Context `conflicts` exists to be able to add more diff3 specific rules, which are not of interest in other syntaxes. I haven't checked if that will be required though.

We could otherwise follow Diff.sublime-syntax and just put everything in `main`. I currently don't see reasons to add each pattern into a dedicated context as they are always required together and chances they need to be overridden individually seem rather low.

`markup` scope is somewhat inspired by Markdown, which scopes most such special things that way. That scope is also rather uncommon in normal source code and thus may provide prominent enough highlighting to clearly stand out from the rest.

Scoping it `comment` would be as correct though.

michaelblyons:
How about `Diff3 Conflict.sublime-syntax`, then? The reason I'm shy about calling it `Diff3.sublime-syntax` is that the "normal" Diff3 output [looks very different](https://www.gnu.org/software/diffutils/manual/diffutils.html#Example-diff3-Normal) from that.

deathaxe:
How about adding support for normal output?

michaelblyons:
That might work. IIRC, one of the (three) `diff3` output modes has a conflict with one kind of normal `diff` output, but they might not conflict with each other. It might be possible to bundle all three `diff3` output styles into a single syntax.

deathaxe:
Well, I'd probably start with normal output mode.

I have no strong meaning about scope names - just explained their origin.

With a look at Diff we could also use `meta.separator punctuation.definition.separator` for `<<<<<` etc.

Anything which allows them to be uniquely addressed by color schemes might be ok. The question is what defaults we prefer. Should they appear like comments in the first place or just like plain text (which Mariana/Monokai) highlight them, currently.

Another question arises. Diff uses `source.diff`. Should we therefore change main scope to `source.diff3`?

michaelblyons:
> With a look at Diff we could also use `meta.separator punctuation.definition.separator` for `<<<<<` etc.

I'm opposed. I think the scope you already chose is the right one, and Diff should change to match.

> Anything which allows them to be uniquely addressed by color schemes might be ok. The question is what defaults we prefer. Should they appear like comments in the first place or just like plain text (which Mariana/Monokai) highlight them, currently.

Oh, the conflict markers? I would not use `comment`. I think your original `punctuation.section` is correct, as I said above.

I was only talking about if a simple syntax is made whose `main` highlights Diff3 conflicts, not the injection of conflict markers into complicated source code.

```
octopus
<<<<< a
tiger
whale
\^^^^^ markup.line.deleted
|||||
koala
\^^^^^ ???? not sure if `comment` or `markup.line.modified`
=====
elephant
\^^^^^^^^ markup.line.inserted
>>>>>
```

> Another question arises. Diff uses `source.diff`. Should we therefore change main scope to `source.diff3`?

Seems fine. No strong opinions. I've changed my mind on this a bunch of times with no obvious right answer.

michaelblyons:
This may have interesting application for [BracketHighlighter](https://packagecontrol.io/packages/BracketHighlighter) and something modeled after [Git Conflict Resolver](https://packagecontrol.io/packages/Git%20Conflict%20Resolver).

deathaxe:
I'll revicit this PR once #4111 is merged.

deathaxe:
Rebased branch and tweaked syntaxes to use _Diff/Diff.sublime-syntax#conflict-markers_.

Yet question remains, whether it is a good idea to create such massive cross package depenedency.

OCaml/OCamllex.sublime-syntax
```diff
@@ -96,7 +96,8 @@ contexts:
           pop: true
         - match: '\\(x[a-fA-F0-9][a-fA-F0-9]|[0-2]\d\d|[bnrt''"\\])'
           scope: comment.block.string.constant.character.escape.ocaml
-  match-patterns:
+    - include: scope:source.diff#conflict-markers  # merge conflict marker  match-patterns:
```
michaelblyons, line 99: Mis-paste

OCaml/OCamlyacc.sublime-syntax
```diff
@@ -60,6 +60,7 @@ contexts:
           pop: true
         - match: '\\(x[a-fA-F0-9][a-fA-F0-9]|[0-2]\d\d|[bnrt''"\\])'
           scope: comment.block.string.constant.character.escape.ocamlyacc
+    - include: scope:source.diff#conflict-markers  # merge conflict marker  declaration-matches:
```
michaelblyons, line 63: Mis-paste

RestructuredText/syntax_test_restructuredtext.rst
```diff
@@ -21,6 +21,22 @@ This is not a comment
...
 .. <- comment.line.double-dot.restructuredtext
 
+
+<<<<<<< HEAD
+.. <- meta.block.conflict.begin.diff punctuation.section.block.begin.diff
+..^^^^^ meta.block.conflict.begin.diff punctuation.section.block.begin.diff
+..     ^ meta.block.conflict.begin.diff - entity - punctuation
+..      ^^^^ meta.block.conflict.begin.diff entity.name.section.diff
+..          ^ meta.block.conflict.begin.diff - entity - punctuation
+
```
michaelblyons, line 31: Did you leave out `========` intentionally?

Rust/tests/syntax_test_conflicts.rs
```diff
@@ -0,0 +1,62 @@
...
+//^^^^^ meta.block.conflict.separator.diff punctuation.section.block.diff
+//     ^ meta.block.conflict.separator.diff - punctuation
+
+>>>>>>> master
+// <- meta.block.conflict.end.diff punctuation.section.block.end.diff
+//^^^^^ meta.block.conflict.end.diff punctuation.section.block.end.diff
+//     ^ meta.block.conflict.end.diff - entity - punctuation
+//      ^^^^^^ meta.block.conflict.end.diff entity.name.section.diff
+//            ^ meta.block.conflict.end.diff - entity - punctuation
+)
```
michaelblyons, line 62: Add trailing newline?

michaelblyons (changes requested):
I also thought I saw a Git Formats tmPreferences file that I was going to suggest moving to Diff, but it seems to be gone now.

deathaxe:
> I also thought I saw a Git Formats tmPreferences file

Yeah, I can always look at PRs for hours and a second after hitting push, I find something.

RestructuredText/syntax_test_restructuredtext.rst
deathaxe: This is one of the ambiguities we are facing in RST and MD. 

This line is already tested as part of a heading in line 48.

OCaml/OCamlyacc.sublime-syntax
deathaxe: Nope, it should be following block comments. Maybe just missing another one in quoted docstrings.

OCaml/OCamllex.sublime-syntax
deathaxe: Also valid here.

OCaml/OCamlyacc.sublime-syntax
michaelblyons: Why is `declaration-matches` part of the comment? (And before your more recent push, it no longer a context.)

OCaml/OCamllex.sublime-syntax
```diff
@@ -152,3 +154,4 @@ contexts:
           scope: constant.character.regexp.escape.ocaml
         - match: '\\(?!(x[a-fA-F0-9][a-fA-F0-9]|[0-2]\d\d|[bnrt''"\\]|[\|\(\)1-9$^.*+?\[\]]|$[ \t]*))(?:.)'
           scope: invalid.illegal.character.string.escape
+        - include: scope:source.diff#conflict-markers  # merge conflict marker  match-patterns:
```
michaelblyons, line 157: Why is `match-patterns:` part of this line comment?

OCaml/OCaml.sublime-syntax
```diff
@@ -384,6 +384,7 @@ contexts:
           pop: true
         - match: '\\(x[a-fA-F0-9][a-fA-F0-9]|[0-2]\d\d|[bnrt''"\\])'
           scope: comment.block.string.constant.character.escape.ocaml
+    - include: scope:source.diff#conflict-markers  # merge conflict marker  constants:
```
michaelblyons, line 387: Why is `constants:` in the comment here?

michaelblyons (dismissed):
Added more context on the OCaml thing I flagged.

OCaml/OCamllex.sublime-syntax
deathaxe: I see.

OCaml/OCamlyacc.sublime-syntax
deathaxe: Maybe some multi-cursor edit or issue with unproper use of "Find in Files", not sure.

OCaml/OCamllex.sublime-syntax
deathaxe: Fixed.

OCaml/OCamlyacc.sublime-syntax
deathaxe: Fixed

OCaml/OCamllex.sublime-syntax
michaelblyons: This one is still here.

michaelblyons (commented):
Switching from "changes requested" to neutral. I think this is a reasonable change, but I would like to hear from others before approving.

deathaxe:
With regards to how shebang is implemented, I tend to think conflict marker contexts should also be implemented individually for each syntax to avoid cross-package dependencies, especially as Diff is not part of Sublime Merge.

deathaxe:
If someone finds unhandled multi-line strings in a syntax, feel free to report it.

Otherwise with dependencies to Diff.sublime-syntax being removed, this PR should be ready.

deathaxe:
Folding would only make sense if ours/theirs/base parts could be folded individually. With `base` part being optional syntax, a begin/end may be missing, causing everything being folded.

It's also not quite sure how it interferes with syntaxes' own folding rules.

Hence I'd hesitate to "just" add some rules for it.

deathaxe:
FWIW, it even makes a difference in Sublime Merge...

grafik

michaelblyons (approved):

deathaxe:
Cool!

PR #4051, opened 2024-10-02
[Groovy] use more specific sub-scopes for keywords

keith-hall:


michaelblyons (approved):

Groovy/Groovy.sublime-syntax
```diff
@@ -147,49 +147,72 @@ contexts:
...
-      scope: keyword.other.throws.groovy
+    - match: \bthrows\b
+      scope: keyword.control.exception.groovy
   keyword-operator:
-    - match: \b(as)\b
-      scope: keyword.operator.as.groovy
-    - match: \b(is)\b
-      scope: keyword.operator.is.groovy
+    - match: \bas\b
+      scope: keyword.context.resource.groovy
```
deathaxe, line 201: I know `as` keyword was topic of discussions multiple times.

Does it make sense to interpret it as kind of `keyword.operator.assignment.[as|alias]`?

It is most often used to ...
1. import modules under a custom local name (`import foo as bar`)
2. assign variables another local name (`define foo as bar`, ...)

All of which look like defining alias names for given variables, with just the syntactical difference of assignee being located on the right.

keith-hall: It is also used in Python to:
3. set the scope of a context manager, syntactic sugar for try, enter context, do work, finally exit context. 

To be fair I don't know how Groovy uses it, I'm not familiar enough with it yet, but I was thinking on type 3 in which case I think the `context.resource` subscope is more accurate than a simple alias would imply.
I certainly do like your proposal for other cases though. (Except I wonder if the assignee being on the right will be confusing if highlighted the same as other assignments as color schemes would do by default until this specific rule is added...)

deathaxe: Python creates over-eager special scopes for all sorts of contexts, which doesn't need to be correct.

My thoughts apply to python and all other syntaxes as well. Context managers instantiate a class and assign it to a variable on the right, which is also some sort of alias for the context manager to refer to.

Actually I have been applying `keyword.operator.assignment` to all `as` keywords, when tweaking syntaxes as well as scoping `in` `keyword.operator.iteration`.

deathaxe: Original Groovy scopes `as` `keyword.operator.as`, because it is a [coercion operator](https://groovy-lang.org/operators.html#_coercion_operator) according to Groovy's language spec. It is a kind of type cast operator.

Hence I'd hesitate to change the `keyword.operator` scope.

If a sematic scope is desired maybe 
- `keyword.operator.type.coercion` 
- or `keyword.operator.type.as`
- or `keyword.operator.assignment.type`

I've also always struggled with scoping `instanceof` keywords, which is also supported by Groovy. They could be scoped `keyword.operator.comparison.type` to re-use existing 3rd-level scope or `keyword.operator.type.instanceof`.

michaelblyons:
Drop the `as` change and merge? I still think it's worth having that conversation, but not worth holding up the PR.

deathaxe (approved):

michaelblyons (approved):

PR #4053, opened 2024-10-03
[JavaScript] Assign typescript to *.cts and *.mts extensions

deathaxe:
This commit adds counter-parts for JavaScript's `*.cjs` and `*.mjs` file extensions to Typescript.


michaelblyons (approved):
It seems likely that someone using ST is not going to want the other things I've dug up using those file extensions: 
- Reallusion cartoon
- Camcorder stream (maybe?)

FichteFoll (approved):

PR #4055, opened 2024-10-10
[CSS] Fix issues with template interpolation

deathaxe:
This PR...

1. changes context include order to give `property-lists` less priority over `{{ interpol }}` interpolation tags, so they are scoped as part of selectors or property names.

2. in case strategy 1 is not enough, provide variables to override block punctuation, without duplicating contexts.

> Note: This commit does not change behavior of CSS itself, just helps inherited
>       CSS variants used in templates like Blade, Go, Jinja, Twig, ...

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -85,6 +85,13 @@ variables:
 
   logical_operators: (?i:and|or|not|only)
 
+  # Block punctuations
+  #  Provides an easy and safe interface for template syntaxes to override
+  #  block punctuation patterns to prevent ambiguities with interpolations.
+  #  It helps to correctly highlight {{ tag }} { {{prop}}: {{value}} }
+  block_start: \{
+  block_end: \}
+
   # Properties and Selectors
   property_or_selector_begin: (?={{ident_begin}}|{{selector_start}})
   property_end: (?=[;@)}])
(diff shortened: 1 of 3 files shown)


CSS/CSS.sublime-syntax
```diff
@@ -355,10 +362,14 @@ contexts:
 
   stylesheet:
     # https://www.w3.org/TR/css-syntax-3/#typedef-stylesheet
+    #
+    # CAUTION:
+    # porperty-lists must follow selector, properties, at-rules
```
keith-hall, line 367:
```suggestion
    # property-lists must follow selector, properties, at-rules
```

CSS/CSS.sublime-syntax
```diff
@@ -689,19 +700,22 @@ contexts:
...
-    - match: \{
+    - match: '{{block_start}}'
       scope: punctuation.section.block.begin.css
       set: at-keyframe-block-body
     - include: comments
     - include: else-pop
 
   at-keyframe-block-body:
+    # CAUTION:
+    # porperty-lists must follow selector, properties, at-rules
```
keith-hall, line 711:
```suggestion
    # property-lists must follow selector, properties, at-rules
```

CSS/CSS.sublime-syntax
```diff
@@ -1508,28 +1522,32 @@ contexts:
...
       scope: punctuation.section.block.begin.css
       push: property-list-body
 
   property-list-body:
     # According to official specs, this is a style-block.
     # It supports declarations, nested style-rules and at-rules.
     # https://www.w3.org/TR/css-syntax-3/#typedef-style-block
+    #
+    # CAUTION:
+    # porperty-lists must follow selector, properties, at-rules
```
keith-hall, line 1542:
```suggestion
    # property-lists must follow selector, properties, at-rules
```

keith-hall (approved):
Just a typo in the comments, otherwise looks good to me

keith-hall (approved):

michaelblyons (approved):

PR #4056, opened 2024-10-12
[CSS] Disable completions in embedded template tags

deathaxe:
This PR suggests to modify CSS completion selector to prevent property completions in comments and embedded template code.

Without this restriction, 

1. template language related completions are often nearly completely hidden.
2. chances are high to commit property completions within template expressions.

... both of which degrates developer experience with regards to completions.

#### CSS

```
div {
  /* |<- no css property completions here */
}
```

#### CSS (PHP)

```
div {
   <?php |<- no css property completions here ?>
}
```

#### CSS (Jinja2)

```
div {
   {% |<- no css property completions here %}
   {{ |<- no css property completions here }}
}
```

same in all the others ...

Note: This PR already excludes `meta.selector` as it locally builds on top of #3983.


keith-hall (dismissed):

deathaxe:
Resolved conflicts with merged #3983.

michaelblyons:
Is there a way to test the negated completions? If not, I suggest adding an extra negation (`- source.css source`) in case someone does not follow the `meta.embedded` convention.

deathaxe:
We'd need some kind of unittest to check whether css_completion plugin returns something based on given cursor positions. Uncertain whether we could applicate UnitTesting to do that or if we want to.

Actually `- source.css source - source.css text` was my first choice either. Was uncertein though whether all template syntaxes apply `source` to their template tags. Maybe having both is foolprove.

michaelblyons (approved):

CSS/CSS.sublime-settings
```diff
@@ -4,11 +4,11 @@
 
     // Controls what scopes default completions will be provided in.
     // Can be a list of strings which are joined before matching.
-    "default_completions_selector": "source.css",
+    "default_completions_selector": "source.css - source.css comment - source.css meta.embedded - source.css source - source.css text",
```
keith-hall, line 7: I'll need to refresh my memory of multiple negative selector operator precedence to understand what this is doing...
A comment which explains this new default could help :)

deathaxe: A reminder was added to settings.

We could also wrap all negative selectors into parens, if that helps improving readability, just like

```
"default_completions_selector": "source.css - (source.css comment | source.css meta.embedded | source.css source | source.css text)",
```

The idea is to exclude CSS comments and all sorts of embedded source/text under source.css.

Current selectors disable completions in `source.css source.php` for instance, but not in `source.php source.css`. Not quite sure it is a realistic example but selectors should just make sure to only exclude stuff under CSS.

keith-hall (approved):

michaelblyons (approved):

deathaxe:
Thanks!

PR #4057, opened 2024-10-14
[Pascal] Fix tests not running

BenjaminSchaaf:
Syntax test files must start with "syntax_test_"


keith-hall (approved):

deathaxe (approved):

michaelblyons:
Did something change? I refer to https://github.com/SublimeText/PackageDev/issues/247#issuecomment-479562673

michaelblyons:
Sure, sure. I don't remember what experimentation I did to test my complaint in PackageDev, but it may have been for running syntax tests from the test file or a single definition file (which is mostly how I do it).

I may have actually used the Pascal test as a "See, they do this in the Defaults, too." 😅😅

PR #4065, opened 2024-10-23
[C] Add support for new preprocessor directives

braewoods:
This adds support for the new #elifdef and #elifndef preprocessor directives that are a combination of #else with #ifdef or #ifndef.

C++, Objective-C, and Objective-C++ are also updated because they derive from the C syntax family as well.

This has already made it into recent C and C++ compilers and is part of the C23 and C++23 standards.

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -620,7 +620,7 @@ contexts:
         - match: '\}'
           scope: meta.function.c meta.block.c punctuation.section.block.end.c
           pop: true
-        - match: (?=^\s*#\s*(elif|else|endif)\b)
+        - match: (?=^\s*#\s*(elif|elifdef|elifndef|else|endif)\b)
           pop: true
         - match: '(?=({{before_tag}})([^(;]+$|.*\{))'
           push: data-structures
@@ -775,7 +775,7 @@ contexts:
       scope: punctuation.section.block.begin.c
       push:
         - meta_scope: meta.block.c
-        - match: (?=^\s*#\s*(elif|else|endif)\b)
+        - match: (?=^\s*#\s*(elif|elifdef|elifndef|else|endif)\b)
           pop: true
         - match: '\}'
           scope: punctuation.section.block.end.c
@@ -1025,11 +1025,11 @@ contexts:
         - include: preprocessor-comments
         - match: \bdefined\b
           scope: keyword.control.c
-        # Enter a new scope where all elif/else branches have their
-        # contexts popped by a subsequent elif/else/endif. This ensures that
-        # preprocessor branches don't push multiple meta.block scopes on
-        # the stack, thus messing up the "global" context's detection of
-        # functions.
+        # Enter a new scope where all elif/elifdef/elifndef/else branches have
+        # their contexts popped by a subsequent elif/elifdef/elifndef/else/endif.
+        # This ensures that preprocessor branches don't push multiple meta.block
+        # scopes on the stack, thus messing up the "global" context's detection
+        # of functions.
         - match: $\n
           set: preprocessor-if-branch-global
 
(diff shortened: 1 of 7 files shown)


braewoods:
Ok. I thought I'd found something I needed to adjust but turns out it wasn't the case.

michaelblyons:
Can you add some assertions in the test modification?

braewoods:
> Can you add some assertions in the test modification?

All I did was extend the existing rules that already handle `elif` and `else` directives to include the new ones, and the only test case I could find for `elif`. What assertions could I even add that would be useful?

C++/syntax_test_c.c
```diff
@@ -437,8 +437,16 @@ if (1) {
 if (2) {
 # elif BAZ
 if (3) {
-# else
+#elifdef BAR2
```
deathaxe, line 440: What Michael is asking for is adding some test assertions.

Example

```suggestion
#elifdef BAR2
/* <- keyword.control.import */
/*^^^^^^ keyword.control.import */
```

The choosen section however seems to rather be for testing correct handling of multiple `{` with intermediate directives.

Tests targeting `#ifdef` seem to be located at https://github.com/sublimehq/Packages/blob/493cb8218a2d77128fb54ebf096256aa5cdeb019/C%2B%2B/syntax_test_c.c#L583-L584

C++/C.sublime-syntax
```diff
@@ -1041,8 +1041,8 @@ contexts:
       captures:
         1: meta.preprocessor.c keyword.control.import.c
       pop: true
-    - match: (?=^\s*#\s*(elif|else)\b)
-      push: preprocessor-elif-else-branch-global
+    - match: (?=^\s*#\s*(elif|elifdef|elifndef|else)\b)
+      push: preprocessor-elif-elifdef-elifndef-else-branch-global
```
deathaxe, line 1045: We probably don't need to rename contexts. Adding keywords to patterns is sufficient.

braewoods: Reverted in latest revision.

braewoods:
I've attempted to revise it given the provided feedback. Is there anything else?

C++/syntax_test_c.c
```diff
@@ -589,9 +589,17 @@ int foo(int val, float val2[])
     if (val == -1) {
 /*  ^^ keyword.control */
 /*                 ^ meta.block meta.block punctuation.section.block.begin */
+#elifdef BAR
+ /* ^ meta.block meta.block */
```
deathaxe, line 593: We probably want to make sure, those new keywords are scoped `keyword` as this is the major change of this PR.

```suggestion
 /* ^ meta.block meta.block keyword.control.import */
```

deathaxe (commented):
A little note/suggestion about syntax tests, but otherwise LGTM!

C++/syntax_test_c.c
braewoods: > We probably want to make sure, those new keywords are scoped `keyword` as this is the major change of this PR.

Well that's the strange thing. When I tried to do that, it failed the assertion when I ran the syntax tests from within Sublime Text 4180. `keyword.control.import` would work for `#elif` but none of my new additions and I can't determine why.

deathaxe: Well, checked out your branch and all scopes are applied as expected. Looks all good.

Have you symlinked C++ and Objective-C++ folders into your ST's _Packages_ directory, so ST can see the changes?

To not bother you with details, I've updated tests for you.

deathaxe (approved):

C++/syntax_test_c.c
braewoods: I re-ran the syntax tests from Sublime Text and got these from your recent changes:
```Packages/User/C++/syntax_test_c.c:592:1
error: scope does not match
592 | #elifdef BAR
593 |  /* <- keyword.control.import */
    |  ^ this location did not match
actual:
    | ^^^^^^^^^ source.c meta.function.c meta.block.c meta.block.c 

Packages/User/C++/syntax_test_c.c:596:1
error: scope does not match
596 | #elifndef BAZ
597 |  /* <- keyword.control.import */
    |  ^ this location did not match
actual:
    | ^^^^^^^^^^ source.c meta.function.c meta.block.c meta.block.c meta.block.c 

FAILED: 2 of 674 assertions in 1 files failed
[Finished]```

braewoods: I see. So that's the issue. I thought it was rather strange. It's working now.

michaelblyons (approved):

PR #4071, opened 2024-10-28
[C] Add support for constexpr storage class specifier

braewoods:
This was added in C23 but is more limited in scope than the C++ version. The C version can only be applied to variables to indicate their usability as a compile time constant.


deathaxe (dismissed):

michaelblyons:
If you're a big C user, would you be interested in looking at #1831, please? A lot of work was sunk into that by ismell, and you and other C users might be interested in the improvements it brings.

FichteFoll (dismissed):
https://en.cppreference.com/w/c/language/constexpr

FichteFoll (approved):

braewoods:
I revamped the original commit to resolve previous conflict with a simpler rebase.

michaelblyons (approved):

PR #4074, opened 2024-10-29
[C] Add support for thread_local and its aliases

braewoods:
This was first added in C11 but has been revised by C23. It is a storage class modifier so all we need to do is add it to the existing list.


michaelblyons (approved):

FichteFoll (approved):
https://en.cppreference.com/w/c/thread/thread_local

PR #4076, opened 2024-10-30
[C] Add more standard types from ISO C

braewoods:
This adds the types defined by stddef.h, wchar.h, and uchar.h. Some of these types were first defined by C++ but are now defined by C11 or C23 so we will defer to C instead now.

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -148,8 +148,6 @@ contexts:
       scope: keyword.operator.word.c++
 
   unique-types:
-    - match: \b(char8_t|char16_t|char32_t|wchar_t|nullptr_t)\b
-      scope: storage.type.c++
     - match: \bclass\b
       scope: keyword.declaration.class.c++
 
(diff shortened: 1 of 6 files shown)


michaelblyons:
Did you intend to remove the types from C++?

michaelblyons:
The inheritance isn't set up. People just muse about how it *should* be.

braewoods:
It isn't? I thought it was given that changes I made to C were also showing up in Objective-C, etc.

michaelblyons:
Where are you getting Obj-C from?

braewoods:
> Where are you getting Obj-C from?

I have both the C++ and Objective-C packages from this repository overriding the builtin ones in my Sublime Text. I noted that anything new that I've been adding syntax highlighting for to C.sublime-syntax would also show up for C++, Objective-C, and Objective-C++ syntax languages. So that's why I thought it was being inherited.

deathaxe:
C is indeed the base syntax for C++, Objective-C and Objective-C++. They don't yet use inheritance but include common contexts from C to share contexts/rules.

Even though patterns/contexts are implemented modified only in C, we still should add relevant assertions to all 4 syntax test files so we can be sure this kind of pattern sharing works or doesn't get broken by any future change.

braewoods:
> Even though patterns/contexts are implemented modified only in C, we still should add relevant assertions to all 4 syntax test files so we can be sure this kind of pattern sharing works or doesn't get broken by any future change.

So then how do you update the tests for the other languages over time? I noticed a lot was previously shared but not necessarily up to date with the one it originally copied from.

deathaxe:
The update of tests is a manual copy & paste process, one of the reasons for possible inconsistency. More over C family syntax definitions and tests are rather a unsorted collection of arbritary contributors' changes.

Thus entropy of a possibly initial structure of contexts and tests has grown over time.

braewoods:
> The update of tests is a manual copy & paste process, one of the reasons for possible inconsistency. More over C family syntax definitions and tests are rather a unsorted collection of arbritary contributors' changes.
> 
> Thus entropy of a possibly initial structure of contexts and tests has grown over time.

For this PR, I can just replace the removed tests with the new one from C if that helps. Given what you've told me, I think I will review the tests and try to rewrite them off the C one once I am finished with my whole set of revisions.

braewoods:
Ok. I've revised the syntax tests for the other shared languages.

deathaxe (approved):

michaelblyons (approved):

PR #4078, opened 2024-10-31
[C] Update keywords-parens for alignof and static_assert

braewoods:
I overlooked this previously but I believe this is an optimization for sizeof that would also apply these two that I previously added.

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -854,7 +854,7 @@ contexts:
           scope: constant.other.c
 
   keywords-parens:
-    - match: '\b(sizeof)\b\s*(\()'
+    - match: '\b(sizeof|alignof|_Alignof|static_assert|_Static_assert)\b\s*(\()'
       captures:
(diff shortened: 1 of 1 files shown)


deathaxe:
Small tests would probably avoid regressing that change by future changes.

braewoods:
What would you suggest then? I already added tests for these previously.

deathaxe:
Obviously not for parenthesed variants, I guess. Otherwise they would have been failing due this commit missing.

braewoods:
> It turns out the `keywords-parens` to be redundant, because both keywords (e.g. `sizeof` and parentheses aka. groups are already matched by other patterns.

Should they be removed then?

deathaxe:
There are more such constructs. Not familiar enough with the syntax definition's architecture to say that for sure. Maybe safer to keep them until a rewrite.

deathaxe (approved):

jfcherng (approved):

PR #4080, opened 2024-10-31
[C] Add C11 _Atomic type qualifier and specifier

braewoods:
This was originally added in C11. It is an optional feature but is often available due to being related to threads. It can be as a type qualifier like const or as a type specifier like typeof. It is implemented under both syntax rules to reflect this.


deathaxe (approved):

braewoods:
Did something happen? I thought I'd have gotten another review at least by now.

michaelblyons (approved):
Thanks for the bump. Sometimes I close things out of my inbox without handling them.

To be clear, though: In the overall TTL for PRs in this repo, yours have been handled with alacrity.

PR #4081, opened 2024-10-31
[C] Add POSIX 2024 includes to available completions

braewoods:
This excludes any already covered by the standard includes. Since these are available to all C family languages, there's no need to be as selective as we are with the standard includes.


braewoods:
I thought I would try my hand at add completions for POSIX headers. I intend to do the same with Win32 if this is accepted.

PR #4082, opened 2024-10-31
[XML] Restrict tag snippets

deathaxe:
This commit makes sure to only provide XML tag snippets in top-level contexts by excluding them from `meta.tag`, comments, strings and possible embedded or interpolated expressions.

Inspired by PR #4056


keith-hall (approved):

michaelblyons:
Is there a reason you have `text.xml` in all the negated scopes?

deathaxe:
I want to make sure not to disable snippets if XML is embedded into another language.
See XML highlighted heredoc in Perl for instance.

That's a general scheme to make sure to only match scopes under a certain main scope but not above it.

michaelblyons (approved):

PR #4087, opened 2024-11-07
[C] Add support for C23 _BitInt type

braewoods:
This is implemented like typeof() because it can only be used with ()s to indicate how many bits the resulting integer type will have.

While I was at it, I also added support for the new suffixes for _BitInt integer constants.


deathaxe (approved):

braewoods:
I noticed there was a related addition in C23 for specifying the exact bits used by the integral type for printf and scanf so I have added that now.

michaelblyons (approved):

deathaxe (approved):

PR #4088, opened 2024-11-07
[C] Add support for alignas and its aliases

braewoods:
This was first added in C11 but has since been revised by C23. It functions like typeof but modifies the alignment of the type or variable instead of being a type itself.


michaelblyons (dismissed):

braewoods:
I updated the commit to resolve conflicts with master.

deathaxe (approved):

michaelblyons (approved):

PR #4091, opened 2024-11-09
[C] Add support for decimal floating-point types

braewoods:
This was introduced in C23 and adds support for the new type keywords, the new suffixes for the new types, and the new length modifiers for printf and scanf.

Furthermore, the existing matching rules for floats was changed to remove the possibility of '1f' being classified as a floating point literal. This is actually considered to be an integral literal with an invalid suffix instead of a floating point literal by all compilers that I know of and such produces compiler errors if such code is compiled.

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -20,8 +20,6 @@ variables:
   dec_digit: '\d'
   hex_digit: '\h'
 
-  dec_digits: '{{dec_digit}}+'
-
   # number exponents
   dec_exponent: '(?:[eE][-+]?{{dec_digit}}*)'
   hex_exponent: '(?:[pP][-+]?{{dec_digit}}*)'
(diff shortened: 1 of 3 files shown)


braewoods:
The only part I believe to be questionable is the changes related to the `1f` matching case. I removed it and updated the tests according because that simply is not valid C or C++ at all. It will be seen as an integer literal with the suffix of `f` or `F` which is a syntax error to any compiler I've ever used to my knowledge.

C++/C.sublime-syntax
```diff
@@ -306,12 +303,11 @@ contexts:
...
-        6: invalid.illegal.numeric.suffix.c
-        7: constant.numeric.value.c
-        8: punctuation.separator.decimal.c
-        9: constant.numeric.suffix.c
-        10: invalid.illegal.numeric.suffix.c
+        5: invalid.illegal.numeric.suffix.c
+        6: constant.numeric.value.c
+        7: punctuation.separator.decimal.c
+        8: constant.numeric.suffix.c
+        9: invalid.illegal.numeric.suffix.c
```
deathaxe, line 310: Without distinct double/float suffixes, we can probably simplify the pattern further 

```yml
    - match: |-
        (?x)
        (
          \b{{dec_digits}} 
          (?:
          # 1., 1.f, 1.L, 1.1, 1.1e1, 1.1e-1, 1.1f, 1.1e1f, 1.1e-1f, 1.1L, 1.1e1L, 1.1e-1L
            (\.) (?: {{dec_digit}}+ | (?!\.) ) {{dec_exponent}}?
          # 1.e1, 1.e-1, 1.e1f, 1.e-1f, 1.e1L, 1.e-1L
          | {{dec_exponent}}
          )
        # .1, .1e1, .1e-1, .1f, .1e1f, .1e-1f, .1L, .1e1L, .1e-1L
        |  (\.) {{dec_digit}}+ {{dec_exponent}}?
        )
        (?: ({{double_suffix}}) | ({{dec_suffix}}) )?
      scope: meta.number.float.decimal.c
      captures:
        1: constant.numeric.value.c
        2: punctuation.separator.decimal.c
        3: punctuation.separator.decimal.c
        4: constant.numeric.suffix.c
        5: invalid.illegal.numeric.suffix.c
```

braewoods:
I have updated the commit to resolve the conflict.

braewoods:
This revises the matching rule to the simplified version presented earlier by @deathaxe.

braewoods:
I've noticed that hexadecimal floats also need to be updated but I think that's best done in a different PR.

michaelblyons (approved):

deathaxe (approved):

PR #4093, opened 2024-11-11
[C] Make function parameter names optional

braewoods:
Function parameter names are now optional in C and C++. Previously the syntax highlighting rules would assume that the last identifier found in a parameter was the parameter name but this was making it impossible to properly highlight parameters without a parameter name.

The only solution I could find was to include the specific types and modifiers before the more general match for the parameter name. With this, the types and modifiers are no longer mistaken for a parameter name.

### DIFF
--- a/C++/syntax_test_cpp.cpp
+++ b/C++/syntax_test_cpp.cpp
@@ -2547,6 +2547,23 @@ MyEnum MACRO1
 // Test preprocessor branching and C blocks
 /////////////////////////////////////////////
 
+int bar(int, int const *, int const * const);
+/*  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.function */
+/*     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.function.parameters meta.group */
+/*     ^ punctuation.section.group.begin */
+/*                                         ^ punctuation.section.group.end */
+/*                                          ^ punctuation.terminator */
+/*      ^^^ storage.type */
+/*         ^ punctuation.separator */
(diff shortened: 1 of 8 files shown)


C++/syntax_test_cpp.cpp
```diff
@@ -2547,6 +2547,23 @@ MyEnum MACRO1
...
+/*      ^^^ storage.type */
+/*         ^ punctuation.separator */
+/*           ^^^ storage.type */
+/*               ^^^^^ storage.modifier */
+/*                     ^ keyword.operator */
+/*                      ^ punctuation.separator */
+/*                        ^^^ storage.type */
+/*                            ^^^^^ storage.modifier */
+/*                                  ^ keyword.operator */
+/*                                    ^^^^^ storage.modifier */
```
michaelblyons, line 2565: Is this intentional or a mis-highlight?

braewoods: > Is this intentional or a mis-highlight?

Is what intentional?

braewoods: Yes. `const` can have multiple layers when used in C. `const` can also be applied to the layers of pointers that a type can have. Hence why it can occur here. Otherwise, only part of the type will be `const`. And `const` usually only applies to whatever immediately precedes, except for the special case where `const` comes before the type. For example, `const char` is equivalent to `char const`. If this was a pointer, the array elements would be `const` but not the pointer itself.

deathaxe:
A general approach to support type only signatures, requires different context structure.

Java syntax would be a good starting point to learn how to do so. But I think it might require more significant changes as we first need to properly support all the generics to be able to do so.

deathaxe (approved):

PR #4094, opened 2024-11-16
[JavaScript] Add SQL embedding

jtojnar:
- [x] My commit messages start with the package name in square brackets, e.g. `[XML] `.
- [x] I have included new or enhanced [syntax tests](https://www.sublimetext.com/docs/syntax.html#testing) to cover the changes.

This is useful for server-side JS.

Especially using one of the following libraries:
https://github.com/felixfbecker/node-sql-template-strings
https://github.com/blakeembrey/sql-template-tag
https://github.com/XeCycle/pg-template-tag

Using `version: 1` for the Embedding syntaxes since they need to match the inherited `SQL.sublime-syntax`.

It does not currently handle interpolation inside SQL strings.

--- 

~~The SQL template is detected but the SQL syntax highlighting is not performed (`source.sql.embedded.js` is the deepest scope anything within the template string gets).~~

**Edit:** If I change `extends: Packages/SQL/SQL.sublime-syntax` to `CSS.sublime-syntax` in `SQL (JS template).sublime-syntax`, it highlights as CSS correctly.

**Edit2** It was caused by a differing version.

### DIFF
--- a/JavaScript/Embeddings/SQL (for JS template).sublime-syntax
+++ b/JavaScript/Embeddings/SQL (for JS template).sublime-syntax
@@ -1,0 +1,14 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/syntax.html
+# highlight tagged template strings
+scope: source.sql.js-template
+version: 1
+hidden: true
+
+extends: Packages/SQL/SQL.sublime-syntax
+
+contexts:
+  prototype:
+    - meta_prepend: true
+    - include: scope:source.js#text-interpolations
--- a/JavaScript/Embeddings/SQL (for TS template).sublime-syntax
+++ b/JavaScript/Embeddings/SQL (for TS template).sublime-syntax
@@ -1,0 +1,14 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/syntax.html
+# highlight tagged template strings
+scope: source.sql.ts-template
+version: 1
+hidden: true
+
+extends: Packages/SQL/SQL.sublime-syntax
+
+contexts:
+  prototype:
+    - meta_prepend: true
+    - include: scope:source.ts#text-interpolations
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1295,6 +1295,18 @@ contexts:
         0: meta.string.template.js string.quoted.other.js
         1: punctuation.definition.string.end.js
       pop: 1
+    - match: (sql|SQL)\s*((\`){{trailing_wspace}}?)
+      captures:
+        1: variable.function.tagged-template.js
+        2: meta.string.template.js string.quoted.other.js
+        3: punctuation.definition.string.begin.js
+      embed: scope:source.sql.js-template
+      embed_scope: meta.string.template.js source.sql.embedded.js
+      escape: '{{leading_wspace}}?(\`)'
+      escape_captures:
+        0: meta.string.template.js string.quoted.other.js
+        1: punctuation.definition.string.end.js
+      pop: 1
     - match: (?:({{identifier_name}})\s*)?(\`)
       captures:
         1: variable.function.tagged-template.js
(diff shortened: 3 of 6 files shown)


michaelblyons:
Does the SQL bump to version 2 then require a bunch of other "host" syntaxes to bump, too?

jtojnar:
It requires the bump of the Embeddings syntaxes. For example, `SQL (for Ruby).sublime-syntax` needs to be `version: 2`, or the SQL example in `syntax_test_ruby.rb` will not be highlighted.

Curiously, `Ruby.sublime-syntax` itself does not seem to need to be bumped.

michaelblyons:
It turns out my local set-up includes #3046, which also bumps to version 2. Sounds good to me.

deathaxe:
Only inherited syntaxes need to be of same version. It is no problem to include or embed a v2 into a v1 or vice versa.

Bumping SQL to v2 and adding support for SQL in tagged template strings however are 2 distinct steps.

The more interesting question is what effect #3046 will have in combination with this one as JS is embedded, extended and used in nearly all template syntaxes. I wonder when we hit context sanity limits again. :D

deathaxe:
With regards to https://github.com/sublimehq/Packages/pull/4035, any purley embedded syntax should be named `<name> (for TS template).sublime-syntax`.

Syntaxes in Embedding are currently not correctly named by accident.

JavaScript/Embeddings/SQL (JS template).sublime-syntax
```diff
@@ -0,0 +1,18 @@
...
+extends: Packages/SQL/SQL.sublime-syntax
+
+contexts:
+  prototype:
+    - meta_prepend: true
+    - include: scope:source.js#text-interpolations
+
+  strings-content:
+    - meta_prepend: true
+    - include: scope:source.js#string-interpolations
```
deathaxe, line 18: Current SQL syntax doesn't use this context. It wouldn't have any effect.

It doesn't support `meta.string string` vs. `meta.string meta.interpolation` in general.

deathaxe: `string-interpolations` removes a single scope from the context it is included to.

This is primarily required as we want to clear `string` scope in interpolations.

```
"Text { interpolated.variable } !"
^^^^^^ meta.string string - meta.interpolation
      ^^^^^^^^^^^^^^^^^^^^^^^^^ meta.string meta.interpolation - string
                               ^^^ meta.string string - meta.interpolation
```

That's required, because `string` is treated special but we want interpolations to be treated as normal source code. And we don't want it to be highlighted like a string. Any unmatched token in interpolations would be highlighted with string color, which may look odd and broken.

Therefore "modern" syntax definitions exclude `prototype` context from strings to enforce special `string-interpolations` contexts to be added by inheriting syntaxes.

Interpolation in all other non-string contexts is handled by `text-interpolations` which does not clear any scopes.

deathaxe:
> Should I move the first commit out into a separate PR?

I'd say yes. Bumping SQL version is a seperate PR. Actually I don't see any reason for it, at the moment. It would be part of merging #3046.

jtojnar:
> I'd say yes. Bumping SQL version is a seperate PR. Actually I don't see any reason for it, at the moment. It would be part of merging #3046.

Well, it is a prerequisite for this, since the embedding syntaxes inherit from the SQL one. So unless we want to wait with this PR for #3046 – but it has already been open for years –  I have moved the SQL change to https://github.com/sublimehq/Packages/pull/4097.

JavaScript/Embeddings/SQL (for JS template).sublime-syntax
```diff
@@ -0,0 +1,14 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/syntax.html
+# highlight tagged template strings
+scope: source.sql.js-template
+version: 2
```
deathaxe, line 6: Those syntaxes don't need to be `version: 2` to be embeddable in JS/TS.

```suggestion
```

deathaxe: > Well, it is a prerequisite for this,

No

jtojnar: Oh, right. I assumed new syntaxes should be the current latest version but I can make it `version: 1` if it is preferable.

deathaxe (approved):

michaelblyons (approved):

PR #4095, opened 2024-11-16
[batch] prevent ^ from counting as an escape for %

God-damnit-all:
In batch scripts (but not on the command line itself), `%` is only escaped by another `%`

You can test this by making a bat file that reads:
```
@echo off
setlocal
set var=A
if ^A==A echo test1
if ^%var%==A echo test2
if "^A"==^"^^A^" echo test3
if "^%var%"==^"^^%var%^" echo test4
if "^%%var%%"==^"^^%%var%%^" echo test5
endlocal
pause
```

### DIFF
--- a/Batch File/Batch File.sublime-syntax
+++ b/Batch File/Batch File.sublime-syntax
@@ -1753,7 +1753,7 @@ contexts:
     - include: quoted-eol-pop
 
   escaped-characters:
-    - match: \^.
+    - match: \^[^%\n]
       scope: constant.character.escape.dosbatch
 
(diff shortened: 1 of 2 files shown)


michaelblyons:
Can you add regression tests?

deathaxe:
Thank's for taking care and contributing. I hope it is not too aggressive to have added some arbritary syntax tests to verify the fix and help not regressing it by future changes.

deathaxe (approved):

michaelblyons (approved):

PR #4098, opened 2024-11-20
[README] add tip to fix confusing problems when renaming/removing files

keith-hall:
add advice to use ST's entire package replacement feature of having an Installed Package with the same name as a shipped package to avoid confusing problems where out-of-date files are still unexpectedly picked up by ST.


FichteFoll (approved):

FichteFoll:
Haven't thought of that. It's a good idea to mention it in the README.

PR #4099, opened 2024-11-22
[C] Revise hexadecimal floats

braewoods:
This updates hexadecimal floats to follow a design similar to the new decimal float syntax. As a byproduct of the update, it also now matches the latest suffixes added in more recent C standards.

### DIFF
--- a/C++/C.sublime-syntax
+++ b/C++/C.sublime-syntax
@@ -325,14 +325,27 @@ contexts:
         5: invalid.illegal.numeric.suffix.c
 
     # hexadecimal float (C99)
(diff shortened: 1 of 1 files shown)


deathaxe (approved):

michaelblyons (approved):

braewoods:
> In general I'd appreciate tests along your contributions, so we can be sure they are covered in the future.

I generally do. This time I was only correcting the syntax highlighting for an existing syntax construct that already had defined tests. I didn't think there were any new cases to write tests for.

PR #4101, opened 2024-11-24
[C] Add digit separator like is already implemented for C++

### LINKED ISSUE #3972: [C/C++] digit separators
### What happened?

the latest C standard introduced digit separators, i.e. `int foo() { return 1'000'000; }`.
Sublime text will highlight '000' as a string instead

braewoods:
C23 added this feature previously available in C++. This should close #3972 as well as both C and C++ now have syntax support for the ' digit separator.

This was implemented by just copying the digit variables from the C++ syntax file as it is the same construct in the end. The existing tests were updated to fix the part of the existing tests were broken by this new feature.

Furthermore, some tests from C++ were adapted for C to add better testing for the new digit seperator. I also included tests for the newer base 2 integer contants. The C file did not have any tests for that yet.

### DIFF
--- a/C++/syntax_test_c.c
+++ b/C++/syntax_test_c.c
@@ -1133,6 +1133,7 @@ static const unsigned char image_png[] = {
 dec0 = 0;
 /*     ^ meta.number.integer.decimal.c constant.numeric.value.c */
 /*      ^ punctuation.terminator - constant */
+
 dec1 = 1234567890;
 /*     ^^^^^^^^^^ meta.number.integer.decimal.c constant.numeric.value.c */
 /*               ^ punctuation.terminator - constant */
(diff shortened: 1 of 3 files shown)


braewoods:
I fixed the test issues with Objectice-C.

C++/syntax_test_c.c
```diff
@@ -1173,13 +1174,19 @@ dec7 = 1234567890uLL;
 /*                  ^ punctuation.terminator - constant */
 
 dec8 = 1'234_567'890s0f;
-/*     ^ meta.number.integer.decimal.c constant.numeric.value.c */
-/*      ^^^^^^^^^ string.quoted.single */
+/*     ^^^^^ constant.numeric.value.c */
+/*          ^^^^ invalid.illegal.numeric.suffix.c */
 /*               ^^^^^^ meta.number.integer.decimal.c */
```
deathaxe, line 1179: I'd suggest to extend existing `meta.number` test to cover whole constant.

```suggestion
/*     ^^^^^^^^^^^^^^^^ meta.number.integer.decimal.c */
/*     ^^^^^ constant.numeric.value.c */
/*          ^^^^ invalid.illegal.numeric.suffix.c */
```

Objective-C/syntax_test_objc.m
```diff
@@ -920,8 +920,8 @@ - (void)debugOutput:(NSString *)format, ... NS_FORMAT_FUNCTION(1,2)
 /*                  ^ punctuation.terminator - constant */
 
 dec8 = 1'234_567'890s0f;
-/*     ^ meta.number.integer.decimal.c constant.numeric.value.c */
-/*      ^^^^^^^^^ string.quoted.single */
+/*     ^^^^^ constant.numeric.value.c */
+/*          ^^^^ invalid.illegal.numeric.suffix.c */
 /*               ^^^^^^ meta.number.integer.decimal.c */
```
deathaxe, line 925: Same here

```suggestion
/*     ^^^^^^^^^^^^^^^^ meta.number.integer.decimal.c */
/*     ^^^^^ constant.numeric.value.c */
/*          ^^^^ invalid.illegal.numeric.suffix.c */
```

deathaxe (commented):
Maybe following 2 `meta.number` tests can be extended to be inline with the other assertions. Otherwise LGTM!

deathaxe (approved):

braewoods:
I rebased upon latest master.

deathaxe:
FWIW, it is not necessary to always rebase unless there are conflicts.

braewoods:
I'm aware but I got into the habit because of the issues that can be caused from being behind.

michaelblyons (approved):

PR #4110, opened 2024-12-05
[Python] Regexp backref identifiers

michaelblyons:
Centralizes named capture group identifiers in a variable. Adds support for capital letters one place they were missing. Adds support for `\g<...>` backref style.

[Relevant documentation in v3.12](https://docs.python.org/3.12/library/re.html#index-18)

Previously included (but now removed) deprecation scopes for non-ASCII capture group identifiers.

### DIFF
--- a/Python/Embeddings/RegExp (for Python).sublime-syntax
+++ b/Python/Embeddings/RegExp (for Python).sublime-syntax
@@ -14,9 +14,8 @@ variables:
   other_modifiers: \?(?:[ixmsLua]*-)?[ixmsLua]+
 
   # capture group identifiers
-  capture_group_name: (?:[a-zA-Z_][a-zA-Z0-9_]*)
-  # non-ASCII is deprecated since v3.11
-  capture_group_name_deprecated: (?:[[:alpha:]][[:alnum:]_]*)
+  capture_group_name: (?:[[:alpha:]_]\w*)
+  capture_group: (?:[1-9][0-9]*|{{capture_group_name}})
 
 contexts:
   base:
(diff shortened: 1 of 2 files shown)


deathaxe (commented):
Python always used to try to even correctly highlight py2. Do we need to bother users with deprecated highlighting, really?

Python/Embeddings/RegExp (for Python).sublime-syntax
```diff
@@ -25,21 +30,52 @@ contexts:
 
   group-start:
     - meta_prepend: true
-    - match: (\?P=)([a-zA-Z_][a-zA-Z_0-9]*\w*)(?=\))
+    - match: (\?P=)({{capture_group_name}})(?=\))
       captures:
         1: keyword.other.back-reference.named.regexp
         2: variable.other.backref-and-recursion.regexp
-    - match: (\?P)(<)([a-z]\w*)(>)
```
michaelblyons, line 32: This part of the change is significant. Capture groups can start with capital letters or underscores, too.

michaelblyons:
I can yank the deprecated highlighting and tests, but there's a real change that is still necessary.

I discovered it trying to name the groups in a capture as specific keys for `groupdict()`.

deathaxe:
I am fine with required fixes, if current patterns don't match all valid names. 

It's just the attempt to do linting by scoping no longer supported nonascii "deprecated". pyright and co. will already highlight them depending on set python version. They should however be highlighted normally by syntax as they are valid in older (even ST's default python versions) - even though actually rather unlikely to appear in real world projects, which is just another reason for not needing that extra complexity.

deathaxe (approved):

Python/Embeddings/RegExp (for Python).sublime-syntax
```diff
@@ -25,21 +29,30 @@ contexts:
...
       captures:
         1: keyword.other.backref-and-recursion.conditional.regexp
         2: punctuation.definition.group.begin.assertion.conditional.regexp
         3: variable.other.back-reference.regexp
         4: punctuation.definition.group.end.assertion.conditional.regexp
+
+  backrefs:
+    - meta_prepend: true
+    - match: \\g(<)({{capture_group}})(>)
+      scope: keyword.other.backref-and-recursion.regexp
```
deathaxe, line 54: The other patterns only seem to scope initial part (e.g. `?P=` or `?P`) keyword, while this pattern scopes `\g` as well as whole name part `keyword`.

Maybe we should decide for one style.

Scoping everything would make it easier to address the whole expression with single simple selector, while it on the other hand seems a bit weird to scope actual name `keyword variable`.

deathaxe: OK, normal Regexp.sublime-syntax also scopes it this way. Maybe the other ones `?P=` need some adjustmant to align them. Can be modified in a separate PR however.

michaelblyons: I agree with you. I don't like the status quo, but I think we should RFC that.

deathaxe:
Following example from LaTeXTools would be fixed by this PR.

grafik

FichteFoll (approved):

PR #4111, opened 2024-12-07
[Diff] Refactor for Diff3, Git, and conflicts

michaelblyons:
Big refactor of Diff syntax according to the GNU and SVN documentation. New Git Diff syntax, also used by Git Log.

Thank you to the reviewers!

## Notable changes

- Lots of documentation links
- Lots of new tests, mostly pulled from relevant documentation
- Some scope tweaking, but preserving the stacking from #2178 for backwards-compatibility
- New syntax for Git Diff emails (`git am` and `git format-patch`)
    - When the normal Diff syntax detects the "magic" first line, it delegates out to Git Diff instead.
- New support for
    - Diff3 formats
    - Conflict sections
    - "Combined" unified diffs, only in Git Diff

## Potentially controversial

-  Conflicts somewhat with #4047, but should now be nearly compatible.
- `storage.modifier` distinguishes between the two Git file modes (namely `644` and `755`)
- Markup scopes color the `++++----` and similar, even though they're not *actually* markup.
- Users and emails are delegated out to Mailmap.
- Whatever you noticed.


Diff/Diff.sublime-syntax
```diff
@@ -1,75 +1,381 @@
...
+      push:
+        - match: '( (-) )(.* (={4})){{eol}}'
+          captures:
+            1: meta.header.diff
+            2: punctuation.separator.sequence.diff
+            3: meta.diff.header.to-file meta.header.to-file.diff
+            4: punctuation.definition.to-file.diff
+          pop: 1
+        - include: pop-immediately
+    - match: ^(\*{15}|={67}){{eol}}
```
michaelblyons, line 121: Also, does anyone have any idea what these rules are for? I left them in there (refactored), but I cannot find any citation, or even an example outside the test file for these specific numbers of `=` or `*` being a separator.

michaelblyons:
Am I supposed to see output on the failing unit tests? I don't *think* anything's failing locally.

keith-hall:
> Am I supposed to see output on the failing unit tests? I don't _think_ anything's failing locally.

Presumably it is the error about Diff mixed inheritance versions at the very top of the CI console output

deathaxe:
Should be fixed by updating Packages/PHP/Embeddings/Diff (for PHP Interpolated).sublime-syntax to version 2.

michaelblyons:
🤦‍♂️ I think my ad blocker was preventing display of the test output. Thanks!

Diff/Diff.sublime-syntax
michaelblyons: The `***************` appears to be a separator between hunks in a file in "Context" mode. Still curious about the 67 consecutive `=`s.

deathaxe: Maybe just a similar app specific separator.

michaelblyons:
> Currently you're only matching English headers. They might not be in English as Git supports a few languages. You'd have to look that up.

For this, you are referring to the `Signed-off-by:`, etc., yes?

Git Formats/Git Diff.sublime-syntax
```diff
@@ -0,0 +1,511 @@
...
+    diff[ ]--(?:git|combined|cc)[ ]
+    | From [ ] \w+@z [ ]
+      Thu [ ] Jan [ ]{2} 1 [ ] 00:00:00 [ ] 1970  # git-send-email
+    | From [ ] (?:[0-9a-fA-F]{40}|[0-9a-fA-F]{64}) [ ]
+      Mon [ ] Sep [ ]   17 [ ] 00:00:00 [ ] 2001  # git-format-patch
+  )
+
+variables:
+  hash: '[0-9a-fA-F]{7,64}'
+  day3: (?:Mon|Tue|Wed|Thu|Fri|Sat|Sun)
```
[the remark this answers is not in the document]
michaelblyons: Found https://github.com/git/git/tree/master/po, but I'm not seeing month and day-of-week abbreviations.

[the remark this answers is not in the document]
deathaxe: FWIW, language specific patterns in _Git Commit_ syntax were added as those are translated to set UI language by git and wouldn't otherwise be highlighted in non-english environments.

AFAIK abbreviation in date-time strings are not translated.

Git Formats/Git Diff.sublime-syntax
```diff
@@ -0,0 +1,567 @@
...
+  # https://git-scm.com/docs/git-format-patch#Documentation/git-format-patch.txt---no-signatureltsignaturegt
+  signature:
+    # This is the Git version unless otherwise configured.  You will notice
+    # that once inside this context, you cannot leave.  That seems to be how
+    # it works for these files.
+    - match: ^(-- ){{eol}}
+      captures:
+        1: meta.separator.diff punctuation.section.block.diff
+      set:
+        - meta_content_scope: comment.block.diff
```
deathaxe, line 411: Is it really worth a dedicated context or can this pattern just be added to `diffs`?

Seems rather unlikely to be subject of inheritance.

Instead probably use `signature` for signature content.

```yml
  diffs:
    - match: ^(diff) ((--)(?:git|combined|cc))
      captures:
        1: variable.function.diff
        2: variable.parameter.diff
        3: punctuation.definition.parameter.diff
      push: [diff-header, file-header-continue]
    # https://git-scm.com/docs/git-format-patch#Documentation/git-format-patch.txt---no-signatureltsignaturegt
    # This is the Git version unless otherwise configured.  You will notice
    # that once inside this context, you cannot leave.  That seems to be how
    # it works for these files.
    - match: ^(-- ){{eol}}
      captures:
        1: meta.separator.diff punctuation.section.block.diff
      set: signature

  signature:
    - meta_content_scope: comment.block.diff
 ```

File: Diff/Diff.sublime-syntax [hunk 4]
```diff
@@ -1,75 +1,402 @@
...
 contexts:
   main:
-    - match: '^((\*{15})|(={67})|(-{3}))$\n?'
+    - match: ^(?={{git_first_line}})
+      push: Packages/Git Formats/Git Diff.sublime-syntax#email-first-line
+    - include: leading-text
+    - include: line-ranges
+    - include: headers
+    - include: deltas
+    - include: diff3-normal
```
deathaxe, line 32: As _Git Diff_ extends from this syntax all these contexts are compiled into _Git Diff_ as well even though they seem unused. It shouldn't cause any performance issues as they are unreferenced in _Git Diff_ but they cause larger syntax cache file.

Hence I wonder how big overlap between _Diff_ and _Git Diff_ really is. I just found a couple of small contexts on a first glance.

Depending on the answer I'd suggest to ...

1. either decouble syntaxes completely, by duplicating those patterns - if there are only few
2. or create a common base syntax, which only contains shared contexts.


So with the comment above, _Diff_ and _Git Diff_ are basically compiled to the same binaries more or less, sharing all their contexts.

That's somehow odd, to be honest.

Diff/Diff.sublime-syntax
```diff
@@ -1,75 +1,402 @@
...
+  | \*\*\*.*\d{4}\s*$
+  | \d+(,\d+)* [adc] \d+(,\d+)* $
+  )
 
-scope: source.diff
 contexts:
   main:
-    - match: '^((\*{15})|(={67})|(-{3}))$\n?'
+    - match: ^(?={{git_first_line}})
+      push: Packages/Git Formats/Git Diff.sublime-syntax#email-first-line
```
deathaxe, line 27: Note this means to literly include all contexts of _Git Diff_ into this syntax as well causing _Diff_ to be a merge of both, while _Gitt Diff_ being extended from this one.

> By commenting out this line _Diff.rcache_ size is reduced from 38kB to 7kB for instance.

One could ask, why 2 syntaxes than, because 99% of git related diff/patch files use `.patch` file extension and hence will be opened with this _Diff_ syntax, requiring this pattern to be highlighted as git diff.

It's an interesting mix and hack of first_line_match, at least.

Git Formats/Git Diff.sublime-syntax
```diff
@@ -0,0 +1,567 @@
...
+name: Git Diff
+scope: source.diff.git
+version: 2
+extends: Packages/Diff/Diff.sublime-syntax
+
+file_extensions:
+  - diff.eml
+  - patch.eml
+  - diff.mbox
+  - patch.mbox
```
deathaxe, line 15: Most likely linux kernel hackers will be the only ones who benefit from opening their patches with this syntax directly.

That's the first time I've learned about these extensions. 

Creating patches with git or Sublime Merge just create `.patch` files, which are opened with _Diff_.

Diff/Diff.sublime-syntax
```diff
@@ -1,75 +1,402 @@
...
+        1: punctuation.definition.from-file.diff
+      push:
+        - match: '( (-) )(.* (={4})){{eol}}'
+          captures:
+            1: meta.header.diff
+            2: punctuation.separator.sequence.diff
+            3: meta.diff.header.to-file meta.header.to-file.diff
+            4: punctuation.definition.to-file.diff
+          pop: 1
+        - include: pop-immediately
```
deathaxe, line 73: Named context?

```suggestion
      push: header-body

  header-body:
    - match: '( (-) )(.* (={4})){{eol}}'
      captures:
        1: meta.header.diff
        2: punctuation.separator.sequence.diff
        3: meta.diff.header.to-file meta.header.to-file.diff
        4: punctuation.definition.to-file.diff
      pop: 1
    - include: pop-immediately
```

Diff/Diff.sublime-syntax
```diff
@@ -1,75 +1,402 @@
...
+        - include: path-separators
+    - match: ^(Prereq)(:)[ \t]*
+      captures:
+        1: meta.mapping.key.diff keyword.other.diff
+        2: punctuation.separator.key-value.diff
+      push:
+        - meta_scope: meta.mapping.diff
+        - meta_content_scope: meta.mapping.value.diff string.unquoted.diff
+        - include: pop-eol
+        - include: path-separators
```
deathaxe, line 54: Named contexts?

```suggestion
  leading-text:
    - match: ^(Index)(:)[ \t]*
      captures:
        1: meta.mapping.key.diff keyword.other.diff
        2: punctuation.separator.key-value.diff
      push: index-body
    - match: ^(Prereq)(:)[ \t]*
      captures:
        1: meta.mapping.key.diff keyword.other.diff
        2: punctuation.separator.key-value.diff
      push: prereq-body

  index-body: 
    - meta_scope: meta.mapping.diff meta.diff.index meta.index.diff
    - meta_content_scope: meta.mapping.value.diff meta.toc-list.file-name.diff
    - include: pop-eol
    - include: path-separators

  prereq-body: 
    - meta_scope: meta.mapping.diff
    - meta_content_scope: meta.mapping.value.diff string.unquoted.diff
    - include: pop-eol
    - include: path-separators
```

deathaxe (commented):
There's not much to say. Really good and clean implementation.

The only thing I can't really wrap the head around is the way how _Diff_ and _Git Diff_ are related to each other, causing ST to basically compile two more or less identical syntax definitions for both. That seems a bit like wasting resources.

Git Formats/Git Diff.sublime-syntax
michaelblyons: I don't remember exactly. I may have invented these extensions. If I just say `patch` here, what happens when a user opens a `patch` file? Would anyone unexpectedly get the Git one?

deathaxe: IIRC syntaxes loaded later, overwrite extension assignments.

Diff/Diff.sublime-syntax
michaelblyons: Thanks for pointing this out. There are a couple ways to handle it that I've thought about (and perhaps more that I hadn't considered). Please feel free to comment.

### Merge Git Diff back into Diff

There's not really a big drawback to this. I originally had them both in the same file. I split them because the Git part started feeling unwieldy compared to the nice, tight Diff part.

### Decouple completely
*Feels* unsatisfying, but there aren't really any practical drawbacks. (Possible exception: Does the Diff "match-Git-first-line / push Git Diff" end up duplicating all those contexts back anyway?)

### Base syntax (simple)
Just centralize the unified diff stuff into a base syntax.

### Base syntax (split)
 Create a set of very small Diff syntaxes each handling their specific case, and then mash them all together for a generic Diff syntax.

- Diff (common)
    - line-marked diffs
        - Diff Normal
        - Diff Context
        - Unified Diff
            - Git Diff
    - section-marked diffs
        - Diff3 Normal
        - Diff3 Edit
        - Conflict (?)
    - ??

With a main Diff syntax that extends everyone else's for convenience. 

#### Pros
- You have the opportunity to rigidly highlight as a particular format if you choose to.
#### Cons
- It's a lot of tiny files that *probably* no one will care to use.
- We may get to find out what happens when one syntax prepends to a context, another one replaces it completely, and a third one inherits from both.

Diff/Diff.sublime-syntax
michaelblyons: Haha. In my commit message from the one that adds this, you can see that I wasn't even confident it would work at all. :stuck_out_tongue_winking_eye:

michaelblyons: Maybe the duplication can be avoided by making an `embed` with an unmatchable escape sequence?

deathaxe: At least git diff-only contexts wouldn't be part of normal diff syntax anymore.

If it is a "first line match", maybe the following would make sense to avoid starting "Git Diff" later on in a document?

Chosen `diffs` as it is the main diff context in _Git Diff_.

```yml
contexts:
  main:
    - match: ^(?={{git_first_line}})
      embed: Packages/Git Formats/Git Diff.sublime-syntax#email-first-line
      escape: (?!)
    - match: ^
      set: diffs

  diffs:
    - include: leading-text
    - include: line-ranges
    - include: headers
    - include: deltas
    - include: diff3-normal
```

Git Formats/Git Diff.sublime-syntax
deathaxe: Not sure, how common they are. At least appear to be mailing-list-workflow-specific.

I can't judge, whether they are useful or may cause confusion in future.

Diff/Diff.sublime-syntax
deathaxe: Last option indeed feels a bit like overkill with regards to amount of files and scattering contexts. If all those patterns don't conflict it is ok to have them in one file.

How much would a common _Diff (Basic)_ syntax contain? If it is enough, it could be a base for various dialects like _SQL (Basic)_ or _RegExp (Basic)_. Maybe most cleanest solution.

If it however contains only 2 - 6 tiny contexts, it might probably also ok to just drop normal Diff contexts in _Git Diff_ by setting them empty via e.g. `diff3-normal: []`. This drops unused contexts but looks a bit messy if there are too many of them.

Diff/Diff.sublime-syntax
```diff
@@ -1,75 +1,408 @@
...
+      scope: punctuation.separator.sequence.diff
+    - match: \B\.\.(?=/|\\)
+      scope: constant.other.path.parent.diff
+    - match: \B\.(?=/|\\)
+      scope: constant.other.path.self.diff
+
+  date:
+    # 1970-01-01 00:00:00 +0000
+    - match: \b1970(-)01(-)01 0?0(:)00(:)00(?:(\.)0+)?(?:[ ]([+-]?0000|UTC))?
+      scope: meta.date.diff constant.language.null.diff
```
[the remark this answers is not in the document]
deathaxe: The scope? The pattern scopes textual representation of "unixtimestamp = 0" special. Why not?

Diff/Diff.sublime-syntax
```diff
@@ -1,75 +1,408 @@
...
+        2: punctuation.separator.key-value.diff
+      push: index-body
+    - match: ^(Prereq)(:)[ \t]*
+      captures:
+        1: meta.mapping.key.diff keyword.other.diff
+        2: punctuation.separator.key-value.diff
+      push: prereq-body
+
+  index-body: 
+    - meta_scope: meta.mapping.diff meta.diff.index meta.index.diff
```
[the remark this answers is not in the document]
deathaxe: https://github.com/sublimehq/Packages/blob/b2d49371e0845034b0e89c556811ff1475afdab2/Diff/Diff.sublime-syntax#L72

Backward compatibility. 

Various color schemes target `meta.diff.[header|index|range]`. Removing them would cause all those color schemes to break.

Diff/Diff.sublime-syntax
michaelblyons: What do you think about https://github.com/michaelblyons/SublimeSyntax-Defaults/pull/7 ? I tried making a "Diff (Basic)," but there were some dependency chains that made it kind of ugly. This way is much cleaner.

Diff/Diff.sublime-syntax
michaelblyons: I don't really understand how, but I think [6826deb](https://github.com/sublimehq/Packages/pull/4111/commits/6826deb45c4e7d29a24264dedd8d6102879ed88b) is messing up other syntaxes that embed Diff.

Diff/Diff.sublime-syntax
```diff
@@ -24,7 +24,12 @@ first_line_match: |-
 contexts:
   main:
     - match: ^(?={{git_first_line}})
-      push: Packages/Git Formats/Git Diff.sublime-syntax#email-first-line
+      embed: Packages/Git Formats/Git Diff.sublime-syntax#email-first-line
```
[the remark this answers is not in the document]
michaelblyons: I thought of that, but I don't think those other syntaxes would be matching on the `git_first_line` pattern anyway.

I should probably update Git Log to use the new Git Diff syntax, but that is independent from this problem.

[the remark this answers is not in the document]
deathaxe: We just want to forward to _Git Diff_ in case of certain patterns. Hence it's not the correct use case for an `meta.embedded` scope.

`meta.embedded` is or should primarily be used to scope evaluated scripts in template languages. Something like `<?php ...?>` or `{% ... %}` is meant to receive that scope.

Diff/Diff.sublime-syntax
deathaxe: Maybe caused by `set: diffs`. `push: diffs` should solve it.

Diff/Diff.sublime-syntax
deathaxe: I wouldn't primarily distinguish between unified as base and the rest, but just put common stuff shared accross GNU Diff and Git Diff into a base syntax.

That said, I take on your approach by removing dedundants and renaming contexts, preparing a suggestion for an (hopefully) optimized extending architecture. I have a feeling of it possibly resolving some of current weirdnesses.

michaelblyons: Sorry I didn't notice your comment before I did some force-pushing. Hopefully not too disruptive. :sweat_smile:

michaelblyons: Thanks for your help! The suggested changes look great.

michaelblyons:
Should we merge *Conflict.sublime-syntax* into the Diff syntax like #4047?

deathaxe:
Unsure. As both don't share anything and merge conflicts appear without anything diff related, a stand-alone syntax is ok, too, I guess. Normal diffs won't ever contain conflict markers, would they?

michaelblyons:
> Normal diffs won't ever contain conflict markers, would they?

No, but after I filed "Conflict?" into my broad organization of diff types, I'm much more willing to concede merging it. Almost none of the other diff types share any commonality with each other. They're just simple enough that they can coexist in the same file with no issue. Conflict could do the same (though I would change the variable names).

I'm considering adding a side-by-side diff highlighter, too, although I haven't looked too hard. (I think it'd be hard to do perfect, but easy to do good-enough.)

deathaxe:
I am fine with both, keeping or merging.

michaelblyons:
I've brought it back in, along with a conservative side-by-side highlighter. A more risky side-by-side syntax at https://github.com/michaelblyons/SublimeSyntax-Defaults/pull/10

Git Formats/tests/syntax_test_git_diff.patch
```diff
@@ -0,0 +1,178 @@
...
++++ b/Gemfile
+@@ -24,6 +24,7 @@ group :development do
+   gem 'listen', '~> 3.0.5'
+   gem 'spring'
+   gem 'spring-watcher-listen', '~> 2.0.0'
++  gem 'rspec-rails'
+ end
+
+ gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+diff --git a/app/assets/images/logo.jpeg b/app/assets/images/logo.jpeg
```
michaelblyons, line 112: In current Git Log format, lines like these are marked as `source.shell`. Should I do that here, too, or is this way fine?

Git Formats/tests/syntax_test_git_diff_combined.patch
```diff
@@ -0,0 +1,50 @@
+\ SYNTAX TEST "Packages/Git Formats/Git Diff.sublime-syntax"
+diff --combined describe.c
+index fabadb8,cc95eb0..4866510 644
```
michaelblyons, line 3: In Git Log, the `..` is marked as `punctuation.sequence`. I have it here as `keyword.operator.logical`. I can make a case either way, but feel free to express an opinion. (For these "combined" diffs, the `,`  is marked as `punctuation`.)

Git Formats/tests/syntax_test_git_diff.patch
deathaxe: The primary reason I think is re-using _Bash_ syntax to highlight this line.

I don't think we need `source.shell`, but maybe a function-call like scope may be usefull, so ST displays those `diff` lines with proper kind icon in symbol list. Currently it displays `v` (variable?).

Git Formats/tests/syntax_test_git_diff_combined.patch
deathaxe: Today, I'd probably use `punctuation.separator.range.diff` as those denote a list of arbitrary commits starting with `cc95eb0` and finalized by `4866510`.

It would match to recent changes made to line range markers in (`diff-line-ranges`).

Git Formats/tests/syntax_test_git_diff.patch
deathaxe: Added a proposal in 15622fc including updated interpretation of argument scopes. I am uncertain whether `entity.name` is a proper scope for path names though. I'd tend to just scope them string. However saw `entity.name` also being used above and wonder whether all paths should receive a common scope.

michaelblyons: I think I borrowed scope names for paths from Git Ignore. I'm fine with a scope change, so long as things are consistent.

michaelblyons: Whatever the scope is, we should make sure at least the paths (ideally together) are available in the local symbols. "Together" so that you don't have to scroll through each one separately in the symbol list. They're almost always the same path anyway.

Git Formats/tests/syntax_test_git_diff_combined.patch
michaelblyons: This sounds good. I think `..` and `...` are technically set operators, but people tend to think of them as ranges.

deathaxe: Like a python set `{a, b, c}`? Not sure I get it.

Git Formats/tests/syntax_test_git_diff.patch
deathaxe: Symbol list should show the whole line due to everything being in `meta.toc-list`.

Git Formats/tests/syntax_test_git_diff_combined.patch
michaelblyons: I was thinking more like `in`.

Git Formats/tests/syntax_test_git_diff.patch
michaelblyons: Ah, good. I was on phone when I wrote that above.

What scope do you like? `meta.path string.unquoted`?

deathaxe: This was what came to my mind first as it would match closest with current shell argument scopes, which use `string`. I haven't however double checked all locations of possible file names and paths and whether `string` would be appropriate there.

Finally those syntaxes are so special purpose, we don't need to turn scope naming into rocket science.

michaelblyons:
Reworking Git Log to use Git Diff contexts at https://github.com/michaelblyons/SublimeSyntax-Defaults/pull/11

michaelblyons:
> let us know when/if you need any review or feedback 😁

Haha. Sorry if you're flooded in notifications. The PR is a lot better after you and DeathAxe's reviews. I think if you're looking for something, check my PRs to this PR (michaelblyons#11 and michaelblyons#10) and decide if you care about any of the things in the "controversial" heading in OP.

Thanks again.

deathaxe:
Empty lines between contexts is ok, but starting arbritary emptly lines between patterns is too much.

michaelblyons:
There are places where I have considered newlines in stacked scopes, but I am not planning to add an empty line between meta and includes. PackageDev scoping is sufficient to make that distinction very obvious.

deathaxe (dismissed):

deathaxe (dismissed):

deathaxe (dismissed):

deathaxe (approved):

Git Formats/Git Diff.sublime-syntax
michaelblyons, line 1: I loaded a patch from patchwork.kernel.org and I was surprised not to get the full highlighting. It did not have a "magic" date. If the more lenient `git_first_line` should be bumped to Diff (Basic), let me know.

Diff/Diff (Basic).sublime-syntax
```diff
@@ -0,0 +1,190 @@
...
+  git_hash40: '[0-9a-fA-F]{40}'
+  git_hash64: '[0-9a-fA-F]{64}'
+  git_hash_full: (?:{{git_hash64}}|{{git_hash40}})
+  git_first_line: |-
+    {{bol}}(?x:
+      From [ ] \w+@z [ ]
+      Thu [ ] Jan [ ]{2} 1 [ ] 00:00:00 [ ] 1970  # git-send-email
+    | From [ ] {{git_hash_full}} [ ]
+      Mon [ ] Sep [ ]   17 [ ] 00:00:00 [ ] 2001  # git-format-patch
+    )
```
deathaxe, line 187: How magic is the date? Are these really only those two days?

michaelblyons: The `git-format-patch` one is listed in the documentation. The `git-send-email` one I saw live in a different document online (context at https://github.com/dandavison/delta/issues/1863), but does not appear in docs.

It wasn't until today that I saw a random date that I think matches the commit date.

deathaxe: As docs explicitly state about those fixed dates as "magic number" to separate patches from normal mails, we should probably keep going with them as is.

michaelblyons: Sounds reasonable to me. The only counter argument I have in favor of putting the variable in Diff (Basic) is that you're already in the Diff syntax. I wouldn't want to loosen the `first_line_match`.

michaelblyons: > So whatever headers should be supported, they seem useful only in Diff (Basic).

Maybe you forget that Diff (Basic) does that thing where if you actually have a Git Diff, it sends to Git Diff syntax instead. I don't have a problem making the `git_first_line` variable more lenient in Diff (Basic).

I am a little wary of changing the `first_line_match` in case you (a) have no MBOX syntax, and (b) load an MBOX file. At this point, with a lenient `first_line_match`, one of the diff syntaxes would load and try to highlight as a Git Diff. This may not actually be a bad thing since it marks email headers, but could conceivably freak out for multi-email MBOX files.

But if Sublime Text has chosen Diff or Git Diff by file extension, I think it would be polite to load up the Git Diff variant if the first line loosely matches.

deathaxe (approved):

jfcherng (approved):

michaelblyons:
Hooray! Thanks, everyone.

PR #4115, opened 2024-12-16
[Common] Add more items to .gitignore

deathaxe:
This commit...

1. excludes Default/ and User/ packages

   That's useful if whole repository is cloned to `${stdatadir}/Packages` to test/develop all bundled packages in a single ST instance.

   Make sure never to add a package with such special name to this repo.

2. excludes known python cache files as they are never to be published, if auto-created by any tool by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,10 @@
+# MacOS specific ignores
 .DS_Store
+
+# excluded packages
+/Default/
+/User/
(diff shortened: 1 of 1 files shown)


FichteFoll (approved):

michaelblyons (approved):
Good idea

PR #4120, opened 2024-12-26
[C#] support class definitions without a body

### LINKED ISSUE #4119: [C#] Primary constructors on raw classes without a body
### What happened?

Sublime seems to think there is a syntax error here because the class has no body.

```csharp
public class InvalidException(string _message) : Exception(_message); // errors on the semicolon
public class NotFoundException(string _message) : Exception(_message);
public class InternalServerErrorException(string _message) : Exception(_message);
```

Since C# 12, primary constructors on classes allows this body-less syntax.

keith-hall:
fixes #4119 

plus converts some anonymous contexts into named contexts

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -446,12 +446,7 @@ contexts:
     - match: \s*(;)
       captures:
         1: punctuation.terminator.statement.cs
-      pop: true
-
-  maybe_terminator:
-    - include: terminator
-    - match: (?=\S)
-      pop: true
+      pop: 1
 
   delegate_end:
     - include: terminator
@@ -477,6 +472,7 @@ contexts:
 
   class_signature:
     - meta_scope: meta.class.cs
+    - include: terminator
     - match: (?=\S)
       set: class_body
 
@@ -517,20 +513,23 @@ contexts:
   data_type_expect_block:
     - match: \{
       scope: punctuation.section.block.begin.cs
-      push:
-        - meta_scope: meta.block.cs
-        - match: \}
-          scope: punctuation.section.block.end.cs
-          pop: 2
-        - include: attribute
-        - include: class_declaration
-        - include: interface_declaration
-        - include: delegate_declaration
-        - include: method_declaration
-        - include: stray_close_bracket
+      push: data_type_inside_block
+    - include: terminator
     - match: \S*
       scope: invalid.illegal.cs
       pop: true
+
+  data_type_inside_block:
+    - meta_scope: meta.block.cs
+    - match: \}
+      scope: punctuation.section.block.end.cs
+      pop: 2
+    - include: attribute
+    - include: class_declaration
+    - include: interface_declaration
+    - include: delegate_declaration
+    - include: method_declaration
+    - include: stray_close_bracket
 
   data_type_parameter:
     - match: '<'
@@ -555,8 +554,7 @@ contexts:
     - match: '{{name}}'
       scope: entity.other.inherited-class.cs
       push:
-        - maybe_terminator
-        - maybe_constructor_parameters
+        - function_call_begin_paren
     - include: type
 
   type_constraint_common:
@@ -800,16 +798,18 @@ contexts:
     - include: attribute
     - include: lambdas
 
-    - match: (?=[^\s\[])
-      set:
-      - match: '\s*\b(out|ref|this|params|in)\s+'
-        captures:
-          1: storage.modifier.parameter.cs
-      - match: \s
-        pop: true
-      - include: type
+    - match: (?=[^\s\[;])
+      set: method_param_type_modifier
     - match: (?=\}|\)|>|\]|;)
       pop: true
+
+  method_param_type_modifier:
+    - match: '\s*\b(out|ref|this|params|in)\s+'
+      captures:
+        1: storage.modifier.parameter.cs
+    - match: \s
+      pop: true
+    - include: type
 
   constructor_prebody:
     - meta_scope: meta.method.constructor.cs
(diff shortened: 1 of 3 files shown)


C#/C#.sublime-syntax
```diff
@@ -446,12 +446,7 @@ contexts:
     - match: \s*(;)
       captures:
         1: punctuation.terminator.statement.cs
-      pop: true
-
-  maybe_terminator:
-    - include: terminator
-    - match: (?=\S)
-      pop: true
+      pop: 1
```
michaelblyons, line 449: Not important, but you use `pop: true` for the rest of the commit (except the `pop: 2`).

C#/tests/syntax_test_C#12.cs
```diff
@@ -0,0 +1,40 @@
...
+///          ^^^^^^^^^^^^^^^^ entity.name.class.cs
+///                          ^^^^^^^^^^^^^^^^^ meta.class.constructor.parameters.cs
+///                          ^ punctuation.section.parameters.begin.cs
+///                           ^^^^^^ storage.type.cs
+///                                  ^^^^^^^^ variable.parameter.cs
+///                                          ^ punctuation.section.parameters.end.cs
+///                                           ^^^^^^^^^^^^^^^^^^^^^^^ meta.class.cs
+///                                            ^ punctuation.separator.type.cs
+///                                              ^^^^^^^^^ entity.other.inherited-class.cs
+///                                                       ^^^^^^^^^^ meta.function-call.cs meta.group.cs
```
deathaxe, line 16: `meta.function-call` feels off somehow in combination with inherited-class.

keith-hall: But logically we want it scoped like a normal function call I believe, a constructor is just a special type of function 🙂 I think other constructor calls are like that, though I'm not at a computer to check right now. Like `new SomeClass("this is a constructor call")`

deathaxe: I read the line as a declaration, which forwards arguments from specialized `InvalidException` class to `Exception` base class in case of an instantiation. The line doesn't actually call any constructors, does it?

The actual instantiation, which calls a constructor as part of object creation would be `new InvalidException("Error message")`, but that's not the topic here.

Even if, I wouldn't scope an object instantiation "meta.function-call" either, as calling one of the available constructors is only one part of the whole object creation process.

keith-hall: What scope can we use here then? :)

btw in the below `InvalidException` and `InvalidException2` are basically identical, just using different syntax.
```cs
public class InvalidException(string _message) : Exception(_message);

public class InvalidException2 : Exception
{
    public InvalidException2(string _message) : base(_message)
    {
    }
}
```
and if we had a breakpoint inside the constructor for `Exception`, it would get hit when calling `new InvalidException("Error message")`, which is why I felt it can be just treated as another function call...

deathaxe: Sure, those constructors are called on `new InvalidException()`. The actual function-call is invoked implicitly when instantiating via `new` operator.

Both, long and short form, however just denote a class definition. See how also `base(_message)` is not scoped `meta.function-call` in current C# syntax, because it does not belong to constructor's normal function body but just denote's which parameters to pass to super class's constructor.

I find this special syntax worth dedicated scopes, especially as they describe a sort of inheritance or dependency tree.

We can find such constructs in Python dataclasses as well as Java's `record class` definitions:

https://github.com/sublimehq/Packages/blob/c7cd5a7dc3eb34b5746b2f87c5ca4967574c9e40/Java/tests/syntax_test_java.java#L2527-L2536

... except none of them supports direct parameter forwarding to parent classes in declaration syntax.

IIRC, Java is the only syntax which applies special `meta.class.[extends|implements]` meta scopes to the inherited `: Exception` part without however needing to address any argument list.

We could go with it and just scope those argument lists `meta.group`....

```cs
public class InvalidException(string _message) : Exception(_message);
//           ^^^^^^^^^^^^^^^^ meta.class.identifier entity.name.class
//                           ^^^^^^^^^^^^^^^^ meta.class.parameters meta.group
//                                           ^^^ meta.class
//                                              ^^^^^^^^^^^^^^^^^^^^ meta.class.[extends|inherites]
//                                               ^^^^^^^^^ entity.other.inherited-class
//                                                        ^^^^^^^^^^ meta.group
```

or add a new `meta.class.extends.[identifier|arguments]` sub-scopes, covering whole `Exception(_message)`, like ...

```cs
public class InvalidException(string _message) : Exception(_message);
//           ^^^^^^^^^^^^^^^^ meta.class.identifier entity.name.class
//                           ^^^^^^^^^^^^^^^^ meta.class.parameters meta.group
//                                           ^^^ meta.class
//                                              ^ meta.class.extends
//                                               ^^^^^^^^^ meta.class.extends.identifier entity.other.inherited-class
//                                                        ^^^^^^^^^^ meta.class.extends.arguments meta.group
```

or just stack another `meta.class.[identifier|arguments]` like...

```cs
public class InvalidException(string _message) : Exception(_message);
//           ^^^^^^^^^^^^^^^^ meta.class.identifier entity.name.class
//                           ^^^^^^^^^^^^^^^^ meta.class.parameters meta.group
//                                           ^^^ meta.class
//                                              ^^^^^^^^^^^^^^^^^^^^ meta.class.extends
//                                               ^^^^^^^^^ meta.class.identifier entity.other.inherited-class
//                                                        ^^^^^^^^^^ meta.class.arguments meta.group
```

_Note: Not sure if it is correct, but I somehow got used to use `arguments` if parens describe arguments being passed to and `parameters` if parens denote a real signature definition._

deathaxe: FWIW, Java scopes instantiations `meta.instantiaton`

https://github.com/sublimehq/Packages/blob/c7cd5a7dc3eb34b5746b2f87c5ca4967574c9e40/Java/tests/syntax_test_java.java#L7973-L7978

It distinguishes it from normal function calls to illustrate it being a special syntax to create a new object and implicitly calling its construtors to initialize attributes/members.

I could also imagine to add sub scopes for identifiers and argument lists like `meta.instantiation.[identifier|arguments]`.

michaelblyons (dismissed):
I'm fine for the conversations to play out, but this is an improvement.

deathaxe:
It however changes inherit class parameter list's scope to `meta.function-call.arguments` without even adding meta.function-call scope to the inherit class name, which is inconsistent and IMHO not correct.

This syntax denotes inheritance definitions, but no function calls.

michaelblyons (dismissed):

C#/C#.sublime-syntax
deathaxe: Let's do it the other way round in another PR?

deathaxe (approved):

michaelblyons (approved):

Issue #3883, opened 2023-12-06
[C++] Macro with arguments right after class keyword breaks all subsequent file


ratijas:
### What happened?

Qt has class annotation macros with string arguments. And it breaks everything else that follows it.

Here, `for` becomes a keyword, and the rest is a string:
```c++
class QT_DEPRECATED_VERSION_X_6_6("this for that")
QPropertyAlias : public QPropertyObserver
```

Here it kinda just works™:

```c++
class 
QT_DEPRECATED_VERSION_X_6_6("this for that")
QPropertyAlias : public QPropertyObserver
```

![image](https://github.com/sublimehq/Packages/assets/6737986/13dbbedf-beac-4d1c-bf51-4167ec0f15a5)

PR #4122, opened 2024-12-29
[C++] Fix issue 3883

simartin:
Codebases such as Qt or GCC use macros to annotate some class/struct/union.

This patch teaches the C++ syntax highlighter about such constructs, and fixes https://github.com/sublimehq/Packages/issues/3883.


deathaxe (approved):
LGTM!

michaelblyons (approved):

PR #4123, opened 2024-12-30
[LaTeX] Add support for IEEEarray environments

deathaxe:
This commit

1. adds support for `\begin{IEEEeqnarray}` math environments. 
   specified by: https://moser-isi.ethz.ch/docs/typeset_equations.pdf 
   inspired by: https://github.com/SublimeText/LaTeXTools/issues/1413

2. fixes `meta.environment.math.block` scope by removing incomplete `.be`.
3. fixes `\\end` scope using `keyword.control.flow.end.latex`.

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -639,7 +639,6 @@ contexts:
         ((\\)begin)
         \s* (\{) \s* ( IEEEeqnarray\*? ) \s* (\})
         \s* (\{) \s* ( {{letter}}+ ) \s* (\})
-      # scope: meta.environment.math.block.latex
       captures:
         1: support.function.begin.latex keyword.control.flow.begin.latex
         2: punctuation.definition.backslash.latex
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -634,16 +634,32 @@ contexts:
         4: variable.parameter.function.latex
         5: punctuation.definition.group.brace.end.tex
       push: block-math-begin-end-command-body
+    - match: |-
+        (?x)
+        ((\\)begin)
+        \s* (\{) \s* ( IEEEeqnarray\*? ) \s* (\})
+        \s* (\{) \s* ( {{letter}}+ ) \s* (\})
+      # scope: meta.environment.math.block.latex
+      captures:
+        1: support.function.begin.latex keyword.control.flow.begin.latex
+        2: punctuation.definition.backslash.latex
+        3: punctuation.definition.group.brace.begin.tex
+        4: variable.parameter.function.latex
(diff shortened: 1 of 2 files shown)


michaelblyons (dismissed):

LaTeX/LaTeX.sublime-syntax
```diff
@@ -634,16 +634,32 @@ contexts:
         4: variable.parameter.function.latex
         5: punctuation.definition.group.brace.end.tex
       push: block-math-begin-end-command-body
+    - match: |-
+        (?x)
+        ((\\)begin)
+        \s* (\{) \s* ( IEEEeqnarray\*? ) \s* (\})
+        \s* (\{) \s* ( {{letter}}+ ) \s* (\})
+      # scope: meta.environment.math.block.latex
```
FichteFoll, line 642: We should avoid comments like this to not cause confusion when looking at them later. Seeing how the meta scope is pushed as a `meta_content_scope` in `block-math-begin-end-command-body`, I suppose we don't need this on the match here. If we wanted it to be included, changing that to `meta_scope` should suffice.

LaTeX/LaTeX.sublime-syntax
```diff
@@ -634,16 +634,32 @@ contexts:
         4: variable.parameter.function.latex
         5: punctuation.definition.group.brace.end.tex
       push: block-math-begin-end-command-body
+    - match: |-
+        (?x)
+        ((\\)begin)
+        \s* (\{) \s* ( IEEEeqnarray\*? ) \s* (\})
+        \s* (\{) \s* ( {{letter}}+ ) \s* (\})
```
FichteFoll, line 641: What is the motivation behind this change? Afaics this marks any begin block not matched otherwise as a math block and that seems a bit over-eager, does it not?

deathaxe: Motivation is support for `\begin{IEEEeqnarray}{c}`

1. `\s* (\{) \s* ( IEEEeqnarray\*? ) \s* (\})` matches `{IEEEeqnarray}`
2. `\s* (\{) \s* ( {{letter}}+ ) \s* (\})` matches `{c}`

Actually, 2. should probably be optional.

deathaxe: Btw.: we need to consume that as otherwise LaTeXTools includes it into math equation preview.

deathaxe: >  Afaics this marks any begin block not matched otherwise as a math block

Nope.

FichteFoll: Ah, I see. I somehow inserted a `|` between these two in my head and did not consider it to be the second parameter.

FichteFoll (approved):

jfcherng (approved):

PR #4124, opened 2024-12-30
[Scala] Add support for Scala 3

### LINKED ISSUE #2795: [Scala] Add support for Scala 3
Scala 3 is currently in its final release candidate, and a release is expected within about a month. While the existing syntax mode does work just fine for the subset of Scala 3 which overlaps with Scala 2, there are a large number of constructs which have been added to the language which are not fully supported. A non-exhaustive list follows:

- `given`, `then`, `as`, and similar new keywords (most of which are "soft" keywords that only parse as such in the proper context)
- Significant indentation (similar to Python; has a major impact on auto-indentation hints)
- New class definition syntax (related to the above)
- Mix-fix method definition syntax
- Higher-rank types (e.g. `[A] =>> (A, A => Int) => Int`)
- Context-bounds included within types (e.g. `Monoid[A] ?=> (A, A) => A`)
- New meaning for wildcards (e.g. `*` in imports, `?` in types, etc)
- New macro syntaxes (related to the new keywords problem)

The list goes on. There's a lot of stuff to update here. I plan to get to this eventually, but my calendar is a bit crunchy right now, so I'm putting this issue out here so that others can be aware of the need.

### LINKED ISSUE #3778: [Scala] "with" keywords highlighted as out of place when comment between them and extends

djspiewak:
Sorry for the massive PR. Scala 3 is a somewhat distressingly large change to the language syntax, as you can see from the kind of invasive surgery I had to perform. Open to bikeshedding on some of the scoping; the constructs themselves are a bit weird (e.g. `given` and `extension`), with existing Scala and Haskell probably being the best benchmark. There are a few things I didn't dig into, like for example type matching doesn't use `variable` scoping where it should (e.g. `X match { case Foo[a] => a }`, but that's consistent with value-level matching which *also* doesn't use `variable` scoping where it should (e.g. `x match { case foo: Foo[a] => foo: a }`).

One final caveat is I tried to obey the guidance around soft vs hard keywords (`extension` is a pretty notable example of this), but I just gave up on `inline`. It can pop up in way too many places so I just made it a hard keyword, even though that's technically over-restrictive.

As an aside, I really *wanted* to go through and clean up this mode, taking advantage of ambiguous parsing to resolve some of the long-standing annoyances in types and newline inference. This is still probably a good thing to do, but it would be a much larger undertaking and Scala 3 support has been waiting for years at this point, so I figured I'd just cake on the tech debt and we can do a proper mode refactor at a later date.

Anyway, each commit is pretty self-contained, so if you're trying to review things step by step, that's the way to go.

Fixes #2795
Fixes #3778
Closes #3779


djspiewak:
Oh one notable bit here…

There's some very old syntax which Scala 3 (and subsequently Scala 2.12 and 2.13) made entirely invalid, which is the following:

```scala
x: Int => 42
```

This type of bare lambda *with an explicit type* and no parentheses was permitted way back in Scala 2.7, and was even relatively idiomatic, but it's now entirely illegal due to the way that braceless syntax interacts with lambdas. Instead, if you want to write this type of thing, you need to wrap the whole thing in curlies or the formals in parens:

```scala
{ x: Int => 42 }
(x: Int) => 42
```

We have the same problem as scalac in parsing this. In particular, it is ambiguous w.r.t. the following syntax:

```scala
foo: a => b
```

In Scala 3, with braceless syntax, the above can be a valid way of writing the following:

```scala
foo { a => b }
```

This is why the bare type ascription form was deprecated. So anyone using Sublime with *really* old versions of Scala (more than a decade at this point) will suddenly see these forms no longer scoping as expected.

djspiewak:
Definitely an oversight. I didn't think about the other issue until after I opened the PR. I'll add a test which shows it's fixed and then add the bit to the description.

Scala/Scala.sublime-syntax
```diff
@@ -157,21 +161,27 @@ contexts:
       scope: constant.character.literal.scala
 
   comments:
-    - match: (//).*$
+    - match: '(//).*\n?$'
```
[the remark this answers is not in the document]
deathaxe: Why quoted?

Scala/Scala.sublime-syntax
```diff
@@ -229,23 +239,43 @@ contexts:
...
+          pop: true
         - match: '{{rightarrow}}'
           scope: keyword.declaration.function.arrow.lambda.scala
-          pop: true
+          set:
+            - meta_scope: meta.block.scala
+            - match: '\}'
+              scope: punctuation.section.block.end.scala
+              pop: true
+            - include: main
```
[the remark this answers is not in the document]
djspiewak: Sadly yes. There is nothing that you can put at the root of a file that you *cannot* put immediately after the opening preamble of a lambda other than package declarations, but even then there are some special cases where it's possible. Scala is a very compositional language in that way.

Scala/Scala.sublime-syntax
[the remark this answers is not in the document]
djspiewak: It's interesting that both orders work. Swapped for yours.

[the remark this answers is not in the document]
djspiewak: That's strange though because it did. I think because the newline was optional.

Scala/Scala.sublime-syntax
```diff
@@ -700,7 +874,61 @@ contexts:
     - match: '(?=\S)'
       set: class-parameter-list
 
+  class-parameter-list-maybe-case:
+    - match: '{{rightarrow}}'
+      fail: case
+    - match: '\('
+      scope: punctuation.section.group.begin.scala
+      push:
+        - match: '\)'
```
deathaxe, line 883: How about adding meta scopes? Those would be required to reliably support syntax based folding.

```suggestion
        - meta_scope: meta.group.scala
        - match: '\)'
```

Scala/Scala.sublime-syntax
```diff
@@ -610,6 +708,43 @@ contexts:
     - match: '(?=\S)'
       pop: true
 
+  function-parameter-list-contents:
```
deathaxe, line 711: How about adding meta scopes? Those would be required to reliably support syntax based folding.

```suggestion
  function-parameter-list-contents:
    - meta_scope: meta.group.scala
```

Maybe also `meta.function.parameters` if it is feasable to scope function definitions.

Scala/Scala.sublime-syntax
```diff
@@ -610,6 +708,43 @@ contexts:
...
+            - match: '(\*)\s*(?=[\),=])'
+              captures:
+                1: keyword.operator.varargs.scala
+            - match: '(?=\))'
+              pop: true
+            - match: ','
+              scope: punctuation.separator.scala
+              pop: true
+            - include: delimited-type-expression
+            - match: '(?=[=])'
```
deathaxe, line 734: Character class shouldn't be required to match single `=` only.

```suggestion
            - match: '(?==)'
```

Scala/Scala.sublime-syntax
```diff
@@ -969,29 +1244,50 @@ contexts:
...
-      captures:
-        1: keyword.control.flow.scala
-        2: punctuation.section.block.begin.scala
+  if-statement:
+    - match: '\bif\b'
+      scope: keyword.control.flow.scala
       push:
-        - match: '\}'
-          scope: punctuation.section.block.end.scala
+        - match: '(?=\n)'
```
deathaxe, line 1251: Same as...

```suggestion
        - match: '$'
```

Scala/Scala.sublime-syntax
```diff
@@ -1090,11 +1387,15 @@ contexts:
...
+    # we treat inline as a hard modifier because it's too convoluted to treat it softly
+    - match: '\b(?:abstract|final|lazy|sealed|implicit|override|using|inline)\b'
+      scope: storage.modifier.other.scala
+    #  soft modifiers
+    - match: '\b(?:infix|opaque|open|transparent)\b(?=\s+(?:private|protected|abstract|final|lazy|sealed|implicit|override|def|val|var|type|given|class|trait|object|enum|case\s+class|case\s+object|inline)\b)'
       scope: storage.modifier.other.scala
 
   # see http://www.scala-lang.org/docu/files/ScalaReference.pdf part 1.3.5-6 (page 18)
-  strings:
+  base-strings:
```
deathaxe, line 1398: I'd suggest to scope strings `meta.string.scala string....` so `meta.string` keeps set within interpolation, which clear string scope. That's common practice accross other default syntaxes. 

At the same time scoping interpolation `${...}` `meta.interpolation` would help aligning with other syntaxes.

Scala/Scala.sublime-syntax
```diff
@@ -1707,28 +2059,39 @@ contexts:
...
     - match: '{{typeid}}'
       scope: support.type.scala
       set: single-type-expression-tail
     - match: (?=@{{plainid}})
       set: single-type-expression-tail
-    - match: \(
+    - match: '\['
+      scope: punctuation.definition.generic.begin.scala
+      set:
+        - match: '\]'
```
deathaxe, line 2072: Meta scope?
```suggestion
        - meta_scope: meta.generic.scala
        - match: '\]'
```

Scala/Scala.sublime-syntax
```diff
@@ -1801,6 +2160,63 @@ contexts:
...
+    # we need to explicitly match this one because "|" is a valid type name
+    - match: '(?=\|)'
+      pop: true
+    - match: '{{typeid}}'
+      scope: support.type.scala
+      set: single-type-expression-tail-no-function-no-or-type-expectation
+    - match: '\{'
+      scope: punctuation.definition.block.begin.scala
+      set:
+        - match: \}
```
deathaxe, line 2208: Meta scope?

```suggestion
        - meta_scope: meta.block.scala
        - match: \}
```

deathaxe (commented):
Can't say much about Scala specific stuff itself, but left some general suggestions.

Scala/Scala.sublime-syntax
djspiewak: This is going to be a long battle lol. There's a LOT of these types of groups.

Scala/Scala.sublime-syntax
djspiewak: This is old code that got moved around, but I *think* I did it because `(?==)` is really hard to read when you have a ligature font (since `==` renders like a long `=`).

Scala/Scala.sublime-syntax
djspiewak: That's a consuming match though, as opposed to a lookahead.

Scala/Scala.sublime-syntax
djspiewak: Iirc, the way we scope strings (particularly with interpolation) was still very much in flux when I wrote this stuff and I haven't gone back and updated it.

Scala/Scala.sublime-syntax
djspiewak: Function definitions are going to be really tricky to scope. Same with class definitions. I'd really *like* to though.

Scala/Scala.sublime-syntax
deathaxe: Both are non-consuming.

Scala/Scala.sublime-syntax
michaelblyons: Do you have PackageDev installed for highlighting? I think the different scopes for `?=` and `=` should prevent collapse into a ligature.

djspiewak: I think I removed it at some point because it was annoying me relative to the base yaml mode. If there's a strong preference for removing the spurious character class, I don't mind merging the suggestion.

deathaxe: There's not. It is not of any functional relevance.

Scala/Scala.sublime-syntax
deathaxe: We can address it in a dedicated follow-up PR as well, as it is not directly related with adding support for Scala 3.

deathaxe (approved):
LGTM!

deathaxe:
Also adding more `meta.group` etc. can be done separately.

PR #4126, opened 2024-12-31
[JavaScript] Fix anonymous class instantiation

### LINKED ISSUE #4125: [TypeScript] `class implements Interface` is broken
### What happened?

```ts
interface B {}
const A = class implements B {
  readonly foo = 1
}
```

Screenshot: 
image

It seems the word `implements` was identified as a class name, because JavaScript.sublime-syntax doesn't exclude it from `reserved_word`.

So the workaround is to give the scoped class a name:

```ts
const A = class A implements B {
  readonly foo = 1
}
```

FYI, my initial code was like this:

```ts
export foo: IFoo = new class implements IFoo {}
```

deathaxe:
Fixes #4125

This commit adds "future reserved words" from [developer.mozilla.org][1] to prevent them being scoped as normal identifiers which fixes anonymous class instantiations in TypeScript.

They are also marked reserved in JavaScript as specified as such by [developer.mozilla.org][1], and `implements` is scoped illegal in class declarations.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -45,11 +45,56 @@ variables:
   # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar
   reserved_word: |-
     (?x:
-      break|case|catch|class|const|continue|debugger|default|delete|do|else|
-      export|extends|finally|for|function|if|import|in|instanceof|new|return|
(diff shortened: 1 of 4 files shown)


michaelblyons (approved):

keith-hall (approved):

deathaxe:
Thanks!

PR #4128, opened 2025-01-04
README: Add instructions on how to create an empty zip file

FichteFoll:
Since I just had to do this and it's not as straight forward as you'd expect, I added instructions for this to the README, seeing how this repo is the only one where such a hack would be relevant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,8 +70,8 @@ After you've finished, keep in mind that you're now overriding a default package
   # or
   $ touch dummy && zip empty.zip dummy && zip -d empty.zip dummy && rm dummy
   ```
-  To make your empty packages stand out in directory listings,
-  you can create them using symlinks to this empty zip file,
+  Optionally, to make your empty packages stand out in directory listings,
+  you can create them using symlinks to this empty zip file instead of copying it,
   e.g. `ln -s empty.zip SQL.sublime-package`.
 
 ## Reference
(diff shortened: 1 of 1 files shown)


deathaxe:
Wouldn't it be easier to just provide a utility command which creates an empty override of given default packages? Sound's like something being in scope of Package Control, OverrideAudit or PackageDev.

FichteFoll:
Would it be easier? Maybe, but since we currently have instructions on how to symlink individual packages of this repo into your ST Packages dir, it also makes sense to have the empty zip file trick in here.

keith-hall (dismissed):

deathaxe:
1. Using the echo command prints plain text to the file on both cygwin bash and WSL bash causing the file not being a valid zip.

2. I know the reason, but still not a fan of this broken line style, but if we want to change it, whole file should follow same style. I wouldn't introduce mixed writing/line styles.

FichteFoll:
> The following works however

Thanks for checking. I'll adopt that since it's more portable.

> Why not creating empty packages directly? They are small enough to not justify additional complexity by creating symlinks.

I mentioned a reason in the text: to make them stand out. Of course, this is an optional thing, so I will emphasise this fact more, but at least for me on Linux it feels more natural (and has symmetry with the Packages folder).

> whole file should follow same style. I wouldn't introduce mixed writing/line styles.

Since I'm an advocator of this style, I'll gladly make the change if we can largely agree on it. In a separate PR, however.

michaelblyons (approved):

deathaxe (approved):

PR #4132, opened 2025-01-07
[Scala] Tweak string and interpolation scopes

deathaxe:
This commit tweaks string and interpolation related scopes after the scheme of Python and ShellScript.

1. from python use `meta.string.interpolated` and `string.quoted.double.block`
2. from shell use interpolation scopes.

   Scope guidelines request `punctuation.section.interpolation.[begin|end]`, but `$` variables are historically scoped `punctuation.definition.variable`.

   This commit (as shellscript) attempt to satisfy both.

   Hence `$` in `${}` is scoped punctuation.definition.variable, which may appear odd in the first place.

   Actually `$` is some sort of expansion operator, which adds a variable's or expression's result into containing string. Perl calls it dereferencing operator.

   Most core syntaxes supporting that kind of syntax struggle with it, currently.

   _TL;TR This commit aligns scopes with existing core syntaxes even though there are still some caveats. Solving them is however a bigger task affecting various syntaxes. Good consistent alternatives are required first._

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -1402,7 +1402,7 @@ contexts:
       scope: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: meta.string.scala string.quoted.double.block.scala
+        - meta_scope: meta.string.scala string.quoted.triple.scala
         - match: '(""")(?!")'
           scope: punctuation.definition.string.end.scala
           pop: true
@@ -1428,7 +1428,7 @@ contexts:
         2: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: meta.string.interpolated.scala string.quoted.double.block.scala
+        - meta_scope: meta.string.interpolated.scala string.quoted.triple.scala
         - match: '"""'
           scope: punctuation.definition.string.end.scala
           pop: true
@@ -1439,7 +1439,7 @@ contexts:
         2: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: meta.string.interpolated.scala string.quoted.double.block.scala
+        - meta_scope: meta.string.interpolated.scala string.quoted.triple.scala
         - match: '{{unicode_char}}'
           scope: constant.character.escape.scala
         - match: '(""")(?!")'
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -1402,17 +1402,17 @@ contexts:
       scope: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: string.quoted.triple.scala
-        - match: '{{unicode_char}}'
-          scope: constant.character.escape.scala
+        - meta_scope: meta.string.scala string.quoted.double.block.scala
         - match: '(""")(?!")'
           scope: punctuation.definition.string.end.scala
           pop: true
+        - match: '{{unicode_char}}'
+          scope: constant.character.escape.scala
     - match: '"'
       scope: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: string.quoted.double.scala
+        - meta_scope: meta.string.scala string.quoted.double.scala
         - match: '"'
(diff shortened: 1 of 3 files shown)


Scala/Scala.sublime-syntax
```diff
@@ -1402,17 +1402,17 @@ contexts:
       scope: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: string.quoted.triple.scala
-        - match: '{{unicode_char}}'
-          scope: constant.character.escape.scala
+        - meta_scope: meta.string.scala string.quoted.double.block.scala
```
djspiewak, line 1405: Triple strings are different from double (similar to Bash).

Scala/Scala.sublime-syntax
```diff
@@ -1402,17 +1402,17 @@ contexts:
...
         - match: '(""")(?!")'
           scope: punctuation.definition.string.end.scala
           pop: true
+        - match: '{{unicode_char}}'
+          scope: constant.character.escape.scala
```
djspiewak, line 1410: Triple strings do not support character escapes. So in other words, we should probably have a test something like this:

```scala
"""\n"""
   ^^ - constant

File: Scala/Scala.sublime-syntax [hunk 3]
```diff
@@ -1427,32 +1427,32 @@ contexts:
...
           pop: true
+        - include: f_string
     - match: '({{alphaid}})(""")'
       captures:
         1: support.function.scala
         2: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: string.quoted.triple.interpolated.scala
+        - meta_scope: meta.string.interpolated.scala string.quoted.double.block.scala
```
djspiewak, line 1442: Similar comment as the above.

djspiewak (changes requested):

michaelblyons (dismissed):
Nice

Scala/Scala.sublime-syntax
michaelblyons: The comparison here is to Python's `"""`.

Scala/Scala.sublime-syntax
deathaxe: see: line 370

Scala/Scala.sublime-syntax
deathaxe: Bash supporting this style of strings?

Took those scopes from Python, which supports both "triple-double" (`"""`) and triple-single (`'''`) quotes and thus wants to distinguish them.

Just see C# and Java also use `quoted.triple` as they only support `"""`. Thus we could probably follow them as well.

Scala/Scala.sublime-syntax
deathaxe: The distinction is `.block` subscope from Python.

deathaxe:
Reverted triple-quoted scope to `string.quoted.triple` as it is also used in C# and Java.

djspiewak (approved):

michaelblyons (approved):

Scala/Scala.sublime-syntax
```diff
@@ -1402,17 +1402,17 @@ contexts:
       scope: punctuation.definition.string.begin.scala
       push:
         - meta_include_prototype: false
-        - meta_scope: string.quoted.triple.scala
-        - match: '{{unicode_char}}'
-          scope: constant.character.escape.scala
+        - meta_scope: meta.string.scala string.quoted.triple.scala
```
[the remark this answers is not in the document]
djspiewak: They're not only multi line but they also allow for the use of special characters (notably quote and backslash) without escaping. So in a sense they're kind of like bash single quote strings.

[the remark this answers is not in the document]
michaelblyons: IMO TextMate did not pick a good convention here, but 🤷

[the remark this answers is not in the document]
michaelblyons: > What about `string.quoted.double.block`?

That's what this PR had originally. 😅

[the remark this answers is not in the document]
deathaxe: see: https://github.com/sublimehq/Packages/pull/4132#discussion_r1905940267

PR #4133, opened 2025-01-08
[Python] no stray parentheses in regexp patterns

deathaxe:
This commit scopes stray brackets normally as those may appear in partial patterns. Correct parentheses matching is not always possible and causes valid closing brackets to be scoped illegal.


michaelblyons:
This does work fine, but what do you think about prepending the "stray" bracket scopes instead of replacing the whole context? Otherwise you need to keep it up to date if anything upstream changes.

michaelblyons (approved):

PR #4134, opened 2025-01-09
[SQL] move heuristics for SQL inside Python and PHP strings to common syntax

keith-hall:
just a random idea I had when thinking about having SQL highlighting inside C# strings, so I thought I'd open a PR to get opinions...

### DIFF
--- a/PHP/PHP Source.sublime-syntax
+++ b/PHP/PHP Source.sublime-syntax
@@ -6,23 +6,14 @@ scope: source.php
 hidden: true
 version: 2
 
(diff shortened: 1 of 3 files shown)


deathaxe:
TBH, that doesn't feel right. What if there were 2 or 3 such constructs. We would barely inherit from them at the same time. Currently, if SQL is disabled/missing, only SQL strings keep plain. With this change whole syntax is driven nonworking. IMHO, that's not worth it.

keith-hall:
good point, I hadn't considered behavior while a package is disabled. Thanks, closing :)

PR #4137, opened 2025-01-13
[LaTeX] Add citation to symbol/reference index

deathaxe:
This commit enables Goto Definition and Goto Reference for citation.


michaelblyons:
Looks good. Maybe add 1-2 symbol test files, too? Even very small ones.

michaelblyons (approved):

deathaxe:
I only see tabs.

PR #4139, opened 2025-01-25
[Rust] Add `+ use<>` syntax

ehuss:
This adds support for the `+ use<>` precise capture syntax to Rust which was introduced in Rust 1.82.
[Documentation](https://doc.rust-lang.org/nightly/reference/trait-bounds.html)

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -564,7 +564,6 @@ contexts:
   types:
     # matches stray identifiers as a types because a type is expected
     - include: comments
-    - include: impl-types
     - include: return-type
     - match: '&'
       scope: keyword.operator.rust
@@ -590,6 +589,7 @@ contexts:
         - include: pattern-param
     - include: dyn-types
     - include: lifetime
+    - include: impl-trait
     - match: (?={{identifier}}::)
       push: type-path
     - match: (?=::)
@@ -710,15 +710,6 @@ contexts:
     - match: '\b(?:r#)?_*{{type_identifier}}'
       scope: storage.type.rust
 
-  impl-types:
-    - match: '\bimpl\b'
-      scope: storage.type.impl.rust
-      push:
-        - include: comments
-        - include: impl-generic
-        - match: '(?=\S)'
-          pop: true
-
   dyn-types:
     - match: '\bdyn\b(?=\s*(?:\(|{{lifetime}}|{{identifier}}))'
       scope: storage.type.trait.rust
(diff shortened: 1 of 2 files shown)


Rust/Rust.sublime-syntax
```diff
@@ -803,6 +794,31 @@ contexts:
...
+    - match: '\bimpl\b'
+      scope: storage.type.impl.rust
+      push:
+        - match: '\buse\b'
+          scope: keyword.other.rust
+          push:
+            - match: '<'
+              scope: punctuation.definition.generic.begin.rust
+              set:
+                - match: '>'
```
deathaxe, line 807: I'd suggest to add a meta scope to generics.

```suggestion
                - match: '>'
                - meta_scope: meta.generic.rust
```

Rust/Rust.sublime-syntax
```diff
@@ -803,6 +794,31 @@ contexts:
...
+                  scope: punctuation.definition.generic.end.rust
+                  pop: true
+                - include: lifetime
+                - match: ','
+                  scope: punctuation.separator.rust
+                - match: \b(Self)\b
+                  scope: keyword.other.rust
+                - match: '{{identifier}}'
```
deathaxe, line 815: `Self` is already defined as storage type in `no-path-type-names` context and identifiers should also receive valuable scope names. Types are currently unscoped as other identifiers are.

I'd therefore suggest to include ordinary `identifiers` context, which would also - even if maybe not required - handle fully qualified identifiers.

```suggestion
                - include: identifiers
```

FichteFoll: While it may not be 100% accurate either, I'd rather include `types` here since that is meant to be used in situations where types are expected. It matches lifetimes as well as stray identifiers and commonly known stdlib types. 

Commas still need to be matched separately, though.

FichteFoll: Actually, reading the spec again, pathed identifiers aren't supported and it is also unlikely for built-in names to be used since this is supposed to refer to generics and a generic with a built-in name would shadow it.

In that case, please instead simply include `all-generic-type-names`.

Rust/tests/syntax_test_generics.rs
```diff
@@ -442,8 +442,96 @@ fn function<const N: u16>() {
...
+//                                     ^^ meta.group punctuation.section.group.end
+//                                        ^^^^^^^^ meta.function meta.function.parameters meta.function.return-type
+//                                        ^^ punctuation.separator
+//                                           ^ meta.function.parameters keyword.operator
+//                                            ^ meta.function.parameters punctuation.section.group.begin
+//                                             ^^ meta.function.parameters storage.type
+//                                               ^ meta.function.parameters punctuation.section.group.end
+
+
+fn impl_trait_with_plus() -> impl Iterator<Item = hir::GenericParam<'hir>> + Captures<'a> + Captures<'s> {}
```
deathaxe, line 524: What is `her::GenericParam` if not a fully qualified type name ?

Rust/Rust.sublime-syntax
```diff
@@ -803,6 +795,40 @@ contexts:
...
+
+  impl-trait-use-generic-body:
+    - meta_scope: meta.generic.rust
+    - match: '>'
+      scope: punctuation.definition.generic.end.rust
+      pop: true
+    - include: lifetime
+    - match: ','
+      scope: punctuation.separator.rust
+    - match: '\bSelf\b'
```
michaelblyons, line 826: I don't know Rust, so apologies if this is stupid: Is `Self` a keyword or is it like `self` or `cls` variables in Python?

deathaxe: lowercase `self` is scoped `variable.language` and is the equivalent of Python.

`Self` is the type of current object. It is scoped `keyword.other` in various other contexts as well.

It should probably receive a unique `storage.type.self` scope, but that's probably another PR.

michaelblyons: Oh, neat. Thanks for explaining.

michaelblyons (approved):

FichteFoll (approved):

Issue #2481, opened 2020-09-15
[C++] Concept syntax not highlighted correctly


friedkeenan:
The new C++20 feature of concepts is not supported by the syntax highlighter. With the syntax shown below the syntax highlighter becomes very confused when used with classes, and with functions it just doesn't highlight everything correctly. With other ways of using concepts (such as `template<std::integral T>`), the issue is less severe.

- Sublime Version: 3211 (also tested with package files as of 4b24b22994df74b2902ad2c2a58813d43f023423)
- OS Version: x86_64 Linux 5.4.64-1-MANJARO

``` cpp
template<typename T> requires std::integral<T>
class Test {
    public:
        T value;
};

template<typename T> requires std::integral<T>
void test(T value) {
    /* ... */
}
```

rwols:
There's also the `concept` keyword now. Unfortunately, adding some contexts to parse `concept Foo = ...` at top-level scope doesn't work, as some libraries use a macro for the concept keyword: https://github.com/boostorg/asio/blob/c77f3b603bf2cd180945184a664c6a05d3bebd3b/include/boost/asio/execution/executor.hpp#L143

So it's going to be fun parsing concepts.

rwols:
I can also imagine libraries having a macro for `requires` to support older C++ version, creating something like

```cpp
#if CPP_VERSION >= 20
#define REQUIRES(expr) requires expr
#else
#define REQUIRES(expr) // empty
#endif

template <typename T> REQUIRES(std::integral<T>)
void test(T value) {
    /* ... */
}

PR #4140, opened 2025-02-01
[C++] Concept Syntax Highlighting

camila314:
Adds long-needed concept/constraint support to C++

### DIFF
--- a/C++/C++.sublime-syntax
+++ b/C++/C++.sublime-syntax
@@ -778,7 +778,7 @@ contexts:
       scope: storage.modifier.c++
       push:
         - meta_scope: meta.constraint.c++
-        - match: (?=\{|{{identifier}}\s*{{identifier}}\(|{{before_tag}})
+        - match: (?=\{|{{identifier}}\s*{{identifier}}\(|{{before_tag}}|:)
           pop: true
         - include: constraint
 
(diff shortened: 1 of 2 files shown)


camila314:
Fixes #2481

C++/C++.sublime-syntax
```diff
@@ -1303,6 +1328,9 @@ contexts:
     - match: '\bunion\b'
       scope: keyword.declaration.union.type.c++
       set: data-structures-union-definition
+    - match: '\bconcept\b'
+      scope: keyword.declaration.concept.c++
```
deathaxe, line 1332: To align with the other data structure definitions, `meta.concept` should also cover the `concept` keyword.

```suggestion
      scope: meta.concept.c++ keyword.declaration.concept.c++
```

C++/C++.sublime-syntax
```diff
@@ -1268,6 +1292,7 @@ contexts:
     - match: '(?=\{)'
       set: function-definition-body
     - include: function-specifiers
+    - include: constraint-requires
```
deathaxe, line 1295: As `constraint-requires` sets a context on stack, the pattern on line 1292 will never match in function concepts, thus body is not scoped `meta.function meta.block`.

C++/C++.sublime-syntax
```diff
@@ -759,6 +762,26 @@ contexts:
               scope: keyword.declaration.c++
             - include: template-common
 
+  requires:
+    - match: \brequires\b
+      scope: keyword.operator.word.c++
```
deathaxe, line 767: What's the purpose of scoping requires `keyword.operator` hier, while using `storage.modifier` in `constraint-requires`?

camila314: they both mean different things, one is a keyword to check if a function is valid and the other is a way to begin a constraint expression outside of a concept.

C++/C++.sublime-syntax
camila314: #[b58a74c](https://github.com/sublimehq/Packages/pull/4140/commits/b58a74c2ba1dc5a6700728e939d64cb5dfe97152)

deathaxe (approved):
LGTM!

michaelblyons (approved):

PR #4143, opened 2025-02-06
[Lua] Handle multi-global statements

mbartlett21:
The syntax now handles multi-global statements correctly.

```lua
    a, b = c, d
--   ^ punctuation.separator.comma
--          ^ punctuation.separator.comma
```

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -211,7 +211,7 @@ contexts:
       scope: invalid.illegal.unexpected-end.lua
 
     - match: (?=\S)
-      push: expression
+      push: expression-list
 
   block-contents:
     - meta_scope: meta.block.lua
--- a/Lua/tests/syntax_test_lua.lua
+++ b/Lua/tests/syntax_test_lua.lua
@@ -729,6 +729,15 @@
 --                   ^ meta.number.integer.decimal constant.numeric.value
 --                    ^ punctuation.terminator.statement
 
+    a, b = c, d
+--  ^ variable.other
+--   ^ punctuation.separator.comma
+--     ^ variable.other
+--       ^ keyword.operator.assignment
+--         ^ variable.other
+--          ^ punctuation.separator.comma
+--            ^ variable.other
+
     local x <const>, y <  const  > = 1, 2;
 --  ^^^^^ storage.modifier.lua
 --        ^ variable.other.lua


Lua/Lua.sublime-syntax
```diff
@@ -205,13 +205,13 @@ contexts:
...
       scope: storage.modifier.lua
-      push: expression-list
+      push: expression-list-stmt
 
     - match: end{{identifier_break}}
       scope: invalid.illegal.unexpected-end.lua
 
     - match: (?=\S)
-      push: expression
+      push: expression-list-stmt
```
michaelblyons, line 215: Did you mean `expression-stmt` here? I don't see where it is used otherwise.

mbartlett21: I noticed that there was already an issue with assigning multiple global variables. That change was added along with the last test in my last commit.
```lua
    a, b = c, d
--   ^      ^ these commas didn't have a scope before
```

Lua/Lua.sublime-syntax
```diff
@@ -457,13 +471,19 @@ contexts:
...
       push: expression-begin
 
+    # assume that == is being typed
+    - match: =(?=\s*(?:{{reserved_word_statement}}|$))
+      scope: keyword.operator.comparison.lua
+      push: expression-begin
+    # only complain once it is clear that it isn't ==
+    - match: =(?!=)
+      scope: invalid.illegal.invalid-operator.lua
+      push: expression-begin
```
deathaxe, line 485: Be it intermediate or not, if a single `=` is illegal, it should be scoped this way. Trying to catch all such intermediate states would using such assumptions would clutter syntax definitions with lots of extra complexity. IMHO, it's not worth it. 

```suggestion
    - match: =
      scope: invalid.illegal.invalid-operator.lua
      push: expression-begin
```

Maybe even move it back to its original location to keep symbolic and alphanumeric operators grouped.

michaelblyons: I'm not sure I agree. It's so annoying to have a flash of invalid as you type.

deathaxe: Which is an argument to not use illegals at all and better keep unknown stuff unscoped. Whatever we try, we'll always run into situations such arbitrary lookaheads are insufficient. Syntaxes would consist of 80% such lookaheads to reliably highlight illegals but avoid them during typing.

michaelblyons: I don't think that's strictly true. Flagging unmatched closing brackets makes sense to me because typing code does not create those. But when the act of typing `==` obviously involves typing a single `=` first, I don't think it should be invalid.

I don't know if this is a good pattern, but I used [a no-scope match](https://github.com/michaelblyons/SublimeSyntax-ssh_config/blob/88f45e0bbd2d806b058e4827314ece57ca91ab7e/SSH%20Common.sublime-syntax#L191-L200) for a config syntax where I *know* unambiguously that you can only type `yes` or `no`. Then, [anything else is marked invalid](https://github.com/michaelblyons/SublimeSyntax-ssh_config/blob/88f45e0bbd2d806b058e4827314ece57ca91ab7e/SSH%20Config.sublime-syntax#L355-L357).

mbartlett21: So then the question would be: Should the second `=` in the following code be highlighted?

```lua
local a = 5 = 7
--          ^ Here
```

If not, then this PR is mostly useless, apart from changing line 214 to `expression-list` rather than `expression`.

deathaxe: The question is how much "linting" a syntax definition should make. Some of them tried too hard to do so. 

The origin of syntax highlighting is to provide certain colors to tokens of special meaning. A `=` is an assignment operator, no matter whether it is placed in correct position or not.

What we primarily need context switching for is to handle situations in which `=` may have different meanings. If it would be used for comparison in `if a = b` for instance, we'd apply a context-specific rule.

2nd `=` in `a = 5 = 7` is clearly a syntax error, but in general there are often many situations of incomplete code in which this is just an intermediate state due to ongoing modifications. Hence any illegal scope provides chances to highlighted such intermediate tokens illegal even though they arn't.

Patterns discussed here actually prove that. As it is impossible to distinguish an intermediate state from an illegal state under all circumstances or with sane effort. I've tried that myself and gave up mostly as all sorts of illegals cause poor editing experience.

If in doubt, keep it as is to avoid illegals flashing up, or accept single `=` in comparisons being scoped illegal. It is not a following keyword which decides whether it is illegal or not.

mbartlett21: I'd probably be inclined to not highlight them then. I'll change this back to having the single expression-list for all of it, and perhaps change the parsing for locals

jfcherng (approved):

michaelblyons (approved):
Thank you!

Issue #3984, opened 2024-05-07
[C][C++] Highlighting user-defined types (classes, structs, typedefs)


fantasai:
### What happened?

Pulling up https://forum.sublimetext.com/t/highlighting-user-defined-data-types-c-in-sublime/67277 into an issue, since I'm also running into this problem: user-defined types (typedefs, classes, structs, enum types) and variables are colored the same way, but should be distinc.

Types are just annotations, whereas the variable and function names are the core of what you're doing, so distinguishing them is somewhat important for C and C++.

I'm unsure if this is fixable, but it's certainly come up on various forums multiple times.

camila314:
LSP has semantic highlighting which is probably the closest you're gonna get. ST can't distinguish between the two with syntax highlighting alone, and trying to implement such a feature in the syntax schema would likely cause major performance issues for large files

midnightveil:
I've opened #4331 which in my eyes the main difference I can see between other editors (notably: Visual Studio Code will recognise `blah_blah_t` as a type, but `blah_blah_foo` it won't); we can pretty easily do the same. This is 95% of the way there, for me.

Issue #3960, opened 2024-04-21
[C/C++] remove switch case auto unindent


StreetwareGames:
### Expected behavior

every time you type `case`, it auto unindents and you end up with each next case being less indented than the previous. No other editor that I'm aware of does this.

### Actual behavior

Don't unindent when you type `case`

### Steps to reproduce

https://github.com/sublimehq/Packages/assets/144807965/998001b1-aa99-4c74-aa6d-c57dd809a1cf

deathaxe:
Indentation rules work on a per-line basis using regular expressions.
Only a single style of indentation can be supported at a time.

The intended style is:

```c
void main(int argc) {
	switch(argc) {
	case 1:
		// statements
		break;
	
	case 2:
		{
			// statements
		}
		break;

	default:
		// statements
		break
	}
}
```

> The way of using brackets directly after case seems rather uncommon. 

Anything else requires customization on users end.

It is not currently planned to change the defaults.

Issue #3920, opened 2024-02-08
Slightly Improve 'C' Syntax Scopes


ketpain:
### Problem description

I really love this program. I've looked around and it seems I could maybe improve scopes myself but I feel like it should be shipped with the program.

### Preferred solution

Would like to have control of self defined variables and such
<img width="478" alt="image" src="https://github.com/sublimehq/sublime_text/assets/42127943/082cd0e6-757d-4cd9-b0df-d4cd0bec09e2">

### Alternatives

No alternatives really

### Additional Information

Thank you for the amazing and lightweight program.

michaelblyons:
Maybe try out #1831?

ketpain:
What I am mentioning would involve modifying "Sublime Text\Packages\Objective-C.sublime-package"

I wish it were as easy as just changing a few lines of scopes

Issue #3715, opened 2023-04-28
[C++] function modifier like [[nodiscard]] and [[noreturn]] break highlighting


Sainan:
### What happened?

```C++
// Good
const void* f() final;
// Bad
[[nodiscard]] const void* f() final;
[[noreturn]] const void* f() final;
```
![image](https://user-images.githubusercontent.com/63328889/235233873-1e532e46-e8fa-4200-9334-c05d50bf4e79.png)

Sainan:
Might be related to #1408 although that's about "find symbol", but I suspect the root cause is the same.

deathaxe:
C family hasn't been updated for a while, thus doesn't support most of the more modern language features right now.

Sainan:
It will be fixed when someone who's a) anal about syntax highlighting, b) cares enough to learn Sublime's weird tmLanguage replacements, c) and cares about C/C++ decides they want to submit some PRs. 😀

Sainan:
To put it more generally: Someone who cares about Sublime Text being able to accurately classify code.

PR #4147, opened 2025-02-07
[C Family] Full rewrite of C family syntaxes

### LINKED ISSUE #1070: [

### LINKED ISSUE #1093: [C/C++] The scope of storage type inside the function parameters incorrectly
For example:
``` cpp
void test(int);
```
The scope of int should be:
```
source.c++ meta.class.c++ meta.block.c++ meta.method.parameters.c++ meta.group.c++ storage.type.c
```
but it is actually:
```
source.c++ meta.class.c++ meta.block.c++ meta.method.parameters.c++ meta.group.c++ variable.parameter.c++
```
but if I add the variable name after the storage type, the scope work correctly:
```cpp
void test(int n);
```

### LINKED ISSUE #1138: [Objective-C] Incorrect syntax highlighting if newline after message target
https://forum.sublimetext.com/t/objective-c-syntax-highlighting-bug/31311

```
#import <Foundation/Foundation.h>

void foo() {
    NSString * foo;
    // Correct syntax highlighting
    [NSString stringWithString:@"foo"];
    [NSString stringWithCString: "foo" encoding: NSASCIIStringEncoding];
    [foo characterAtIndex:0];

    // Incorrect syntax highlighting if newline immediately after the
    // message target. 'stringWithString', 'stringWithCString', 'encoding',
    // and 'characterAtIndex' should all be blue.
    [NSString
        stringWithString:@"foo"];
    [NSString
        stringWithCString: "foo" encoding: NSASCIIStringEncoding];
    [foo
        characterAtIndex:0];
}
```

image

### LINKED ISSUE #1438: [Objective-C] Multi-parameter method names do not appear correctly in symbol list
For the following code snippet:
```
#import <Foundation/Foundation.h>
@interface Foo
-(void)methodWithArg:(BOOL)arg1 arg2:(BOOL)arg2;
@end
```
I would expect to see the method `methodWithArg:arg2:` as one symbol on the symbol list. Instead, this appears as two different symbols: `methodWithArg:` and `arg2:`.

This is extremely common and problematic for ObjC users.

image

### LINKED ISSUE #1830: [C] struct field is seen as a declaration inside a struct

### LINKED ISSUE #2047: [C] Bug with “register” keyword
I just spent hours banging my head off the table because of this. I forgot that the word `register` is a keyword when writing C code. Clang and gcc both kept throwing totally random errors when I was writing the function `void print_single_register(char *register, pid_t child_pid);`. I finally brought the function into Xcode which showed register in the keyword color. It turns out that Sublime will not recognize it as a keyword if it comes after the type definition, only if it comes before the type definition like it should (`register int i = 0;`). This is most likely a bug in the definition of the C syntax highlighting built into Sublime. The “C Improved” syntax definition I have installed through Package Manager highlights it correctly. First screenshot is the built in C syntax, second is “C Improved"

Screen Shot 2019-08-14 at 9 20 42 AM
Screen Shot 2019-08-14 at 9 24 28 AM

### LINKED ISSUE #2221: [C] Syntax highlighting around C namespaces is broken
When return types are on a separate lines:
* argument their types become a different colour
* argument list in the first function is broken but not in the third
* there is a red mark on the closing brace
Screenshot 2019-12-02 at 07 46 01

Without newline it looks OK:
Screenshot 2019-12-02 at 07 38 31

Inside a struct it's busted again but only for pointers:
Screenshot 2019-12-02 at 07 53 13

These are some of the examples but this is common and all over the place. `u64` is coloured because I added it to `custom_types` in `C.sublime-syntax:26` but otherwise the syntax file is standard. Sublime v3.2.2 build 3211

### LINKED ISSUE #2334: [C] Macro struct converter
- Sublime Version: Build 3211
- OS Version: Macos 10.15.3

``` 'C'
STRUCTPACKER(typedef struct unpacked_t {bool hello;} ) packed_t;
```

Screen Shot 2020-04-16 at 1 05 30 PM

### LINKED ISSUE #2762: [C] Incorrect error indication when a function return type is on the line before the function name and struct/enum types are used
- Sublime Version: 3211
- OS Version: Ubuntu 20.04

``` C
enum foo {
	SOME_VAL
};

struct bar {
};

enum foo
baz(struct bar *a)
{
	return SOME_VAL;
}
```

Screenshot from 2021-03-29 14-57-41

This code compiles without error with GCC:

```
gcc -c sublime_text_test.c
```

### LINKED ISSUE #2815: [C++]could not parse the function-call without left bracket in the same line
- Sublime Version: 4102
- OS Version: Ubuntu 20.04 and Windows 7/10

this printf could not be parsed, when the left bracket in the next line. (github could not, too)

``` c++
printf
    ("Hello!");
```

And I met it here: https://github.com/Avnu/gptp/blob/master/common/ptp_message.cpp#L1375

Please ,reference: sublimehq/sublime_text#4152

I am not sure if I need to create this new issue here again and close that old one?

### LINKED ISSUE #3224: [C++] Empty preprocessor line
### What happened?

Empty or null preprocessor lines (a single '#' and a newline) are not highlighted like 'normal' preprocessor lines.
This shocked me, since I didn't even know null preprocessor lines were legal. But they are.

Example from boost;
```C++
#ifndef BOOST_CONFIG_HPP
#  include <boost/config.hpp>
#endif
#                         <------------ this line will not be colored like the rest of the preprocessor directives
#if defined(BOOST_HAS_PRAGMA_ONCE)
#  pragma once
#endif
```

Expected to have the pound sign / hashtag / octothorpe to be highlighted as the start of a preprocessor directive.

### LINKED ISSUE #3317: [PackageName] C: pointer to function: return type is not highlighted
### What happened?

C: pointer to function: return type is not highlighted:
image

Please fix.

ST4, build 4126.

### LINKED ISSUE #3504: [C/C++] typedef marked `const` as an indexed entity.name.type
### What happened?

```c++
typedef bool (ParseResult::*BooleanType)() const;
```

Sublime Text renders it as:

typedef with const

and naturally indexes it, so it becomes a target for GoTo Definition:

Definition of const

### LINKED ISSUE #3592: [C] Wrong scoping of number if sign and number are on different lines.
### What happened?

The following test case fails with the digit `1` not being scoped as part of the number.
```c
int x = -
//^ storage.type.c
//    ^ keyword.operator.assignment.c
//      ^ keyword.operator.arithmetic.c
   1.53;
// ^^^^ meta.number.float.decimal.c constant.numeric.value.c
//  ^ punctuation.separator.decimal.c
//     ^ punctuation.terminator.c
```

Interestingly, the scoping works if there are no leading spaces.

### LINKED ISSUE #3805: [C++] Shift/stream operators considered arithmetic operators
### What happened?

In C++, the shift operators `>>`, `<<`, `>>=`, and `<<=` should not be put in the same scope as basic arithmetic operators `>` and `<`. Instead, a new scope such as `keyword.operator.shift.c` should be introduced.

Example:
image

### LINKED ISSUE #3901: [C++] Cannot goto functions after a very long function.
### Expected behavior

The symbols after the function `JS_CallInternal` like `JS_Call` have a scope named `entity.name.function.c` which make them be indexed by Sublime Text.
image

### Actual behavior

The symbols after the function `JS_CallInternal` does not have the `entity.name.function.c` scope and cannot be indexed by Sublime Text.
image
image

### Steps to reproduce

The [source code](https://github.com/bellard/quickjs) repository.

### LINKED ISSUE #3938: [C++] Go to function append the return type of the function to the function name if separated on a different line if the return type is a custom type
### Expected behavior

The same behavior that exists for built in types (int, void)

### Actual behavior

The return type is being appended to the function name in goto function as shown in the video below:

https://github.com/sublimehq/Packages/assets/53301143/fceef6bd-933b-406a-8b77-96ef519862e3

Note the issue does not happen with void and int types for example (I have not tested but probably all built in types works fine)

### Steps to reproduce

define Internal as static `#define Internal static` 
define your own struct  
separate the return type prefixed by `Internal` in its own line, while the function name in the line after it.

### LINKED ISSUE #3956: [C] New C23 keywords
### What happened?

New types got added to C23 like `nullptr` and `constexpr` and need syntax highlighting.
Here's the [full list](https://en.wikipedia.org/wiki/C23_(C_standard_revision))

### LINKED ISSUE #4058: Incorrect syntax highlighting in C with GNU Statement Expressions
### What happened?

Hi, this piece of C code has incorrect syntax highlighting in sublime text

image

```
struct test {
    int x;
};

int main() {
    struct test* a = ({
        nullptr;
    });

    return 0;
}
```

### LINKED ISSUE #4104: [C] Sublime doesn't index a function name on its own line preceded with an enumeration return type
### Expected behavior

Expecting `mix()` to be indexed, that is, to have the `entity.name.function.c` scope.

```
enum meow {
    X, Y, Z,
};

enum allo
mix()
{
    return Z;
}
```

### Actual behavior

`mix()` is not indexed.

The `mix` part has the `meta.enum.c` scope.

However, it works as expected like this:

```
enum meow {
    X, Y, Z,
};

enum allo mix()
{
    return Z;
}
```

### Steps to reproduce

1. In a C file, write:

   ```
   enum meow {
       X, Y, Z,
   };
   
   enum allo
   mix()
   {
       return Z;
   }
   ```
2. Try to go to the `mix` symbol.

camila314:
I am currently working on a full rewrite of the C family using modern Sublime Text syntax highlighting features. I plan to rewrite C, Objective-C, C++, and Objective-C++ in that order. Note that as of right now, only the C syntax has been edited and the syntax test file does not yet have any annotations.

## Changes made to C Highlighting
- Removed `#if 0` and `#if 1 #else` creating a commented out block
- Added `i` and `j` to numerical suffixes (GNU extension)
- Add `_Alignof`, `alignof`, and `offsetof`, `_Static_assert`, `static_assert`, `_Pragma` as word operators
- Every double-underscored identifier without another scope is treated as support constant
- `__cplusplus` is a `support.constant`
- Add `nullptr_t` (C23) and `max_align_t` (C11) to list of default types
- Add `_Thread_local`, `thread_local`, `_Atomic`, `_Alignas` and `_Noreturn` as storage modifiers
- Fixed bug where a function parameter followed by a newline wouldn't be highlighted properly
- Add more `__declspec` properties
- Add common `__attribute__` properties
- Add digit separator `'` (C23)
- Add scopes for `__VA_ARGS__` and `__VA_OPT__` (C23)
- Give defined macro a `suport.macro.c` scope
- Add `#embed`, `#elifdef`, and `#elifndef` directives (C23)
- Add multi-line double-slashed comments via `\`
- Add `typedef` support for function ptrs
- Function pointer arguments in a `typedef` now share the same scoping as regular function defs

## Issues
- [x] Fixes #4104
- [x] Fixes #4058
- [x] Fixes #4104
- [x] Fixes #3956
- [x] Fixes #3938
- [x] Fixes #3901
- [x] Fixes #3805
- [x] Fixes #3592
- [x] Fixes #3504
- [x] Fixes #3317
- [x] Fixes #3224
- [x] Fixes #2815
- [x] Fixes #2762
- [x] Fixes #2334
- [x] Fixes #2221
- [x] Fixes #2047
- [x] Fixes #1830
- [x] Fixes #1438
- [x] Fixes #1138
- [x] Fixes #1093 
- [x] Fixes #1070


camila314:
My motivations are simple
image

C++/C.sublime-syntax
```diff
@@ -614,6 +669,9 @@ contexts:
   eol-pop:
     - match: $\n?
       pop: 1
+  consume-pop:
```
michaelblyons, line 672: I'm very suspicious of this context. Are you using it somewhere?

camila314: oh i guess i removed whatever functionality it had, so i will remove it in the next commit

deathaxe:
Some general advices:

`main` and `prototype` are primary entry points and thus shouldn't be located somewhere in the middle of anywhere. They should be the first contexts.

I'd recommend to organize contexts in a logical tree-like top-down structure, grouping related things like preprocessors, declarations, statements, expressions, keywords, identifiers, constants, operators, variables etc. together.

Various syntaxes even use `###[ SECTION NAME]###` comments to visually separate bigger blocks of related contexts. This also enables easy folding of irrelevant blocks.

All of that will certainly help to keep track of structure as syntax definition grows. Keeping same context order and structure especially is important when extending a syntax, which will be the casse for C++.

The more syntaxes in this repo follow a common schema the easier it is to maintain them.

_Premature ordering can be kill of creativity, so order is not that important for prototyping. Maybe however don't wait too long to add some structure._

---

Please don't just add some code snippets to syntax test files, but also design assertions on the way going. It sometimes takes more time to write tests than syntax, they are however the only way to ensure not to break wanted/designed behavior by automated syntax tests. 

When writing rules for certain constructs you have in mind what you find important. Tests are just an expression of those thoughts or tools to verify your ideas/expectations against reality and vice versa.

Tests can (or even should) also contain incomplete statements/expressions, to ensure code doesn't break too badly. At least it helps to ensure to limit breaking highlighting to small regions or makes you think about how to achieve it.

Assertions should also contain some negative tests to ensure meta scope boundaries are maintained correctly or to avoid unwanted overlapping meta scopes. These are the less obvious issues, which may however impact various functionalities such as syntax based folding.

deathaxe:
FWIW, context order and structure of existing C syntaxes may not be ideal nor a good template to rely on. Better look at something more recent, such as Java, PHP or Python.

camila314:
> I don't think those are the actual tests. 😉 I note the path to the syntax def in `User`, and the final scope being `c-2`.
> 
> May I recommend similar to your PackageDev clone, you look at the suggestions in the ReadMe of this repo? Then you can modify the C syntax directly. If you want to keep the old one for comparison, you can rename it or move _that_ one to `User`.
> 
> I'm also curious why you changed the assertion comment from `//` to `/*` ... `*/`.

the only reason I haven't done that is because the old c++ will break due to missing imports, most unfortunately

C++/C.sublime-syntax
```diff
@@ -505,103 +512,151 @@ contexts:
...
+    - match: \\$\n
       scope: punctuation.separator.continuation.c
       push:
         - match: ^
           pop: 1
+  preprocessor.include:
+    - meta_scope: meta.preprocessor.c
+    - match: \"
+      scope: punctuation.definition.string.begin.c
+      push: string.double
```
deathaxe: Strings are commonly used in various contexts and thus may deserve a dedicated context. 

Maybe something like:

```suggestion
    - include: string.quoted.double
```

And in "literal" or "constants" section:

```yml
  string.quoted.double:
    - match: \"
      scope: punctuation.definition.string.begin.c
      push: string.quoted.double.body

  string.quoted.double.body:
    - match: \"
      scope: punctuation.definition.string.end.c
      pop: 1
  
```

C++/C.sublime-syntax
```diff
@@ -160,46 +167,46 @@ contexts:
         4: constant.numeric.suffix.c
         5: invalid.illegal.numeric.suffix.c
     # binary integer (C23)
-    - match: \b(0[bB])([0-1]*)(?:({{integer_suffix}})|({{bin_suffix}}))?
+    - match: \b(0[bB])([01](?:[01]|'(?=[^']))*)?(?:({{integer_suffix}})|({{bin_suffix}}))?
```
deathaxe, line 549: If we want to enforce at least a digit between each `'`:

```suggestion
    - match: \b(0[bB])([01](?:'?[01])*)?(?:({{integer_suffix}})|({{bin_suffix}}))?
```

But for the sake of supporting intermediate states patterns could probably be a bit more lazy, if there are no other conflicting terms/expressions to be expected.

```suggestion
    - match: \b(0[bB])(['01]*)(?:({{integer_suffix}})|({{bin_suffix}}))?
```

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,708 @@ first_line_match: |-
...
       push:
-        - match: ':'
-          scope: punctuation.separator.c
-          pop: true
-        - include: expressions
-
-  modifiers-parens:
+        - match: \(
+          scope: punctuation.section.group.begin.c
+          push: declspec
```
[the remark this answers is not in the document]
deathaxe: I'd also try to keep primary popping contexts (e.g.: `- match: \)`) at the beginning, so they always have highest priority.

Also note added meta scopes, which may help to provide more context for completions, plugins or color schemes to enable special highlighting.

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,733 @@ first_line_match: |-
...
+    - include: operator.word
+    - include: type
+    - include: keyword
+    - include: constant
 
-  line-comment-documentation-body:
-    - meta_include_prototype: false
-    - meta_scope: comment.line.documentation.c
-    - include: line-comment-end
+  ###[SIMPLE STRUCTURES]###
```
michaelblyons, line 260: If you add an extra space on either side of the words, PackageDev will mark as a heading (`###[ SIMPLE STRUCTURES ]###`).

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,733 @@ first_line_match: |-
...
+      scope: keyword.operator.variadic.c
+  operator.increment-decrement:
+    - match: \+\+|--
+      scope: keyword.operator.arithmetic.c
+  operator.accessor:
+    - match: \.|\-\>
+      scope: punctuation.accessor.c
+  operator.pointer:
+    - match: \&|\*
+      scope: keyword.operator.arithmetic.c
```
michaelblyons, line 515: Is this the right scope? I think the old tests had `keyword.operator.arithmetic` just because they couldn't identify that these were pointers.

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,733 @@ first_line_match: |-
...
+    - match: \bcontinue\b
+      scope: keyword.control.flow.continue.c
+    - match: \bbreak\b
+      scope: keyword.control.flow.break.c
+    - match: \breturn\b
+      scope: keyword.control.flow.return.c
+    - match: \bgoto\b
+      scope: keyword.control.flow.goto.c
+  keyword.control:
+    - match: \b(?:if|else|for|while|do|switch|case|default)\b
```
michaelblyons, line 663: Some languages have extra subscopes for some of these.

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,311 @@ first_line_match: |-
...
-      scope: constant.language.objc
-
-  unique-modifiers:
-    - match: (@)(synchronized|public|private|protected)\b
-      scope: storage.modifier.objc
+  constant:
+    - meta_append: true
+    - match: \b(nil|Nil)\b
+      scope: constant.language.null.objc
+    - match: (\b|\@)(YES|NO|true|false)\b
```
michaelblyons, line 218: Split into `.true` and `.false` subscope items like you did with C.

michaelblyons:
In no particular order, a bunch of this repository's open C Family issues: #409 #888 #1034 #1070 #1082 #1093 #1107 #1138 #1156 #1249 #1333 #1408 #1438 #1749 #1760 #1772 #1774 #1794 #1830 #1901 #1906 #1913 #1914 #1931 #1993 #2047 #2094 #2121 #2221 #2251 #2334 #2352 #2385 #2471 #2762 #2815 #2861 #2883 #2937 #3091 #3113 #3114 #3145 #3224 #3317 #3343 #3487 #3504 #3592 #3714 #3715 #3805 #3901 #3920 #3938 #3956 #3960 #3984 #4058 #4104

Not all of these may be valid, fixable, or even things we want. But it's at least worth looking at them to see if they happen to be fixed.

C++/C.sublime-syntax
michaelblyons: Okay, so 79. :sweat_smile:

uncenter:
I can't tell if these have been fixed yet, but some issues I noticed for C++ files on the latest stable that don't seem to be mentioned above:

CleanShot 2025-02-12 at 09 30 46

The names of namespaces being used don't have any specific scopes.

CleanShot 2025-02-12 at 09 30 58

I would expect the `Typedef` type here to have some scope related to type names.

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,735 @@ first_line_match: |-
...
+      push: enclosure.block
+  enclosure.parenthesis:
+    - meta_scope: meta.group.c
+    - include: expression
+    - match: \)
+      scope: punctuation.section.group.end
+      pop: 1
+    - match: \}|\] # Fallback
+      scope: invalid.illegal.stray-bracket-end.c
+      pop: 1
```
deathaxe, line 378: This may easily cause owning contest to also break or not being terminated correctly due to brackets already being consumed here.

I'd recommend to use lookaheads and consume stray brackets only in top-level context.

michaelblyons:
Edited OP to add GitHub's format.

C++/C.sublime-syntax
```diff
@@ -193,6 +193,18 @@ contexts:
...
+      captures:
+        1: variable.function.c support.function.std.c
+    - match: ({{identifier}})\s*$
+      captures:
+        1: variable.function.c
+    - match: \(
+      scope: punctuation.section.group.begin.c
+      set: enclosure.parenthesis
+    - match: (?=\S)
+      scope: invalid.illegal.bad
```
deathaxe: A lookahead doesn't consume anything and thus doesn't receive any scope.

You most likely want to `fail: try-call`?

C++/C.sublime-syntax
```diff
@@ -193,6 +193,18 @@ contexts:
         - include: else-pop
     - match: (?=\S)
       fail: try-function
+  expression.try-call:
+    - match: ({{std_funcs}})\s*$
+      captures:
+        1: variable.function.c support.function.std.c
```
deathaxe: IMHO, we don't need to stack those scopes.

```suggestion
        1: support.function.std.c
```

C++/C.sublime-syntax
camila314: oops this is an artifact from debugging yes you are correct

deathaxe:
Not sure how useful #1831 still is. It came with some interesting parts, but scoping variable definitions `entity.name` is not that great. Maybe some things can even be solved with newer features.

Disclaimer: It's a while (some years) I had a look at it, the last time.

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,773 @@ first_line_match: |-
...
+        1: keyword.operator.c
+        2: entity.name.type.typedef.c
+        3: punctuation.section.group.begin.c
+      set: function.params
+
+    - match: ({{identifier}})\s*(?=;)
+      captures:
+        1: entity.name.type.typedef.c
+      pop: 1
+    - match: ; # Fallback
```
deathaxe, line 321: Missing `scope`? Alternatively use a lookahead and let a parent context consume/scope it, if it is required to handle multiple context pops.

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
+    - match: (?:\+|\-)\s*(?=\()
+      push: [selector, selector.return]
+    - include: expression
 
-  properties:
-    - match: ((@)property)\s*(\()
+  property:
+    - match: \(
+      scope: punctuation.section.group.begin.objc
+      push: property.attrs
```
deathaxe, line 78: If `@property ()()()` is not a valid statement, I'd suggest to `set` property.attr.

```suggestion
      set: property.attrs
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
+    - include: else-pop
+  property.attrs:
+    - match: \)
```
deathaxe, line 81: All groups should receive a corresponding meta scope.

```suggestion
    - meta_scope: meta.group.objc
    - match: \)
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
+      scope: variable.parameter.objc
+      pop: 1
+
+  protocol-list:
+    - match: \<(?=({{identifier}}|,|\s|\*)*\>)
       scope: punctuation.section.scope.begin.objc
       push:
+        - include: expression-identifiers
+        - include: punctuation.comma
+        - include: operator.pointer
```
deathaxe, line 134: Please, always start contexts with `meta_scope` followed by real patterns.

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
+        1: keyword.operator.bitwise.objc
+      branch_point: block-fn
+      branch: [block-fn, pop-immediately]
+    - match: '(@catch)\s*(\()'
+      captures:
+        1: keyword.control.exception.objc
+        2: punctuation.section.group.begin.objc
+      push: function.params
+    - match: (?=\b(self)\s*(\.))
+      push: self-access
```
deathaxe, line 153: Maybe just...

```suggestion
    - match: \bself\b
      scope: variable.language.objc
      push: self-access

...


  self-access:
    - match: \s*(\.)\s*({{identifier}})
      captures:
        1: keyword.operator.accessor.objc
        2: variable.other.member.objc
    - include: pop-immediately


```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
 
-  ns-predicate:
-    - match: '(?=\])'
-      pop: true
-    - match: \bargument(?:Array|s)(:)
-      scope: support.function.any-method.name-of-parameter.objc
+  block-fn:
+    - match: \(
+      scope: punctuation.section.group.begin.objc
+      push: function.params
```
deathaxe, line 159: Note this would also match `void func()()() {}`

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
-    - match: (@)(try|catch|finally|throw)\b
-      scope: keyword.control.exception.objc
+  selector-obj:
+    - meta_scope: meta.selector.objc
+    - meta_content_scope: meta.selector.method-name.objc
+    - match: \)
+      scope: punctuation.section.group.end.objc
+      pop: 1
+    - match: .
+      scope: support.function.objc
```
deathaxe, line 214: A `.` pattern causes each character to become a single token, which is not ideal for various reasons and should be avoided if possible.

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
+
+  keyword:
+    - meta_prepend: true
+    - match: \b(self|super)\b
+      scope: variable.language.objc
+    - match: \@(property|synthesize|dynamic)\b
+      scope: keyword.other.objc
+    - match: '(@available)\b(\()'
```
deathaxe, line 222: A `\b` in between is not required as `\(` is already a word boundary. Optionally supporting whitespace via `\s*` may be useful however.

```suggestion
    - match: '(@available)\s*(\()'
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
+      push: available
+    - match: '(@selector)\b(\()'
       captures:
-        1: punctuation.definition.keyword.objc
-    - match: (@)(defs|encode)\b
-      scope: keyword.other.objc
+        1: keyword.operator.word.objc
+        2: punctuation.section.group.begin.objc
+      push: selector-obj
+    - match: '@(selector|encode|available)'
```
deathaxe, line 232: Tiny optimization, but avoiding unused capture groups helps with performance a little bit.

```suggestion
    - match: '@(?:selector|encode|available)'
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,334 @@ first_line_match: |-
...
+      scope: storage.modifier.objc
+    - match: \@(private|public|protected|package|optional|required)\b
+      scope: storage.modifier.objc
+  keyword.type:
+    - meta_append: true
+    - match: \b(?:BOOL|Class|SEL|IMP|id|instancetype|IBAction)\b
+      scope: storage.type.objc
+    - match: \@(?:interface|class|implementation|protocol|end)\b
+      scope: keyword.declaration.objc
+    - match: '(?=\bNS\w+\b)'
```
deathaxe, line 252: Not quite sure, but the following might be enough

```suggestion
    - match: '(?=\bNS)'
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,328 @@ first_line_match: |-
...
-          scope: constant.other.placeholder.objc
-        - include: scope:source.c#string_placeholder
+      push: string.single
+  string.double-ns:
+    - meta_scope: string.quoted.double.objc
+    - include: string.double
+    - match: '%@'
+      scope: constant.other.placeholder.objc
+
+  available:
```
deathaxe, line 192: 1. Missing meta content scope.
2. Maybe move `\)` pattern to beginning of context for max prio.

```suggestion
  available.body:
    - meta_scope: meta.available.objc  # or meta.directive.available ???
    - meta_content_scope: meta.group.objc
    - match: \)
      scope: meta.group.objc punctuation.section.group.end.objc
      pop: 1
```

Note: using `meta_content_scope` to help with properly scoping only opening \( in `- match: '(@available)\s*(\()'`

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,328 @@ first_line_match: |-
...
+    - match: \@(?:property|synthesize|dynamic)\b
+      scope: keyword.other.objc
+    - match: '(@available)\s*(\()'
       captures:
-        1: punctuation.definition.keyword.objc
-    - match: (@)(synchronized)\b
-      scope: keyword.control.synchronize.objc
+        1: keyword.operator.word.objc
+        2: punctuation.section.group.begin.objc
+      push: available
```
deathaxe, line 222:
```suggestion
        1: keyword.operator.word.objc
        2: meta.group.objc punctuation.section.group.begin.objc
      push: available.body
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,328 @@ first_line_match: |-
...
-    - match: \b(NS_DURING|NS_HANDLER|NS_ENDHANDLER)\b
-      scope: keyword.control.macro.objc
-    - match: (@)(try|catch|finally|throw)\b
-      scope: keyword.control.exception.objc
+  selector-obj:
+    - meta_scope: meta.selector.objc
+    - meta_content_scope: meta.selector.method-name.objc
+    - match: \)
+      scope: punctuation.section.group.end.objc
+      pop: 1
```
deathaxe, line 208: As punctuation uses `meta.section.group` a corresponding `meta.group` should be present.

```suggestion
    - meta_scope: meta.selector.objc
    - meta_content_scope: meta.group.objc meta.selector.method-name.objc
    - match: \)
      scope: meta.group.objc punctuation.section.group.end.objc
      pop: 1
```

Objective-C/Objective-C.sublime-syntax
```diff
@@ -13,200 +15,328 @@ first_line_match: |-
...
+        1: keyword.operator.word.objc
+        2: punctuation.section.group.begin.objc
+      push: available
+    - match: '(@selector)\s*(\()'
       captures:
-        1: punctuation.definition.keyword.objc
-    - match: (@)(defs|encode)\b
-      scope: keyword.other.objc
+        1: keyword.operator.word.objc
+        2: punctuation.section.group.begin.objc
```
deathaxe, line 226:
```suggestion
        2: meta.group.objc punctuation.section.group.begin.objc
```

deathaxe (commented):
Here are some general notes and advices about common/best practices to help improve your already increadible great syntax writing style.

C++/C++.sublime-syntax
```diff
@@ -26,2340 +28,616 @@ first_line_match: |-
...
+      captures:
+        1: punctuation.section.brackets.end.c++
+        2: meta.group.c++ punctuation.section.group.begin.c++
+      set: [lambda-try-not, enclosure.parenthesis]
 
-  function-definition-body:
-    - meta_content_scope: meta.function.c++ meta.block.c++
-    - match: '\{'
-      scope: punctuation.section.block.begin.c++
+    - match: (\])\s*(\{|\-\>|mutable\b)
```
deathaxe, line 343: `fail`-patterns don't need capture groups nor do they need to consume text as it is rewinded anyway. Thus lookahead patterns are optimal choise.l

C++/C++.sublime-syntax
```diff
@@ -26,2340 +28,616 @@ first_line_match: |-
...
-    - match: '\b(namespace)\s+(?={{path_lookahead}}\s*\{)'
-      scope: meta.namespace.c++
+        1: keyword.control.exception.c++
+        2: meta.group.c++ punctuation.section.group.begin.c++
+      push: function-params
+    - match: \b(try|catch|finally|throw)\b
+      scope: keyword.control.exception.c++
+    - match: \bco_(await|yield|return)\b
+      scope: keyword.control.c++
+    - match: \b(requires)\s*(\()\b
```
deathaxe, line 554: A `\(` is already a word boundary. Thus trailing `\b` is redundant or even may cause trouble.

```suggestion
    - match: \b(requires)\s*(\()
```

C++/C++.sublime-syntax
```diff
@@ -26,2340 +28,616 @@ first_line_match: |-
...
-          pop: true
-    - include: expressions
-
-  preprocessor-practical-workarounds:
-    - include: preprocessor-convention-ignore-uppercase-ident-lines
-    - include: scope:source.c#preprocessor-convention-ignore-uppercase-calls-without-semicolon
+        1: keyword.operator.word.c++
+        2: meta.group.c++ punctuation.section.group.begin.c++
+      push: [function-tail, function-params]
+    - match: \b(requires)\b
```
deathaxe, line 559: No capture group needed here. Just causes (tiny) overhead.

```suggestion
    - match: \brequires\b
```

C++/C++.sublime-syntax
```diff
@@ -26,2340 +28,616 @@ first_line_match: |-
...
+      scope: keyword.operator.bitwise.c++
+  operator.word:
+    - meta_prepend: true
+    - match: \b((const|static|reinterpret|dynamic)_cast|static_assert|delete|new|typeid|noexcept)\b
+      scope: keyword.operator.word.c++
+    - match: \b(and|x?or|not)(_eq)?\b
+      scope: keyword.operator.logical.c++
+    - match: \b(bit(and|or)|compl)\b
+      scope: keyword.operator.bitwise.c++
+    - match: \bxor(_eq)\b
```
deathaxe, line 590: What's that? `xor(?:_eq)?` or `xor_eq`?

C++/C++.sublime-syntax
```diff
@@ -26,2340 +28,616 @@ first_line_match: |-
...
+    - meta_prepend: true
+    - include: template
+  operator.bitwise:
+    - match: \&|\||\^|~(?!{{identifier}}\s*\()|<<|>>
+      scope: keyword.operator.bitwise.c++
+  operator.word:
+    - meta_prepend: true
+    - match: \b((const|static|reinterpret|dynamic)_cast|static_assert|delete|new|typeid|noexcept)\b
+      scope: keyword.operator.word.c++
+    - match: \b(and|x?or|not)(_eq)?\b
```
deathaxe, line 586:
```suggestion
    - match: \b(?:and|x?or|not)(?:_eq)?\b
```

or maybe for readability even

```suggestion
    - match: \b(?:and|xor|or|not)(?:_eq)?\b
```

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,833 @@ first_line_match: |-
...
+      scope: invalid.illegal.syntax.c
+    - match: (\.|\-\>)(?=\s*{{identifier}})
+      scope: punctuation.accessor.dot.c
+      branch_point: access
+      branch: [access, pop-immediately]
+    - match: \.|\-\>
+      scope: punctuation.accessor.c
+  operator.pointer:
+    - match: \&|\*
+      scope: keyword.operator.c
```
michaelblyons, line 603: I think I've seen these scoped as `keyword.operator.reference `.

C++/C++.sublime-syntax
```diff
@@ -51,9 +51,9 @@ contexts:
     - match: (?={{identifier}}\s*<)
       branch_point: function-template
       branch: [function-template, consume-ident]
-    - match: \b{{identifier}}\b
+    - match: '{{identifier}}'
```
michaelblyons, line 54: Was there a problem with the word breaks?

camila314: no, but {{identifier}} already has word breaks so i thought it was redundant

rwols:
First of all let me say I'm happy someone is looking at these syntaxes. I'm impressed by the sheer speed you are displaying.

I haven't read the 50+ posts on this pull request but I'm trying it out locally now, specifically the C++ syntax.

### Problem 1: Eager scoping of entity.name.function

missing paragraph?

The first thing I'm noticing is that for plain function calls, it assigns entity.name.function. Regrettably, I would want this behavior as it was before. Function calls should be scoped variable.function. Examples where it too eagerly assigns entity.name.function:

```cpp
C& C::f(std::string_view z) noexcept
{
    if (const auto e = a().b(z)) {
        return f(*e); // f *should* be scoped as variable.function
    }
    return *this;
}
```

```cpp
C& C::f(const X& x) noexcept
{
    const std::vector<X*> xs = f().g<X>(); // the call to g<X> *should* be variable.function
    m = std::distance(xs.begin(),
        std::find_if(xs.begin(), xs.end(), [&](const X* c) -> bool {
            return c == std::addressof(x);
        })
    );
    return *this;
}
```

```cpp
void Frame::OnOpenProject(wxCommandEvent& /*event*/)
{
    if (/*...current content has not been saved...*/ 0) {
        const wxString body("Current content has not been saved! Proceed?");
        //             ^^^^ variable.function
        const wxString title("Please confirm");
        //             ^^^^^ variable.function
        if (wxMessageBox(body, title, wxICON_QUESTION | wxYES_NO, this) == wxNO)
            return;
    }
    wxFileDialog openFileDialog(this, wxT("Open a project file"), "", "", "JSON files (*.json)|*.json", wxFD_OPEN);
    //           ^^^^^^^^^^^^^^ variable.function
    if (openFileDialog.ShowModal() == wxID_CANCEL)
        return;
    try {
        Project project(Project::LoadFromFile(openFileDialog.GetPath()));
        //      ^^^^^^^ variable.function
        SetCurrentWorkingDirectory(project.GetBaseDir());
        // ^^^^^^^^^^^^^^^^^^^^^^^ variable.function
    } catch (const std::exception& ex) {
        wxLogError("Failed to load project: %s", ex.what());
        // ^^^^^^^ variable.function
        SetStatusText("project NOT loaded");
        // ^^^^^^^^^^ variable.function
        return;
    }
    SetStatusText("project loaded!");
}
```

This eager scoping of `entity.name.function` also happens in class initializer lists:

```cpp
SomeClass::SomeClass(const S& title, const P& pos, const Z& size)
    : m_member(std::make_unique<G>(m_other_member))
    //              ^^^^^^^^^^^ variable.function
{
}
```

It also too eagerly assigns entity.name.function to macro invocations:

```cpp
void C::Update()
{
    LOG_INFO(FACILITY) << "hello there general kenobi";
    // ^^^^^ - entity.name.function
    //                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ string.quoted.double
    const auto& state = getState();
    if (m.Update(state)) {
        LOG_INFO(FACILITY) << "frobnicating";
        // ^^^^^ - entity.name.function
        //                    ^^^^^^^^^^^^^^ string.quoted.double
        getFoo().f(getBlorb());
        LOG_INFO(FACILITY) << "did the frobnicating";
        // ^^^^^ - entity.name.function
        //                    ^^^^^^^^^^^^^^^^^^^^^^ string.quoted.double
    }
}
```

C++/C.sublime-syntax
```diff
@@ -13,1427 +14,893 @@ first_line_match: |-
...
+      push: type.typedef
+    - include: enum-def
+    - match: \bstruct\b
+      scope: keyword.declaration.struct.c
+      branch_point: struct
+      branch: [type.struct, pop-immediately]
+    - match: \bunion\b
+      scope: keyword.declaration.union.c
+      branch_point: union
+      branch: [type.union, pop-immediately]
```
deathaxe, line 338: Maybe organize all sorts of definitions in separate contexts, grouped logically? This way we don't need to introduce exceptions like `enum-def` in case an extending syntax needs to override it.

deathaxe (commented):
OK, seems low hanging fruits are pruned and we can start to the hard part of writing syntax defintions - the details. ;-)

A formal note ahead:

Not separating contexts with emptly lines in combination with separating patterns is odd and undesirable.

For readability reasons all contexts should be separated by an emptly line.

Use `###[ ... ]###` to group related contexts instead.

---

Highlighting seems to work, but tests are not sufficent to prove it being ready for the wild, to be honest. Especially meta scopes seem some general refinements to ensure correct order, boundaries or not overlapping them unintentionally. That's actually the most tricky part of writing syntax definitions.

So here are some examples, of what may need some attention:

rwols:
### Problem 2: Does not handle templates well

The following are all valid C++ template classes:

```cpp
template <class T> struct X {
//               ^ punctuation.definition.generic.end
};

template <class T, class S> class Y {
//                        ^ punctuation.definition.generic.end
};

template <typename T, typename S> class Z {
//                              ^ punctuation.definition.generic.end    
};

template <template <class> T> struct W {
//                       ^ punctuation.definition.generic.end
//                          ^ punctuation.definition.generic.end
};

```

Compiler Explorer link: https://godbolt.org/z/o5P3MrjGM

Template functions are also not scoped correctly:

```cpp
#include <type_traits>
#include <utility>

template <typename T> struct is_pyobject : std::false_type {};
//                  ^ punctuation.definition.generic.end

struct object {};

template <typename T, std::enable_if_t<is_pyobject<T>::value, int> = 0>
//                                                                    ^ punctuation.definition.generic.end
auto object_or_cast(T &&o) -> decltype(std::forward<T>(o)) {
    return std::forward<T>(o);
}

template <typename T, std::enable_if_t<!is_pyobject<T>::value, int> = 0>
//                                                                     ^ punctuation.definition.generic.end

object object_or_cast(T &&o);
//                          ^ - invalid.illegal

int main() {
    return 0;
}
```

Compiler Explorer link: https://godbolt.org/z/xb7zKb85j

For working with templates, I invite you to clone this project locally and see if you can make the template-heavy code in the header files work: https://github.com/pybind/pybind11/tree/master/include/pybind11.

I'm just reporting global themed problems as I go :) thanks once again for tackling this!

C++/syntax_test_c.c
```diff
@@ -1,1555 +1,1250 @@
...
+    1 > 2 >= 3 == 4 != 5 <= 6 < 7
+//  ^ meta.number.integer.decimal.c constant.numeric.value.c 
+//    ^ keyword.operator.comparison.c 
+//      ^ meta.number.integer.decimal.c constant.numeric.value.c 
+//        ^^ keyword.operator.comparison.c 
+//           ^ meta.number.integer.decimal.c constant.numeric.value.c 
+//             ^^ keyword.operator.assignment.c 
+//                ^ meta.number.integer.decimal.c constant.numeric.value.c 
+//                  ^ keyword.operator.arithmetic.c 
+//                   ^ keyword.operator.assignment.c
```
deathaxe, line 201: Operator patterns is a sensible part. You need to carefully organize them in correct order, starting with those of 3 chars, followd by 2 and then 1 to ensure they are matched correctly.

```suggestion
//                  ^^ keyword.operator.comparison.c 
```

camila314:
I am pretty tired of maintaining this so i will probably not do much more commits. anyone else is very free to pick it up for me

camila314:
guess who's back 😁

FichteFoll:
Marking it as draft until it is ready for review/merge again.

michaelblyons:
Camila copied to an external package at https://github.com/camila314/C-Family-Rewrite

PR #4153, opened 2025-02-12
[CSS] Fix escaped identifiers

### LINKED ISSUE #4151: [CSS] escaped psuedo class is still identified as psuedo class
### What happened?

```css
.component\:profile {
    
}
```

The `:profile` should not be identified as a psuedo-selector since it's escaped with a backslash.

Image

Image

deathaxe:
This PR ...

1. fixes #4151, caused by premature termination of identifiers with escaped characters
2. fixes ascii alphanumeric patterns in identifiers (as `[:alpha:]` consumes all unicode alphanumerics)

### DIFF
--- a/CSS/CSS.sublime-syntax
+++ b/CSS/CSS.sublime-syntax
@@ -15,8 +15,8 @@ variables:
   # https://www.w3.org/TR/css3-selectors/#lex
   unicode: \\\h{1,6}[ \t\n\f]?
   escape: (?:{{unicode}}|\\[^\n\f\h])
-  nmstart: (?:[_[:alpha:][^\x{0}-\x{127}]]|{{escape}})
-  nmchar: (?:[-_[:alnum:][^\x{0}-\x{127}]]|{{escape}})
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):

CSS/CSS.sublime-syntax
```diff
@@ -125,7 +125,7 @@ variables:
 
   # Foreign Tags
   # https://html.spec.whatwg.org/multipage/parsing.html#tag-name-state
-  tag_name_begin: (?=[[:alpha:]])
+  tag_name_begin: (?=[A-Za-z])
```
keith-hall, line 128: I'm pretty sure HTML allows unicode tag names, no?

deathaxe: These patterns were originally copied from _HTML (Plain).sublime-syntax_, which implements tag parsing according to https://html.spec.whatwg.org/multipage/parsing.html#tag-open-state.

Those rules start tag names only with ASCII letters. Any following char can be unicode.

keith-hall (approved):

PR #4154, opened 2025-02-12
[SQL] improve MySQL syntax

keith-hall:
- support charset in return types
- support $$ as a statement terminator
- support `SET` ... `:=` syntax
- support `SET` with system variable assignment

### DIFF
--- a/SQL/MySQL.sublime-syntax
+++ b/SQL/MySQL.sublime-syntax
@@ -677,9 +677,9 @@ contexts:
     - include: pop-on-top-level-reserved-word
 
   variable-member:
+    - meta_include_prototype: false
     - meta_content_scope: variable.other.member.mysql
-    - match: ''
-      pop: 1
+    - include: immediately-pop          
 
 ###[ GRANT STATEMENTS ]########################################################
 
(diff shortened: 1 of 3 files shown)


SQL/MySQL.sublime-syntax
```diff
@@ -662,6 +676,11 @@ contexts:
     - include: expressions
     - include: pop-on-top-level-reserved-word
 
+  variable-member:
+    - meta_content_scope: variable.other.member.mysql
+    - match: ''
+      pop: 1
```
deathaxe, line 682: Exclude prototype and re-use `immediately-pop`?

```suggestion
    - meta_include_prototype: false
    - meta_content_scope: variable.other.member.mysql
    - include: immediately-pop          
```

michaelblyons: Maybe without the trailing spaces? :wink:

SQL/MySQL.sublime-syntax
```diff
@@ -677,9 +677,9 @@ contexts:
     - include: pop-on-top-level-reserved-word
 
   variable-member:
+    - meta_include_prototype: false
     - meta_content_scope: variable.other.member.mysql
-    - match: ''
-      pop: 1
+    - include: immediately-pop
```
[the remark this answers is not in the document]
keith-hall:
```suggestion
    - include: immediately-pop
```

michaelblyons (approved):

deathaxe (approved):

PR #4156, opened 2025-02-13
[AppleScript] sregex compatibility

michaelblyons:
See #481

There's lots more that could be done to improve this package, but I've killed the lookbehinds and added a couple of subscopes. Also, there's a test file for the first time.

### DIFF
--- a/AppleScript/AppleScript.sublime-syntax
+++ b/AppleScript/AppleScript.sublime-syntax
@@ -118,40 +118,24 @@ contexts:
     - match: ^\s*(considering)\b
       push:
         - meta_scope: meta.block.considering.applescript
+        - meta_content_scope: meta.array.attributes.considering.applescript
         - match: ^\s*(end(?:\s+considering)?)(?=\s*(--.*?)?$)
-          pop: true
-        - match: (?<=considering)
-          push:
-            - meta_scope: meta.array.attributes.considering.applescript
-            - include: line-continuations
-            - include: pop-eol
-            - include: attributes.considering-ignoring
-        - match: (?<=ignoring)
-          push:
-            - meta_scope: meta.array.attributes.ignoring.applescript
-            - include: line-continuations
-            - include: pop-eol
-            - include: attributes.considering-ignoring
+          pop: 1
+        - include: line-continuations
+        - include: pop-eol
+        - include: attributes.considering-ignoring
         - match: \bbut\b
           scope: keyword.control.but.applescript
         - include: main
     - match: ^\s*(ignoring)\b
       push:
         - meta_scope: meta.block.ignoring.applescript
+        - meta_content_scope: meta.array.attributes.ignoring.applescript
         - match: ^\s*(end(?:\s+ignoring)?)(?=\s*(--.*?)?$)
-          pop: true
-        - match: (?<=considering)
-          push:
-            - meta_scope: meta.array.attributes.considering.applescript
-            - include: line-continuations
-            - include: pop-eol
-            - include: attributes.considering-ignoring
-        - match: (?<=ignoring)
-          push:
-            - meta_scope: meta.array.attributes.ignoring.applescript
-            - include: line-continuations
-            - include: pop-eol
-            - include: attributes.considering-ignoring
+          pop: 1
+        - include: line-continuations
+        - include: pop-eol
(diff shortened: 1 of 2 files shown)


michaelblyons:
Okay, now I'm doing more work on it.

michaelblyons:
Not a full rebuild from spec and there are definitely some broken pieces, but it's lots better. Has some good building blocks if someone wants to push forward with Apple's crazy blocks.

michaelblyons:
>     * Language constants (as an example, there are many others), are they actually `\b(?i:false)\b` or rather `\b(?:[Ff]alse|FALSE)\b`? Seemed odd at first glance over.

Most of those were inherited. I guess I could check them against the spec, but I did see a line in the spec saying that almost everything is case insensitive.

>     * If you match a group of words, sometimes you use `(one two)`, sometimes `(one\s+two)`.

Most of this is inherited, too. The only part I remember with a space is in the raw Unicode part, which isn't fully spec-conforming yet anyway.

>     * You could make use of:
>       ```yaml
>       variables:
>         leading_wspace: (?:^\s*)
>       ```

I guess I could, but `{{leading_wspace}}` is longer than `^\s*`. 

>     * You can just indent the lines you test a bit, makes writing the assertions on the following lines easier. Like:
>       ```
>         test
>       # ^^^^ scope
>       ```
>         
>       instead of:
>       ```
>       test
>       # <- scope
>       #^^^ scope
>       ```

Good thought. Will change.

michaelblyons:
> * Language constants (as an example, there are many others), are they actually `\b(?i:false)\b` or rather `\b(?:[Ff]alse|FALSE)\b`? Seemed odd at first glance over.

According to the spec, these are actually supposed to be lower case. Now the question: Do I follow the spec, or do I cajole a Mac user to try out mixed case to see if the case-insens version was determined experimentally?

michaelblyons:
I originally set out to reach sregex compatibility. I got a little carried away, but my aim was not to produce a perfect spec-conforming document.

There is specific handling for certain program's arguments in there. And I cannot test as I do not have MacOS.

Review away. 🙂

AppleScript/AppleScript.sublime-syntax
```diff
@@ -12,16 +14,52 @@ file_extensions:
 first_line_match: |-
   (?xi:
     ^ \#! .* \bosascript(?!\s+-l\s+JavaScript)\b  # shebang
```
deathaxe, line 16: Maybe introduce `{{shebang_language}}` variable and related shebang context to highlight language name?

related with: https://github.com/sublimehq/Packages/pull/4219

AppleScript/AppleScript.sublime-syntax
```diff
@@ -1021,15 +1134,19 @@ variables:
...
+    | on | onto | out\s+of | over
+    | thru
+    | under
     )\b
   # https://developer.apple.com/library/archive/documentation/AppleScript/Conceptual/AppleScriptLangGuide/reference/ASLR_classes.html#//apple_ref/doc/uid/TP40000983-CH1g-SW8
   unit_type: |-
     \b(?x:
       # Optional length dimensions (cubic cannot be miles/km)
-      (?:(?: square | cubic (?! \s+ kilomet | miles )) \s+)?
+      (?:(?: square | cubic (?!\s+kilomet | miles )) \s+)?
```
michaelblyons, line 1149: I realize I have a bug here, not `\s+`ing before the `miles` exception to cubic units.

AppleScript/AppleScript.sublime-syntax
```diff
@@ -552,25 +553,61 @@ contexts:
         # is / isn't
           is (?: n't |\s+ not )? (?:\s+
           (?: equal (?:\s+ to)?
-          | (?: less | greater )\s+ than (?:\s+ or \s+ equal (?:\s+ to )?)?
-          | in | contained \s+ by
+          | (?: less | greater )\s+ than (?:\s+ or\s+equal (?:\s+ to )?)?
+          | in | contained\s+by
```
michaelblyons, line 557: I spaced out the `\s+` to give space for space-collapsing merge strategies to do the right thing.

deathaxe (approved):
Can't say too much about AppleScript itself, but syntax LGTM!

PR #4161, opened 2025-02-17
[SQL] Add support for MySql LOAD DATA statements

keith-hall:


SQL/MySQL.sublime-syntax
```diff
@@ -622,6 +625,45 @@ contexts:
       scope: keyword.other.dml.sql
       set: maybe-column-name-list
 
+###[ DML LOAD STATEMENTS ]#####################################################
+  dml-load-data:
```
deathaxe, line 629: For the pedantic:

```suggestion
###[ DML LOAD STATEMENTS ]#####################################################

    dml-load-data:
```

SQL/MySQL.sublime-syntax
```diff
@@ -622,6 +625,45 @@ contexts:
...
+    - include: numbers
+    - match: (?=;)
+      pop: 1
+
+  meta-filepath:
+    - meta_content_scope: meta.filepath.mysql
+    - include: immediately-pop
+
+  single-string:
+    - include: strings
```
deathaxe, line 664: Would it make sense to introduce a `string` context right next to `strings` in SQL (Basic) and re-use it here to ensure single string being consumed?


```yml
  string:
    - match: \'
      scope: punctuation.definition.string.begin.sql
      set: inside-single-quoted-string
```

keith-hall: I thought about that, and was a little wary of having a duplicated context whose sole change is to perform a `set` rather than a `push`... Plus on a base syntax, where every inheriting syntax also needs to do such repeating contexts when extending one of them, it doesn't seem ideal - it would be easy to forget. I feel like the current solution solves the problem reasonably enough.

FichteFoll (approved):

deathaxe (approved):

PR #4163, opened 2025-02-18
[Java] Restrict contextual keyword `record`

### LINKED ISSUE #4162: [Java] Method invocation of `record` matched as a keyword
### What happened?

This should be matched as a normal method call, not as `keyword.declaration.record.java`.

```java
meterRegistry.timer("cool metric name").record(20000, TimeUnit.MILLISECONDS);

class Clazz {
    void fn() {
        meterRegistry.timer("cool metric name").record(20000, TimeUnit.MILLISECONDS);
    }
}
```

Image

An inner function call confuses it even more and causes the closing parenthesis to be marked as invalid:
```java
meterRegistry.timer("more metric").record(someFunc(), TimeUnit.MILLISECONDS);
```

Also occurs without chaining:
```java
meterRegistry.record(20000, TimeUnit.MILLISECONDS);
```

deathaxe:
Fixes #4162

This commit avoids contextual keyword `record` being matched in unappropriated context. It is a valid keyword only in record declarations.

A record declaration is treated as valid if it starts with `record ident`, terminated by a `{..}` block. Anything in between is treated optional.

see: https://docs.oracle.com/javase/specs/jls/se23/html/jls-3.html#jls-3.9

### DIFF
--- a/Java/tests/syntax_test_java.java
+++ b/Java/tests/syntax_test_java.java
@@ -2449,37 +2449,33 @@ Bar             // comment
  * https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-8.10
  *****************************************************************************/
 
-record
-// <- meta.class.java keyword.declaration.record.java
-//^^^^ meta.class.java keyword.declaration.record.java
-//    ^ meta.class.identifier.java
+record ;
+// <- - keyword.declaration
+//^^^^ - keyword.declaration
 
-record RecordTest
+record RecordTest ;
+// <- support.class.java
+//^^^^ support.class.java
+//     ^^^^^^^^^^ variable.other.java
+
+record RecordTest<> ;
+// <- support.class.java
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

Java/tests/syntax_test_java.java
```diff
@@ -2570,6 +2566,76 @@ record CompactConstructorTests(int foo) {
 }
 // <- meta.class.java meta.block.java punctuation.section.block.end.java
 
+/*
+ * Record is a `contextual keyword` valid only in record declaration statements.
+ * see: https://docs.oracle.com/javase/specs/jls/se23/html/jls-3.html#jls-3.9
+ */
+
+record record(record record) { record record(record record) { record record = 10 } }
```
michaelblyons, line 2574: 🪆

michaelblyons (approved):

PR #4165, opened 2025-02-19
[SQL] More MySql improvements

keith-hall:
- improve `SELECT INTO` highlighting
- improve `SHOW` highlighting

### DIFF
--- a/SQL/MySQL.sublime-syntax
+++ b/SQL/MySQL.sublime-syntax
@@ -95,6 +95,11 @@ contexts:
     - meta_prepend: true
     - include: rename-statements
     - include: show-statements
+    - match: \b(?i:table)\b
+      scope: keyword.context.mysql
+      push:
+        - table-name
+        - single-identifier
 
 ###[ COMMENTS ]################################################################
 
@@ -603,6 +608,7 @@ contexts:
     - match: \b(?i:replace\s+(?:into)?)\b
       scope: keyword.other.dml.sql
       push: expect-table-name
+    - include: into
     - match: \b(?i:limit)\b
       scope: keyword.other.dml.sql
     - match: \b(?i:on\s+duplicate\s+key)\b
@@ -645,6 +651,9 @@ contexts:
       push:
         - maybe-partition
         - expect-table-name
+    - include: import-export-file-params
+
+  import-export-file-params:
     - match: \b(?i:character\s+set)\b
       scope: storage.modifier.mysql
       push: charset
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):

SQL/MySQL.sublime-syntax
```diff
@@ -941,6 +977,11 @@ contexts:
 
   expressions:
     - meta_prepend: true
+    - match: \b(?i:declare)\b
+      scope: keyword.declaration.mysql
+      push:
+        - expect-type
+        - single-identifier
```
deathaxe, line 984: What's the reason declare not being a `statement`? Is it realy valid within `( ... )` groups etc.?

SQL/MySQL.sublime-syntax
```diff
@@ -95,6 +95,11 @@ contexts:
     - meta_prepend: true
     - include: rename-statements
     - include: show-statements
+    - match: \b(?i:table)\b
+      scope: keyword.context.mysql
+      push:
+        - table-name
+        - single-identifier
```
deathaxe, line 102: To follow other "identifier" use cases, which all ensure scope (e.g. `meta.table-name`) to start with first non-whitespace character, I'd suggest the following change:

```suggestion
      push: expect-table-name
```

SQL/MySQL.sublime-syntax
keith-hall: well spotted, thanks, this was a mistake :)

SQL/MySQL.sublime-syntax
keith-hall: nice catch - done :)

SQL/tests/syntax/syntax_test_mysql.sql
```diff
@@ -4919,3 +4951,211 @@ LOAD DATA INFILE 'file.txt'
...
+--                                                   ^ punctuation.section.group.end.sql
+--                                                    ^ punctuation.section.group.end.sql
+--                                                      ^^ keyword.operator.assignment.alias.sql
+--                                                         ^ meta.alias.table.sql
+    INTO OUTFILE '/tmp/select-values.txt';
+
+TABLE employees ORDER BY lname LIMIT 1000
+-- ^^ keyword.context.mysql
+--    ^^^^^^^^^ meta.table-name.sql
+-- ^^^ - meta.table-name
```
deathaxe, line 5074: Just for the sake of perfection ;)

```suggestion
-- ^^^ - meta.table-name
--    ^^^^^^^^^ meta.table-name.sql
```

deathaxe (approved):

jfcherng (approved):

PR #4169, opened 2025-02-23
[Diff] Fix conflict markers

deathaxe:
This PR...

1. restricts merge conflict markers to exactly 7 consecutive chars (e.g. `<<<<<<<` or `>>>>>>>`) according to https://www.gnu.org/software/diffutils/manual/diffutils.html#Generating-the-Merged-Output-Directly

   If less chars, it's not invalid, but just no conflict marker.

2. ensure reference names are separated from markers by at least one whitespace

These changes are required to reduce risk of ambiguities with headings in Markdown and Restructured Text.

This is an attempt to avoid existing syntax tests failing, when merging current master into #4047.

### DIFF
--- a/Diff/Diff.sublime-syntax
+++ b/Diff/Diff.sublime-syntax
@@ -307,10 +307,11 @@ contexts:
 ###############################################################################
 
 variables:
-  conflict_begin: <{7}
-  conflict_end: '>{7}'
-  conflict_split: ={7}
-  conflict_base: \|{7}
+  conflict_marker_size: '{7}'
+  conflict_begin: <{{conflict_marker_size}}
+  conflict_end: '>{{conflict_marker_size}}'
+  conflict_split: ={{conflict_marker_size}}
+  conflict_base: \|{{conflict_marker_size}}
   conflict_any: |-
     (?x:
       {{conflict_begin}}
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

michaelblyons:
Git concurs that conflict marker size is default 7 chars. https://git-scm.com/docs/gitattributes#_conflict_marker_size

Diff/Diff.sublime-syntax
```diff
@@ -307,10 +307,10 @@ contexts:
...
 
 variables:
-  conflict_begin: <{5,}
-  conflict_end: '>{5,}'
-  conflict_split: ={5,}
-  conflict_base: \|{5,}
+  conflict_begin: <{7}
+  conflict_end: '>{7}'
+  conflict_split: ={7}
+  conflict_base: \|{7}
```
michaelblyons, line 313: Is it feasible to do `conflict_marker_size: '{7}'` and then `conflict_begin: <{{conflict_marker_size}}` etc.?

Sorry I'm AFK and cannot test. I would approve this as-is, but wanted to suggest.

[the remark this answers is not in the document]
michaelblyons: Sorry double post

deathaxe: Applied.

michaelblyons (approved):

jfcherng (approved):

deathaxe:
Thanks!

PR #4171, opened 2025-02-24
[Git Config] Some scope tweaks

deathaxe:
This PR...

1. simplifies `meta.section` scope, by only applying it to `[...]` lines
2. changes section bracket scopes to `punctuation.section.brackets` to comply with scope naming guidelines
3. changes key scopes to `meta.mapping.key string.unquoted`

Note: Sub-sections are also scoped `entity.name.section string`, which is something we are facing with TOML as well. Luckily `key` question is easier here as only unqualified unquoted keys are allowed.

### DIFF
--- a/Git Formats/Git Config.sublime-syntax
+++ b/Git Formats/Git Config.sublime-syntax
@@ -33,17 +33,17 @@ contexts:
     # Note that this does not match [color] (without the subsection). This is
     # because [color] doesn't actually take color values.
     - match: ^\s*(?=\[\s*color\s+\")
-      push: [key-color-pair, section-header]
+      push: [key-color-pair, section-header-meta, section-header]
     # section-other matches all sections except [color "subsection"].
     - match: ^\s*(?=\[)
-      push: [key-value-pair, section-header]
+      push: [key-value-pair, section-header-meta, section-header]
 
 ###[ SECTION HEADERS ]#########################################################
 
   section-header:
     - match: \[
       scope: punctuation.definition.brackets.begin.git.config
-      set: [section-header-meta, section-header-end, section-name]
+      set: [section-header-end, section-name]
 
   section-header-meta:
     - meta_scope: meta.section.git.config
--- a/Git Formats/Git Config.sublime-syntax
+++ b/Git Formats/Git Config.sublime-syntax
@@ -17,7 +17,7 @@ hidden_file_extensions:
 first_line_match: ^\[core\]  # .git/config files always start with [core]
 
 variables:
-  variable_name: '[a-zA-Z][\w-]*'
+  key_name: '[a-zA-Z][\w-]*'
   zero_to_255: 25[0-5]|2[0-4][0-9]|1\d\d|[1-9][0-9]|[0-9]
 
 contexts:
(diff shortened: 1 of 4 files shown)


jfcherng (dismissed):

Git Formats/Git Config.sublime-syntax
```diff
@@ -115,11 +115,10 @@ contexts:
   # changed = red
   # untracked = bold green
   key-color-pair:
-    - meta_scope: meta.section.body.git.config
-    - match: ^(\s*)({{variable_name}})(\s*(\=)\s*)
+    - match: ^(\s*)({{key_name}})(\s*(=)\s*)
       captures:
         1: meta.mapping.git.config
-        2: meta.mapping.key.git.config variable.other.readwrite.git.config
+        2: meta.mapping.key.git.config string.unquoted.git.config
```
michaelblyons, line 121: I really find this to be a negative change. I could tweak my color scheme to compensate, but it's going to be a downside for most people.

Git Formats/Git Config.sublime-syntax
```diff
@@ -130,11 +129,10 @@ contexts:
 
   # key = val
   key-value-pair:
-    - meta_scope: meta.section.body.git.config
-    - match: ^(\s*)({{variable_name}})(\s*(\=)\s*)
+    - match: ^(\s*)({{key_name}})(\s*(=)\s*)
       captures:
         1: meta.mapping.git.config
-        2: meta.mapping.key.git.config variable.other.readwrite.git.config
+        2: meta.mapping.key.git.config string.unquoted.git.config
```
michaelblyons, line 135: As above.

Git Formats/Git Config.sublime-syntax
```diff
@@ -33,20 +33,20 @@ contexts:
...
   section-header:
     - match: \[
-      scope: punctuation.definition.brackets.begin.git.config
-      set: [section-header-meta, section-header-end, section-name]
+      scope: punctuation.section.brackets.begin.git.config
+      set: [section-header-end, section-name]
 
   section-header-meta:
-    - meta_scope: meta.section.header.git.config
+    - meta_scope: meta.section.git.config
```
michaelblyons, line 49: Needs change to *Symbol List.tmPreferences* to match scope.

deathaxe: Fixed

Git Formats/Git Config.sublime-syntax
michaelblyons: I have added an override in my color scheme to make `meta.mapping.key string` a distinct color. We'll see if I hate it in some other context like, say, Python.

Who knows, maybe it'll be better for dict keys there, too.

deathaxe: That's a general issue JSON and YAML suffer from as well.

see: https://github.com/sublimehq/sublime_text/issues/3590

michaelblyons: Yeah. I forgot I used to have an override on the old computer.

michaelblyons (approved):

PR #4175, opened 2025-02-25
[meta] fix gitignore & file display in browser

jrappen:
when viewing files on GitHub.com

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,14 @@
+*                           text=auto
+
+*.sublime-build             linguist-language=JSON-with-Comments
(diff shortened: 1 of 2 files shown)


michaelblyons:
Don't they have support for sublime jsonc files already?

https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml#L3352

Fun trip down memory lane: https://github.com/github-linguist/linguist/issues/2850

michaelblyons:
Was something broken on GitHub?

Also, were you having to exclude any of those gitignore patterns locally? Those seem fine to add if you are having issues.

PR #4176, opened 2025-02-25
[R] refactor of syntax files

jrappen:
- bump v2
- add section headers
- add prototype contexts
- move vars to EOF
- add named contexts for numbers and constants

Co-authored-by: Michael B. Lyons <michaelblyons@users.noreply.github.com>
Co-authored-by: deathaxe <deathaxe@users.noreply.github.com>

### DIFF
--- a/R/R.sublime-syntax
+++ b/R/R.sublime-syntax
@@ -46,7 +46,6 @@ contexts:
 ###[ COMMENTS ]################################################################
 
   codesection:
-    - meta_include_prototype: false
     - match: ^\s*((\#+)\s+(.+?)\s+(?:-{4,}|={4,}|#{4,})[ \t]*$\n?)
       captures:
         1: comment.line.number-sign.r
@@ -60,9 +59,12 @@ contexts:
   line-comments:
     - match: \#+(?!.*(?:-{4,}|={4,}|#{4,})\s*$\n?)
       scope: punctuation.definition.comment.r
-      push:
-        - meta_scope: comment.line.number-sign.r
-        - include: eol-pop
+      push: line-comment-body
+
+  line-comment-body:
+    - meta_include_prototype: false
+    - meta_scope: comment.line.number-sign.r
+    - include: eol-pop
 
   roxygen:
     - match: (#')\s*(?!.*(?:-{4,}|={4,}|#{4,})\s*$\n?)
@@ -187,6 +189,7 @@ contexts:
       scope: keyword.operator.word.r
 
   accessor:
+    - meta_include_prototype: false
     - match: '\$'
       scope: keyword.accessor.dollar.r
       push:
@@ -253,9 +256,7 @@ contexts:
         2: punctuation.definition.string.begin.r
       push:
         - meta_include_prototype: false
-        - meta_scope: >-
-            meta.string.r
-            string.quoted.double.r
+        - meta_scope: meta.string.r string.quoted.double.r
         - match: \)(\3)"
           scope: punctuation.definition.string.end.r
           pop: 1
@@ -266,9 +267,7 @@ contexts:
         2: punctuation.definition.string.begin.r
       push:
         - meta_include_prototype: false
-        - meta_scope: >-
-            meta.string.r
-            string.quoted.double.r
+        - meta_scope: meta.string.r string.quoted.double.r
         - match: \}(\3)"
           scope: punctuation.definition.string.end.r
           pop: 1
@@ -279,9 +278,7 @@ contexts:
         2: punctuation.definition.string.begin.r
       push:
         - meta_include_prototype: false
-        - meta_scope: >-
-            meta.string.r
-            string.quoted.double.r
+        - meta_scope: meta.string.r string.quoted.double.r
         - match: \](\3)"
           scope: punctuation.definition.string.end.r
           pop: 1
@@ -292,9 +289,7 @@ contexts:
         2: punctuation.definition.string.begin.r
       push:
         - meta_include_prototype: false
-        - meta_scope: >-
-            meta.string.r
-            string.quoted.single.r
+        - meta_scope: meta.string.r string.quoted.single.r
         - match: \)(\3)'
           scope: punctuation.definition.string.end.r
           pop: 1
@@ -305,9 +300,7 @@ contexts:
         2: punctuation.definition.string.begin.r
       push:
         - meta_include_prototype: false
-        - meta_scope: >-
-            meta.string.r
-            string.quoted.single.r
+        - meta_scope: meta.string.r string.quoted.single.r
         - match: \}(\3)'
           scope: punctuation.definition.string.end.r
           pop: 1
@@ -318,9 +311,7 @@ contexts:
         2: punctuation.definition.string.begin.r
       push:
         - meta_include_prototype: false
-        - meta_scope: >-
-            meta.string.r
-            string.quoted.single.r
+        - meta_scope: meta.string.r string.quoted.single.r
         - match: \](\3)'
           scope: punctuation.definition.string.end.r
           pop: 1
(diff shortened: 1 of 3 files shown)


R/R.sublime-syntax
```diff
@@ -41,6 +41,18 @@ contexts:
     - include: function-calls
     - include: general-variables
 
+###[ PROTOTYPES ]##############################################################
```
michaelblyons, line 44: Move prototype to bottom, since it's never very interesting.

michaelblyons (changes requested):

michaelblyons:
Are there any more places to add `meta_include_prototype: false`? Like strings? Otherwise LGTM.

michaelblyons (approved):

R/R.sublime-syntax
```diff
@@ -187,123 +235,242 @@ contexts:
...
+      captures:
+        1: punctuation.definition.string.escape.r
+    - match: '(\\)u\{?\h{1,4}\}?'
+      scope: constant.character.escape.r
+      captures:
+        1: punctuation.definition.string.escape.r
+    - match: '(\\)U\{?\h{1,8}\}?'
+      scope: constant.character.escape.r
+      captures:
+        1: punctuation.definition.string.escape.r
```
deathaxe, line 427: 1. Few syntaxes certainly scope `\\` as `punctuation.definition.escape`. As most of them consist of only 2 characters, it adds up more noise than use. Hence I'd rather vote to remove existing punctuation scope from backslash escapes instead of adding more.
2. It is not useful to add a single pattern for each individual escape sequence. I'd rather suggest to group them. Good examples (also for scoping) is Python.

   https://github.com/sublimehq/Packages/blob/5fa6fb77a536dabaf5abb5e6c6bc309d03708d75/Python/Python.sublime-syntax#L4039-L4055

R/R.sublime-syntax
```diff
@@ -187,123 +235,242 @@ contexts:
...
     - match: (R|r)("(-*)\()
       captures:
         1: storage.type.string.r
         2: punctuation.definition.string.begin.r
       push:
-        - meta_scope: string.quoted.double.r
+        - meta_include_prototype: false
+        - meta_scope: >-
+            meta.string.r
+            string.quoted.double.r
```
deathaxe, line 258: I'd follow FichteFolls arguments to only break up long scope chains into multiple rows, like Bash/Zsh which would otherwise heavily exceed 100 chars. Even then `>-` is not required.

Those simple ones can keep on a single line.

R/R.sublime-syntax
```diff
@@ -41,24 +43,61 @@ contexts:
     - include: function-calls
     - include: general-variables
 
+###[ COMMENTS ]################################################################
+
   codesection:
+    - meta_include_prototype: false
```
deathaxe, line 49: This context is never directly pushed to via `push: codesection`. Thus `meta_include_prototype` is redundant.

R/R.sublime-syntax
```diff
@@ -41,24 +43,61 @@ contexts:
...
+    - include: line-comments
+
+  line-comments:
+    - match: \#+(?!.*(?:-{4,}|={4,}|#{4,})\s*$\n?)
       scope: punctuation.definition.comment.r
       push:
         - meta_scope: comment.line.number-sign.r
-        - match: \n
-          pop: true
+        - include: eol-pop
```
deathaxe, line 65: Maybe add named context `line-comment-body` if we are up to refactor?

R/R.sublime-syntax
```diff
@@ -145,10 +192,11 @@ contexts:
       push:
```
deathaxe, line 192: It feels like now that comments are included via `prototype` you need to exclude them here, cause `$` only ever introduces special function or variable names.

R/R.sublime-syntax
```diff
@@ -187,123 +238,188 @@ contexts:
...
           scope: punctuation.definition.string.end.r
-          pop: true
+          pop: 1
+
+  single-quoted-string-escapes:
+    - meta_include_prototype: false
+    - match: (\\)\'
+      scope: constant.character.escape.r
+      captures:
+        1: punctuation.definition.string.escape.r
```
deathaxe, line 344: Punctuation suggestions also apply here.
```suggestion
    - match: \\\'
      scope: constant.character.escape.r
```

R/R.sublime-syntax
```diff
@@ -187,123 +238,188 @@ contexts:
...
+    - include: string-escapes
+
+  double-quoted-string-escapes:
+    - meta_include_prototype: false
+    - match: (\\)\"
+      scope: constant.character.escape.r
+      captures:
+        1: punctuation.definition.string.escape.r
```
deathaxe, line 352: and here 

```suggestion
    - match: \\\"
      scope: constant.character.escape.r
```

R/R.sublime-syntax
```diff
@@ -187,123 +238,188 @@ contexts:
...
+    - include: string-escapes
+
+  string-escapes:
+    - meta_include_prototype: false
+    - include: invalid-string-escapes
+    - include: valid-string-escapes
+
+  invalid-string-escapes:
+    - meta_include_prototype: false
+    - match: '(\\)0'                                        # null character
```
deathaxe, line 362: remove unused capture group

```suggestion
    - match: '\\0'                                        # null character
```

R/R.sublime-syntax
```diff
@@ -187,123 +238,188 @@ contexts:
...
+  valid-string-escapes:
+    - meta_include_prototype: false
+    - match: '\\[\\nrtbafv]'
+      scope: constant.character.escape.r
+    - match: '\\[0-7]{1,3}'
+      scope: constant.character.escape.octal.r
+    - match: '\\x\h{1,2}'
+      scope: constant.character.escape.hex.r
+    - match: '\\u\{?\h{1,4}\}?'
+      scope: constant.character.escape.unicode.r
```
deathaxe, line 374: With regards to the `true/false/null` debate, maybe enable distinction of unicode escapes as already applied in python and some others?

```suggestion
      scope: constant.character.escape.unicode.16bit.r
```

R/R.sublime-syntax
```diff
@@ -187,123 +238,188 @@ contexts:
...
+    - match: '\\U\{?\h{1,8}\}?'
+      scope: constant.character.escape.unicode.r
```
deathaxe, line 376:
```suggestion
      scope: constant.character.escape.unicode.32bit.r
```

R/R.sublime-syntax
```diff
@@ -187,123 +238,184 @@ contexts:
...
+    - match: '\\[\\nrtbafv]'
+      scope: constant.character.escape.r
+    - match: '\\[0-7]{1,3}'
+      scope: constant.character.escape.octal.r
+    - match: '\\x\h{1,2}'
+      scope: constant.character.escape.hex.r
+    - match: '\\u\{?\h{1,4}\}?'
+      scope: constant.character.escape.unicode.16bit.r
+    - match: '\\U\{?\h{1,8}\}?'
+      scope: constant.character.escape.unicode.32bit.r
```
deathaxe, line 372: Sure we need dedicated contexts for valid and invalid escapes?

It's probably rather unlikely an inheriting syntax wants to specifically only replace `\\0`. Even if, it still could to by prepending patterns to `string-escapes` context.

This separation doesn't hurt beyond a little overhead, but also doesn't add much benefit.

```suggestion
  string-escapes:
    - match: '\\[\\nrtbafv]'
      scope: constant.character.escape.r
    - match: '\\[0-7]{1,3}'
      scope: constant.character.escape.octal.r
    - match: '\\x\h{1,2}'
      scope: constant.character.escape.hex.r
    - match: '\\u\{?\h{1,4}\}?'
      scope: constant.character.escape.unicode.16bit.r
    - match: '\\U\{?\h{1,8}\}?'
      scope: constant.character.escape.unicode.32bit.r
    - match: '\\0'
      scope: invalid.illegal.constant-character-escape.r
```

deathaxe (commented):
Just realized 2 more little things, otherwise LGTM!

deathaxe (approved):

michaelblyons (approved):

PR #4182, opened 2025-03-03
[Rust] add enum members to indexed symbol list

keith-hall:
[Rust] add enum members to indexed symbol list to support Goto Definition etc.

### DIFF
--- a/Rust/Symbol List.tmPreferences
+++ b/Rust/Symbol List.tmPreferences
@@ -10,7 +10,7 @@
 		entity.name.struct.rust,
 		entity.name.trait.rust,
 		entity.name.union.rust,
-        meta.enum.rust storage.type - meta.struct
+		meta.enum.rust storage.type - meta.struct
 	</string>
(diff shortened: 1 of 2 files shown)


jfcherng (approved):

Rust/Symbol List.tmPreferences
```diff
@@ -9,7 +9,8 @@
 		entity.name.module.rust,
 		entity.name.struct.rust,
 		entity.name.trait.rust,
-		entity.name.union.rust
+		entity.name.union.rust,
+        meta.enum.rust storage.type - meta.struct
```
FichteFoll, line 13: This line uses spaces for indentation while the rest of the file uses tabs.

Can you also add a "negative" test for the `meta.struct` case?

keith-hall: fixed and done, thanks :)

FichteFoll (approved):

michaelblyons (approved):

PR #4183, opened 2025-03-03
[SQL] support comments in more places and MySQL LATERAL derived tables

### LINKED ISSUE #4181: [SQL] Comment highlighting is broken in "left join lateral", array closing bracket is highlighted as invalid
### What happened?

Heya! With the SQL syntax rewrite, with the following SQL query, I get weird highlighting.

The query:

```sql
-- comment one
select
    col1,
    (col_array)[1]
-- comment two
from table1
-- comment three -- left join lateral
left join lateral (
    -- comment four
    select col2
    -- comment five
    from table2
    -- comment six
    where col2 = 'string'
    and col3 = 123
    -- comment seven
    and col4 = '123'
    -- comment eight
    and col5 = 123
    -- seems to toggle the commenting when using
    -- certain keywords
    group by col6
    -- comment nine with single quote: don't use the wrong columns
    order by col7
)
```

The highlighting:

<details open><summary>Screenshot 1</summary>

Image

</details>

The issue seems to be `lateral` is interpreted as a function call, and I guess it changes the "mode" that tokens are classified.

If `lateral` is taken away, the highlighting is much better, but still has some quirks -- line 23 should be interpreted as a comment, but isn't:

<details open><summary>Screenshot 2</summary>

Image

</details>

There's also the issue with the right square bracket `]` being highlighted as invalid, though that one's not as annoying since it doesn't affect subsequent tokens.

I'm using Sublime Text 4195 (dev) on Mac OS Sequoia 15.2.

keith-hall:
Fixes #4181

### DIFF
--- a/PHP/Embeddings/SQL (for PHP Interpolated).sublime-syntax
+++ b/PHP/Embeddings/SQL (for PHP Interpolated).sublime-syntax
@@ -17,25 +17,11 @@ contexts:
     - match: \)
       scope: punctuation.section.group.end.sql
 
-  single-identifier:
-    - meta_prepend: true
-    - meta_include_prototype: false
-    - include: Packages/PHP/PHP Source.sublime-syntax#interpolation
-
-  inside-like-single-quoted-string:
-    - meta_prepend: true
-    - include: string-interpolation
-
-  inside-single-quoted-string:
-    - meta_prepend: true
-    - include: string-interpolation
-
-  string-interpolation:
-    # this context is included in anonymous string contexts
-    - meta_prepend: true
-    - include: Packages/PHP/PHP Source.sublime-syntax#string-interpolations
-
 ###[ COMMENTS ]################################################################
+
+  inside-number-sign-comment:
+    - meta_append: true
+    - include: Packages/PHP/PHP Source.sublime-syntax#interpolations
 
   inside-double-dash-comment:
     - meta_append: true
@@ -48,3 +34,18 @@ contexts:
   inside-comment-block:
     - meta_append: true
     - include: Packages/PHP/PHP Source.sublime-syntax#interpolations
+
+###[ LITERALS ]################################################################
+
+  inside-like-single-quoted-string:
+    - meta_prepend: true
+    - include: Packages/PHP/PHP Source.sublime-syntax#string-interpolations
+
+  inside-single-quoted-string:
+    - meta_prepend: true
+    - include: Packages/PHP/PHP Source.sublime-syntax#string-interpolations
+
+  string-interpolations:
+    # this context is included in anonymous string contexts
+    - meta_prepend: true
+    - include: Packages/PHP/PHP Source.sublime-syntax#string-interpolations
--- a/PHP/Embeddings/SQL (for PHP Interpolated).sublime-syntax
+++ b/PHP/Embeddings/SQL (for PHP Interpolated).sublime-syntax
@@ -34,3 +34,17 @@ contexts:
     # this context is included in anonymous string contexts
     - meta_prepend: true
     - include: Packages/PHP/PHP Source.sublime-syntax#string-interpolations
+
+###[ COMMENTS ]################################################################
+
+  inside-double-dash-comment:
+    - meta_append: true
+    - include: Packages/PHP/PHP Source.sublime-syntax#interpolations
+
+  inside-comment-docblock:
+    - meta_append: true
+    - include: Packages/PHP/PHP Source.sublime-syntax#interpolations
+
+  inside-comment-block:
+    - meta_append: true
+    - include: Packages/PHP/PHP Source.sublime-syntax#interpolations
--- a/Ruby/Embeddings/SQL (for Ruby).sublime-syntax
+++ b/Ruby/Embeddings/SQL (for Ruby).sublime-syntax
@@ -12,3 +12,7 @@ contexts:
     - meta_prepend: true
     - include: Ruby.sublime-syntax#escaped-char
     - include: Ruby.sublime-syntax#embedded-ruby
+
+  comments:
+    - include: Ruby.sublime-syntax#escaped-char
+    - include: Ruby.sublime-syntax#embedded-ruby
--- a/Ruby/Embeddings/SQL (for Ruby).sublime-syntax
+++ b/Ruby/Embeddings/SQL (for Ruby).sublime-syntax
@@ -12,10 +12,24 @@ contexts:
     - meta_prepend: true
     - include: embedded-ruby
 
+  embedded-ruby:
+    - include: Ruby.sublime-syntax#escaped-char
+    - include: Ruby.sublime-syntax#embedded-ruby
+
+###[ COMMENTS ]################################################################
+
   number-sign-comments:
     - meta_prepend: true
     - include: embedded-ruby
 
-  embedded-ruby:
-    - include: Ruby.sublime-syntax#escaped-char
-    - include: Ruby.sublime-syntax#embedded-ruby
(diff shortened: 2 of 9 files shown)


Ruby/Embeddings/SQL (for Ruby).sublime-syntax
```diff
@@ -12,3 +12,7 @@ contexts:
     - meta_prepend: true
     - include: Ruby.sublime-syntax#escaped-char
     - include: Ruby.sublime-syntax#embedded-ruby
+
+  comments:
+    - include: Ruby.sublime-syntax#escaped-char
+    - include: Ruby.sublime-syntax#embedded-ruby
```
deathaxe, line 18: Are you sure about this one? 

`comments` is included in `prototype` and replacing all SQL comments may not have intentional effects.

Maybe we should think about removing `- meta_include_prototype: false` from comment contexts in SQL (Basic) instead, so `prototype` and thus ruby extensions are applied normally. 

Alternatively prepend those rules to
- inside-double-dash-comment
- inside-comment-docblock
- inside-comment-block

keith-hall: You're absolutely right that there is a mistake here - I thought I had done a `meta_append`...

I like your alternative suggestions. I will be able to take a look at fixing it tomorrow night. If anyone is impatient and want to work on it before then, be my guest 🙂

Ruby/Embeddings/SQL (for Ruby).sublime-syntax
```diff
@@ -10,5 +10,12 @@ contexts:
 
   prototype:
     - meta_prepend: true
+    - include: embedded-ruby
+
+  number-sign-comments:
```
deathaxe, line 15: `number-sign-comments` is included into `comments` which is included in `prototype` and thus prepanding `embedded-ruby` here is of little use beyond just adding duplicate to `prototype`.

You probably want to...

```suggestion
  inside-number-sign-comment:
```

if not removing `- meta_include_prototype: false` from MySQL as for SQL (Basic).

Maybe also add a test to verify interpolation working as expected?

PHP/Embeddings/SQL (for PHP Interpolated).sublime-syntax
```diff
@@ -34,3 +34,8 @@ contexts:
     # this context is included in anonymous string contexts
     - meta_prepend: true
     - include: Packages/PHP/PHP Source.sublime-syntax#string-interpolations
+
+  table-name-or-subquery:
+    - meta_include_prototype: false
+    - meta_prepend: true
+    - include: comments
```
deathaxe, line 41: In case `meta_include_prototype` is required to correctly handle interpolation, we should probably (keep) it in core SQL syntax. Otherwise extending syntaxes need to add too many special-case contesxts like this one, increasing dependency surface.

Ruby/Embeddings/SQL (for Ruby).sublime-syntax
keith-hall: I think the problem here is that Ruby uses the `#` character to start an interpolation, which is the start of line comment char in some SQL dialects. So adding support for interpolations when already inside the comment is too late.

`number-sign-comments` is included into `comments`, which is part of `prototype` but as we need `meta_include_prototype: false` in some places, and explicitly include the `comments` context instead, the interpolations need to specifically be prepended before `#` is matched as a SQL comment char.

keith-hall: The existing Ruby and PHP interpolation tests seem to be enough to catch these cases.

PHP/Embeddings/SQL (for PHP Interpolated).sublime-syntax
keith-hall: the `meta_include_prototype` for `table-name-or-subquery`  is now back in the core/basic SQL syntax :+1:

Ruby/Embeddings/SQL (for Ruby).sublime-syntax
deathaxe: I haven't spotted any, which tests for interpolation within SQL comments, yet.

### Ruby

```
-- id: #{id}
```

### PHP


```
-- id: $id
```

Ruby/Embeddings/SQL (for Ruby).sublime-syntax
```diff
@@ -10,5 +10,26 @@ contexts:
...
     - meta_prepend: true
+    - include: embedded-ruby
+
+  embedded-ruby:
     - include: Ruby.sublime-syntax#escaped-char
     - include: Ruby.sublime-syntax#embedded-ruby
+
+###[ COMMENTS ]################################################################
+
+  number-sign-comments:
```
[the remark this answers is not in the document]
keith-hall: we still need the `number-sign-comments` so that the Ruby interpolations take priority over `#` comments in SQL, but I have added the `inside-number-sign-comment`as well now :)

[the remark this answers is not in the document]
deathaxe: I see.  To avoid duplicate includes, I'd suggest one of:

1. ensure to match comments only if they don't look like an interpolation

```yml
  number-sign-comments:
    - match: \#(?![{@$])
      scope: punctuation.definition.comment.sql
      push: inside-number-sign-comment
```

2. prepend `comments` to catch all dialects and **remove** `prototype` as comments are included in prototype.

```yml
  comments:
    # Notes:
    # 1. comments are included in `prototype` thus apply everywhere
    # 2. prepend patterns here to ensure Ruby interpolation `#{...}` takes
    #    precedence over any kind of comment (e.g. number-sighn comments).
    - meta_prepend: true
    - include: embedded-ruby
```

Ruby/syntax_test_ruby.rb
```diff
@@ -282,6 +282,13 @@ def #{sym}(*args, &block)
 #^^^^^^^^^^^^^ meta.string.heredoc.ruby source.sql.embedded.ruby - meta.interpolation
 #             ^^^^^^^ meta.string.heredoc.ruby source.sql.embedded.ruby meta.interpolation.ruby variable.other.readwrite.global
 #                    ^ meta.string.heredoc.ruby source.sql.embedded.ruby - meta.interpolation
+-- id: #{id}
+#^^^^^^^^^^^ meta.string.heredoc.ruby source.sql.embedded.ruby comment.line.double-dash.sql
+#^ punctuation.definition.comment.sql
+#      ^^^^^ meta.interpolation.ruby
+#      ^^ punctuation.section.interpolation.begin.ruby
+#        ^^ source.ruby.embedded.ruby
+#          ^ punctuation.section.interpolation.end.ruby
```
deathaxe, line 291: Maybe also verify `#{..}` takes precedence over `#` comments?

```suggestion
#          ^ punctuation.section.interpolation.end.ruby
#{no_comment}
# <- meta.string.heredoc.ruby source.sql.embedded.ruby meta.interpolation.ruby punctuation.section.interpolation.begin.ruby
#^^^^^^^^^^^^ meta.string.heredoc.ruby source.sql.embedded.ruby meta.interpolation.ruby
#^ punctuation.section.interpolation.begin.ruby
# ^^^^^^^^^^ source.ruby.embedded.ruby
#           ^ punctuation.section.interpolation.end.ruby
```

deathaxe:
Some proposals can be found at https://github.com/forkeith/Packages/pull/32

deathaxe (approved):

michaelblyons (approved):

PR #4185, opened 2025-03-04
[Python] Fix completions in f-string replacement

### LINKED ISSUE #4184: [Python] Missing autocompletions in format strings
### Expected behavior

Using Sublime Text Build 4192

Autocomple suggestions within the replacement fields of a format string should be the same as they are outside the format string.  In the example below, I expect that typing "nu" after "func(" will suggest "num" inside the format string replacement field.

```python
num = 1

def func(x):
	return x

func(num) 
# autocompletion for func and num worked

s = f'{func(num)} ' 
# autocompletion for func worked but num was not shown as an autocomplete suggestion
```

Similarly:

```python
key = ''
dictionary = {key: None}

dictionary[key]
# autocompletion for dictionary and key worked

s2 = f'{dictionary[key]}'
# autocompletion for dictionary worked but key was not shown as an autocomplete suggestion

```

### Actual behavior

Autocompletions for python variable names are not shown when a function is called or when a dictionary is accessed within a format string.

### Steps to reproduce

1. Type the python code above into a python file in Sublime Text.
2. Note which autocompletion suggestions show up when typing the format strings.

deathaxe:
fixes #4184

This commit removes `source.python.embedded` scope from f-string replacements as this causes ST to treat it as external syntax, which prevents word-completions of e.g. variables from being suggested.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -2979,7 +2979,6 @@ contexts:
         - triple-double-quoted-f-string-replacement-expression
 
   triple-double-quoted-f-string-replacement-expression:
-    - meta_content_scope: source.python.embedded
     - include: f-string-replacement-expression
 
   triple-double-quoted-f-string-replacement-formatspec:
@@ -3324,7 +3323,6 @@ contexts:
         - double-quoted-f-string-replacement-expression
 
   double-quoted-f-string-replacement-expression:
-    - meta_content_scope: source.python.embedded
     - include: f-string-replacement-expression
     - include: eol-pop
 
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):

jfcherng (approved):

FichteFoll:
Any suggesttions for my color scheme rule to highlight the background of embedded and interpolated sources that does not include the punctuation? Previously, I was able to simply use `source source` and catch Python f-strings. Now I assume I will have to use `meta.interpolation - punctuation.section.interpolation`, which is fine by me but just checking if we have this use case on the radar and agree that this is the proper way forward.

deathaxe:
As only few (none I am aware of) syntaxes add separate source.<name>.embedded for interpolation with their own syntax, meta.interpolation - punctuation is probably the most widely working approach.

You may however handle nested f-strings somehow to apply background without interruptions.

PR #4190, opened 2025-03-10
[C#] fix null coalescing operator followed by a space

### LINKED ISSUE #4187: [C#] Space after a null-conditional element access operator causes rendering issues
### What happened?

Test code:

```cs
using System;
using System.Xml;

public class Program {
	public static int Main (string [] args)
	{
		var xml = new XmlDocument ();
		var outcome = xml.SelectSingleNode ("/whatever")?.Attributes? ["nope"];
		if (outcome is not null) {
			Console.WriteLine ("true");
		} else {
			Console.WriteLine ("false");
		}
		return 0;
	}
}
```

renders like this:

Image

Note the `else` word is rendered as a syntax error.

This is valid C#: https://dotnetfiddle.net/Q7akvD.

Removing the space after `Attributes?` makes it render correctly:

```cs
using System;
using System.Xml;

public class Program {
	public static int Main (string [] args)
	{
		var xml = new XmlDocument ();
		var outcome = xml.SelectSingleNode ("/whatever")?.Attributes?["nope"];
		if (outcome is not null) {
			Console.WriteLine ("true");
		} else {
			Console.WriteLine ("false");
		}
		return 0;
	}
}
```

keith-hall:
Fixes #4187

### DIFF
--- a/C#/tests/syntax_test_Operators.cs
+++ b/C#/tests/syntax_test_Operators.cs
@@ -12,7 +12,7 @@ Customer? first = customers?[0];
 ///       ^^^^^ variable.other
 ///             ^ keyword.operator.assignment.variable
 ///               ^^^^^^^^^ variable.other
-///                        ^ keyword.operator.null-coalescing
+///                        ^ keyword.operator.null-coalescing - meta.brackets
 ///                         ^^^ meta.brackets
 ///                         ^ punctuation.section.brackets.begin
 ///                           ^ punctuation.section.brackets.end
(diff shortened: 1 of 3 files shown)


C#/tests/syntax_test_Operators.cs
```diff
@@ -9,8 +9,12 @@
 Customer? first = customers?[0];
 /// <- support.type
 ///     ^ storage.type.nullable
-///                        ^^^^ meta.brackets
-///                        ^^ punctuation.section.brackets.begin
+///       ^^^^^ variable.other
+///             ^ keyword.operator.assignment.variable
+///               ^^^^^^^^^ variable.other
+///                        ^ keyword.operator.null-coalescing
```
michaelblyons, line 15: Should there be a `- meta.brackets` here, or does that make things more complicated?

keith-hall: good idea, added :slightly_smiling_face:

michaelblyons (approved):

jfcherng (approved):

PR #4194, opened 2025-03-20
[Lua] Highlight strings used by string.{format,gsub,pack,etc}

mbartlett21:
This PR adds highlighting to the strings that are passed to `string.{format,find,match,gsub,gmatch,pack,unpack,packsize}`.

They are split into three categories:

- Format strings [1].
  These are done similarly to C's `sprintf` and friends, except that Lua only supports some specifiers and adds `q` as an option.

- Patterns [2].
  They are passed as the second argument to the four functions that take patterns, meaning that I can also match on `(whatever_value):match 'this%s*that'`. For scopes, I have roughly followed what the builtin regular expressions package does.
  I haven't worked on nested groups as that got quite hard with handling it inside the three different types of strings that Lua provides.
  The special capture `()` returns a position. I'm not sure what scope this should be. I've set it to `punctuation.section.brackets.lua` for the moment.

- Pack format strings [3].
  The alignment and endian specifiers I have scoped as `storage.modifier` and I have scoped the actual type designations as `storage.type`
  For the padding that can be specified with `x` or `X`, I've scoped it to `punctuation.separator.padding` as I didn't see any better one, and that helped differentiate it from ones that have values.

Since format strings are always the first argument, I did not find a way of handling them when the function is called as a method. (e.g. `('%3.5f'):format(3)`).

[1]\: https://www.lua.org/manual/5.4/manual.html#pdf-string.format
[2]\: https://www.lua.org/manual/5.4/manual.html#6.4.1
[3]\: https://www.lua.org/manual/5.4/manual.html#6.4.2

### DIFF
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -465,7 +465,11 @@ contexts:
       push: expression-begin
 
   accessor:
-    - match: '[.:]'
+    - match: ':'
+      scope: punctuation.accessor.lua
+      push: property-method
+
+    - match: '\.'
       scope: punctuation.accessor.lua
       push: property
 
@@ -498,6 +502,18 @@ contexts:
       pop: 1
     - include: reserved-word-pop
     - include: else-pop
+
+  property-method:
+    - match: '(?:find|gmatch|gsub|match){{identifier_break}}'
+      scope: meta.property.lua meta.function-call.lua variable.function.lua
+      set:
+        - match: (?={{function_args_begin}})
+          pop: 1
+          push:
+            - function-arguments-meta
+            - function-arguments-pattern-at1
+        - include: else-pop
+    - include: property
 
   function-arguments-meta:
     - meta_scope: meta.function-call.arguments.lua
@@ -757,6 +773,142 @@ contexts:
     - meta_scope: meta.function.lua
     - include: immediately-pop
 
+  function-arguments-pattern-at2:
+    - include: string
+    - include: table-constructor
+
+    - match: \(
+      scope: punctuation.section.group.begin.lua
+      set:
+        - meta_scope: meta.group.lua
+        - match: \)
+          scope: punctuation.section.group.end.lua
+          pop: 1
+        - include: reserved-word-expression-pop
+        - match: (?=\S)
+          push:
+            - expression-list-end-pattern-next
+            - expression-begin
+
+  function-arguments-pattern-at1:
+    - include: pattern-string
+    - include: table-constructor
+
+    - match: \(
+      scope: punctuation.section.group.begin.lua
+      set:
+        - meta_scope: meta.group.lua
+        - match: \)
+          scope: punctuation.section.group.end.lua
+          pop: 1
+        - include: reserved-word-expression-pop
+        - match: (?=\S)
+          push:
+            - expression-list-end
+            - pattern-expression
+
+  pattern-expression:
+    - include: pattern-string
+    - include: expression-begin
+
+  expression-list-end-pattern-next:
+    - match: ','
+      scope: punctuation.separator.comma.lua
+      set:
+        - expression-list-end
+        - pattern-expression
+    - include: expression-end
+
+  pattern-nonnested:
+    - match: '\('
+      scope: punctuation.section.group.begin.lua
+    - match: '\)'
+      scope: punctuation.section.group.end.lua
+    - match: '(\[)(\^)?\]?'
+      captures:
+        1: punctuation.definition.set.begin.lua
+        2: keyword.operator.logical.lua
+    - match: '\]'
+      scope: punctuation.definition.set.end.lua
+    - match: '\(\)'
+      scope: punctuation.section.brackets.lua
+    - match: '[+*?\-]'
+      scope: keyword.operator.quantifier.lua
+    - match: \.
+      scope: keyword.other.any.lua
+
+  pattern-string:
+    - match: \[(=*)\[
+      scope: punctuation.definition.string.begin.lua
+      set:
+        - pattern-multiline-string-body
+        - pattern-test-first-caret
+    - match: \'
+      scope: punctuation.definition.string.begin.lua
+      set:
+        - pattern-single-quoted-string-body
+        - pattern-test-first-caret
+    - match: \"
+      scope: punctuation.definition.string.begin.lua
+      set:
+        - pattern-double-quoted-string-body
+        - pattern-test-first-caret
+
+  pattern-test-first-caret:
+    - match: '\^'
+      scope: keyword.control.anchor.lua
+      pop: 1
+    - include: immediately-pop
+
+  pattern-single-quoted-string-body:
+    - meta_include_prototype: false
+    - meta_scope: meta.string.lua string.quoted.single.lua
+    - include: single-quoted-string-body
+    - match: \$(?=\')
+      scope: keyword.control.anchor.lua
+    - match: '%'
+      scope: constant.character.escape.lua
+      push:
+        - match: (?=[\\\'])
+          pop: 1
+        - match: .
+          scope: constant.character.escape.lua
+          pop: 1
+    - include: pattern-nonnested
+
+  pattern-double-quoted-string-body:
+    - meta_include_prototype: false
+    - meta_scope: meta.string.lua string.quoted.double.lua
+    - include: double-quoted-string-body
+    - match: \$(?=\")
+      scope: keyword.control.anchor.lua
+    - match: '%'
+      scope: constant.character.escape.lua
+      push:
+        - match: (?=[\\\"])
+          pop: 1
+        - match: .
+          scope: constant.character.escape.lua
+          pop: 1
+    - include: pattern-nonnested
+
+  pattern-multiline-string-body:
+    - meta_include_prototype: false
+    - meta_scope: meta.string.lua string.quoted.multiline.lua
+    - match: (?:(\$)|(%)|)(\]\1\])
+      captures:
+        1: keyword.control.anchor.lua
+        2: constant.character.escape.lua
+        3: punctuation.definition.string.end.lua
+      pop: 1
+    - match: '%'
+      scope: constant.character.escape.lua
+      push:
+        - match: .
+          scope: constant.character.escape.lua
+          pop: 1
+    - include: pattern-nonnested
+
   builtin-modules:
     - match: coroutine{{identifier_break}}
       scope: support.constant.builtin.lua
@@ -800,9 +952,18 @@ contexts:
         - match: \.
           scope: punctuation.accessor.lua
           set:
+            - match: (?:find|gmatch|gsub|match){{identifier_break}}
+              scope: meta.property.lua support.function.builtin.lua
+              set:
+                - match: (?={{function_args_begin}})
+                  pop: 1
+                  push:
+                    - function-arguments-meta
+                    - function-arguments-pattern-at2
+                - include: else-pop
             - match: |-
                 (?x:
-                  byte|char|dump|find|format|gmatch|gsub|len|lower|match|pack
+                  byte|char|dump|format|len|lower|pack
                   |packsize|rep|reverse|sub|unpack|upper
                 ){{identifier_break}}
               scope: meta.property.lua support.function.builtin.lua
--- a/Lua/Lua.sublime-syntax
+++ b/Lua/Lua.sublime-syntax
@@ -824,6 +824,23 @@ contexts:
             - expression-list-end
             - fmtstr-expression
 
+  function-arguments-packstr-at1:
+    - include: packstr-string
+    - include: table-constructor
+
+    - match: \(
+      scope: punctuation.section.group.begin.lua
+      set:
+        - meta_scope: meta.group.lua
+        - match: \)
+          scope: punctuation.section.group.end.lua
+          pop: 1
+        - include: reserved-word-expression-pop
+        - match: (?=\S)
+          push:
+            - expression-list-end
+            - packstr-expression
+
   pattern-expression:
     - include: pattern-string
     - include: expression-begin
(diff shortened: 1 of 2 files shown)


mbartlett21:
@jrappen 

I've added (roughly) the tests from the Regular Expression package and worked on getting the syntax to work with character classes and captures now.

> Changes suggest anchor keywords only match at beginning and end of single-line strings, but anywhere in multi-line strings. Maybe double-check that, I don't know Lua.

This is fixed now so that the anchor only matches at the very start and very end. (Because of how multiline strings work, it has to account for a possible preceding newline)

---
@michaelblyons 
Were you meaning the sections under `pattern-{single-quoted,double-quoted,multiline}-string-body`? I don't think there is any nested ones left in what's been added now.

Lua/Lua.sublime-syntax
```diff
@@ -499,6 +505,18 @@ contexts:
     - include: reserved-word-pop
     - include: else-pop
 
+  property-method:
+    - match: '(?:find|gmatch|gsub|match){{identifier_break}}'
+      scope: meta.property.lua meta.function-call.lua variable.function.lua
+      set:
+        - match: (?={{function_args_begin}})
+          pop: 1
+          push:
```
deathaxe, line 514: This would be the following then:

```suggestion
          set:
```

Lua/Lua.sublime-syntax
```diff
@@ -680,6 +698,38 @@ contexts:
...
+    # unicode can make multiple, but that is a bit hard to handle
+    - match: \\u\{\h+\}
+      scope: constant.character.escape.unicode.lua
+      pop: 1
+
+    # \z turns into nothing, so we don't pop
+    - match: \\z
+      push:
+        - meta_scope: constant.character.escape.whitespace.lua
+        - include: else-pop
```
deathaxe, line 724: Looks like just consumming `\\z` without pushing or popping contexts is enough

```suggestion
    - match: \\z
      scope: constant.character.escape.whitespace.lua
```

mbartlett21: that won't be enough, since it has to consume whitespace, including newlines. and newlines will get caught on the next match if this doesn't have the extra context

mbartlett21:
I've added some lookahead matching on the one line for immediate calls so that it can work with `("string"):{pack,format}`.

Lua/Lua.sublime-syntax
```diff
@@ -757,6 +814,442 @@ contexts:
...
+
+  pattern-unquoted-charclass:
+    - meta_include_prototype: false
+    - include: pattern-shared-charclass
+    - match: \%b
+      scope: keyword.control.balanced.lua
+      push:
+        - pattern-unquoted-balanced-char
+        - pattern-unquoted-balanced-char
+    - match: \%
```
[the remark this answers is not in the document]
mbartlett21: I'm not sure. With the quoted ones, it needs to be in a separate pattern so that the additional prototype can end the string if it is `%'`, but I'll test what happens with an embedded syntax

mbartlett21:
> find a way to get rid of the anonymous contexts five (?!) levels deep

This seems to be to do with `with_prototype`, which I've used in order to deduplicate code between double-quoted and single-quoted strings.

Or are you meaning in the `builtin-modules` pattern structure?

michaelblyons:
Instead of something like

```yml
strings:
  - match: '"'
    scope: punctuation.definition.string.begin
    push:
      - meta_scope: string.quoted.double
      - match: '"'
        scope: punctuation.definition.string.end
        pop: 1
```

Use


```yml
strings:
  - match: '"'
    scope: punctuation.definition.string.begin
    push: string-double-body

string-double-body:
  - meta_scope: string.quoted.double
  - match: '"'
    scope: punctuation.definition.string.end
    pop: 1
```

Lua/Lua.sublime-syntax
```diff
@@ -760,185 +793,756 @@ contexts:
...
+        - pattern-quoted-balanced-char
+        - pattern-quoted-balanced-char
+    - match: \%
+      scope: constant.character.escape.lua
+      push:
+        - meta_include_prototype: false
+        - include: string-escape-pop
+        - match: .
+          scope: constant.character.escape.lua
+          pop: 1
```
michaelblyons, line 963: In the comments about the `pattern-unquoted-charclass`, you say that this needs the `push` to escape if there's a `%'`, but can you not just use the below to do that? I'm not sure I grasp the full implications of `string-escape-pop`.

```yaml
    - match: \%[^'\n]
      scope: constant.character.escape.lua
```

Lua/Lua.sublime-syntax
```diff
@@ -9,82 +9,29 @@ file_extensions:
 
 first_line_match: |-
   (?xi:
-    ^ \#! .* {{shebang_language}}           # shebang
-  | ^ \s* -- .*? -\*- .*? \blua\b .*? -\*-  # editorconfig
+    ^ \#! .* {{first_line_language}}                                            # shebang
```
deathaxe, line 12: `shebalng_language` is what is used in all other syntaxes and so Lua should keep doing so.

Lua/Lua.sublime-syntax
deathaxe: There should certainly be some tests to clarify the situation. With "any non-alphanumeric characters after %" lua spec includes backslash-escaped characters. The following escape sequences are all valid and equal:

```
(''):match('% ')
--          ^^ constant.character.escape.lua

(''):match('%\x20')
--          ^ constant.character.escape.lua
--           ^^^^ constant.character.escape.hexadecimal.lua

(''):match('%\u{32}')
--          ^ constant.character.escape.lua
--           ^^^^^^ constant.character.escape.unicode.lua

```

The following is also valid

```
(''):match('%\n')
--          ^^^ constant.character.escape.lua
```

We could argue about `- match: .` being valid as `[^[:alnum:]]` would be more precise, with regards to unicode, but I am not sure if spec targets unicode or ascii alphanumerics. The latter are already covered by patterns and wouldn't apply here and thus require:

```
    - match: \%(?![[:alnum:]])
```

Not sure if Lua is used in template languages like Jinja which would want to scope `%` escape even when directly followed an interpolation. _I don't any other syntax supports that, currently._

```
(''):match('%{{interpolation}}')
--          ^ constant.character.escape.lua
--           ^^^^^^^^^^^^^^^^^ meta.string.lua meta.interpolation.lua
```

With regards to quoted patterns following tests would also be useful

```
(''):match('%')
--          ^ constant.character.escape.lua
--           ^ punctuation.definition.string.end.lua

(''):match('%"')
--          ^^ constant.character.escape.lua
--            ^ punctuation.definition.string.end.lua

(''):match("%")
--          ^ constant.character.escape.lua
--           ^ punctuation.definition.string.end.lua

(''):match("%'")
--          ^^ constant.character.escape.lua
--            ^ punctuation.definition.string.end.lua
```

mbartlett21: That dot there is pretty much just matching anything except for:
- alphabetic (acdglpsuwxz are character classes, f is a frontier pattern, b is a balanced pattern, and the others are treated as invalid escapes)
- numeric (a backreference to that capture)
- $ or ^ (a simple escape, matched in one match to prevent them from being classed as special characters)
- backslash escapes (they are done through string-escape-pop.) (which also handles the case of \z)

It could probably be replaced with `- match: \%(?:[^\w\\'"])?` which would do approximately the same.

The escaping with \x etc isn't actually that useful, as `%` is only necessary to escape the magic characters `^$()%.[]*+-?`.

Which if you are happy with it highlighting other combinations of `%`, could be done with
`- match: '\%(?:[$^()%>\]\[*+\-?])?'`
(with a bit of extra handling for the end of long strings and also making other %-character combinations invalid escapes).

mbartlett21: I've changed the implementation to just allow any character there and simplified it a bit.
For the unquoted/long strings, it works just matching directly, because of it being an embedded syntax. For the quoted strings, matching with ' or " still has to be done over multiple matches to allow it to escape the string properly.

Lua/Lua.sublime-syntax
```diff
@@ -757,6 +814,442 @@ contexts:
...
+      scope: storage.modifier.lua
+
+    - match: '[bBhHlLjJTfdnz]|[Iis][0-9]*|c[0-9]+'
+      scope: storage.type.lua
+
+    - match: '[xX]'
+      scope: punctuation.separator.padding.lua
+
+    - match: (?:[^\\\"\'\]\s])
+      scope: invalid.illegal.invalid-storage.lua
```
deathaxe, line 1238: Previous comments apply here as well.

```suggestion
  packstr-string:
    - match: \[(=*)\[
      scope: punctuation.definition.string.begin.lua
      set: multiline-packstr-body
    - match: \'
      scope: punctuation.definition.string.begin.lua
      set: single-quoted-packstr-body
    - match: \"
      scope: punctuation.definition.string.begin.lua
      set: double-quoted-packstr-body

  single-quoted-packstr-body:
    - meta_include_prototype: false
    - meta_scope: meta.string.lua string.quoted.single.lua
    - include: single-quoted-string-body
    - include: packstr-patterns

  double-quoted-packstr-body:
    - meta_include_prototype: false
    - meta_scope: meta.string.lua string.quoted.double.lua
    - include: double-quoted-string-body
    - include: packstr-patterns

  multiline-packstr-body:
    - meta_include_prototype: false
    - meta_scope: meta.string.lua string.quoted.multiline.lua
    - include: multiline-string-body
    - include: packstr-patterns

  packstr-patterns:
    - match: '[=<>]|![0-9]*'
      scope: storage.modifier.lua
    - match: '[bBhHlLjJTfdnz]|[Iis][0-9]*|c[0-9]+'
      scope: storage.type.lua
    - match: '[xX]'
      scope: punctuation.separator.padding.lua
    - match: '[^"''\\\]\s]'
      scope: invalid.illegal.invalid-storage.lua
```

Lua/Lua.sublime-syntax
```diff
@@ -757,6 +814,442 @@ contexts:
...
+    - match: .
+      pop: 1
+
+  # this gets included in expressions that want to have a pattern string
+  pattern-string:
+    - match: \[(=*)\[
+      scope: punctuation.definition.string.begin.lua
+      set:
+        - pattern-multiline-string-body
+        - pattern-test-first-caret-multiline
```
deathaxe, line 1085: To reliably escape on proper amount of `=` chars, embed in combo with pop could be used here:

```suggestion
    - match: \[(=*)\[
      scope: meta.string.lua string.quoted.multiline.lua punctuation.definition.string.begin.lua
      embed: pattern-multiline-string-body
      embed_scope: meta.string.lua string.quoted.multiline.lua
      escape: \]\1\]
      escape_captures:
        0: meta.string.lua string.quoted.multiline.lua punctuation.definition.string.end.lua
      pop: 1
```

Lua/Lua.sublime-syntax
```diff
@@ -757,6 +814,442 @@ contexts:
...
+    # in order to work with the multiline below
+    - match: (?=[^\^])
+      pop: 1
+
+  pattern-test-first-caret-multiline:
+    - meta_include_prototype: false
+    - match: $\n?
+      set: pattern-test-first-caret
+    - include: pattern-test-first-caret
+
```
deathaxe, line 1117: Obsolete with previous change

```suggestion
```

Lua/Lua.sublime-syntax
```diff
@@ -757,6 +814,442 @@ contexts:
...
+  pattern-multiline-string-body:
+    - meta_include_prototype: false
+    - meta_scope: meta.string.lua string.quoted.multiline.lua
+    - include: multiline-string-body
+    - match: ''
+      embed: pattern-unquoted
+      # we don't seem to be able to use the capture to get the right one
+      # this means that patterns that contain something that matches this
+      # will lose their context half way through.
+      escape: (?=\]=*\])
```
deathaxe, line 1151: Changes to simple multi-push, optionally consuming `\n` directly after opening punctuation.

```suggestion
  pattern-multiline-string-body:
    - meta_include_prototype: false
    - match: \n?
      push:
        - pattern-unquoted
        - pattern-test-first-caret
```

Lua/Lua.sublime-syntax
```diff
@@ -692,9 +706,19 @@ contexts:
...
           pop: 1
         - include: reserved-word-expression-pop
+        - match: (?="(?:{{string_char_escape}}|[^\\"])*"\s*\)\s*:\s*format{{identifier_break}})
+          push: fmtstr-expression
+        - match: (?='(?:{{string_char_escape}}|[^\\'])*'\s*\)\s*:\s*format{{identifier_break}})
+          push: fmtstr-expression
+        - match: (?="(?:{{string_char_escape}}|[^\\"])*"\s*\)\s*:\s*(?:pack|unpack|packsize){{identifier_break}})
+          push: packstr-expression
+        - match: (?='(?:{{string_char_escape}}|[^\\'])*'\s*\)\s*:\s*(?:pack|unpack|packsize){{identifier_break}})
+          push: packstr-expression
```
deathaxe, line 716: Those patterns expect pack/format strings to be one-liners.

But even expressions of concatted strings (e.g.: `("one %0s two" + "another %d"):format();`) are not scoped format string.

Other syntax definitions like python, don't restrict placeholder this way, as it is also valid to specify a format string in a separate variable.

```
fmt_str = "My %s format"
fmt_str:format("name")
```

I'd suggest to highlight placeholders everywhere or head for a more flexible approach using `branching` to scope all strings in a group followed by `:format()`, ... accordingly.

jrappen:
@mbartlett21 maybe you could address the change-requests?

mbartlett21:
@deathaxe / @jrappen 

I've finally gotten around to doing this. Will I need to rebase to get CI to work or is there no dev build at the moment?

jrappen:
SublimeHQ were doing some infrastructure updates for their linux builds over the last couple of days and re-built the last dev and stable builds. I guess something broke?

/cc @BenjaminSchaaf

jrappen:
Ben said he fixed their side.

Lua/Lua.sublime-syntax
```diff
@@ -683,17 +664,114 @@ contexts:
...
+    - match: \\.
+      scope: invalid.illegal.invalid-escape.lua
+      pop: 1
+
+###[ PARENTHESIZED EXPRESSIONS ]###############################################
+
   parenthesized-expression:
     - match: \(
       scope: punctuation.section.group.begin.lua
+      branch_point: special-string-paren
```
jrappen: It might make sense to name branch point the same as the context containing it.

jrappen:
I didn't have a lot of time to look through this.

- you might want to review tests where `meta.*` scopes are used so that:
  - `meta.mapping` vs. `meta.mapping.key` vs. `meta.mapping.value` are correctly tested like
    ```text
    -- SYNTAX TEST "Packages/Lua/Lua.sublime-syntax"

        KEY SEPARATOR VALUE
    --  ^^^^^^^^^^^^^^^^^^^ meta.mapping
    --  ^^^ meta.mapping.key
    --     ^ - meta.mapping.key
    --               ^^^^^ meta.mapping.value
    --                     ^ - meta.mapping
    ```
- also check strings are correctly tested, especially multiline ones.
  Use character combinations for the beginning and the end of test lines using the definition in the first line, compare:
  ```python
  # SYNTAX TEST "Packages/x/x.sublime-syntax \

  multilinestring = "whatever \
  #                           ^ separator.line \
  "
  ```
  unless lua multiline string work with """ whatever """ then you'd not need to define line separators like above. But you'll figure it out.
- the structure of the context for multiline strings looks different from other languages, I didn't have too much time to look through those. If not everything is working as intended, maybe check other language syntaxes with multiline strings.
- some added changes to the syntax file are not reflected in the tests added

PR #4197, opened 2025-03-28
Add "set_unsaved_view_name_for_syntax" setting

BenjaminSchaaf:
This will be shipping in a future build of Sublime Text to control which syntaxes the `set_unsaved_view_name.py` plugin applies to, replacing a hardcoded check for the "Plain text" syntax.

### DIFF
--- a/Markdown/Markdown.sublime-settings
+++ b/Markdown/Markdown.sublime-settings
@@ -1,0 +1,3 @@
+{
+    "set_unsaved_view_name_for_syntax": true,
+}
--- a/Markdown/MultiMarkdown.sublime-settings
+++ b/Markdown/MultiMarkdown.sublime-settings
@@ -1,0 +1,3 @@
+{
+    "set_unsaved_view_name_for_syntax": true,
+}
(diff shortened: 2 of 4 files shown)


deathaxe:
More (text-like) syntaxes can be added at any point in future.

I wonder if the new setting just replaces the syntax check in `Default/set_unsaved_view_name.py`.

MarkdownEditing for instance uses first heading to name a document, because it is possible the first line just contains `---` or `+++` marker denoting frontmatter.

Assuming MD, RST or Textile containing structured text, with a primary header, this one would probably also provide most valuable name.

deathaxe:
`Default/set_unsaved_view_name.py` seems to contain various ST2 specific details, which are no longer required since ST3 due to API being threadsafe.

- global `self.pending` while trying to debounce view-specific events is suspicious.

- `self.setting_name` seems obsolete.

- debouncing by 20ms is questionable as I doubt anyone being able to type that fast. An update rate of at least 50 or even 100 would probably make more sense if debouncing is required at all.

deathaxe:
Let's wait for activity until end of weekend.

PR #4207, opened 2025-04-07
[Perl] Add postfix dereference syntax

### LINKED ISSUE #4204: [Perl] Postderef asterisk highlights together with a following character
### Description

Best visible using ayu and Monokai as they colour the sigil and the variable name the same.

It seems that the `*` in the postderef all operator (`%*`, `@*`, etc.) isn't considered part of the phrase with the sigil, but starts another one itself (as if it were a glob variable I guess).

### Examples
```perl
for my $blah (sort keys $foo->%*) {
#                              ^^
}

($href1->%*, $href2->%*)
#         ^^          ^^
```

### Actual result
In these examples, I see a blue (same colour as magic variables and globals, like `$@` or `%ENV`) "variable" `*)` and `*,`

### Expected result
As the `*` is a part of the `%*` (or `@*`, etc.) dereferencing operation it should not behave like it were a glob variable.

deathaxe:
Fixes #4204

This commit implements:

1. https://perldoc.perl.org/perlref#Postfix-Dereference-Syntax
2. https://perldoc.perl.org/perlref#Postfix-Reference-Slicing

### DIFF
--- a/Perl/Perl.sublime-syntax
+++ b/Perl/Perl.sublime-syntax
@@ -1639,8 +1639,17 @@ contexts:
         - match: '{{identifier}}(?!::)'
           scope: meta.function-call.perl variable.function.member.perl
           pop: true
-        # item access like $array->[0]
-        - match: (?=[{\[])
+        # item access like $array->[0] or postfix dereferce slicing $array->@[0]
+        # https://perldoc.perl.org/perlref#Postfix-Reference-Slicing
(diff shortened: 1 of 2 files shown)


michaelblyons:
Sorry. I meant to check this out, and then forgot about it. Will revisit.

michaelblyons (approved):

deathaxe:
Thanks

PR #4215, opened 2025-04-11
[Perl] Fix POD comments

### LINKED ISSUE #4211: [Perl] The first POD "title" is highlighted differently from the rest
### What happened?

It seems like that the POD command starting the POD block (such as `=head1`) will format its argument differently from any successive commands.  It might be intentional but then I can't figure out what that could be.  Example attached in Mariana.

Image

deathaxe:
Fixes #4211 

This commit...

1. restructures POD related contexts and patterns to fix a wrong assumption of them needing to be wrapped into =pod and =cut tags.

   Instead any =<name> tag starts a new paragraph.

2. removes `meta.comment` scope

3. removes `meta.interpolation` scope for highlighted code blocks, as this is just not what they are.

### DIFF
--- a/Perl/syntax_test_perl.pl
+++ b/Perl/syntax_test_perl.pl
@@ -6,62 +6,95 @@
 
 ###[ POD TESTS ] #############################################################
 
-=pod
-# <- meta.comment.perl comment.block.documentation.perl entity.name.tag.pod.perl
-#^^^ meta.comment.perl comment.block.documentation.perl entity.name.tag.pod.perl
-=encoding utf8
-# <- meta.comment.perl comment.block.documentation.perl entity.name.tag.pod.perl
-#^^^^^^^^^^^^^ meta.comment.perl comment.block.documentation.perl
-#^^^^^^^^ entity.name.tag.pod.perl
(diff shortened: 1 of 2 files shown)


Perl/syntax_test_perl.pl
```diff
@@ -869,7 +904,7 @@ =head1 B<--param>
 # <- meta.string.heredoc.perl text.html.embedded.perl
 #^^^^^^^^^ meta.string.heredoc.perl text.html.embedded.perl
    HTML
-#  ^^^^ meta.string.heredoc.perl text.html.embedded.perl - constant.other.language-name.html.perl
+#  ^^^^ meta.string.heredoc.perl text.html.embedded.perl - constant.other.language-name.perl
```
michaelblyons, line 907: This negation still feels oddly specific to me. Did you mean for `- constant` or `- constant.other` instead?

deathaxe: That was too long ago to remember. This line is just a result of search and ~~destroy~~ replace.

michaelblyons (approved):

PR #4217, opened 2025-04-16
[Rust] Add support for shebang and frontmatter

### LINKED ISSUE #4164: [Rust] Frontmatter syntax support
### What happened?

[RFC #3503](https://github.com/rust-lang/rfcs/pull/3503) adds support for a new frontmatter syntax.  The contained content format is defined by the calling tool with a tool identifier.

In [RFC #3502](https://github.com/rust-lang/rfcs/pull/3502), the `cargo` infostring identifier would be used to signify the block uses the `Cargo.toml` format/schema.  Cargo will assume that an empty infostring is meant to be used with it.

**Note:** These are not stable.  Stabilization is being tracked in:
- https://github.com/rust-lang/rust/issues/136889
- https://github.com/rust-lang/cargo/issues/12207

e.g.
```rust
#!/usr/bin/env cargo

---
[dependencies]
regex = "1"
---

fn main() {
}
```

deathaxe:
Resolves #4164

### DIFF
--- a/Rust/Rust.sublime-syntax
+++ b/Rust/Rust.sublime-syntax
@@ -9,7 +9,8 @@ file_extensions:
 
 first_line_match: |-
   (?xi:
-    ^ \s* // .*? -\*- .*? \brust\b .*? -\*-  # editorconfig
+    ^ \#! .* \b(cargo|rust)\b                # shebang
+  | ^ \s* // .*? -\*- .*? \brust\b .*? -\*-  # editorconfig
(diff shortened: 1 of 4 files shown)


michaelblyons (approved):

deathaxe:
Most recent commit primarily targets syntaxes to be implemented equally with regards to how shebang context is pushed on stack.

michaelblyons (approved):

PR #4222, opened 2025-04-19
[Markdown] Add support for TOML frontmatter

deathaxe:
This commit adds support for TOML highlighted markdown frontmatter.

caused by: https://github.com/SublimeText-Markdown/MarkdownEditing/issues/736


michaelblyons:
Code looks good. Add a test file?

deathaxe:
As first line requies to contain only `+++` for frontmatter to be detected, how would we add required syntax test header?

michaelblyons:
😞 Oh, it needs to be first line? I usually cheat and let comments be available for first-line sections... but Markdown doesn't even have comments.

michaelblyons (approved):

PR #4227, opened 2025-04-27
[Common] Scope `in` iteration

deathaxe:
This PR scopes `in` keywords in loop statements `keyword.operator.iteration`.

Uses "iteration" rather than "iterator" as it is more aligned with "assignment" or "comparison".

This commit is part of an attempt to introduce a single scope for that keyword as various (3rd-party) packages use different ones, like:

-  `keyword.control.flow`
-  `keyword.control.loop.for.in`
-  `keyword.operator`
-  `keyword.operator.logical`
-  `keyword.operator.in`
-  `keyword.operator.iterator`
-  `keyword.operator.iteration`
-  `keyword.other.in`

3rd-party packages known to already use proposed scope:

- CoffeeScript
- Jinja2
- Liquid
- Ngx HTML
- PostCSS
- Sass
- Twig
- XSL

### DIFF
--- a/ASP/ASP.sublime-syntax
+++ b/ASP/ASP.sublime-syntax
@@ -663,7 +663,7 @@ contexts:
   control_flow_foreach_in:
     - meta_scope: meta.for.block.asp
     - match: '\b(?i:In){{whitespace_or_end_of_statement}}'
-      scope: keyword.control.flow.asp
+      scope: keyword.operator.iteration.in.asp
       set: [inside_control_flow_for, expression_until_end_of_statement]
 
   control_flow_forto:
--- a/ASP/syntax_test_asp.asp
+++ b/ASP/syntax_test_asp.asp
@@ -878,7 +878,7 @@
    '<- - meta.between-if-and-then.asp
    '^^^^^^^^ keyword.control.flow.asp
    '         ^^^^^^ variable.other.asp
-   '                ^^ keyword.control.flow.asp
+   '                ^^ keyword.operator.iteration.in.asp
    '                                  ^ meta.for.block.asp
         Response.Write(vbCrLf & cookie)
        '^^^^^^^^ support.class.asp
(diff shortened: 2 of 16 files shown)


michaelblyons:
Is this like the `in` in Python `[x for x in ...]`? I kind of appreciate how that doesn't use `keyword.operator`, so it is different from `if x in [1, 2, 3]`.

Maybe I'm wrong about how to handle them, though. I'm open to other opinions. I do agree that the scopes across languages should be standardized.

deathaxe:
In all use cases `in` iterates over (operates on) given list of items returning them one by one (in `for i in ...` or `foreach i in ...` loops) or returning `true` if any matches (in `if x in ...`) statements.

`keyword.control.flow, used by some syntaxes is reserved for keywords like `break`, `continue`, `return`, ... .

Hence `keyword.operator` feels most appropriate. One may argue about iteration in context of if-clauses as those may be interpreted as `keyword.operator.comparison` or `keyword.operator.logical`, but nontheless iteration through given list of items is involved.

keith-hall:
Probably it doesn't matter too much, but remember that in Python, `if key in dict` performs no iteration, it's just a lookup by the hashed key...

michaelblyons:
Interesting. I would still *prefer* different scopes for "presence in set" than for "iterating through set" (even if they're all `keyword.operator.*`), but looking at some of the changed tests, this Ruby one might be a bit strange:

```ruby
while item in items do
```

deathaxe:
> but remember that in Python, if key in dict performs no iteration, it's just a lookup by the hashed key...

What does a lookup do? It iterates over items to do the lookup, but by just returning boolean if item was found.

That being said, scope of `in` in python's if statement are not changed by this PR:

deathaxe:
Ruby doesn't provide enough context to distinguish them, atm.

michaelblyons:
Keith, can you elaborate your upvote? Do you mean you would also prefer a split (and if so, how vehemently?) or you agree that the Ruby thing is funny?

I am inclined to approve even if I'm not 100% on scopes in every spot. I think it's a clear improvement over the present circumstances. I'm happy to wait for a few weeks, but I don't want good-but-slightly-disputed PRs float in limbo indefinitely.

deathaxe:
Note, Ruby has been adjusted to distinguish iteration (in for loops) and presence tests (in case, while, until statements). As ruby case statements doing pattern matching, using keyword.operator.logical scope (as before) is valid.

keith-hall:
> Keith, can you elaborate your upvote? Do you mean you would also prefer a split (and if so, how vehemently?) or you agree that the Ruby thing is funny?

both :) but I'm happy with this PR now

keith-hall (approved):

michaelblyons (approved):

FichteFoll:
I second @michaelblyons's concerns in that the `in` keyword as a part of a loop statement should *not* be scoped as `keyword.operator`. Python's `in` in expressions clearly is an operator, but in a `for x in y` statement the `in` acts more like a separator than an operator and is thus of the same `keyword.flow` nature as the initial `for`.

PR #4229, opened 2025-04-28
[Markdown] Reduce unique fenced codeblock scopes

deathaxe:
This commit...

1. removes syntax name from `meta.code-fence.definition.[begin|end].[syntax]`.

   As a result each code block uses common scope: `meta.code-fence.definition.[begin|end].markdown-gfm`

   Amount of unique scopes which need to be handled by ST is reduced.

2. adds `meta.code-fence.body` to content part, so code fences can be addressed or searched by `meta.code-fence` scope.

   Example:

       view.sel().add_all(view.find_by_selector("meta.code-fence"))

   selects all fenced code blocks in a document.

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -524,7 +524,7 @@ contexts:
           {{fenced_code_block_language}}?
           .*?(\s*$\n?)   # all characters until EOL
       captures:
-        0: meta.code-fence.definition.begin.text.markdown-gfm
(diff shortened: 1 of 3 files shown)


deathaxe:
This wouldn't include header/footer lines (those with `` ``` ``).

Furthermore it's a bit inconsistent to not apply a common meta scope to whole region of fenced code blocks, especially as we already have `meta.heading`, `meta.paragraph`, `meta.list` etc. covering all the other block types.

PR #4235, opened 2025-05-02
[Bash] Interpolation "block comment"

michaelblyons:
The main objective is to make toggle (block) comment work for sections of command arguments. Only works single-line; this may be because I can't append (and then remove) `\` to the end of each line.

I did not include some edits to Git Config comment tmPreferences that make `[alias]` definitions using shell prefer this style.


deathaxe:
To verify intented behavior ; with regards to 

```shell
curl \
	--proxy=socks5://localhost:4321 \
	--url=https://example.com
```

in order to comment out proxy arguments...

1. select `--proxy=socks5://localhost:4321`
2. hit `ctrl+shift+/` to comment it out

Once commented out

```shell
curl \
	`# --proxy=socks5://localhost:4321` \
	--url=https://example.com
```

same binding removes converts comment back in arguments?

That works for me only, when removing `string` from comment selector in _Comments - String.tmPreferences_. Otherwise only leading `#` is added/removed.

deathaxe (approved):

keikoro:
Was this meant to affect only how block comments work for command line arguments?

Because it's not usable for block comments for regular shell scripting, i.e. for files using Bash syntax.

A snippet like:
```sh
if ! [ -w "$dir" ]; then
   echo Path not writeable.
fi
```
now turns into the following when block commenting is invoked once:
```sh
`# if ! [ -w "$dir" ]; then
   echo Path not writeable.
fi`
```
and this mess when trying to toggle it off:
```sh
`# `# if ! [ -w "$dir" ]; then
   echo Path not writeable.
fi`` 
```

and it just goes on and on like this with every further invocation.

deathaxe:
As shellscripts don't have concepts for block comments, they are faked by converting selection into a subshell with only a line comment in it. Primary use case is commenting out single arguments of commands.

     cmd active arg1 `# inactive-arg` active-arg2

It doesn't however work well for multi-line statements as only first line right after `` ` `` is commented out via `#`. All following lines (e.g.: `echo` and `fi`) are still actively evaluated statements in a subshell.

This PR leverages ST's default toggle comment functionality to provide exactly this.

Neither are comments in subshells meant for huge block comments nor is ST's toggle comment functionality prepared to handle this.

Don't expect those commments to work like C style `/* */`.

Issue #4244, opened 2025-05-17
[zsh] misc. issues


okdana:
### What happened?

i've finally started using the new zsh highlighting (current master), which is very nice, but i've noticed a few issues, some of which can break the entire file. i'm not able to study the definitions myself atm so i'll just document them for now:

- [x] 1. this is highlighted as a command, but it should be a parameter assignment #4245

   ```zsh
   1=foo
   ```

- [x] 2. it highlights expansion flag arguments as if they're in the normal context where expansion, substitution, etc are valid (and therefore they have to follow those rules). but they're not. flag arguments are always literals, except in the presence of `(p)` where a limited form of parameter expansion is allowed (`${(pj<$j>)arr}`) #4246

   ```zsh
   : ${(j<$j>)arr} ${(j<$(ls)>)arr} ${(j<${>)arr}
   ```

- [x] 3. it treats everything within an array subscript as arithmetic, so subscript flags are not highlighted, glob syntax is marked as invalid, etc. this means it doesn't even like plain strings used as association keys. this is a big one for a lot of shell code shipped with zsh https://github.com/sublimehq/Packages/pull/4273

   ```zsh
   : ${arr[(r)([^@]##@|)foo]}
   : ${assoc[my.key]}
   ```

- [x] 4. i guess it wants everything after `autoload` to be a literal function name so it doesn't highlight paths correctly and it breaks completely on globs #4247

   ```zsh
   autoload /path/to/func
   autoload $dir/*/*~*.zwc(#q.N:t)
   ```

- [x] 5. i think it doesn't know that `q-` and `q+` are variants of `q`, not pairs of separate flags, and since `+` isn't a valid flag on its own it's not highlighted right #4246

   ```zsh
   : ${(q-)foo} ${(q+)foo}
   ```

- [x] 6. doesn't seem to like dynamically named functions, nor functions with quoted/escaped names, nor multi defs, unless they're preceded by `function` https://github.com/sublimehq/Packages/pull/4262

   ```zsh
   $foo() { : }
   \$foo() { : }
   'foo'() { : }
   foo bar() { : }
   ```

- [x] 7. the parameter name `fd` is highlighted as a command 
      _(WONTFIX, working as expected)_

   ```zsh
   exec {fd}>&-
   ```

[...]

PR #4245, opened 2025-05-18
[ZSH] Fix parameter identifier pattern

deathaxe:
Addresses #4244 issue 1

This commit fixes parameter/variable identifier patterns to comply with section 15.1 of ZSH language specification.

    A parameter has a name, a value, and a number of attributes. A name may be
    any sequence of alphanumeric characters and underscores, or the single
    characters ‘*’, ‘@’, ‘#’, ‘?’, ‘-’, ‘$’, or ‘!’. A parameter whose name
    begins with an alphanumeric or underscore is also referred to as a
    variable.

Variables consist of alphanumeric characters and thus can even start with
numeric characters.

Bash's `identifier` and `variable` patterns are tweaked in order to be able to implement different ZSH behavior by overriding a single variable, only.

Various tests are added to verify this pattern not only in variable assignment statements but also parameter expansions.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -3120,7 +3120,7 @@ variables:
 
   # Variable names
   variable_begin: (?={{variable_first_char}})
-  variable_first_char: '[[:alpha:]_''"$%]'
+  variable_first_char: '[{{identifier_first_char}}''"$%]'
 
   # POSIX identifiers (alpha-numeric)
   identifier: '{{identifier_first_char}}{{identifier_char}}*'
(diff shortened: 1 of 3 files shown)


keith-hall (approved):

deathaxe:
Misleading? The spec is a mess. It should provide proper ABNF grammer.

FichteFoll (approved):
Since the variable pattern is currently split into a start and continue character set, we don't currently have a way to validate whether a variable only consists of up to two numerical characters. The change addresses the false negative and introduces a false positive. However, the shell itself will yell on various false positives, so this is still an improvement.

deathaxe:
Finally the only difference between Bash and ZSH is, ZSH supporting new values being assigned to positional parameters, which Bash syntactically doesn't.

Therefore positional variables are already scoped by Bash everywhere but in l-value position of assignments. We could avoid false-positives by (partly) reverting this PR and just prepend `def-variable` context by a dedicated pattern for positional parameter assignments.

But these are the details, which are hard to learn from ZSH docs, especially for those whose primary OS doesn't support ZSH ;-)

PR #4256, opened 2025-05-21
[SQL] Fix: Postgre comments don't need a space after the double dash

keith-hall:
[reported on the forum](https://forum.sublimetext.com/t/sql-syntax-highlighting-doesnt-recognize-double-dash-with-no-space-after-it-as-comments/76241?u=kingkeith)

### DIFF
--- a/SQL/PostgreSQL.sublime-syntax
+++ b/SQL/PostgreSQL.sublime-syntax
@@ -66,6 +66,14 @@ contexts:
         2: punctuation.section.brackets.end.psql
       set: maybe-group
 
+###[ COMMENTS ]################################################################
+
+  double-dash-comments:
+    - meta_append: true
+    - match: '--'
+      scope: punctuation.definition.comment.sql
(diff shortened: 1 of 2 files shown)


SQL/PostgreSQL.sublime-syntax
```diff
@@ -66,6 +66,14 @@ contexts:
         2: punctuation.section.brackets.end.psql
       set: maybe-group
 
+###[ COMMENTS ]################################################################
+
+  double-dash-comments:
+    - meta_append: true
```
deathaxe, line 72: As the context contains just this one rule in both SQL (basic) and MySQL, we probably don't need to append, just replace again to apply SQL (basic) pattern.

```suggestion
```

michaelblyons:
ANSI doesn't need a space, does it? Maybe MySQL should be the one with the override.

michaelblyons (approved):

deathaxe (approved):

PR #4261, opened 2025-05-22
[SQL] Fix ALTER TABLE ADD FOREIGN KEY

### LINKED ISSUE #4260: [SQL] Incorrect syntax error highlighting in mysql
### What happened?

Recognizing there are a few other issues opened surrounding the recent changes in 4200 that brought more specification strict SQL syntax highlighting, so please close this as needed relative to MRs going in for those issues (#4258, others mentioned in the forums).

In 4200, I'm seeing closing parens being flagged as syntactically invalid in this scenario when using the SQL, mySQL and postgres syntax highlighter.  

Image

```sql
ALTER TABLE Orders
ADD FOREIGN KEY (PersonID) REFERENCES Persons(PersonID);

ALTER TABLE Orders
ADD CONSTRAINT FK_PersonOrder
FOREIGN KEY (PersonID) REFERENCES Persons(PersonID);

```

keith-hall:
Fixes #4260

### DIFF
--- a/SQL/MySQL.sublime-syntax
+++ b/SQL/MySQL.sublime-syntax
@@ -457,13 +457,13 @@ contexts:
         1: keyword.other.ddl.sql
         2: keyword.other.ddl.sql
       set:
-        - maybe-column-modifier
         - alter-table-add-index-args
         - expect-index-name
         - maybe-condition
 
   alter-table-add-index-args:
     - include: column-name-lists
+    - include: column-reference-definitions
     - include: comment-keywords
     - include: index-algorithm-args
     - include: index-clustering-args
--- a/SQL/MySQL.sublime-syntax
+++ b/SQL/MySQL.sublime-syntax
@@ -457,6 +457,7 @@ contexts:
         1: keyword.other.ddl.sql
         2: keyword.other.ddl.sql
       set:
+        - maybe-column-modifier
         - alter-table-add-index-args
         - expect-index-name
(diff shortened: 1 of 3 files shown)


michaelblyons (dismissed):

SQL/MySQL.sublime-syntax
```diff
@@ -457,6 +457,7 @@ contexts:
         1: keyword.other.ddl.sql
         2: keyword.other.ddl.sql
       set:
+        - maybe-column-modifier
```
deathaxe, line 460: Not all column modifiers are valid in that context.

`REFERENCES` seems to belong to `reference_definitions` which is supported after FOREIGN KEYS only.

We should maybe add related pattern directly to `alter-table-add-index-args`.

see: 

https://github.com/sublimehq/Packages/blob/77c99b4cbc4d2e65f65efd146184d3b2fd82d52d/SQL/tests/syntax/syntax_test_mysql.sql#L2413-L2415

Note: That's the section, related tests should be added to.

deathaxe: We should keep closely to provided BNF representations, to not mix up things and cause other issues.

keith-hall: Good point, I will try to bear that in mind 👍

deathaxe (approved):

michaelblyons (approved):

PR #4265, opened 2025-05-23
[ZSH] Support more sophisticated function definition names

deathaxe:
This PR is a follow-up of #4262 to add support for all sorts of quoting and interpolation within function definition names beyond scope of issue #4244.

### DIFF
--- a/ShellScript/Zsh.sublime-syntax
+++ b/ShellScript/Zsh.sublime-syntax
@@ -1338,7 +1338,8 @@ variables:
 
   illegal_group_separators: '[|&;>]'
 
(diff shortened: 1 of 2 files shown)


deathaxe:
This one results in various false positives to solve.

Examples:

```
# local scoped as function
local mark='_.!~*''()-'

# [[ and ]] scoped as function
[[ ( ) ]]
```

deathaxe:
This one is fixed by this PR

```
# local scoped as function
local mark='_.!~*''()-'
```

This one is to be fixed by modified import order introduced by #4282

```
# [[ and ]] scoped as function
[[ ( ) ]]
```

michaelblyons (approved):

PR #4266, opened 2025-05-23
[ZSH] Fix declare/typeset/unset variable assignments

deathaxe:
Addresses #4244 issue 15

This commit...

1. reduces strictness of `declare` and `typeset` option patterns, in order to ensure `-m` option (or any new one) doesn't break distinction between literal/arithmetic/array/mapping value highlighting.

   Only important flags `i`, `a`, `A` and `f` are taken into account.

2. adds `\\` and `{` as valid variable name begin characters to fix identifiers starting with escaped characters or brace expansions.

Note: Fixes Bash and Zsh.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -448,7 +448,7 @@ contexts:
 
   cmd-declare-args:
     # option contains -f or -F
-    - match: ([-+])[aAgiIlnrtux]*[Ff][aAgiIlnrtux]*{{opt_break}}
+    - match: ([-+])[A-Za-z]*[Ff][A-Za-z]*{{opt_break}}
       scope:
         meta.parameter.option.shell
         variable.parameter.option.shell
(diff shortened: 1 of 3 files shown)


ShellScript/Bash.sublime-syntax
```diff
@@ -448,31 +448,31 @@ contexts:
 
   cmd-declare-args:
     # option contains -f or -F
-    - match: ([-+])[aAgiIlnrtux]*[Ff][aAgiIlnrtux]*{{opt_break}}
+    - match: ([-+])[A-Za-z]*[Ff][A-Za-z]*{{opt_break}}
```
michaelblyons, line 451: At this point, is it worth just `([-+])(?i:[a-z]*f[a-z]*){{opt_break}}`? Or does the explicit upper/lower in every set help recognition?

I guess the pattern breaks down for the other ones that need explicit upper or lower case for the mode letter down below. Variable for `[A-Za-z]*?` then, maybe?

michaelblyons (commented):
No particular objections. Just some questions.

ShellScript/Bash.sublime-syntax
deathaxe: General `(?i)` flag isn't that ideal, as `i` needs to be matched case-sensitive and all patterns should follow same style.

To apply same pattern as for ordinary options, we can however leverage `{{word_char}}` to replace `[A-Za-z]`.

michaelblyons:
It doesn't work with the whole `*?`? 🥴 I feel silly now.

deathaxe:
Not sure I follow, but I don't think adding a dedicated variable just to also augment `*?` is worth it. It would work, but without any benifit with regards to readability or maintainability. Just another var with another level of information required to follow.

deathaxe:
FWIW, there are oddities to be supported, which are not covered by this PR.

Basically quotation marks can appear at any position. Both of the following lines assign new value to given array element, but non of which is currently scoped properly.

```
typeset "var[1]=value"

typeset "var[1]"=value
```

keith-hall (approved):

PR #4273, opened 2025-05-26
[ShellScript] Add Zsh subscript flags

deathaxe:
Addresses #4244 issue 3

This PR...

1. implements subscripts with branching to heuristically distinguish indexed array item access from key based access of associative array items.

   Branches try to ...
   1. match subscript flags followed by patterns or plain text (only ZSH)
   2. match valid arithmetic expressions
   3. fallback to plain string subscripts

   Expressions related contexts have been slightly refactored to re-use most
   parts in subscripts while accounting for unique bailout requirements aka.
   failing related branch in case of illegal syntax being found.

2. restricts patterns to consume at most one subscript bracket pair after each parameter name to comply with language behavior.

   This change involves pushing a dedicated context to consume simple variable expansions (e.g. `$var`) onto stack (see: simple-parmeter-expansion), which significantly reduces syntax cache size of both Bash and Zsh by about 25%.

3. organizes related contexts in a dedicated section and changes context names by using the term `subscript` instead of `subscription`.

Syntax tests are organized by Bash/Zsh manual chapters.

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -1560,7 +1560,7 @@ contexts:
     # attempt to scope subscript content arithmetic expression
     # if anything unrecognized was found, fallback to plain string
     - match: ''
-      branch_point: subscript
+      branch_point: subscript-content
       branch:
         - subscript-expression-content
         - subscript-plain-content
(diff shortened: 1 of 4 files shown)


deathaxe:
Thanks for reviewing!

PR #4279, opened 2025-05-30
[SQL] Add default SQL syntax selector command

deathaxe:
This PR adds a naive command for users to choose default SQL dialect to be assigned to `source.sql`.

grafik

The `sql_set_default_syntax` basically works for any syntax, but is prefixed `sql_` as it is located in and dedicated to SQL syntax package.

Alternatively a general purpose `set_default_syntax` could be implemented by ST core or Default package.

Ideally, dialect syntaxes could be identified by attributes in `Syntax()` class, but until this becomes true, rely on a static list of files specified via `dialect_file` argument to choose from. 

FWIW, Right now SQL seems to be the only use case for what is called a "dialect".

### DIFF
--- a/SQL/default_syntax.py
+++ b/SQL/default_syntax.py
@@ -111,10 +111,5 @@ class SqlSetDefaultSyntaxCommand(sublime_plugin.WindowCommand):
             return
 
         # write modified default syntax to extracted Packages path
-        with open(
-            file=Path(sublime.packages_path()).parent / default_syntax.path,
-            mode="w",
-            encoding="utf-8",
-            newline="\n",
-        ) as out:
-            out.write(new_content)
+        file = Path(sublime.packages_path()).parent / default_syntax.path
+        file.write_text(new_content, encoding="utf-8", newline="\n")


predragnikolic:
If I have a `*.sql` file open, 
and I would like to specify a dialect for that file,
intuitively I would look in the command palette to see if it offers `Set Syntax: SQL (MySQL)` or `Set Syntax: SQL (PostgreSQL)`.

deathaxe:
You can do so and even assign *.sql extension to any of MySQL, PostgreSQL or T-SQL.

Syntax names like `<Name> (<Base>)` denote extended syntaxes based on another. That's not really the case here, because `SQL` is by default just an alias for `MySQL`, which can be changed to anything. All dialects are based on `SQL (basic)` _(yes it violates given scheme above)_.


The point of this PR and provided command is to be able to specify the SQL syntax/dialect to use in syntax highlighted strings of Python, PHP, ... as those don't provide any other way to do so - and just embed `source.sql`.

That being said, we could hide "SQL" syntax and just assign *.sql extension to MySQL by default - somewhat following the scheme of ShellScript.

jrappen (dismissed):

jrappen:
Looks fine, needs SublimeHQ core team review & decision.

michaelblyons:
Bummer. For anyone missing this, you can install https://github.com/deathaxe/DefaultSyntaxChooser

jrappen:
@deathaxe Could you please comment on why you close your PRs when you do so?

Issue #4272, opened 2025-05-26
[zsh] globqual issues


okdana:
### What happened?

hi again,

```zsh
[[ $a == (e|b) ]]
```

this breaks the rest of the file, apparently because it thinks the `e|` is an unterminated `e` glob qualifier. there are a few issues with this:

* glob qualifiers must be preceded by a pattern, so a stand-alone `(...)` shouldn't be interpreted as one

* bare glob qualifiers -- the `(...)` form rather than the `(#q...)` one -- can't contain `|`, so even if it was preceded by a pattern it shouldn't be interpreted as one

* bare glob qualifiers (even legal ones) can't appear inside a `[[` condition

* extended glob qualifiers `(#q...)` *can* appear inside a `[[` condition, but not on the right-hand side of a pattern-matching expression like `a == b` or `a != b`. i think they can appear anywhere else in one

(this is generally covered under [Glob Qualifiers](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Glob-Qualifiers), but lmk if i can improve it, short of a formal grammar lol)

so

```zsh
echo (e::)    # not globqual (no preceding pattern)
echo (#qe::)  # not globqual (no preceding pattern)
echo x(e||)   # not globqual (bare globqual may not contain '|')
echo x(e::)   # is  globqual
echo x(#qe::) # is  globqual
echo x(#qe||) # is  globqual

[[ $a == x(.) ]] # not globqual (bare globqual not valid in [[ ... ]])
[[ x(.) == $a ]] # not globqual (bare globqual not valid in [[ ... ]])

[[ $a == x(#q.) ]] # not globqual (globqual not valid on rhs of [[ a == b ]])
[[ x(#q.) == $a ]] # is  globqual
```

thanks again for your work on this

deathaxe:
How does the following example fit into the picture of bare qualifiers not being allowed to contain `(` within parentheses?

```
echo *(e:'reply=(${REPLY}{1,2})':)
```

Actually whole description of 

```
14.8.7 Glob Qualifiers

...

estring
+cmd

...

```

seems to violate those restrictions. Or are those valid only within extended qualifiers (#qe...)?

PR #4282, opened 2025-06-02
[ShellScript] Improve built-in and compound test expressions

deathaxe:
Addresses #4272 / no bare glob qualifiers in compound tests

This PR refactors built-in `test ...` and `[ ... ]` as well as compound `[[ ... ]]` test expressions to apply more detailed parsing rules, which follow more closely closely to language specifications and shell parsing behavior.

This fixes various edge cases and enables more fine gained and accurate highlighting.

Parsing performance isn't affected by this change, syntax cache size is also nearly unchanged.

For details, please refer to commit messages.

Some highlights:

- highlight math expressions before and after `-eq` and friends

  grafik
   
- support some of Bash/Zsh's insanity

  grafik

- compare two redirections

  grafik

- fix arbitrarily nested groups in expressions

  grafik

  grafik

- no bare globs in compound tests

  grafik

  grafik

- various other fixes

### DIFF
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -742,6 +742,7 @@ contexts:
       set:
         - compound-test-eregexp-body
         - eregexp-unexpected-quantifier
+        - maybe-tilde-interpolation
 
   compound-test-eregexp-body:
     - meta_include_prototype: false
--- a/ShellScript/Bash.sublime-syntax
+++ b/ShellScript/Bash.sublime-syntax
@@ -694,6 +694,8 @@ contexts:
       scope: keyword.operator.comparison.shell
     - match: '<(?!<)|>(?!>)'
       scope: keyword.operator.comparison.shell
+    - match: -[ao]{{opt_break}}(?!\s*(=~|!=|==?))
+      scope: keyword.operator.logical.shell
     - include: cmd-args-end
     - include: test-expression-common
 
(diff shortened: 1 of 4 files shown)


keith-hall (dismissed):

deathaxe:
Most recent commit "just" creates dedicated section for compound tests, renames contexts and includes compound tests earlier to fix ambiguity with function definitions.

jrappen:
Are we not sticking to syncing branch_point names with the names of the contexts containing them?

Maybe you could replace images with code blocks in the opening comment.

deathaxe:
If we create a dedicated context with a single pattern forming a branch_point, we likely want to do so.

In this case the patterns forming branch_points are one of many in their context, because it is the one and only location of usage. There could likely be multiple branch_point forming patterns each of which must have a unique name and thus can't equal the context name.

The names have been carefully choosen, except the `cmd-` prefix probably not being needed.

IF those 2 patterns would be organized in a dedicated context, it would be named like the branch_points. But there are no reasons to do so.

Current implementation is a short form of

```diff
@@ -714,6 +714,9 @@
       captures:
         1: punctuation.definition.parameter.shell
       push: cmd-compound-test-pattern
+    - include: cmd-compound-test-comparison
+
+  cmd-compound-test-comparison:
     # comparison expressions
     - match: '{{word_begin}}'
       branch_point: cmd-compound-test-comparison
```

which was avoided due to unique usage of cmd-compound-test-comparison pattern.

keith-hall (approved):

PR #4284, opened 2025-06-05
JSON.sublime-syntax: support jsonl

pbsds:
.jsonl is a simple newline delimited array of json objects

https://jsonlines.org/

### DIFF
--- a/JSON/JSON.sublime-syntax
+++ b/JSON/JSON.sublime-syntax
@@ -6,6 +6,7 @@ version: 2
 
 file_extensions:
   - json
+  - jsonl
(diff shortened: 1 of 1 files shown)


deathaxe:
Note that just adding .jsonl file extension will cause LSP to treat it as normal JSON and complain about multiple top-level objects/items being present.

grafik

pbsds:
would `hidden_file_extensions` work?

pbsds:
Ah, it seems the upstream project which uses these definitions already has a case for jsonl, it's just bugged

PR #4307, opened 2025-06-14
[ZSH] Fix qualifier and expansion modifiers

### LINKED ISSUE #4272: [zsh] globqual issues

deathaxe:
Addresses #4244 issue 12

Fixes #4272

This PR...

1. restricts allowed characters in bare qualifiers (but not in extended qualifiers).

   Chars like `|`, `(` or `~` in `(...)` signal it being a pattern group rather than a bare qualifier, in order to avoid ambiguities such as `(e|expr|)`.

2. adds history modifiers to parameter expansions.

Note: requires (includes) PR #4297.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -14,10 +14,12 @@ file_extensions:
 
 first_line_match: |-
   (?xi:
-    ^ \s* // .*? -\*- .*? \b(c\#|cs|csharp)\b .*? -\*-  # editorconfig
+    ^ \#! .* {{shebang_language}}                       # shebang
+  | ^ \s* // .*? -\*- .*? \b(c\#|cs|csharp)\b .*? -\*-  # editorconfig
   )
 
 variables:
+  shebang_language: \bdotnet\s+run\b
 
   bin_op: '(?:\+|->|-|\*|/|%|\|\||&&|\||&|\^|<<|>>|=>|<=|<|>=|>|==|!=|\?\?)'
   unary_op: '(?:\+\+|--|-|~|!|&|\*)'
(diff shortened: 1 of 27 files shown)


deathaxe:
That's the case.

deathaxe:
Thanks!

PR #4308, opened 2025-06-16
[JavaScript] add symbol tests, minor syntax fixes

jrappen:
This pull request fixes the highlighting of line comments at end of
file, increases the readability of a regex for string escape characters
and adds some missing syntax tests.


jrappen:
⚠️ Another symbol transformation bug with regard to syntax tests? Tests break with symbols starting and ending with quotes (single or double).

jrappen:
I wanted to test with v4201, but the workflow still fetches the v4199 binary. Will try again with newer builds.

PR #4309, opened 2025-06-17
[SQL] Add pgsql as the main file extension for Postgres dialect

keith-hall:
ref: https://discord.com/channels/280102180189634562/280157121897562114/1383880358952435774


michaelblyons:
I don't know which should be the default, but having both in the list seems like a good plan.

deathaxe (approved):

PR #4310, opened 2025-06-19
[C#] Support `dotnet run script.cs`

keith-hall:
- support shebangs the same way other syntax definitions do
- improve the preprocessor directive (`preprocessor_option`) context
  - no longer mark anything unexpected as illegal
  - after a colon, scope the following word
  - this ensures it can correctly highlight the following officially supported keywords:
    - `sdk`
    - `package`
    - `property`
  - currently it leaves the text after it unscoped. We can adjust this based on feedback later if necessary.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -200,9 +200,6 @@ contexts:
         2: keyword.other.preprocessor.cs
         3: keyword.other.preprocessor.cs
 
-    - match: '!'
-      scope: punctuation.definition.comment.cs
-      set: shebang-body
     - match: (:)(\w+)
       captures:
         1: punctuation.definition.preprocessor.cs
--- a/C#/tests/syntax_test_dotnetrun.cs
+++ b/C#/tests/syntax_test_dotnetrun.cs
@@ -5,9 +5,6 @@
 
 // https://devblogs.microsoft.com/dotnet/announcing-dotnet-run-app/
 
-#!/usr/bin/dotnet run
-#!^^^^^^^^^^^^^^^^^^^ comment.line.shebang.cs
-#!         ^^^^^^^^^^ constant.language.shebang.cs
 #:package Humanizer@2.14.1
 #!^^^^^^^^^^^^^^^^^^^^^^^^ meta.preprocessor.cs
 #!^^^^^^^ variable.language.cs


jrappen:
`shebang`s are usually only at BOF, why this exception?

keith-hall:
Mainly so I could make a syntax test for it probably, and it falls under preprocessor directives which are supported anywhere

jrappen:
`Packages/C#/C#/tests/syntax_test_shebang_dotnet_run.cs`:


```csharp
#! SYNTAX TEST "Packages/C#/C#.sublime-syntax" dotnet run
#! <- comment.line.shebang.cs punctuation.definition.comment.cs
  #! <- comment.line.shebang.cs constant.language.shebang.cs
#!^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ comment.line.shebang.cs

```

C#/C#.sublime-syntax
```diff
@@ -15,9 +15,11 @@ file_extensions:
 first_line_match: |-
   (?xi:
     ^ \s* // .*? -\*- .*? \b(c\#|cs|csharp)\b .*? -\*-  # editorconfig
+  | ^ \#! .* {{shebang_language}}                       # shebang
```
deathaxe, line 18: Can we use same order of patterns as in all other syntaxes?

```suggestion
    ^ \#! .* {{shebang_language}}                       # shebang
  | ^ \s* // .*? -\*- .*? \b(c\#|cs|csharp)\b .*? -\*-  # editorconfig
```

C#/C#.sublime-syntax
```diff
@@ -227,6 +252,11 @@ contexts:
       scope: invalid.illegal.cs
 
   main:
+    - meta_include_prototype: false
+    - match: ''
+      push: [statements, shebang]
```
deathaxe, line 257: With a dedicated `statements` context, can `main` be moved to the beginning next to `prototype`?

Seeing the main entry positioned in the middle of the file feels a bit strange.

C#/C#.sublime-syntax
```diff
@@ -192,8 +194,14 @@ contexts:
         1: keyword.other.preprocessor.cs
         2: keyword.other.preprocessor.cs
         3: keyword.other.preprocessor.cs
-    - match: .*
-      scope: invalid.illegal.cs
+
+    - match: '!'
+      scope: punctuation.definition.comment.cs
+      set: shebang-body
```
deathaxe, line 200: This pattern causes #! being scoped as 2 tokens is not ideal, even if it visually looks as one.

The main purpose of `shebang` context is to only scope first line as such. With this pattern scoping it everywhere `shebang` was redundant.

If shebang is not supported in the middle of files we could probably accept test assertions not being highlighted or just use a dedicated a dedicated `syntax_test_shebang.cs` for shebang tests, while keeping test assertions in normal comment style for `#:`.

C#/C#.sublime-syntax
keith-hall: yes, done :)

C#/C#.sublime-syntax
keith-hall: `#`in this case represents preprocessor directives, not comments. I checked and `dotnet run` throws an error if the shebang is not at the first line of the file, so I removed this part. See what you think of the latest changes :)

deathaxe (approved):

PR #4318, opened 2025-06-25
[JavaScript] Fix nested template strings

### LINKED ISSUE #4220: [JavaScript] Template string inside tag function breaks syntax highlighting
### What happened?

Template string inside tag function:
```js
let x = html`<div>
  ${`<p>World</p> `}
  </div>
  <p>This</p>
`;
```

or tag function inside tag function
```
let x = html`<div>
  ${html`<p>World</p> `}
  </div>
  <p>This</p>
`;
```
breaks syntax highlighting and I see this:
Image

deathaxe:
Resolves #4220

This PR...

1. introduces _JavaScript (Plain)_, _TypeScript (Plain)_, _JSX (Plain)_ and _TSX (Plain)_ without syntax highlighted template strings as possible base syntax for template languages which don't need them or would fail loading due to complexity otherwise.

3. replaces `embed` directives by `set` and includes all syntaxes used for highlighting template strings, to prevent them being prematurely terminated by interpolations with nested template strings.

4. Nested template strings don't highlight any syntax to keep syntax cache size in sane limits.

Note:

Compiled syntax cache of each language grows from about 170kB to 1.2MB also causing all template languages which make use of it to grow. I didn't hit any limits yet and Python is also 2.5MB now with new SQL syntax being included a couple of times. So maybe it is safe enough for production.


TerminalFi:
I wouldn’t say it only goes from 4.2 to 5.4. That’s 25% or so increase in size. Relatively large

deathaxe:
Not limiting to "plain" syntaxes in nested templates it would go from 4.2 to 12MB, which the "only" was compared with.

deathaxe:
This one maybe overcomplicates things.

PR #4325, opened 2025-07-12
[ZSH] Fix eval argument scopes

### LINKED ISSUE #4324: [Bash] Zsh function definition scope with `eval` extends to far
### What happened?

My background hilighting on `entity.name.function` runs amok on this.

```sh
# OpenSSL cert shorthands
ssl_hashes=( sha512 sha256 sha1 md5 )

for sh in ${ssl_hashes}; do
    eval 'ssl-cert-'${sh}'() {
        emulate -L zsh
        if [[ -z $1 ]] ; then
            printf '\''usage: %s <file>\n'\'' "ssh-cert-'${sh}'"
            return 1
        fi
        openssl x509 -noout -fingerprint -'${sh}' -in $1
    }'
done; unset sh
```

Image

deathaxe:
Resolves #4324

This PR adds a simple workaround to prevent certain functions being scoped bare function definition due to structure of their arguments. It includes functions whose string arguments may be likely to cause ambiguities.

### DIFF
--- a/ShellScript/Zsh/tests/syntax_test_scope.zsh
+++ b/ShellScript/Zsh/tests/syntax_test_scope.zsh
@@ -9726,6 +9726,7 @@ echo 'ssl-cert-'${sh}'() {
 #^ meta.function-call.arguments.shell meta.string.glob.shell string.quoted.single.shell
 #^ punctuation.definition.string.end.shell
 
+
 ###############################################################################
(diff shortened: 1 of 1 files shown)


deathaxe:
FWIW, waiting for #4307 to avoid merge conflicts

FichteFoll (approved):

deathaxe:
Ok, then the other way round.

PR #4328, opened 2025-07-25
[Rust] String format: support named width/precision and debug hexa type

TheClams:


deathaxe:
I'd vote for some more tests so identifiers are tested in all format parts - not just the width part.

PR #4331, opened 2025-08-02
[C] Recognise types ending in _t

midnightveil:
This is one place where VS Code does better with the default syntax, because it recognises the majority of custom types as-is without any sort of structural type analysis. (This also works better than analysis, as that necessarily requires the LSP to be able to perform analysis, which isn't always the case).

https://github.com/microsoft/vscode/blob/3aef8be5e46aa7e97e8b0ed115f6147cd0b24634/extensions/cpp/syntaxes/c.tmLanguage.json#L195-L198


jrappen:
please add tests

jrappen:
Also, since you suggest this to fix a specific issue, please reference it as in the GitHub docs:

<https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword>

midnightveil:
> Also, since you suggest this to fix a specific issue, please reference it as in the GitHub docs:

I didn't added fixes because it's only a partial fix; if I did fixes it would close the original issue.

EDIT: And I've now added a test.

deathaxe (approved):

PR #4332, opened 2025-08-02
[JavaScript] Fix premature regex parsing termination

systemcrash:
If the regexp contains an escaped `/` in it, the detection prematurely concluded and an expression like the following is incorrectly evaluated:
```
/^(\/\S+)\s+\S+\s+(\d+)/
    ^--- finishing here
```
Now the regexp evaluates correctly.

Unindented one level also to align with the rest of the file.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1467,23 +1467,23 @@ contexts:
       set: regexp
 
   regexp:
-      - meta_include_prototype: false
-      - meta_scope: meta.string.js string.regexp.js
-      - match: "/"
-        scope: punctuation.definition.string.end.js
-        set:
-          - meta_include_prototype: false
-          - meta_content_scope: meta.string.js string.regexp.js
-          - match: '[gimyus]'
-            scope: keyword.other.js
-          - match: '[A-Za-z0-9]' # Ignore unknown flags for future-compatibility
-          - include: immediately-pop
-      - match: '(?=.|\n)'
-        push:
-          - meta_include_prototype: false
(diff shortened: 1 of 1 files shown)


jrappen:
please add tests

systemcrash:
`syntax_test_js_regexp.js`?

JavaScript/JavaScript.sublime-syntax
```diff
@@ -1467,23 +1467,23 @@ contexts:
...
-          - include: immediately-pop
-      - match: '(?=.|\n)'
-        push:
-          - meta_include_prototype: false
-          - match: '(?=/)'
-            pop: 1
-          - include: scope:source.regexp.js
+    - meta_include_prototype: false
+    - meta_scope: meta.string.js string.regexp.js
+    - match: (?<![\\\/])/
```
deathaxe, line 1472: Lookbehinds are not supported by sregex and cause fallback to slower Oniguruma.

While Oniguruma is still supported for backward compatibility reasons, such patterns must be avoided in current syntax definitions.

systemcrash: Where can I find supported syntax? I used https://www.sublimetext.com/docs/syntax.html#match-patterns

deathaxe: Yes.

jrappen:
I don't understand. Why was this closed?

deathaxe:
Because a fix requires more than a one-liner.

systemcrash:
I found a fix, but it turns out the original clause was fine.

PR #4337, opened 2025-08-05
[C#] Allow any modifier order in nested methods

### LINKED ISSUE #4334: [C#] static async task method does not highlight correctly
### What happened?

Hi, when I declare a static method as async and returning a `Task` it seems to think there is a syntax issue.

Image

If I remove `Task` or `async` it seems to highlight correct, so I think it's something to do with the number of tokens before the method declaration.

This is the codes:

```csharp
        static async Task OnExportCommand(FileInfo outputfile, CancellationToken token)
        {
            try
            {
                Console.WriteLine("Starting export operation...");
                var bookmarks = service.GetAll();
                string json = JsonSerializer.Serialize(bookmarks, new JsonSerializerOptions { WriteIndented = true });
                await File.WriteAllTextAsync(outputfile.FullName, json, token);   
            }
            catch(Exception ex)
            {
                Helper.ShowErrorMessage([$"An unknown exception occurred.", 
                                         $"Error message {ex.Message}"]);
            }
        }
```

deathaxe:
Fixes #4334

This PR allows any number of `async`, `ref` and `static` in any order in front of nested method definitions.

### DIFF
--- a/C#/tests/syntax_test_GeneralStructure.cs
+++ b/C#/tests/syntax_test_GeneralStructure.cs
@@ -1460,7 +1460,7 @@ public class TestModifierOrder
 ///                   ^^^^^^^^^^^^^^^ entity.name.function
 ///                                  ^ punctuation.section.parameters.begin.cs
 ///                                                                               ^ punctuation.section.parameters.end.cs
-    static async Task OnExportCommand(FileInfo outputfile, CancellationToken token) => await Task.CompletedTask;
+    static async Task OnExportCommand(FileInfo outputfile, CancellationToken token) => await Task.CompletedTask {
 /// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - invalid
 /// ^^^^^^ storage.modifier
 ///        ^^^^^ storage.modifier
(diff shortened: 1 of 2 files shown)


C#/tests/syntax_test_GeneralStructure.cs
```diff
@@ -1460,12 +1460,26 @@ public class TestModifierOrder
...
 ///                                                                               ^ punctuation.section.parameters.end.cs
-    static async Task OnExportCommand(FileInfo outputfile, CancellationToken token) => await Task.CompletedTask;
+    static async Task OnExportCommand(FileInfo outputfile, CancellationToken token) => await Task.CompletedTask {
 /// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - invalid
 /// ^^^^^^ storage.modifier
 ///        ^^^^^ storage.modifier
 ///              ^^^^ support.type
 ///                   ^^^^^^^^^^^^^^^ entity.name.function
 ///                                  ^ punctuation.section.parameters.begin.cs
 ///                                                                               ^ punctuation.section.parameters.end.cs
```
keith-hall, line 1470: technically this isn't valid C# code, I would feel happier approving if we changed it to:
``````suggestion
    static async Task OnExportCommand(FileInfo outputfile, CancellationToken token)
/// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - invalid
/// ^^^^^^ storage.modifier
///        ^^^^^ storage.modifier
///              ^^^^ support.type
///                   ^^^^^^^^^^^^^^^ entity.name.function
///                                  ^ punctuation.section.parameters.begin.cs
///                                                                               ^ punctuation.section.parameters.end.cs
    {
                await NestedMethod();
```

PR #4343, opened 2025-08-26
[C#] Support C#14 extension members

keith-hall:
https://devblogs.microsoft.com/dotnet/csharp-exploring-extension-members/


jrappen:
- why did choose doc comments for tests?
- some of the tests will not run, as you set to use doc comments but some are declared as (usual) line comments

keith-hall:
Thanks, fixed

PR #4347, opened 2025-09-23
[Makefile] avoid partial keyword matches; add `undefine`

mitranim:
This fixes syntax highlighting for makefiles (such as [this one](https://github.com/jart/blink/blob/98f95e8383d1032eb4d2dc6aae937b23539e915e/Makefile)) which use `undefine`, which was broken due to partial matching of `define` (missing `\b`).


mitranim:
Applied suggestions and squashed.

deathaxe (approved):

PR #4348, opened 2025-09-25
[Ruby] fix highlighting of __END__

Andriamanitra:
It was being highlighted as `variable.other.constant.ruby` previously. ~~The Ruby interpreter also allows whitespace after `__END__` (but not before).~~

*edit: I was actually totally wrong about that last part, I had just set up my editor to automatically remove trailing whitespace when testing...*

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -111,6 +111,7 @@ contexts:
       pop: 1
 
   expressions:
+    - include: data-section
     - include: constants
     - include: class
     - include: module
@@ -122,7 +123,6 @@ contexts:
     - include: method
     - include: strings
     - include: comments
-    - include: data-section
     - include: heredocs
     - include: punctuation
     - include: operators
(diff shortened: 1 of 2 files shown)


michaelblyons:
You might tighten the scope assignment to just `__END__` with a capture group. Seems fine in general, though. Do you have a link to Ruby docs?

Andriamanitra:
Ruby docs don't have much to say about it: https://docs.ruby-lang.org/en/master/Object.html#DATA

michaelblyons:
YAML uses `entity.other.document.end` for something similar, but doesn't push into a text-only context after encountering it.

https://github.com/sublimehq/Packages/blob/76b1f045062d46da789e0c03d03197dce9062d1d/YAML/YAML.sublime-syntax#L226-L227

Does anyone else know of other comparable patterns and their scopes?

Andriamanitra:
My understanding is that Ruby got the idea for this feature from [Perl](https://perldoc.perl.org/perldata#Special-Literals). Perl's sublime-syntax seems to use `constant.language.perl`:
https://github.com/sublimehq/Packages/blob/3b2e548ca24d0a2057819b5832bc515fe97992df/Perl/Perl.sublime-syntax#L784-L788

PR #4352, opened 2025-10-01
[CSS] Simplify indentation rules

### LINKED ISSUE #4344: Bad indentation when an opening parenthese is quoted inside CSS rule
### Description of the bug

For accessibility, as [explained on a W3schools page](https://www.w3schools.com/cssref/pr_gen_content.php), a famous CSS trick when you print a webpage is to inserts the value of the href attribute in parenthesis after each <a> element like this:

```
a::after {
	content: ' (' attr(href) ')';
}
```

But when you want to add another CSS instruction into this bloc after the "content" line, the indentation breaks :

```
a::after {
	content: ' (' attr(href) ')';
		font-size: 10pt;
	}
```

I tried a lot of things like chaging double quotes to simples. The Edit > Line > Reindent option is also broken. It comes from the opening parenthesis "(" : if I delete it, indentation back right, but I need it and i can't fin a workaround!

### Steps to reproduce

Copy-paste the following CSS code:
```
a::after {
	content: ' (' attr(href) ')';
	font-size: 10pt;
}
```

Go Edit > Line > Reindent and see broken code indentation:

```
a::after {
	content: ' (' attr(href) ')';
		font-size: 10pt;
	}
```

### Expected behavior

The right indentation, like this:

```
a::after {
	content: ' (' attr(href) ')';
	font-size: 10pt;
}
```

### Actual behavior

Bad indentation:

```
a::after {
	content: ' (' attr(href) ')';
		font-size: 10pt;
	}
```

### Sublime Text build number

4200

macOS 13.7.7

deathaxe:
Resolves #4344.

This commit removes attempts to be too smart with regards to finding unbalanced brackets, as it doesn't work reliable enough using single patterns.

Instead just accept trailing block comments after an opening bracket to indent line after.

### DIFF
--- a/CSS/Indentation Rules.tmPreferences
+++ b/CSS/Indentation Rules.tmPreferences
@@ -13,25 +13,10 @@
 			^.*
 			# followed by opening bracket
 			[(\[{]
-			# optionally followed by...
-			(?<balanced>
-			# balanced parentheses
-			  \( \g<balanced>* \)
-			# balanced brackets
-		  	| \[ \g<balanced>* \]
-			# balanced braces
-			| \{ \g<balanced>* \}
-			# double quoted string with ignored escaped quotation marks
-			| \" .*? (?<![^\\]\\)(?<![\\]{3})\"
-			# single quoted string with ignored escaped quotation marks
-			| \' .*? (?<![^\\]\\)\'
-			# block comment
-			| /\* .*? \*/
-			# anything but closing brackets
-			| [^])}]
-			)*
+			# optionally followed by block comment
+			(?: \s* /\* ((?!\*/).)* \*/ )*
 			# ... until end of line
-			$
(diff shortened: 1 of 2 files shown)


michaelblyons:
Haha. Were you trying `(?R)` things for this when you made the other PR?

deathaxe:
Those (short) considerations were a result of finding `(?R)` while attempting to fix a highlighting issue in a regexp pattern used by LaTeXTools. I quickly dropped that idea though after those patterns caused significant lag when trying out in find panel.

Comparing to how most other syntaxes work, I just realized it being a dead end, to expect indenation, created from a single pair of regexp patterns, ever being able to be smart. So drop it and do what all the others do. Just handle the naive parts.

michaelblyons (approved):

PR #4357, opened 2025-10-05
[Perl/Ruby] Fix raw data sections

deathaxe:
This commit scopes `__END__` in Ruby/Perl, and `__DATA__` in Perl as end of script and beginning of raw literal data section.

Content after is scoped as plain text. 

Note: Ruby continues supporting html syntax.


keith-hall (dismissed):

michaelblyons:
Did you want to include `__DATA__` as well as `__END__` in one or both languages?

deathaxe:
Does Ruby support `__DATA__`? Haven't checked it, tbh, as even Perl supports it only for backward compatibility - recommending `__END__` instead.

michaelblyons (approved):

keith-hall (approved):

PR #4360, opened 2025-10-10
[C] Compile File with MSVC on Windows

saccarosium:
Hi,
I've noticed that the build file only works on machines that have `gcc` witch is not the case by default on Windows nor Mac. I think would be a good idea to use the default toolchains available on the system so maybe in a future commit we could add `clang` for mac.

Please tell me if I made a mistake on the syntax for defining os specific commands.

### DIFF
--- a/C++/C Single File.sublime-build
+++ b/C++/C Single File.sublime-build
@@ -7,6 +7,19 @@
 	"variants":
 	[
 		{
+			"name": "MSVC",
+			"shell_cmd": "cl /nologo \"${file}\"",
+			"file_regex": "^(.*)\\(([0-9]*)\\):( )(.*)",
+			"windows": true,
+		},
+		{
+			"name": "MSVC Run",
+			"shell_cmd": "cl /nologo \"${file}\" && ${file_base_name}",
+			"file_regex": "^(.*)\\(([0-9]*)\\):( )(.*)",
+			"interactive": true,
+			"windows": true,
+		},
+		{
 			"name": "Run",
 			"shell_cmd": "gcc \"${file}\" -o \"${file_path}/${file_base_name}\" && \"${file_path}/${file_base_name}\"",
 			"interactive": true,


michaelblyons:
It is also my understanding from [the documentation][docs] that the variants' `file_regex` does not need to be set, since it keeps the existing value. If someone knows otherwise for when there are multiple platforms, please mention it.

[docs]: https://www.sublimetext.com/docs/build_systems.html

saccarosium:
> It is also my understanding from [the documentation](https://www.sublimetext.com/docs/build_systems.html) that the variants' file_regex does not need to be set, since it keeps the existing value. If someone knows otherwise for when there are multiple platforms, please mention it.

I don't know. Is it sure that the `file_regex` is propagated even if not at the top level and only in the `"windows"` or `"linux"` keys?

deathaxe:
It is properly propagated.

Hierarchy is: global > platform > variant.

A global platform specific key is also only available in a variant's platform specific scope.

Hence this PR disables build system for MacOS in its current form.

I also wouldn't just disable gcc on Windows as it can also be used on that platform. Hence, if MSVC is to be supported, it should be done by a windows-only variant.

saccarosium:
> It is properly propagated.
>
> Hierarchy is: global > platform > variant.
>
> A global platform specific key is also only available in a variant's platform specific scope.

Got it! Makes sense

> I also wouldn't just disable gcc on Windows as it can also be used on that platform. Hence, if MSVC is to be supported, it should be done by a windows-only variant.

Do you mean having gcc in the top level and then having a variant like `MSVC` so when is displayed like `C Compile File - MSVC`?

deathaxe:
Yes, a windows-only MSVC variant on top of globally available gcc would probably make most sense.

saccarosium:
Ok I've made the changes, also I did add the `/nologo` flag that prevent the Microsoft watermark to show.

before
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35217 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

main.c
Microsoft (R) Incremental Linker Version 14.44.35217.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:main.exe
main.obj

```

after:
```
main.c

```

deathaxe (commented):
Regardless invalid build variant specification, it doesn't seem to be possible to hide MSVC variant from Linux/Windows.

While on Linux gcc is normally available via default environment, running "cl" compiler on Windows, requires proper environment to be set-up, first. Most likely by starting "Development Console". Hence required environment would be available only, when running ST from that console. Otherwise "cl" or required libs/includes are not found.

After all this build file is just a naive example, which is insufficient for most serious projects anyway.

It also doesn't provide clang/llvm commands as well. So maybe anything beyond this naive gcc example is probably better provided as separate (platform-specific) package.

saccarosium:
> It also doesn't provide clang/llvm commands as well. So maybe anything beyond this naive gcc example is probably better provided as separate (platform-specific) package.

Make sense but maybe clang should be provided as the default on macos but that is for another PR.

Thank you for the feedback!

PR #4362, opened 2025-10-20
[Python] Improve named backref regexp highlighting

deathaxe:
This commit prepares named backreferences for correctly handling interpolation.

Primary use case are raw f-strings, but it applies to all other sorts of template languages as well.

Notes:

1. to align with _Regular Expressions_ package, `keyword.other.backref-and-recursion` covers all, leading `?` and group names.
2. adds `?&backref` to support patterns from "regexp" package.

### DIFF
--- a/Python/Embeddings/RegExp (for Python).sublime-syntax
+++ b/Python/Embeddings/RegExp (for Python).sublime-syntax
@@ -13,10 +13,6 @@ variables:
   deactivate_x_mode: (?:\?[imsLua]*-[imsLua]*x[imxsLua]*)
   other_modifiers: \?(?:[ixmsLua]*-)?[ixmsLua]+
 
-  # capture group identifiers
-  capture_group_name: (?:[[:alpha:]_]\w*)
-  capture_group: (?:[1-9][0-9]*|{{capture_group_name}})
-
   # quantifiers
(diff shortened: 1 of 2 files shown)


michaelblyons:
Looks good.

Would interpolations still work if you added an `invalid.illegal` for `[^:[alnum]:_]` in `backref-name`? (Or are there other concerns for annoyance while typing? I would _expect_ any nonconforming characters to be inside the interpolations and not really a problem if interpolations are working.)

deathaxe:
With regards to related Discord discussion, it may require additional messures to ensure place holders like `%s` to not be highlighted illegal. Those are currently not included into with_prototype in Python. 

~~A point of concern is `[^:[alnum]:_]` pattern creating individual tokens for each character and `[^:[alnum]:_]+` potentially preventing interpolation from being highlighted, when preceeded by an non-word character. But maybe individual tokens being created may be unlikely and expected ammount small enough to be acceptable.~~

It doesn't currently produce good results, as normal patterns don't handle placeholders as those cause ambiguities with `{1}` quantifiers. Adding illegals results in ...

grafik

With current PR, they at least don't look too broken.

grafik

jrappen (approved):

PR #4363, opened 2025-10-21
[Markdown] Fix typing ``` in the middle of a document

jskinner:
Typing ``` was inserting ```` (except when typing at EOF), because the initial paired `` is not recognized by the syntax definition as a code block, and therefore the second backtick was inserting, rather than skipping over the paired character.

This change tweaks the key binding to work in this situation.

Additionally deleted the bindings that would delete ** and __ sequences with one keypress, because they're not auto-paired, so it doesn't make any sense to do this.

### DIFF
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -10,18 +10,6 @@
             { "key": "selector", "operand": "text.html.markdown - markup.raw" }
         ]
     },
-    {
-        "keys": ["backspace"],
-        "command": "run_macro_file",
-        "args": {"file": "res://Packages/Default/Delete Left Right.sublime-macro"},
-        "context": [
-            { "key": "setting.auto_match_enabled" },
-            { "key": "selection_empty", "match_all": true },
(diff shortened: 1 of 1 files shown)


FichteFoll (approved):
Great! This has been bothering me for a while but not enough to look into fixing it.

deathaxe:
FWIW, I don't understand the `**` auto-pairing arguments as removed backspace binding is the counterpart for auto-pairng bindings of bold/italic emphasis tags which are located directly above those removed hunks.

michaelblyons:
Before approving, i tested typing an undersxore. it did not pair for me, so i figured it was safe to remove the delete.

deathaxe:
_Selecting some text and typing `_` creates italics, for instance._

Auto-pairing is enabled for selected text only, as at least auto-pairing `*` requires some extra key bindings to get list items done right, which may be a bit beyond.

But maybe such kinds of smart features are better handled by dedicated plugins, such as MarkdownEditing, anyway.

I can't also reproduce any issues with regards to the primary issue which is said being fixed here, but this may be related with backticks being treated as dead keys on my german windows keyboard. At least it seems orthogonal and thus doesn't break anything.

PR #4364, opened 2025-10-21
[Markdown] Remove auto-pairing of asterisk and underscore

deathaxe:
As a consequence of #4363, this commit also removes key bindings for auto-pairing `_` and `*`.

### DIFF
--- a/Markdown/Default.sublime-keymap
+++ b/Markdown/Default.sublime-keymap
@@ -1,28 +1,4 @@
 [
-    // Auto-pair asterisks
-    {
-        "keys": ["*"],
-        "command": "insert_snippet",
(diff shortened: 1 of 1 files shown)


FichteFoll:
I'm against this change. While auto-pairing `*` and `_` does not happen while typing, it does make plenty sense to me (and also various online markdown editor inputs) to wrap the selection when these characters are input. Like it was mentioned in #4363, just typing `*` and `_` did not auto-pair and thus using backspace while in-between the pair should not undo it either.

deathaxe:
Both bindings belong to each other though. I finally don't care, as all of it is augmented by MarkdownEditing anyway.

PR #4368, opened 2025-11-01
[CSS] upgrade to py3.13

jrappen:


deathaxe:
From my understanding .python-versions is needed until python 3.3 plugin host will be removed in the future.

CSS/css_completions.py
```diff
@@ -87,16 +88,10 @@ def on_query_completions(self, view, prefix, locations):
...
         text = view.substr(sublime.Region(pt, view.line(pt).end()))
         matches = self.re_value.search(text)
-        if matches:
-            colon, space, value, term = matches.groups()
-        else:
-            colon = ""
-            space = ""
-            value = ""
-            term = ""
+        colon, space, value, term = matches.groups() if matches else ("", "", "", "")
```
deathaxe: This change is orthogonal, IMHO. 

I prefer the more explicit former form as it doesn't require unpacking to be applied in case of no match.

CSS/.python-version
```diff
@@ -1 +1 @@
-3.8
\ No newline at end of file
+3.13
```
deathaxe: ST 4201+ reads 3.8 as 3.13. Why not just leaving those files alone until py33 is removed?

FichteFoll: Having them say 3.8 when the files actually require 3.13 is just wrong, though, even when ST 4201+ can handle it. This should also prevent issues when trying to run these Python files in an older ST build that does not have the 3.13 runtime.

I would be in favor of just removing them, if they are not required, but the 3.3 host is not dead yet and still the default, afaik.

CSS/css_completions.py
```diff
@@ -1,9 +1,11 @@
-import re
 import sublime
 import sublime_plugin
+
+import re
 import timeit
 
 from functools import cached_property, wraps
+from typing import List, Optional, Tuple
```
FichteFoll: Same as in the other PR, importing List, Optional and Tuple is generally not necessary in 3.10+.

michaelblyons:
Force-pushing a new squashed commit every time makes reviewing more difficult.

CSS/.python-version
deathaxe: We don't need typing in such simple plugins and beyond that the modules and this set of PRs doesn't add anything which judges breaking things for users which probably use cloned repos - at least not until public betas are available.

FichteFoll:
In general, I would decide whether to squash/rebase a PR on a case-by-case basis, but since we squash all PRs on this repo anyway, avoiding force-pushes is really the better workflow here.

jrappen:
I'll fix the left-overs tonight, thanks for the feedback.

jrappen:
I'll redo this at a later date.

PR #4369, opened 2025-11-01
[Diff] upgrade to py3.13

jrappen:


jrappen:
I'll fix the left-overs later tonight, thanks for the feedback.

jrappen:
I'll redo this at a later date.

PR #4370, opened 2025-11-10
[JavaScript] Add key binding for parentheses auto-indentation

dpjohnst:
with the cursor between parentheses:

```javascript
    function myFunc(|)
```

becomes:

```javascript
    function myFunc(
        |
    )
```

after hitting <kbd>Enter</kbd>.

### DIFF
--- a/JavaScript/Default.sublime-keymap
+++ b/JavaScript/Default.sublime-keymap
@@ -90,6 +90,26 @@
         ]
     },
 
+    // Add indented line in parentheses
+    { "keys": ["enter"], "command": "run_macro_file", "args": {"file": "res://Packages/Default/Add Line in Braces.sublime-macro"}, "context":
+        [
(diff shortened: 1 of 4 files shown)


deathaxe (approved):
Given tests prove this PR being an improvement and this indentation style is already enforced this way in syntaxes like Go or R.

General questions (applying to all use cases) therefore are: 

1. Is support for `"indent_to_bracket": true` setting of any interest for those languages?
2. If "yes", how could it be implemented?

The setting changes argument indentation style to:

```
thisisafunction(test,
                foo,
                bar)
```

Key binding could be disabled for this indentation style via `"context"`, but it doesn't work for indentation rules.

jrappen (approved):

jrappen:
Slightly edited comments above to fix readability.

PR #4372, opened 2025-11-10
[Python] Add key binding to unindent line after break/continue/return

deathaxe:
This commit adds a key binding to auto-dedent after certain control flow keywords, which indicate end of block, in python.

source https://forum.sublimetext.com/t/automatically-unindent-in-python/77454


keith-hall (approved):

michaelblyons (approved):

jrappen (approved):

jfcherng:
Why we come up with this until now?

jfcherng (approved):

deathaxe:
Sometimes just the right questions need to be asked.

PR #4373, opened 2025-11-15
[C#] support C#14

keith-hall:


deathaxe (dismissed):

C#/C#.sublime-syntax
```diff
@@ -51,6 +51,8 @@ variables:
   name_normal: '{{start_char}}{{other_char}}*\b'
   cap_name: '(\p{Lu}{{other_char}})'
 
+  method_param_type_modifier: \b(out|ref|this|params|in)\s+
```
michaelblyons: Is there a good way to convert to a non-capturing group and drop the `\s+`? I think there is already a place where this has made a scoping mistake...

michaelblyons (approved):

PR #4389, opened 2025-11-21
[Ruby] Fix regexp patterns at the beginning of blocks

### LINKED ISSUE #3817: [Ruby] Incorrect syntax highlighting in Regexp literal with single quote
### What happened?

If a single quote (**'**) is placed inside a Regexp literal expression the code afterwards is highlighted as text

```ruby
let(:error_msg) { /can't be blank/ }

before do
  record.repository_owner = nil
  record.repository_name = nil
end
```
Captura desde 2023-08-03 12-30-21

Sublime Version: 4152
OS Version: Linux fedora 6.4.4-100.fc37.x86_64

deathaxe:
fixes #3817

This commit fixes `/.../` patterns appearing directly after `{` in blocks.

### DIFF
--- a/Ruby/Ruby.sublime-syntax
+++ b/Ruby/Ruby.sublime-syntax
@@ -484,7 +484,7 @@ contexts:
       scope: meta.block.parameters.ruby punctuation.definition.parameters.begin.ruby
       set: block-parameters
     - match: (?=\s*[^\s\|])
-      pop: true
+      set: try-regex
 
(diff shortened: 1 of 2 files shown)


jrappen (approved):

jrappen:
Next time a link instead of the issue number in the test comments would make more sense as reviewers can just right-click the link and open it in their default browser.

PR #4391, opened 2025-11-22
[LaTeX] Refine macro definitions

deathaxe:
This commit...
1. reduces required context switches in macro definitions
2. scopes brackets in macro signatures
3. scopes macro variables `#1` as `variable.parameter` in signatures and as `variable.other` in macro code.
4. drops stacked `storage.modifier` scopes from `\def` `\newcommand` etc. declaration keywords.


michaelblyons (approved):

jfcherng (approved):

jrappen:
thanks for this one, appreciate it

PR #4394, opened 2025-11-23
[LaTeX] Fix math variable scopes

deathaxe:
This commit removes `markup.italic` scope from math variables.

It is up to a color scheme to decide, whether a scope is to be rendered with italic fonts. `markup` scope is dedicated for plain text markups.

### DIFF
--- a/LaTeX/TeX.sublime-syntax
+++ b/LaTeX/TeX.sublime-syntax
@@ -735,7 +735,7 @@ contexts:
 
   math-variables:
     - match: '[A-Za-z]+'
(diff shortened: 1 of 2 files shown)


michaelblyons:
For the record: I'm in favor of this change, but I expect it to get some dispute after release.

deathaxe:
FWIW, `markup.italic` was introduced by the patchset collected by PR#3575 and is released in ST4174+.

It was neither part of orignal TextMate syntaxes, nor is it scoped this way by VS Code or Zed.

PR #4396, opened 2025-11-23
[LaTeX] Fix internal macro names

deathaxe:
This commit...

1. adds support for internal macro names, which in addition to normal commands, may contain `_` and `:` for structuring purposes.

   Typesetting commands and internal macro names are assigned distinct scopes.

2. changes scope of command definitions to `entity.name.function`.

### DIFF
--- a/LaTeX/LaTeX.sublime-syntax
+++ b/LaTeX/LaTeX.sublime-syntax
@@ -181,6 +181,11 @@ contexts:
     - meta_scope: meta.function.body.latex
     # but in principle, a single token is also valid,
     # thus either a command sequence ...
+    - match: (\\){{macroname}}
+      scope: support.function.internal.latex
+      captures:
+        1: punctuation.definition.backslash.latex
+      pop: 1
     - match: (\\){{cmdname}}
       scope: support.function.general.latex
       captures:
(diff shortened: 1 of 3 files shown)


deathaxe:
FWIW, commands may only consist of letters (char category 11). By default, this excludes `_`.

Various macrofiles in common TeX distros however include `_` into letter categories by following commands.

```
\chardef\letter=11
\chardef\uscode=\catcode`\_
\catcode`\_=\letter
```

Another prominent use case is EXPL3 language specification for internal LaTeX3 macro names. (see: https://texdoc.org/serve/expl3.pdf/0)

EXPL3 commands require programming layer to be loaded e.g. via `\ExplSyntaxOn`.

Most common files, those internal macro names are used within, are \*.cls, \*.sty or even DocTeX (\*.dtx) files. There are however also some \*.tex files containing them.

The question arising with all of that is: Do we accept possible false positives in normal LaTeX documents by just scoping identifiers with underscore as commands/macros, or would it be useful to create a dedicated "LaTeX Macro", "LaTeX Class" or "LaTeX EXPL3" syntax, which focuses on the programming layer rather than typesetting?

PR #4397, opened 2025-11-24
[Diff] Ensure word boundaries on hashes

michaelblyons:


jrappen:
On mobile right now and haven't checked, but are you sure all these forms of valid hashes would count as valid words?

Looking at the settings for Diff, word boundaries haven't been changed so they'd be the default ones.

michaelblyons:
> On mobile right now and haven't checked, but are you sure all these forms of valid hashes would count as valid words?

This is not the `"word_separators"` setting. It's requiring a `\b` (change from `\w` to `\W` or vice versa) at the edges of the Git SHA so `[0-9a-f]{40}` doesn't match in the middle of a >40-char string.

michaelblyons:
The biggest functional effect[^1] is for SHAs beginning with 7 zeroes[^2] not to have that section flagged as `constant.language.null`.

[^1]: and maybe I should add a test for this...
[^2]: Extremely rare, but not impossible

jrappen (approved):

deathaxe (approved):

PR #4400, opened 2025-11-28
[LaTeX] Simplify character-code contexts

deathaxe:
This commit...

1. removes obsolete/unused character-code-assignment context
2. replaces character-code-meta context by assigning meta scope via character-code-value context.


jrappen:
nice catch

PR #4401, opened 2025-11-28
[LaTeX] Refactor TeX macro definitions and introduce common braces

deathaxe:
This PR...

1. renames and refactors macro definition related contexts to `macro-...`.

2. merges `macro-braces` and `braces` related contexts, as allowed content is
   equal. They may contain any command, but included contexts must be prepared
   to match partly, only. So huge `\begin...\end` environments with custom
   syntax highlighting may be candidates to exclude.

3. introduces an `expressions` context containing only simple patterns, which
   don't push large contexts onto stack and therefore may be used in any nested
   context, such as braced groups.

4. includes `expressions` context to all appropriate locations.

5. moves simple patterns from `latex` to `expressions`, which may appear in
   nested braces.

6. splits external package related contexts into blocks/envs and inline commands
   to include them into `main` and `expression` accordingly.

The goal is to have a single brace-group-body context, which fits for almost
all needs, to help improve scoping consistency.

### DIFF
--- a/LaTeX/DocStrip.sublime-syntax
+++ b/LaTeX/DocStrip.sublime-syntax
@@ -15,6 +15,9 @@ contexts:
   main:
     - meta_prepend: true
     - include: docstrip-preamble
+
+  expressions:
+    - meta_prepend: true
     - include: docstrip-config
     - include: docstrip-user-io
     - include: docstrip-keywords
@@ -71,7 +74,7 @@ contexts:
 
   docstrip-file-argument-path:
     - meta_content_scope: meta.path.tex.docstrip
-    - include: macro-braces-body
+    - include: brace-group-body
 
   docstrip-config:
     - match: (\\)(?:usedir|showdirectory|BaseDirectory|DeclareDir|UseTDS|maxfiles|maxoutfiles){{endcs}}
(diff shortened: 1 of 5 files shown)


michaelblyons:
RE (2) the changes to test scopes look good. Is there anything you are concerned about getting included in blocks that wasn't there before?

deathaxe:
As literly any command seems to be evaluated in any location, it should be ok or even required to scope them anywhere. Does it make sense to support each command in any braced group? Semantically probably not, but as we can't add patterns for each latex command, we actually may not have enough information to decide when to include which command.

This is what I've seen in various situations. Specifically supported commands being scoped inconsistently, just due to `general-commands` being used in various contexts, only.

Most commands pop as early as possible. Hence I don't expect this change to break syntax.

It just increases syntax cache size due to more rules being duplicated in more contexts. But this topic may be addressed at a later point by using a `(?=\\)` lookahead to group all commands in a single context. This requires some more cleanup, first, though.

deathaxe:
I am heading towards a consistent base context structure which we can build (or refactor existing) special-purpose contexts for certain command upon. If those special commands will be known to only support basic commans, I'll fine tune included commands further.

michaelblyons (dismissed):

deathaxe:
Seems we may need to be careful with environments `\begin...\end`, which push dedicated contexts on stack. This may cause issues in statements like `\cmd[key=[\begin{...}]]{...}` which appears in texmf's markdown.dtx

deathaxe:
Here's a 2nd revision with a common "braces" approach and a new `expressions` context dedicated for usage within braces.

jrappen (approved):

PR #4413, opened 2025-12-10
[Python] Unify block begin punctuation

deathaxe:
This commit scopes all colons `punctuation.section.block.begin`, which start an indented block of code, regardless their actual context.

The point is colon indicating a start of a code block, like an opening brace in a C-style language. We just don't scope the block `meta.block` as we don't want or can't reliably count indentation to pop it off, correctly.


deathaxe:
Exactly, what I wouldn't have expected anyone to do :-D  

FWIW, most of them are covered by corresponding meta scopes, if distinct colors is desired.

PR #4416, opened 2025-12-18
[C#] embed SQL instead of using with_prototype

keith-hall:
https://github.com/SublimeText/PowerShell/issues/213

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1975,10 +1975,8 @@ contexts:
   inside_triple_quoted_raw_string_literal_syntax:
     - meta_include_prototype: false
     - match: (?={{sql_indicator}})
-      set: scope:source.sql
-      with_prototype:
-        - match: (?=""")
-          pop: true
+      embed: scope:source.sql
+      escape: (?=""")
     - match: (?=\S)
       pop: true
 


michaelblyons:
Are you going to do long string and/or long format string as well?

keith-hall:
Probably we will want to extend the SQL syntax for C# embedding and add some string formatting contexts for those, it isn't quite as straightforward. I agree that it needs to be done, probably this PR doesn't bring much benefit without it... Maybe a job for my next chunk of spare time 😅

michaelblyons:
I'm not sure if you saw that my "Ripping SQL out of C#" comment on SublimeText/PowerShell#213 was accompanied by https://github.com/SublimeText/PowerShell/commit/cf8df095f2485cacb1c24dfcd3e5a0889f66ebd7 that actually does so when embedding the C#.

I don't know how valuable it is to PowerShell users to have the fancy SQL embedding in their C# strings. And it's not like they're losing something they were used to. So no pressing need.

deathaxe:
Extending and embedding SQL

pro:
- core C# syntax becomes smaller again
- more control over which SQL contexts format string placeholders etc. are injected into
- SQL is loaded only, if used in a file

contra:
- overall package size significantly increases as each SQL variant contains various C# contexts, required to render extended placeholders or interpolation patterns

_FWIW, I also have a Python (and C#) branch with SQL embedded insead of being included. Current Python with SQL  included is about 1.5MB. Embedding SQL shrinks Python to 150kB, but increases whole package's syntax cache size to about 7MB._

It is however the more bullet proove solution, regardless overall size, unless any other syntax pushes C# (or Python) using `with_prototype`.

deathaxe:
For those looking for some inspiration ... https://github.com/deathaxe/sublime-packages/tree/pr/cs/embed-sql

Current C# is about 1.4MB of size.

With SQL embedded (3x), each SQL is of about 1MB of size + C# using about 740kB.

michaelblyons:
Are caches generated for embedded syntaxes whose activating pattern is never encountered?

deathaxe:
Not until they are loaded, which they are when running syntax tests. Comparing cache sizes, may primarily indicate how good ST can dedup or re-use existing patterns to achieve a certain kind of highlighting. The larger caches are the more distinct patterns are involved, which may also increase RAM usage. It may however also be of no sense, to compare them.

michaelblyons:
See also #4464

deathaxe:
FWIW, embedding SQL into C# doesn't help with PowerShell etc. failing due to context sanaty limits being hit, as each `embed` is converted into `push...with_prototype` if any including syntax (e.g. PowerShell) includes C# using `with_prototype`.

michaelblyons:
Yeah, PowerShell still has a TODO at https://github.com/SublimeText/PowerShell/issues/218 to fix that part on their side.

jrappen:
⚠️ superseded by #4547 (which has already been reviewed and approved)

PR #4420, opened 2025-12-25
[JavaScript] Use more-specific scopes for in/of/is

### LINKED ISSUE #4419: [JavaScript] Use more-specific scopes for `in`/`of`
### What happened?

When switching to JavaScript from other languages, I routinely type `for x in y` instead of `for x of y`. I make this mistake so often that I made a custom JavaScript syntax to highlight `in`/`of` in different colors.

In the default syntax, both the `in` and `of` keywords are assigned the `keyword.operator.word.js` scope.

In my syntax, I assign scopes as follows:

| Keyword | Scope |
| - | - |
| `in` | `keyword.operator.word.in.js` |
| `of` | `keyword.operator.word.of.js` |

This matches `new` being assigned `keyword.operator.word.new.js`, it also matches other keywords which get specific scopes.

Could the default syntax adopt this change and use more-specific scopes? I'd be happy to submit a pull request if needed.

iccir:
This implements #4419 and updates all test cases. For consistency, I also changed the `is` operator in TypeScript to be `keyword.operator.word.is.js` (I can back this out if needed). All `keyword.operator.word.js` scopes should now be specific.

Closes #4419

### DIFF
--- a/JavaScript/TypeScript.sublime-syntax
+++ b/JavaScript/TypeScript.sublime-syntax
@@ -854,7 +854,7 @@ contexts:
       scope: keyword.operator.type.union.js
       push: [ts-type-expression-end-no-line-terminator, ts-type-expression-begin]
     - match: is{{identifier_break}}
-      scope: keyword.operator.word.is.js
+      scope: keyword.operator.type.is.js
       push: [ts-type-expression-end-no-line-terminator, ts-type-expression-begin]
     - match: as{{identifier_break}}
       scope: keyword.operator.type.js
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

JavaScript/JavaScript.sublime-syntax
```diff
@@ -983,8 +983,11 @@ contexts:
       fail: for-in-of
 
   for-in-of-word:
-    - match: (?:of|in){{identifier_break}}
-      scope: keyword.operator.word.js
+    - match: in{{identifier_break}}
+      scope: keyword.operator.word.in.js
+      pop: 1
+    - match: of{{identifier_break}}
+      scope: keyword.operator.word.of.js
```
deathaxe: When changing scopes, I'd suggest to follow general scoping used for such operator keywords in other languages, such as [Python](https://github.com/sublimehq/Packages/blob/6a88bab4a59e6c5127bb89348232426bf135cf0c/Python/Python.sublime-syntax#L2288-L2289) instead of inventing something new again.

With a look onto the context name, it however might mean a bit more work as we now distinguish e.g. `in` as operator in conditional expressions and `in` keyword in for-loops.

`in` targets objects, `of` targets type checks.

PR #4421, opened 2025-12-27
[JavaScript] Use more specific scope for logical in keyword

deathaxe:
This commit scopes logical `in` keyword `keyword.operator.logical`, following the scheme of PR #4227 and #4271 as a follow-up of #4420.

It fixes a syntax test, using invalid syntax `for (var in list)`, in which `in` was scoped logical keyword due to `var` not being a valid iterator variable.

### DIFF
--- a/JavaScript/JavaScript.sublime-syntax
+++ b/JavaScript/JavaScript.sublime-syntax
@@ -1567,7 +1567,7 @@ contexts:
       scope: keyword.operator.js
       push: expression-begin
     - match: in{{identifier_break}}
-      scope: keyword.operator.js
+      scope: keyword.operator.logical.js
       push: expression-begin
     - match: '=(?![=>])'
       scope: keyword.operator.assignment.js
--- a/JavaScript/tests/syntax_test_js.js
+++ b/JavaScript/tests/syntax_test_js.js
@@ -1338,7 +1338,7 @@ a = /\//u + 0;
 
     x
     in y;
-//  ^^ keyword.operator
+//  ^^ keyword.operator.logical
 
     x
     instanceof y;
--- a/JavaScript/tests/syntax_test_js_bindings.js
+++ b/JavaScript/tests/syntax_test_js_bindings.js
@@ -280,6 +280,6 @@ using [ x ] = 0;
 await using in x;
 //^^^ keyword.control.flow.await
 //    ^^^^^ variable.other.readwrite
-//          ^^ keyword.operator
+//          ^^ keyword.operator.logical
 //             ^ variable.other.readwrite
 //              ^ punctuation.terminator.statement
--- a/JavaScript/tests/syntax_test_js_control.js
+++ b/JavaScript/tests/syntax_test_js_control.js
@@ -179,7 +179,7 @@
 //  ^^^ keyword.control.loop.for
 //      ^^^^^^^^^^^^^^ meta.group
 //       ^ punctuation.separator.expression
-//           ^^ keyword.operator
+//           ^^ keyword.operator.logical
 //                  ^ punctuation.separator.expression
 
     for (a[x in list];;;) {}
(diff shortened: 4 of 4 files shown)


michaelblyons (approved):

jwortmann:
Note that this is a step backwards regarding #3694, because until now it turned out that all keywords with the `keyword.operator.js` scope were word operators, so color schemes could use that full scope to highlight them differently from operator symbols like for example `&&`, which all use a more specific (sub-)scope.

Apart from that, `in` is in fact not a logical operator, but rather a comparison operator (if it should be put into one of the existing subscope categories), because it kind of compares the value on the left side with the iterable on the right side. The defining property of a logical operator in the mathematical sense, but afaik also for programming languages, is that it combines one or more *logical variables* (or expressions that can be evaluated as a boolean value). The fact that the resulting value also turns out to be a boolean is merely a byproduct, and it is the case as well for some other operators, like the `in` here. The logical operators are things like NOT, AND, OR, XOR, see e.g. https://en.wikipedia.org/wiki/Logical_connective.

For instance in the Python docs `in` is listed next to things like `<`, `<=`, `==` as a comparison: https://docs.python.org/3/reference/expressions.html#operator-precedence

One could maybe think about a new sub-category `keyword.operator.membership` for the operators like `in` and `instanceof`, but within the existing scope categories "comparison" is more appropriate than "logical".

deathaxe:
> this is a step backwards regarding https://github.com/sublimehq/Packages/issues/3694,

Linked RFC was not agreed on and this PR just aligns scope of `in` keyword with various other languages. 

> ... is listed next to things like <, <=, == as a comparison: https://docs.python.org/3/reference/expressions.html#operator-precedence

Operator precedence doesn't say anything about operator types. It just says something about priority or order of execution.

> Apart from that, in is in fact not a logical operator, but rather a comparison operator 

This can actually be argued about. `comparison` was considdered earlier, as `iterator` or `iteration` was, but finally dropped. It is not a comparison, because it just checks whether value of an operator is within a given sequence of elements, returning a logical result. This is a higher level of logical operation with an iteration involved.

> One could maybe think about a new sub-category `keyword.operator.membership`

Possibly, but actually it would be wrong to scope `instanceof` with that scope, as `instanceof` just logically checks, whether given operator matches a given type.

jwortmann:
> It is not a comparison, because it just checks whether value of an operator is within a given sequence of elements, returning a logical result. This is a higher level of logical operation with an iteration involved.

> Operator precedence doesn't say anything about operator types. It just says something about priority or order of execution.

Well, the description in that table does:

> Comparisons, including membership tests and identity tests

The word "including" means that membership tests are considered to be part of the comparisons category. And if you open the section on comparison operators https://docs.python.org/3/reference/expressions.html#comparisons, you can see that `in` is formally defined as a "comp_operator":

> comp_operator: "<" | ">" | "==" | ">=" | "<=" | "!="
>               | "is" ["not"] | ["not"] "in"

---

I rarely use JavaScript and I don't know if there is an official documentation/specification for it, but the website at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#relational_operators also lists both `in` and `instanceof` under "Relational operators", right next to `<`, `<=`, etc., and not under the "Binary logical operators" section.

---

You can deny it if you want, and put some more :thumbsdown: at my comments if that makes you happy :shrug:
But when I write that `in` is not a logical operator, I'm not pulling that out of thin air.

---

It seems that the D syntax in this repository scopes `in` as a comparison operator as well:

https://github.com/sublimehq/Packages/blob/4c73d165a9785064ec4c491ed442d1b0578cfd43/D/D.sublime-syntax#L1551-L1552

Also there is a typo at

https://github.com/sublimehq/Packages/blob/4c73d165a9785064ec4c491ed442d1b0578cfd43/R/R.sublime-syntax#L231-L232

which seemingly let that scope slip through for the R syntax in #4271.

michaelblyons:
I think jwortmann makes some compelling points, but I also think that `in` should be consistent across languages.

I don't have an answer to what should be used.

deathaxe:
Than lets keep it as is, if that's the prefered solution.

FichteFoll:
FWIW, I also think that a `comparison` scope is more appropriate for a membership check than `logical`, which to me is reserved to boolean operations.

PR #4424, opened 2025-12-30
[Git Formats] Improve Git Log diffs

michaelblyons:
- Fix diff sections not recognized when using `git log -p --cc`, `git log -p --combined`
- Use specialized header metadata from Git Diff. ~~We might want to change the trailing scopes on some of these to `.git` instead of `.diff` and/or further move them to Git Common.~~ *done*
- Refactor the decorate content, and explicitly support tags. ~~I'm not super-happy with the scopes here either, but I didn't want to rock the boat too much.~~ *now sporting updated scopes*

~~I'm also a little annoyed by the `markup.raw` in the diff headers, but I didn't want to change that without discussion.~~ *Scopes changed by deathaxe*

### DIFF
--- a/Git Formats/Git Log.sublime-syntax
+++ b/Git Formats/Git Log.sublime-syntax
@@ -42,15 +42,16 @@ contexts:
     - match: \bHEAD(?![\w-])
       scope: support.type.git.log
     - match: ->
-      scope: punctuation.separator.mapping.git.log
+      scope: keyword.operator.reference.git.log
     - match: \b(?:master|main)\b
-      scope: entity.other.branch-name.master.git.log
+      scope: variable.language.master.git.log
     - match: \b(tag)(:)\s*([\w./-]+)
       captures:
+        1: storage.type.label.git.log
         2: punctuation.separator.log
-        3: entity.other.tag.git.log
+        3: entity.name.reference.tag.git.log
     - match: ([\w.-]+)(?:(/)([\w.-]+))?
-      scope: entity.other.branch-name.git.log
+      scope: entity.name.reference.branch.git.log
     - match: ','
       scope: punctuation.separator.sequence.git.log
 
--- a/Git Formats/Git Commit.sublime-syntax
+++ b/Git Formats/Git Commit.sublime-syntax
@@ -60,7 +60,7 @@ contexts:
     - match: \b({{on_branch}})\s+(.*)
       captures:
         1: markup.heading.on-branch.git.commit
-        2: constant.language.branch-name.git.commit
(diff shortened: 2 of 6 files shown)


deathaxe:
I've been keen enough to push a change suggestion to hopefully simplify context switching a bit.

`markup.raw` was primarily added as most color schemes add a slighly tweaked background color and I wanted to make headers stand out. I aggree it being a bit off, as there's no semantic or syntactical reason for it.

The `meta.header.git.commit` scope was choosen to seemlessly integrate with git commit message scopes `meta.subject.git.commit` and `meta.message.git.commit`. Adjusting it to `meta.block.header` like diff's headers block, would probably also be fine.

I agree with email-headers' trailing scope being `.diff` looking a bit off in Git Log. I'd therefore support moving them to Git Common and changing trailing scope to something common.

michaelblyons:
I don't mind `meta.header.git.commit`. But the `markup.raw` was low contrast with my `source source` rules,  making it hard to see where the next commit came.

Your changes look good. Thanks for those.

I can refactor the headers into Common.

Any opinion on scope for the word "tag"? It's not a mapping because you can have multiple. Keyword?

Lastly, I am tempted to change tag and branch names to `entity.name.label` and `HEAD` to `variable.language`, but I don't have to do that.

Git Formats/Git Log.sublime-syntax
```diff
@@ -7,67 +7,65 @@ version: 2
...
-      push:
-        - meta_scope: meta.header.git.commit markup.raw.block.git.log
-        - meta_content_scope: string.unquoted.log
-        - match: \n
-          pop: 1
-        - include: Git Common.sublime-syntax#email
-    # using push instead of include as workaround for
-    # https://github.com/SublimeTextIssues/Core/issues/2395
+        2: punctuation.separator.log
+        3: entity.other.tag.git.log
```
deathaxe: > Any opinion on scope for the word "tag"?

This one?

`entity.other.tag` was just used as branch names are also scoped `entity.other`.

It's actually not a mapping, hence the `->` above may also just be simple separator.

In general I have no strong opinion about best choise for scopes as all git related syntaxes are such special purpose, they don't fit into common language's scope schemes.

I guess that's the reason for `entity` instead of `variable` vs. `constant`. But well, the git commit hash is also scoped `constant`. 

We should probably only try to use same sort of scopes in all git syntaxes.

michaelblyons: I meant this part with the part you quoted...

```
  tag: 0.1.0
  ^^^
```

But I am also interested in the tag itself, so no worries.

PR #4435, opened 2026-02-04
[TOML] Fix missing scope before first tables

### LINKED ISSUE #4434: [TOML] No root scope before a first table
### What happened?

Most scopes before a first table don't start with `source.toml`:

Image

It's not noticeable with a default theme. But if you have custom overrides (which usually start with `source.xxx` prefix), they won't be applied to every part of a toml file.

deathaxe:
Fixes #4434

This commit sets away from main context to ensure `source.toml` is never removed by clear_scope directives in e.g. table-keys.


jrappen:
@deathaxe Thanks, appreciate the quick fix.

PR #4441, opened 2026-02-14
[Python] Reorganize variables

deathaxe:
This is a pure maintanance PR to...

1. move global `variables:` context to the bottom 
2. merge some uniquely used contexts into `statements`

It does not change any pattern or modify behavior.

Primary intent is to move the huge variables list to the bottom, as it is less important compared to normal contexts.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -36,243 +36,6 @@ first_line_match: |-
   | ^ \s* \# .*? -\*- .*? \bpython(?:\d(?:\.\d+)?)?\b .*? -\*-  # editorconfig
   )
 
-variables:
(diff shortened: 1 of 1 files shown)


michaelblyons:
1. Makes sense. Move the variables down. :+1: 
2. Is there a performance gain for consolidating statements?

    The way the contexts are now will let downstream inheritors prepend/append in more places. I don't have any intention to do so myself, and I guess they can always replace the whole `statements` context if they need to.

    I'm not opposed, just curious.

deathaxe:
All kinds of statements (optionally) push dedicated contexts onto stack if certain keyword is consumed. There's no benefit of separating them. Prepending to `statements` is sufficient. Merging saves some disk space as ST doesn't need to preserve dedicated precompiled patterns for the two separate contexts.

michaelblyons (approved):

PR #4446, opened 2026-03-02
[C#] correctly scope generics inside annotations

keith-hall:


deathaxe:
The scope changes seem to have quite noticible impact on various color schemes, due to `support.type` becoming the dominant scope, rather than `variable.annotation`.

before:

grafik

after:

grafik

Java at least scopes final identifier parts `variable.annotation` to respect scope naming guidelines, which explicitly advice that scope.

Python scopes qualifiers, `variable.annotation`, too, to ensure them to follow a quite distinct highlighting by only addressing that scope.

The overall question is - do we want to follow a somewhat common scope schema of annotations in all langauges?

Annotations are just expressions (mostly function calls) in Python and classes in C#/Java, but stacking `variable.annotation support.type`, etc. seems to be to much. We could completely drop `variable.annotation` stuff and just scope them as normal expressions, but that's not what I'd expect for highlighting to happen, at least.

keith-hall:
Thanks for the feedback, I have now reverted the `support.type` change and left `meta.annotation` as the dominant scope

PR #4452, opened 2026-03-07
[Python] Scope PascalCased identifiers as classes

deathaxe:
This is a rather opinionated PR to align highlighting of modern python3 code with languages like C# or Java and Github's way of treating pascal case identifiers as classes.

The suggestion relies on modern code style conventions to distinguish ordinary variables or functions from class references or instantiations.

Python itself doesn't technically require classes to be pascal case or variables to be snake_case. Thus a syntax focused purly on technical constraints might choose a common scope for all identifiers.

This PR however increases consistency of highlighting class names accross all locations, assuming proper code styling being applied.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -614,8 +614,22 @@ contexts:
       push:
         - function-call-argument-list
         - unqualified-function-call-name
-    - match: '{{identifier}}'
+    - match: (?={{identifier_start}})
+      push: unqualified-base-class-name-content
+
+  unqualified-base-class-name-content:
+    - include: builtin-exception
+    - include: builtin-type
+    - include: class-variable
+    - include: magic-function
+    - include: magic-variable
+    - include: wildcard-variable
+    - include: constant-name
+    - match: '{{identifier_class}}'
       scope: entity.other.inherited-class.python
+      pop: 1
+    - include: generic-name
+    - include: immediately-pop
 
   qualified-base-class-separator:
     - meta_include_prototype: false
@@ -640,9 +654,13 @@ contexts:
       set:
         - function-call-argument-list
         - qualified-function-call-name
-    - match: \s*({{identifier}})
+    - match: \s*({{identifier_class}})
       captures:
         1: entity.other.inherited-class.python
+      pop: 1
+    - match: \s*({{identifier}})
+      captures:
+        1: variable.other.python
       pop: 1
     - include: line-continuation-immediately-pop
 
--- a/Python/tests/syntax_test_python.py
+++ b/Python/tests/syntax_test_python.py
@@ -2676,17 +2676,23 @@ class MyClass(Inherited, \
 #   ^^^ comment.block.documentation.python
     pass
 
-class MyClass(*bases, *more_bases):
+bases = [Foo, Bar]
+# <- variable.other.python
+#^^^^ variable.other.python
+#        ^^^ support.class.python
+#             ^^^ support.class.python
+
+class MyClass(*bases, *MORE_BASES):
 #^^^^^^^^^^^^ meta.class.python
 #^^^^ keyword.declaration.class.python
 #     ^^^^^^^ entity.name.class.python
 #            ^^^^^^^^^^^^^^^^^^^^^ meta.class.inheritance.python
 #            ^ punctuation.section.inheritance.begin.python
 #             ^ keyword.operator.unpacking.sequence.python
-#              ^^^^^ entity.other.inherited-class.python
+#              ^^^^^ variable.other.python
 #                   ^ punctuation.separator.inheritance.python
 #                     ^ keyword.operator.unpacking.sequence.python
-#                      ^^^^^^^^^^ entity.other.inherited-class.python
+#                      ^^^^^^^^^^ variable.other.constant.python
 #                                ^ punctuation.section.inheritance.end.python
 #                                 ^ meta.class.python punctuation.section.block.begin.python
 


keith-hall (approved):

FichteFoll:
I am generally in favor of this change – it has been something I wanted to explore for a long time – but I wonder about whether using `support.class` is what we want to use even for user-defined classes. Is this also how it is done for C# and similar? Aren't we losing the differentiation between built-in classes and user-defined classes like this? Is there precedence for a `variable.class` scope?

This is one of the core concerns of #1842, after all, and affects every class usage in Python, so I'd need at least some form of a survey before approving. I'd also like to try it out locally first, but no promises right now as to when that would happen (ideally over the next week).

deathaxe:
I haven't seen `variable.class`, yet, and it feels odd, as a class is a data type, but not a variable. I wouldn't even want to consider such change for those reasons.

`support.class` is however what all syntaxes are using. An attempt to distinguish user defined classes from built-in classes in Java has been heavily criticised and thus was reverted to exclusively using `support.class`.

Following the scope naming guidelines basic built-in types, like `str`, `int`, ... would have to be scoped `storage.type`.

The primary issue with `support` is it having been mis-interpreted recently as being something "built-in" or provided by languages. According to original TextMate specs, it rather targets all user generated entities, created either by libraries or user code.

A distinction between user generated or built-in classes could be achived via 3rd-level scopes, which is still the case for exceptions. This PR just removes collections of `typing` module classes, which are not built-in but provided by a library.

FichteFoll:
> I haven't seen `variable.class`, yet, and it feels odd, as a class is a data type, but not a variable. 

I can make the same argument for `variable.type`, which matches how we attach meaning to arbitrary identifiers like `variable.constant` or `variable.function`, but I also know that this is not how it was used historically and it would definitely be a breaking change that I would not feel comfortable making in some PR.

> `support.class` is however what all syntaxes are using.

No, Rust uses `storage.type`.

> Following the scope naming guidelines basic built-in types, like `str`, `int`, ... would have to be scoped `storage.type`.

I generally do not have an issue with that, but my preference would be to differentiate identifiers (even identical ones) depending on how they are used. For example, in Python I would argue to scope type annotations with `storage.type` but class instantiation as `support.type` in `s: str = str(1)`. I made the same argument in #1842 already, so if we want to discuss this topic more, perhaps we should move the discussion there.

Just for the Record, Rust currently does *not* differentiate here, partially because I did not put in the effort, partially because I did not consider using `support` for user-defined identifiers to be proper, and partially because there was no precedence for highlighting a type differently depending on where it was used (outside of inheritance). At least I think so since it's been a while.

> The primary issue with `support` is it having been mis-interpreted recently as being something "built-in" or provided by languages. According to original TextMate specs, it rather targets all user generated entities, created either by libraries or user code.

That isn't exactly a "recent" thing because the [TextMate manual](https://macromates.com/manual/en/language_grammars#naming_conventions) that we've been referencing for scope names for a long time has been saying the following for the past 15 years or so:

> - support — things provided by a framework or library should be below support.
>   -  function — functions provided by the framework/library. For example NSLog in Objective-C is support.function.
>   -  class — when the framework/library provides classes.

This very much sounds like it is about something that a, well, framework or library provides. Whether the "standard library" also counts as a library may be a point of debate, but I would say yes. 

> A distinction between user generated or built-in classes could be achived via 3rd-level scopes, which is still the case for exceptions.

Sounds good to me. We should document this at some place, though.

> This PR just removes collections of `typing` module classes, which are not built-in but provided by a library.

That is a reasonable consequence when all class-like identifiers are scoped the same.

That said, I wonder about type variables, i.e. single upper-case letter identifiers. Are those more like variables or more like types?

deathaxe:
`variable.constant` is jet another Rust specific choice. I am aware of `variable.other.constant` only in other languages, but which is IMHO, also an oximoron. In most cases it is not possible to distinguish normal variables from constants, and using `constant` scope for them would be more appropriate anyway.

`storage.type` is what was proposed for user defined classes in Java, but caused the said critics for the same reason `support.class` is now discussed here - separation from built-in basic types. `storage.type.class` was originally intented for `class` keywords (or is already used for that by e.g. ASP). We use `keyword.declaration` for them, today.

I don't argue for `support.class` being the best choice nor am I a fan of `support` in general. This PR only intents to align Python with what seems to be used in most (maybe not all) other syntxes. 

Anything beyond is not scope of this PR, as any different scope choise may require a concuded RFC discussion and would affect many other syntaxes.

deathaxe:
FWIW, here's a list of all scopes containing `class`, except those starting with `meta.` and `punctuation.` from current master:

<details>

```
entity.name.class.asp
entity.name.class.c++
entity.name.class.cs
entity.name.class.d
entity.name.class.forward-decl.c++
entity.name.class.forward-decl.objc++
entity.name.class.haskell
entity.name.class.java
entity.name.class.js
entity.name.class.lisp
entity.name.class.matlab
entity.name.class.objc++
entity.name.class.php
entity.name.class.python
entity.name.class.scala
entity.name.type.class.actionscript
entity.name.type.class.groovy
entity.name.type.class.ocaml
entity.name.type.class-type.ocaml
entity.other.attribute-name.class.css
entity.other.attribute-name.class.html
keyword.declaration.class.actionscript
keyword.declaration.class.c++
keyword.declaration.class.cs
keyword.declaration.class.d
keyword.declaration.class.end.matlab
keyword.declaration.class.haskell
keyword.declaration.class.java
keyword.declaration.class.js
keyword.declaration.class.matlab
keyword.declaration.class.objc++
keyword.declaration.class.python
keyword.declaration.class.record.cs
keyword.declaration.class.ruby
keyword.declaration.class.scala
keyword.other.class-definition.ocaml
keyword.other.class-type-definition.ocaml
storage.type.class.groovy
storage.type.return-type.class.groovy
support.class.actionscript
support.class.asp
support.class.built-in.applescript
support.class.builtin.js
support.class.builtin.php
support.class.built-in.time.applescript
support.class.built-in.unit.applescript
support.class.cocoa
support.class.cocoa.leopard.objc
support.class.cocoa.objc
support.class.cocoa.objc++
support.class.collection.asp
support.class.dom.js
support.class.exception.python
support.class.finder.items.applescript
support.class.finder.type-definitions.applescript
support.class.finder.window-classes.applescript
support.class.itunes.applescript
support.class.java
support.class.js
support.class.latex
support.class.node.js
support.class.php
support.class.prelude.haskell
support.class.python
support.class.quartz.objc
support.class.ruby
support.class.scala
support.class.standardadditions.file.applescript
support.class.standardadditions.internet.applescript
support.class.standardadditions.miscellaneous.applescript
support.class.standardadditions.user-interaction.applescript
support.class.standard-suite.applescript
support.class.system-events.audio-file.applescript
support.class.system-events.disk-folder-file.applescript
support.class.system-events.folder-actions.applescript
support.class.system-events.movie-file.applescript
support.class.system-events.other.applescript
support.class.system-events.processes.applescript
support.class.system-events.property-list.applescript
support.class.system-events.quicktime-file.applescript
support.class.system-events.xml.applescript
support.class.textmate.applescript
support.class.text-suite.applescript
support.class.typing.python
support.function.classobj.php
variable.language.class.java
variable.other.readwrite.class.ruby
variable.parameter.class-inheritance.python
```
</details>

TL;TR: majority of them is of `entity.name.class` or `support.class`, regardless belonging to framework or being user defined. 

Finally that distinction is not possible with sane efforts for most languages. Anything could be a framework or number of classes shipped via stdlibs being just too huge to sanely provide all classes belonging to it. Not even arguing static name lists possibly targetting possibly user defined classes of same names.

The only language which seems to use `storage.type.class` is Groovy.

Hence, like it or not, the common sense for class variables applied to majority of languages is `support.class` and the scope of this PR is about aligning python's class names with that common sense.

michaelblyons:
FWIW, I'm okay with this change, but I've been holding off approval in case of additional comment.

deathaxe:
> That said, I wonder about type variables, i.e. single upper-case letter identifiers. Are those more like variables or more like types?

Type generics' parameters are scoped `variable.parameter.type` in Java and `variable.parameter.generic` in TypeScript. References are scoped `support.class` as it is impossible to distinguish them from normal types syntactically, except assuming all single-char types being type parameter references.

jrappen:
@deathaxe You got two reviews. Was there anything else to discuss?

deathaxe:
It still seems not all maintainers are comfortable with this change, entirely, and I don't want to offend anyone.

FWIW, I am not super happy with `support.class` either, but current strategy is to increase cross-syntax consistency by applying known concepts and best practices to those which would not end up in an entire rewrite. This should help to apply future concept changes to most/all affected languages more easily, with lower risk to miss any.

FichteFoll:
> I don't argue for `support.class` being the best choice nor am I a fan of `support` in general. This PR only intents to align Python with what seems to be used in most (maybe not all) other syntxes.
> 
> Anything beyond is not scope of this PR, as any different scope choise may require a concuded RFC discussion and would affect many other syntaxes.

I agree with this, hence you have my approval. I tried out the branch on a couple of Python files and the difference is admittedly less noticeable than I expected in those selected example files because I already used a lot of built-in types in there and this change aligns user-defined classes with this highlighting, resulting in more consistency.

We can warm up a general discussion about identifier scopes some time later when there is more time and motiviation on multiple sides. Sorry for not getting around to re-visit this earlier.

PR #4453, opened 2026-03-09
[Python] Fix parameter scope

deathaxe:
This commit fixes an unexpected side effect of renaming meta.generic-name by replacing stacked `variable.parameter variable.other` scopes.


michaelblyons:
Maybe add a `- variable variable`?

michaelblyons (approved):

PR #4456, opened 2026-03-11
[SQL] Refactor DECLARE statements

deathaxe:
This PR ...

1. fixes duplicate `after-type` contexts being set/pushed onto stack
2. refactors MySQL/PostgreSQL specific DECLARE statements to fix various existing highlighting issues and some introduced by (1.)

Notes:

- Postgre's DECLARE statement is more limited than MySQL. Hence various contexts are replacemed in Postgre, to disable/suppress MySQL specific syntax.

- `@` variables seem not supported in variable declarations, according to syntax specs.
- T-SQL's DECLARE statement is untouched. It declares cursors `meta.cursor-name`, but MySQL and PostgreSQL assign ordinary `variable` scope to keep context structure as simple a s possible.

### DIFF
--- a/SQL/MySQL.sublime-syntax
+++ b/SQL/MySQL.sublime-syntax
@@ -162,6 +162,7 @@ contexts:
         2: keyword.other.of.sql
       set:
         - type-declaration-index-by
+        - maybe-type-modifier
         - expect-type
     - include: else-pop
 
@@ -195,6 +196,7 @@ contexts:
     - match: (?=\S)
       set:
         - maybe-default-value
+        - maybe-type-modifier
         - expect-type
 
   condition-declaration-value:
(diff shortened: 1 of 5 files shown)


deathaxe:
T-SQL's DECLARE statement specs differ from what I see in existing tests to some degree, but it requires more investigation to judge.

However added some more refinements with regards to type modifiers and default values as those share some parts but still differ between column and variable declarations.

keith-hall:
> T-SQL's DECLARE statement specs differ from what I see in existing tests to some degree, but it requires more investigation to judge.

FWIW I'm pretty sure all the TSQL tests for `DECLARE` are valid syntax, but it's definitely possible something is misaligned with the specs

keith-hall (approved):

deathaxe:
I have a feeling, documented specs don't neccesarily tell the whole story of what's supported.

PR #4461, opened 2026-03-20
[Diff] Fix delta punctuation

deathaxe:
This PR excludes whitespace from punctuation scopes in Diff syntaxes.

before:

grafik

after:

grafik

### DIFF
--- a/Diff/Diff.sublime-syntax
+++ b/Diff/Diff.sublime-syntax
@@ -160,24 +160,24 @@ contexts:
 
   diff-deltas:
     # https://www.gnu.org/software/diffutils/manual/diffutils.html#Example-Normal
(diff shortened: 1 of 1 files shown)


keith-hall:
Arguably that space is a kind of punctuation because it separates the diff marker from the text, and some users might want to highlight it differently...

keith-hall (dismissed):
I'm not sure I like this change personally, but I can see how others would, and it is technically sound, so it gets my approval

michaelblyons:
Part of the trouble comes from my lack of distinction between Unified diff punctuation and the others where there's a space. Arguably, the space isn't really a part of the punctuation though.

Change seems fine to me.

PR #4472, opened 2026-03-27
[JavaScript] Align class instantiation scopes with Java/PHP

deathaxe:
This PR...

1. scopes class names `support.class` in `new ClassName()` expressions.
2. scopes whole `new ...` expression `meta.instantiation`.
3. converts related anonymous into named contexts.
4. renames contexts to better match their content.
5. directly match `.target` without additional lookahead patterns.

This aligns scopes with e.g. Java and PHP.

Note: `new` keyword is kept untouched at this point as finding a common scope for it seems to be another story.


deathaxe:
OK, this was the last one.

keith-hall (approved):

PR #4480, opened 2026-03-28
[Scala] Tweak varargs scopes

deathaxe:
This commit scopes anonymous and named varargs expressions using existing scopes of other languages.

see also: https://www.scala-lang.org/api/current/scala/quoted/Varargs$.html

Basically current Scala syntax definition supports two forms of varargs

  1. expr*
  2. _*

Considerations:

  1. Both look like regexp patterns followed by `*` quantifier.
  2. The underscore is used whenever a variable name is not needed.
  3. Python supports a comparable unpacking expression `*_`.

Combining those two points, results in ...

  1. scoping `*` as `keyword.operator.quantifier` which RegExp uses for it.
  2. scoping `_` in front of `*` as `variable.language.anonymous`.

### DIFF
--- a/Scala/Scala.sublime-syntax
+++ b/Scala/Scala.sublime-syntax
@@ -712,7 +712,7 @@ contexts:
           set:
             - match: '(\*)\s*(?=[\),=])'
               captures:
-                1: keyword.operator.varargs.scala
+                1: keyword.operator.quantifier.scala
             - match: '(?=\))'
               pop: 1
             - match: ','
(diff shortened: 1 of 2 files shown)


michaelblyons:
Python's `*_` would be `keyword.operator.unpacking.sequence` followed by the same `variable.language.anonymous`. In case anyone wanted to compare.

michaelblyons:
@djspiewak Any concerns?

djspiewak:
This seems a bit strange to me.

`_*` is not used when variable names are unneeded, it's used for argument *expansion*. So the exact syntax:

```scala
def foo(bar: String*): String = bar.mkString(", ")

foo("Me", "myself", "I")  // => "Me, myself, I"

val args = List("Me", "myself", "I"
foo(args: _*)  // => "Me, myself, I"
```

So `*` is used alone at the declaration site, while `_*` is used alone at the call site. `*` at the call site is not supported. This syntactic form is special: `_*` is effectively a *single* operator, similar to how `=>` is a single operator for defining a lambda. I don't think it makes sense to split the scoping between the two characters.

Issue #4468, opened 2026-03-26
[C#] Tuples, Qualified-Types, and Multiline-Strings


Nieb:
### What happened?

~Tuple aliases are very broken.~
Qualified types are inconsistently highlighted depending on context.
~Tuple assignments have issues with custom types.~
```cs
using F2 = (float x, float y);
using F3 = (float x, float y, float z);
using F4 = (float x, float y, float z, float w); // <- tuple aliases are all kinds of wonky

static class BLARG {
    Some.Thing SomeThing = new(); // <- perhaps this is intentional...

    void blarg(Some.Thing SomeThing) { // <- ...
        Some.Thing SomeThing = new();  // <- ...

        (int a, float b) = blarg();
        (vec2 a, vec3 b) = blarg();         // <- 
        (vec2 a, vec3 b, vec4 c) = blarg(); // <- last type isn't highlighted.
    }

    //  MultiLine strings support triple curly-brace for interpolation:
    private static string GLSL__BlargVertex = $$$"""
    #version 460 core
    #line {{{LINE_NUMBER()+1}}}

        layout(location = 0) in vec4 VertPos; //  Position XYZW

        void main() {
            gl_Position = VertPos;
        }

    """;
}
```

deathaxe:
FWIW, the amount of `$` chars in front of a multi-line string specifies the amount of braces needed to denote interpolation. A dedicated set of rules is needed for each in a static syntax definition and the question just is how many make sense. C# currently supports up to 2 `$` signs.

Nieb:
> FWIW, the amount of `$` chars in front of a multi-line string specifies the amount of braces needed to denote interpolation. A dedicated set of rules is needed for each in a static syntax definition and the question just is how many make sense. C# currently supports up to 2 `$` signs.

I see, could we do 3?  There is an, admittedly rare, chance that 2 curly-braces could show up inside a code string.  Also, odd numbers are aesthetically nice. :P

By the way, I have found several edge cases where the syntax goes wonky.  I've just fixed most of them in an edited syntax.  I could revert to a vanilla syntax and report them as I come across them again.
Would you prefer new issues be made for each?

michaelblyons:
> Would you prefer new issues be made for each?

Please feel free to report bugs! If you have fixes already, PRs are welcome.

Not all bugs can be patched since some languages are ambiguous, but it sounds like you've fixed some of your problems already.

deathaxe:
[...]

PR #4484, opened 2026-03-29
[C#] Fix various string literal highlighting issues

deathaxe:
Addresses #4468 "Multi-line" strings.

This PR is another iteration of refactoring and improving C# string literals.

It ...

1. contains changes to structure in order to organize related contexts in a more tree like structure, grouped by string literal types.
2. renames string contexts by their language specific types. "Long strings" become "Verbatim Strings".
3. removes invalid patterns for e.g. `@$"""..."""`. Those are just verbatim strings starting with escaped quotes.
4. fixes various escaping issues
5. adds various placeholder and interpolation related bailouts to handle incomplete expressions.
6. add support for `$$$"""..."""` and `$$$$"""..."""` raw strings
7. adjusts string scopes to align with Python

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -2031,6 +2031,7 @@ contexts:
     - match: '"'
       scope: punctuation.definition.string.end.cs
       pop: true
+    - include: string_prototype
     - include: string_escapes
     - include: string_placeholder_escapes
     - include: string_placeholders
@@ -2057,6 +2058,7 @@ contexts:
     - match: '"'
       scope: punctuation.definition.string.end.cs
       pop: true
+    - include: string_prototype
     - include: string_escapes
     - include: string_placeholder_escapes
     - match: \{
@@ -2109,6 +2111,7 @@ contexts:
     - match: '"'
       scope: punctuation.definition.string.end.cs
       pop: true
+    - include: string_prototype
     - include: string_placeholder_escapes
     - include: string_placeholders
     - include: extended_verbatim_string_placeholders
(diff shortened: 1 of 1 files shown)


deathaxe:
This PR seems to blow up PowerShell syntax due to context sanity limit being exceeded.

In Python we explicitly excluded SQL syntax highlighting in interpolated strings to discourage interpolated queries for security reasons.

Should we therefore remove SQL highlighting from C#'s verbatim format strings? It would resolve PowerShell.

michaelblyons:
PowerShell [explicitly disables SQL](https://github.com/SublimeText/PowerShell/blob/master/syntax/embed/C%23%20(for%20PowerShell).sublime-syntax) in nested C#. If this PR renames one of those contexts, it needs a change in PowerShell, too.

Or do https://github.com/sublimehq/Packages/pull/4416#issuecomment-3671069244

deathaxe:
Gotcha! Actually, all are renamed. Luckily those are cheap enough to have both versions in overridng syntax. I had to rename them as I wasn't able to wrap my head around the old naming vs. naming in specs.

michaelblyons:
Good catch, BTW, on 

> 3. removes invalid patterns for e.g. `@$"""..."""`. Those are just verbatim strings starting with escaped quotes.

michaelblyons:
(I wrote most of this a couple days ago but forgot to post.)

For casual lurkers, C# somewhat annoyingly calls its `"""` strings "Raw" and its older `@"` strings "Verbatim." Deathaxe has chosen here to set `"""` to `string.quoted.double.block` like Python uses and to create a new `string.quoted.double.verbatim` scope for `@"`.

I originally held off on approval for a while because I wanted to consider leaving `string.quoted.double.verbatim` as `.raw`. But if "raw" is actually C#'s name for another kind of string, that might be misleading. There are arguable trade-offs both ways.

deathaxe:
I am also not very confident about the best solution as `raw` is basically used for various `verbatim` terms in other languages as well. On the other hand these kind of detailed scopes is barely standardized and at some point it should be possible to denote language specific requirements. Suggestions wellcome.

PR #4493, opened 2026-04-01
[C#] Refactor delegate statements

deathaxe:
This PR implements `delegate ...` statements using multi-push and introduces commonly usable `maybe_type_parameters` and `type_arguments` contexts. Java uses them a lot.


jrappen:
tbh not a fan of having `pop: true` and `pop: 2` in the same syntax file, but that's out of scope for this PR.

Issue #2507, opened 2020-09-19
[RFC] Discussion on finding common scopes for import export related keywords.


Ultra-Instinct-05:
I have mainly opened this as an RFC issue to initiate some discussion on import export related keywords in various languages & to find (maybe ?) some common scopes.

This was mainly due to the fact that in JavaScript, `import` & `export` are both scoped as `keyword.control.import-export` (as well as any other keyword(s) like `except`, `as`, `from` in the said statements), which doesn't feel correct since one can explicitly mark an `import` keyword as `keyword.control.import` (|||ly for `export`) rather than having the current scope.

Some of the other references I have been able to find are :-
1. D -> `keyword.control.import`
2. Erlang -> `keyword.control.directive.import`
3. Go -> `keyword.other.import`
4. Haskel -> `keyword.control.import`
5. Java -> `keyword.control.import`
6. Python -> `keyword.control.import`

Thom1729:
I don't have a strong opinion about this, except that JavaScript's `keyword.control.import-export` is not great and I would welcome a new scope.

`keyword.control.import` seems to be the clear leader. I don't particularly care for `control`, because I think that's supposed to be used for flow control keywords like `if`, `switch`, and `for`, but I wouldn't be upset if that's where we landed.

Import syntax varies wildly across different languages. Off the top of my head, I might mentally divide it into three cases:

#### Dedicated syntax

In these cases, there's probably a clearly identifiable `import` keyword to scope.

JavaScript:

```js
import React from "react";
import "./style.css";
```

Python:

```py
import os.path
from os.path import join
```

#### Just another function

When imports are done with a function or with function-like syntax, any of `variable.function`, `support.function`, and/or `keyword` may be appropriate.

JavaScript:

```js
const fs = require("fs"); // Literally just a function call. Currently has a `support` scope.
import("./style.css"); // Technically special syntax, but works almost exactly like a function. Currently scoped `keyword.import`.
```

#### Some kind of directive

Languages that have multiple directives, including an import directive, should probably scope directives consistently, but we might add a `keyword` scope to the name of an import directive.

Erlang:

```erlang
-import(lists, [map/2,foldl/3,foldr/3]).
```

C:

```c
#include <stdio.h>
```

---

Exports might be more diverse. A lot of languages have an explicit `export` statement. Some implicitly export everything. Some are weird:

JavaScript (CommonJS):

```js
function foo () {}
module.exports = { foo };
```

Python:
```py
__all__ = ["foo"]

[...]

PR #4508, opened 2026-04-11
[JavaScript] Fix import keyword scope

deathaxe:
This commit applies a commonly used `keyword.control.import` scope.

This is the only usage of `keyword.import`, which is otherwise not known to be agreed part of scope naming guideline.


michaelblyons:
Lots of commentary at #2507.

This change seems reasonable until that RFC is resolved.

michaelblyons (approved):

jrappen (approved):

PR #4509, opened 2026-04-11
[JavaScript] Fix scope of generator operator

deathaxe:
This commit scopes `*` after `yield` and `function` `storage.modifier.generator` as current scope is unique and scope naming guideline doesn't contain an aggreed `generator` 2nd level scope.

The asterisk turns a function into a generator when following the `function` keyword.

The `yield*` expression delegates value generation to another generator like `yield from` in Python.


michaelblyons (approved):
I like it.

jrappen (approved):

PR #4522, opened 2026-04-15
[C#] Add regex to argument strings in known functions

### LINKED ISSUE #692: [C#] Add scoping of Regular Expressions

michaelblyons:
Possibly naive adaptation of something Keith and I had been tossing around.

Fixes #692

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1259,6 +1259,15 @@ contexts:
         - match: \(
           scope: meta.group.cs punctuation.section.group.begin.cs
           set: [constructor_arguments, arguments]
+        - match: ((?:(?:System(\.))?Text(\.))?RegularExpressions(\.))?(Regex)\s*(\()
+          captures:
+            1: meta.namespace.cs
+            2: punctuation.accessor.dot.cs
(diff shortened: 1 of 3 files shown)


keith-hall (dismissed):

C#/C#.sublime-syntax
```diff
@@ -2134,6 +2162,12 @@ contexts:
       pop: 1
     - include: immediately_pop
 
+  regex_arguments:
+    - match: (?=[@$]*")
+      push:
+        - include: regex_string
```
deathaxe: Are there reasons to not push directly?

```suggestion
      push: regex_string
```

deathaxe (approved):

jrappen (approved):

PR #4523, opened 2026-04-16
[C/C++] Add attributes

12Thanjo:
I added support for syntax highlighting of C/C++ attributes based on the [cppreference documentation for C](https://en.cppreference.com/w/c/language/attributes.html) and [C++](https://en.cppreference.com/w/cpp/language/attributes.html).

In addition, this fixes a bug where the syntax highlighting breaks when an attribute is in the same line as a function (it no longer considered it to be a function).

### DIFF
--- a/C++/syntax_test_c.c
+++ b/C++/syntax_test_c.c
@@ -1151,6 +1151,40 @@ __notdeclspec(deprecated("bla")) void func2(int) {}
 /* <- meta.function-call variable.function                    */
 /*                                    ^ entity.name.function  */
 
+
+
+inline [[nodiscard]] [[gnu::hot]] static void nodiscard_func();
+/* <- storage.modifier.c++ */
+/*     ^^^^^^^^^^^^^ meta.attribute.c++ */
+/*     ^^ punctuation.section.attribute.begin.c++ */
+/*       ^^^^^^^^^ storage.modifier.c++ */
+/*                ^^ punctuation.section.attribute.end.c++ */
+/*                   ^^^^^^^^^^^^ meta.attribute.c++ */
+/*                   ^^  punctuation.section.attribute.begin.c++ */
+/*                             ^^  punctuation.section.attribute.end.c++ */
+/*                                ^^^^^^ storage.modifier.c++ */
+/*                                            ^^^^^^^^^^^^^^ entity.name.function.c++ */
+
+
+[[deprecated("no longer used")]] void attribute_with_arg_func();
+/* <-- meta.attribute.c++ punctuation.section.attribute.begin.c++> */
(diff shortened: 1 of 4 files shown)


michaelblyons:
- It looks like the `.c` and `.c++` got mixed up in some of the test assertions.
- Most syntax files avoid deeply nested `push` or `set`
- If you're interested in the C/C++ family, perhaps consider taking #4147 across the finish line! (No pressure. This is not a trivial quest.)

C++/syntax_test_c.c
```diff
@@ -1151,6 +1151,40 @@ __notdeclspec(deprecated("bla")) void func2(int) {}
 /* <- meta.function-call variable.function                    */
 /*                                    ^ entity.name.function  */
 
+
+
+inline [[nodiscard]] [[gnu::hot]] static void nodiscard_func();
```
jrappen, line 1156: some newly added test assertions wrong, scopes correct

12Thanjo: Yup, copy-paste mistake. Fixed it.

12Thanjo:
> If you're interested in the C/C++ family, perhaps consider taking [[C Family] Full rewrite of C family syntaxes #4147](https://github.com/sublimehq/Packages/pull/4147) across the finish line! (No pressure. This is not a trivial quest.)

I'll definitely consider it, but not soon. I don't understand how it all works well enough yet to do what I'd consider a good enough job yet. However, I am working on a programming language, so I'll have practice writing the syntax highlighting for that - it's (in part) designed to be much easier to parse than C++, so it should be easier to do and good practice before I try full production-grade C++ syntax.

In addition, I have never touched Objective-C or Objective-C++, so I wouldn't be able to complete those rewrites.

michaelblyons:
> However, I am working on a programming language, so I'll have practice writing the syntax highlighting for that

Cool! If you have questions, feel free to visit the ST forum or Discord.

Edit: I should also mention that if you're the one designing the language, and you're using formal syntax to do so, there are projects to convert them to Sublime Syntax: [SBNF original in Rust](https://github.com/BenjaminSchaaf/sbnf), [SBNF re-implementation in Python](https://github.com/nuchi/sublime-from-cfg), and [Langspec](https://github.com/LordMartron94/Langspec).

But if you're writing a syntax the by-hand way, people are happy to help. (Maybe help with the formal things, too, just not me.)

rwols:
From what I understand, “Objective-X is a strict superset of X”, so all you have to do is understand the Objective syntax extensions to bolt it on top. (Contrast with the fact that C++ is not a strict superset of C)

michaelblyons (approved):

deathaxe (approved):

PR #4529, opened 2026-04-20
[PHP] Replace lone '&' in snippet scope with '&amp;'

### LINKED ISSUE #4528: [PHP] Possible Bug in Packages/PHP/Snippets/php (begin tag).sublime-snippet
### Expected behavior

In ST build 4200 on a Windows 10 x64 system, I was parsing snippet files with Python Package `xml.etree.ElementTree` and one of the shipped `.suplime-snippet` files causes an exception when parsing it as XML, namely this one:

https://github.com/sublimehq/Packages/blob/master/PHP/Snippets/php%20(begin%20tag).sublime-snippet

The reason is that the ampersand ('&') at character position 141 (inside the `<scope>` tag) is causing the `ET.fromstring(snippet_str)` to generate this exception:
```
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 4, column 41
```
I tested it by replacing that `&` with `&amp;` instead and that DOES parse without an exception.  And the text parsed out of it is correct:
```
'(embedding.php | text.html.php) & punctuation.section.embedded.begin.php'
```

While this snippet does work in Sublime Text (producing a `<?php | ?>` result), I mention this because it MAY be more correct to replace that ampersand with the `&amp;` entity.  I could find nothing specifically about this in either official or community docs.  However, one example was given in the official docs that uses the `&lt;` and `&gt;` entities and so I suspect that is the correct way to include an ampersand in the scope as well?

Anyway, I'm the type to "speak up" when I see something wrong.  I hope this helps.

Kind regards,
Vic

### Actual behavior

see above.

### Steps to reproduce

In a plugin:
```py
from xml.etree import ElementTree as ET
import sublime

path = 'Packages/PHP/Snippets/php (begin tag).sublime-snippet'
snippet_xml_str = sublime.load_resource(path)
snippet_tree = ET.fromstring(snippet_xml_str)
  File "__main__", line 1, in <module>
  File "./python3.8/xml/etree/ElementTree.py", line 1320, in XML
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 4, column 41
```
And proving the ampersand => `&amp;` resolves it:
```py
left = snippet_xml_str[:141]
right = snippet_xml_str[:142]
new_str = left + '&amp;' + right
snippet_tree = ET.fromstring(snippet_xml_str)
scope = snippet_tree.find('scope').text
print(scope)
# produces:  "(embedding.php | text.html.php) & punctuation.section.embedded.begin.php"
```

vwheeler63:
Fixes #4528

### DIFF
--- a/PHP/Snippets/php (begin tag).sublime-snippet
+++ b/PHP/Snippets/php (begin tag).sublime-snippet
@@ -1,6 +1,6 @@
 <snippet>
 	<content><![CDATA[${TM_PHP_OPEN_TAG:php} $0 ?>]]></content>
 	<tabTrigger>php</tabTrigger>
-	<scope>(embedding.php | text.html.php) & punctuation.section.embedded.begin.php</scope>
+	<scope>(embedding.php | text.html.php) &amp; punctuation.section.embedded.begin.php</scope>
 	<description>&lt;?php … ?&gt;</description>
 </snippet>


michaelblyons (approved):

jrappen (approved):

vwheeler63:
I've never had a PR turn-around so fast.  Thanks, guys!  :-)

PR #4530, opened 2026-04-20
[C#] Fix for loops

### LINKED ISSUE #4527: [C#] Multiple indices in for-loop
### What happened?

```cs
internal static partial class BLARG {
    public static void blarg(vec2[] Poly) {
        for (int iB = 0, iA = Poly.Length-1; iB < Poly.Length; iA = iB++) {
            //         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  nothing highlights after comma.
        }
    }
}
```

deathaxe:
Fixes #4527

This PR fixes issues with multiple variable declarations in for loops.

### DIFF
--- a/C#/C#.sublime-syntax
+++ b/C#/C#.sublime-syntax
@@ -1340,12 +1340,12 @@ contexts:
       captures:
         1: keyword.control.loop.for.cs
         2: meta.group.cs punctuation.section.group.begin.cs
-      set: [for_block, line_of_code_in, line_of_code_in, for_var_assignment, var_declaration]
+      set: [for_block, for_increment, line_of_code_in, variables_declaration, var_declaration]
     - match: \b(for)\s*(\()
       captures:
         1: keyword.control.loop.for.cs
         2: meta.group.cs punctuation.section.group.begin.cs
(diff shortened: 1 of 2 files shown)


C#/C#.sublime-syntax
```diff
@@ -1646,14 +1646,10 @@ contexts:
...
 
-  for_var_assignment:
-    - match: '='
-      scope: keyword.operator.assignment.cs
-      set:
-        - line_of_code_in
-        - maybe_pointer
-    - match: (?=;|\)|\})
-      pop: 1
+  for_increment:
```
keith-hall, line 1649: Technically there is no need for it to be an increment expression, but naming is hard and it is clear enough from looking at this context that it supports decrements etc as well

keith-hall (approved):

michaelblyons (approved):

C#/C#.sublime-syntax
deathaxe: Agree, but this is only a quick fix. I have already identified (and partly fixed) dozens of issues, all of which is part of a larger changeset. It needs some time to mature first in order to limit noise.

Issue #4525, opened 2026-04-16
[C++] Parentheses in macro arguments break syntax highlighting


GreenLightning:
### What happened?

In Unreal Engine programming it is very common to annotate code with macros like the following, however, the extra parentheses (`(Abstract)`) in the macro arguments break the syntax highlighting (see how the public keyword is highlighted differently than in `MyPlainStruct`).

<img width="188" height="162" alt="Image" src="https://github.com/user-attachments/assets/aaa2b7b9-3cb6-4eb1-a88e-4181d9735fec" />

<img width="170" height="173" alt="Image" src="https://github.com/user-attachments/assets/e6e3209e-2042-41e0-8748-081cabdf6fb2" />

```c++
struct MyPlainStruct
{
public:
};

USTRUCT(meta = (Abstract))
struct MyUStruct
{
public:
	// ....
};
```

deathaxe:
Maybe duplicate of #2861

PR #4534, opened 2026-04-23
[C++] Fix parentheses in macro arguments (#4525)

GreenLightning:
Fixes parentheses in macro arguments breaking the syntax highlighting (see #4525).


michaelblyons:
Maybe pull the pattern out into a variable? Then you don't need as much commenting to describe what you're doing, and anyone adding enhancements to the lookahead can add it to both places at once.

GreenLightning:
> Maybe pull the pattern out into a variable? Then you don't need as much commenting to describe what you're doing, and anyone adding enhancements to the lookahead can add it to both places at once.

I updated the commit. What do you think?

michaelblyons (approved):

deathaxe (approved):

Issue #4540, opened 2026-05-02
[python] many false positives in marking CamelCase names as classes


chrisjbillington:
### What happened?

Since upgrading to ST 1405, I'm seeing that Python classes in imports and some other places are highlighted as distinct from other symbols, in most colour schemes they're rendered in italics and a different colour to other names. This appears to be based on them being in CamelCase. This would be good if it could detect classes more specifically, but in practice I see a lot of false positives, as in the below screenshot. Most of these are just constants, not classes. Whilst the standard is to write constants in all caps, I notice a lot of my code doesn't follow this in practice, resulting in such false positives.

Ideally there would be a way to identify classes unambiguously, but if not I think I'd prefer them not to be distinguished from other names in order to avoid the false positives.

Edit: I didn't intend to add the "C: Syntax" label and don't see how to remove it. Apologies.

<img width="389" height="415" alt="Image" src="https://github.com/user-attachments/assets/4b2a61ee-5121-43f0-9a16-4e028601fa29" />

keith-hall:
> Ideally there would be a way to identify classes unambiguously, but if not I think I'd prefer them not to be distinguished from other names in order to avoid the false positives

Perhaps there are some improvements which can be made about false positives (probably after an arithmetic operator for example it is unlikely to be a bare class name), but in general I'm not convinced that just because your code doesn't follow conventions that the highlighting should be degraded for other code which does... Have you considered tweaking your colour scheme to make Python classes look the same as variables for example?

chrisjbillington:
I haven't and that's an option I suppose.

One false positive you might agree with special-casing is single-letter capital letters, these are following the convention for constants of being in all caps, but seem to also match on being CamelCase:

<img width="636" height="226" alt="Image" src="https://github.com/user-attachments/assets/4c755c1a-f2f1-421c-bc9b-e0bb0e3a200a" />

chrisjbillington:
Whilst I can change it for myself, I suppose I would like to push back just a little on the idea of variable capitalisation conventions playing quite as a strong a role as this in highlighting given the fields Python is used in.

[...]

PR #4541, opened 2026-05-02
[Python] Adjust type patterns

deathaxe:
Addresses #4540

This PR restricts which tokens are scoped `support.class`.

Scope tokens as type only, if they

1. follow `PascalCase` or `_PascalCase` without underscores in between.
2. consist of at least 2 chars
~~3. do not end with uppercase letter~~

Single char upper case tokens are now scoped `variable.other`, which is probably an appropriate scope for "type variables" in generics as well.

_It would be possible to keep them scoped `support.type` in type annotations with further adjustments to related contexts. Those currently use normal expression rules, which causes some ambiguities to resolve - especially with regards to `after-expression` context._

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -4134,7 +4134,7 @@ variables:
   identifier_start: '[[:alpha:]_]'
   identifier_char: '[[:alnum:]_]'
   identifier_class: '_?[[:upper:]](?:[[:upper:][:digit:]]*[[:lower:]]+[[:alnum:]]*)?\b'
-  identifier_constant: '(?:[\p{Lu}_][\p{Lu}_\d]*)?[\p{Lu}]{2,}[\p{Lu}_\d]*\b'  # require 2 consecutive upper-case letters
+  identifier_constant: '(?:[_[:upper:]][_[:upper:][:digit:]]*)?[[:upper:]]{2,}[_[:upper:][:digit:]]*\b'  # require 2 consecutive upper-case letters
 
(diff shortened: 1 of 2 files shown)


keith-hall (dismissed):

Python/Python.sublime-syntax
```diff
@@ -4133,7 +4133,7 @@ variables:
   identifier: '{{identifier_start}}{{identifier_char}}*'
   identifier_start: '[[:alpha:]_]'
   identifier_char: '[[:alnum:]_]'
-  identifier_class: '_?[[:upper:]](?:[[:upper:][:digit:]]*[[:lower:]]+[[:alnum:]]*)\b'
+  identifier_class: '_?(?:[[:upper:]]+[[:digit:][:lower:]]+)+\b'
```
FichteFoll: This change will also cause a class named `ClassA` to not be matched as a class anymore, which seems like a perfectly valid class name to me that comes up in the wild. I don't think we can generalize it like this to exclude `NaT` and `NaN` (via https://github.com/sublimehq/Packages/issues/4540#issuecomment-4363226291) specifically.

chrisjbillington: Agree, I gave it as an example of code in the wild, but it doesn't seem possible to exclude without like, specifically matching `NaT` exactly. FWIW, `numpy` removed `np.NaN` (replaced by`np.nan`) in numpy 2.0, presumably for code style reasons.

FichteFoll (approved):

jrappen (approved):

keith-hall (approved):

PR #4544, opened 2026-05-05
[Python] Fix commit autocompletion with enter after control flow keywords

### LINKED ISSUE #4543: [Python] Auto-complete not working on lines starting with `return`
### Expected behavior

When typing on a line starting with `return` the auto-complete should work.

### Actual behavior

When tying on a line starting with return auto-complete is non-functional. This was introduced by 41c1f3a931fdc4e7de979888e67d37851e2291c0. This would also break auto-complete on lines starting with `break|continue|pass`, when typing a comment after them.

### Steps to reproduce

Type the following:

```python
def foo():
    return f
```

The auto-complete should suggest `foo`, but pressing *enter* just enters a newline.

deathaxe:
Fixes #4543

This PR...

1. fixes `enter` not commiting completions after keywords like `return`.
2. adds `keypad_enter` binding to dedent next line for completeness.

It aligns `enter` behavior with that for all other languages, depending on value of `auto_complete_commit_on_tab` setting.

### DIFF
--- a/Python/Default.sublime-keymap
+++ b/Python/Default.sublime-keymap
@@ -152,4 +152,14 @@
             { "key": "preceding_text", "operator": "regex_match", "operand": "^\\s+(?:break|continue|pass|return\\b.*)$", "match_all": true }
         ],
     },
+    { "keys": ["keypad_enter"], "command": "run_macro_file", "args": {"file": "Packages/Python/Add Dedented Line.sublime-macro"}, "context":
+
+        [
+            { "key": "setting.auto_indent", "operand": true },
+            { "key": "selector", "operand": "source.python - source.python comment - source.python string" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "auto_complete_visible", "operand": false },
+            { "key": "preceding_text", "operator": "regex_match", "operand": "^\\s+(?:break|continue|pass|return\\b.*)$", "match_all": true }
+        ],
+    },
 ]
--- a/Python/Default.sublime-keymap
+++ b/Python/Default.sublime-keymap
@@ -148,7 +148,6 @@
             { "key": "setting.auto_indent", "operand": true },
             { "key": "selector", "operand": "source.python - source.python comment - source.python string" },
             { "key": "selection_empty", "match_all": true },
-            { "key": "auto_complete_visible", "operand": false },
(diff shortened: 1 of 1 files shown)


jrappen (dismissed):

michaelblyons (approved):

jrappen (approved):

Python/Default.sublime-keymap
```diff
@@ -151,4 +151,28 @@
...
+        [
+            { "key": "setting.auto_indent", "operand": true },
+            { "key": "selector", "operand": "source.python - source.python comment - source.python string" },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_match", "operand": "^\\s+(?:break|continue|pass|return\\b.*)$", "match_all": true }
+        ],
+    },
+
+    // Prefer committing completions if "auto_complete_commit_on_tab" is disabled
+    { "keys": ["enter"], "command": "commit_completion", "context":
```
FichteFoll, line 164: Wouldn't it make more sense to adjust the context of the previous enter and keypad_enter bindings to not trigger when `auto_complete_visible` and `setting.auto_complete_commit_on_tab = false`?

deathaxe: How would you express it exactly?

FichteFoll: Something like this (untested due to lack of time):
```js
    // Insert dedented line after certain keywords
    { "keys": ["enter"], "command": "run_macro_file", "args": {"file": "Packages/Python/Add Dedented Line.sublime-macro"}, "context":
        [
            { "key": "setting.auto_indent", "operand": true },
            { "key": "selector", "operand": "source.python - source.python comment - source.python string" },
            { "key": "preceding_text", "operator": "regex_match", "operand": "^\\s+(?:break|continue|pass|return\\b.*)$", "match_all": true },
            { "key": "auto_complete_visible", "operator": "not_equal" },
            { "key": "setting.auto_complete_commit_on_tab", "operator": "not_equal", "operand": false }
        ],
    },
```

FichteFoll: Admittedly, the semantics are little different because you cannot invert the 
```
            { "key": "auto_complete_visible" },
            { "key": "setting.auto_complete_commit_on_tab", "operand": false }
```
context cleanly since or-combinations are not possible. I suppose it's not a perfect implementation, then.

PR #4547, opened 2026-05-16
[C#] Rewrite Syntax

### LINKED ISSUE #4481: [C#] Type-Cast false-positive
### What happened?

Singular items in parenthesizes are detected as a type cast.

```cs
internal static partial class BLARG {
    public static partial class Thing;  //  OK!   
    public static partial struct Thing; //  OK!   After updating, these appear to work now.

    public static void blarg() {
        int i = RayCoord.x  +  RayCoord.y*(Vs.x)  +  RayCoord.z*(Vs.x*Vs.y); //  <- It's not a type-cast if there isn't a literal or identifier to the right of it.

        int a = b >>> 16;   //  ok?
        int b = a >>> 24;   //  <- Unsigned-right-shift     This appears to highlight now.  
                            //                              Though, I don't see >>> or \>\>\> in the syntax file.
                            //                              Perhaps it is highlighting as '>>' and '>'.

        uint c = 0xFFFF_0123u;         //  <- Support type-suffixes on hex & bin literals.
        uint d = 0b10101010_01010101u; //  <- 
    }
}
```

### LINKED ISSUE #4531: [C#] Question mark should be considered part of an operator type
### What happened?

This snippet is incorrectly highlighted starting at `Bar` itself, but if you remove the question mark `?` then it looks alright:

```cs
public class Foo
{
    operator Bar?(Foo reference)
    {
        if (reference == null) return null;
```

Image

Image

vs.

Image

Why are we trying `member_plain` after an `operator` keyword? Doesn't make sense.

### LINKED ISSUE #4545: [C#] C# 14 has evolved yet again
### What happened?

In C# 14, they added record classes to specify is not a struct type. Also, you can type alias tuples. Sublime is simply showing syntax errors but this compiles correctly.

```csharp
// record classes
// A mutable record class.
public record class C1
{
  public string? Name { get; set; }
}

// An immutable record class.
public record class C2(string? Name);
```
Then this:
```csharp
using Fruit = (string Name, int Number); // Aliasing a tuple type.
```
Here are screenshots of what I see.

Image

Image

deathaxe:
fixes #4481
fixes #4531
fixes #4545
... and many unreported

supersedes #4416

General
-------

This commit rewrites C# syntax definition, following current best practices
with regards to architecture, context names and scope naming.

It implements the top-down (global declarations, member declarations,
statements, expressions) structure, primarily inspired by Java, but with
simpler approaches to avoid heavy use of branch points.

Closely follows specification from

    https://learn.microsoft.com/en-us/dotnet/csharp/specification/

Highlights
----------

- adds full support for language features up to C#15
- improved support for C# specific regular expression highlighting
- highlights SQL queries using T-SQL by default
  via Packages/C#/Embeddings/SQL (for C#).sublime-syntax
- improves UX with incomplete statements and expressions
- 20-25% faster parsing


michaelblyons:
Can you `embed` SQL instead of prototyping it? I think you had a POC somewhere.

deathaxe:
Embedding SQL caused various more complex SQL statements to fail highlighting, which needs more investigation to find out why. We either need more complex SQL extensions or possibly even structural changes in SQL itself to fix it.

michaelblyons:
> Note: Required structural changes break backward compatibility with packages such as [Razor C# Syntax](https://packages.sublimetext.io/packages/Razor%20C%23%20Syntax) or [HTML (C#)](https://packages.sublimetext.io/packages/HTML%20(C%23)).

It would be nice if we can get this solidified before the next stable ST. HTML (C#) already has a couple brief release series for C# compatibility issues, and I hope not to make too many more.

Owing to circumstances, I no longer have access to 100k lines of interesting C# to test this against. I'll do my best with the test files, but I'm not likely to find problems not covered in tests.

michaelblyons:
Was `var` a thing that conventions moved to `keyword.declaration` instead of `storage.type`?

deathaxe:
For some languages, probably yes, if it is interpreted as general type agnostic declaration keyword.

I followed Java approach, however, which interprets `var` as variant datatype. Maybe we can clarify that by adding the 3rd-level scopes from Java as well.

jrappen:
@LoneBoco You might want to take a look at this as it might show up soon in one of the next `dev` builds.

michaelblyons:
> For some languages, probably yes, if it is interpreted as general type agnostic declaration keyword.
> 
> I followed Java approach, however, which interprets `var` as variant datatype. Maybe we can clarify that by adding the 3rd-level scopes from Java as well.

For C# it seems to be type-implied on first assignment, but thereafter concrete. https://www.ideone.com/hg25D5

```csharp
using System;

public class Test
{
	public static void Main()
	{
		var foo = 12;
		foo = "abc";
	}
}
```
```text
/home/UlFlcn/Project/Program.cs(8,9): error CS0029: Cannot implicitly convert type 'string' to 'int' [/home/UlFlcn/Project/Project.csproj]
    0 Warning(s)
    1 Error(s)
```

michaelblyons:
A pity you couldn't split that goto reference in Stable. That was a cute idea. Soon™️ .

deathaxe:
Yeah, but excluding extra `meta` was required anyway to avoid `variable.other` in r-values being indexed. Anything else might need more complex meta scoping, to distinguish l-values from r-values in declarations.

C#/Embeddings/Regex (for C#).sublime-syntax
```diff
@@ -70,7 +74,14 @@ contexts:
         1: punctuation.definition.capture-group-name.begin.regexp
         2: variable.other.capture-group.regexp
         3: punctuation.definition.capture-group-name.end.regexp
-    # substituations
+
+  substituations:
```
michaelblyons:
```suggestion
  substitutions:
```

C#/Embeddings/Regex (for C#).sublime-syntax
```diff
@@ -20,6 +20,11 @@ variables:
   capture_group_name: (?:[a-zA-Z_]\w*)
 
 contexts:
+
+  base-common:
+    - meta_prepend: true
+    - include: substituations
```
michaelblyons:
```suggestion
    - include: substitutions
```

michaelblyons (changes requested):

jrappen:
Could you set this to draft while you're busy? Then I don't need to check after every push if you're ready for review.

deathaxe:
Beyond some bugfixes found when skimming various Unity or other larger code bases, C#15 features, SQL and RegExp integration without `with_prototype` caused some unexpected amount of work to ensure syntax highlighting works as expected.

There are no plans for further changes beyond reported bugs at this point.

jrappen:
Ok, I'll try to make some time this week.

LoneBoco:
https://github.com/sublimehq/Packages/pull/4547/commits/efef118543ad50f3e8e9e4522a77d28b73584b4b breaks ST4 4205.  It causes the error message: "missing variable escape"

It seems to be an issue in `inside-verbatim-regex-string`, with the `- include: scope:source.regexp.embedded.cs.verbatim-string` line.  If I comment that line out, it works fine.  I'm not entirely sure why this is happening, though.

michaelblyons:
> [efef118](https://github.com/sublimehq/Packages/commit/efef118543ad50f3e8e9e4522a77d28b73584b4b) breaks ST4 4205. It causes the error message: "missing variable escape"

You need to use the Regular Expressions from this PR as well.

michaelblyons:
Thoughts on the syntax tests:

- Looks good!
- There are lots of new C# concepts I haven't used. I remember being all excited that C#6 had null-coalescing operators.
- Speaking of those operators, I find it a little awkward for `?.` to be `punctuation` but `?[` to be `keyword` on the `?` and `punctuation` for `[`. It's an okay choice. Just a bit weird.
- Someone might want to check that writing format spec strings doesn't flash `illegal` back and forth.
- I think the scope for a context `using` was different before. I should read up again on the discussions about those.

Syntax file:

- Good comments.
- Why is the VARIABLE DECLARATIONS main context called `local-declarations`?
- The `\.` match in `expression-common` could be `include: accessors`.
- `stray-punctuations` is grammatically incorrect, but adding the `s` helps show that it's the *plural* of `punctuation` rather than the singular (which is also `punctuation`).

C#/C#.sublime-syntax
```diff
@@ -395,2112 +415,2274 @@ contexts:
...
+  inside-nameof-group:
+    - meta_scope: meta.group.cs
+    - match: \)
+      scope: punctuation.section.group.end.cs
+      pop: 1
+    - include: accessors
+    - include: namespace-variables
+    - include: instance-variables
+    - include: base-types
+    # We don't if these are types of variables, so we guess that
```
jrappen: maybe:

```yaml
# We can't be sure that these are types of variables. We
# assume uppercase names to be types, and all remaining
# ones to be variables.
```

C#/C#.sublime-syntax
```diff
@@ -395,2112 +415,2274 @@ contexts:
...
+    # uppercase names are types, and others are variables
+    - match: (?=\p{Ll}){{name}}(?=\s*\))
```
jrappen: Any time I see one of those I ponder whether it would make sense to reference the unicode properties section of the oniguruma docs or not, at least in the header section of the syntax file. Really depends who's editing the file, though.

C#/C#.sublime-syntax
```diff
@@ -2541,154 +2732,179 @@ contexts:
...
-      scope: constant.language.boolean.false.cs
-    - match: \btrue\b
-      scope: constant.language.boolean.true.cs
-    - match: \bnull\b
-      scope: constant.language.null.cs
 
 ###[ REGULAR EXPRESSIONS ]#####################################################
 
-  regex_string:
+  regex-string:
```
jrappen, line 2738: unused capturing groups

C#/C#.sublime-syntax
```diff
@@ -2541,154 +2732,179 @@ contexts:
...
+    - match: '"(?!")'
+      scope: punctuation.definition.string.end.cs
       pop: 1
-    - include: else_pop
+    - include: scope:source.regexp.embedded.cs.verbatim-string
+      apply_prototype: true
 
 ###[ STRINGS ]#################################################################
 
   strings:
```
jrappen, line 2799: unused capturing groups

deathaxe:
> ... main context called local-declarations ...

1. It contains `var` keyword which is valid only within local variable declarations. 
2. It may start a local (nested) method declaration.

Thus `variable-declarations` doesn't seem to be suitable and just `declarations` would be too generic as it might indicate to also include class/struct/... declarations.

It is included in top-level contexts for fallback reasons only. But well, the naming is not that important. 

> I think the scope for a context using was different before. 

It was `keyword.control.` before, maybe compared to python's `with...` etc.

I am uncertain about that, because it basically does the same regardless being used in global import part or in normal control flow.

C# used to even scope global `using` different depending on whether it imports namespaces or declares new types or variables, but this is IMHO not good choise when those can appear right next to each other, especially as we can't determine whether an import imports a namespace, type or function. Hence back then I decided to scope them equally across all use cases.

The only difference in control flow is them automatically clearing resources when leaving the introduced code block.

> Speaking of those operators, I find it a little awkward for ?. 

Yeah, this part is untouched, yet.

deathaxe:
> The \. match in expression-common could be include: accessors

Maybe low severity, but accessors may be extended by 3rd-parties and order of operator patterns in expression-common is quite sensitive. Hence I found the explicit pattern there a more obvious choice. Size of final syntax file is unaffected. And I don't like arbitrary includes in the middle of normal patterns :D

C#/C#.sublime-syntax
deathaxe: TBH, I prefer more verbose `[[:lower:]]` etc. overall. This is just another remainder, which was not touched. IIRC, posix char classes were not supported back when C# was written.

C#/C#.sublime-syntax
deathaxe: All of them are required to pop context on matching amount of quotation marks via `\1` backref.

C#/C#.sublime-syntax
deathaxe: Same here. Needs to stay as it is.

michaelblyons:
> It was `keyword.control.` before, maybe compared to python's `with...` etc.

Right. IronPython and recently `pythonnet` implement this kind of `using` as `with`.

> I am uncertain about that, because it basically does the same regardless being used in global import part or in normal control flow.

Do you mean the one in global that is more like a `from`/`import` to distinguish between namespaces?


> C# used to even scope global `using` different depending on whether it imports namespaces or declares new types or variables, but this is IMHO not good choise when those can appear right next to each other, especially as we can't determine whether an import imports a namespace, type or function. Hence back then I decided to scope them equally across all use cases.

For global scope, I think you made the right choice. Unless I misunderstood the situation described.


> The only difference in control flow is them automatically clearing resources when leaving the introduced code block.

It wasn't `keyword.control.flow` was it? I thought it was `keyword.control.context`.

I would still approve as-is, but I disagree on this thing being an import.

deathaxe:
Recent commits introduce branching to avoid illegal highlighting of incomplete string placeholders or interpolation.

deathaxe:
> It wasn't keyword.control.flow was it? I thought it was keyword.control.context.

Maybe the latter, yes.

Comment about python was just about same meaning.

I am aware of discussions proposing `keyword.control.context`. We may also return to that, for `using (foo = ...) { }` statements. I just priorized consistent highlighting of `using` keyword and found `keyword.declaration` not worse with regards to what the statement does.

> ... I disagree on this thing being an import.

That's the point. Some global `using` keywords were scoped `keyword.control.import` before, which may have been a proper choice back when it was used to declare used namespaces only.

As it may also just declare new types or aliases, these days, `keyword.declaration` appears to be a better fit.

> Do you mean the one in global that is more like a from/import to distinguish between namespaces?

The global one can do anything. Importing namespaces or declaring types and variables.

```
using Any.Namespace;

using static Any.Namespace.MaybeType;

using alias = Any.Namespace;

using TupleType = (int, string, foo);

using DisposableType myVar;
```

The last one is pretty similar to what `using ()` does, just for the whole module's scope.

michaelblyons:
> The global one can do anything. Importing namespaces or declaring types and variables.
> 
> ```cs
> using Any.Namespace;
> using static Any.Namespace.MaybeType;
> using alias = Any.Namespace;
> using TupleType = (int, string, foo);
> using DisposableType myVar;
> ```
> 
> The last one is pretty similar to what `using ()` does, just for the whole module's scope.

You make some compelling arguments.

> C# uses T-SQL

I like the SQL change. That probably (🤞) makes it better out of the box for C# users. I used to have my own default SQL flavor that took T-SQL and added some NHibernate stuff and some (private) conventions. Your changes make it slightly more awkward to set a single shared SQL default, but I think it is worth it.

deathaxe:
> ... more awkward to set a single shared SQL default.

C#/Embeddings/SQL (for C#).sublime-syntax is designed to be the same sort of common override like SQL/SQL.sublime-syntax, so plugins could provide a "Set SQL Dialect for C#" command. It requires just an additional override of SQL (for C#) to point it to SQL/SQL to continue with a common SQL syntax.

michaelblyons (approved):

jrappen (approved):

jrappen:
Thanks a lot for working on this.

Haven't checked the opening comment against commit messages from later commits, but you can still do that when merging.

PR #4549, opened 2026-05-18
[Go] Extend struct tag key matching

ilkiri23:
Some Go packages use hyphens in struct tag keys. For example, [cleanenv](https://github.com/ilyakaznacheev/cleanenv#read-configuration) uses \`env-default\`:
```go
type ConfigDatabase struct {
  Port string `yaml:"port" env:"PORT" env-default:"5432"`
  ...
}
```


michaelblyons (approved):
Looks good to me. Thanks for making a PR and for linking the example.

deathaxe (approved):

PR #4552, opened 2026-05-22
[Python] Add support for explicit lazy imports

deathaxe:
This PR implements https://peps.python.org/pep-0810/, which is supported as of python 3.15

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -427,13 +427,17 @@ contexts:
 ###[ IMPORT STATEMENTS ]######################################################
 
   import-statements:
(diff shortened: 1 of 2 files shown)


michaelblyons (approved):
Works for me.

Technically, `lazy` isn't valid in all the same places as an `import` was, but that's fine to leave to linters. (And if you had a nested `import`, *the point of `lazy` is that you can move it to the top!*)

deathaxe:
As python syntax doesn't (can't) track blocks or nested contexts, context aware treatment is not  possible.

jrappen (approved):

PR #4556, opened 2026-05-29
[Go] Update meth snippet

ilkiri23:
This change replaces leading spaces with tabs and adds additional field markers


jrappen:
what about placeholders for 4 and 5?

ilkiri23:
As far as I understand, if we add placeholders `${4:arguments}` and `${5:returnType}`, the user will need to press Tab+Space instead of just Tab when the method does not take arguments or return nothing

jrappen:
Just an informative question, as I don't use go.

jrappen (approved):

michaelblyons (approved):

PR #4566, opened 2026-06-14
[JavaScript] Fix nested template strings

### LINKED ISSUE #4220: [JavaScript] Template string inside tag function breaks syntax highlighting

deathaxe:
Resolves #4220

This PR splits changes of #4318 into two initial commits to better track required changes to achieve proper handling of nested template strings.

Step 1: 

Rename existing syntaxes to `<name> (Plain)` and scopes to `source.<...>.plain`. Adjust includes to keep everything working.

Step 2: 

1. (Re-)Create JavaScript, TypeScript, JSX, TSX syntaxes by extending the plain ones from Step 1.
2. Move literal template string related contexts to those new syntaxes.
3. Refactor and directly include CSS, HTML, JS, ... syntaxes in order to properly handle nested template strings.

Notes: 

1. Nested template strings don't highlight any syntax to keep syntax cache size in sane limits.
2. Any 3rd-party package continues to only see JavaScript, TypeScript, JSX and TSX syntaxes with unchanged scopes. Hence this change has no side effects for those.
3. 3rd-party packages could make use of `.plain` variants to prevent syntax highlighted template strings cause infinite include loops.


jrappen (dismissed):

jrappen:
I wasn't sure if adding explicit child scope match tests would make sense (requires 4201 or later), it'd work on the `dev` channel test binary only.

deathaxe:
A first-class example of what this PR resolves can be observed at [thecrawl/eleventy.install-chart.mjs#L427-L465](https://github.com/packagecontrol/thecrawl/blob/567991431ea59f05cbad06ede04f53c90f6518ba/eleventy.install-chart.mjs#L427-L465).

Finally we should decide whether syntax highlighting tagged templates is worth the extra complexity (layer of syntax files) and merge this PR or remove syntax highlighting from the package again.

Currently it is just too likely to badly break syntax highlighting in real world code, which is not ideal for keeping as is for an upcoming stable release.

deathaxe:
Rebased on current master as this is the easiest way to move forward without conflicts.

michaelblyons:
> A first-class example of what this PR resolves can be observed at [thecrawl/eleventy.install-chart.mjs#L427-L465](https://github.com/packagecontrol/thecrawl/blob/567991431ea59f05cbad06ede04f53c90f6518ba/eleventy.install-chart.mjs#L427-L465).
>
> Currently it is just too likely to badly break syntax highlighting

You make a compelling point.

> Finally we should decide whether syntax highlighting tagged templates is worth the extra complexity (layer of syntax files) and merge this PR or remove syntax highlighting from the package again.

I don't have issues with the work you've done here, but I hope more people can check it out before merging.

deathaxe:
> ...  I hope more people can check

Given the age of the whole topic including the former PC I am just concerned it not happening anytime soon, if it ever would.

---

The step backward - dropping all nested syntax highlighting - would basically be to drop latest commit from this PR and remove ` (Plain)` from filenames and `.plain` from scope names.

michaelblyons (approved):

jrappen (approved):

PR #4568, opened 2026-06-14
[D] Fix `auto` attribute in if and while conditions

### LINKED ISSUE #4567: [D] (new) while condition assignment
### What happened?

```d
while (auto n = expression)
{
}
```
incorrect:

Image

correct:

Image

https://dlang.org/changelog/2.097.0.html#while-condition-assignment

deathaxe:
Fixes #4567

This PR removes a redundant `auto` pattern from `basic-type` context and adds a special `condition-attribute` context to consume it at the beginning of `if` and `while` conditions.

Note: Adding `condition-attribute` context is a conservative decision to avoid undesired side-effects. It may also be valid to prepend `auto` pattern to `first-value`.

### DIFF
--- a/D/D.sublime-syntax
+++ b/D/D.sublime-syntax
@@ -2355,7 +2355,7 @@ contexts:
   condition-attribute:
     - match: '\bauto\b'
       scope: storage.modifier.d
-      pop: 1
+      pop: true
     - match: '(?=\S)'
       pop: true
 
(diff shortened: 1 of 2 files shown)


jrappen:
not a fan of the mix of `pop: 1` and `pop: true`

michaelblyons (approved):

jrappen (approved):

PR #4573, opened 2026-06-19
[Java] Add support for JEP401 value class objects

deathaxe:
This commit adds support for `value` class modifier, supported as of JDK28.

see: https://openjdk.org/jeps/401


jrappen (dismissed):

deathaxe:
Resolved conflict with master. FWIW, JEP401 has been merged and will be part of JDK28

jrappen (approved):

PR #4577, opened 2026-06-21
[ci] use `actions/checkout@v7`

jrappen:


michaelblyons (approved):
They finally fixed the bug where you could checkout a commit from anywhere in the fork network!? Big news.

deathaxe (approved):

jrappen:
I was just surprised to see them jump to v7 so soon after v6.

PR #4581, opened 2026-07-01
[TOML] Add 'mise.lock' to file_extensions

FichteFoll:
https://mise.en.dev/dev-tools/mise-lock.html

### DIFF
--- a/TOML/TOML.sublime-syntax
+++ b/TOML/TOML.sublime-syntax
@@ -15,6 +15,7 @@ file_extensions:
   - pdm.lock
   - poetry.lock
   - uv.lock
+  - mise.lock
 
 contexts:
   main:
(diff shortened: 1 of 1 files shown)


michaelblyons:
Should some of these be hidden file extensions?

Otherwise, sounds good to me.

deathaxe (dismissed):
Hiding lock files is probably a good idea but can be a sepearate PR.

michaelblyons (approved):

michaelblyons (changes requested):
Alphabetical order?

michaelblyons (approved):
Yay, thanks!

FichteFoll:
Just for the record, main `file_extensions` list is obviously not sorted because the first entry shall be the most likely one since it will be appended to the file name by default if no extension is provided (on some platforms).

jrappen (approved):

PR #4586, opened 2026-07-10
[Python] Fix raw string replacements

### LINKED ISSUE #4585: [Python] False-positive invalid highlighting of escaped character in template placeholder
### What happened?

The `\$` "escape" sequence that is not actually an escape sequence because it occurs within a raw string is highlighted as `invalid.deprecated.character.escape`:

```py
R"{\$}"
```

deathaxe:
Fixes #4585

This PR...

1. introduces dedicated `raw-string-replacement` contexts without escape sequences
2. reorganizes string replacement related contexts by moving them directly below their related/owning string type
3. renames former `string-replacement` to `u-string-replacement` to express its unique use in normal unicode strings.

### DIFF
--- a/Python/Python.sublime-syntax
+++ b/Python/Python.sublime-syntax
@@ -2746,6 +2746,47 @@ contexts:
     - include: triple-double-quoted-raw-string-replacements
     - include: escaped-raw-quotes
 
+  triple-double-quoted-raw-string-replacements:
+    - include: escaped-string-braces
+    - match: (?=\{)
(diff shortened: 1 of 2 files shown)


michaelblyons:
I think it would be funny to collect stats per language of "line and context count relating to strings" as a percentage of the total.

michaelblyons (approved):

jrappen (approved):

jrappen:
Even worse are the languages with significant whitespace and need for branching, with capturing across contexts. And then those that are that on top of being markup languages.

PR #4589, opened 2026-07-22
[Common] Auto-pair triple-quoted strings

### LINKED ISSUE #4584: [Python] Auto-pair triple quotes
### Expected behavior

It would be nice to auto-pair the triple quotes in python, like zed does

Something like this seems to work for `"`
(we should do the same for `'`)

Edit: there’s a better behavior than the poc bellow, see discussion
```json
  {
    "keys": [
      "\""
    ],
    "command": "insert_snippet",
    "args": {
      "contents": "\"$0\"\"\""
    },
    "context": [
      {
        "key": "setting.auto_match_enabled"
      },
      {
        "key": "selector",
        "operand": "source.python"
      },
      {
        "key": "selection_empty",
        "match_all": true
      },
      {
        "key": "preceding_text",
        "operator": "regex_contains",
        "operand": "\"\"$",
        "match_all": true
      },
      {
        "key": "following_text",
        "operator": "not_regex_contains",
        "operand": "^\"",
        "match_all": true
      }
    ]
  },
  {
    "keys": [
      "\""
    ],
    "command": "move",
    "args": {
      "by": "characters",
      "forward": true
    },
    "context": [
      {
        "key": "setting.auto_match_enabled"
      },
      {
        "key": "selector",
        "operand": "source.python"
      },
      {
        "key": "selection_empty",
        "match_all": true
      },
      {
        "key": "following_text",
        "operator": "regex_contains",
        "operand": "^\"",
        "match_all": true
      },
      {
        "key": "selector",
        "operator": "not_equal",
        "operand": "punctuation.definition.string.begin",
        "match_all": true
      },
      {
        "key": "eol_selector",
        "operator": "not_equal",
        "operand": "string.quoted.double - punctuation.definition.string.end",
        "match_all": true
      }
    ]
  },
```

### Actual behavior

```
"| + " -> ""
""| + " -> """
""" + " -> """|"""
"""|""" + """ -> """"""|
[aaa] + " -> "[aaa]"
"[aaa]" + "" -> """[aaa]"""
```

### Steps to reproduce

-

deathaxe:
Resolves #4584

This PR adds key bindings to auto-pair triple quoted strings in Groovy, Java, Python and TOML.


jrappen (dismissed):

michaelblyons:
I haven't tried the macros yet, but I'm curious about a minor thing: Can we use trailing commas in Sublime JSONC so that adding and removing lines doesn't touch the lines around it?

jrappen:
while we're at it, (based on what github was showing me online) maybe add empty lines at EOF?

deathaxe:
With regards to trailing commas, the style follows style of files in Default package, vast majority of which avoid them.

With regards to trailing newline at eof, it is of no syntactical use.

michaelblyons:
I'm trying the Python one, and it's not doing what I expect for expansion.

```python
foo = |
foo = '|'
foo = ''|
foo = '''|
# Expected '''|''', right?
```

Same for `"`

For deletions, I think it looks good:

```py
foo = '''|'''
foo = '|'
foo = |
```

michaelblyons (approved):

PR #4601, opened 2026-08-14
[Markdown] Disable indentation rules

### LINKED ISSUE #4597: [Markdown] Indentation level increased after sequence `<text>`
### Expected behavior

Indentation remains on the same level for all three lines

### Actual behavior

After the second line, indentation is increased by one level.

### Steps to reproduce

1. Create a new view and set the syntax to "Markdown".
2. Write or paste `` some `text` `` and press enter.
3. Write or paste `` with `<tags>` `` and press enter.

deathaxe:
Fixes #4597 

This PR...

1. renames existing indentation rules file as it targets raw code blocks, only.
2. disables indentation rules in normal content to prevent rules from html being applied.

### DIFF
--- a/Markdown/Indentation Rules.tmPreferences
+++ b/Markdown/Indentation Rules.tmPreferences
@@ -1,14 +1,15 @@
-<?xml version="1.0" encoding="UTF-8"?>
 <plist version="1.0">
 <dict>
 	<key>scope</key>
-	<string>text.html.markdown markup.raw</string>
+	<string>text.html.markdown</string>
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string>^(.*\*/)?\s*\}[;\s]*$</string>
+		<string></string>
 		<key>increaseIndentPattern</key>
-		<string>^.*(\{[^}"']*|\([^)"']*)$</string>
+		<string></string>
+		<key>bracketIndentNextLinePattern</key>
+		<string></string>
 	</dict>
 </dict>
 </plist>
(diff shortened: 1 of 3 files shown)


Markdown/Indentation Rules.tmPreferences
```diff
@@ -1,14 +1,15 @@
...
 <plist version="1.0">
 <dict>
 	<key>scope</key>
-	<string>text.html.markdown markup.raw</string>
+	<string>text.html.markdown</string>
 	<key>settings</key>
 	<dict>
 		<key>decreaseIndentPattern</key>
-		<string>^(.*\*/)?\s*\}[;\s]*$</string>
+		<string></string>
```
keith-hall, line 13: do we have any indentation test files? I don't remember offhand whether an empty regex would have the behavior of disabling auto indentation or always enabling it because an empty regex matches everything...

deathaxe: No, we don't have any, yet. Just quickly tested it working, manually.

Anyone is welcome to add some.

Markdown/Indentation Rules.tmPreferences
```diff
@@ -1,14 +1,15 @@
...
 		<key>increaseIndentPattern</key>
-		<string>^.*(\{[^}"']*|\([^)"']*)$</string>
+		<string></string>
+		<key>bracketIndentNextLinePattern</key>
+		<string></string>
```
FichteFoll, line 17: Perhaps we can add a comment here explaining why we are setting everything to an empty string?

Markdown/Indentation Rules.tmPreferences
deathaxe: FWIW, indentation tests in markdown are more or less useless as indentation system can't teach a `reindent` to not change most of the blocks/content in unexpected (wrong) ways.

FichteFoll (approved):

PR #4603, opened 2026-08-19
[Markdown] Improve table performance with fast reject

guille:
- [X] My commit messages start with the package name in square brackets, e.g. `[XML] `.
 - [ ] I have included new or enhanced [syntax tests](https://www.sublimetext.com/docs/syntax.html#testing) to cover the changes.
 
 No tests added because it's just performance: no regressions introduced.
 
 Found this through syntect, which I use in one of my projects. Sublime's optimized engine does not have the issue, or it's not as bad, and neither does Oniguruma, but `fancy-regex` does.
 
 `{{table_first_row}}` expands to ~25 KB of regex whose first alternative is `(?:{{table_cell}}?\|){2}`, and `{{table_cell}}` is itself a `(?:…)+` over alternations. That nesting is exponential on failure, and it always fails on ordinary paragraph text.
 
 Both alternatives of `{{table_first_row}}` require a literal pipe on the line, and without it the nested quantifiers in `{{table_cell}}` backtrack over every paragraph line. The lookahead acts as a faster reject for the same condition.
 
 I suppose the main risk is the possibility of drift, but hopefully the existing test coverage is enough to prevent that.
 
 The fast reject, on top of fixing the pathological case in `fancy-regex`, also speeds up Sublime's syntax in prose-heavy Markdown docs, when there are no tables (around 10% on my measurements, YMMV).

(Side note: #4600 is also great for performance, amazing stuff!)

### DIFF
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -2571,7 +2571,7 @@ contexts:
 ###[ LEAF BLOCKS: TABLES ]####################################################
 
   tables:
-    - match: ^(?=[^|\n]*\|)(?={{table_first_row}})
+    - match: ^(?=[^|]*\|)(?={{table_first_row}})
       push: table-header
 
   table-header:
--- a/Markdown/Markdown.sublime-syntax
+++ b/Markdown/Markdown.sublime-syntax
@@ -2571,7 +2571,7 @@ contexts:
 ###[ LEAF BLOCKS: TABLES ]####################################################
 
   tables:
-    - match: ^(?={{table_first_row}})
+    - match: ^(?=[^|\n]*\|)(?={{table_first_row}})
       push: table-header
 
   table-header:


Markdown/Markdown.sublime-syntax
```diff
@@ -2571,7 +2571,7 @@ contexts:
 ###[ LEAF BLOCKS: TABLES ]####################################################
 
   tables:
-    - match: ^(?={{table_first_row}})
+    - match: ^(?=[^|\n]*\|)(?={{table_first_row}})
```
keith-hall: Does the `\n` bring any benefit here? Only one line is ever matched at a time
```suggestion
    - match: ^(?=[^|]*\|)(?={{table_first_row}})
```

keith-hall (dismissed):
Nice, thanks. Friendly hint: If you are interested in fancy-regex performance, you may want to bump fancy-regex to 0.19 and enable/call `seek(true)` on the RegexBuilder 😉 likely this pattern wouldn't have been slow

guille:
PS: It does seem like fancy-regex 0.19 doesn't hit the bad case, the performance is much much better. Looking forward to getting those improvements once it lands on syntect then! 😉

keith-hall (approved):

Markdown/Markdown.sublime-syntax
```diff
@@ -2571,7 +2571,7 @@ contexts:
 ###[ LEAF BLOCKS: TABLES ]####################################################
 
   tables:
-    - match: ^(?={{table_first_row}})
+    - match: ^(?=[^|]*\|)(?={{table_first_row}})
```
deathaxe, line 2574: The alternation should be added to the pattern of `table_first_row`.

That being said, patterns in `table_first_row` already add various heuristics to detect lines with pipes while trying to avoid various false positives.

This change appears to make at least some of those existing ones obsolete or may have other maybe unwanted side effects?

There are likely reasons why at least 2 unescaped pipes are expected in a line.

deathaxe (changes requested):
As this change basically seems to make the first part of `table_first_row` pattern obsolete or never match, further considerations are needed.

Markdown/Markdown.sublime-syntax
deathaxe: I just realize it is an AND lookahead. So nevermind the comments above.

deathaxe (approved):

PR #4604, opened 2026-08-24
[C/C++/C#/Java] Add missing keywords

mikomikotaishi:
This PR adds missing keywords to the C, C++, C# and Java languages.

- C is missing keywords like `_Generic`, `_Countof`, the Annex H floating types, and macros (`offsetof`, `unreachable`, `countof`; akin to `assert`). This PR adds these.
- Add `contract_assert`, `pre`, and `post` to C++.
- Add `scoped`, `field`, match `allows ref struct`, `extern alias`, detect `\e` escape for C#.
- Add `non-sealed` for Java.


C#/C#.sublime-syntax
```diff
@@ -3394,7 +3414,7 @@ variables:
   integer_suffix: '[uU][lL]?|[lL][uU]?'
 
   # characters
-  escape_base: \\[0abfnrtv"'\\]
```
mohamedniyaz0506, line 3397: #enhance:\\[$0&abfnrtv&0$"'\\]

C#/C#.sublime-syntax
```diff
@@ -431,6 +432,17 @@ contexts:
...
+    - match: (extern)\s+(alias)\b
+      captures:
+        1: storage.modifier.cs
+        2: keyword.declaration.alias.cs
+      push:
+        - match: '{{unreserved_name}}'
+          scope: variable.namespace.cs
+          pop: 1
+        - include: else-pop
+
```
deathaxe: 1. To make sure inherited syntaxes can access everything, named contexts are preferred.

2. A dedicated meta scope (e.g. `meta.alias.cs`) may help plugins to identify the context.

3. separate "sections" for each construct are preferred

```suggestion
###[ ALIAS DECLARATIONS ]######################################################

  extern-alias-declarations:
    - match: (extern)\s+(alias)\b
      captures:
        1: storage.modifier.cs
        2: keyword.declaration.alias.cs
      push: extern-alias-declaration-name

  extern-alias-declaration-name:
    - meta_scope: meta.alias.cs
    - match: '{{unreserved_name}}'
      scope: variable.namespace.cs
      pop: 1
    - include: else-pop

###[ USING DECLARATIONS ]######################################################
```

C#/C#.sublime-syntax
deathaxe: The list of documented valid escape sequences can be found at https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/#string-escape-sequences.

deathaxe (commented):
One suggestion to improve `extern alias` implementation in C#.

Otherwise LGTM!

C#/C#.sublime-syntax
mikomikotaishi: I concur with the change, I'll apply it

deathaxe (approved):

michaelblyons (approved):

mohamedniyaz0506 (approved):

PR #4606, opened 2026-08-30
[JavaScript] Add support for svg tagged template strings

deathaxe:
This commit assigns html syntax to svg tagged template strings.

grafik


michaelblyons (approved):

keith-hall (approved):
SVG is technically an XML dialect but I guess it makes sense to use the HTML syntax for the CSS etc

deathaxe:
Yes, it is XML, but not worth a separate syntax, which would need to be embedded everywhere, when it has to parse all the things html already does perfectly, well, just with a slightly more strict set of tag name rules. Furthermore current XML is not well suited for all the template syntaxes and the way they inject their template tags.

PR #4611, opened 2026-09-10
[Rust] Add basic reference indexing

BenjaminSchaaf:
This commit adds indexing for both function calls and macro invocations.


Rust/Indexed Reference List.tmPreferences
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>scope</key>
+    <string>
+        variable.function,
```
keith-hall: arguably the Default package's `Indexed Reference List.tmPreferences` already handles `variable.function` - the Rust syntax definition is missing the `meta.function-call` scope for it to apply.

BenjaminSchaaf: The syntax doesn't currently parse function calls, so adding that would be a much larger effort. I've fixed the scopes not being rust-specific.

dpjohnst (approved):
LGTM

keith-hall (approved):

